Issue #7, opened 2020-04-23
Enhance AST dump


NalaGinrut:
I'm working on an experimental branch to enhance `-frust-dump-parse` which is a little inspired by gccgo's AST dump.
For example:
```rust
 fn abc(x:u32, y:u32) -> u32 {
    return x+y;
 }
 
 fn main() {
     println!("Hello World!");
 }
```
The dumped result looks like this:
```
Crate: 
 inner attributes: none
 items: 
  
u32 abc(x : u32, y : u32)
 {
 BlockExpr: 
 outer attributes: none
 inner attributes: none
 statements: 
  ExprStmtWithoutBlock: return ArithmeticOrLogicalExpr: x + y
 final expression: none
 }

  
void main()
 {
 BlockExpr: 
 outer attributes: none
 inner attributes: none
 statements: 
  ExprStmtWithoutBlock: println!("Hello World!")
 final expression: none
 }
```

The branch is [nala/enhance/ast-dump](https://github.com/philberty/gccrs/tree/nala/enhance/ast-dump)

Comments are welcome.

philberty:
I really like this output it makes it more in like with how -fdump-tree-gimple looks also

SimplyTheOther:
There's two main comments I'd like to make here:

- Some of the non-dump-related code uses the as_string() method to convert some things into string form, such as paths (e.g. converting `something::something_else` to "something::something_else"). As such, if you convert all object dumping processes to output the same form, either the code requesting a string form would have to be changed (i.e. new method), or you might have to add a new method like `to_dump_string()` or something.

- The main issue that I experienced with creating a nice-looking dump is that indentation levels are pretty much impossible to create with the current system of recursive `as_string()` calls, as the called object has no conception of the indentation level of the parent, which prevents any indentation from being preserved if the called object adds a new line to the string. You'd probably have to pretty heavily modify the dumping mechanism to do this - the best solution that I can currently think of would be a separate "Dump Creator" object that holds state on the current indentation level and the string so far, and a theoretical `dump_string()` method could add new lines to that object, and then the final string could be extracted from it at the end. This would probably be a lot of work and may involve unforeseen issues, however. 

There's also one minor nitpick with the current example you have shown - strictly speaking, `BlockExpr` includes the curly braces surrounding it (as that is how BlockExpr is defined).

NalaGinrut:
@SimplyTheOther 
[...]

PR #8, opened 2020-04-28
Enhance AST dump

NalaGinrut:
See #7

### DIFF
--- a/gcc/rust/ast/rust-ast-full-test.cc
+++ b/gcc/rust/ast/rust-ast-full-test.cc
@@ -1,4596 +1,6309 @@
+/*
+   Copyright (C) 2009-2020 Free Software Foundation, Inc.
+
+This file is part of GCC.
+
+GCC is free software; you can redistribute it and/or modify it under
+the terms of the GNU General Public License as published by the Free
+Software Foundation; either version 3, or (at your option) any later
+version.
+
+GCC is distributed in the hope that it will be useful, but WITHOUT ANY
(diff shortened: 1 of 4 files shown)


philberty:
the review in github is too large to review. It would be ideal for future PR's to split up the work into smaller commits if possible and expand on your commit message a little bit more. I am extremely guilty of this but its a habit i want to get out of.

Can you supply an example rust program your resting with as part of creating the review.

Off topic: I will review this properly locally tomorrow still working on a PR to do some static analysis work which i will ask both of you to review also.

NalaGinrut:
@philberty I'm sorry but this PR is special since it mixed with the indentation fix. I understand they should be separated, but this work is before the topic of the indentation standard discussion. My idea is to merge it then fix the rest indentation with another PR, then we can start the better cooperation for code review. I think it's the cost for us in the very beginning cooperation. 

What do you think?
cc @SimplyTheOther

philberty:
Ah ok yeah that makes sense if this includes some of the indentation fix do you want to add in the rest of the indentation fix into this PR and we can merge it so its done or do  you want to get this out of the way first?

NalaGinrut:
@philberty The AST dump is not complete yet, but the mixed indentation issue may affect other contributors to make new things. So my idea is to merge it ASAP, and I'll raise a new PR for indentation ASAP. Then the world is back to normal. Alas... ;-)

philberty:
cool beans i merged it now :)

Issue #19, opened 2020-05-13
Perform basic compilation of a single function


philberty:
We should be able to compile a function with variables and parameters down to gimple

PR #20, opened 2020-05-13
This will now compile a void function declaration to gimple

philberty:
-fdump-tree-gimple should provide output of the skeleton function

Addresses #19


philberty:
This is really basic i have more code which will compile parameters and some of the code but this PR allows us to get this output in gimple:
```
philbert@philbert-XPS-13-7390 {~/workspace/gccrs-build} $ cat test.rs.005t.gimple 
main ()
{

}
```

PR #33, opened 2020-05-21
WIP more GIMPLE conversion

philberty:
This is the initial setup for the pass to convert to GIMPLE. On thing that needed done in type resolution was to find all locals per block as these must be setup up front for GIMPLE but this was easy enough with our visitor pattern.

Note the creation of a cscope.h (needs a better name) but something like this might be useful for our static analysis passes to be more unified.

### DIFF
--- a/gcc/rust/analysis/rust-type-resolution.cc
+++ b/gcc/rust/analysis/rust-type-resolution.cc
@@ -23,6 +23,7 @@ TypeResolution::TypeResolution (AST::Crate &crate, TopLevelScan &toplevel)
   : Resolution (crate, toplevel)
 {
   functionScope.Push ();
+  localsPerBlock.Push ();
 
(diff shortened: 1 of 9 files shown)


gcc/rust/analysis/rust-type-resolution.cc
```diff
@@ -670,11 +675,16 @@ TypeResolution::visit (AST::Function &function)
     }
 
   // walk the expression body
+  localsPerBlock.Push ();
   for (auto &stmt : function.function_body->statements)
     {
       stmt->accept_vis (*this);
     }
 
+  auto localMap = localsPerBlock.Pop ();
```
philberty, line 684: This seems to be a C++17 thing and gcc is warning about it. Not sure if we should allow it or now. Might be something to create an issue out of for what C++ version will we limit ourselves to.

PR #44, opened 2020-08-12
Add target hooks for basic cfg usage

SimplyTheOther:
This pull request adds some target hooks for the Rust frontend, which allow platform-specific information to be provided to the frontend.

Specifically, it adds many defines for use with the `cfg` attribute and macro, including ones for `target_arch`, `target_feature`, `target_os`, `target_vendor`, `target_family` and `target_env`, as well as providing information on pointer width and endianness. If you are not familiar with the `cfg` attribute, it is used for conditional compilation, and more information can be found on it [here](https://doc.rust-lang.org/reference/conditional-compilation.html).

Note that this is _not_ a finalised implementation of `cfg` attribute handling - other, somewhat less important defines, such as the ones related to the availability of atomic integers of various sizes, have not been added yet. However, I thought that this PR was already big enough and took long enough to code without taking the time to investigate GCC's atomic support, so I settled on just adding target defines for each platform supported by GCC.

I am so sorry about the size of this PR, but I thought it would be better that the defines worked on all platforms before merging, and I grossly underestimated the amount of platforms supported by GCC. I'm not sure if it would be possible to break this up into smaller PRs without significant work because I made subsequent improvements to earlier target hooks after learning more about GCC's internals as I went along.


philberty:
Thanks for the PR I will get to look at this on Saturday I am getting
married tomorrow so I am on vacation for a dew days.

SimplyTheOther:
Congratulations! I hope you have a great wedding and vacation. Wishing you both the best.

PR #45, opened 2020-08-27
Lexer cleanup

SimplyTheOther:
This PR improves code quality in the lexer. 
Notably, the giant `build_token` method had many aspects split off into separate methods. 
I also fixed some location tracking issues in the lexer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,4 @@ REVISION
 test.code-workspace
 
 gcc/rust/test3-tiny/*
+.clang-format.swap
--- a/gcc/rust/lex/rust-codepoint.h
+++ b/gcc/rust/lex/rust-codepoint.h
@@ -1,11 +1,6 @@
 #ifndef RUST_CODEPOINT_H
 #define RUST_CODEPOINT_H
 
-#include "config.h"
-#include "system.h"
-#include "coretypes.h"
-// config, system, coretypes - TODO: ensure all are needed
-
 #include <string>
 
 namespace Rust {
@@ -16,11 +11,11 @@ struct Codepoint
   // Creates a zero codepoint.
   Codepoint () : value (0) {}
 
-  // Creates a codepoint from UTF-8 value.
-  Codepoint (uint32_t value_) : value (value_) {}
+  // Creates a codepoint from an encoded UTF-8 value.
+  Codepoint (uint32_t value) : value (value) {}
 
-  // Returns a C++ string containing value of codepoint.
-  ::std::string as_string ();
+  // Returns a C++ string containing string value of codepoint.
+  std::string as_string ();
 };
 } // namespace Rust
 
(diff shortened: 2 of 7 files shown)


bjorn3:
Would it be feasible to do a 1-to-1 translation of librustc_lexer? The main file https://github.com/rust-lang/rust/blob/118860a7e76daaac3564c7655d46ac65a14fc612/src/librustc_lexer/src/lib.rs is 810 lines long. This would ensure full compatibility between rustc and gccrs on the lexer side and if there are any changes to the rustc lexer those would be very easy to port to gccrs. I don't know if that is acceptable from a license point of view though.

SimplyTheOther:
A *complete* 1-1 translation? My understanding is no. librustc_lexer makes heavy use of tagged unions (“enum” in Rust - things like TokenKind in that file), which are not supported in C++11. There are workarounds for this, e.g. inheritance, but it would reduce performance (due to virtual function call) and also no longer be a 1-1 translation.

There’s also the issue of differences in how gccrs stores locations (rustc uses a “span”), and also the token format expected by each parser. Also, in my opinion, the lexer is probably the least complex part of the compiler (and the easiest to update), so I don’t think we’d get many advantages by using a 1-1 conversion of rustc’s.

There’s also the issue that rustc’s lexer system is a bit more complex than that. There is a second lexer [here](https://github.com/rust-lang/rust/blob/118860a7e76daaac3564c7655d46ac65a14fc612/src/librustc_parse/lexer/mod.rs), which apparently produces “rich tokens” or something. To get full functionality of their lexing, we’d need to copy that too.

bjorn3:
> My understanding is no. librustc_lexer makes heavy use of tagged unions (“enum” in Rust - things like TokenKind in that file), which are not supported in C++11.

```c++
enum Base {
    Binary,
    Octal,
    Hexadecimal,
    Decimal,
};

struct LiteralKind {
    enum LiteralKindVariant {
        Int,
        Float,
    } kind;
    union LiteralKindData {
        struct Int { Base base; bool empty_int; } Int;
        struct Float { Base base; bool empty_exponent; } Float;
    } data;
};
```

works with `-std=c++11` with gcc 10.2.



> Also, in my opinion, the lexer is probably the least complex part of the compiler (and the easiest to update), so I don’t think we’d get many advantages by using a 1-1 conversion of rustc’s.

I see

SimplyTheOther:
``` 
    union LiteralKindData {
        struct Int { Base base; bool empty_int; } Int;
        struct Float { Base base; bool empty_exponent; } Float;
    } data;
```

I was referring to type-safe tagged unions, i.e. something like ``std::variant`` in C++14. It would be very easy to accidentally use this kind of union incorrectly if you forgot to check the tag, or assumed that it should be a certain element, etc. ``union``s themselves actually have several other issues in addition to not being type-safe - they don't work with non-trivially constructible inner types unless you do hacks like placement new, you have to explicitly call destructors if you ever change the active element, and they're just a pain in general.

philberty (approved):

philberty:
I agree with @SimplyTheOther I don't believe a 1-1 copy of the rust code is going to help us

PR #46, opened 2020-10-01
Parser cleanup, enhancements and conversion to template

SimplyTheOther:
This PR performs some cleanup and enhancement of the parser, as well as converting it to a template.

Many of the expression parsing methods were consolidated (e.g. parsing addition expressions and subtraction expressions are now done by one method instead of several), and others were modified to reduce error incidence by having "final token" inputs. The right angle brackets at the end of generics should also be disambiguated in a better way now - token "splitting" instead of just "replacing". Also, match expression parsing and also AST storage/modelling of it was improved to be closer to the spec, faster, and simpler.

It also converts the parser to a template class (over a "ManagedTokenSource" type rather than the Lexer class in particular) in anticipation of the macro expansion code, which will require the parser to operate on tokens from a token stream (rather than tokens lexed from characters in the lexer) - I considered templating a better option than using a polymorphic "ManagedTokenSource" (which would be too slow) or misusing the Lexer (which would be dodgy and buggy). As the parser is a template class, its implementation had to be moved to a header file.


philberty:
This is looking good but just wondering is it worth potentially squashing some of the cleanup/debug commits?

philberty:
If its a pain i will just merge soon

gcc/rust/lex/rust-lex.h
```diff
@@ -9,6 +9,36 @@
 #include <tuple>
 
 namespace Rust {
+// Simple wrapper for FILE* that simplifies destruction.
+struct RAIIFile
+{
```
philberty, line 14: is it worth pulling this class out into its own file? It could be handy later for reading/exporting information in the object files when we start supporting imports

gcc/rust/ast/rust-stmt.h
```diff
@@ -204,28 +173,21 @@ class ExprStmtWithoutBlock : public ExprStmt
 class ExprStmtWithBlock : public ExprStmt
 {
 public:
-  // ExprWithBlock* expr;
-  ::std::unique_ptr<ExprWithBlock> expr;
-
-  /*~ExpressionStatementWithBlock() {
-      delete expr;
-  }*/
+  std::unique_ptr<ExprWithBlock> expr;
```
philberty, line 176: I know i have added these as public fields for the passes which seems kind of messy but i cant think of anything better right now. I think the answer will become clear over time i think we should raise an issue about them.

SimplyTheOther:
> This is looking good but just wondering is it worth potentially squashing some of the cleanup/debug commits?

I have squashed the debug commits. Unfortunately, this meant that I had to commit the work I had started after this PR before rebasing, which seems to have been incorporated into this PR. I have not tested this new code and it may not compile (though behaviour shouldn't change).

gcc/rust/lex/rust-lex.h
SimplyTheOther: Possibly. I could do so now if you wanted, but it's basically self-contained so it could be easily done at a future date.

gcc/rust/ast/rust-stmt.h
SimplyTheOther: I agree. Other ASTs using this inheritance model (e.g. clang) seem to have getters and setters, though they also don't seem to manage memory in the same way.

philberty:
Sounds good thanks for that. I will merge when travis says it builds.

PR #61, opened 2020-12-04
Phil/build changes

philberty:
This updates the README and removes C/C++ from the enabled languages as simply building rust seems to work since the rebase against GCC

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ this is why a build directory is created.
 ```bash
 $ mkdir gccrs-build
 $ cd gccrs-build
-$ ../gccrs/configure --prefix=$HOME/gccrs-install --disable-bootstrap --enable-multilib --enable-languages=c,c++,rust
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -27,14 +27,14 @@ this is why a build directory is created.
 ```bash
 $ mkdir gccrs-build
 $ cd gccrs-build
-$ ../gccrs/configure --prefix=$HOME/gccrs-install --disable-bootstrap --enable-multilib --enable-languages=c,c++,rust
+$ ../gccrs/configure --prefix=$HOME/gccrs-install --disable-bootstrap --enable-multilib --enable-languages=rust
 $ make
 ```
 
 Running the compiler itself without make install we can simply imvoke the compiler proper:
```
bjorn3, line 34: typo:
```suggestion
Running the compiler itself without make install we can simply invoke the compiler proper:
```
(you might want to manually commit it as this change is small enough to not require licensing)

philberty: thanks good find :+1:

Issue #17, opened 2020-05-05
Macro expansion


NalaGinrut:
The macro expansion happens before the resolution.

This issue is opened for the record of pipeline in kanban.

NalaGinrut:
Here're related RFCs:
- [1584-macros](https://github.com/rust-lang/rfcs/blob/master/text/1584-macros.md) This is the macros 2.0
- [2361-dbg-macro](https://github.com/rust-lang/rfcs/blob/master/text/2361-dbg-macro.md) 
[0378-expr-macros](https://github.com/rust-lang/rfcs/blob/master/text/0378-expr-macros.md) It seems the nightly haven't implemented it yet,  But I'll open an issue for it.

NalaGinrut:
The current parser will return the `MacroItem` as `unique_ptr`, so that we lose the chance to store the macro definition into a table for expansion step. It's too slow if we query its definition by traversing the AST since the AST is unnecessarily balanced.

If we just store its `unique_ptr` into the table and don't put `MacroItem` on AST tree, then it can be a solution. But we will lose the chance to dump the macro definition in AST dumper.

So I think we may have to maintain 2 copies of each macro definition. This is not the best way, but could be optimized in the future refactoring.

Any better idea? 
@philberty @SimplyTheOther

SimplyTheOther:
Wouldn’t you be able to store a reference to the MacroItem (or rather a reference to a unique_ptr containing the MacroItem) in the table?

Alternatively, you could parse the macro definition into a different form during the expansion phase and put the results of that into the table, and maybe even remove the macro definition item from the constructed AST after that stage if it isn’t required anymore.

NalaGinrut:
> Wouldn’t you be able to store a reference to the MacroItem (or rather a reference to a unique_ptr containing the MacroItem) in the table?

Yes, I can, but sounds not safe to store the unwrapped reference into the table, since it's not unique anymore.

> Alternatively, you could parse the macro definition into a different form during the expansion phase and put the results of that into the table, and maybe even remove the macro definition item from the constructed AST after that stage if it isn’t required anymore.

I think you're suggesting a pass to eliminate dead macros, yes I can do that.
But I don't think it's related to my question, my question is about how to query it quickly. The elimination has to rely on it.

> Or just move the MacroItem unique_ptr into the table while in the expansion phase, so that the MacroItem is still in the AST during the AST dump phase and so remains in the dump.

[...]

PR #62, opened 2020-12-14
Expansion code for cfg attributes, bug fixes and minor enhancements in parser

### LINKED ISSUE #42: StructFieldExpr are missing location info
Currently initializing struct members in StructExprFieldIndexValue and StructExprFieldIdentifierValue don't have any location info.

### LINKED ISSUE #52: Parser failed to handle expressions within index portion ArrayIndexExpr
```test.rs:8:22: error: expecting ‘]’ but ‘+’ found
    8 |     let c = xs[index + 1];
      |                      ^
left denotation is null; returning null for parse_expr
```

### LINKED ISSUE #53: Parser failed to handle CopiedArray
test.rs:8:36: error: expecting ‘;’ but ‘]’ found                                                                           
    8 |     let mut array: [i32; 3] = [0; 3];                                                                              
      |                                    ^                 
test.rs:8:5: error: failed to parse statement or expression without block in block expression
    8 |     let mut array: [i32; 3] = [0; 3];                                                                              
      |     ^                                                                                                              
test.rs:8:36: error: unrecognised token ‘]’ for start of item                                                              
    8 |     let mut array: [i32; 3] = [0; 3];                                                                              
      |                                    ^                                                                               
test.rs:8:36: error: failed to parse item in crate

SimplyTheOther:
This PR adds the code used to expand the AST to allow the proper handling of the `cfg` conditional compilation attribute, as well as related attributes such as `cfg_attr` and theoretically the `cfg!` macro as well. This will ensure that allowed aspects of Rust, e.g. functions, can be stripped based on these attributes. This is one aspect of the "macro expansion" phase (#17), and is probably the first one that should be executed (as name resolution is not possible before this aspect). 

Additionally, there are also bug fixes and parser enhancements. #53 (can't handle copied array expression), #52 (can't parse some expressions inside array index expression), and #42 (no locations in struct expression fields) should all be fixed. Additionally, the use tree parsing was enhanced to be closer to the required behaviour. 

There were also some minor changes to the AST - fields/member variables that were made public for access in the resolution stages have been made private again, and now use getter functions. This decision was made to allow changing the representation of the variables later (e.g. moving variables into parent or child classes).
Also, the copy constructors and copy assignment operators of many AST classes were made safer by adding checks for null pointers before attempting to dereference them (when cloning nodes).

Closing keywords for issues:
Fixes #53, fixes #52, resolves #42.


philberty:
nice work!

gcc/rust/analysis/rust-scan.cc
```diff
@@ -353,7 +353,7 @@ TopLevelScan::visit (AST::UseDeclaration &use_decl)
 void
 TopLevelScan::visit (AST::Function &function)
 {
-  functions[function.function_name] = &function;
```
philberty, line 356: thanks for doing this my branch to move to HIR is updating the pipeline with a new name/type resolution algorithm so we wont need those extra fields on LetStmt and Function inside the AST which i will remove with that PR

PR #64, opened 2020-12-16
HIR and new TyTy classes

philberty:
Apologies for this big PR but i think it needed to go all together to make sure it all fits. I am trying to follow rustc a bit more now i think its going to solve alot of problems down the road.

HIR classes are based off the AST classes for now eventually this should remove the macro classes once the expand pass is done.

We should also reorder the pipeline a bit, a new name resolution pass should be added and the name resolution pass here should be called name resolution late.

The type resolution converts all types over to a new TyTy module similar to rust this allows for the combination of types the type resolution before was never going to work for cases like:

let mut x;
x = None
x = 1;

there for x is an option or i32.

Or

let mut x = Vec<_>::new()
x.push(1)

therefore x is a vec or i32.

This is going to vastly similify the GIMPLE transformation. My plan so far is that given MIR is a control flow graph it maps fairly close to GIMPLE maybe we have enough info with it to avoid adding MIR down the line.


gcc/rust/resolve/rust-ast-resolve.cc
```diff
@@ -0,0 +1,252 @@
...
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#include "rust-ast-resolve.h"
+#include "rust-ast-resolve-toplevel.h"
+#include "rust-ast-resolve-item.h"
+#include "rust-ast-full.h"
+#include "rust-tyty.h"
+
+#define MKBUILTIN_TYPE(_X, _R, _TY)                                            \
```
SimplyTheOther, line 25: This is more of a stylistic choice than anything else, but making this a template member function (if required - otherwise just a regular member function) might be better than having it as a macro. I believe that doing so would be possible as long as `_R` is passed in by reference - no other modification of parameters seems to occur.

gcc/rust/resolve/rust-ast-resolve.cc
```diff
@@ -0,0 +1,252 @@
...
+
+  *rib = it->second;
+  return true;
+}
+
+void
+Resolver::insert_builtin_types (Rib *r)
+{
+  auto builtins = get_builtin_types ();
+  for (auto it = builtins.begin (); it != builtins.end (); it++)
```
SimplyTheOther, line 109: I believe a range-based for loop would be better here for brevity.

gcc/rust/resolve/rust-ast-resolve.cc
```diff
@@ -0,0 +1,252 @@
...
+{
+  // setup parent scoping for names
+  resolver->get_name_scope ().push (crate.get_node_id ());
+  resolver->push_new_name_rib (resolver->get_name_scope ().peek ());
+  // setup parent scoping for new types
+  resolver->get_type_scope ().push (mappings->get_next_node_id ());
+  resolver->push_new_type_rib (resolver->get_type_scope ().peek ());
+
+  // first gather the top-level namespace names then we drill down
+  for (auto it = crate.items.begin (); it != crate.items.end (); it++)
```
SimplyTheOther, line 243: A range-based for loop would also be more concise here and at line 247 below.

gcc/rust/hir/tree/rust-hir.h
```diff
@@ -0,0 +1,1467 @@
...
+
+  virtual void accept_vis (HIRVisitor &vis) = 0;
+
+protected:
+  // pure virtual clone implementation
+  virtual MacroMatch *clone_macro_match_impl () const = 0;
+};
+
+// A token is a kind of token tree (except delimiter tokens)
+class Token : public TokenTree, public MacroMatch
```
SimplyTheOther, line 117: Assuming that (unexpanded) attributes and macros don't exist in the HIR, tokens shouldn't be required either, I don't think. The AST `Token` is only used for storing the token data in macro invocations or attributes. Maybe a note of this could be made or something instead of actual deletion at this point, just in case it turns out that tokens are in fact needed.

gcc/rust/hir/tree/rust-hir-item.h
```diff
@@ -0,0 +1,3259 @@
...
+	     std::unique_ptr<Type> type = nullptr,
+	     Attribute outer_attr = Attribute::create_empty ())
+    : outer_attr (std::move (outer_attr)),
+      type_representation (std::move (type_representation)),
+      type_param_bounds (std::move (type_param_bounds)),
+      type (std::move (type)), locus (locus)
+  {}
+
+  // Copy constructor uses clone
+  TypeParam (TypeParam const &other)
```
SimplyTheOther, line 73: Again, the modifications to copy constructors that have been made in the AST to prevent null pointer dereferences (having an if statement guard on whether the `other` variable is null before calling the clone method) should be applied to here, as well as other applicable classes in the HIR.

gcc/rust/hir/tree/rust-hir-macro.h
```diff
@@ -0,0 +1,628 @@
...
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#ifndef RUST_HIR_MACRO_H
+#define RUST_HIR_MACRO_H
```
SimplyTheOther, line 20: If macros are all expanded in the HIR, then this file should probably not exist. However, just in case they are needed, I would not remove the code here yet.

SimplyTheOther (approved):
This looks great. I had a couple of nitpicks (which I included as inline comments), but nothing too major.

gcc/rust/backend/rust-compile-context.h
```diff
@@ -0,0 +1,218 @@
...
+{
+public:
+  Context (::Backend *backend)
+    : backend (backend), resolver (Resolver::Resolver::get ()),
+      tyctx (Resolver::TypeCheckContext::get ()),
+      mappings (Analysis::Mappings::get ())
+  {
+    // insert the builtins
+    auto builtins = resolver->get_builtin_types ();
+    for (auto it = builtins.begin (); it != builtins.end (); it++)
```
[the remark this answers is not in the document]
philberty: thanks yeah i should be more consistent i think my brain keeps going between C++11 and old times :)

gcc/rust/hir/tree/rust-hir-cond-compilation.h
```diff
@@ -0,0 +1,248 @@
+// Copyright (C) 2020 Free Software Foundation, Inc.
```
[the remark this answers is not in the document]
philberty: Thanks i will make an issue to remove these. I basically copied the AST as the beginning point for HIR

Issue #12, opened 2020-05-01
Basic Type inferencing of code within the same file


philberty:
We need to create a method to do name resolution and type basic type inferencing

PR #71, opened 2020-12-18
Add type unification as part of typecheck.

philberty:
Rust must examine each usage of a name and unify their types. For example:

let mut x;
x = 1

This means the declaration is determined to be an inference variable then
the assignment can be resolved to an Integer TyTy which can be combined
as part of the rules to now make the let x decl be an i32.

#12


SimplyTheOther (approved):
PR looks great.

PR #74, opened 2020-12-18
When compiling a block it might reference a GIMPLE node that is not

### LINKED ISSUE #72: HIR to GIMPLE crash when compiling forward decl
When we are processing the HIR into gimple there is a crash when we call a function that has not been compiled yet.

```fn main() {
    let mut an_integer = 5;
    an_integer = test(1) + 3;

    let call_test = test(1);
}

fn test(x: i32) -> i32 {
    return x + 1;
}
```

The resolution and type resolution correctly setup all mappings required to compile the main function and then compile the test function on request but there is a core in this code.

philberty:
resolved yet. Such as:

fn main() -> i32 { call() }
fn call() -> i32 { return 1; }

The compilation pass acts on the first function main and detects the other
node call is not compiled yet we can go resolve it. This is a toplevel
item since it has a local_def_id and therefore it has a NULL enclosing
scope for the block.

Fixes #72

### DIFF
--- a/gcc/rust/backend/rust-compile-item.h
+++ b/gcc/rust/backend/rust-compile-item.h
@@ -48,11 +48,7 @@ public:
       {
 	// has this been added to the list then it must be finished
 	if (ctx->function_completed (lookup))
-	  {
-	    printf ("returning early the function [%s] is completed!\n",
-		    function.as_string ().c_str ());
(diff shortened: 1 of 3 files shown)


SimplyTheOther (approved):
PR seems great to me.

PR #75, opened 2020-12-21
This brings arrays back into the new framework. It resolves ArrayType

philberty:
ArrayExpr, ArrayExprElems and ArrayIndexExpr. Still need to do
ArrayElemsCopied.

I expect there to be some changes to cleanup the rust-tyty-resolver
this code is to resolve all ribs within a scope but its getting a bit
hairy in there now.

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-type.h
+++ b/gcc/rust/typecheck/rust-hir-type-check-type.h
@@ -25,6 +25,42 @@
 namespace Rust {
 namespace Resolver {
 
+class ArrayCapacityConstant : public TypeCheckBase
+{
+public:
+  static bool fold (HIR::Expr *expr, size_t *folded_result)
+  {
+    ArrayCapacityConstant folder;
+    expr->accept_vis (folder);
+    *folded_result = folder.result;
+    return folder.ok;
+  }
+
+  virtual ~ArrayCapacityConstant () {}
+
+  void visit (HIR::LiteralExpr &expr)
+  {
+    switch (expr.get_lit_type ())
+      {
+	case HIR::Literal::LitType::INT: {
+	  ok = true;
+	  std::stringstream ss (expr.as_string ());
+	  ss >> result;
+	}
+	break;
+
+      default:
+	return;
+      }
+  }
+
+private:
+  ArrayCapacityConstant () : TypeCheckBase (), ok (false), result (-1) {}
+
+  bool ok;
+  size_t result;
+}; // namespace Resolver
+
 class TypeCheckType : public TypeCheckBase
 {
 public:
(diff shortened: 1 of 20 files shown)


gcc/rust/typecheck/rust-hir-type-check-type.h
```diff
@@ -25,6 +25,42 @@
 namespace Rust {
 namespace Resolver {
 
+class ArrayCapacityConstant : public TypeCheckBase
```
philberty, line 28: this will need to change when we add in size types. It seems rust needs a constant when declaring an array type.

gcc/rust/typecheck/rust-tyty-resolver.h
```diff
@@ -63,9 +64,25 @@ class TyTyResolver
 	  TyTy::TyBase *resolved = nullptr;
 	  if (!context->lookup_type (hir_node_ref, &resolved))
 	    {
-	      rust_fatal_error (mappings->lookup_location (hir_node_ref),
-				"failed to lookup type for reference");
-	      return false;
+	      // this could be an array/adt type
```
philberty, line 67: this pice of code will get abstracted more with tuples it a bit hacky at the moment. Its the code responsible to handling the case of:

let mut x;
x = 1;

so inference variables get resolved correctly.

gcc/rust/typecheck/rust-hir-type-check-type.h
SimplyTheOther: By constant do you mean a constant expression (i.e. an expression that evaluates to a constant, like C++ constexpr) or a constant itself? Because I believe the requirement is for a constant expression, meaning that, for example, a call to a constant function could be listed as the array size. 
This is at least what the spec page [here](https://doc.rust-lang.org/reference/types/array.html) says.

SimplyTheOther (approved):
PR looks great. A bit confused about one of the comments made about future stuff, but code itself looks fine.

gcc/rust/typecheck/rust-hir-type-check-type.h
philberty: That is interesting I was testing this with the rust compiler:

```
fn main() {
    let x = 3;
    let arr: [i32; x];
}
```

But this errors with:

```
error[E0435]: attempt to use a non-constant value in a constant
 --> test.rs:3:20
  |
3 |     let arr: [i32; x];
  |                    ^ non-constant value

For more information about this error, try `rustc --explain E0435````

philberty: I think grammar wise its probably an expression for the capacity but it wants a constant for the static analysis.

PR #94, opened 2020-12-22
Shadowing rules are done as part of name resolution.

### LINKED ISSUE #82: Shadowing rules
Shadowing of declarations need to be respected for core data structures such as:

let mut x;
x = 4;
let mut x;
x = true;

philberty:
When a new name is defined the name resolver freezes the previous
declartion such that all new references point to the latest decl.

This patch fixes a crash when we shadow and get a type mismatch and the
combination of types fails. See the failure test case for the crash.

Fixes #82


SimplyTheOther (approved):
PR looks great.

PR #96, opened 2020-12-23
Add missing license text

### LINKED ISSUE #66: Missing the gpl headers in some files
We are missing the GPL info at the top of some files.

unseddd:
Add GPL license text to files without the header

Resolves #66

### DIFF
--- a/gcc/rust/ast/clone-test.h
+++ b/gcc/rust/ast/clone-test.h
@@ -1,3 +1,21 @@
+// Copyright (C) 2020 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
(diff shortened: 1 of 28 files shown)


gcc/rust/ast/clone-test.h
```diff
@@ -1,3 +1,21 @@
...
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
 #ifndef CLONE_TEST_H
```
SimplyTheOther, line 19: This file should probably be removed. It isn't actually used in the current codebase for anything - it was a failed test in seeing whether I could get clone to work automatically without the boilerplate (of having to manually define copy constructors and whatever) that currently exists.

gcc/rust/ast/rust-ast-containers.h
```diff
@@ -1,3 +1,21 @@
...
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
 #ifndef RUST_AST_CONTAINERS_H
```
SimplyTheOther, line 19: This file is similar to the last one in that it should also be removed, for similar reasons. This was also a failed "first attempt" at an AST layout that should be unused in the current codebase.

SimplyTheOther (approved):
PR looks fine to me.

PR #100, opened 2020-12-23
Remove impl_block test. This is part of the next milestone and we want

philberty:
to turn on CI build failure if any test fails and this will block that
change until the next milestone if not removed.


bjorn3:
Rustc has several test cases that ICE. Instead of removing them, they are marked as compile-fail so when they are fixed, CI will fail. This would then require you to mark it as build-pass/run-pass, ensuring that it doesn't regress in the future. In addition [glacier](https://github.com/rust-lang/glacier) has a big list of reported ICE's that are checked once a day to see if they are fixed yet.

philberty:
I could move this test to be part of the failure tests is actually not a bad idea.

PR #117, opened 2021-01-04
This brings structs back in post HIR changes. 

philberty:
It supports structs where no base struct is referenced and the constructor is in order.

This is a big change to get the basic ADT type implemented. More smaller changes are to come in subsequent PRs to support tuples and base struct references.

With this change the test suite should be fully passing. Such that we can start failing builds if failing tests are popping up in PRs.


SimplyTheOther (approved):
PR looks great.

PR #126, opened 2021-01-06
Respect the f32 and f64 suffix on literals

philberty:
Rust is permissive for integers being marked as floats so the check in the lexer can be removed here.

### DIFF
--- a/gcc/rust/lex/rust-lex.cc
+++ b/gcc/rust/lex/rust-lex.cc
@@ -1938,16 +1938,6 @@ Lexer::parse_decimal_int_or_float (Location loc)
       PrimitiveCoreType type_hint = type_suffix_pair.first;
       length += type_suffix_pair.second;
 
-      if (type_hint == CORETYPE_F32 || type_hint == CORETYPE_F64)
-	{
-	  rust_error_at (
-	    get_current_location (),
-	    "invalid type suffix %qs for integer (decimal) literal",
-	    get_type_hint_string (type_hint));
-	  // ignore invalid type suffix as everything else seems fine
-	  type_hint = CORETYPE_UNKNOWN;
-	}
-
       current_column += length;
 
       str.shrink_to_fit ();
--- a/gcc/rust/backend/rust-compile-expr.h
+++ b/gcc/rust/backend/rust-compile-expr.h
@@ -156,8 +156,6 @@ public:
 	return;
 
 	case HIR::Literal::FLOAT: {
-	  printf ("FLOATY BOYO: [%s]\n", expr.as_string ().c_str ());
-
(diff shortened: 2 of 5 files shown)


gcc/rust/lex/rust-lex.cc
```diff
@@ -1938,16 +1938,6 @@ Lexer::parse_decimal_int_or_float (Location loc)
...
-	    get_current_location (),
-	    "invalid type suffix %qs for integer (decimal) literal",
-	    get_type_hint_string (type_hint));
-	  // ignore invalid type suffix as everything else seems fine
-	  type_hint = CORETYPE_UNKNOWN;
-	}
-
       current_column += length;
 
       str.shrink_to_fit ();
```
SimplyTheOther, line 1943: The code below here (the `make_int`) would be incorrect (assuming that an integer with a type hint of a float type is treated as a float). What should happen instead is that `make_float` is called if the type hint is 'f32' or 'f64', while 'make_int' should be called otherwise. This is assuming that I have understood the [spec](https://doc.rust-lang.org/reference/tokens.html#floating-point-literals) properly. 

I suppose it depends on whether rust is actually permissive on integers being marked as floats (i.e. allows integers to be given a float type hint) or whether the float type hint means that the integer is treated as a float.

If my interpretation is wrong, feel free to ignore this.

SimplyTheOther (approved):
PR looks great. I had one query about whether integers given a float type hint are actually treated as floats, or whether they are treated as integers with the type hint being ignored (or complained about during type check or whatever).

gcc/rust/lex/rust-lex.cc
philberty: Yeah this was not clear from the spec to me either. From testing the rust compiler i found this is permissive:

```
fn main() {
    let a: i32 = 5i32;
    let b: f32 = 5f32;
}
```

Without this change we determine:

```
float1.rs:3:18: error: invalid type suffix ‘f32’ for integer (decimal) literal
    3 |     let b: f32 = 5f32;
      |   
```

philberty: What if i split this change into two parts one where we support f32 and f64 suffix but remove this integer support for float suffix as a separate PR

bjorn3: > Yeah this was not clear from the spec to me either.

It may be a good idea to document this kind of things in the reference (https://github.com/rust-lang/reference/), so other alternative implementations don't have to discover it again.

philberty: I've created https://github.com/rust-lang/reference/pull/922

PR #127, opened 2021-01-06
Bring in a clang format linter as part of CI

philberty:
Lets see if it works


philberty:
Not sure what the best way to do this yet is will close this PR and investigate more

PR #128, opened 2021-01-06
Add test to cover handling hex, binary and octal number literals.

philberty:
This also ensure the type suffix is respected against the number.


SimplyTheOther (approved):
PR looks great.

PR #131, opened 2021-01-07
Implicit Returns support.

philberty:
For implict returns we must consider cases with a block having multiple
returns:

 HIR::BlockExpr Stmts {
   ...
   return x
 }
 HIR::BlockExpr final_expression {
   x + 1
 }

Although the code above is bad this is valid rust code and the rust
compiler correctly identifies the final_expression as unreachable.

This dead code eliminiation is done as part of AST to HIR lowering.
Type resolution examines all blocks to identifiy if they terminate
a function with a return/final expression it must correspond accordngly.

If the block is the final block the resulting termination of the block
must match the return type of the function, else the block must conform
to unit type.

### DIFF
--- a/gcc/rust/ast/rust-stmt.h
+++ b/gcc/rust/ast/rust-stmt.h
@@ -184,6 +184,8 @@ class ExprStmt : public Stmt
   Location locus;
 
 public:
+  Location get_locus_slow () const final override { return get_locus (); }
+
   Location get_locus () const { return locus; }
 
 protected:
--- a/gcc/rust/ast/rust-expr.h
+++ b/gcc/rust/ast/rust-expr.h
@@ -2838,6 +2838,8 @@ public:
       }
   }
 
+  size_t num_statements () const { return statements.size (); }
+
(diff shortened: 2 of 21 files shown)


gcc/rust/ast/rust-stmt.h
```diff
@@ -184,6 +184,8 @@ class ExprStmt : public Stmt
   Location locus;
 
 public:
+  virtual Location get_locus_slow () const override { return get_locus (); }
```
SimplyTheOther, line 187: Pretty pedantic, but this may optimise better with a `const final override` qualifier, since `get_locus_slow()` won't be overrided again at any later point. Also, since it isn't a new virtual method (it overrides the `Stmt` one), it doesn't need the `virtual` qualifier at the start.

SimplyTheOther (approved):
PR looks great. Had one minor nitpick, but otherwise completely fine.

gcc/rust/ast/rust-stmt.h
philberty: good point i will fix it before merge

PR #132, opened 2021-01-08
Make TyTyVisitor a pure abstract class

philberty:
This is a code cleanup as there are some inconsistencies with type inferencing rules and making the class force implementing the visitors will help fix these later on.


SimplyTheOther (approved):
PR looks great.

PR #133, opened 2021-01-08
Ensure scoping rules are enforced correctly for BlockExpr's

### LINKED ISSUE #80: Scoping rules
We should respect the basic lexical scoping of rust code. This will need extended later for lifetimes/generics etc.

philberty:
There was a bug with LetStmts where we name resolved the identifier
pattern first before the init expression. This lead to a situation
where shadowing rules were broken.

Example:

let x = 1;
let x = x + 1;

In this example the code was referencing the 2nd LetStmt as the resolved
name for the Identifier reference.

Fixes #80


SimplyTheOther (approved):
PR looks great.

PR #134, opened 2021-01-08
Functions with parameters must receive their own scoping Rib

philberty:
This means that parameters are scoped and shadowed correctly. Currently the resolver is treating all parameters are shadowing each other if they have the same name which is incorrect.

#80


SimplyTheOther (approved):
PR looks great.

PR #135, opened 2021-01-09
Add in a check for unused decls within ribs.

philberty:
This lead to cleanup of the name resolver as the usage of mappings means that in a given rib if there are no references to a decl NodeId that means it was not used.

To get on par with the offical rust compiler it should be allowed to have a reference where the name was assigned but not used but this might be a separate pass.

This gives us warnings for example:

```
test.rs:10:9: warning: unused name
   10 |     let a = 1;
      |         ^
test.rs:11:9: warning: unused name
   11 |     let b = test();
      |         ^
test.rs:5:1: warning: unused name
    5 | fn unused() -> i32 {
      | ^
```

test

```
fn test() -> i32 {
    1
}

fn unused() -> i32 {
    2
}

fn main() {
    let a = 1;
    let b = test();
}
```

The check is very generic since it is based on node id mappings so it should work as more rust is supported over time.

### DIFF
--- a/gcc/rust/resolve/rust-ast-resolve-unused.h
+++ b/gcc/rust/resolve/rust-ast-resolve-unused.h
@@ -1,0 +1,45 @@
+// Copyright (C) 2020 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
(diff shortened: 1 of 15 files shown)


gcc/rust/resolve/rust-ast-resolve-unused.h
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Rust {
+namespace Resolver {
+
+class ScanUnused : public ResolverBase
+{
+public:
+  static void Scan (Rib *r)
+  {
+    r->iterate_decls ([&] (NodeId decl_node_id, Location locus) -> bool {
```
philberty, line 32: Al thought this could be part of the Rib class, this is a diagnostic pass to look for unused which will need some extra work to check for assigned but unused later.

SimplyTheOther (approved):
PR looks great.

PR #137, opened 2021-01-10
Mark all functions as DECL_PUBLIC such that they can be linked against.

### LINKED ISSUE #136: Link error for main on Linux x86
this is likely the same bug as #36 but lets track it separately for now.

```
started lowering AST
finished lowering AST
started type resolution
finished type resolution
COLLECT_GCC_OPTIONS='-fsplit-stack' '-v' '-g' '-O0' '-o' 'test' '-shared-libgcc' '-mtune=generic' '-march=x86-64'
 as -v --64 -o /tmp/cchYC833.o /tmp/ccPPyd64.s
GNU assembler version 2.34 (x86_64-linux-gnu) using BFD version (GNU Binutils for Ubuntu) 2.34
COMPILER_PATH=/opt/gccrs-dev/libexec/gcc/x86_64-pc-linux-gnu/11.0.0/:/opt/gccrs-dev/libexec/gcc/x86_64-pc-linux-gnu/11.0.0/:/opt/gccrs-dev/libexec/gcc/x86_64-pc-linux-gnu/:/opt/gccrs-dev/lib/gcc/x86_64-pc-linux-gnu/11.0.0/:/opt/gccrs-dev/lib/gcc/x86_64-pc-linux-gnu/
LIBRARY_PATH=/opt/gccrs-dev/lib/gcc/x86_64-pc-linux-gnu/11.0.0/:/opt/gccrs-dev/lib/gcc/x86_64-pc-linux-gnu/11.0.0/../../../../lib64/:/lib/x86_64-linux-gnu/:/lib/../lib64/:/usr/lib/x86_64-linux-gnu/:/usr/lib/../lib64/:/opt/gccrs-dev/lib/gcc/x86_64-pc-linux-gnu/11.0.0/../../../:/lib/:/usr/lib/
COLLECT_GCC_OPTIONS='-fsplit-stack' '-v' '-g' '-O0' '-o' 'test' '-shared-libgcc' '-mtune=generic' '-march=x86-64' '-dumpdir' 'test.'
 /opt/gccrs-dev/libexec/gcc/x86_64-pc-linux-gnu/11.0.0/collect2 -plugin /opt/gccrs-dev/libexec/gcc/x86_64-pc-linux-gnu/11.0.0/liblto_plugin.so -plugin-opt=/opt/gccrs-dev/libexec/gcc/x86_64-pc-linux-gnu/11.0.0/lto-wrapper -plugin-opt=-fresolution=/tmp/ccBljRM2.res -plugin-opt=-pass-through=-lgcc_s -plugin-opt=-pass-through=-lgcc -plugin-opt=-pass-through=-lc -plugin-opt=-pass-through=-lgcc_s -plugin-opt=-pass-through=-lgcc --eh-frame-hdr -m elf_x86_64 -dynamic-linker /lib64/ld-linux-x86-64.so.2 -o test /lib/x86_64-linux-gnu/crt1.o /lib/x86_64-linux-gnu/crti.o /opt/gccrs-dev/lib/gcc/x86_64-pc-linux-gnu/11.0.0/crtbegin.o -L/opt/gccrs-dev/lib/gcc/x86_64-pc-linux-gnu/11.0.0 -L/opt/gccrs-dev/lib/gcc/x86_64-pc-linux-gnu/11.0.0/../../../../lib64 -L/lib/x86_64-linux-gnu -L/lib/../lib64 -L/usr/lib/x86_64-linux-gnu -L/usr/lib/../lib64 -L/opt/gccrs-dev/lib/gcc/x86_64-pc-linux-gnu/11.0.0/../../.. /tmp/cchYC833.o -lm --wrap=pthread_create -lgcc_s -lgcc -lc -lgcc_s -lgcc /opt/gccrs-dev/lib/gcc/x86_64-pc-linux-gnu/11.0.0/crtend.o /lib/x86_64-linux-gnu/crtn.o
/usr/bin/ld: /lib/x86_64-linux-gnu/crt1.o: in function `_start':
(.text+0x24): undefined reference to `main'
collect2: error: ld returned 1 exit status```

### PR DESCRIPTION
**Pull request by philberty**
This will change when we implement visibility but the main method will be reserved case that will always be DECL_PUBLC.

Fixes #136

### DIFF
--- a/gcc/rust/backend/rust-compile-item.h
+++ b/gcc/rust/backend/rust-compile-item.h
@@ -138,10 +138,24 @@ public:
     // convert to the actual function type
     auto compiled_fn_type = TyTyCompile::compile (ctx->get_backend (), fnType);
 
+    unsigned int flags = 0;
+    bool is_main_fn = function.function_name.compare ("main") == 0;
+
+    // if its the main fn or pub visibility mark its as DECL_PUBLIC
+    // please see https://github.com/Rust-GCC/gccrs/pull/137
+    if (is_main_fn || function.has_visibility ())
+      flags |= Backend::function_is_visible;
+
+    std::string asm_name = function.function_name;
+    if (!is_main_fn)
+      {
+	// FIXME need name mangling
+	asm_name = "__" + function.function_name;
+      }
+
     Bfunction *fndecl
       = ctx->get_backend ()->function (compiled_fn_type, function.function_name,
-				       "" /* asm_name */, 0 /* flags */,
-				       function.get_locus ());
+				       asm_name, flags, function.get_locus ());
     ctx->insert_function_decl (function.get_mappings ().get_hirid (), fndecl);
 
     // setup the params

### CODE REVIEW

**Comment by bjorn3** [comment]
Generic and `#[inline]` functions must never be public as there could be multiple instances of them in the same crate graph.

As for the linker error in #136, rustc always defines the main shim in the same codegen unit as the rust main function.

**Comment by philberty** [comment]
> Generic and `#[inline]` functions must never be public as there could be multiple instances of them in the same crate graph.
> 
> As for the linker error in #136, rustc always defines the main shim in the same codegen unit as the rust main function.

Thanks for the review, I have changed this to only apply to the main function and any function with pub visibility. There will be changes to this later on when we start working on the cases you mention.  I think this is a middle ground for now to fix the bug.

File: gcc/rust/backend/rust-compile-item.h [hunk 1]
```diff
@@ -138,10 +138,19 @@ class CompileItem : public HIRCompileBase
...
+    unsigned int flags = 0;
+    bool is_main_fn = function.function_name.compare ("main") == 0;
+
+    // if its the main fn or pub visibility mark its as DECL_PUBLIC
+    // please see https://github.com/Rust-GCC/gccrs/pull/137
+    if (is_main_fn || function.has_visibility ())
+      flags |= Backend::function_is_visible;
+
+    // FIXME need name mangling
+    std::string asm_name = function.function_name;
```
NalaGinrut, line 150: The `main` function doesn't need the name mangling. I mention it here in case you miss it.

NalaGinrut (approved):

gcc/rust/backend/rust-compile-item.h
bjorn3: The rust main function needs symbol mangling. Only the main shim doesn't.

SimplyTheOther (approved):
PR looks great.

gcc/rust/backend/rust-compile-item.h
NalaGinrut: Alright, I see.
In Rust, the `main` function will be mangling first, then connect to the main shim. And the main shim was generated as a convention automatically.
This is different from C/C++, the `main` function is always the default symbol linked to `_start` defined in the linker script.

Thanks for correcting me! @bjorn3

bjorn3: Yeah, libstd does some extra work in addition to what the libc startup does like setting up a stack guard, adding a handler for SIGSEGV to give a nice error message on stack overflows, storing the program arguments in a static and running the main function within `std::panic::catch_unwind` to ensure that panicking doesn't immediately abort due to a missing unwind catcher. For reference the main shim calls https://github.com/rust-lang/rust/blob/a2cd91ceb0f156cb442d75e12dc77c3d064cdde4/library/std/src/rt.rs#L60 with the main function as first argument when using libstd. There are also a few other (unstable) options like marking a function with `#[start]`, in which case argc and argv are directly passed to that function without any additional arguments: https://github.com/bjorn3/rustc_codegen_cranelift/blob/3ea8915d4a247b5b3c4cfb3424c230ccd2645b17/example/alloc_example.rs#L30

bjorn3: Should I open an issue for handling the main shim?

philberty: > Should I open an issue for handling the main shim?

Yes @bjorn3 that would be great can you reference this PR in the issue too.

bjorn3: #144

philberty: Thanks for that @bjorn3

PR #148, opened 2021-01-13
Enforce type checking on implicit returns

philberty:
This change enforces strict type checkon on all ExprWithBlock. This includes unreachable tail expressions. Specific test cases have been added for context to ensure the failure cases are hit accordingly.

This change also improves the diagnostic errors for mismatched types where the reference HirId was lost leading to default locations.

There is a general hardening of the typechecker in this patch which will be used for further changes later on.


SimplyTheOther (approved):
PR looks great.

PR #149, opened 2021-01-14
This brings in resolution and type checking of the unit-type.

### LINKED ISSUE #93: Unit type support
Rust uses a builtin akin to void in other languages this is a core data structure

philberty:
It is possible to assign and declare variables of unit-type which translate down to zero sized void_type_node.

More work is needed to handle array and ADT types using unit-type when emitting gimple. The name+type resolution should be generic enough.

Fixes #93


SimplyTheOther (approved):
PR looks great.

Issue #78, opened 2020-12-21
Tuples compilation


philberty:
We should be able to compile TupleStructs and Anon Tuples.

philberty:
fixed in #153

PR #153, opened 2021-01-16
This adds supports for tuples

philberty:
More testing is required but this adds tuples apart from TupleStructs
which are parsed as CallExpr. This will be the primitives required to
finish that work.

Addresses #78


philberty:
```
void main ()
{
  struct (i32,bool) a;
  struct (i32,f32) b;

  try
    {
      a.0 = 123;
      a.1 = 1;
      b.0 = 456;
      b.1 = 5.0e+0;
    }
  finally
    {
      a = {CLOBBER};
      b = {CLOBBER};
    }
}
```

bjorn3:
Just to check: does `(a,)` work? That should give a tuple containing only `a`.

philberty:
> Just to check: does `(a,)` work? That should give a tuple containing only `a`.

in theory it should i will add a test case.

philberty:
Thanks @bjorn3 test case added

NalaGinrut (approved):
LGTM

Issue #173, opened 2021-01-19
CompoundAssignmentExpr Parser bug for |=


philberty:
```
beginning null denotation identifier handling
current peek token when starting path pratt parse: '%='
current token (just about to return path to null denotation): '%='
finished null denotation identifier path parsing - next is branching 
test.rs:21:9: error: found unexpected token ‘=’ in null denotation
   21 |     e %= 5;
      |         ^
null denotation is null; returning null for parse_expr
left denotation is null; returning null for parse_expr
test.rs:21:5: error: failed to parse expression for expression without block (pratt-parsed expression is null)
   21 |     e %= 5;
      |     ^
test.rs:21:5: error: failed to parse statement or expression without block in block expression
test.rs:21:11: error: unrecognised token ‘integer literal’ for start of item
   21 |     e %= 5;
      |           ^

```

The parser only seems to fail with:

```
    e %= 5;
    g |= 7;
    h ^= 8;
```

philberty:
%= and ^= are fixed in #175 

|= shows:

```
```
beginning null denotation identifier handling
current peek token when starting path pratt parse: '|='
current token (just about to return path to null denotation): '|='
finished null denotation identifier path parsing - next is branching 
test.rs:24:7: error: expecting ‘}’ but ‘|=’ found
   24 |     g |= 7;
      |       ^
test.rs:24:5: error: error may be from having an expression (as opposed to statement) in the body of the function but not last
   24 |     g |= 7;
      |     ^

```
```

PR #175, opened 2021-01-19
Fixes the ^= and %= compound expression parsing.

philberty:
The input needed to be skipped once the operator was parsed to ensure
the lexer did not duplicate the operator. |= still fails but this looks
to be a parser problem.

The output for |= shows:

```
current peek token when starting path pratt parse: '|='
current token (just about to return path to null denotation): '|='
test.rs:24:7: error: expecting ‘}’ but ‘|=’ found
   24 |     g |= 7;
      |       ^
test.rs:24:5: error: error may be from having an expression (as opposed to statement) in the body of the function but not last
   24 |     g |= 7;
      |     ^

```

I am unsure what the bug is there yet

Addresses #173

### DIFF
--- a/gcc/rust/lex/rust-lex.cc
+++ b/gcc/rust/lex/rust-lex.cc
@@ -466,7 +466,9 @@ Lexer::build_token ()
 	  if (peek_input () == '=')
 	    {
 	      // modulo-assign
+	      skip_input ();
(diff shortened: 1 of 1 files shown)


SimplyTheOther (commented):
This fix looks correct.

PR #182, opened 2021-01-26
Add mutablity checks and left hand size assignee checker

### LINKED ISSUE #77: mutability
We should respect the mut keyword on declarations such as:

let mut x;
x = 1;
x = 2;

philberty:
In order to assign to a name we must ensure the LHS is a valid expression
to assign to. This leads onto actually checking if this is a mutable
declaration or not.

Once these checks pass the name resolver we can in GIMPLE create immutable
types for these declarations to help with optimization.

Fixes #77


bjorn3:
This doesn't handle interior mutability, right? If you have multiple immutable references to an `UnsafeCell` you can derive a mutable reference to the contents of this `UnsafeCell` from any of these immutable references for as long as only one such mutable reference exists at any point. libcore implements `Freeze` for all types that don't contain an `UnsafeCell` to help the compiler with handling interior mutability.


```rust
let foo = &UnsafeCell::new(Vec::new());
let bar = &*foo;
unsafe { (*foo.get()).push(1); }
unsafe { (*bar.get()).push(2); }
unsafe { assert_eq!((*foo.get()).len(), 2); }
```

is not UB.

bjorn3:
> Is it ok if I use your comment to raise a new issue for interior mutability?

Sure

Issue #111, opened 2021-01-03
Impl Block support


philberty:
Support adding receiver methods to structs:
https://doc.rust-lang.org/rust-by-example/fn/methods.html

https://doc.rust-lang.org/std/keyword.impl.html

PR #203, opened 2021-02-02
Add support for impl blocks for functions and constants.

philberty:
This is the building block in adding support for methods and the self keyword.
    
    Fixes: #111
    
    Fix type resolution null pointer crash
    
    This also enhance ArithmeticOrLogicalExpr rules and associated error
    handling to propagate up the stack.


gcc/rust/ast/rust-ast.h
```diff
@@ -1333,6 +1333,8 @@ class InherentImplItem
 
   virtual void mark_for_strip () = 0;
   virtual bool is_marked_for_strip () const = 0;
+
+  virtual Location get_locus_slow () = 0;
```
SimplyTheOther, line 1337: Nitpick - this should be const-qualified.

gcc/rust/ast/rust-ast.h
```diff
@@ -1459,6 +1461,10 @@ class MacroInvocationSemi : public MacroItem,
   const std::vector<Attribute> &get_outer_attrs () const { return outer_attrs; }
   std::vector<Attribute> &get_outer_attrs () { return outer_attrs; }
 
+  Location get_locus () const { return locus; }
+
+  Location get_locus_slow () { return get_locus (); }
```
SimplyTheOther, line 1466: It would probably be best to qualify this with `const final override`.

gcc/rust/ast/rust-item.h
```diff
@@ -860,6 +867,14 @@ class Method : public InherentImplItem, public TraitImplItem
     return where_clause;
   }
 
+  Identifier get_method_name () const { return method_name; }
+
+  NodeId get_node_id () const { return node_id; }
+
+  Location get_locus () const { return locus; }
+
+  Location get_locus_slow () { return get_locus (); }
```
SimplyTheOther, line 876: It would probably be best to qualify this with `const final override`.

gcc/rust/ast/rust-item.h
```diff
@@ -1503,6 +1518,8 @@ class Function : public VisItem, public InherentImplItem, public TraitImplItem
 
   Location get_locus () const { return locus; }
 
+  Location get_locus_slow () { return get_locus (); }
```
SimplyTheOther, line 1521: It would probably be best to qualify this with `const final override`.

gcc/rust/ast/rust-item.h
```diff
@@ -2536,6 +2553,8 @@ class ConstantItem : public VisItem,
...
```
SimplyTheOther, line 2556: It would probably be best to qualify this with `const final override`.

SimplyTheOther (approved):
PR looks great. Had some minor nitpicks regarding `get_locus_slow`.

gcc/rust/ast/rust-ast.h
philberty: Thanks for the review i will fix these before merge

Issue #112, opened 2021-01-03
self keyword for receiver methods params


philberty:
support the self keyword for impl blocks
https://doc.rust-lang.org/std/keyword.self.html

as far as i can tell self is sugar for a parameter like:

```
struct foo(i32)

impl foo {
  fn bar(self: Self) {
  }
}
```

And Self is covered in #204

PR #205, opened 2021-02-04
Initial method resolution with Impl Block, Self Types and self arguments

### LINKED ISSUE #191: Method Call Expressions
https://doc.rust-lang.org/reference/expressions/method-call-expr.html

### LINKED ISSUE #204: Support Self within impl blocks
Self can be used as a reference back to the Type on an impl block such as:

```
struct foo(i32)

impl foo {
  fn new() -> Self {
    Self(123)
  }
}
```

philberty:
Add in support to compile Methods and MethodCallExpr
    
    There is more work to be done here with adjustments to the self argument
    such as borrows and mutability checking.
    
    Method resolution is basic, for now there is code to scan for all possible
    matches but traits are not supported at the moment so this resolves quite
    simply for now.
    
    Fixes #191 #112

    Support Self used within impl blocks
    
    This is simply a type reference so we can use the name resolver to resolve
    Self as a type reference back to the Type used for the impl block.
    
    Fixes #204

    Fix unreachable crash when a struct contains another algebraic data type
    
    When we have a nested struct this must resolve to the existing compiled
    struct and not create a new record type.


gcc/rust/ast/rust-pattern.h
```diff
@@ -91,6 +91,15 @@ class IdentifierPattern : public Pattern
       is_mut (is_mut), to_bind (std::move (to_bind)), locus (locus)
   {}
 
+  IdentifierPattern (NodeId node_id, Identifier ident, Location locus,
+		     bool is_ref = false, bool is_mut = false,
+		     std::unique_ptr<Pattern> to_bind = nullptr)
+    : Pattern (), variable_ident (std::move (ident)), is_ref (is_ref),
+      is_mut (is_mut), to_bind (std::move (to_bind)), locus (locus)
+  {
+    this->node_id = node_id;
```
SimplyTheOther, line 100: Wouldn't it be better to construct the node_id in the constructor list rather than the body?

SimplyTheOther (approved):
PR looks great, had one minor nitpick.

gcc/rust/ast/rust-pattern.h
philberty: Good find will fix.

PR #206, opened 2021-02-05
Move scan for unused names to be after type resolution

philberty:
Methods are resolved as part of type resolution so scanning ribs early
on will results in false warnings about unused methods.

### DIFF
--- a/gcc/rust/rust-session-manager.cc
+++ b/gcc/rust/rust-session-manager.cc
@@ -35,6 +35,7 @@
 #include "rust-ast-lower.h"
 #include "rust-hir-type-check.h"
 #include "rust-tycheck-dump.h"
+#include "rust-ast-resolve-unused.h"
 #include "rust-compile.h"
 
 extern Linemap *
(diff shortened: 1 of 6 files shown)


gcc/rust/rust-session-manager.cc
```diff
@@ -477,66 +473,41 @@ Session::parse_file (const char *filename)
...
 
   // expansion pipeline stage
   expansion (parsed_crate);
-  fprintf (stderr, "\033[0;31mSUCCESSFULLY FINISHED EXPANSION \n\033[0m");
-
   if (options.dump_option == CompileOptions::EXPANSION_DUMP)
     {
-      // dump AST with expanded stuff
-      fprintf (stderr, "BEGIN POST-EXPANSION AST DUMP\n");
-      parser.debug_dump_ast_output (parsed_crate);
```
SimplyTheOther, line 522: I think that this expansion dump should be kept, at least temporarily if you are changing the way dumps work. It is useful to compare the results of the AST pre-macro expansion and post-macro expansion.

SimplyTheOther (approved):
PR looks good. Had one issue with removing the expansion dump code.

gcc/rust/rust-session-manager.cc
philberty: Sounds good i will keep that in. I want to make a change separate to this to allow multiple dump options to be specified but it will be a separate PR

philberty: I will revert these changes actually for now and add in the scan or unused to make this change cleaner.

philberty: Reverted thanks again!

PR #207, opened 2021-02-05
Enforce Duplicate definition rules

### LINKED ISSUE #208: Duplicate Items are not allowed to shadow
Currently, the compiler is not failing duplicate definitions for Items such as:

```
struct S1 {
    x: f64,
    y: f64,
}

struct S1(i32, bool);

fn main() {}
```

This will lead to bad type resolution later on and should not be allowed.

philberty:
Rust does not allow functions/methods/constants/static definitions to
shadow otherwise you end up with unusable items.

Fixes #208


SimplyTheOther (approved):
PR looks great.

PR #209, opened 2021-02-05
Fix overloaded virtual and unused parameters warnings

philberty:


gcc/rust/backend/rust-compile-context.h
```diff
@@ -222,11 +222,20 @@ class TyTyResolveCompile : public TyTy::TyVisitor
 
   virtual ~TyTyResolveCompile () {}
 
-  void visit (TyTy::ErrorType &type) override { gcc_unreachable (); }
+  void visit (TyTy::ErrorType &type ATTRIBUTE_UNUSED) override
```
SimplyTheOther, line 225: Apparently a more portable means of preventing the warning is to simply omit the parameter name.

For example, visiting the ErrorType would be written as so:
```
void visit (TyTy::ErrorType &) override
{
  gcc_unreachable ();
}
```

I have been told that this is more portable (something about attributes not being interpreted in the same way with different C++ compilers) but I don't actually know whether this is truly the case or not.

gcc/rust/typecheck/rust-hir-type-check-base.h
```diff
@@ -34,209 +34,285 @@ class TypeCheckBase : public HIR::HIRVisitor
...
-  virtual void visit (HIR::TraitObjectTypeOneBound &type) {}
-  virtual void visit (HIR::TupleType &type) {}
-  virtual void visit (HIR::NeverType &type) {}
-  virtual void visit (HIR::RawPointerType &type) {}
-  virtual void visit (HIR::ReferenceType &type) {}
-  virtual void visit (HIR::ArrayType &type) {}
-  virtual void visit (HIR::SliceType &type) {}
-  virtual void visit (HIR::InferredType &type) {}
-  virtual void visit (HIR::BareFunctionType &type) {}
+  virtual void visit (HIR::Token &tok ATTRIBUTE_UNUSED) override {}
```
SimplyTheOther, line 37: For nitpicky cleanup purposes only, it is probably better to omit the `virtual` keyword, since `override` does not require it to be present to cause the compiler to check it overrides another function and omitting the word makes the line shorter.

gcc/rust/typecheck/rust-hir-type-check-item.h
```diff
@@ -31,14 +31,16 @@ namespace Resolver {
 
 class TypeCheckItem : public TypeCheckBase
 {
+  using TypeCheckBase::visit;
```
SimplyTheOther, line 34: Does this line actually need to be here? `TypeCheckBase::visit` doesn't seem to be referenced in the class anywhere.

gcc/rust/typecheck/rust-tycheck-dump.h
```diff
@@ -27,6 +27,8 @@ namespace Resolver {
 
 class TypeResolverDump : public TypeCheckBase
 {
+  using TypeCheckBase::visit;
```
SimplyTheOther, line 30: I would have the same question here about this line.

SimplyTheOther (approved):
Cleanup is mostly good but had a few minor nitpicks.

gcc/rust/backend/rust-compile-context.h
philberty: I think your right it looks really ugly too see the unused attributed like that too.

gcc/rust/typecheck/rust-hir-type-check-base.h
philberty: I am still having issues with the overloaded virtual method is hidden warning so i will do as you say here for now

gcc/rust/typecheck/rust-hir-type-check-item.h
philberty: good find i was testing something out to do with overloaded virtual and it didnt work

PR #210, opened 2021-02-05
This adds support for basic BlockExpressions

### LINKED ISSUE #189: Block Expressions
https://doc.rust-lang.org/reference/expressions/block-expr.html

Most of the code should be in place for this but will provide a chance to clean up some work from implicit returns #129.

philberty:
We keep temporary's for each block in order for the result to be
referenced. For example:

  let x = { test() + 1 };

This can be resolved into:

  {
    let x:i32;

    _tmp1:i32;
    {
      _tmp2:i32 = test();
      _tmp1 = _tmp2 + 1;
    }

    x = _tmp1;
  }

Fixes #189

Depending on the order of merging PRs the method resolution one will need an amendment to use this shared function block code.

### DIFF
--- a/gcc/rust/backend/rust-compile-base.h
+++ b/gcc/rust/backend/rust-compile-base.h
@@ -237,9 +237,13 @@ public:
 protected:
   HIRCompileBase (Context *ctx) : ctx (ctx) {}
 
+  Context *ctx;
+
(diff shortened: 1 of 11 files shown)


SimplyTheOther (approved):
PR looks great.

PR #211, opened 2021-02-06
With the merge of BlockExprs

philberty:
Further testing found regressions with implicit returns and type
resolution.

This unifies the type resolution in block expressions to be more strict
and ensure everything bar the final statement should be UnitType.

It also gets rid of the is_final_expr hack used for implicit returns before

#212 if fixed would have picked this up


SimplyTheOther (approved):
PR looks great.

PR #215, opened 2021-02-08
Conditional blocks with else can resolve into a non unit type expression

### LINKED ISSUE #213: If elseif and else can resolve into a type
TypeResolution correctly handles if with else block resolving into a type but this can also be applied to if blocks containing else if and a final else for example:

```
fn test(x: i32) -> i32 {
    if x == 10 {
        123
    } else if x < 10 {
        456
    } else {
        789
    }
}

fn main() {
    let a = test(1);
}
```

philberty:
When we compile if blocks the else gives us a final branch that can
resolve. Single IfExpr cannot resolve as there is no final branch.

This also fixes a small regression in compiling elif blocks where the
gimple body was empty.

Fixes #213

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.h
+++ b/gcc/rust/backend/rust-compile-expr.h
@@ -417,7 +417,6 @@ public:
 
   void visit (HIR::IfExprConseqElse &expr)
   {
-    // this can be a return expression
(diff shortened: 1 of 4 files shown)


SimplyTheOther (approved):
PR looks great.

PR #216, opened 2021-02-08
Fixes parsing if expressions as part of an expression.

### LINKED ISSUE #214: Error parsing rust ternery expressions of if else
Preparing to parse files. 
Attempting to parse file: test.rs
current peek token when starting path pratt parse: '>'
current token (just about to return path to null denotation): '>'
test.rs:12:13: error: found unexpected token ‘if’ in null denotation
   12 |     let c = if a > 10 { true } else { false };
      |             ^
test.rs:12:16: error: failed to parse expression in let statement
   12 |     let c = if a > 10 { true } else { false };
      |                ^
test.rs:12:5: error: failed to parse statement or expression without block in block expression
   12 |     let c = if a > 10 { true } else { false };
      |     ^
test.rs:13:1: error: unrecognised token ‘}’ for start of item
   13 | }
      | ^
test.rs:13:1: error: failed to parse item in crate

```
fn main() {
    let a = 123;

    let b = {
        if a > 10 {
            true
        } else {
            false
        }
    };

    let c = if a > 10 { true } else { false };
}
```

Here the parser fails to parse the assignment to c.

philberty:
This allows for rust style ternery expressions.

Fixes #214

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -7504,12 +7504,23 @@ Parser<ManagedTokenSource>::parse_loop_label ()
 template <typename ManagedTokenSource>
 std::unique_ptr<AST::IfExpr>
 Parser<ManagedTokenSource>::parse_if_expr (
-  std::vector<AST::Attribute> outer_attrs ATTRIBUTE_UNUSED)
+  std::vector<AST::Attribute> outer_attrs ATTRIBUTE_UNUSED, bool pratt_parse)
 {
   // TODO: make having outer attributes an error?
-
-  Location locus = lexer.peek_token ()->get_locus ();
-  skip_token (IF);
+  Location locus = Linemap::unknown_location ();
+  if (!pratt_parse)
+    {
+      locus = lexer.peek_token ()->get_locus ();
+      if (!skip_token (IF))
+	{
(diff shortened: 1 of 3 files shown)


bjorn3:
Does this also allow the following?

```rust
fn main() {
    let a = loop {};
    let b = while true {};
    let c = for _ in &[()] {};
    let d = match loop {} {};
    let e = return;
    f(return);
}
```

And the gems at

https://github.com/rust-lang/rust/blob/921ec4b3fca17cc777766c240038d7d50ba98e0d/src/test/ui/weird-exprs.rs

philberty:
> Does this also allow the following?
> 
> ```rust
> fn main() {
>     let a = loop {};
>     let b = while true {};
>     let c = for _ in &[()] {};
>     let d = match loop {} {};
>     let e = return;
>     f(return);
> }
> ```
> 
> And the gems at
> 
> https://github.com/rust-lang/rust/blob/921ec4b3fca17cc777766c240038d7d50ba98e0d/src/test/ui/weird-exprs.rs

I know offhand that this will fail with loop expr but I am working on that in a separate branch. I think I will raise this as a separate issue. The names of those functions are brilliant. Thanks for the heads up once again.

SimplyTheOther (approved):
PR looks great.

PR #218, opened 2021-02-08
Fix bad type resolution for CallExpr when fn is not simply PathInExpression

### LINKED ISSUE #217: Function Pointer calls inside FieldAccessExpressions
```
bjorn312:45
Are function pointers already supported? If so does calling a function pointer in a field require (val.field)(arg) like rustc? This is to disambiguate field function pointer calls from method calls when both have the same name.
```

There is no simple way to differentiate between a MethodCallExpr and CallExpr when the form is of FieldAccessExpr which will end up looking like a MethodCallExpr.

To test our how GCCRS works in this scenario it was found using the rustic workaround syntax GCCRS crashes.

Without the parenthesis we get similar errors in both compilers:

```
test.rs:17:13: error: failed to resolve the PathExprSegment to any Method                                                                                                                                                                              
   17 |     let e = d.a(d.b);                                                                                                                                                                                                                          
      |             ^
```

Rustc gives:

```
error[E0599]: no method named `a` found for struct `Foo` in the current scope
  --> test.rs:17:15
   |
1  | struct Foo {
   | ---------- method `a` not found for this
...
17 |     let e = d.a(d.b);
   |               ^ field, not a method
```

philberty:
When we have a test case that is, for example, a FieldAccessExpression the
compiler must resolve and compile that field into a reference which can be
called. This is not a simple direct call to a function in that scenario.

Fixes #217

### DIFF
--- a/gcc/rust/backend/rust-compile.cc
+++ b/gcc/rust/backend/rust-compile.cc
@@ -53,33 +53,9 @@ CompileCrate::go ()
 void
 CompileExpr::visit (HIR::CallExpr &expr)
 {
-  // this can be a function call or it can be a constructor for a tuple struct
-  Bexpression *fn = ResolvePathRef::Compile (expr.get_fnexpr (), ctx);
-  if (fn != nullptr)
+  Btype *type = ResolvePathType::Compile (expr.get_fnexpr (), ctx);
+  if (type != nullptr)
     {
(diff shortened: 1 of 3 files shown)


SimplyTheOther (approved):
PR looks great.

PR #221, opened 2021-02-09
Fix parsing |= as a compound assignment expression.

### LINKED ISSUE #173: CompoundAssignmentExpr Parser bug for |=

philberty:
Fixes #173

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -386,6 +386,8 @@ Parser<ManagedTokenSource>::left_binding_power (const_TokenPtr token)
       return LBP_MOD_ASSIG;
     case AMP_EQ:
       return LBP_AMP_ASSIG;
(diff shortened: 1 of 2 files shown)


SimplyTheOther (approved):
PR looks great.

PR #223, opened 2021-02-09
Simple Loops

### LINKED ISSUE #106: Loop expressions
This should make gccrs handle loop { } statements

### LINKED ISSUE #219: Loops fail to be parsed as an expression:
example:

```
fn main() {
    let mut a = 1;
    let mut b = 1;

    // first number in Fibonacci sequence over 10:
    let fib = loop {
        if b > 10 {
            break b;
        }
        let c = a + b;
        a = b;
        b = c;
    };
}
```

Gives:

```
test.rs:6:15: error: found unexpected token ‘loop’ in null denotation
    6 |     let fib = loop {
      |               ^
test.rs:6:20: error: failed to parse expression in let statement
    6 |     let fib = loop {
      |                    ^
test.rs:6:5: error: failed to parse statement or expression without block in block expression
    6 |     let fib = loop {
      |     ^
test.rs:9:9: error: unrecognised token ‘}’ for start of item
    9 |         }
      |         ^
test.rs:9:9: error: failed to parse item in crate

```

This is similar issue to #214

philberty:
Support Simple LoopExpr
    
    This is the building block for the rest of loops where we have a basic
    infinite loop. Break/Continue reliest on the resolution of labels and
    breaks can also make a loop into a BlockExpr
    
    Fixes #106

    Fix parse of LoopExpr as part of a normal expression
    
    For example this fixes the form of: let x = loop { ... };
    
    Fixes #219

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -1139,11 +1139,14 @@ private:
 
   Location locus;
 
+  NodeId node_id;
+
 public:
   // Constructor
   Lifetime (LifetimeType type, std::string name = std::string (),
 	    Location locus = Location ())
-    : lifetime_type (type), lifetime_name (std::move (name)), locus (locus)
+    : lifetime_type (type), lifetime_name (std::move (name)), locus (locus),
+      node_id (Analysis::Mappings::get ()->get_next_node_id ())
   {}
 
   // Creates an "error" lifetime.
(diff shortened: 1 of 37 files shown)


SimplyTheOther (approved):
PR looks great. I have one question though - it appears that according to the [GCC docs](https://gcc.gnu.org/onlinedocs/gccint/Basic-Statements.html#Basic-Statements) that GENERIC supports infinite loops as a specific node type (`LOOP_EXPR`). Would it be better to create that node type instead for optimisation purposes?

philberty:
> PR looks great. I have one question though - it appears that according to the [GCC docs](https://gcc.gnu.org/onlinedocs/gccint/Basic-Statements.html#Basic-Statements) that GENERIC supports infinite loops as a specific node type (`LOOP_EXPR`). Would it be better to create that node type instead for optimisation purposes?

Good point i forgot about that i will change that

PR #227, opened 2021-02-11
Changes to AST structure in preparation for macro expansion and cleanup

SimplyTheOther:
This PR makes some changes to the structure of the AST for macro expansion and cleanup purposes.

For example, the representation of a macro invocation has been partially unified - both `MacroInvocation` and `MacroInvocationSemi` now have a `MacroInvocData` element that can be operated on, leading to greater code reuse. Also, now AST tokens do not store a copy of the lexer token data, but rather store a `shared_ptr` to a lexer token, enabling much cheaper and easier conversion between lexer tokens (useful for parser) and AST tokens (useful for storage as `TokenTree`s). I have added a new `ASTFragment` class and `SingleASTNode` class in order to represent the results of macro invocations.

Also, some deficiencies with the AST were addressed, such as increasingly complicated code that relied on `Expr` storing outer attributes itself (rather than all child expressions storing them), and the lack of outer attribute support in `IfExpr`, for instance. These movements of outer attribute storage down to child classes should enable better AST dumps in the future as well. 

I have also made some general cleanup, such as adding `final override` to several virtual functions that shouldn't be overriden further, such as `get_locus_slow` in classes that containing a definition of `get_locus`. This could improve optimisation.


philberty (approved):
nice one

Issue #85, opened 2020-12-21
String literals


philberty:
We should add support to compile raw str strings to GIMPLE.

let x = "mystring"

PR #230, opened 2021-02-11
Char type

philberty:
Add char type
    
    This might need changes in the Lexer to allow for wchar_t to be preserved.
    
    Addresses #85

### DIFF
--- a/gcc/rust/backend/rust-compile-context.h
+++ b/gcc/rust/backend/rust-compile-context.h
@@ -422,6 +422,14 @@ public:
     translated = compiled_type;
   }
 
+  void visit (TyTy::CharType &type) override
+  {
+    ::Btype *compiled_type = nullptr;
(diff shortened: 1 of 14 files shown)


SimplyTheOther (approved):
PR looks good. I'm a little concerned about "slicing" of the character (since the value is derived from the first byte of the string only) in rust-compile-expr.h, but I imagine this just needs a method call to fix (e.g. encode string to UTF-32 with a method call or something).

bjorn3:
`wchar_t` may be a 16 bit number on some systems, right? A rust `char` is a 32 bit number where every value outside of the range of valid unicode codepoints is invalid and transmuting such an invalid value to `char` is UB.

philberty:
Yep, i think it needs some work to fix the slicing of the character but the GIMPLE tree is a 32 bit unsigned so the backend should be correct at least. I will raise an issue out of this PR for that part.

philberty:
I've raised https://github.com/Rust-GCC/gccrs/issues/233

Issue #196, opened 2021-01-27
Inferred type


philberty:
https://doc.rust-lang.org/reference/types/inferred.html

Issue #170, opened 2021-01-19
dereference expression


philberty:
https://doc.rust-lang.org/reference/expressions/operator-expr.html#the-dereference-operator

Issue #169, opened 2021-01-19
Borrow Expressions


philberty:
https://doc.rust-lang.org/reference/expressions/operator-expr.html#borrow-operators

PR #231, opened 2021-02-12
Add ReferenceType with BorrowExpr and DereferenceExpr

philberty:
This also adds in the mising InferenceType _ which was mostly implemented
before as part of Data Structures 1.

We create GENERIC REFERENCE_TYPES for these this is the building block
to finish work on mutability rules and pointers.

Fixes: #196
Addresses: #169 #170

Note more work is needed here for mutable references and double references.


SimplyTheOther (approved):
PR looks great.

Issue #187, opened 2021-01-27
Add is_equal to TyTy types


philberty:
Currently to check if types are equal or can be unified we call combine. This is inefficient for many cases like FnTypes we want a strict is_equal check.

We can add this as an abstract method to the TyTy::TyBase, it might be worth renaming this class at the same time as the name bugs me.

RuchiGitHu:
Good Evening,
I want to fix the issue but this is my first step towards contributing in open source project. Please let me know about this issue.
Thanks in advance..

philberty:
Hi @RuchiGitHu thanks for your interest. Please be aware in this project we require contributors to follow the copyright assignment process https://gcc.gnu.org/contribute.html 

There are 2 parts to this issue.

1. Where inside https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/typecheck/rust-tyty.h we rename and refactor the usage of TyBase I think the name is bad at the moment, maybe call it BaseType. It represents the abstract base class for all potential types within the rust language. Such as ADTType which is the algebraic data type to represent structs for example. The purpose here is a cleanup effort on naming.
2. The next piece is we need to add a new virtual method to this BaseType class in the same file called: virtual bool is_equal(TypeBase&other) const {} For now this can be implemented within the base class like this {  return get_type() == other.get_type(); }

Some more thought is needed once this block of work is in as to where we use this instead of calling combine to unify types.

Hope this helps.

RuchiGitHu:
Thanks @philberty for the information you provided to me. I will try to fix it.

philberty:
Fixed in #238

PR #238, opened 2021-02-16
Cleanup tyty

YizhePKU:
Todos:
- [x] Rename `TyBase` to `BaseType`
- [x] Rename `combine` to `unify`
- [x] Add `is_equal` method to `BaseType`(Issue #187)
- [ ] Replace `unify` calls with `is_equal` calls where appropriate
- [x] ~~Extract BaseRules::combine as a standalone function~~
- [x] Remove `base` field from `BaseRules`
- [x] Documentation

### DIFF
--- a/gcc/rust/backend/rust-compile-context.h
+++ b/gcc/rust/backend/rust-compile-context.h
@@ -53,7 +53,7 @@ public:
 	rust_assert (
 	  tyctx->lookup_type_by_node_id ((*it)->get_node_id (), &ref));
 
-	TyTy::TyBase *lookup;
+	TyTy::BaseType *lookup;
 	rust_assert (tyctx->lookup_type (ref, &lookup));
 
 	auto compiled = TyTyCompile::compile (backend, lookup);
@@ -261,7 +261,7 @@ private:
 class TyTyResolveCompile : public TyTy::TyVisitor
 {
 public:
-  static ::Btype *compile (Context *ctx, TyTy::TyBase *ty)
+  static ::Btype *compile (Context *ctx, TyTy::BaseType *ty)
   {
     TyTyResolveCompile compiler (ctx);
     ty->accept_vis (compiler);
@@ -355,7 +355,7 @@ public:
     std::vector<Backend::Btyped_identifier> fields;
     for (size_t i = 0; i < type.num_fields (); i++)
       {
-	TyTy::TyBase *field = type.get_field (i);
+	TyTy::BaseType *field = type.get_field (i);
 	Btype *compiled_field_ty
 	  = TyTyCompile::compile (ctx->get_backend (), field);
 
--- a/gcc/rust/backend/rust-compile-expr.h
+++ b/gcc/rust/backend/rust-compile-expr.h
@@ -57,7 +57,7 @@ public:
 	return;
       }
 
-    TyTy::TyBase *tyty = nullptr;
+    TyTy::BaseType *tyty = nullptr;
     if (!ctx->get_tyctx ()->lookup_type (expr.get_mappings ().get_hirid (),
 					 &tyty))
       {
@@ -177,7 +177,7 @@ public:
 	      return;
 	    }
 
-	  TyTy::TyBase *tyty = nullptr;
+	  TyTy::BaseType *tyty = nullptr;
 	  if (!ctx->get_tyctx ()->lookup_type (
 		expr.get_mappings ().get_hirid (), &tyty))
 	    {
@@ -203,7 +203,7 @@ public:
 	      return;
 	    }
 
-	  TyTy::TyBase *tyty = nullptr;
+	  TyTy::BaseType *tyty = nullptr;
 	  if (!ctx->get_tyctx ()->lookup_type (
 		expr.get_mappings ().get_hirid (), &tyty))
 	    {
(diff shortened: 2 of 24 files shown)


philberty:
don't worry about the greetings failure I think I might remove it and try something else.

philberty:
Something which might be a future project is to add a can_eq method but it will be a separate piece of work later probably for traits milestone

YizhePKU:
After adding some documentation, `BaseRules::unify` now looks clear enough that I decide to leave it as is. Plus it's an internal function anyway, so no need to put it in the global scope.

YizhePKU:
@philberty What's the original purpose of `is_unit`? [rustc_middle::ty::TyKind](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/enum.TyKind.html) doesn't list the unit type, so rustc probably used a zero-element tuple as the unit type?

philberty:
> @philberty What's the original purpose of `is_unit`? [rustc_middle::ty::TyKind](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/enum.TyKind.html) doesn't list the unit type, so rustc probably used a zero-element tuple as the unit type?

I originally thought it unit type and unit-structs are subtly different but actually we can just get around this by representing them as a unit-struct TyTy so i think the goal should be to get rid of that in a separate PR. We also need to separate FnDef and FnPtr currently everything is an FN typekind which is ok... but i reckon later on it will be best to separate them.

bjorn3:
> so rustc probably used a zero-element tuple as the unit type?

Indeed: https://github.com/rust-lang/rust/blob/f1c47c79fe8438ed241630f885797eebef3a6cab/compiler/rustc_middle/src/ty/sty.rs#L1676-L1681

YizhePKU:
It occurs to me that using double dispatch to implement type inference might be overkill. As it stands, two types can be unified if and only if
1. They're the same kind, and have the same underlying types
2. At least one of the types is an inference variable

philberty:
> It occurs to me that using double dispatch to implement type inference might be overkill. As it stands, two types can be unified if and only if
> 
>     1. They're the same kind, and have the same underlying types
> 
>     2. At least one of the types is an inference variable

I am not convinced this will hold up for all types since we have not implemented them all yet. Currently, I am working on Generics which introduces the Param type which might have some changes here.

As a personal preference I prefer keeping code as explicit as possible for example we have unify code for InferTyype + i32, which means we take others to be i32. Or we have the case of i32 + InferType then that means we take self. This simple case of ordering does not need an algorithm to figure out what's going on we see the 2 types and can make the correct decision for that situation.

Some of what you are talking about could be addressed with a can_eq method which is something we will need for traits.

Also, I don't see where the double dispatch is, the only dispatch for type unification occurs in TyBase::combine where we visit the other type so we don't need to cast (GCC does not allow RTTI in C++), we rely on the method combine in each of the TyTy types so there is no need to dispatch there.

philberty:
If you are implementing the is_equal please be careful where you use this. The main places i can think of using it is in the Fndef type unification.

The other place is in method resolution, but i think that will need a can_eq case because rust adjusts receivers to be able to match the methods self. So if we have:

let a:Foo = Foo{...};

a.method();

but method has a Borrowed self param the compiler should add in the adjustment to take the reference here. At the moment this will fail and needs a can_eq.

YizhePKU:
> As a personal preference I prefer keeping code as explicit as possible

I agree. The current code orgnization is great for accomodating future changes.

> the only dispatch for type unification occurs in TyBase::combine where we visit the other type

Here's how I understand the situation. We need to potentially implement a different unification logic for each pair of types. It's like filling in a 2d table. From this viewpoint, `TyBase::combine` acts as the first dispatch(a row in the table), and `visit` acts as the second dispatch(a column in the table). My original thought was that a full 2d table is not necessary, since every row has at most two columns. But keeping the entire table is certainly more explicit and flexible.

YizhePKU:
There're two ways to address the duplicated `base` field problem:
1. Remove the `base` field from derived rules. The disadvantage is that derived rules need to type-cast every time they access the `base` field. It's conceptually simple but quite verbose.
2. Remove the `base` field from `BaseRules`. The problem is that `BaseRules` cannot access the `base` field directly any more, but we can work around this by introducing a (private) virtual method `get_base` that derived rules implement.

I went for the second approach.

YizhePKU:
@philberty I'm ready to merge. Please take a look.

`is_equal` is implemented, but currently I don't understand the codebase well enough to recognize where it should be used. Maybe later.

About copyright issues: GNU has responded and asked me for some extra info. I should receive the copyright assignment form in a few day, which will cover my "past and future changes" on this project.

philberty:
> @philberty I'm ready to merge. Please take a look.
> 
> `is_equal` is implemented, but currently I don't understand the codebase well enough to recognize where it should be used. Maybe later.
> 
> About copyright issues: GNU has responded and asked me for some extra info. I should receive the copyright assignment form in a few day, which will cover my "past and future changes" on this project.

That's brilliant great work. Thanks for persevering through the TyTy code base, its really helped what you have done here. I want this PR to go in before I raise my PR for the first part of generics which you might like to review when its up probably early next werek.

philberty (approved):
great work lets go for it

YizhePKU:
Thanks for the encouragement. It means a lot to me. I'll do my best:)

PR #242, opened 2021-02-17
Format all files in gcc/rust/ using clang-format and add clang format GitHub Action

humancalico:
Signed-off-by: Akshat Agarwal <humancalico@disroot.org>

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -1,0 +1,20 @@
+name: Clang Format Lint
+
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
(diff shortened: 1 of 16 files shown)


SimplyTheOther (commented):
PR looks fine to me, at least the reformatting of the .h and .cc files. I have no idea about the clang-format action though.

philberty:
> PR looks fine to me, at least the reformatting of the .h and .cc files. I have no idea about the clang-format action though.

@SimplyTheOther the goal is that the checker will fail the build on PRs if the code is not formatted I think it will help make code reviews easier. I think its a good idea but could be annoying if your editor does not automatically clang-format for you.

philberty (approved):

PR #245, opened 2021-02-17
Add debug interface to GCC debug_tree on GENERIC graphs

philberty:
This adds in a debugging interface for the backend IR to help diagnose
issues with the tree.


SimplyTheOther (approved):
PR looks great.

PR #251, opened 2021-02-25
Support Generic arguments to Structs and Tuple Structs

### LINKED ISSUE #235: Support Generic Params on Structs
https://doc.rust-lang.org/rust-by-example/generics/gen_fn.html

philberty:
This removes StructFieldType from the TyTy base as it is not a type that
can be unified against.

It adds in a substition mapper implementation which will likely change
over time when this this support is extended over to Functions.

Note generic argument binding is not supported as part of this yet.
This is the building block needed to move forward with generics.

Fixes #235


SimplyTheOther (approved):
PR looks great.

PR #254, opened 2021-02-25
Update default configs in ExpansionCfg

NalaGinrut:
Keep ExpansionCfg configs complying with `librustc_expand`.


SimplyTheOther (approved):
PR looks great.

PR #255, opened 2021-02-25
Add in support for string literals

### LINKED ISSUE #85: String literals

philberty:
This gives the apropriate reference type over const char *.

Fixes #85

### DIFF
--- a/gcc/rust/backend/rust-compile-context.h
+++ b/gcc/rust/backend/rust-compile-context.h
@@ -268,8 +268,6 @@ public:
     return compiler.translated;
   }
 
-  virtual ~TyTyResolveCompile () {}
-
   void visit (TyTy::ErrorType &) override { gcc_unreachable (); }
 
   void visit (TyTy::InferType &) override { gcc_unreachable (); }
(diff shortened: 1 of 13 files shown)


bjorn3:
> This gives the apropriate reference type over const char *.

Is this `&str` a fat pointer? And does the literal not contain a trailing '\0'?

(By the way raw strings usually refer to the `r#"fo"o"#` syntax which doesn't require escaping the contents.)

philberty:
> > This gives the apropriate reference type over const char *.
> 
> Is this `&str` a fat pointer? And does the generated data not contain a trailing '\0'?
> 
> (By the way raw strings usually refer to the `r#"fo"o"#` syntax which doesn't require escaping the contents.)

Thanks @bjorn3 yep its a fat pointer and it does contain the null terminator. This was just an initial block for this str data type, and byte strings etc are not implemented yet so they will be separate tickets/prs.

SimplyTheOther (approved):
PR looks great. Although I do agree that the use of "raw string" is a bit confusing - perhaps "string literals" could be better?

philberty:
> PR looks great. Although I do agree that the use of "raw string" is a bit confusing - perhaps "string literals" could be better?

Yeah my bad there, I updated the commit message etc to reflect this. I am pushing to get this in since it would be ideal to have some kind of macro interception to implement println over to a printf call for example as well as the assert macros.

PR #257, opened 2021-02-26
Fix to parse inner attribute correctly

NalaGinrut:
This patch fixes the inner attribute parsing. And throw errors properly.

Here's a test code:
```rust
// sum.rs
#![recursion_limit="128"]
macro_rules! sum {
    ($base:expr) => { $base };
    ($a:expr, $($rest:expr),+) => { $a + sum!($($rest),+) };
}

fn main() {
println!("{}", sum!(1,2,3,4,5));
```

Compile:
```bash
gccrs -frust-dump-parse sum.rs -c | less gccrs.ast.dump
```
We can get:
```rust
Crate: 
 inner attributes: 
 recursion_limit = 128
 items: 
......
```

BTW, `rustc` lacks of necessary checks, but we make it better, for example, if we write some bad code:
```rust
#!,recursion_limit="128"]
...
```
`rustc` doesn't complain at all. But in patched `gccrs`:
```bash
sum.rs:1:3: error: Expect `['
    1 | #!,recursion_limit="128"]
      |   ^
sum.rs:1:3: error: unrecognised token ‘,’ for start of item
sum.rs:1:3: error: failed to parse item in crate
```

PS: In this test case, the invocation of `println!` will throw an error that is unrelated to this patch. I'll take look at macro expansion for that issue.

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -206,11 +206,11 @@ Parser<ManagedTokenSource>::skip_generics_right_angle ()
       return true;
       case RIGHT_SHIFT: {
 #if 0
-	/* shit. preferred HACK would be to replace this token in stream with
-	 * '>', but may not be possible at this point. */
-	// FIXME: ensure locations aren't messed up
-	TokenPtr right_angle = Token::make (RIGHT_ANGLE, tok->get_locus () + 1);
-	lexer.replace_current_token (right_angle);
+        /* shit. preferred HACK would be to replace this token in stream with
+         * '>', but may not be possible at this point. */
+        // FIXME: ensure locations aren't messed up
+        TokenPtr right_angle = Token::make (RIGHT_ANGLE, tok->get_locus () + 1);
+        lexer.replace_current_token (right_angle);
 #endif
 
 	// new implementation that should be better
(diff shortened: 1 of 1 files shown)


NalaGinrut:
I think I'm using the correct `clang-format` config in `gcc/contrib`, and my last patch has no redundant indentation, but it seems this patch has some unexpected indentation. I don't know if it's my config problem. Could you double-check it? @philberty @

SimplyTheOther (approved):
PR looks good.

bjorn3:
> BTW, rustc lacks of necessary checks, but we make it better, for example, if we write some bad code:

That is because rustc interprets it as shebang like a shell. This makes it possible to have the same file be a shell script and valid rust file. See for example https://github.com/bjorn3/rustc_codegen_cranelift/blob/86c594426f247ae34ef12f5e4dfd97f5ffe093c8/scripts/filter_profile.rs#L1-L8

NalaGinrut:
@bjorn3 Thanks for the explanation! 
However, even it's used for shebang, it should check typo for users when it's an inner attribute.

And I can even write a ridiculous syntax:
```rust
#!,,,,,,,recursion_limit="128"]
```
This can be compiled with `rustc` silently. Looks like a potential security issue.

But anyway, folks may always leave the first line blank, since I found `rustc` will check since the second line.

Issue #212, opened 2021-02-06
TestSuite Fail compilation tests are passing when they shouldn't


philberty:
Fail compilation tests are designed to compile some code and test for valid failures such as bad type found and pass only when the compiler fails, if a crash is found it should also fail

I think the test suite is looking for assembly output which is empty but it should fail with the non zero exit code. Any help would be greatly appreciated here.

tschwinge:
@philberty, would you appreciate if I worked through the GCC/Rust testsuite, to change some _local idioms_ into _standard GCC upstream idioms_ -- which I suppose would include addressing the issue raised here?

philberty:
Yes @tschwinge that would be brilliant, I copied from the gdc and go test suites to bootstrap the dejagnu stuff but I don't have much experience with dejagnu so any help would be really appreciated.

I am working on hacking in a basic println and assert_eq macro support so we can write runnable automated tests which would be nice to have soon.

YizhePKU:
I think we can close this issue now.

philberty:
thanks

PR #261, opened 2021-03-01
Make failing tests actually fail

YizhePKU:
As mentioned in issue #212, some tests in fail_compliation are passing when they shouldn't. The reason is that `{ dg-do run { xfail *-*-* } }` will be successful if the compilation failed _or_ if the resulting executable returns a non-zero value.

This PR fixes this by replacing it with `{ dg-do compile }`, which detects if the compiler crashed or not, followed by `{ dg-final { output-exists-not } }`, which ensures the compilation actually failed.

Now the failing tests are `fail_compilation/implicit_returns_err1.rs` and `fail_compilation/tuple_struct1.rs`.

### DIFF
--- a/gcc/testsuite/rust.test/rust-test.exp
+++ b/gcc/testsuite/rust.test/rust-test.exp
@@ -314,7 +314,8 @@ proc dmd2dg { base test } {
 	}
 
 	fail_compilation {
-	    puts $fdout "// { dg-do run { xfail *-*-* } }"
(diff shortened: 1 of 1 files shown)


philberty:
Brilliant thanks for this let me have a look at the failures and i will get back to this PR we might merge this anyway.

philberty:
let's go for this and then we can fix the failures as a priority

Issue #234, opened 2021-02-15
Support Generic Params on functions


philberty:
https://doc.rust-lang.org/rust-by-example/generics/gen_fn.html

PR #268, opened 2021-03-08
TyTy Cleanup and Fix Inference Type bug

### LINKED ISSUE #258: Infertype used on Array TyTy fails
```
../array.rs:2:13: error: unable to determine type: please give this a type: 18
    2 |     let a: [_; 5] = [1, 2, 3, 4, 5];
      |             ^
../array.rs:3:9: warning: unused name
    3 |     let b: u32 = a[1];
      |         ^
```

This fails to compile as the Array TyTy object holds onto a concrete TyTy object but needs to follow Tuple Type to simply use an HirId to represent the Type so that it can change.

### LINKED ISSUE #263: Compiler is defaulting floats to f32 instead of f64
https://rust-lang.github.io/rfcs/0212-restore-int-fallback.html

We default to i32 and f32 for inference variables on numbers inside typecheck/rust-hir-type-check.cc

philberty:
Introduce new TyCtx wrapper which denotes types that must be inferred. Which cleans up the code for ArrayTypes, TupleTypes and ReferenceTypes.

As an InferType is a BaseType during unification we iterate the chain of references involved to update the context when something has been resolved. When we resolve against raw numbers we get back another InferType but with InferKind non General so this is a case where the type is resolved since this can default to i32. Further refactoring here will tidy up code as part of #121 

This also separates fndef into function definitions and function pointers, since function definitions will support substations as part of #234 but function pointers do not. Leveraging the TyCtx here for function pointers we also gain the ability to use inference variables here as well fixing #269 

Fixes #258 and #269


YizhePKU:
Would you explain why you introduced `TyCtx`? What's the problem that it solves?

Is `TyCtx` a "context" like `TypeCheckingContext`?(It doesn't look like one.)

philberty:
> Would you explain why you introduced `TyCtx`? What's the problem that it solves?
> 
> Is `TyCtx` a "context" like `TypeCheckContext`?(It doesn't look like one.)

I have updated the comment for the PR with more information. But really this denotes types that can change. For instance, when constructing a ArrayType it used to take a BaseType as the element type which is ambiguous, using TyCtx denotes this can be an inference variable in the code and usages of this element type must be looked up to get the BaseType proper.

This was then reused for Function pointers to allow for inference variables there. fixing #269

philberty:
@YizhePKU i created a new issue TyVar might be a better name lets change it.

PR #271, opened 2021-03-09
Remove unused header inclusions.

dkm:
Both headers are not used and trigger warning on DEFHOOK macros redefinitions.


dkm:
While browsing/building, I saw some warnings about redef. This simple patch gets rid of it... One less warning :)

philberty:
nice find

PR #276, opened 2021-03-12
Proof of Concept: using s-expression for AST dump

YizhePKU:
This is a proof-of-concept implementation for dumping AST into s-expression. I chose s-expression for demonstration because it's simple enough to write a serialization procedure for it by hand. If we end up using JSON we might need to pull in additional dependency.

Two new functions are added:
* `to_sexp()` converts an object to a s-expression string. It's a overloaded template function, similar to `std::to_string()`.
* `sexp()` creates a new compound s-expression containing some sub-expressions(which is to say, wrap them with a pair of parenthesis). `sexp()` accepts an arbitrary number of parameters(via some template magic).

The follow types of objects are serializable:
* numbers
* `std::string`
* any class that implements `SexpSerializable`, which is an interface containing a single `to_sexp` method
* containers of serializable objects, e.g. `std::vector<SexpSerializable*>`
* pointers to serializable objects, e.g. `std::unique_ptr<std::string>`

Calling `to_sexp()` or `sexp()` on containers or pointers will recursively call these functions for the underlying objects. As a result, implement the `SexpSerializable` interface is often a matter of listing what values you want to dump; the rest is handled automatically.

A new header file, `util/rust-expr.h`, contains implementation of these functions. It's a big pile of meta-programming magic(I had way too much fun doing these meta-programming.)

---

As an example, here's how you would implement `SexpSerializable` for the `AST::Crate` object:

```c++
struct Crate : public SexpSerializable {
    std::vector<std::unique_ptr<Item>> items;
    NodeId node_id;

    std::string to_sexp() const override {
        return sexp("Crate", sexp("node_id", node_id),
                             sexp("items", items));
    }

    // ...other methods
}
```

Assuming the underlying types also implements `SexpSerializable`, calling `to_sexp()` on a crate will generate this:

```lisp
(Crate 
    (node_id 123)
    (items 
        (Function main 
            (returns void) 
            (params ) 
            (body 
                (Block 
                    (statements [Stmt placeholder] [Stmt placeholder]) 
                    (final_expr [nullptr]))))))
```

---

Limitations:
* I only implemented `SexpSerializable` for a few classes for demonstration purpose. It will need a lot of polish.
* Currently `to_sexp()` only generates a compact, minimal s-expression. The previous output is formatted by one of my editor plugins.

I'm happy to hear any thoughts and suggestions. This is just a phototype so I'm OK rewriting it from scratch if necessary.

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -28,6 +28,8 @@
 #include "rust-token.h"
 #include "rust-location.h"
 
+#include "util/rust-sexp.h"
+
 namespace Rust {
 // TODO: remove typedefs and make actual types for these
 typedef std::string Identifier;
@@ -801,7 +803,7 @@ struct MetaListNameValueStr;
 
 /* Base statement abstract class. Note that most "statements" are not allowed in
  * top-level module scope - only a subclass of statements called "items" are. */
-class Stmt
+class Stmt : public SexpSerializable
 {
 public:
   // Unique pointer custom clone function
@@ -813,6 +815,9 @@ public:
   virtual ~Stmt () {}
 
   virtual std::string as_string () const = 0;
+
+  // Placeholder for Stmt that hasn't implemented to_sexp
+  virtual std::string to_sexp () const override { return "[Stmt placeholder]"; }
 
   virtual void accept_vis (ASTVisitor &vis) = 0;
 
@@ -863,9 +868,12 @@ protected:
 class ExprWithoutBlock;
 
 // Base expression AST node - abstract
-class Expr
+class Expr : public SexpSerializable
 {
 public:
+  // Placeholder for Expr that don't serialize to s-expression yet
+  std::string to_sexp () const override { return "[Expr placeholder]"; }
+
   // Unique pointer custom clone function
   std::unique_ptr<Expr> clone_expr () const
   {
@@ -1045,7 +1053,7 @@ protected:
 class TraitBound;
 
 // Base class for types as represented in AST - abstract
-class Type
+class Type : public SexpSerializable
 {
 public:
   // Unique pointer custom clone function
@@ -1058,6 +1066,8 @@ public:
   virtual ~Type () {}
 
   virtual std::string as_string () const = 0;
+
+  std::string to_sexp () const override { return "[Type placeholder]"; }
 
   /* HACK: convert to trait bound. Virtual method overriden by classes that
    * enable this. */
@@ -1542,7 +1552,7 @@ protected:
 };
 
 // A crate AST object - holds all the data for a single compilation unit
-struct Crate
+struct Crate : public SexpSerializable
 {
   bool has_utf8bom;
   bool has_shebang;
@@ -1598,6 +1608,11 @@ public:
 
   // Get crate representation as string (e.g. for debugging).
   std::string as_string () const;
+
+  virtual std::string to_sexp () const override
+  {
+    return sexp ("Crate", sexp ("items", items));
+  }
 
   // Delete all crate information, e.g. if fails cfg.
   void strip_crate ()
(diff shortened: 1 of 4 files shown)


SimplyTheOther:
Have you considered a "visitor"-based implementation of this instead? I would think it would be cleaner, especially if a more complex format such as JSON is used.

philberty:
Yeah I think this might be better served via a visitor based implementation like https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/typecheck/rust-tycheck-dump.h

philberty:
Nice work though HIR dumps in rustc are Sexpressions and i wish we had a similar output since we want to output the NodeMappings alongside the tree item itself

philberty:
Sorry accidentally pressed close keep up the good work.

YizhePKU:
Thanks for the encouragement! I'll do my best.

YizhePKU:
> Have you considered a "visitor"-based implementation of this instead? I would think it would be cleaner, especially if a more complex format such as JSON is used.

Could you elaborate why you think visitor-based implementation is cleaner?

YizhePKU:
The current design is to implement dumping logic in a `to_sexp` method in every class. If we were to use a visitor-based solution, we end up putting all dumping logic into a single "Resolver" class. I argue this is not the best approach:

1. The dumping logic is closely coupled with internals of the class being dumped, so it should stay close to the class itself. In particular, a `visit()` function cannot access private members of the class being dumped.
2. It requires two virtual dispatches(first `accept_vis()`, then `visit()`) instead of one. Visitor pattern is often used to simulate double dispatch, but in our case we only need single dispatch.

SimplyTheOther:
> 
> 
> The current design is to implement dumping logic in a `to_sexp` method in every class. If we were to use a visitor-based solution, we end up putting all dumping logic into a single "Resolver" class. I argue this is not the best approach. The dumping logic is closely coupled with internals of the class being dumped, so it should stay close to the class itself. In particular, a `visit()` function cannot access private members of the class being dumped.

It depends how you conceptualise the dumping process. Either you consider it a property of the class, and hence the method belongs in the class, or you consider it a property of a "dumper". 
But that isn't the main issue. There are two main issues in my opinion:

- Having lots of different types of "dump" methods pollutes the class interface
- Dump methods with complicated formatting routines are easier to implement with a separate class (that keeps track of more state)

Having only an s-expression dump doesn't really pollute the class interface, but if we added routines for JSON dumps, YAML dumps, clang-style dumps, etc., this would definitely do so. As such, I thought it could be better to create a generic "dump" framework now rather than have to refactor this all later.

The second point also doesn't really apply to s-expression dumps as much, depending on how exactly you've implemented it. If you add the "prettyprint" indentation formatting, then you'd need to store the current indentation level, and possibly other information relating to parentheses, I would think. The best way to store this would be in a "SExprDumper" class that does the formatting, I would think. Doing so with JSON or other file formats would be even more complicated.

As for the `visit()` functions not being able to access private members - that is true, but basically all AST and probably HIR classes have (const or non-const) getter methods for all relevant members, I think.

YizhePKU:
> Having lots of different types of "dump" methods pollutes the class interface

Admittedly so. I think it's acceptable; it's just one method per dump format.

> Dump methods with complicated formatting routines are easier to implement with a separate class (that keeps track of more state)

Actually, this doesn't call for a visitor pattern. It's straightforward to add something like `DumpingContext` to a method. We already have `TypeCheckContext`, for example.

YizhePKU:
To summarize the discussion so far.

* methods couples closely with the class, while visitors are separated from the class
* methods pollute the class interface(a little)
* methods can access private members of the class
* methods can access private members

I'll admit I have a tendency to avoid visitors when possible, mainly because they're really hard to read and understand--convoluted dispatch route, unhelpful function names, no return values...Sometimes we have no choice, but in this case we do have a viable alternative.

YizhePKU:
Closing in favor of #318.

Issue #270, opened 2021-03-09
Change CI workflow to make -j 


philberty:
Currently the CI job runs make on its own we should try and take advantage of make -j $(nproc) or something if the GitHub workflow allows.

PR #277, opened 2021-03-13
CI workflow make -j commit

therealansh:
issue #270 resolved, added make -j $(nproc) to CI workflow.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -44,7 +44,7 @@ jobs:
     - name: Build
       run: |
            cd gccrs-build; \
-           make
+           make -j $(nproc)
 
     - name: Test
       run: |


.github/workflows/ccpp.yml
```diff
@@ -44,10 +44,10 @@ jobs:
...
       run: |
            cd gccrs-build; \
-           make
+           make -j $(nproc)
 
     - name: Test
       run: |
            cd gccrs-build; \
-           make check-rust | tee results.log; \
+           make -j $(nproc) check-rust | tee results.log; \
```
philberty, line 52: make check-rust is a little bit different there is code in place call dejagnu in parallel mode. https://github.com/Rust-GCC/gccrs/blob/b8549556123da80e60922eaa928cfd284f831303/gcc/rust/Make-lang.in#L98

Can you remove the -j for the test section?

philberty:
Nice one it brings the CI build time down from 37 min to 19 min or so. :D Thanks for this!

philberty:
Before we merge this can you squash your commits into 1?

bjorn3:
You could use the squash and merge option.

philberty (approved):

.github/workflows/ccpp.yml
tschwinge: @philberty: the DejaGnu support code that you referenced is just to provide functionality for testing in parallel mode, but not to actually enable that by default; you still do need to run `make -j[...] check-rust` to run testing in parallel mode.  So, it would seem to me that the change as originally proposed by @therealansh has actually been correct.

In practice, it doesn't make a big difference right now: `check-rust` is quick to complete even for (default) `-j1`.  Leaving it to you guys if you'd like to re-visit this issue now -- or if we do that at the point when `check-rust` takes longer than a few seconds.  ;-)

If running testing in parallel mode (`-j[...]`), the test log currently written by `[...] | tee results.log` will (likely) be different for each run.  You should instead inspect the `[build]/gcc/testsuite/rust/rust.sum` file.

PR #284, opened 2021-03-15
added to_string() for TypeKind

### LINKED ISSUE #280: Implement TyTy::TypeKind::to_string
https://github.com/Rust-GCC/gccrs/blob/b8549556123da80e60922eaa928cfd284f831303/gcc/rust/typecheck/rust-tyty.h#L29

We have an enum of Types we should implement a to_string method for debugging.

therealansh:
resolves #280 , this PR adds to_string() to TyTy::TypeKind. The method is implemented using switch case to map enum to string.

### DIFF
--- a/gcc/rust/typecheck/rust-tyty.h
+++ b/gcc/rust/typecheck/rust-tyty.h
@@ -101,8 +101,10 @@ public:
 
   virtual bool has_subsititions_defined () const { return false; }
 
-  virtual std::string to_string () const {
-    switch(kind){
+  virtual std::string to_string () const 
+  {
+    switch(kind)
+      {
       case TypeKind::INFER:
         return "Infer";
         break;
@@ -118,7 +120,7 @@ public:
       case TypeKind::PARAM:
         return "PARAM";
         break;
-      case: TypeKind::ARRAY:
+      case TypeKind::ARRAY:
         return "ARRAY";
         break;
       case TypeKind::FNDEF:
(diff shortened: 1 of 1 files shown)


gcc/rust/typecheck/rust-tyty.h
```diff
@@ -101,6 +101,70 @@ class BaseType
 
   virtual bool has_subsititions_defined () const { return false; }
 
+  virtual std::string to_string () const
```
philberty, line 104: I don't think we want a to_string implemented on TyTy::BaseType. This class already has a BaseType::to_string which seems like a conflict here.

philberty:
I would suggest that you add a new class TypeKindFormat which has a static std::string to_string (TypeKind kind) { switch .... }

therealansh:
would this TypeKindFormat class have more methods, in any case, to format the TypeKind in other ways?

SimplyTheOther (approved):
PR looks great.

philberty:
@therealansh can you run clang format ok? There is a clang format file within the toplevel/contrib folder

philberty:
> would this TypeKindFormat class have more methods, in any case, to format the TypeKind in other ways?

Yes we have a basic dump for the TypeResolution using -frust-dump-all but this could be improved by having this class.

philberty:
Are you able to rebase this work so it can be merged I think this is good to go in since it is a small PR.

philberty (approved):

Issue #260, opened 2021-03-01
Crash with zero-length array


YizhePKU:
Input program:
```
fn main() {
    let array: [i64; 0] = [];
}
```

Output:
```
Preparing to parse files.
SUCCESSFULLY PARSED CRATE
ran register_plugins (with no body)
SUCCESSFULLY REGISTERED PLUGINS
started injection
finished injection
SUCCESSFULLY FINISHED INJECTION
started expansion
finished expansion
SUCCESSFULLY FINISHED EXPANSION
rust1: internal compiler error: in get_array_elems, at rust/ast/rust-expr.h:1196
0x85cc83 Rust::AST::ArrayExpr::get_array_elems()
	../../gccrs/gcc/rust/ast/rust-expr.h:1196
0x85cc83 Rust::Resolver::ResolveExpr::visit(Rust::AST::ArrayExpr&)
	../../gccrs/gcc/rust/resolve/rust-ast-resolve-expr.h:200
0x747759 Rust::AST::ArrayExpr::accept_vis(Rust::AST::ASTVisitor&)
	../../gccrs/gcc/rust/ast/rust-ast-full-test.cc:5093
0x862650 Rust::Resolver::ResolveExpr::go(Rust::AST::Expr*, unsigned int)
	../../gccrs/gcc/rust/resolve/rust-ast-resolve-expr.h:36
0x862650 Rust::Resolver::ResolveStmt::visit(Rust::AST::LetStmt&)
	../../gccrs/gcc/rust/resolve/rust-ast-resolve-stmt.h:56
0x7484e9 Rust::AST::LetStmt::accept_vis(Rust::AST::ASTVisitor&)
	../../gccrs/gcc/rust/ast/rust-ast-full-test.cc:5741
0x85ac7c Rust::Resolver::ResolveStmt::go(Rust::AST::Stmt*, unsigned int)
	../../gccrs/gcc/rust/resolve/rust-ast-resolve-stmt.h:37
0x85ac7c operator()
	../../gccrs/gcc/rust/resolve/rust-ast-resolve.cc:336
0x85ac7c _M_invoke
	/usr/include/c++/9/bits/std_function.h:285
0x84275d std::function<bool (Rust::AST::Stmt*)>::operator()(Rust::AST::Stmt*) const
	/usr/include/c++/9/bits/std_function.h:688
0x85b862 Rust::AST::BlockExpr::iterate_stmts(std::function<bool (Rust::AST::Stmt*)>)
[...]

PR #288, opened 2021-03-17
Fix crash with zero length array

YizhePKU:
This PR fixes the crash that results from zero-length arrays, such as `let arr = []`(issue #260). `ArrayExpr` now uses `ArrayElemsValues` to represent an empty array like that, instead of `nullptr`.

However, currently such code still doesn't compile. Four tests were added, two of which report an error:

```Rust
// compilable/array_empty_list.rs
fn main() {
    let arr = [];
}

// rust1: error: unable to determine type: please give this a type: 24
```

```Rust
// compilable/array_zero_length_fold.rs
fn main() {
    let arr = ["Hello"; 123 - 123];
}

// array_zero_length_fold.rs:2:25: fatal error: failed to fold capacity constant
```

I think these are best treated as separate bugs. We can merge first and fix later, or we can delay this PR. Either is OK.


YizhePKU:
Whoops, just realized `array_empty_list.rs` actually fails to compile with rustc. I'll move it to fail_compilations then.

YizhePKU:
The formatting check failed because I'm using clang-format 11, while CI is using clang-format 10. I'll find a way to fix that.

gcc/rust/typecheck/rust-hir-type-check-expr.h
```diff
@@ -676,14 +680,25 @@ class TypeCheckExpr : public TypeCheckBase
...
+
+    for (auto &type : types)
       {
-	infered_array_elems = infered_array_elems->unify (types.at (i));
+	infered_array_elems = infered_array_elems->unify (type);
       }
-
     for (auto &elem : types)
-      infered_array_elems->append_reference (elem->get_ref ());
+      {
```
dkm, line 699: IIRC GNU style does not force braces around single statement loop/if. Is this something clang-format added?

YizhePKU: Nah, I added it myself. It's just personal preference. Thanks for looking at my PR :)

SimplyTheOther (approved):
PR looks mostly fine to me.

One minor exception was the lack of a space between the nested template ending brackets - I thought this was required by the gcc formatting style? At least gcc's configuration for clang-format on my computer seems to cause this to occur. Ignore this if they changed it or something.

YizhePKU:
I experimented a little. It seems that the space is optional; clang-format preserves the space if you have it, but doesn't add the space eagerly.

philberty (approved):

philberty:
Nice work, this kind of change I fear we may have to hold off on for copyright assignment, how is that going? If it is in transit back to the FSF then it can be ok.

YizhePKU:
Sadly I'm still waiting for a response from my school officials. I'll notify you when it's done.

gcc/testsuite/rust.test/fail_compilation/array_empty_list.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
+    let arr = [];
```
philberty, line 2: With the new changes to the testsuite @dkm has done you should be able to add a dg-error line here if you rebase this PR.

YizhePKU: Cool. I'll do that.

dkm:
The commit pushed in this PR seems incorrect. Maybe you did some merges at some point?

YizhePKU:
OK I may have screwed up the commits. I did a merge a while ago to keep up with master. Yesterday I rebased against master and that resulted in that merge getting replayed on top of master. Maybe just removing that merge commit is good enough?

dkm:
Rebasing a merge commit is rarely a good idea :) The easiest would simply be to revert to master and reapply your commits by hand using `cherry-pick` (or `rebase --onto` with commit range), or you could also `merge --squash` everything (basically doing a diff with master and creating a commit with this). Let me know if you need help with this, I can have a closer look if needed

YizhePKU:
I fixed the problem by force-pushing. I think this is all right?

dkm:
Yes, looks correct now!

gcc/rust/typecheck/rust-hir-type-check-expr.h
```diff
@@ -702,14 +706,25 @@ class TypeCheckExpr : public TypeCheckBase
...
-    for (size_t i = 1; i < types.size (); i++)
+    // Create an infer type and register it in TypeCheckContext
+    // TODO: This is very leaky. Need a refactor of some sort.
+    auto crateNum = Analysis::Mappings::get ()->get_current_crate ();
+    auto nodeId = UNKNOWN_NODEID;
+    auto hirId = mappings->get_next_hir_id ();
+    auto localDefId = UNKNOWN_LOCAL_DEFID;
+    auto mappings = Analysis::NodeMapping (crateNum, nodeId, hirId, localDefId);
+    infered_array_elems
+      = new TyTy::InferType (hirId, TyTy::InferType::InferTypeKind::GENERAL);
```
philberty, line 717: I have a wrapper over this that you can use:

https://github.com/Rust-GCC/gccrs/blob/e997db044394dd7ae751f52c209320a0097218be/gcc/rust/typecheck/rust-tyty.cc#L51

You should be able to call:

```c++
infered_array_elems = TyTy::TyVar::get_implict_infer_var ();
```

YizhePKU:
@philberty I've sent my copyright assignment papers to FSF. We can merge this PR now.

Also, I noticed a typo in the method name `get_implict_infer_var`(should be `get_implicit_infer_var`). I'll fix that in a second.

philberty:
> @philberty I've sent my copyright assignment papers to FSF. We can merge this PR now.
> 
> Also, I noticed a typo in the method name `get_implict_infer_var`(should be `get_implicit_infer_var`). I'll fix that in a second.

Thats great news @YizhePKU thanks for working though it this is great work.

Issue #224, opened 2021-02-10
Rewrite parser error message system to better enable backtracking


SimplyTheOther:
In order to enable what is effectively backtracking (required when parsing macros during macro expansion), there must be a way of recovering from parse errors being thrown from the parser.

My best idea for implementation would require the error to essentially not be thrown in the first place (in the sense that `error_at` should not be called) and rather have any parse errors be registered somewhere, and then if a calling function doesn't expect any errors, the actual `error_at` functions for those registered errors will be called.

This implementation method would require rewrite each parse function to return both a result and a list of errors, with errors being stored a new data structure that would contain an error message and a location, and maybe other information that could be useful for diagnostics. These errors would be propagated up functions.

In a way, this implementation method is somewhat similar to Rust's tagged-union "possibly error" types, which I believe are used extensively in rustc.

An alternative implementation method could be to use exceptions, though I believe gcc does not allow this for frontends.

philberty:
What if you introduce something like rust_parse_error_at(Location locus, bool macro_mode, __VA_ARGS__)

So if it is macro_mode you can store the errors off into a side table and then do something with them. You could even mark them with the NodeId of the particular macro that is being expanded. This may not work but might be another option to think about.

SimplyTheOther:
One issue with simply calling a function like that is that it would require global state for the error table. If you rewrote it to not be a global error table (e.g. creating an instance of error table stored by each Parser instance), then it would be essentially the same as creating Error class instances as in my idea, except instead of propagating them up functions it would be stored per-Parser. 

As for storing errors per-Parser rather than propagating them up functions, this method seems like it would have advantages and disadvantages. The main advantage is that there would likely be less boilerplate code required for error handling in each parsing method. The disadvantage is that it may not interact very well with some of the parsing methods that are effectively "try to parse this but if you can't it isn't an error". I'll see if I can implement in this way first, and if it doesn't work very well, I'll switch back to my original plan.

SimplyTheOther:
That PR should've closed this issue, not sure why it didn't.

PR #290, opened 2021-03-17
Modify parser error message reporting system to allow recovery

SimplyTheOther:
Basically, this modifies the parser error message reporting system to create `Error` structs instead of directly calling `rust_error_at`, so that errors can be ignored if need be. This effectively allows backtracking, which is useful for macro expansion and possibly regular parsing in the future. 

This was implemented by having a stored error table per `Parser` instance, and then emitting the stored errors if desired after parsing has completed. This should allow several parsers to coexist without issues relating to error handling.

Addresses #224


bjorn3:
@SimplyTheOther The username `@redbrain` has been registered by someone else. You may want to update your remote url.

SimplyTheOther:
> 
> 
> @SimplyTheOther The username `@redbrain` has been registered by someone else. You may want to update your remote url.

Thanks for the heads-up. I have now changed it in my local repo.

philberty:
Yeah I used to use that username a few years ago but I forgot to keep my blog going and it got registered by some shopping webapp lol. My wife calls me philbert so I thought I would try and stick with that.

philberty (approved):
nice one

PR #291, opened 2021-03-17
Fix Backend:: mark 2 methods pure

dkm:
virtual `loop_expression` and `exit_expression` were not marked pure and do not have
a default impl, causing error during linking when optimisation are disabled.

This fixes this error:
```
/usr/bin/ld: rust/rust-gcc.o: in function `Backend::~Backend()':
...gccrs/build/gcc/../../gcc/rust/rust-backend.h:67: undefined reference to `vtable for Backend'
/usr/bin/ld: rust/rust-gcc.o: in function `Backend::Backend()':
...gccrs/build/gcc/../../gcc/rust/rust-backend.h:64: undefined reference to `vtable for Backend'
collect2: error: ld returned 1 exit status
make: *** [../../gcc/rust/Make-lang.in:92: rust1] Error 1
```

### DIFF
--- a/gcc/rust/rust-backend.h
+++ b/gcc/rust/rust-backend.h
@@ -468,10 +468,10 @@ public:
     = 0;
 
   // infinite loop expressions
-  virtual Bexpression *loop_expression (Bblock *body, Location);
+  virtual Bexpression *loop_expression (Bblock *body, Location) = 0;
 
   // exit expressions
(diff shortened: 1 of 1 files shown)


SimplyTheOther (approved):
PR looks great.

philberty:
Nice find i saw this once last week and i couldn't find where the issue was. Thanks!

philberty (approved):

PR #292, opened 2021-03-17
wip: default float to f64

### LINKED ISSUE #263: Compiler is defaulting floats to f32 instead of f64

therealansh:
this PR resolves #263 and defaults float to f64


therealansh:
PR #294 solves the issue linked closing this PR

PR #294, opened 2021-03-17
Refactoring TypeResolution Resolve with new InferType::default_type and defaulting float to f64

thomasyonug:
I'm not sure that my default_type is ok or not.   #263

### DIFF
--- a/gcc/rust/typecheck/rust-tyty.cc
+++ b/gcc/rust/typecheck/rust-tyty.cc
@@ -106,6 +106,29 @@ InferType::clone ()
 			get_combined_refs ());
 }
 
+bool
+InferType::default_type (BaseType **type) const
+{
+  auto context = Resolver::TypeCheckContext::get ();
+  bool ok = false;
+  switch (infer_kind)
+    {
+    case GENERAL:
+      return false;
+      case INTEGRAL: {
+	ok = context->lookup_builtin ("i32", type);
+	rust_assert (ok);
+	return ok;
+      }
+      case FLOAT: {
+	ok = context->lookup_builtin ("f64", type);
+	rust_assert (ok);
+	return ok;
+      }
+    }
+  return false;
+}
+
 void
 ErrorType::accept_vis (TyVisitor &vis)
 {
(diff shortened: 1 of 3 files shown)


thomasyonug:
By the way, the code style of GCC looks so weird to me.

philberty (approved):

philberty:
@thomasyonug this looks good to me, this is on the edge of needing copyright assignment do you have any progress with that? https://gcc.gnu.org/contribute.html

gcc/rust/typecheck/rust-tyty.cc
```diff
@@ -106,6 +106,29 @@ InferType::clone ()
...
+InferType::default_type (BaseType **type) const
+{
+  auto context = Resolver::TypeCheckContext::get ();
+  bool ok = false;
+  switch (infer_kind)
+    {
+    case GENERAL:
+      break;
+      case INTEGRAL: {
+	ok = context->lookup_builtin ("i32", type);
```
philberty, line 119: this could be simpler with just return context->lookup and remove the assertions

gcc/rust/typecheck/rust-tyty.cc
```diff
@@ -106,6 +106,29 @@ InferType::clone ()
...
+	rust_assert (ok);
+      }
+      break;
+      case FLOAT: {
+	ok = context->lookup_builtin ("f64", type);
+	rust_assert (ok);
+      }
+      break;
+    }
+  return ok;
```
philberty, line 129: then this can be simply return false.

gcc/rust/typecheck/rust-tyty.cc
philberty: Actually, keep the assertions but you can return ok instead of break. Since this is very unlikely to fail but incase it does an assertion will help alot.

gcc/rust/typecheck/rust-tyty.cc
thomasyonug: Yep, this seems more readable.

thomasyonug:
> @thomasyonug this looks good to me, this is on the edge of needing copyright assignment do you have any progress with that? https://gcc.gnu.org/contribute.html

Sorry, I'll try to sign this stuff tonight.

thomasyonug:
I had sent a email to request my assignment form for my past and future change. Hope I can sign it as soon as possible.

thomasyonug:
@philberty 
Hi, I had signed the disclaimer to cover my future changes to GNU software, as well as my past changes. 
And my school signed a disclaimer too.

philberty:
Nice one in that case your covered i can merge this now. Thanks for bearing with the process! great work!

Issue #236, opened 2021-02-15
 Support Generic Params on TupleStructs


philberty:
Tuple Structs have generic params

philberty:
is a dup of #250

PR #297, opened 2021-03-18
Generics continued this adds more type resolution to ADT and Functions

philberty:
Recursive function substution is not supported here with this patch. As the
patch was getting bigger it seems better to merge sooner than later to
avoid conflicts.

#235 #234 #236


SimplyTheOther (approved):
PR looks great.

Issue #252, opened 2021-02-25
Remove TyTy::UnitType


philberty:
> so rustc probably used a zero-element tuple as the unit type?

Indeed: https://github.com/rust-lang/rust/blob/f1c47c79fe8438ed241630f885797eebef3a6cab/compiler/rustc_middle/src/ty/sty.rs#L1676-L1681

_Originally posted by @bjorn3 in https://github.com/Rust-GCC/gccrs/issues/238#issuecomment-779841432_

philberty:
marking this as a good first project though I think there might be some dragons here.

dkm:
I'm trying to resolve this issue, hope to have something to push shortly :)

philberty:
Good luck this will in the end fix #155 which i never finished. I think you might find some issues with GCC it does not have support for zero sized types. So there needs to be care in the Gimple code generation to not emit any expressions using it since it is simply void.

dkm:
In fact, I tried the naive approach of removing it and see what breaks. And after patching all errors, it seems to work. At least, all tests are passing. I still have to really understand all the parts so I may have overlooked something, but haven't had any problems.

philberty:
Nice work thanks for working on it. Since this is a bigger piece of code being changed I think this will be the right time to remind you about FSF copyright assignments btw. If its something in progress we can probably merge because it allows for past and future PRs.

dkm:
The FSF documents are already in the process of being signed, but still on my side, so better wait for them to be transmitted to the FSF before merging. I hope it's only a matter of days now.

philberty:
Yeah, as soon as you send them back I can merge. Really great work though :D I hope it was an enjoyable process at least. Doing that change i think you will have got a good feel for where things are going on in the code i will mark more issues for good first PR.

dkm:
Yes, it was really enjoyable. In particular because the task was not too complex for someone still not used to the code base. Feel free to add more simple PR, I'll pick some gladly :D

PR #301, opened 2021-03-21
Refactor testsuite using more standard DejaGNU

dkm:
This is a first draft for a little refactoring of the testsuite.
I've tried to use a more "standard" way of testing (inspired by c torture tests and other lang).
I kept the separate directory for the test that are expected to raise error but renamed it `xfail_compile`.

I've used the default for torture tests, so each compile job is using a combination of optim level, debug and lto.
I think that the testsuite is small enough for that to be ok, but this can easily be reduce.

```
                === rust Summary ===

# of expected passes            1332
# of unexpected failures        7
# of expected failures          63
```
The `expected failures` are coming from the tests in `xfail_compile`.

The new execute tests are working, but still need some care to handle the xfail cases (they are the `unexpected failures` above, where the non-zero return is not correctly xfailed by dg).

This is a first draft, but any input welcome :)


dkm:
There are some workarounds to handle the debug-verbosity of the compiler (`dg-excess-error` and pruning directives).

tschwinge:
Generally, doing *torture testing* (different optimizations levels etc.) seems fine to me.

If we agree that the `rust.test/xfail_compile/` testcases are exactly those where the GCC/Rust front end terminates compilation, then we don't need *torture testing*, because we never ever get that far that optimization levels etc. would matter.

tschwinge:
PASS: rust.test/xfail_compile/continue1.rs   -O0   (test for errors, line 8)
    PASS: rust.test/xfail_compile/continue1.rs   -O0   (test for errors, line 8)
    FAIL: rust.test/xfail_compile/continue1.rs   -O0  (test for excess errors)
    [...]

The *excess errors* are the usual GCC/Rust front end debugging messages.  It first wasn't clear to me why I'm seeing these as *excess errors* not generally but only for this (and a few other) testcases, but then I saw the `additional_prunes` in `gcc/testsuite/lib/rust.exp:rust_target_compile`.  What we're seeing for `rust.test/xfail_compile/continue1.rs` is `rust1: error: expected () got <tyty::error>` which does match the `error` regexp, so that's to be expected.  As done in a few other files, adding `{ dg-excess-errors "Noisy error and debug" }` to `rust.test/xfail_compile/continue1.rs` does cure this (all to be removed once resolved properly, of course).

tschwinge:
Instead of absolute line numbers, for example `rust.test/xfail_compile/continue1.rs`:

    /* { dg-error "cannot `continue` outside of a loop" "" { target { *-*-* } } 8 } */
    /* { dg-error "failed to type resolve expression" "" { target { *-*-* } } 8 } */

... you may use relative numbers; move these two lines to lines 9, 10 (or rather then: lines 7, 8), and use the `.-N` syntax (also `.+N`, if applicable`):

    /* { dg-error "cannot `continue` outside of a loop" "" { target { *-*-* } } .-1 } */
    /* { dg-error "failed to type resolve expression" "" { target { *-*-* } } .-2 } */

tschwinge:
In `rust.test/execute/exit_error.rs` add `dg-xfail-run-if` to turn `FAIL: rust.test/execute/exit_error.rs   -O0  execution test` etc. into `XFAIL` (as this test indeeed is expected to fail per the current implementation).

tschwinge:
As the `gcc/testsuite/rust.test/*/*.exp` files are (a) simple and (b) new, I'd use just `Copyright (C) 2021 [...]`, and remove the `This file was written by [...]` attributions.

At this time, I haven't reviewed the `*.exp` files in much detail (in particular, the pre-existing `gcc/testsuite/lib/rust-dg.exp`, `gcc/testsuite/lib/rust.exp`), but since you've removed more code than adding new one, that ought to be good.  ;-)

dkm:
Thanks very much for the comments, I'll refresh my changes and push something !

dkm:
I did some more cleanup:
- `rust-dg.exp` : remove the `rust-dg-runtest` as it was in fact not used.
- only use defaut options for xfail_compile tests
- use some relative `dg-error` directives (still have to cleanup remaining tests)
- fixed the xfail execute test

I'll continue this and hopefully have it ready when I can finally contribute something :)

gcc/testsuite/rust.test/xfail_compile/xfail_compile.exp
```diff
@@ -14,19 +14,21 @@
...
-
 #
-# Rust torture test suite.
+# Rust test suite.
+# These tests are expected to raise an error in the frontend
 #
 
 # Load support procs.
 load_lib rust-dg.exp
+load_lib torture-options.exp
```
tschwinge, line 24: Remove that, and...

gcc/testsuite/rust.test/xfail_compile/xfail_compile.exp
```diff
@@ -14,19 +14,21 @@
...
 
 # Initialize `dg'.
 dg-init
 
+# Only use the default options
+set-torture-options [list {}]
+
```
tschwinge, line 31: ... remove that: this `set-torture-options` override will leak into other DejaGnu `*.exp` files executed after this one in the same `runtest` process...  (Yes, this is _ugly_, I don't like this behavior...)

Instead of `gcc-dg-runtest`, you can simply use `dg-runtest` below.

tschwinge (changes requested):

gcc/testsuite/rust.test/xfail_compile/continue1.rs
```diff
@@ -1,11 +1,11 @@
-/* { dg-error "cannot `continue` outside of a loop" "" { target { *-*-* } } 8 } */
-/* { dg-error "failed to type resolve expression" "" { target { *-*-* } } 8 } */
 fn main() {
```
tschwinge, line 1: Hmm, I still need `dg-excess-errors` in this file, to avoid the corresponding `FAIL`.

gcc/testsuite/rust.test/xfail_compile/xfail_compile.exp
dkm: Will do !

gcc/testsuite/rust.test/xfail_compile/continue1.rs
dkm: That's what I would like to better understand. Sometime, simply moving dg-error directives raise the excess-error. I don't really understand exactly why it's needed for some tests and for some others not. It's strange that you need it, I'll rerun it to check that's not needed on my side..

dkm: After re-running with this single commit, I have the same behavior. So, not really sure what caused some previous run to be ok without the `dg-excess-errors`.

dkm:
I've cleaned all the test to use relative line ref where applicable and also used the C++ style comment instead of the C one.
I get this :
```
                === rust Summary ===

# of expected passes            769
# of expected failures          27
```

dkm:
Ok, I finally understand the problem with `continue1.rs` where I didn't have the same behavior as @tschwinge . In my dev branch I have #252 applied and the message 
```
rust1: error: expected () got <tyty::error>
```
is not here anymore.

Not really sure if this is a problem in #252 or simply a side-effect. The error message is not very clear :). I guess it's armless

tschwinge:
@dkm: nice work -- let me suggest that you remove _draft_ status from this pull request.  :-)

dkm:
Thanks ! In fact, I was leaving draft on my non-trivial PRs until I get the FSF document signed (should be the case before end of the week, hopefully) to avoid their merge by error...

dkm:
Here's what I get on current master:
```
		=== rust Summary ===

# of expected passes		769
# of expected failures		27
```

PR #302, opened 2021-03-24
Fix error while unexpected EOF when scanning for end of comment

dkm:
Emit an error when EOF is reached before end of C-style comment.

fix #300

PR #315, opened 2021-03-27
Fix test after several PR were merged

dkm:
the PR #302 and #301 were conflicting because they are both modifying the tests.
Move the conflicting test and add correct dg-error directive.

### DIFF
--- a/gcc/testsuite/rust.test/xfail_compile/continue1.rs
+++ b/gcc/testsuite/rust.test/xfail_compile/continue1.rs
@@ -1,4 +1,3 @@
-// { dg-excess-errors "Noisy error and debug" }
 fn main() {
     let mut a = 1;
     let mut b = 1;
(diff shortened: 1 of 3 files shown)


dkm:
I'll test this with an updated tree and check the tests are correct

dkm:
```
                === rust Summary ===

# of expected passes            793
# of expected failures          26
```

Also fixed the `continue1.rs` test.

philberty:
Nice one @dkm do you have the ability to merge now? If so you should give it a go :)

philberty (approved):

dkm:
> Nice one @dkm do you have the ability to merge now? If so you should give it a go :)

I don't seem to have this kind of power :)

dkm:
> > Nice one @dkm do you have the ability to merge now? If so you should give it a go :)
> 
> I don't seem to have this kind of power :)

Scratch that, I have permission to merge... Not sure if you added them or if I missed it yesterday :)

PR #318, opened 2021-03-28
Proof of concept 2: First-class s-expression, inline visitors, and polymorphic values

YizhePKU:
Here I present three pieces of "experimental technology" that may be useful for the project. I took the idea from various sources and implemented them from scratch. Maybe some of these can make it into the codebase? I'd love to hear your thoughts and suggestions.

# First-class s-expression

Taking inspiration from [nlohmann's excellent JSON library](https://github.com/nlohmann/json), this new design stores an s-expression using an intermediate data structure. This enables the use of braces to construct compound expressions. Further more, custom conversion can be implemented via functions instead of methods, so no more polluting the class interface:)

Example:

```c++
// use sexp() as a conversion function
auto e1 = sexp("hello");

// serialize by calling to_string()
auto s1 = e1.to_string(); // "hello"

// use brace initializer lists to create compound expressions
// supported datatypes are strings, integers, and floats
auto e2 = sexp{e1, 123, {"inner", "expression"}};
auto s2 = e2.to_string(); // "(hello 123 (inner expression))

// define conversion function for custom types
struct MyArray {
    std::vector<std::string> items;
};

sexp to_sexp(const MyArray& arr) {
    return sexp{"MyArray", {"item_count", items.size}, {"items", items}};
}
```

This design also plays nicely with visitor patterns. A custom conversion function only needs to be defined for the base type, either implemented as you normally would with visitor patterns, or with inline visitors(see below).

# Inline visitors

A lot of my complaints about visitors are due to their verbosity--for every operation on an ADT, you have to write an "Resolver" class. [Inline visitors](https://members.accu.org/index.php/articles/2021) make operations on ADTs much more lightweight.

Example:

```c++
// The usual setup. Skip to main() for the interesting part.
class BaseVisitor {
  public:
    virtual void visit(DerivedA& a) = 0;
    virtual void visit(DerivedB& b) = 0;
}

class Base {
  public:
    virtual ~Base() = default;

    virtual accept_vis(BaseVisitor& vis) = 0;
};

class DerivedA : public Base {
  public:
    int some_operation() { return 42; }

    virtual void accept_vis(BaseVisitor& vis) override { vis.visit(*this); }
}

class DerivedB : public Base {
  public:
    int some_other_operation() { return 13; }

    virtual void accept_vis(BaseVisitor& vis) override { vis.visit(*this); }
}

int main() {
    Base *obj = new DerivedA();

    // define an inline visitor
    // can use captured variables to return results
    int ret;
    auto vis = begin_visitor<BaseVisitor>()
                 .on<DerivedA>([&ret](DerivedA& a) {
                     ret = a.some_operation();
                 })
                 .on<DerivedB>([&ret](DerivedB& b) {
                     ret = b.some_other_operation();
                 })
                 .end_visitor();

    // apply the visitor as usual
    obj->accept_vis(vis);
}
```

# Polymorphic values

As a compiler project we're mostly concerned with tree-like ADTs, such as AST and HIR. To implement the visitor pattern, we store sub-trees using pointers, instead of composing values directly. On the other hand, we would like these data to be copy-able. Unfortunately, that means implementing copy constructors and `clone()` methods for every single data class. It is a mess.

Wouldn't it be nice if pointers could copy their underlying value automatically? That's what [polymorphic values](https://github.com/jbcoe/polymorphic_value) accomplish. It's a wrapper around `std::unique_ptr` that implements deep copying for you.

Example:

```c++
class Base {
  public:
    virtual ~Base() = default;
    virtual std::string get_content() = 0;
};

class Derived : public Base {
  private:
    std::string _content;
  public:
    Derived(const std::string& content) : _content(content) {}

    virtual std::string get_content() override { return _content; }
};

int main() {
    // create a polymorphic value
    poly<Base> obj = make_poly<DerivedA>("Hello");
    obj->get_content(); // "Hello"

    // copy a polymorphic value
    auto obj2 = obj; // calls Derived's copy constructor
    obj2->get_content(); // "Hello"

    // you can put poly inside structs or containers, and it just works
    auto arr = std::vector<poly<Base>>{obj, obj2}; // call copy constructor twice
}
```

---

With these experiments done, I'll get back to the usual bug fixing and house keeping. I'll also be preparing for my GSoC proposal--probably on the topic of HIR dump.

### DIFF
--- a/gcc/rust/util/rust-inline-visitor.h
+++ b/gcc/rust/util/rust-inline-visitor.h
@@ -1,0 +1,63 @@
+/* An implementation of the inline visitor.
+   https://members.accu.org/index.php/articles/2021 */
+
+#include <utility>
+
+#ifndef RUST_INLINE_VISITOR
+#define RUST_INLINE_VISITOR
+
+namespace Rust {
+
+/* VisitorWrapper is a wrapper over a (possibly incomplete) Visitor.
+   Add another visit() method by calling on(). These lambdas are stored
+   in _args as a linked list, and passed to the actual visitor when
+   end_visitor() is called. */
+template <typename BaseVisitor, typename Args> struct VisitorWrapper
+{
+  Args _args;
+
+  VisitorWrapper (Args &&args) : _args (args) {}
+
+  // The actual visitor as an inner type.
+  // Each visitor inherits from the last one, and implements one more visit().
+  template <typename T, typename F> struct Visitor : BaseVisitor
+  {
+    F _f;
+    Visitor (std::pair<F, Args> args)
+      : BaseVisitor (args.second), _f (args.first)
+    {}
+
+    using BaseVisitor::visit;
+    virtual void visit (T &t) final override { _f (t); }
+  };
+
+  template <typename T, typename F>
+  VisitorWrapper<Visitor<T, F>, std::pair<F, Args>> on (F &&f)
+  {
+    return VisitorWrapper<Visitor<T, F>, std::pair<F, Args>> (
+      std::make_pair (f, _args));
+  }
+
+  BaseVisitor end_visitor () { return BaseVisitor (_args); }
+};
+
+/* An empty visitor. A wrapper around the target Visitor we're matching against.
+   Consumes the final nullptr of the linked list. */
+template <typename TargetVisitor> struct EmptyVisitor : TargetVisitor
+{
+  EmptyVisitor (std::nullptr_t ptr) {}
+
+  using TargetVisitor::visit;
+};
+
+// The entry point
+template <typename TargetVisitor>
+VisitorWrapper<EmptyVisitor<TargetVisitor>, std::nullptr_t>
+begin_visitor ()
+{
+  return VisitorWrapper<EmptyVisitor<TargetVisitor>, std::nullptr_t> (nullptr);
+}
+
+} // namespace Rust
+
+#endif
--- a/gcc/rust/util/rust-poly.h
+++ b/gcc/rust/util/rust-poly.h
@@ -1,0 +1,112 @@
+/* An implementation of polymophic_value for C++11.
+   https://github.com/jbcoe/polymorphic_value */
+
+#include <memory>
+#include <type_traits>
+
+#include "rust-make-unique.h"
+
+#ifndef RUST_POLY
+#define RUST_POLY
+
+namespace Rust {
+
+template <bool U>
+using enable_if_t = typename std::enable_if<U, std::nullptr_t>::type;
+
+/* Cloner acts as a type-erasure device.
+   The concrete type, U, is erased.
+   However, clone() still creates an instance of U. */
+template <typename T> class Cloner {
+public:
+  virtual ~Cloner() = default;
+  virtual std::unique_ptr<Cloner<T>> clone() = 0;
+  virtual T *get() = 0;
+};
+
+template <typename T> class ConcreteCloner : public Cloner<T> {
+private:
+  static_assert(std::is_copy_constructible<T>::value,
+                "T is not copy constructible");
+
+  std::unique_ptr<T> _t;
+
+public:
+  ConcreteCloner(std::unique_ptr<T> &&t) : _t(std::move(t)) {}
+
+  virtual std::unique_ptr<Cloner<T>> clone() override {
+    auto t2 = Rust::make_unique<T>(*_t);
+    return Rust::make_unique<ConcreteCloner<T>>(std::move(t2));
+  }
+
+  virtual T *get() override { return _t.get(); }
+};
+
+template <typename T, typename U,
+          typename Rust::enable_if_t<std::is_base_of<T, U>::value> = nullptr>
+class DelegateCloner : public Cloner<T> {
+private:
+  std::unique_ptr<Cloner<U>> _worker;
+
+public:
+  DelegateCloner(std::unique_ptr<Cloner<U>> worker)
+      : _worker(std::move(worker)) {}
+
+  virtual std::unique_ptr<Cloner<T>> clone() override {
+    return Rust::make_unique<DelegateCloner<T, U>>(_worker->clone());
+  }
+
+  virtual T *get() override { return _worker->get(); }
+};
+
+/* Value wrapper that supports polymophic dispatch.
+   It happens to be implemented on top of std::unique_ptr, but
+   conceptually it's a value, not a pointer. In particular,
+   nullptr is not allowed. As a wrapper, poly is immutable--its
+   binding cannot be changed once created.
+ */
+template <typename T> class poly {
+public:
+  std::unique_ptr<Cloner<T>> _cloner;
+
+  // Construct from raw pointer
+  // Will take ownership of the value.
+  poly(std::unique_ptr<T> &&t) {
+    if (t == nullptr) {
+      throw std::runtime_error("Attempting to create a poly from nullptr");
+    }
+    _cloner = Rust::make_unique<ConcreteCloner<T>>(std::move(t));
+  }
+
+  // Copy constructor
+  poly(const poly &rhs) { _cloner = rhs._cloner->clone(); }
+
+  // Move constructor
+  poly(poly &&rhs) noexcept : _cloner(std::move(rhs._cloner)) {}
+
+  // Covariant "copy" constructor
+  template <typename U,
+            typename Rust::enable_if_t<std::is_base_of<T, U>::value> = nullptr>
+  poly(const poly<U> &rhs)
+      : _cloner(Rust::make_unique<DelegateCloner<T, U>>(rhs._cloner->clone())) {
+  }
+
+  // Covariant "move" constructor
+  template <typename U,
+            typename Rust::enable_if_t<std::is_base_of<T, U>::value> = nullptr>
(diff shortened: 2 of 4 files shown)


SimplyTheOther:
This all looks really cool. 

Unfortunately, there do seem to be some issues with polymorphic value types at least:

- The description for the repository says C++14. From a glance at the code, I can see that it at least uses `make_unique` (although I suppose that can be fairly trivially implemented for C++11). 
- It uses exceptions, which GCC doesn't allow. I also suppose that these can be removed and replaced with `assert` or something.
- It would have somewhat more overhead than `unique_ptr` - there is an abstract class stored in `poly` in addition to the pointer, so that's at least double the space requirement, and also has a cost for virtual lookup (although I suppose the current implementation also has that cost).
- It uses RTTI (although this can be removed, I think, with potential loss of runtime type safety)

The other issue is that I'm not sure how it would interact with the AST structure - things like having multiple levels of "cloneability" (e.g. Type vs TypeNoBounds), multiple inheritance from multiple different "cloneable" types (e.g. Expr and Type), etc.

YizhePKU:
> Accessing the pointer's value requires a dereference, a virtual lookup, and another dereference. This probably adds up.

I haven't tried it yet, but I think it's possible to store the pointer in the outer-most `poly` class instead of `Cloner`. That way we only incur a runtime cost when cloning.

> I'm not sure why you would add a "can't be reseated" aspect to poly. This would prevent the use of move or copy assignment. Wouldn't this prevent usage in a vector, since move or copy assignment is required?

You're right. I didn't think too much when making that decision. Implementing copy assignment and move assignment should be easy.

philberty:
This is pretty interesting nice work if you can solve @SimplyTheOther 's comment can you show what you imagine an example usage of this would look like on the AST for example? You don't have to have a fully compilable example just a draft of your idea. As always nice work.

YizhePKU:
> Can you show what you imagine an example usage of this would look like on the AST for example?

Suppose we would like to dump an AST tree as s-expressions. Here's how we can achieve that:

1. Create a new file in the AST folder, maybe `rust-ast-sexp.h`
2. Implement `to_sexp` method on some base type, say `AST::Expr`, using inline visitors:

```c++
sexp to_sexp(Expr& expr) {
    sexp result;
    auto vis = begin_visitor<ASTVisitorBase>()
               .on<ArrayExpr>([&result](ArrayExpr& arr) {
                   result = sexp{"ArrayExpr", {"elements", arr.get_array_elements()}};
               })
               .on<TupleExpr>([&result](TupleExpr& tuple) {
                   result = sexp{"TupleExpr", tuple.get_tuple_elems()};
               })
               /* and many others */
               .end_visitor();
    expr.accept_vis(vis);
    return result;
}
```

3. Then we can convert any `AST::Expr` into s-expression by calling `sexp()` on it. 

Because this works recursively, once we implement `to_sexp` on `Crate` we can dump the whole crate just by calling `sexp(crate).to_string()`.

---

Originally I wanted to make a demonstration as part of the PR, but these implementations took me long enough that I thought it would be best to publish and get some feedback first.

philberty:
Interesting, how do you handle the case of an arithmetic expression where you will need to visit recursively? Does the inline visitor need to have a 2nd level of inline visitors?

YizhePKU:
> Interesting, how do you handle the case of an arithmetic expression where you will need to visit recursively? Does the inline visitor need to have a 2nd level of inline visitors?

Do you mean e.g. `OperatorExpr` then `ArithmeticOrLogicalExpr`? Splitting into two levels is possible(demonstrated below), but I don't see why it's necessary. `ASTVisitor` is already flat. We can just ignore `OperatorExpr` and implement for concrete classes only.

```
sexp to_sexp(Expr& expr) {
    sexp result;
    auto vis = begin_visitor<ASTVisitorBase>()
               .on<OperatorExpr>([](OperatorExpr& expr) {
                   result = to_sexp(expr);
               })
               /* and other level 1 stuff */
               .end_visitor();
    expr.accept_vis(vis);
    return result;
}

sexp to_sexp(OperatorExpr& expr) {
    sexp result;
    auto vis = begin_visitor<ASTVisitorBase>()
               .on<ArithmeticOrLogicalExpr>([](ArithmeticOrLogicalExpr& expr) {
                   /* implement actual logic */
               })
               /* and other level 2 stuff */
               .end_visitor();
    expr.accept_vis(vis);
    return result;
}
```

philberty:
I was wondeing about this piece:

```c++
sexp to_sexp(Expr& expr) {
    sexp result;
    auto vis = begin_visitor<ASTVisitorBase>()
               .on<OperatorExpr>([](OperatorExpr& expr) {
                   sexpr lhs = ....;
                   sexpr rhs = ....;
                   result = to_sexp(expr.operator.to_string(), lhs, rhs);
               })
               .end_visitor();
    expr.accept_vis(vis);
    return result;
}
```

How do you get the lhs and rhs and create an sexpr from the elemt parts here?

YizhePKU:
> How do you get the lhs and rhs and create an sexpr from the elemt parts here?

We call `sexp{expr.operator.to_string(), lhs, rhs}`. `lhs` and `rhs` will be converted to sexp automatically.

> is there a constructor for sexpr to contain the operator and a list of sexprs?

I'm not sure what you meant. Do you mean converting a `std::vector<sexp>` to a single `sexp`? We can call `sexp` on that vector. In fact we don't even have to build that vector beforehand. We can call `sexp` on any `std::vector<T>` where T can be converted to `sexp`.

philberty:
Just to clarify one last thing in the example before would you call this?

```c++
sexp to_sexp(Expr& expr) {
    sexp result;
    auto vis = begin_visitor<ASTVisitorBase>()
               .on<OperatorExpr>([](OperatorExpr& expr) {
                    sexpr lhs = to_sexpr(expr.lhs);
                    sexpr rhs = to_sexpr(expr.rhs);
                   result = sexp(expr.operator.to_string(), lhs, rhs);
               })
               .end_visitor();
    expr.accept_vis(vis);
    return result;
}
```

to get the sexpr for lhs and rhs like above? Or do you call a visitor again?

Overall I can see some utility in inline visitors for name and type resolution in order to do some operations.  I would open for this to move forward so long as the existing Visitor path are uneffected and these work side by side so we can opt into using this if we wish.

Whats your opinion @SimplyTheOther,

gcc/rust/util/rust-inline-visitor.h
```diff
@@ -0,0 +1,63 @@
...
+   https://members.accu.org/index.php/articles/2021 */
+
+#include <utility>
+
+#ifndef RUST_INLINE_VISITOR
+#define RUST_INLINE_VISITOR
+
+namespace Rust {
+
+/* VisitorWrapper is a wrapper over a (possibly incomplete) Visitor.
```
philberty, line 11: Style nit here: I prefer comments to be // for multiline. Clang format helps here.

gcc/rust/util/rust-poly.h
```diff
@@ -0,0 +1,111 @@
+/* An implementation of polymophic_value for C++11.
+   https://github.com/jbcoe/polymorphic_value */
```
philberty, line 2: is your implementation sufficiently different from the reference that it won't cause copyright assignment issues?

gcc/rust/util/rust-sexp.h
```diff
@@ -0,0 +1,215 @@
...
+  std::vector<sexp> value;
+
+  template <typename It>
+  SList (It begin, It end) : value (std::vector<sexp> (begin, end))
+  {}
+
+  virtual void accept_vis (SValueVisitor &vis) override { vis.visit (*this); }
+};
+
+class sexp
```
philberty, line 100: this class is excellent

gcc/rust/util/rust-sexp.h
```diff
@@ -0,0 +1,215 @@
+#include <algorithm>
```
philberty, line 1: you need to move the includes below the ifdef guards

YizhePKU:
> Would you call this to get the sexpr for lhs and rhs like above? Or do you call a visitor again?

We don't have to call a visitor again. Your code sample has the right idea. It should look like this:

```c++
sexp to_sexp(Expr& expr) {
    sexp result;
    auto vis = begin_visitor<ASTVisitorBase>()
               .on<OperatorExpr>([](OperatorExpr& expr) {
                   result = sexp{expr.operator.to_string(), expr.lhs, expr.rhs};
               })
               .end_visitor();
    expr.accept_vis(vis);
    return result;
}
```

gcc/rust/util/rust-poly.h
YizhePKU: Currently it looks quite similar, but as discussed earlier this implementation has a runtime overhead. I might end up rewriting it again to solve the performance problem(which should make it distinct enough).

YizhePKU:
These implementation will need some cleanup and testing before they're ready.

I have some bad news though. My school is somewhat unwilling to sign my copyright assignment paper, not because they object but because they don't understand why it's necessary. I'll sort it out eventually but it will take time.

YizhePKU:
I'm closing this due to inactivity.

* Inline visitors have been merged into master. 
* I'll improve polymorphic values and use it to simplify trees later.
* As for s-expressions, I may or may not use it in my GSoC project; it's certainly the least useful one in general.

PR #319, opened 2021-03-29
Fix infinite loop when expanding cfg attributes

### LINKED ISSUE #320: Infinite loop when expanding cfg_attr macro
The following program produces an infinite loop:

```rust
#[cfg_attr(feature = "somefeature", attribute = "someattr")]
struct Feature;

fn main() {
}
```

CohenArthur:
Well turns out the issue was very simple to fix and probably a typo :D

Should the test case be moved to compilable? Considering the features don't actually exist, but the syntax is correct.

Closes #320


philberty:
Can you remove this test case from the other PR then? Also, can you create an associated issue for this PR for the bug it fixes?

Nice work!

philberty (approved):

philberty:
perfect moving this to the xfail section should solve the build issue I should have mentioned that earlier. Sorry.

CohenArthur:
> perfect moving this to the xfail section should solve the build issue I should have mentioned that earlier. Sorry.

No worries!

Issue #311, opened 2021-03-27
Support already aubstituted types with Impl blocks


philberty:
Rust allows for an impl block to be implemented on a fully substituted type for example:

```rust
struct GenericStruct<T>(T, usize);

impl GenericStruct<i32> {
    fn new(a: i32, b: usize) -> Self {
        GenericStruct(a, b)
    }

    fn get(self) -> i32 {
        self.0
    }
}

fn main() {
    let a: GenericStruct<i32> = GenericStruct::<i32>::new(123, 456);
    let aa: i32 = a.get();
}

```

philberty:
This is a better test case:

```
struct Foo<A = (isize, char)> {
    a: A,
}

impl Foo<isize> {
    fn bar(self) -> isize {
        self.a
    }
}

impl Foo<char> {
    fn bar(self) -> char {
        self.a
    }
}

impl Foo {
    fn bar(&self) {
        let a: (isize, char) = self.a;
        let b = a.0;
        let c = a.1;

        let aa: Foo<isize> = Foo { a: b };
        let bb: isize = aa.bar();
    }
}

fn main() {
    let a = Foo { a: (123, 'a') };
    a.bar();
}
```

PR #322, opened 2021-03-30
Initial TypeAlias support

### LINKED ISSUE #312: Type Alias Support
https://doc.rust-lang.org/std/keyword.type.html

philberty:
commit 42e13fb08b14f638ca1becd3ca9386972648f372 (HEAD -> phil/generics4, origin/phil/generics4)
Author: Philip Herron <philip.herron@embecosm.com>
Date:   Tue Mar 30 17:53:09 2021 +0100

    Add support for TypeAlias
    
    This allows for a typedef style alias of any type in rust code, as the
    types are abstracted behind the TyTy interface this falls out quite nicely.
    
    In order to support TypeAliases we needed to be able to supported already
    substituted types in #311.
    
    More testing is needed here the GenericParameters on TypeAlias are not
    being used in the type resolution pass here.
    
    Fixes #312
    Addresses #311

commit 52e63128c5491c5432960b62e664d7ac479d140a
Author: Philip Herron <philip.herron@embecosm.com>
Date:   Tue Mar 30 18:23:53 2021 +0100

    Fix bad tracking on used substitutions on generic types.
    
    When implementing generics we need to keep track of used arguments to
    recursively substitute ParamTypes when nessecary. This also applies for
    cases such as: Foo<i32> where we can check wether we need to actually
    perform a substitution.

### DIFF
--- a/gcc/testsuite/rust.test/xfail_compile/type-alias1.rs
+++ b/gcc/testsuite/rust.test/xfail_compile/type-alias1.rs
@@ -1,0 +1,6 @@
+type TypeAlias = (i32, u32);
+
+fn main() {
+    let a: TypeAlias;
+    a = (123, 456f32); // { dg-error "expected .u32. got .f32." }
+}
--- a/gcc/rust/ast/rust-item.h
+++ b/gcc/rust/ast/rust-item.h
@@ -1720,6 +1720,8 @@ public:
     return existing_type;
   }
 
+  Identifier get_new_type_name () const { return new_type_name; }
+
 protected:
   /* Use covariance to implement clone function as returning this object
    * rather than base */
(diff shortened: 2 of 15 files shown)


gcc/testsuite/rust.test/xfail_compile/type-alias1.rs
```diff
@@ -0,0 +1,6 @@
+type TypeAlias = (i32, u32);
+
+fn main() {
+    let a: TypeAlias;
+    a = (123, 456f32); // { dg-error "expected .u32. got .f32." }
```
philberty, line 5: @dkm is this the correct way to do this the error message is:

```
test.rs:5:15: error: expected [u32] got [f32]
    5 |     a = (123, 456f32); // { dg-error "expected .u32. got .f32." }
      |
```

dkm: Looks correct yes. You'll quickly see if it works correctly as the test will FAIL if the regex does not match an error message at this exact line !

PR #328, opened 2021-03-31
Fix CI checks

dkm:
Checking for 'unexpected' will also catch XFAIL tests being PASS.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -46,8 +46,17 @@ jobs:
            cd gccrs-build; \
            make -j $(nproc)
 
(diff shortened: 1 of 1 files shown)


philberty (approved):

philberty:
Nice one I didn't realise gcc/testsuite/rust/rust.sum was a log file for a long time :).

dkm:
You have rust.log and rust.sum that are created. They are correctly merged in case of parallel testing (instead of filtering the output of `make check`). The .sum only contains the tests status and can be used by non-regression script (dg-compare in gcc for example) :)

Issue #336, opened 2021-04-02
GCC '--enable-bootstrap' build


tschwinge:
If building a native toolchain, GCC by default does a 3-stage bootstrap build (<https://gcc.gnu.org/install/configure.html>).  In addition to making sure that GCC is able to reproduce itself bit-by-bit, this also means that stages 2+ are built with `-Werror` (turning most *warning* into *error* diagnostics; see `--enable-werror`, possibly enabled by default).  This helps to catch a good number of bugs, because it enforces that GCC compiles without compiler diagnostics; it's a requirement for upstream patch submission (<https://gcc.gnu.org/contribute.html#testing>).

The GCC/Rust instructions explicitly specify `--disable-bootstrap` (which of course is much faster):

  - <https://github.com/Rust-GCC/gccrs/blob/master/README.md>
  - <https://github.com/Rust-GCC/gccrs/blob/master/.github/workflows/ccpp.yml>
  - <https://github.com/Rust-GCC/gccrs/blob/master/Dockerfile>

I think it's OK to keep it that way (I'm also doing that for my day-to-day development), but at least as a background task, we should also make sure that `--enable-bootstrap` works -- which it currently doesn't.  I've produced a number of patches, that I'm going to submit individually.

philberty:
We could add a workflow to do bootstrap builds on master only so it is not required on PRs

PR #343, opened 2021-04-02
Address '[...]/gcc/rust/lex/rust-lex.cc:1322:21: error: comparison is always false due to limited range of data type [-Werror=type-limits]' diagnostic [#336]

tschwinge:
#336

Should this maybe be `unsigned char`, or something else changed?

We don't seem to have any testcases?


SimplyTheOther:
This seems to be the same issue as the other one, so I'll copy my response from there.

IIRC `char` is not specifically defined as `unsigned char` or `signed char` in the C++ standard - it is implementation-defined. As such, values allowed by `char` may (depending on target platform) be larger than 127. This is the reason for this check - ASCII characters that Rust requires do not have values larger than 127.

PR #344, opened 2021-04-02
Address '[...]/gcc/rust/lex/rust-lex.cc:1397:27: error: comparison is always false due to limited range of data type [-Werror=type-limits]' diagnostic [#336]

tschwinge:
#336

Should this maybe be `unsigned char`, or something else changed?

We don't seem to have any testcases?

### DIFF
--- a/gcc/rust/lex/rust-lex.cc
+++ b/gcc/rust/lex/rust-lex.cc
@@ -1394,7 +1394,7 @@ Lexer::parse_byte_string (Location loc)
 	  else
 	    length += std::get<1> (escape_length_pair);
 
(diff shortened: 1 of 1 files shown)


SimplyTheOther:
IIRC `char` is not specifically defined as `unsigned char` or `signed char` in the C++ standard - it is implementation-defined. As such, values allowed by `char` may (depending on target platform) be larger than 127. This is the reason for this check - ASCII characters that Rust requires do not have values larger than 127.

tschwinge:
Closed #344 via 75be04262789545634d6459235b249b01560742c.

PR #349, opened 2021-04-02
Address '[...]/gcc/rust/rust-target.h:23: error: "DEFHOOK" redefined' diagnostic [#336]

tschwinge:
#336

I haven't quickly been able to understand the target hooks handling here -- seems to be WIP?

My proposed changed doesn't change any behavior.

### DIFF
--- a/gcc/rust/rust-session-manager.cc
+++ b/gcc/rust/rust-session-manager.cc
@@ -28,7 +28,17 @@
 #include "tm.h"
 #include "tm_p.h"
 
-#include "rust-target.h"
+//#include "rust-target.h"
+/*TODO This isn't (currently?) necessary, but if '#include'd after '#include "target.h"', causes:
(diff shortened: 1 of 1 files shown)


dkm:
I think I tried this fix in : https://github.com/Rust-GCC/gccrs/pull/273
but @philberty had reason for having this file included.

In my dev branch, I have fixed this error by removing the inclusion of `target.h`. You can see this here : https://github.com/dkm/gccrs/commit/fcdcee0dc2dd6559aadc270d217871aa15791a09
As I was still trying to understand how these files/hook should be used, I haven't opened a PR for it. I'll let you decide if it's a correct fix or not :)

philberty:
Yeah maybe we should just comment this out for now untill we end up using the target hooks the reason i had to leave it in as i was worried with merges with upstream gcc it would cause issues we might miss but have a working bootstrap build might be more important.

Issue #330, opened 2021-04-01
[WIP] Create liveness analysis for dead code detection.


thomasyonug:
This is a working issue for the 2021 GSoC. I'm little busy last two weeks, sorry about my late.

I had checked the method from rustc: [https://github.com/rust-lang/rust/blob/master/compiler/rustc_passes/src/dead.rs#L569](https://github.com/rust-lang/rust/blob/master/compiler/rustc_passes/src/dead.rs#L569)
That implementation separates two phases to detect unused code.
1. liveness analysis to get all reachable symbols.
2. code called from live codes are live， otherwise dead.

Implementation for our project.
1. first of all, we need a test for warning features. But it seems like currently our new test system doesn't supports warnings?
2. add an analysis dir which contains our Liveness class, and try to start after resolve.
3. once obtain liveness variables, we can simply scan all dead codes which not belongs to the live variable.
``` c++
  // type resolve
  Resolver::TypeResolution::Resolve (hir);
  if (options.dump_option_enabled (CompileOptions::TYPE_RESOLUTION_DUMP))
    {
      dump_type_resolution (hir);
    }

  // liveness analysis
  Analysis::LivenessAnalysis::liveness(hir);
  if (options.dump_option_enabled (CompileOptions::ANALYSIS_DUMP))
    {
      // I have no idea about what should be dumped. Maybe some bit-vectors?
    }

  if (saw_errors ())
    return;

  // scan deadcode
  Analysis::ScanDeadcode::scan();
  if (options.dump_option_enabled (CompileOptions::DEADCODE_DUMP))
    {
      // what should be dumped?
    }

  if (saw_errors ())
    return;

  // scan unused has to be done after type resolution since methods are resolved
  // at that point
  Resolver::ScanUnused::Scan ();

  if (saw_errors ())
    return;
```

Now, this is just an entry point design. I will continuously update any details on design liveness analysis and dead-code detection.
If anyone has any suggestions, plz just comment on this issue directly or @ me in zulip, thank you!

dkm:
>     1. first of all, we need a test for warning features. But it seems like currently our new test system doesn't supports warnings?

The current setup will not tolerate any warnings in `compile` directory by default. If you want to test for specific warning you can use `dg-warning` as documented here :  https://gcc.gnu.org/onlinedocs/gccint/Directives.html

I guess that's what you are looking for ?

thomasyonug:
Yes, dkm.This is very helpful, thanks.

[...]

PR #357, opened 2021-04-06
Create liveness analysis for dead code detection

thomasyonug:
This is the very beginning work for dead code detection. #330 
1. create a new test for this feature.
2. handle an extremely simple case.

After compiling the rust1, and feed "testsuite/rust.test/xfail_compile/unused.rs" to it.
``` rust

fn bar() { // {dg-warning "function is never used: `bar`"}
    foo();
}

fn foo() { // {dg-warning "function is never used: `foo`"}
    bar();
}

fn f() {

}

fn main() {
    f();
}

```
we will get some warnings.
``` bash
../gccrs/gcc/testsuite/rust.test/xfail_compile/unused.rs:2:1: warning: function is never used: `[bar]`
    2 | fn bar() { // {dg-warning "function is never used: `bar`"}
      | ^
../gccrs/gcc/testsuite/rust.test/xfail_compile/unused.rs:6:1: warning: function is never used: `[foo]`
    6 | fn foo() { // {dg-warning "function is never used: `foo`"}
      | ^

```

### DIFF
--- a/gcc/rust/analysis/rust-hir-liveness.cc
+++ b/gcc/rust/analysis/rust-hir-liveness.cc
@@ -1,0 +1,141 @@
+// Copyright (C) 2021 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#include "rust-hir-liveness.h"
+#include "rust-hir-full.h"
+#include "rust-name-resolver.h"
+
+namespace Rust {
+namespace Analysis {
+
+class FindEntryPoint : public LivenessBase
+{
+  using Rust::Analysis::LivenessBase::visit;
+
+public:
+  static std::vector<HirId> find (HIR::Crate &crate)
+  {
+    FindEntryPoint findEntryPoint;
+    for (auto it = crate.items.begin (); it != crate.items.end (); it++)
+      {
+	it->get ()->accept_vis (findEntryPoint);
+      }
+    return findEntryPoint.getEntryPoint ();
+  }
+
+  // TODO not only fn main can be a entry point.
+  void visit (HIR::Function &function) override
+  {
+    if (function.get_function_name () == "main")
+      {
+	entryPoints.push_back (function.get_mappings ().get_hirid ());
+      }
+  }
+
+private:
+  FindEntryPoint () : LivenessBase () {}
+  std::vector<HirId> entryPoints;
+  std::vector<HirId> getEntryPoint () { return entryPoints; }
+};
+
+std::set<HirId>
+Liveness::Analysis (HIR::Crate &crate)
+{
+  Liveness liveness (FindEntryPoint::find (crate));
+  liveness.go (crate);
+
+  return liveness.liveSymbols;
+}
+
+void
+Liveness::go (HIR::Crate &crate)
+{
+  while (!worklist.empty ())
+    {
+      HirId hirId = worklist.back ();
+      worklist.pop_back ();
+      scannedSymbols.emplace (hirId);
+      HIR::Item *item
+	= mappings->lookup_hir_item (crate.get_mappings ().get_crate_num (),
+				     hirId);
+      if (item == nullptr)
+	continue;
+      liveSymbols.emplace (hirId);
+      item->accept_vis (*this);
+    }
+}
+
+void
+Liveness::visit (HIR::ExprStmtWithoutBlock &stmt)
+{
+  stmt.get_expr ()->accept_vis (*this);
+}
+
+void
+Liveness::visit (HIR::CallExpr &expr)
+{
+  expr.get_fnexpr ()->accept_vis (*this);
+}
+
+void
+Liveness::visit (HIR::PathInExpression &expr)
+{
+  NodeId ast_node_id = expr.get_mappings ().get_nodeid ();
+  NodeId ref_node_id = UNKNOWN_NODEID;
+  if (resolver->lookup_resolved_name (ast_node_id, &ref_node_id))
+    {
+      Resolver::Definition def;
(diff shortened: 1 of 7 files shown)


gcc/rust/rust-session-manager.cc
```diff
@@ -547,6 +549,22 @@ Session::parse_file (const char *filename)
       dump_type_resolution (hir);
     }
 
+  // liveness analysis
+  std::set<HirId> live_symbols = Analysis::Liveness::Analysis (hir);
+  if (options.dump_option_enabled (CompileOptions::ANALYSIS_DUMP))
+    {
+      // I have no idea about what should be dumped. Maybe some bit-vectors?
```
philberty, line 556: This is an interesting one thanks for thinking about it. We also have an incomplete name resolution dump. What if we simply emit the IDs for the live/dead code sets. HIR dumps when complete should/must contain the NodeMappings information which can be used for reference.

philberty: Dumps should be a separate PR/commit i think

gcc/rust/analysis/rust-hir-liveness.cc
```diff
@@ -0,0 +1,154 @@
...
+{
+  expr.get_fnexpr ()->accept_vis (*this);
+}
+
+void
+Liveness::visit (HIR::PathInExpression &expr)
+{
+  NodeId ast_node_id = expr.get_mappings ().get_nodeid ();
+  NodeId ref_node_id = UNKNOWN_NODEID;
+  if (resolver->lookup_resolved_name (ast_node_id, &ref_node_id))
```
philberty, line 113: This piece is a bit confusing with the definition piece, I am currently working on cleaning up how this works well done for getting this bit right.

gcc/rust/analysis/rust-hir-liveness.cc
```diff
@@ -0,0 +1,154 @@
...
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#include "rust-hir-liveness.h"
+#include "rust-hir-full.h"
+#include "rust-name-resolver.h"
+
+extern bool
+saw_errors (void);
```
philberty, line 24: I don't think you need this extern symbol here.,

philberty:
Very well done @thomasyonug this is a great first step! What is the status of your copyright assignment stuff?

gcc/testsuite/rust.test/xfail_compile/unused.rs
```diff
@@ -0,0 +1,16 @@
+
+fn bar() { // {dg-warning "function is never used: `bar`"}
```
philberty, line 2: @dkm do you think this test case should be moved into compile? Since this is simply about warnings

thomasyonug:
> Very well done @thomasyonug this is a great first step! What is the status of your copyright assignment stuff?

Thank you, Phil. I had sent all required files to FSF about two weeks ago. I thought the process is finished before, but Craig from FSF told me they can't download the files from google driver today.

So I have to separate 3 emails to attach my signed pdf because of the size problem.

I believe everything is going fine this time.

gcc/testsuite/rust.test/xfail_compile/unused.rs
dkm: I guess so yes. `xfail` should really be about compiler returning with an error

dkm (changes requested):
As hinted by @philberty the test should be moved to compile directory :)

philberty:
bors ping

philberty:
bors try

gcc/rust/rust-session-manager.cc
thomasyonug: Yeah, this is a sparse analysis currently, dump hirid for live/dead code sets might be a suitable way. I will separate this dump part for other PR.

gcc/testsuite/rust.test/xfail_compile/unused.rs
thomasyonug: I had moved unused.rs to compile dir.

tschwinge: > Okay, is there any chance to create a new test dir for specific warning outputs? Because I want to make sure that the test will fail if the dg-warning doesn't match the real compiler warning.

We don't need a new category/directory for that; it should already behave in that way (have you checked?) -- if not, that's a bug in the test harness.

philberty:
bors try

philberty (approved):

dkm (approved):

dkm:
wooops, we should have squashed the changes before merging? There are 2 merges and a WIP commit

dkm:
What do you thing @philberty ?

philberty:
I agree with @dkm @thomasyonug can you rebase your PR against master and fix up the commit message? This looks good to merge to me or do you wish to keep this as a draft pr?

thomasyonug:
OK, I will remove wip and rebase my commit.

thomasyonug:
Hi, @philberty , @dkm .  Commit had rebased, and any problems now? After this merging, I would like to write more codes here. :)

dkm:
Without being too nick-picky, I guess a commit message describing a bit more what your changes are doing would be better?
You should also mentions the issue #330 .

thomasyonug:
> Without being too nick-picky, I guess a commit message describing a bit more what your changes are doing would be better?
> You should also mentions the issue #330 .

This is a good point. Let me figure out a suitable commit msg. Thanks.

philberty:
An example a commit message might be:

```
Add unused code pass

In order to find more examples of dead code we must create liveness variables and scan after type resolution. In the case, if a function is unused and it calls another function the 2nd function is now unused since the caller is not used etc. This is a WIP but demonstrates the pattern we can follow to add more dead code analysis warnings like rustc does.

Fixes #...
```

philberty:
Just as a heads up I think this PR should go in before mine which is nearly ready which changes PathInExpression a lot so it actually implements the Turbo fish properly, in theory, it should not affect the code you have written here but it would be easier for me to figure out any change in behaviour that might affect your code.

See #358

thomasyonug:
> An example a commit message might be:
> 
> ```
> Add unused code pass
> 
> In order to find more examples of dead code we must create liveness variables
> and scan after type resolution. In this case, if a function is unused and it calls
> another function the 2nd function is now unused since the caller is not used etc.
> This is a WIP but demonstrates the pattern we can follow to add more dead code
> analysis warnings like rustc does.
> 
> Addresses #...
> ```

Thanks for the example, I'll take it as my commit msg.

thomasyonug:
> Just as a heads up I think this PR should go in before mine which is nearly ready which changes PathInExpression a lot so it actually implements the Turbo fish properly, in theory, it should not affect the code you have written here but it would be easier for me to figure out any change in behaviour that might affect your code.
> 
> See #358

Cool! I'd love to help debug if any incorrect behaviours come up.

Issue #353, opened 2021-04-05
Need check for Inherent Impl Overlapping


philberty:
see rustc_typeck/src/coherence/inherent_impls_overlap.rs

```rust
struct Foo<A> {
    a: A,
}

impl Foo<isize> {

    fn bar(self) -> isize {
        self.a
    }
}

impl Foo<char> {

    fn bar(self) -> char {
        self.a
    }
}

impl<T> Foo<T> {
    
    fn bar(self) -> T {
        self.a
    }
}

fn main() {
    let a = Foo { a: 123 };
    a.bar();
}
```

```
error[E0592]: duplicate definitions with name `bar`
  --> src/main.rs:9:5
   |
9  |     fn bar(self) -> isize {
   |     ^^^^^^^^^^^^^^^^^^^^^ duplicate definitions for `bar`
...
23 |     fn bar(self) -> T {
   |     ----------------- other definition for `bar`
```
when we define a function/methods on impl blocks it can break method resolution if we have overlapping names. This occurs when there is a generic definition vs concrete definitions and only when names overlap

for example chaning the generic impl method name will compile

```rust
impl<T> Foo<T> {
   
    fn test(self) -> T {
        self.a
    }
}
```

Issue #335, opened 2021-04-02
Canonical Paths


philberty:
https://doc.rust-lang.org/reference/paths.html#canonical-paths

As part of my ongoing work into generic data type and blocks, I fuzzed over the name resolution for PathInExpression and TypePath to use AST::SimplePath which has limitations such as: #325 when you specify constants/methods on the impl block.

Path canonicalization will fix bugs in the name resolver to allow multiple impl blocks for same GenericType which has be substituted.

Issue #325, opened 2021-03-31
Duplicate symbol defined bug with methods


philberty:
Rust allows for impl blocks on generic types this means we have separate implementations for methods based on the receiver being used.

```rust
struct Foo<A> {
    a: A,
}

impl Foo<isize> {
    fn bar(self) -> isize {
        self.a
    }
}

impl Foo<char> {
    fn bar(self) -> char {
        self.a
    }
}

fn main() {
    let a = Foo { a: 123 };
    a.bar();
}
```

This looks like a scoping/rib problem in the name resolver.

philberty:
I just realised I have already followed most of this https://doc.rust-lang.org/reference/expressions/method-call-expr.html and do a probe to lookup methods I think I just need a tweak to my name resolver here in this case in theory the type resolution pass should work to resolve the correct method.

PR #358, opened 2021-04-07
Canonical Paths for Name Resolution and handle TurboFish properly

### LINKED ISSUE #355: Need check for multiple applicable items in scope
```rust
struct Foo<A> {
    a: A,
}

impl Foo<isize> {
    fn test() -> i32 {
        123
    }

    fn bar(self) -> isize {
        self.a
    }
}

impl Foo<char> {
    fn test() -> i32 {
        123
    }
    
    fn bar(self) -> char {
        self.a
    }
}

fn main() {
    let a:i32 = Foo::test();
}
```

This error only occurs when the compiler needs to resolve the PathExpr because it can resolve to multiple items.

```
error[E0034]: multiple applicable items in scope
  --> src/main.rs:29:22
   |
29 |     let a:i32 = Foo::test();
   |                      ^^^^ multiple `test` found
   |
note: candidate #1 is defined in an impl for the type `Foo<isize>`
  --> src/main.rs:8:5
   |
8  |     fn test() -> i32 {
   |     ^^^^^^^^^^^^^^^^
note: candidate #2 is defined in an impl for the type `Foo<char>`
```

philberty:
Add Canonical paths to name resolution
    
In order to support name resolution and checks for duplicate definitions
of names we need canonical paths for all DefId items such as inherent impl
items and normal items. Consider:

```rust
  struct Foo<T>(T);

  impl Foo<f32> {
    fn name()...
  }

  impl Foo<i32> {
    fn name()...
  }
```

Each of the impl blocks have a name function but these are separate due to
the concrete impl of the Parameter type passed in.

The caveat here is that to call this Function name the programmer must be
explicit in which implentation they wish to call such as:

```rust
  let a = Foo::<f32>::name();
```

This lets the Path probe lookup the appropriate impl block. The problem here
is that rust also allows for the compiler to infer the impl you wish such
as:

```rust
  let a = Foo::name();
```

This should fail since there are multiple candidates possible for this
Path. Unless there might have only been one name function in which case
it would have worked.

This patch is also responsible to implement PathInExpression by iterating
each segment and applying generic arguments as we go.

Fixes #355 #335 #325 #353

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.h
+++ b/gcc/rust/backend/rust-compile-expr.h
@@ -526,7 +526,7 @@ public:
 
   void visit (HIR::PathInExpression &expr) override
   {
-    translated = ResolvePathRef::Compile (&expr, ctx);
+    translated = ResolvePathRef::Compile (expr, ctx);
   }
 
   void visit (HIR::LoopExpr &expr) override
--- a/gcc/rust/backend/rust-compile-resolve-path.h
+++ b/gcc/rust/backend/rust-compile-resolve-path.h
@@ -30,10 +30,10 @@ class ResolvePathRef : public HIRCompileBase
   using Rust::Compile::HIRCompileBase::visit;
 
 public:
-  static Bexpression *Compile (HIR::Expr *expr, Context *ctx)
+  static Bexpression *Compile (HIR::PathInExpression &expr, Context *ctx)
   {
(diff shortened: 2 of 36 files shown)


philberty:
bors try

philberty:
bors try

philberty:
bors try

Issue #317, opened 2021-03-28
Invalid expression discarding when using semicolons


CohenArthur:
`rustc` does not produce any warnings for the following code:
```rust
fn main() {
    8;
}
```

(which is very different from this code, which does not compile with rustc either, obviously)
```rust
fn main() {
    8
}
```

What's happening is that, despite `8` being an "expression", the use of a semicolon afterwards allows not using it. This would be similar to something like
```rust
fn not_void() -> i32 {
    8
}

fn main() {
    not_void();
}
```
where discarding the return value of a function is fine, unless said function or type is marked with `#[must_use]`.

However, gccrs does not seem to be able to discard the value of an expression (and thus transforming said expression into a "statement"). Therefore, the snippets of code above fail with the following error:
```sh
/tmp/main.rs:1:1: error: expected [()] got [<integer>] # or `got [i32]` if we use the `not_void()` function
    1 | fn main() {
      | ^
```

I haven't found any issues about the subject, but might not have looked hard enough! If so, I apologize

YizhePKU:
I checked with a debugger. The syntax is parsed correctly(`8` is parsed as `ExprStmtWithoutBlock`, not as the final expression of the block). The error happened in the stage of type resolution.

YizhePKU:
The offending code is at `rust-hir-type-check.cc:82`. The implemented logic is that, if no trailing expression exists, the type of the last statement will be used as the type of the block, which is incorrect.

```c++
void
TypeCheckExpr::visit (HIR::BlockExpr &expr)
{
  TyTy::BaseType *block_tyty
    = new TyTy::TupleType (expr.get_mappings ().get_hirid ());

  expr.iterate_stmts ([&] (HIR::Stmt *s) mutable -> bool {
    bool is_final_stmt = expr.is_final_stmt (s);
    bool has_final_expr = expr.has_expr () && expr.tail_expr_reachable ();
    bool stmt_is_final_expr = is_final_stmt && !has_final_expr;

    auto resolved = TypeCheckStmt::Resolve (s, inside_loop);
    if (resolved == nullptr)
      {
	rust_error_at (s->get_locus_slow (), "failure to resolve type");
	return false;
      }

    if (stmt_is_final_expr)
      {
	delete block_tyty;
	block_tyty = resolved;
      }
    else if (!resolved->is_unit ())
      {
	rust_error_at (s->get_locus_slow (), "expected () got %s",
		       resolved->as_string ().c_str ());
      }

[...]

PR #364, opened 2021-04-11
Fix issues about block expression evaluation

lrh2000:
This PR introduces the `!` type (named "never", see https://doc.rust-lang.org/std/primitive.never.html), so that the following code works.
```rust
fn foo(c: i32) -> i32 {
    let b = if c > 0 {
        return 1;
    } else {
        c
    };
    b
}
```
Note the code also works even without the PR, however, it is an accident because `return 1` resolves to `<integer>` before this PR, which is not reasonable.

This PR is roughly complete:
- [x] The trailing expression is not necessarily without a block
- [x] Add termination check for GENERIC compiling
- [x] Reinforce termination check for HIR lowering
- [x] Introduce the `!` type (named "never")
- [x] Perform more code testing
- [x] Fix existing tests and add new tests
- [x] Document the code

TODOs (I'll open separate issues or PRs for them):
- Integration of the `!` type (example code: `fn foo() -> ! { /* ...*/}`)
- Detection of infinite loops (which should resolve to `!`)
- Lazy evaluation for lazy boolean operators (buggy code: `func() || return -1;` and `func() && { bar(); false };`)


philberty:
Nice work, I would suggest that you run the 3 xfail tests manually. They might be failing only because an error message might be updated/changed as part of this PR so you might find it is all working as you expect. I would also suggest that you add any test cases you can from the relevant issue: https://github.com/Rust-GCC/gccrs/issues/317

philberty:
Also I would like to hear any opinions you have on the code you had to work with on this change.

gcc/rust/typecheck/rust-tyty-rules.h
```diff
@@ -1069,12 +1107,66 @@ class StrRules : public BaseRules
...
 };
 
+class NeverRules : public BaseRules
+{
+  using Rust::TyTy::BaseRules::visit;
+
+public:
+  NeverRules (NeverType *base) : BaseRules (base), base (base) {}
+
+  virtual void visit (TupleType &type) override { resolved = type.clone (); }
```
philberty, line 1125: I would have assumed that the only methods you would implement for a never type would be InferType and NeverType as only those types can be unified. This class will allow ints to unify with never types and coerce ints to be !.

gcc/rust/typecheck/rust-tyty-rules.h
```diff
@@ -478,6 +486,8 @@ class InferRules : public BaseRules
     BaseRules::visit (type);
   }
 
+  void visit (NeverType &type) override { resolved = base->clone (); }
```
philberty, line 489: I think the only one that should be implementing this should be InferRules

gcc/rust/typecheck/rust-tyty-rules.h
lrh2000: Yes, but actually this problem does not exist now. It is because `!` has not been integrated and code like `let a: !` or `fn foo() -> ! { /* ... */ }` simply does not compile.

In the future, when we integrate the `!` type, I think we need to explicitly distinguish `unify` and `assign`. For example, currently both of the following two code segments will unify `!` and `<integer>`, but the first is legal and the second is illegal.
```rust
let a = if x > y {
    return -1;
} else {
    y
};
```
```rust
let a: !;
a = 1;
```

gcc/rust/typecheck/rust-tyty-rules.h
lrh2000: But we have to add a new one for NeverType, don't we? After all we cannot use InferType directly because general InferType that cannot be inferred to a concrete type will generate an error message.

And here `unify` is just `unify`, instead of `assign`. Refer to my comment above.

gcc/rust/typecheck/rust-tyty-rules.h
lrh2000: And what's more, assigning `<integer>` to `!` is illegal and assigning `!` to `<integer>` is legal. At present, it seems that we do not distinguish between `A->unify(B)` and `B->unify(A)`.

lrh2000:
@philberty Thanks for your review.

> Nice work, I would suggest that you run the 3 xfail tests manually. They might be failing only because an error message might be updated/changed as part of this PR so you might find it is all working as you expect. I would also suggest that you add any test cases you can from the relevant issue: #317

Yes, the failed tests did fail as expected. After I perform more code testing and (probably) discover and fix some bugs, I'll fix the existing tests and add new tests. (But I have not noticed the relevant issue before and I will take a closer look at it soon.)

> Also I would like to hear any opinions you have on the code you had to work with on this change.

I've explained the NeverType in the above comments. Here I'll explain the termination check.

Since NeverType is not a real type (it contains no data and just means that it is not reachable), the termination check at GENERIC compiling is used to make sure that it won't mess up the generated GENERIC trees. It will exit as soon as it process one of the `return`, `break` and `continue` expression.

The termination check at HIR lowering is reinforced in this PR, but it is still not wonderful. It tries its best to remove all unreachable **statements**, instead of expressions. It matters because something like `a = return -1;` is always legal regardless of `a`'s concrete type. So extra termination check at GENERIC compiling is needed.

gcc/rust/typecheck/rust-tyty-rules.h
philberty: I see what you mean since the ! is not lexically in scope as a builtin type yet. Thats true.

The other problem such as your 2nd example

```
let a: !;
a = 1;
```

The way your unify rules are in this PR will allow this to work. In order to fix it all you should have to do is remove the visitors in NeverRules to simply be

```
class NeverRules {
  virtual void visit (InferType &type) override { if type.get_infer_kind() == InferTypeKind::GENERAL { resolved = type.clone() } }
  
  virtual void visit (NeverType &type) override { resolved = type.clone (); }
}
```

When you dont implement the other visitors it means they defaulty back to a type mis match error for you. At the moment your unify rules will let that be valid.

philberty: I think for the ordering problem, that is up to the caller to make sure it is done in order i don't think there is a general way to enforce that at present.

gcc/rust/typecheck/rust-tyty-rules.h
philberty: InferTypes are inference variables if they uinify against a float/integer they will get a concrete type or if they are unified against a number literal they can then default to i32/f64 respectively.

The only type other than NeverType that should care about NeverType should be general inference variables. At the moment if you have code such as:

```
let a:_ = !
```

a will remain a General Inference variable with your rules here.

If you don't implement the NeverType visitor your rules will correctly give the type mismatch errors which we need at the moment they will be hidden.

gcc/rust/typecheck/rust-tyty-rules.h
philberty: Oh wait i just reread your comment is <integer> unified vis ! never actually legal? Ignore my review comments if this is true.

philberty: Ok i just tried this: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=0d0d292b1139de6201392255007c0691

It shows that assigning integer to ! is illegal. So in this case my comments still stand all you have to do is a tweak to your rules here.

lrh2000: > When you dont implement the other visitors it means they defaulty back to a type mis match error for you. At the moment your unify rules will let that be valid.

Then won't the following code generate a type-mismatched error?
```rust
    let b = if c > 0 {
        return 1; // ! type
    } else {
        c // <integer> type
    };
```
```rust
    let b = if c > 0 {
        c // <integer> type
    } else {
        return -1; // ! type
    };
```
Here we need to unify `!` with any other types.

lrh2000: > Oh wait i just reread your comment is integer unified against never actually legal in rust? Ignore my review comments if this is true. Thanks
Here is the code I've tested with rustc.
```rust
let a: !;
a = 1; // It does NOT compile.
```
```rust
let a: u32;
a = return -1; // It DOES compile.
```

philberty: Interesting, maybe rust playgorund is not right but i was trying this:

```
#![feature(never_type)]
fn main() {
    let x: ! = {
        return 123
    };
}
```

This gives:

```
error[E0308]: mismatched types
 --> src/main.rs:5:12
  |
5 |     return 123
  |            ^^^ expected `()`, found integer
  ```

Comment (philberty) [reply gcc/rust/typecheck/rust-tyty-rules.h:1125]: Sorry for being pedantic here its looking likely your code is correct here but lets be explicit about what should work or shouldn't.

Comment (philberty) [reply gcc/rust/typecheck/rust-tyty-rules.h:1125]: Whats interesting it that the documentation says the example i ran should have worked as you said:

https://doc.rust-lang.org/std/primitive.never.html

File: gcc/rust/typecheck/rust-tyty-rules.h (hunk 2 above)
Comment (philberty) [reply gcc/rust/typecheck/rust-tyty-rules.h:489]: ignore my comment here i think if we are sure never type works as this then it looks good. Maybe my rust playground examples are not correct since the rust documentation says otherwise.

Comment (lrh2000) [reply gcc/rust/typecheck/rust-tyty-rules.h:489]: > a will remain a General Inference variable with your rules here.

Maybe surprisingly, it is correct (partially). The below code is tested with rustc.
```rust
fn main() {
    let a: _ = {
        return;
    };
    a = 1;
    a = 1.5; // expected integer, found floating-point number
}
```
But thanks, there is a problem I have not noticed before.
```rust
fn main() {
    let a: _ = {
        return;
    };
}
```
The code above compiles with rustc. But gccrs complains that `a`'s type cannot be determined.

gcc/rust/typecheck/rust-tyty-rules.h
lrh2000: Oh actually, it expected `()` simply because the `main` function's return type is `()`.

gcc/rust/typecheck/rust-tyty-rules.h
philberty: Interesting that's going to be a difficult one to fix with the code at present. I have a patch in my PR that should help with that:

https://github.com/Rust-GCC/gccrs/pull/358/commits/3087d72eae82d75b3b4e1abaf35cdeabda5546cc

I think lets fix that test case in a different PR. Ignore my review comments I think you are correct sorry for the questions.

Lets create a separate issue for the never type to track this work: https://github.com/Rust-GCC/gccrs/issues/367

As for this PR i think if you can fix the 3 xfail test cases and add a few new test cases this looks good to me to be merged.

philberty (approved):

gcc/rust/typecheck/rust-tyty-rules.h
philberty: I think there must be a case that rustc must default a to be a (). I think we could amend my patch in https://github.com/Rust-GCC/gccrs/commit/3087d72eae82d75b3b4e1abaf35cdeabda5546cc to do that for left over general inference variables. At the end of type resolution in the name scope.

lrh2000: > As for this PR i think if you can fix the 3 xfail test cases and add a few new test cases this looks good to me to be merged.

Thanks. Maybe I'll do this tomorrow. Tonight (UTC+8) let me get my GSoC proposal done first. It is really close to the deadline (sorry for the late).

lrh2000: > I think there must be a case that rustc must default a to be a (). I think we could amend my patch in 3087d72 to do that for left over general inference variables. At the end of type resolution in the name scope.

And this does NOT compile.
```rust
fn main() {
    let a; // consider giving `a` a type
}
```

lrh2000:
There was one more thing I forgot to mention yesterday. Removing unreachable statements at HIR lowering will make gccrs and rustc behaves differently. But I'm not sure whether or not it is a problem that we need to address.

For example, this code fails to compile with rustc.
```rust
fn foo() -> i32 {
    return 1;
    let tmp: f32 = 0.0;
    tmp // expected `i32`, found `f32`
}
```
But at HIR lowering, gccrs will remove the `let tmp: f32` statement. So gccrs cannot generate the same error message as rustc.

philberty (approved):
Thanks for adding more test cases here it helps with the review.

philberty:
> > Also I would like to hear any opinions you have on the code you had to work with on this change.
> 
> There was one more thing I forgot to mention yesterday. Removing unreachable statements at HIR lowering will make gccrs and rustc behaves differently. But I'm not sure whether or not it is a problem that we need to address.
> 
> For example, this code fails to compile with rustc.
> 
> ```rust
> fn foo() -> i32 {
>     return 1;
>     let tmp: f32 = 0.0;
>     tmp // expected `i32`, found `f32`
> }
> ```
> 
> But at HIR lowering, gccrs will remove the `let tmp: f32` statement. So gccrs cannot generate the same error message as rustc.

Let's make that a new issue I did this thinking it would make it easier for the dead code caes but we can change that. In a seperate PR.

gcc/rust/typecheck/rust-tyty-rules.h
```diff
@@ -1087,6 +1092,58 @@ class StrRules : public BaseRules
...
+  using Rust::TyTy::BaseRules::visit;
+
+public:
+  NeverRules (NeverType *base) : BaseRules (base), base (base) {}
+
+  virtual void visit (TupleType &type) override { resolved = type.clone (); }
+
+  virtual void visit (ADTType &type) override { resolved = type.clone (); }
+
+  virtual void visit (InferType &type) override { resolved = type.clone (); }
```
flip1995, line 1106: This case isn't specified in (stable) Rust yet. If `InferType` can't be inferred to anything, it will fall back to the `()` (unit) type when unified with the `NeverType`. This is unstable in Rust and may change to be inferred to the `!` type in the future.

I don't think you have to change anything here yet, but I would add a `FIXME` here, since this will probably cause problems down the road. That is, because if you don't unify it with anything here, it will error after type inference iff no type could be inferred otherwise.[¹] 

Since this is unstable Rust, I don't think you have to address this yet, but I would add a note. Some more explanation about this: https://hackmd.io/@FohtAO04T92wF-8_TVATYQ/SJ0vcjyWL?type=view

---

[¹]: So either you will have to deal with this here in some way (which I don't think you can do?) or mark it here somehow, that it might be the `!` type and implement a fallback after type inference (or however this is down with integer types).

gcc/testsuite/rust.test/compile/unused.rs
```diff
@@ -14,4 +14,4 @@ fn f() {
 
 fn main() {
     f();
-}
\ No newline at end of file
+}
```
flip1995, line 17: No newline at end of file.

gcc/testsuite/rust.test/xfail_compile/continue1.rs
```diff
@@ -5,6 +5,6 @@ fn main() {
     let _fib = {
         continue; // { dg-error "cannot `continue` outside of a loop" }
         // { dg-error "failed to type resolve expression" "" { target { *-*-* } } .-1 }
-        123
+        123 // { dg-warning "unreachable expression" }
```
flip1995, line 8: This warning appearing here is wrong-ish. It is true in the sense that it can't be reached, because the program doesn't compile. From a UX perspective this produces unnecessary clutter though. This shouldn't block the PR from going through.

gcc/testsuite/rust.test/xfail_compile/implicit_returns_err2.rs
```diff
@@ -1,9 +1,6 @@
-// { dg-error "expected .* got .*" "" { target { *-*-* } } 0 }
-
 fn test(x: i32) -> i32 {
-    return 1;
-    // { dg-warning "unreachable expression" "" { target *-*-* } .+1 }
-    true
+    return true; // { dg-error "expected .i32. got .bool." }
```
flip1995, line 2: I would keep the previous test and add this modification as a new test case. This tests a different thing. Before it checked that `true` is unreachable (warning, but compile-pass) AND has the wrong type (hard error, compile fail), where the first warning should not prevent detecting the second error.

With this change it removes the dependency of those errors.

Before: "This test fails even though the expression is unreachable"
Now: "This test fails because it returns the wrong type and emits a warning"

gcc/rust/typecheck/rust-tyty-rules.h
```diff
@@ -242,6 +242,11 @@ class BaseRules : public TyVisitor
 		   type.as_string ().c_str ());
   }
 
+  virtual void visit (NeverType &type) override
+  {
+    resolved = get_base ()->clone ();
+  }
```
flip1995, line 248: > And what's more, assigning `<integer>` to `!` is illegal and assigning `!` to `<integer>` is legal. At present, it seems that we do not distinguish between `A->unify(B)` and `B->unify(A)`.

Wouldn't emitting an error here solve this? 

If `A=!` and `B=?`, then this is the direction `B->unify(A)` where you have a type `B` and try to make it work with the `!` type, aka assigning B to A (?). This is only allowed if `B==A==!`, which is covered by the `NeverRules`. 

On the other hand, the `A->unify(B)` is handled in the `NeverRules`, which means that you try to make `A` work with `B`, aka assigning A to B, which is always allowed.

flip1995 (changes requested):
Note: I don't know anything about the GCC(rs) compiler internals, so take my suggestions about the code with a grain of salt.

gcc/testsuite/rust.test/compile/unused.rs
flip1995: Oh you added it, oops. nvm

gcc/rust/typecheck/rust-tyty-rules.h
lrh2000: Thanks for your review first. I agree with you and I won't deal with this (at least in this PR). Actually when implementing NeverType, I tested what you said with rustc and its behavior really confuses me. I will add a `FIXME` later.

gcc/testsuite/rust.test/xfail_compile/continue1.rs
lrh2000: Yeah, you are right. This code cannot compile with rustc.

gcc/testsuite/rust.test/xfail_compile/implicit_returns_err2.rs
lrh2000: I agree with you. But I will justify my changes. Before this PR, all unreachable statements are removed but unreachable trailing expressions are kept. So here is a type-mistached error. But it is not reasonable at all since the unreachable expression may refer removed variables:
```rust
fn foo() -> i32 {
    return 1;
    let tmp: f32 = 0.0; // removed because it is unreachable
    tmp // gccrs cannot find tmp since it is removed
}
```
So in this PR, unreachable statements AND trailing expressions are removed. Thus the type-mismatched error disappears and the test is modified accordingly.

gcc/rust/typecheck/rust-tyty-rules.h
lrh2000: I don't think it helps. My opinion is that we must distinguish `unify` and `assign` explicitly. For example, we do need unify `!` with `<integer>` as well as unifying `<integer>` with `!` type. Consider the following two code segments.
```rust
let a = if x > y {
    return -1; // ! type
} else {
    1 // <integer> type
}; // `!`->unify(`<integer>`);
```
```rust
let a = if x > y {
   1 // <integer> type
} else {
    return -1; // ! type
}; // `<integer>`->unify(`!`);
````

File: gcc/testsuite/rust.test/xfail_compile/continue1.rs (hunk 5 above)
Comment (lrh2000) [reply gcc/testsuite/rust.test/xfail_compile/continue1.rs:8]: But I won't deal with this (at least in this PR). I've explained the reason below, see https://github.com/Rust-GCC/gccrs/pull/364#discussion_r613216808. (Note removing unreachable statements/expressions happens eariler than type check.)

File: gcc/rust/typecheck/rust-tyty-rules.h (hunk 3 above)
Comment (flip1995) [reply gcc/rust/typecheck/rust-tyty-rules.h:1106]: I think the confusion comes mainly from `!` getting inferred to `()` currently. But yeah, the never type is just confusing by nature I guess.

File: gcc/testsuite/rust.test/xfail_compile/implicit_returns_err2.rs (hunk 6 above)
Comment (flip1995) [reply gcc/testsuite/rust.test/xfail_compile/implicit_returns_err2.rs:2]: I'm not saying that your changes are wrong. I'm just saying that you should keep the old test and that it should still fail the compilation. It doesn't _really_ matter what/how many errors/warnings are emitted as long as it still fails.

Generally removing regression tests isn't something you should do. This practically removes one test and replaces it with another. I would just add the new test and keep the old test (with the `dg` comments adjusted).

File: gcc/rust/typecheck/rust-tyty-rules.h (hunk 7 above)
Comment (flip1995) [reply gcc/rust/typecheck/rust-tyty-rules.h:248]: Ah yeah, I tried to think about a situation where the other direction also works. Should have thought about it longer :smile: Nvm then. This has to be addressed, but I would say this is too big for this PR. I suggest to add a `FIXME` here as well. Maybe also add a `XPASS` test that does `let a: ! = 123;` (not sure if GCC has xpass tests though).

**Comment by philberty** [comment]
In terms of distinguishing unify and assign we do have BaseType::equals Base::Type::can_eq which equate to similar methods in rustc tyctx.

I think this PR is the right move forward in terms of Never type and having a distinct change in control flow type like return.
And I have to reiterate well done on this work truly excited that you were able to get this far for your first contribution.

The issue I have as @flip1995 mentioned is the unifying of ! with <integer> etc it doesn't seem right to me. I know it is solving a problem here but it feels as though we are going to lead to bad type bugs with those rules. But we also have the valid issue of unreachable code as you mention so I see 2 separate issues coming out of this PR one for the unify rules at least more testing and comparison with rustc and the unreachable code needs to be updated accordingly.

File: gcc/rust/typecheck/rust-hir-type-check.cc [hunk 8]
```diff
@@ -123,44 +123,25 @@ TypeResolution::Resolve (HIR::Crate &crate)
...
-    }
-
-  infered = block_tyty->clone ();
+    infered
+      = TypeCheckExpr::Resolve (expr.get_final_expr ().get (), inside_loop)
+	  ->clone ();
+  else if (expr.is_tail_reachable ())
+    infered = new TyTy::TupleType (expr.get_mappings ().get_hirid ());
+  else
+    infered = new TyTy::NeverType (expr.get_mappings ().get_hirid ());
```
philberty, line 144: are we saying the end of every block is a NeverType now?

philberty (changes requested):
Can we move forward by splitting this PR up?

The first commit is about tracking the semi-colon like rustc in the AST. That is a single simpler PR to merge.

That work could be built upon to remove the unused code termination stuff.

Then that work will fix #317 

Then that leaves the NeverType open

gcc/rust/typecheck/rust-tyty-rules.h
lrh2000: Exact. I wonder why `!` becomes `()` sometimes. Thanks for the article you provided, which answers the question.

gcc/testsuite/rust.test/xfail_compile/implicit_returns_err2.rs
lrh2000: OK, I understand now. But maybe the key problem is that the original test won't fail anymore, because the trailing expression is removed.

In my opinion, the type check against unreachable statements and expressions is totally broken and unreasonable before this PR. It skips all unreachable statements and checks the unreachable expression. The original test is just an accident so we have got the right error message. (You can see the bad case in the example above.)

As far as I'm concerned, I would rather make the type check skip all unreachable statements and the trailing expressions temporarily (and then the original test does compile). How do you think about it?

gcc/rust/typecheck/rust-tyty-rules.h
lrh2000: Yeah, I totally agree with this.

gcc/rust/typecheck/rust-hir-type-check.cc
lrh2000: It says if the block's tail is unreachable (for example, there is a `return`/`continue`/`break`), its type is `!`, otherwise its type is `()`. I don't think there is a problem.

gcc/testsuite/rust.test/xfail_compile/implicit_returns_err2.rs
flip1995: > the type check against unreachable statements and expressions is totally broken and unreasonable before this PR

AKAICT it complies with the Rust reference and `rustc`. You can't skip unreachable code in the semantic analysis as long as the `unreachable` analysis only produces a warning and not a hard error.

> As far as I'm concerned, I would rather make the type check skip all unreachable statements AND the trailing expressions temporarily (and then the original test does compile). How do you think about it?

This is against the Rust reference though, so it breaks the compiler. This test _must not_ compile.

gcc/rust/typecheck/rust-hir-type-check.cc
flip1995: This is abusing the `!` type while assuming that the type analysis works in a specific way. This is too much technical debt IMO.

lrh2000:
> Can we move forward by splitting this PR up?

I'll consider this. At least I didn't realize there are so many problems about the type system when introducing the NeverType, especially when the goal of the PR was not fully integrating the NeverType. (`let a: !` or `fn foo() -> ! {}` is not supported by this PR. Does closing https://github.com/Rust-GCC/gccrs/issues/367 requires this?)

gcc/testsuite/rust.test/xfail_compile/implicit_returns_err2.rs
dkm: Indeed, that would be really awkward as some unreachable code would build fine until one day it's not unreachable anymore and causes type errors without having touched it...

lrh2000: The original implementation is also against the Rust reference. After all it skips all unreachable statements. So maybe I should 
either do not change it at all, or make it completely right?

flip1995:
> At least I didn't realize there would be so many problems about the type system when introducing the NeverType, especially when the goal of the PR was not fully integrating the NeverType. 

The `!` type is _really_ complicated. Not even `rustc` has fully figured out how to get it right. Integrating it in a known-to-be broken state complicates the type checking system even more and will most likely cause headaches down the road. I don't think we can justify integrating the never type like this without making sure we've done it right, as-in complying with `rustc`. 

This should be done in a separate PR and not in conjunction with a bug fix of an unrelated issue. If the bug fix should depend on the `!` type being implemented fully (and correctly), then this should be done right and not only in a good-enough-for-now state.

gcc/rust/typecheck/rust-hir-type-check.cc
lrh2000: For example, rustc can compile this. I don't think it is abusing the `!` type?
```rust
fn foo() -> ! {
    loop {};
    let mut a = 1;
    a = a + a;
}
```

lrh2000:
> > At least I didn't realize there would be so many problems about the type system when introducing the NeverType, especially when the goal of the PR was not fully integrating the NeverType.
> 
> The `!` type is _really_ complicated. Not even `rustc` has fully figured out how to get it right. Integrating it in a known-to-be broken state complicates the type checking system even more and will most likely cause headaches down the road. I don't think we can justify integrating the never type like this without making sure we've done it right, as-in complying with `rustc`.
> 
> This should be done in a separate PR and not in conjunction with a bug fix of an unrelated issue. If the bug fix should depend on the `!` type being implemented fully (and correctly), then this should be done right and not only in a good-enough-for-now state.

But the original purpose of the PR is simply making the following code work
```rust
fn foo(c: i32) -> i32 {
    let b = if c > 0 {
        return 1;
    } else {
        c
    };
    b
}
```

It should be a basic support and shouldn't require us to figure out the complicated `!` type (since it is even still unstable in rustc).

flip1995:
> It should be a basic support and shouldn't require us to figure out the complicated `!` type (since it is even still unstable in rustc).

Yes exactly. So this PR should not implement the `!` type at all.

gcc/rust/typecheck/rust-hir-type-check.cc
flip1995: Yes, but only because the types unify successfully in the unreachable code. If the unreachable code produces an error the compilation will fail.

gcc/testsuite/rust.test/xfail_compile/implicit_returns_err2.rs
flip1995: > Indeed, that would be really awkward as some unreachable code would build fine until one day it's not unreachable anymore and causes type errors without having touched it...

Ah yes, the python way 🐍 

> So maybe I should either do not change it at all, or make it completely right?

No, but you shouldn't break existing regression tests, aka introducing new bugs (that were fixed before hence the regression test).

lrh2000:
> > It should be a basic support and shouldn't require us to figure out the complicated `!` type (since it is even still unstable in rustc).
> 
> Yes exactly. So this PR should not implement the `!` type at all.

OK, I got it. Then there is a question, which type should `return`/`break`/`continue` expressions resolve to?

Indeed when I started to write the code, I just thought that it was necessary to add a new type that could represent the tail of block is unreachable. And therefore we can make our code cleaner (otherwise we have to check whether or not the tail of each block is reachable when checking the type of `if/else` statement, which is a little awkward, in my opinion). After a while I found out there is already such a type in Rust named "never", then I introduce such a type. Maybe the decision is wrong and I'd better think it more deeply before.

How about keeping the first three commits in this PR (thus adding a few changes, we can make the above example work properly) and making everything related to NeverType a separate PR? The termination check is still necessary. Do you think it is OK?

gcc/testsuite/rust.test/xfail_compile/implicit_returns_err2.rs
lrh2000: OK. I'll revert the change.

lrh2000:
> In terms of distinguishing unify and assign we do have BaseType::equals Base::Type::can_eq which equate to similar methods in rustc tyctx. Also i don't think we should have to distinguish assign/unify it kind of breaks the HM algo a bit really we are saying here that now ordering matters to the unify rules so if we have:
> 
> ```
> <integer> -> <i32> -> <never> = !
> <integer> -> <i32> -> <never> -> i32 = i32
> ```
> 
> We are also saying that never can_eq every other type too which doesn't seem right either.

Just now, I had a fast investigation on how rustc deals with this (so things may be inaccurate).

There is a `lub` which computes the least-upper-bound type where the order of two types (arguments) does NOT matter ([at.rs#L147](https://github.com/rust-lang/rust/blob/16bf626a31cb5b121d0bca2baa969b4f67eb0dab/compiler/rustc_infer/src/infer/at.rs#L147)) and a `sup` which makes `actual <: expected` where the two types (arguments) are actual and expected respectively so their order DOES matter ([at.rs#93](https://github.com/rust-lang/rust/blob/16bf626a31cb5b121d0bca2baa969b4f67eb0dab/compiler/rustc_infer/src/infer/at.rs#L93)). In our context, I think the first is `unify` and the second is `assign`. You can have a look at the code comment to verify what I have said.

When type-checking the assignment, we can follow the code to from [expr.rs#L874](https://github.com/rust-lang/rust/blob/16bf626a31cb5b121d0bca2baa969b4f67eb0dab/compiler/rustc_typeck/src/check/expr.rs#L874) to [coercion.rs#L864](https://github.com/rust-lang/rust/blob/16bf626a31cb5b121d0bca2baa969b4f67eb0dab/compiler/rustc_typeck/src/check/coercion.rs#L864). Here the `use_lub` flag is not set so it will call `sub` finally.

However, when type-checking the `if/else` statement, we can follow the code from [expr.rs#L777](https://github.com/rust-lang/rust/blob/16bf626a31cb5b121d0bca2baa969b4f67eb0dab/compiler/rustc_typeck/src/check/expr.rs#L777) to [coercion.rs#L1019](https://github.com/rust-lang/rust/blob/16bf626a31cb5b121d0bca2baa969b4f67eb0dab/compiler/rustc_typeck/src/check/coercion.rs#L1019). Note the `use_lub` flag is set ([coercion.rs#L1013](https://github.com/rust-lang/rust/blob/16bf626a31cb5b121d0bca2baa969b4f67eb0dab/compiler/rustc_typeck/src/check/coercion.rs#L1013)) so it will call `lub` finally.

They finally both calls a function named `unify` ([coercion.rs#L126](https://github.com/rust-lang/rust/blob/16bf626a31cb5b121d0bca2baa969b4f67eb0dab/compiler/rustc_typeck/src/check/coercion.rs#L126)), but I think this difference matters and it indeed distinguishes `unify` and `assign`.

gcc/rust/typecheck/rust-hir-type-check.cc
lrh2000: I think there is nothing to do with this PR. After all, the problem already exists before this PR.

lrh2000:
I checked how rustc deals with NeverType. It is roughly like this:

In the rustc's function `coerce`, `!` is allowed to coerce to other type but NOT vice versa ([coercion.rs#L157](https://github.com/rust-lang/rust/blob/16bf626a31cb5b121d0bca2baa969b4f67eb0dab/compiler/rustc_typeck/src/check/coercion.rs#L157)).

So there is no problems in assignment expressions. As for `if/else/match` expressions, rustc use `CoerceMany`, which will try coercion in two directions (in our context, both `A->unify(B)` and `B->unify(A)` will be tried), see [coercion.rs#L1015](https://github.com/rust-lang/rust/blob/16bf626a31cb5b121d0bca2baa969b4f67eb0dab/compiler/rustc_typeck/src/check/coercion.rs#L1015) and [coercion.rs#L1034](https://github.com/rust-lang/rust/blob/16bf626a31cb5b121d0bca2baa969b4f67eb0dab/compiler/rustc_typeck/src/check/coercion.rs#L1034).

Is rustc's way better and should we do similar things?

lrh2000:
> Can we move forward by splitting this PR up?
> 
> The first commit is about tracking the semi-colon like rustc in the AST. That is a single simpler PR to merge.
> 
> That work could be built upon to remove the unused code termination stuff.
> 
> Then that work will fix #317
> 
> Then that leaves the NeverType open

It is not trivial. Consider the following code:
```rust
fn foo(x: i32) -> i32 {
    if x >= 0 {
        return x;
    } else {
        return -x;
    }
}
```
Here the `if/else` expression should be the trailing expression. And its type must be unified with `i32`. So which type should it resolve to?

Then I think there are only two available options:
- Mark the type by whether or not it is reachable. Then we'll essentially introduce the NeverType.
- Pass the expected type from top to bottom during the type check. So if the tail is not reachable, we will get the expected type.

Now it looks like we have to make much more changes to the code if we choose the second option.

I wonder if we can adopt the NeverType but treat it as an internal type (at least for now). In the last commit, I revert support for NeverType partially. Now only fairly limited support of NeverType is introduced. How do you think about it?

@philberty @flip1995 And thanks for your review again!

gcc/testsuite/rust.test/xfail_compile/implicit_returns_err2.rs
lrh2000: This issue should be fixed completely. It is actually very easy to fix (sorry for this since I didn't think it carefully before and I thought it should be a separate issue).

philberty:
> > Can we move forward by splitting this PR up?
> > The first commit is about tracking the semi-colon like rustc in the AST. That is a single simpler PR to merge.
> > That work could be built upon to remove the unused code termination stuff.
> > Then that work will fix #317
> > Then that leaves the NeverType open
> 
> It is not trivial. Consider the following code:
> 
> ```rust
> fn foo(x: i32) -> i32 {
>     if x >= 0 {
>         return x;
>     } else {
>         return -x;
>     }
> }
> ```
> 
> Here the `if/else` expression should be the trailing expression. And its type must be unified with `i32`. So which type should it resolve to?
> 
> Then I think there are only two available options:
> 
>     * Mark the type by whether or not it is reachable. Then we'll essentially introduce the NeverType.
> 
>     * Pass the expected type from top to bottom during the type check. So if the tail is not reachable, we will get the expected type.
> 
> 
> Now it looks like we have to make much more changes to the code if we choose the second option.
> 
> I wonder if we can adopt the NeverType but treat it as an internal type (at least for now). In the last commit, I revert support for NeverType partially. Now only fairly limited support of NeverType is introduced. How do you think about it?
> 
> @philberty @flip1995 And thanks for your review again!

That test case currently works in master and there are plenty of test cases for it. The way it should work is is that the return expressions are of type <integer> and these unify with each block when its this type of if expression. This means the entire IfExpressions if of type <integer>.

The awkward part is when we have an expected failure test case such as:

```rust
fn foo(x: i32) -> i32 {
    if x >= 0 {
        x
    } else {
        -x
    }
    123
}
```

GCCRs and rustc both identify that the if expression result in an i32 and should be of type ().  The awkward part of this is Rustc allows for:

```
fn foo(x: i32) -> i32 {
    if x >= 0 {
        return x;
    } else {
        return -x;
    }
    123
}
```

GCCRS works the same as rustc but only because the final expression is marked as unreachable. This is where the internal never type comes in to fix it. As the PR has been updated you have fixed the unify rules and can_eq etc which looks good. I havent finished this 2nd review but something I was thinking though was the the notion of the fallback type similar as we have in Inference variables.

So when we encounter a ReturnExpr without any expression we create a NeverType(hir_id) but when we do get an expression we can add in a HIRID TyVar reference for that such that we end up with:

```
NeverType::NeverType(Hirid ref, HirId ty_ref, ... HirId fallback_hir_id);

BaseType* get_fallback() const { return TyVar(fallback_hir_id).get_tyty(); )

private:
fallback_hir_id
```

That means we get everything we need.

gcc/rust/typecheck/rust-hir-type-check.cc
```diff
@@ -123,44 +123,25 @@ TypeResolution::Resolve (HIR::Crate &crate)
...
 	return false;
       }
 
-    if (stmt_is_final_expr)
-      {
-	delete block_tyty;
-	block_tyty = resolved;
-      }
-    else if (!resolved->is_unit ())
-      {
```
philberty, line 147: We still need this check for unit type for the case of:

```
fn test() -> i32 {
    if (...) { 123} else { 456 }
    789
}
```

gcc/rust/backend/rust-compile-expr.h
```diff
@@ -45,7 +47,11 @@ class CompileExpr : public HIRCompileBase
     HIR::Expr *tuple_expr = expr.get_tuple_expr ().get ();
     TupleIndex index = expr.get_tuple_index ();
 
-    Bexpression *receiver_ref = CompileExpr::Compile (tuple_expr, ctx);
+    Bexpression *receiver_ref
+      = CompileExpr::Compile (tuple_expr, ctx, &terminated);
```
philberty, line 51: Please remove the unreachable code termination in GENERIC generation, GCC should be able to optimize this. It is worth while doing but it's not the right time to be adding this to the GENERIC backend yet.

gcc/rust/typecheck/rust-hir-type-check-item.h
```diff
@@ -82,7 +82,8 @@ class TypeCheckItem : public TypeCheckBase
 
     context->pop_return_type ();
 
-    expected_ret_tyty->unify (block_expr_ty);
+    if (block_expr_ty->get_kind () != TyTy::NEVER)
+      expected_ret_tyty->unify (block_expr_ty);
```
philberty, line 86: This is not right i think this is where we want to be able to call:

never_type->get_fallback()->unify(block_expr_ty)

lrh2000: But it has been checked, in https://github.com/Rust-GCC/gccrs/blob/3c251b99b80c05a7b655aa41b3285b7ce5215ede/gcc/rust/typecheck/rust-hir-type-check-expr.h#L165
Hasn't it?

lrh2000: Oh, I noticed if the expression is empty, it won't be checked. https://github.com/Rust-GCC/gccrs/blob/3c251b99b80c05a7b655aa41b3285b7ce5215ede/gcc/rust/typecheck/rust-hir-type-check-expr.h#L150
But we can still fix it in that place?

gcc/rust/typecheck/rust-hir-type-check.cc
lrh2000: Now it is checked in https://github.com/Rust-GCC/gccrs/blob/3c251b99b80c05a7b655aa41b3285b7ce5215ede/gcc/rust/typecheck/rust-hir-type-check-stmt.h#L46-L50
since the following code is correct.
```rust
fn test() {
   123;
}
```

gcc/rust/typecheck/rust-hir-type-check-item.h
philberty: Thats only if its a return expression. What about the case where there is an empty block such as:

```
fn test() -> i32 {
}
```

Here the block will resolve to NeverType in this PR and not be checked against i32 which means we lose the existing check that will sa expected i32 got ().

philberty: > Oh, I noticed if the expression is empty, it won't be checked.
> 
> https://github.com/Rust-GCC/gccrs/blob/3c251b99b80c05a7b655aa41b3285b7ce5215ede/gcc/rust/typecheck/rust-hir-type-check-expr.h#L148-L152
> 
> But we can still fix it in that place?

I don't think we can generally assume that if the expression resolves to Never it must be checked against the return type.
What about nested if-expressions where there is a missing empty block, since nested if statement need to be dependant on the context of their parent block not nessecarily the return type of a function.

gcc/rust/backend/rust-compile-expr.h
lrh2000: The reason why I add it is that I'm afraid that the following code will mess up the GENERIC tree.
```rust
fn main() {
    let a: (i32, i32, i32) = if true {
        return;
        // Here, we'll assign `()` to `(i32, i32, i32)`. 
    } else {
        (1, 2, 3)
    };
}
```
Are you sure there is no problems at all? If there are no problems, I'll remove the unreachable code termination in GENERIC generation. (Maybe I'll have a try later.)

lrh2000:
> That test case currently works in master and there are plenty of test cases for it. The way it should work is is that the return expressions are of type and these unify with each block when its this type of if expression. This means the entire IfExpressions if of type integer.

It shouldn't. Besides what you have said, it also means the following code won't work.
```rust
fn foo() -> f32 {
    let a = if true {
        return 3.0; // `<float>`
    } else {
        3 // `<integer>`
    };
    0.0
}
```
And what's more, the semicolon after the return expression means it is a statement. I don't think it should affect the block expression's type in such a way as you described. In the following code, `let a = 1` and `let b = 1` are not removed during the type check. What's the type of the `if/else` expression?
```rust
fn foo(x: i32) -> i32 {
    if x >= 0 {
        return x;
        let a = 1;
    } else {
        return -x;
        let b = -1;
    }
}
```

gcc/rust/typecheck/rust-hir-type-check-item.h
lrh2000: > Here the block will resolve to NeverType in this PR and not be checked against i32 which means we lose the existing check that will sa expected i32 got ().
It will resolve to `()`, since the tail is reachable.

lrh2000: > What about nested if-expressions where there is a missing empty block, since nested if statement need to be dependant on the context of their parent block not nessecarily the return type of a function.

If there is a missing empty block, it means the tail is reachable so we won't get a NeverType.

philberty:
> > That test case currently works in master and there are plenty of test cases for it. The way it should work is is that the return expressions are of type and these unify with each block when its this type of if expression. This means the entire IfExpressions if of type integer.
> 
> It shouldn't. Besides what you have said, it also means the following code won't work.
> 
> ```rust
> fn foo() -> f32 {
>     let a = if true {
>         return 3.0; // `<float>`
>     } else {
>         3 // `<integer>`
>     };
>     0.0
> }
> ```
> 
> And what's more, the semicolon after the return expression means it is a statement. I don't think it should affect the block expression's type in such a way as you described. In the following code, `let a = 1` and `let b = 1` are not removed during the type check. What's the type of the `if/else` expression?
> 
> ```rust
> fn foo(x: i32) -> i32 {
>     if x >= 0 {
>         return x;
>         let a = 1;
>     } else {
>         return -x;
>         let b = -1;
>     }
> }
> ```

Yes, Let a = block_expr is a different case as to what i mentioned.

This PR removes the unit type check for block-expr type resolution:

```rust
fn foo(x: i32) -> i32 {
    if x >= 0 {
        x
    } else {
        -x
    }
    123
}
```

The if the expression is resolving to integer but it should be ().

gcc/rust/backend/rust-compile-expr.h
philberty: It shouldn't cause issues you can do it in C/C++ FE and they directly create GENERIC tress. Even if it did cause issues it shouldn't be in this PR since this is an optimization outside of the scope of what this PR is about and it is something we should discuss/design first on zulip or a github iussue.

lrh2000:
> This PR removes the unit type check for block-expr type resolution:
> 
> ```rust
> fn foo(x: i32) -> i32 {
>     if x >= 0 {
>         x
>     } else {
>         -x
>     }
>     123
> }
> ```
> 
> The if the expression is resolving to integer but it should be ().

No, I've explained in https://github.com/Rust-GCC/gccrs/pull/364#discussion_r614737279.

philberty:
Can we close this PR. The current state of the commit history is not mergeable as one big squash this PR and there many distinct changes which make it too difficult to review and agree on the behaviour:

1. Changes to the Parser Behaviour for block-expr
2. Changes to HIR lowering
3. Introduces unreachable code optimizations to GENERIC backend
4. Introduces never type on return, break and continue expressions
5. Changes how we type resolve block expressions in general
6. Indirectly means we start implementing unstable rust features although this is a case we probably need

There are already alot of comments mentioning that we wish this PR to be split up several times.

gcc/rust/backend/rust-compile-expr.h
lrh2000: Yeah, it is OK now since only fairly limited part of NeverType is introduced. There is a different case when I was writing this code.

lrh2000: Yeah, it is OK now since only fairly limited part of NeverType is introduced. There is a different case when I was writing this code.

lrh2000:
> Can we close this PR. The current state of the commit history is not mergeable as one big squash. This PR introduces many distinct changes which make it too difficult to review and agree on the behaviour:
> 
> 1. Changes to the Parser Behaviour for block-expr
> 2. Changes to HIR lowering
> 3. Introduces unreachable code optimizations to GENERIC backend
> 4. Introduces never type on return, break and continue expressions
> 5. Changes how we type resolve block expressions in general
> 6. Indirectly means we start implementing unstable rust features although this is a case we probably need
> 
> There are already alot of comments mentioning that we wish this PR to be split up several times.

OK, I'll try to split this PR into several ones. Sorry for that.

Issue #363, opened 2021-04-10
Unexpected diagnostics with unused code pass (#330, #357)


tschwinge:
The _unused code pass_ (#330, #357, @thomasyonug) seems to produce (a good number of?) unexpected diagnostics for our current testsuite.  Please correct me if I've misunderstood this, but for `rust.test/compile/block_expr2.rs`, for example, it prints:

```
[...]/gcc/testsuite/rust.test/compile/block_expr2.rs:1:1: warning: function is never used: `[test]`
```

This doesn't seem correct as `test` is used in `let a = { test() };`.  Similar for other testcases; I haven't analyzed in detail.  Can we get this fixed easily, or should we otherwise temporarily disable these unexpected _function is never used_ diagnostics (possibly conditionally enabled by a default-off command-line flag)?

These diagnostics are "conflicting" with change that I'm working on: #362 "[testsuite] Scan for warning diagnostics, too".  ("Conflicting" in quotes as I might also match these with `dg-bogus`, XFAILed, but I'd like to avoid the overhead of adding all these directives.)

thomasyonug:
As I mentioned in #357, it's very beginning work. 
I will have a better look to check if it's an easily fixed one in one day. If not, maybe we should disable it by default.

PR #362, opened 2021-04-10
[testsuite] Scan for warning diagnostics, too

tschwinge:

PR #365, opened 2021-04-11
add more path to collect live symbols

thomasyonug:
Addresses #363

### DIFF
--- a/gcc/rust/analysis/rust-hir-liveness.cc
+++ b/gcc/rust/analysis/rust-hir-liveness.cc
@@ -65,31 +65,27 @@ Liveness::Analysis (HIR::Crate &crate)
 void
 Liveness::go (HIR::Crate &crate)
 {
+  CrateNum crateNum = crate.get_mappings ().get_crate_num ();
   while (!worklist.empty ())
     {
       HirId hirId = worklist.back ();
       worklist.pop_back ();
       scannedSymbols.emplace (hirId);
-      HIR::Item *item
-	= mappings->lookup_hir_item (crate.get_mappings ().get_crate_num (),
-				     hirId);
-      if (item == nullptr)
-	continue;
+      HIR::Item *item = mappings->lookup_hir_item (crateNum, hirId);
       liveSymbols.emplace (hirId);
-      item->accept_vis (*this);
+      if (item != nullptr)
+	{
+	  item->accept_vis (*this);
+	}
+      else
+	{ // the item maybe inside a trait impl
+	  HirId parent_impl_id = UNKNOWN_HIRID;
+	  HIR::InherentImplItem *implItem
+	    = mappings->lookup_hir_implitem (crateNum, hirId, &parent_impl_id);
+	  if (implItem != nullptr)
+	    implItem->accept_vis (*this);
+	}
     }
-}
-
-void
-Liveness::visit (HIR::ExprStmtWithoutBlock &stmt)
-{
-  stmt.get_expr ()->accept_vis (*this);
-}
-
-void
-Liveness::visit (HIR::CallExpr &expr)
-{
-  expr.get_fnexpr ()->accept_vis (*this);
 }
 
 void
@@ -114,7 +110,7 @@ Liveness::visit (HIR::PathInExpression &expr)
 	  rust_error_at (expr.get_locus (), "reverse lookup failure");
 	  return;
 	}
-      if (scannedSymbols.find (ref) != scannedSymbols.end ())
+      if (scannedSymbols.find (ref) == scannedSymbols.end ())
 	{
 	  worklist.push_back (ref);
 	}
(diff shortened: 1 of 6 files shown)


thomasyonug:
Hi @tschwinge , I had check out ur pr and pass almost tests in this pr. 
But some tests such as compile/unused.rs are still failed. I don't know how to modify the dg-warning msg in the test file to get past it.  

``` bash
../gccrs/gcc/testsuite/rust.test/compile/unused.rs:2:1: warning: function is never used: `bar`
    2 | fn bar() { // { dg-warning "function is never used: `bar`" "" { target *-*-* } .-1 }
      | ^
../gccrs/gcc/testsuite/rust.test/compile/unused.rs:6:1: warning: function is never used: `foo`
    6 | fn foo() { // { dg-warning "function is never used: `foo`" "" { target *-*-* } .-1 }
      | ^
```

Could you plz help me figure out it. I'm really not familiar with dg-test suites.

dkm:
Problem is the final `.-1` that is used to say "the warning will be located on the line preceding this comment". Simply use `{ dg-warning "function is never used: `foo`" }`, that should work as expected. The relative reference can be used when several errors/warning happen on the same line for example :)

gcc/rust/analysis/rust-hir-liveness.cc
```diff
@@ -114,7 +110,10 @@ Liveness::visit (HIR::PathInExpression &expr)
 	  rust_error_at (expr.get_locus (), "reverse lookup failure");
 	  return;
 	}
-      if (scannedSymbols.find (ref) != scannedSymbols.end ())
+      for (auto s : scannedSymbols)
+	{
```
philberty, line 114: is this a loop with an empty body?

gcc/rust/analysis/rust-hir-liveness.cc
```diff
@@ -65,33 +65,29 @@ Liveness::Analysis (HIR::Crate &crate)
...
+      HIR::Item *item = mappings->lookup_hir_item (crateNum, hirId);
       liveSymbols.emplace (hirId);
-      item->accept_vis (*this);
+      if (item != nullptr)
+	{
+	  item->accept_vis (*this);
+	}
+      else
+	{ // the item maybe inside a trait impl
+	  HirId parent_impl_id = UNKNOWN_HIRID;
```
philberty, line 82: This is something i hope we can fix soon. Where you look up an item but give it a visitor instead of trying to do lookup_{item,implitem... etc} so that it can be done in one go. It might be worth fixing as part of your work here. I will make an issue for it. see #365

thomasyonug:
> Problem is the final `.-1` that is used to say "the warning will be located on the line preceding this comment". Simply use `{ dg-warning "function is never used: `foo`" }`, that should work as expected. The relative reference can be used when several errors/warning happen on the same line for example :)

Thanks! This is very helpful!

gcc/rust/analysis/rust-hir-liveness.cc
thomasyonug: Okay, do you mean #366? I'll take a look.

gcc/rust/analysis/rust-hir-liveness.cc
thomasyonug: Yeah, the gcc code indent style really makes me confused about my own code. :(
I have to spend more time adapting.

tschwinge:
@thomasyonug, thanks for looking into this so quickly!

Yeah, DejaGnu takes some time to familiarize with -- don't feel bad about that.

I'm confirming that these changes here do resolve #363 "Unexpected diagnostics with unused code pass (#330, #357)", and I'm then able to easily make #362 "[testsuite] Scan for warning diagnostics, too" work.  (If still applicable, I'll then also fix up the `dg-warning` discussed here.)

That one has been mentioned already:

```
[...]/gcc/rust/analysis/rust-hir-liveness.cc: In member function 'virtual void Rust::Analysis::Liveness::visit(Rust::HIR::PathInExpression&)':
[...]/gcc/rust/analysis/rust-hir-liveness.cc:113:17: warning: unused variable 's' [-Wunused-variable]
       for (auto s : scannedSymbols)
                 ^
```

gcc/rust/analysis/rust-hir-liveness.cc
philberty: Yeah i know what you mean are you able to take advantage of the clang-format style it helps me.

gcc/rust/analysis/rust-hir-liveness.cc
philberty: yeah I don't mean to fix that as part of this PR but it was something on my mind.

philberty:
Once that empty for loop is removed I think this is good to merge but I think it's up to you @tschwinge and @thomasyonug as to what order you want to merge these for your test suite work.

thomasyonug:
> Once that empty for loop is removed I think this is good to merge but I think it's up to you @tschwinge and @thomasyonug as to what order you want to merge these for your test suite work.

It's all up to @tschwinge, just @me if any errors come up here. Some cases have not been covered yet.

thomasyonug:
In this pr, I have to remove dg-warning in unused.rs in order to avoid warning check. Maybe you can add it back in #362? @tschwinge

tschwinge:
bors merge

Issue #366, opened 2021-04-11
Mappings improvements


philberty:
https://github.com/Rust-GCC/gccrs/blob/edc4a9549145861a2e71c552f06e53c860e39d9a/gcc/rust/util/rust-hir-map.h#L117

For the compiler to be query-based we keep mappings to IDs for all AST nodes and Hir Mappings for all HIR trees. This means we can look up any node at any time.

Instead of having to do lookup_{expr, stmt....} etc to lookup a node we should have a generic:

```c++
bool lookup_hir_node (CrateNum crateNum, HirId id, HIRLookupVisitor& vis);
```

This should take a vistior and attempt to lookup item if that fails lookup impl item etc if no node was found return false to indicate an error in lookup.

The visitor should be something like:

```c++
class HIRLookupVisitor {
  void visit(HIR::Item&) = 0;
  void visit(HIR::InherentImplItem&) = 0;
  // fill the rest out with the available mappings we have    
};
```

This would be helpful for PR #365

thomasyonug:
I see. Is vis like a callback to handle the following process of found hir node?

thomasyonug:
Should we replace the original lookup_* api with new one? Or keep original one for existence codes?

philberty:
I think for this issue it might be best to keep the original lookup_* api for a while I think it would be nice to make that a separate issue/PR to get rid of those.

philberty:
> I see. Is vis like a callback to handle the following process of found hir node?

Yes I think @YizhePKU has a PR for inline visitors that might make this really cool https://github.com/Rust-GCC/gccrs/pull/318

Infact @YizhePKU do you think it is possible to have a PR to implement the inline visitors part on its own? And also do you have any opinions on this issue?

YizhePKU:
Sure! We can implement inline visitors on its own. My currently implementation doesn't compile on GCC yet. After I fix that I'll open a PR.

The proposed improvement seems great. Maybe I'll add that returning a `bool` is not strictly necessary, since a visitor usually "returns" its result by modifying a local variable. Probably both are fine.

CastilloDel:
I would be interested in working on this!

CohenArthur:
@CastilloDel I'm assigning you then. Thanks!

philberty:
I think we should close this issue as the code has changed quite a lot since 2021

PR #369, opened 2021-04-13
Implement the inline visitor

YizhePKU:
This is a clean implementation of the inline visitor. Should be useful for various cleanups such as #366.

### DIFF
--- a/gcc/rust/util/rust-inline-visitor.h
+++ b/gcc/rust/util/rust-inline-visitor.h
@@ -1,0 +1,95 @@
+// Copyright (C) 2021 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+// An improved implementation of the inline visitor.
+// Original idea from https://members.accu.org/index.php/articles/2021
+
+#ifndef RUST_INLINE_VISITOR
+#define RUST_INLINE_VISITOR
+
+#include <utility>
+#include <type_traits>
+
+namespace Rust {
+
+// Wrapper for the target Visitor we're matching against.
+// Consumes the final nullptr of the _args linked list.
+template <typename TargetVisitor> struct EmptyVisitor : TargetVisitor
+{
+  EmptyVisitor (std::nullptr_t ptr) {}
+
+  using TargetVisitor::visit;
+};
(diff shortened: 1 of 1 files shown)


YizhePKU:
About the different behavior of GCC and Clang, here's a snippet of code that compiles in Clang but not in GCC.

```c++
struct AbstractStruct {
    virtual void not_implemented() = 0;
};

template <typename T>
struct Foo {
    T foo() { return T(); }
};

int main() {
    Foo<AbstractStruct>();
}
```

philberty (approved):

philberty:
This looks good to me. Can you post the example usage of this again? So others can see what this gives us.

YizhePKU:
Inline visitor is a ergonomic way to create a visitor without writing an entire class. It it especially useful in dealing with ADTs such as AST nodes.

Here's an example:

```c++
// Suppose we're trying to create a visitor for AST nodes
class ASTVisitor {
public:
    virtual void visit(Token& token) = 0;
    virtual void visit(LiteralExpr& expr) = 0;
    // etc.
};

// It's common to define operations only for a subset of types.
// A "base" visitor type would provide default "no-op" operations
class ASTVisitorBase : public ASTVisitor {
    virtual void visit(Token& token) {}
    virtual void visit(LiteralExpr& expr) {}
};


int main() {
    // Let's define an operation
    auto vis = Rust::begin_visitor<ASTVisitorBase>()
        .on<Token>([](Token& token) {
            // Token-specific logic
        })
        .on<LiteralExpr>([](LiteralExpr& expr) {
            // LiteralExpr-specific logic
        })
        .end_visitor();

    // We can now use that visitor as usual
    auto expr = new LiteralExpr();
    expr->accept_vis(vis); // triggers LiteralExpr-specific logic

    // Visitors can't have return values(it's not specific to inline visitors)
    // You can use capture-by-reference in lambdas as a workaround.
}
```

dkm:
This looks really nice (not sure if the builder pattern is very common in C++, but it's a nice reminder of rust) for some uses of the pattern. I'm still in the process of understanding how the different parts are organized in gccrs so can't really measure how much it could replace existing code or if it's simply for future use ? 
IIUC, this could be used when there's a need for a simple visitor without having to write a class?

What are the pro/cons for this ?
I'm wondering what is the cost of this vs the regular approach? It may also be a bit harder to debug but I have no experience with this.

YizhePKU:
> IIUC, this could be used when there's a need for a simple visitor without having to write a class?

That is correct.

> What are the pro/cons for this ?
> I'm wondering what is the cost of this vs the regular approach? It may also be a bit harder to debug but I have no experience with this.

The [original paper](https://members.accu.org/index.php/articles/2021) says that with optimization turned on, inline visitors compile to the same assembly code as traditional visitors.

Advantages of inline visitor:
* Less boilerplate, can be defined inside another function without polluting the global namespace
* Can capture local variables, eliminating the need to pass around parameters and return values

Disadvantages of inline visitor:
* Gets clumsy when the definition of the visitor is too long
* Might produce weird error messages. I haven't tried it in a real setting yet, so I'm not sure.

philberty:
I think it will be useful for cases where you want a visitor but the code is not likely to be reused for anything else so I think this makes a lot of sense for some cases in mappings and I think for substitutions I am doing at the moment it might be useful.

philberty:
One downside with lambdas is in my experience gdb doesn't seem to be great at giving location info for them.

PR #372, opened 2021-04-15
Create CODE_OF_CONDUCT.md

philberty:
Adding a code of conduct lays out rules that we can all follow in the project.

Although GCC does not have one of these we can at least adopt one here for the project development.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,128 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, religion, or sexual identity
+and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+* Demonstrating empathy and kindness toward other people
+* Being respectful of differing opinions, viewpoints, and experiences
+* Giving and gracefully accepting constructive feedback
+* Accepting responsibility and apologizing to those affected by our mistakes,
+  and learning from the experience
+* Focusing on what is best not just for us as individuals, but for the
+  overall community
+
+Examples of unacceptable behavior include:
+
+* The use of sexualized language or imagery, and sexual attention or
+  advances of any kind
+* Trolling, insulting or derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or email
+  address, without their explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+  professional setting
+
+## Enforcement Responsibilities
+
+Community leaders are responsible for clarifying and enforcing our standards of
+acceptable behavior and will take appropriate and fair corrective action in
+response to any behavior that they deem inappropriate, threatening, offensive,
+or harmful.
+
+Community leaders have the right and responsibility to remove, edit, or reject
+comments, commits, code, wiki edits, issues, and other contributions that are
+not aligned to this Code of Conduct, and will communicate reasons for moderation
+decisions when appropriate.
+
+## Scope
+
+This Code of Conduct applies within all community spaces, and also applies when
+an individual is officially representing the community in public spaces.
+Examples of representing our community include using an official e-mail address,
+posting via an official social media account, or acting as an appointed
+representative at an online or offline event.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported to the community leaders responsible for enforcement at
+https://github.com/orgs/Rust-GCC/people.
+All complaints will be reviewed and investigated promptly and fairly.
+
+All community leaders are obligated to respect the privacy and security of the
+reporter of any incident.
+
+## Enforcement Guidelines
+
+Community leaders will follow these Community Impact Guidelines in determining
+the consequences for any action they deem in violation of this Code of Conduct:
+
+### 1. Correction
+
+**Community Impact**: Use of inappropriate language or other behavior deemed
+unprofessional or unwelcome in the community.
+
+**Consequence**: A private, written warning from community leaders, providing
+clarity around the nature of the violation and an explanation of why the
+behavior was inappropriate. A public apology may be requested.
+
+### 2. Warning
+
+**Community Impact**: A violation through a single incident or series
+of actions.
+
+**Consequence**: A warning with consequences for continued behavior. No
+interaction with the people involved, including unsolicited interaction with
+those enforcing the Code of Conduct, for a specified period of time. This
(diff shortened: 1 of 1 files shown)


flip1995 (approved):
With CoCs involved it is important to note, that those are mostly just common sense. The TL;DR is "Don't be an a**". This does not restrict anyone from contributing or expressing their opinion, but guards Rust-GCC from malicious actors from outside the project, since we can point them to the CoC without them using an "I didn't know" excuse.

I'm definitely in favor of this!

bjorn3:
I am in favor of having a code of conduct. This PR adapts the contributor convenant as code of conduct, but there are many other code of conduct like documents like the [rust](https://www.rust-lang.org/policies/code-of-conduct) code of conduct or the [gnu](https://www.gnu.org/philosophy/kind-communication.html) kind communication guidelines. While I do not have any objections to the contributor convenant, I think it is worthwhile to investigate alternative code of conducts. Any problems with the chosen code of conduct could be used by community leaders to wrongly reject people (has happened before. eg https://www.fast.ai/2020/10/28/code-of-conduct/) or by contributors to justify inappropriate behavior.

dkm:
there's also the work done by some GNU hackers here:
https://wiki.gnu.tools/gnu:social-contract

A new website (https://gnu.tools/ ) will be pushed online today I think, so we may want to read this before deciding, maybe GCCrs could use it.

philberty:
Sounds good @dkm there is no rush to merge this so we can wait.

flip1995:
IIUC the GNU assembly is also just adopting the CCCoC. The social contract is not a CoC, but rather the Vision of the GNU assembly project.

dkm:
> IIUC the GNU assembly is also just adopting the CCCoC. The social contract is not a CoC, but rather the Vision of the GNU assembly project.

Yes you are right, thanks for correcting me !

NalaGinrut:
Looks great to me!
For any inappropriate words, I think it's better to publicly correct first. Try to not to blame first. This will be helpful to make the community more friendly and attractive.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported to the community leaders responsible for enforcement at
+https://gcc-rust.zulipchat.com/.
+All complaints will be reviewed and investigated promptly and fairly.
+
+All community leaders are obligated to respect the privacy and security of the
+reporter of any incident.
```
tschwinge, line 67: As it requires registration, and messages by default are public, does the Zulip chat suffice as point of contact (`[INSERT CONTACT METHOD]` in the original CCCoC v2.0 text), in particular the _[obligation] to respect the privacy [...]_?  Should this maybe rather be an email address?  What are other software communities putting there?

tschwinge:
A few "random" comments.  ;-)

I personally always wish that one doesn't need (to actually make use of) these formal rules, and instead everybody just behave reasonable by default -- but, oh well, reality teaches us...  (No issues in this group here, though, as far as I can tell.)  :-)

So, I'm not objecting to adapting this, or another set of reasonable rules.

GCC (and GNU generally) don't publish a CoC, instead GNU proposes "GNU Kind Communications Guidelines", <https://www.gnu.org/philosophy/kind-communication.html>, <https://lwn.net/Articles/769167/>.

Just as one other data point: the GNU Tools Cauldron 2019, as an in-person meeting, did have a CoC: <https://gcc.gnu.org/wiki/cauldron2019#Code_of_Conduct>, _British Computer Society Open Source Special-ist Group Code of Conduct, <https://ossg.bcs.org/coc/>_.

philberty:
Thanks, @tschwinge, I found this also https://gnu.tools/en/documents/code-of-conduct/ which might be more appropriate. Github encourages projects to have a contributor guideline and CoC in the security settings iirc.

I raised this PR to make a discussion more than anything if we think we don't need this then don't have to merge it.

flip1995:
To summarize the discussion: 

- We want a CoC.
- There are multiple different CoCs out there 
- Each CoC might have its pros and cons
- Suggested CoCs:
    - [Contributor Covenant Code of Conduct](https://www.contributor-covenant.org/) (used by GNU Assembly)
    - [Rust Code of Conduct](https://www.rust-lang.org/policies/code-of-conduct)
    - [GNU Kind Communications Guidelines](https://www.gnu.org/philosophy/kind-communication.html)
    - [BCS Code of Conduct](https://ossg.bcs.org/coc/)

Should we vote on which CoC we should adopt? How should we go ahead and decide this? Next community meeting?

---

<details>
<summary>Now facts end and my opinion starts:</summary>

The CCCoC is a pretty widely adopted CoC and I don't really see a problem why we couldn't adopt it. The problems that arose from this CoC, that  @bjorn3 pointed out (https://www.fast.ai/2020/10/28/code-of-conduct/) came mostly from the people applying the CoC rather than from the CoC itself (IMO). 

The Rust CoC would also make sense in a modified form, since this is also kind of a Rust project. Personally I like the Rust CoC the most, since it does the best job at communicating what is and isn't allowed in the community.

The GNU Kind Communications Guidelines would make sense, since this should be upstreamed to GCC in the future.

The BCS CoC would make the least sense IMO, since the BCS isn't related to this project.
</details>

tschwinge:
Additional item to resolve: point of contact, <https://github.com/Rust-GCC/gccrs/pull/372#pullrequestreview-639212843>.

---

I agree that the BCS CoC makes the least sense; I just mentioned that one for completeness, as is appeared in the GNU Tools (Cauldron) context.

I have no strong preference between the other ones.

---

> The GNU Kind Communications Guidelines would make sense, since this should be upstreamed to GCC in the future.

In some way, yes, but on the other hand: the CoC is not about the source code itself (that's what is to be upstreamed) but it's rather about the community of people (which ideally at some point also should "move upstream"), so in my opinion, there's no problem if GCC upstream and GCC/Rust are governed by different CoCs.

---

> Should we vote on which CoC we should adopt? How should we go ahead and decide this? Next community meeting?

Yeah, why not.

Thus, anyone who's not able to attend the next meeting, should be sure to voice their opinions before that one.

bjorn3:
> * [Contributor Covenant Code of Conduct](https://www.contributor-covenant.org/) (used by GNU Assembly)
> 
> * [Rust Code of Conduct](https://www.rust-lang.org/policies/code-of-conduct)
> 
> * [GNU Kind Communications Guidelines](https://www.gnu.org/philosophy/kind-communication.html)

I am fine with all of them.

philberty:
As per our last online call, I have updated the contact details to point to the organisation members, and I will be merging this. We can always change if this doesn't suit our project so please feel free to discuss this further.

tschwinge:
> Although GCC does not have [a code of conduct]

As of last week, it now does: <https://inbox.sourceware.org/CADzB+2=V=dfUK3caGE6rjYPw93NWnvyAT8vFn6bf-KoPd33bKQ@mail.gmail.com> "Announcing GCC Code of Conduct".  :tada:

> we can at least adopt one here for the project development.

Which we then did.

Should we now switch to the upstream "GCC Code of Conduct", <https://gcc.gnu.org/conduct.html>?  (..., which does seem very reasonable to me.)  I'd offer to implement the necessary (few) changes in GCC/Rust.

Maybe a discussion item for the next monthly GCC/Rust call?

PR #373, opened 2021-04-15
Create CONTRIBUTING.md

philberty:
This contributing guide should answer frequently asked questions when people submit code to GCCRS.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,46 @@
+## How to contribute to GCC Rust
+
+#### **Did you find a bug?**
+
+* **Ensure the bug was not already reported** by searching on GitHub under [Issues](https://github.com/Rust-GCC/gccrs/issues).
+
+* If you're unable to find an open issue addressing the problem, [open a new one](https://github.com/Rust-GCC/gccrs/issues/new). Be sure to include a **title and clear description**, as much relevant information as possible, and a **code sample** or an **executable test case** demonstrating the expected behavior that is not occurring.
+
+#### **Did you write a patch that fixes a bug?**
+
+* Open a new GitHub pull request with the patch.
+
+* Ensure the PR description clearly describes the problem and solution. Include the relevant issue number if applicable.
+
+* Before submitting, GCCRS requires copyright assignment. Please read the [Contributing to GCC](https://gcc.gnu.org/contribute.html) guide to know more.
+
+#### **Do you intend to add a new feature or change an existing one?**
+
+* Suggest your change in the [Zulip](https://gcc-rust.zulipchat.com/) and start writing code.
+
+* Do not open an issue on GitHub until you have collected positive feedback about the change. GitHub issues are primarily intended for bug reports and fixes.
+
+#### **Do you have questions about the source code?**
+
+* Ask any question about how to use GCCRS in [Zulip](https://gcc-rust.zulipchat.com/).
+
+### **PR Policy**
+
+* The PR policy: Everything has to go through a PR
+  - An exception to this rule will be the merge commits of updating the repo against upstream GCC
+
+* Reviewers/Maintainers of the project (aka people who have bors rights) should be pinged for reviews/questions.
+
+* A PR can have one or several commits (split should have a technical/logical reason, ie. no fixup-ish commit)
+
+* Avoid PR's with merge commit unless there's a good reason
+
+* Where possible please add test cases to gcc/testsuite/rust.test for all PRs. Some issues may not be testable via dejagnu/automation such as debug dump changes.
+
+* PR's cannot be merged untill clang format and the build and tests pass.
+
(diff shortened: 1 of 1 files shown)


flip1995 (commented):
Some things that may be worth mentioning:

1. The PR policy: Everything has to go through a PR
2. Reviewers/Maintainers of the project to ping/ask for reviews/questions (aka people who have bors rights)

CONTRIBUTING.md
```diff
@@ -0,0 +1,29 @@
...
+
+* If you're unable to find an open issue addressing the problem, [open a new one](https://github.com/Rust-GCC/gccrs/issues/new). Be sure to include a **title and clear description**, as much relevant information as possible, and a **code sample** or an **executable test case** demonstrating the expected behavior that is not occurring.
+
+#### **Did you write a patch that fixes a bug?**
+
+* Open a new GitHub pull request with the patch.
+
+* Ensure the PR description clearly describes the problem and solution. Include the relevant issue number if applicable.
+
+* Before submitting, GCCRS requires copyright assignment please read the [Contributing to GCC](https://gcc.gnu.org/contribute.html) guide to know more.
```
bjorn3, line 15:
```suggestion
* Before submitting, GCCRS requires copyright assignment. Please read the [Contributing to GCC](https://gcc.gnu.org/contribute.html) guide to know more.
```

philberty:
Thanks for the suggestions they look good to me. The only caveat I would have is that we need to periodically merge with GCC which is a very big merge commit which I don't think should require it to be via PR but I am also not opposed to that either.

flip1995:
> The only caveat I would have is that we need to periodically merge with GCC which is a very big merge commit which I don't think should require it to be via PR but I am also not opposed to that either.

Yes, you should also add this as the exception for this rule.

CONTRIBUTING.md
```diff
@@ -24,6 +24,13 @@
 
 * Ask any question about how to use GCCRS in [Zulip](https://gcc-rust.zulipchat.com/).
 
+### ** PR Policy **
+
+* The PR policy: Everything has to go through a PR
+  - An exception to this rule will be the merge commits of updating the repo against upstream GCC
+* Reviewers/Maintainers of the project (aka people who have bors rights) should be pinged for reviews/questions.
```
flip1995, line 31: Catch-22: The only people who can see this list are people who have bors rights.

dkm:
Looks good !

dkm (approved):

SimplyTheOther (approved):
Looks good to me.

NalaGinrut:
For the new features, it's better to have related tests.
Bugfix should have a test if any possible.

However, some cases are not suitable for auto testing, so I think this should be a suggestion.

philberty:
Sounds good i will add a note about tests

NalaGinrut (approved):

dkm:
Is this the correct place to ask for how the PR should look ? 
- a PR can have one or several commits (split should have a technical/logical reason, ie. no fixup-ish commit)
- no merge commit unless there's a good reason

TLDR: changes should be rebased over latest master branch and nicely split.

flip1995:
I would say this is the perfect place. But to something like this, I would add an additional note, that this only applies when merging the PR, not while the PR is WIP. Otherwise people will start squash-force pushing which is a pain to review (I call this `git canpf` – `git commit --amend --no-edit && git push --force` and may or may not have an alias for it).

dkm:
Ok, I'll try to add something about this. I must confess I'm guilty of push-forcing... I'll avoid this for future review :D

CONTRIBUTING.md
dkm: I would add :
```
* Changes must be rebased (no merge commit) over latest `master` branch and be reasonably split
* During the review process, new changes should not be forced-pushed except for the last grooming step
```

flip1995: Force pushing isn't really the problem. If you rebase a PR on top of master you'll have to force push, but the commits still exist in the previous order. The problem is, if you only have one commit in a PR and every change is amended to that commit and then force pushed.

I would write the second point as 

```
* During the development/review process, changes to the PR should be pushed as new commits and only squashed together once the PR is ready to get merged.
```

NalaGinrut:
For the merge commit issue, I think we need to make it flexible:
1. In normal cases there shouldn't be any merge commit. We should cooperate with rebasing. Because in this model, the conflicts can be resolved and `push -f` that will not affect the master.
2. However, the world is not perfect.  If there's a conflict when we behind the upstream many commits, the rebase mode requires the maintainer to fix the conflicts repeatedly. In this situation, the maintainer may choose a better way.

philberty:
I've added some more to the PR policy based on comments here. Let's merge this and please feel free to raise issues/discussion around this. This is something to help and not hinder development.

Issue #296, opened 2021-03-18
Fails to fold capacity constant


YizhePKU:
It seems like constants folding doesn't work for array capacity yet.

```Rust
fn main() {
    let arr = ["Hello"; 1 + 2];
}

// test.rs:2:25: fatal error: failed to fold capacity constant
```

```Rust
fn main() {
    let arr: [i64; 1 + 2];
}

// array_fold_type.rs:2:20: error: non-constant value
```

Maybe constant folding is simply not implemented yet? If so I'll mark the test in PR #288 as skip.

philberty:
Yeah this is a good bug to fix. The reason it is like this was the abstraction of generic were using specified a size_t for array capacity and instead of fixing that i went with this hack over in:

https://github.com/Rust-GCC/gccrs/blob/fbe9e08d952187cb70e71d284ce1e4cd50a302fe/gcc/rust/typecheck/rust-hir-type-check-type.h#L28

This class should be removed.

Then ArrayType should reference the HIR::Expr for the constant and we can remove the unify checks for size for now until we get a better solution for folding constants in the Front-end specifically. Though GCC might give us this in the middle-end for free.


https://github.com/Rust-GCC/gccrs/blob/fbe9e08d952187cb70e71d284ce1e4cd50a302fe/gcc/rust/backend/rust-compile-context.h#L402

This call should take the referenced expression which could be compiled into a GCC::Bexpresion* with CompileExpr::Compile(array_capacity_tree).

It also means your going to handle the case were we infer the size from a simple initilizer like:

let a = [1,2,3]; The code in place should count this for you but it will need turned into an implicit HIR::LiteralExpr for the ArrayType

philberty:
I started to rethink this one a bit and realised we can reuse fold-const.c from the backend code. It will be a little wierd but we might as well reuse it. Not sure how well error handling will be but I will knock up a PoC to see how it might work this week hopefully.

dkm:
>  fold-const.c

wouldn't this make gccrs dependent on gcc ? Or are you thinking on wrapping this as you did with the other part of GCC ?

philberty:
Yes it would but in PR #383 I added another wrapper over them I would really appreciate @tschwinge and @dkm or @davidmalcolm to review this it feels very janky in 2 places.

1. where we want to be able to print the array capacity for nice error messages
2. how we handle copied arrays

PR #383, opened 2021-04-20
Reuse fold-const.c from GCC to enforce const array capacities

philberty:
Rust allows for constant eval for cases like array capacity, the initial
version of this code forced the programmer to only use literals which was
not correct but got the type system off the ground.

This now takes advantage of the GCC backend object to offer 3 helprs.

 1. check for constant Bexpression* are equal
 2. extract a size_t constant value from the Bexpression*
 3. to string the Bexpression constant

We can get away with the extraction of the value here because we know its going to be a usize for array capacity but some thought is needed if these methods are to be reused in other cases.

There is a new ConstFold namespace which should be extensible for const
functions later on and const generics.

Fixes: #296

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.h
+++ b/gcc/rust/backend/rust-compile-expr.h
@@ -275,7 +275,12 @@ public:
 	return;
       }
 
-    Btype *array_type = TyTyResolveCompile::compile (ctx, tyty);
+    rust_assert (tyty->get_kind () == TyTy::TypeKind::ARRAY);
+    TyTy::ArrayType *array_tyty = static_cast<TyTy::ArrayType *> (tyty);
+    capacity_expr = array_tyty->get_capacity ();
+
+    Btype *array_type = TyTyResolveCompile::compile (ctx, array_tyty);
+    rust_assert (array_type != nullptr);
 
     expr.get_internal_elements ()->accept_vis (*this);
     std::vector<unsigned long> indexes;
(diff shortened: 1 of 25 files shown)


gcc/rust/rust-gcc.cc
```diff
@@ -176,6 +175,46 @@ class Gcc_backend : public Backend
...
+      return false;
+
+    size_t *tmp = (size_t *) buf;
+    *result = *tmp;
+    return true;
+  }
+
+  std::string const_size_val_to_string (Bexpression *expr)
+  {
+    rust_assert (TREE_CONSTANT (expr->get_tree ()));
```
philberty, line 196: This is really janky @dkm @tschwinge, the capacity constant in arrays are all constant items so fold-const.c correctly folds it all down to a constant value. But i still need to be able to print the capacity for type checking purposes but this just feels very error-prone to me here.

Also the same with the above we want we need to be able to access the value in the case of Copied arrays where we need to create an array constructor of the capacity and pass the same element.

gcc/rust/backend/rust-compile-expr.h
```diff
@@ -302,7 +307,11 @@ class CompileExpr : public HIRCompileBase
     Bexpression *translated_expr
       = CompileExpr::Compile (elems.get_elem_to_copy (), ctx);
 
-    for (size_t i = 0; i < elems.get_num_elements (); ++i)
+    size_t capacity;
+    bool ok = ctx->get_backend ()->const_size_cast (capacity_expr, &capacity);
+    rust_assert (ok);
+
+    for (size_t i = 0; i < capacity; ++i)
```
philberty, line 314: @tschwinge @dkm this is the copied array piece i mentioned

Issue #382, opened 2021-04-19
'unsafe' ICE


tschwinge:
A simple:

```rust
fn main() {
    unsafe {
    }
}
```

... causes ICE:

```
Program received signal SIGSEGV, Segmentation fault.
(anonymous namespace)::(anonymous namespace)::TypeCheckExpr::Resolve (expr=0x0, inside_loop=<optimized out>)
    at [...]/gcc/rust/typecheck/rust-hir-type-check-expr.h:45
45          expr->accept_vis (resolver);
```

tschwinge:
See also <https://github.com/Rust-GCC/gccrs/pull/380#pullrequestreview-642249188>.

PR #308, opened 2021-03-26
Add ICE failing tests

CohenArthur:
As per @philberty 's advice, I've tried finding a few failing test cases for the compiler. In both of those cases, the behaviour is unexpected: `match_self.rs` produces a segfault, while `cfg_feature.rs` ends up in an infinite loop. Would you like me to also open up issues so that they can be documented?

I'd also love to start hacking at the compiler to try and figure them out 

Here is the log for both files:

<details><summary>match_self.rs</summary>
<p>

```sh
>  gccrs ../gccrs/gcc/testsuite/rust.test/compilable/match_self.rs 
Preparing to parse files.
Attempting to parse file: ../gccrs/gcc/testsuite/rust.test/compilable/match_self.rs
beginning null denotation self/self-alias/dollar/crate/super handling
current peek token when starting path pratt parse: '{'
current token (just about to return path to null denotation): '{'
just finished parsing path (going to disambiguate) - peeked token is '{'
about to start parsing match arm patterns
successfully parsed initial match arm pattern
successfully parsed match arm patterns
successfully parsed match arm
successfully parsed inherent impl
SUCCESSFULLY PARSED CRATE
ran register_plugins (with no body)
SUCCESSFULLY REGISTERED PLUGINS
started injection
finished injection
SUCCESSFULLY FINISHED INJECTION
started expansion
finished expansion
SUCCESSFULLY FINISHED EXPANSION
rust1: internal compiler error: Segmentation fault
0xe9170f crash_signal
	../../gccrs/gcc/toplev.c:330
0x98e2ea Rust::Resolver::TypeCheckExpr::Resolve(Rust::HIR::Expr*, bool)
	../../gccrs/gcc/rust/typecheck/rust-hir-type-check-expr.h:41
0x98e773 Rust::Resolver::TypeCheckStmt::visit(Rust::HIR::ExprStmtWithBlock&)
	../../gccrs/gcc/rust/typecheck/rust-hir-type-check-stmt.h:44
0x989e81 Rust::Resolver::TypeCheckStmt::Resolve(Rust::HIR::Stmt*, bool)
	../../gccrs/gcc/rust/typecheck/rust-hir-type-check-stmt.h:38
0x989e81 operator()
	../../gccrs/gcc/rust/typecheck/rust-hir-type-check.cc:93
[...]

PR #384, opened 2021-04-20
Fix dg-ice handling in rust testsuite

dkm:
Fix the pruning regex to also match ICE message and allow for dg-ice use.
Add the test from #382 as a sample example.
Should help #308

### DIFF
--- a/gcc/testsuite/lib/rust.exp
+++ b/gcc/testsuite/lib/rust.exp
@@ -181,7 +181,7 @@ proc rust_target_compile { source dest type options } {
     ## FIXME: until the compiler is made less verbose, we need to prune its output almost completely.
     # Only keep lines containing certain diagnostics so that we can check these.
     global additional_prunes
-    lappend additional_prunes "^((?!(error: |warning: )).)*$"
+    lappend additional_prunes "^((?!(error: |warning: |\[Ii\]nternal compiler error: )).)*$"
 
     return [target_compile $source $dest $type $options]
 }
--- a/gcc/testsuite/rust.test/xfail_compile/unsafe.rs
+++ b/gcc/testsuite/rust.test/xfail_compile/unsafe.rs
@@ -1,0 +1,4 @@
+fn main() { // { dg-ice "#382" }
+    unsafe {
+    }
+}


gcc/testsuite/lib/rust.exp
```diff
@@ -181,7 +181,7 @@ proc rust_target_compile { source dest type options } {
     ## FIXME: until the compiler is made less verbose, we need to prune its output almost completely.
     # Only keep lines containing certain diagnostics so that we can check these.
     global additional_prunes
-    lappend additional_prunes "^((?!(error: |warning: )).)*$"
+    lappend additional_prunes "^((?!(error: |warning:|\[Ii\]nternal compiler error* )).)*$"
```
tschwinge, line 184: That should be `"^((?!(error: |warning: |\[Ii\]nternal compiler error: )).)*$"`.  ;-)

gcc/testsuite/rust.test/xfail_compile/unsafe.rs
```diff
@@ -0,0 +1,4 @@
+fn main() { // { dg-ice "Segmentation fault" }
```
tschwinge, line 1: To cross-reference, would it be useful to use `dg-ice "#382"` (or similar)?

tschwinge (changes requested):
Great, thanks!  Just minor comments.

gcc/testsuite/lib/rust.exp
dkm: nice catch, sorry, quick copy/paste from the dg-ice handling code!

gcc/testsuite/rust.test/xfail_compile/unsafe.rs
dkm: Hmmm, maybe yes. If the string is informational, "Segmentation fault" is not really a good pick. I thought the string was matched after the 'Internal compiler error' (so that you could check you are ICEing in a specific function), but I did not really read the doc and simply (badly it seems) inferred from examples

tschwinge:
bors merge

PR #388, opened 2021-04-25
Assign outer_attrs in IfLetExpr constructor

### LINKED ISSUE #360: IfLetExpr does have `outer_attrs`.
`IfLetExpr` does have `outer_attrs`. The actual issue here is that the primary constructor does not initialise the `outer_attrs` field from the `outer_attrs` parameter.

_Originally posted by @SimplyTheOther in https://github.com/Rust-GCC/gccrs/pull/348#discussion_r606641271_

CohenArthur:
Closes #360 

This fixes IfLetExpr's behavior and removes the previous unused warning as well

### DIFF
--- a/gcc/rust/ast/rust-expr.h
+++ b/gcc/rust/ast/rust-expr.h
@@ -4374,11 +4374,11 @@ public:
 
   IfLetExpr (std::vector<std::unique_ptr<Pattern> > match_arm_patterns,
 	     std::unique_ptr<Expr> value, std::unique_ptr<BlockExpr> if_block,
-	     std::vector<Attribute> /* outer_attrs */, Location locus)
-    : match_arm_patterns (std::move (match_arm_patterns)),
+	     std::vector<Attribute> outer_attrs, Location locus)
+    : outer_attrs (std::move (outer_attrs)),
+      match_arm_patterns (std::move (match_arm_patterns)),
       value (std::move (value)), if_block (std::move (if_block)), locus (locus)
   {}
-  // outer attributes not allowed on if let exprs either
 
   // copy constructor with clone
   IfLetExpr (IfLetExpr const &other)


dkm:
Can't really say anything about the change (it looks correct from the C++ pov, but maybe the field should be removed as the comment is saying it is not allowed).
I would still suggest adding more content to the commit message ? I'm not really clear on what is the expected flow with bors as it adds a commit with the PR details.

CohenArthur:
> Can't really say anything about the change (it looks correct from the C++ pov, but maybe the field should be removed as the comment is saying it is not allowed).
> I would still suggest adding more content to the commit message ? I'm not really clear on what is the expected flow with bors as it adds a commit with the PR details.

From @philberty in #380:
> IfLetExpr does have outer_attrs. The actual issue here is that the primary constructor does not initialise the outer_attrs field from the outer_attrs parameter.

My understanding is that the `outer_attrs` were not used because it was believed at the time that they were not allowed for `IfLetExpr`s. I can definitely add more to the commit message, but I'd like to get some more clarification too to make sure I'm not spewing anything wrong

SimplyTheOther (approved):
PR looks great.

SimplyTheOther:
> > Can't really say anything about the change (it looks correct from the C++ pov, but maybe the field should be removed as the comment is saying it is not allowed).
> > I would still suggest adding more content to the commit message ? I'm not really clear on what is the expected flow with bors as it adds a commit with the PR details.
> 
> From @philberty in #380:
> 
> > IfLetExpr does have outer_attrs. The actual issue here is that the primary constructor does not initialise the outer_attrs field from the outer_attrs parameter.
> 
> My understanding is that the `outer_attrs` were not used because it was believed at the time that they were not allowed for `IfLetExpr`s. I can definitely add more to the commit message, but I'd like to get some more clarification too to make sure I'm not spewing anything wrong

This is correct. The field was once thought to not be allowed, but that was an incorrect assumption (or has changed in a more recent version of the Rust reference).

CohenArthur:
> 
> 
> > > Can't really say anything about the change (it looks correct from the C++ pov, but maybe the field should be removed as the comment is saying it is not allowed).
> > > I would still suggest adding more content to the commit message ? I'm not really clear on what is the expected flow with bors as it adds a commit with the PR details.
> > 
> > 
> > From @philberty in #380:
> > > IfLetExpr does have outer_attrs. The actual issue here is that the primary constructor does not initialise the outer_attrs field from the outer_attrs parameter.
> > 
> > 
> > My understanding is that the `outer_attrs` were not used because it was believed at the time that they were not allowed for `IfLetExpr`s. I can definitely add more to the commit message, but I'd like to get some more clarification too to make sure I'm not spewing anything wrong
> 
> This is correct. The field was once thought to not be allowed, but that was an incorrect assumption (or has changed in a more recent version of the Rust reference).

Thanks for shedding light on the situation! I'll rebase and add more explanation in the commit message

philberty (approved):

philberty:
is this ready to merge @CohenArthur ?

CohenArthur:
> is this ready to merge @CohenArthur ?

Yes, thanks! I've added a bit more background in the commit's message

PR #390, opened 2021-04-25
Clean up the compilation of block expressions

lrh2000:
The original implementation of compiling block expressions and function bodies is incorrect. Here is an example.
```rust
fn foo() -> i32 {
    1;
    2;
    0
}
```
The above Rust code will compile into the below GIMPLE code.
```c
i32 foo ()
{
  i32 D.199;

  D.199 = 1;
  return D.199;
  D.199 = 2;
  return D.199;
  D.199 = 0;
  return D.199;
}
```
It is obviously incorrect.

*(Original in https://github.com/Rust-GCC/gccrs/pull/364)*


philberty (approved):

philberty:
I have 1 small nit, when it comes to git commit messages you could have used your message here in your PR which gives context to your change. This really helps when reviewing code in git to know what this change is doing.

lrh2000:
> I have 1 small nit, when it comes to git commit messages you could have used your message here in your PR which gives context to your change. This really helps when reviewing code in git to know what this change is doing.

I thought it was unnecessary before, because `bors` will automatically do it when a PR gets merged. I'll do this in the future.

Issue #359, opened 2021-04-10
Lifetime parameter not handled correctly


dkm:
```
fn foo3<'a>(t: &'a str)-> &'a str {
    t
}
```

Raises an error : 
```
slice_str.rs:8:9: error: unresolved generic parameter
```

dkm:
I'm currently working on a fix for having the lifetime parameters better handled: correct usage of parameters, but no lifetime related check.

dkm:
```
struct Test<'a, T> {
    t:&'a T
}

fn struct_use() {

    let s = Test { t: &3i32 };
}
```

```
gccrs/gcc/testsuite/rust.test/unsupported/lifetime_param2.rs:2:13: error: failed to setup generic parameter
rust1: internal compiler error: in insert_type, at rust/typecheck/rust-tyctx.cc:79
0x10cffbc Rust::Resolver::TypeCheckContext::insert_type(Rust::Analysis::NodeMapping const&, Rust::TyTy::BaseType*)
        ../../gcc/rust/typecheck/rust-tyctx.cc:79
0xffe1cd Rust::HIR::StructStruct::accept_vis(Rust::HIR::HIRVisitor&)
        ../../gcc/rust/hir/tree/rust-hir-full-test.cc:5531
0x109af62 Rust::Resolver::TypeResolution::Resolve(Rust::HIR::Crate&)
        ../../gcc/rust/typecheck/rust-hir-type-check.cc:38
0xf1290f Rust::Session::parse_file(char const*)
        ../../gcc/rust/rust-session-manager.cc:561
0xf12500 Rust::Session::parse_files(int, char const**)
        ../../gcc/rust/rust-session-manager.cc:451
0xe7e0b4 grs_langhook_parse_file
        ../../gcc/rust/rust-lang.cc:170
Please submit a full bug report,
with preprocessed source if appropriate.
Please include the complete backtrace with any bug report.
See <https://gcc.gnu.org/bugs/> for instructions.
```

philberty:
I think this seems to be fixed: https://godbolt.org/z/Wedb3M1EE

philberty:
Lets add this as a test case to cover that ICE to close out the issue for now

dkm:
This one can probably be closed.

PR #392, opened 2021-04-26
Basic support for lifetime parameter

dkm:
refs #359
This change allow for basic lifetime usage. Only syntax is checked, nothing
more. Goal is to be able to compile known-to-be-valid code.

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-implitem.h
+++ b/gcc/rust/typecheck/rust-hir-type-check-implitem.h
@@ -59,12 +59,22 @@ public:
       {
 	for (auto &generic_param : function.get_generic_params ())
 	  {
-	    auto param_type
-	      = TypeResolveGenericParam::Resolve (generic_param.get ());
-	    context->insert_type (generic_param->get_mappings (), param_type);
+	    switch (generic_param.get ()->get_kind ())
+	      {
+	      case HIR::GenericParam::GenericKind::LIFETIME:
+		// Skipping Lifetime completely until better handling.
+		break;
+		case HIR::GenericParam::GenericKind::TYPE: {
+		  auto param_type
+		    = TypeResolveGenericParam::Resolve (generic_param.get ());
+		  context->insert_type (generic_param->get_mappings (),
+					param_type);
 
-	    substitutions.push_back (
-	      TyTy::SubstitutionParamMapping (generic_param, param_type));
+		  substitutions.push_back (
+		    TyTy::SubstitutionParamMapping (generic_param, param_type));
+		}
+		break;
+	      }
 	  }
       }
 
--- a/gcc/rust/typecheck/rust-hir-type-check-toplevel.h
+++ b/gcc/rust/typecheck/rust-hir-type-check-toplevel.h
@@ -152,12 +152,21 @@ public:
       {
 	for (auto &generic_param : function.get_generic_params ())
 	  {
-	    auto param_type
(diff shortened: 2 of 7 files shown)


dkm:
I have been trying to add very basic checks (like check that lifetime parameter are defined) but it's a bit more complex than expected. So this is a first minimal step and there are some known issues. I'll try to make this a bit better, at least reject things like : 
```
fn foo1(t: &'a str) -> &'a str { // 'a is not defined
    t
}

fn foo4<'a>(t: &'a str)-> &'b str { // 'b is not defined
    t
}
```

gcc/rust/typecheck/rust-hir-type-check-implitem.h
```diff
@@ -59,12 +59,22 @@ class TypeCheckTopLevelImplItem : public TypeCheckBase
...
-	    context->insert_type (generic_param->get_mappings (), param_type);
-
-	    substitutions.push_back (
-	      TyTy::SubstitutionParamMapping (generic_param, param_type));
+	    switch (generic_param.get ()->get_kind ())
+	      {
+	      case HIR::GenericParam::GenericKind::LIFETIME:
+		// Skipping Lifetime completely until better handling.
+		break;
+		case HIR::GenericParam::GenericKind::TYPE: {
```
dkm, line 67: Grrr, clang-format always does incorrect indent for switch case with `{ }`. I've seen it in other parts in the code. Will fix this.

gcc/rust/typecheck/rust-hir-type-check-toplevel.h
```diff
@@ -152,12 +152,21 @@ class TypeCheckTopLevel : public TypeCheckBase
...
-	    context->insert_type (generic_param->get_mappings (), param_type);
-
-	    substitutions.push_back (
-	      TyTy::SubstitutionParamMapping (generic_param, param_type));
+	    switch (generic_param.get ()->get_kind ())
+	      {
+	      case HIR::GenericParam::GenericKind::LIFETIME:
+		// Skipping Lifetime completely until better handling.
+		break;
+		case HIR::GenericParam::GenericKind::TYPE: {
```
dkm, line 160: Same here.

philberty: I really like this I have some changes locally that could benifit from this enum.

philberty:
Would it be possible to extract the HIR enum for the generic parameters? This would be useful for me.

dkm:
Of course :D Go ahead! Take/modify whatever you need from this :D

philberty:
> Of course :D Go ahead! Take/modify whatever you need from this :D

I've cherry-picked your commit over into #399 which helps because I am explicitly casting GenericParam into TypeParam and the only way to make it safe was another visitor or this enum. thanks for this you are still marked as the author etc.

dkm:
> > Of course :D Go ahead! Take/modify whatever you need from this :D
> 
> I've cherry-picked your commit over into #399 which helps because I am explicitly casting GenericParam into TypeParam and the only way to make it safe was another visitor or this enum. thanks for this you are still marked as the author etc.

Oh, thanks ! But don't try too hard to keep the authorship for this very little part, my code is still in very early stage so it may be problematic for you. Do not hesitate to simply extract the enum, I'll rebase my changes over this as you'll probably merge it en master before me :D

dkm:
Ok, so I guess I should close this as it's has been merge via #399.

PR #393, opened 2021-04-26
Add support for 'unsupported' test cases in testsuite

dkm:
Add support for 'unsupported' test cases in testsuite

Introduce an 'unsupported/' directory in the testsuite. It should contains tests
case for which the compiler currently has an incorrect behavior:
- its accepts invalid input
- it rejects valid input

Some basic guidelines are provided in unsupported.exp.

### DIFF
--- a/gcc/testsuite/rust.test/unsupported/struct_field_vis.rs
+++ b/gcc/testsuite/rust.test/unsupported/struct_field_vis.rs
@@ -1,0 +1,15 @@
+// { dg-xfail-if "pub visibility not supported #432"  *-*-* }
+
+mod foomod {
+    pub struct Foo {
+        pub f: i32,
+        pub g: u32,
+    }
+}
+
+fn test() -> foomod::Foo {
+    foomod::Foo{
+        f:1,
+        g:3,
+    }
+}
--- a/gcc/testsuite/rust.test/unsupported/lifetime_param.rs
+++ b/gcc/testsuite/rust.test/unsupported/lifetime_param.rs
@@ -1,0 +1,11 @@
+// { dg-excess-errors "warnings" }
+
+// { dg-error "lifetime not defined" "#359" { xfail *-*-* } .+1 }
+fn lifetime_undefined(t: &'a str) -> &'a str {
+    t
+}
+
+// { dg-error "lifetime not defined" "#359" { xfail *-*-* } .+1 }
+fn lifetime_undefined_bis<'a>(t: &'a str)-> &'b str {
+    t
+}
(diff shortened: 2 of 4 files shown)


dkm:
I'm not really sure about the best way to handle these. Should all test be marked as XFAIL, or should we leave them as FAIL ? I would think leaving as FAIL is the best think to do, but maybe I missing something

dkm:
Replying to my comment: FAIL is not a good idea as it prevents regular use of dg. Marking everything as xfail or should-fail seems like a better idea.

philberty:
This is a reallly cool PR

philberty (approved):
I guess some of these crash i guess? I would love to be able to have a folder of stuff that _should_ compile but currently fails for any reason. I think it should ignore warnings, so evetrually we get an XPASS and can update the test cases as we go.

I think it would be brilliant for people to be able to add test cases for things that fail that we miss etc as a PR.

gcc/testsuite/rust.test/unsupported/struct_field_vis.rs
```diff
@@ -0,0 +1,15 @@
+// { dg-xfail-if "lifetime not defined" { *-*-* } }
```
bjorn3, line 1: I don't see any lifetimes. Should this be something else?

dkm:
On May 14, 2021 1:55:15 PM UTC, bjorn3 ***@***.***> wrote:
***@***.*** commented on this pull request.
>
>
>
>> @@ -0,0 +1,15 @@
>+// { dg-xfail-if "lifetime not defined" { *-*-* } }
>
>I don't see any lifetimes. Should this be something else?
>

Yes, this testcase has been modified and doesn't use any lifetime now, you're right ! I'll rework this PR a bit !

If we still want to have DG exit with 0, we need to use PASS/XFAIL. If any test is FAIL/XPASS then it will report a failure (and prevent merge). We can change this (could skip it for merge, or maybe have some DG black magic to change this behavior).

dkm:
> I guess some of these crash i guess? I would love to be able to have a folder of stuff that _should_ compile but currently fails for any reason. I think it should ignore warnings, so evetrually we get an XPASS and can update the test cases as we go.
> 
> I think it would be brilliant for people to be able to add test cases for things that fail that we miss etc as a PR.

Yes, that would be an easy way to contribute ! We would need some guidelines for naming and how to do the tests. I'll try to propose something!

dkm:
I've cleaned a bit the few tests provided with this change.
I've also added some simple guidelines in the .exp to help adding new tests.
I'm afraid this directory can get a bit messy if/when too many unsupported tests are added (I don't doubt it can be easily populated), so it may need to have naming rules also if it's getting too crowded?

philberty:
I agree the naming of test cases will become more important. It is something I have been bad at. Let's open an issue to talk about test case name.

tschwinge:
In #465 "Re-organize GCC/Rust testsuites" I'm changing `rust.test/unsupported` -> `rust/compile/xfail`.
Instead of *xfail*, another name would be DejaGnu's *kfail* (*known* to fail), but that isn't used/supported elsewhere in GCC, so we cannot easily use that even if we wanted to.  (Also, it's not completely clear what the perceived difference between *xfail* and *kfail* is.)

Issue #385, opened 2021-04-21
Desugar the HIR struct


yizhepku:
Unit structs(structs with no fields) currently fail to compile in various ways:

```rust
struct S;

fn main() {
    let s = S{};
}

// fatal error: Failed to lower expr: [StructExprStruct (or subclass): Path: S inner attributes: none]
```

```rust
struct S;

fn main() {
    let s = S;
}

// error: unresolved node: ( S ([C: 0 Nid: 11 Hid: 20]))
```

```rust
// Technically a tuple struct
struct S();

fn main() {
    let s = S();
}

// error: unrecognised token ‘)’ in type
```

This might be related to #368.

philberty:
I think your right @YizhePKU this is causing #368 thanks for this.

philberty:
#526 fixes these issues so i don't think this is a bug anymore but it should be a refactoring exercise to desugar the HIR via the pr #405

philberty:
this is working

PR #405, opened 2021-05-04
Simplify StructExpr in trees

YizhePKU:
Todos:

- [ ] Simplify AST::StructExpr
- [ ] Simplify HIR::StructExpr
- [ ] Update the parser
- [ ] Update AST-to-HIR lowering
- [ ] Update backend code generation
- [ ] Documentation & Testing


philberty:
This looks great so far!

YizhePKU:
The motivation for this PR is issue #385.

There were 11 AST classes representing StructExpr related stuff, three of which need to be lowered to HIR. As it turns out, only two were implemented. Instead of adding even more code to patch the issue, this PR aims to reduce 11 classes to 2 and eliminate the root cause.

YizhePKU:
Sorry for the inactivity. I'll come back to this when I figure out my summer intern.

PR #427, opened 2021-05-13
Fix duplicate function generation in the backend

### LINKED ISSUE #403: Dup generic function generation
Example:

```rust
struct Foo<A>(A);

impl<T> Foo<T> {
    fn new (a:T) -> Self {
        Self(a)
    }
}

pub fn main() {
    let a;
    a = Foo::new(123);

    let b;
    b = Foo::new(123);
}
```

gives two Foo::new functions for i32 this needs similar dedup code like structs

philberty:
Fix duplicated function generation for generics
    
When emitting the backend IR we need to check if we have already compiled
one already. This is all due to the fact when we do generic substitutions
in type resolution its only upon usage of a function the function is
substituted it gains a new unique ty_ref HIR ID to make sure new
substituted versions of a type have unique ids.
    
Fixes #403

### DIFF
--- a/gcc/rust/backend/rust-compile-context.h
+++ b/gcc/rust/backend/rust-compile-context.h
@@ -157,13 +157,38 @@ public:
     return true;
   }
 
-  void insert_function_decl (HirId id, ::Bfunction *fn)
+  void insert_function_decl (HirId id, ::Bfunction *fn,
+			     const TyTy::BaseType *ref)
   {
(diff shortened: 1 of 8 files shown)


bjorn3:
Bors is currently having trouble: https://github.com/bors-ng/bors-ng/issues/1246

philberty:
bors ping

PR #433, opened 2021-05-18
Allow unused underscore identifiers

### LINKED ISSUE #361: Names prefixed with underscore shouldn't get 'unused name' diagnostic
That's at least what `rustc` does; I haven't looked up any details.

For example, as present `rustc` doesn't print any diagnostics, but with the underscore removed from `_fib` in `gcc/testsuite/rust.test/compile/loop5.rs`, it prints:

```
warning: unused variable: `fib`
 --> gcc/testsuite/rust.test/compile/loop5.rs:6:9
  |
6 |     let fib = loop {
  |         ^^^ help: if this is intentional, prefix it with an underscore: `_fib`
  |
  = note: `#[warn(unused_variables)]` on by default
```

We currently print the `unused name` diagnostic for both variants.

CohenArthur:
This fixes #361

### DIFF
--- a/gcc/rust/resolve/rust-ast-resolve-unused.h
+++ b/gcc/rust/resolve/rust-ast-resolve-unused.h
@@ -30,7 +30,13 @@ public:
   static void ScanRib (Rib *r)
   {
     r->iterate_decls ([&] (NodeId decl_node_id, Location locus) -> bool {
-      if (!r->have_references_for_node (decl_node_id))
+      CanonicalPath ident = CanonicalPath::create_empty ();
+
+      bool ok = r->lookup_canonical_path (decl_node_id, &ident);
+      rust_assert (ok);
+
+      if (!r->have_references_for_node (decl_node_id)
+	  && ident.get ().at (0) != '_')
 	{
 	  rust_warning_at (locus, 0, "unused name");
 	}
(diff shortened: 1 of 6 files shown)


philberty:
This looks good to me. Though to me this looks like either 1 commit or two commits. I think two commits makes sense:

1. Add NodeId to CanonicalPath mappings (this is actually useful for name mangling later on)
2. Check for underscore before warning.

CohenArthur:
> 
> 
> This looks good to me. Though to me this looks like either 1 commit or two commits. I think two commits makes sense:
> 
>     1. Add NodeId to CanonicalPath mappings (this is actually useful for name mangling later on)
> 
>     2. Check for underscore before warning.

Gotcha, I'll rebase right away

gcc/rust/resolve/rust-ast-resolve-unused.h
```diff
@@ -30,7 +30,14 @@ class ScanUnused
   static void ScanRib (Rib *r)
   {
     r->iterate_decls ([&] (NodeId decl_node_id, Location locus) -> bool {
-      if (!r->have_references_for_node (decl_node_id))
+      CanonicalPath ident = CanonicalPath::create_empty ();
+      if (!r->lookup_id (decl_node_id, &ident))
+	{
+	  return false;
```
philberty, line 36: Can you put an assertion that this returns ok? Like:

```
bool ok = r->lookup_id();
rust_assert(ok);
```

 If this breaks the make check-rust then leave it as is.

philberty (approved):

philberty (approved):

PR #436, opened 2021-05-21
Fixed #419: only parse inner_attributes if it starts with "#!" not only "#"

### LINKED ISSUE #419: Outer attributes fail to parse at the start of a file
These could either be tool attributes:

```rust
#[rustfmt::ignore]
pub fn square(num: i32) -> i32 {
    num * num
}
```

Or proc macros:

```rust
#[tokio::main]
async fn main() {}
```

In both cases gccrs fails with:

```
<source>:1:2: error: expected '!' or '[' for inner attribute or shebang
    1 | #[tokio::main]
      |  ^
<source>:1:2: error: unrecognised token '[' for start of item
<source>:1:2: error: failed to parse item in crate
```

thomasyonug:
Fixed #419

only parse inner_attributes if it starts with "#!" not only "#"

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -442,7 +442,9 @@ Parser<ManagedTokenSource>::parse_inner_attributes ()
 {
   std::vector<AST::Attribute> inner_attributes;
 
-  while (lexer.peek_token ()->get_id () == HASH)
+  // only try to parse it if it starts with "#!" not only "#"
+  while (lexer.peek_token ()->get_id () == HASH
+	 && lexer.peek_token (1)->get_id () == EXCLAM)
     {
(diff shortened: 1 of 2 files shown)


thomasyonug:
In order to check this little change, should I add a new test that has a top attr?

philberty:
Yes, I think even for small bug fixes like this it would be ideal to add a test case. Though I know we have not implemented any attributes but you could add one just to make sure it does not fail to parse in the compiler.

philberty (approved):

thomasyonug:
Actually, there is one file that contains the attribute feature [file](https://github.com/Rust-GCC/gccrs/blob/master/gcc/testsuite/rust.test/compile/cfg_attr.rs) but it does not trigger parse error.
I give a new test to make sure the bug was fixed.

philberty (approved):

Issue #426, opened 2021-05-11
Raw identifier not supported


dkm:
Raw identifiers : https://doc.rust-lang.org/rust-by-example/compatibility/raw_identifiers.html
are not supported by gccrs:
```
pub fn square(num: i32) -> i32 {
    r#num * num
}
```
gives:
```
<source>:3:5: error: failed to find name: nnum
    3 |     r#num * num
      |     ^
```

See https://godbolt.org/z/oYaxGxGcv (beware that it is using nightly, so this link might work when bug fixed :)

PR #443, opened 2021-05-23
Fix raw identifier parsing.

dkm:
Lexer::parse_raw_identifier added the first character twice.

Adds tests for a simple raw identifier, a keyword as raw identifier and
xfail tests for a single underscore and forbidden keyword (crate) as raw
identifier.

To help error diagnostics continue after parse_raw_identifier failed in
Lexer::build_token.

Fixes: https://github.com/Rust-GCC/gccrs/issues/426


dkm:
bors r+

PR #445, opened 2021-05-23
Fix assert in rust-buffered-queue peek

dkm:
When end + num_items_to_read > buffer.size the buffer is extended to make sure it
can hold at least the extra num_items_to_read. The assert at the end of this block
however checks against end + num_queued_items. Fix the assert. This makes check-rust
zero fail on arm64.


dkm:
bors r+

PR #446, opened 2021-05-23
Fix aarch64 config for rust

dkm:
Define aarch64_rust_target_cpu_info in aarch64-protos.h.
AARCH64_ISA_V8_1 isn't defined directly, check aarch64_isa_flags.


dkm:

PR #447, opened 2021-05-23
Mention the identifier that was unused in ScanUnused.

dkm:
The ScanUnused class would only mention some name was unused.
Add the actual identifier name that is unused to help the user.

Note that this warning might produce a similar warning as the
ScanDeadcode class produces if the indentifier is a function.


dkm:
bors r+

PR #449, opened 2021-05-24
Add remark automation

philberty:
This adds linting checks for any markdown documents added to the repo. Being able to add some docs to the source code could help with sharing knowledge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+language: c
+
+arch:
+  - amd64
+  - ppc64le
+  - s390x
+  - arm64
+  
+os: linux
+
+before_install:
+  - sudo apt-get install -y automake autoconf libtool autogen bison flex libgmp3-dev libmpfr-dev libmpc-dev build-essential gcc-multilib g++-multilib dejagnu
+
+script:
+  - mkdir -p gccrs-build;
+  - cd gccrs-build; ../configure --enable-languages=rust --disable-bootstrap --enable-multilib; make -j $(nproc); make check-rust
+  - cd gccrs-build; \
+           if grep "# of unexpected" gcc/testsuite/rust/rust.sum;\
+           then \
+              echo "some tests are not correct"; \
+              exit 1; \
+            else \
+              exit 0; \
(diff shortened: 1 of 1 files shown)


.github/workflows/remark.yml
```diff
@@ -0,0 +1,29 @@
...
+
+    - name: Setup Node.js
+      uses: actions/setup-node@v1.4.4
+
+    - name: Install remark
+      run: npm install remark-cli remark-lint remark-lint-maximum-line-length remark-preset-lint-recommended
+
+    # Run
+    - name: Check *.md files
+      run:  git ls-files  '*.md' | grep -v "gcc/config*" | grep -v "gcc/testsuite" | grep -v "gcc/common.md" | xargs -I {} ./node_modules/.bin/remark {} -u lint -f > /dev/null
```
dkm, line 29: I think you don't need to have all these grep here, ls-files already behaves like ls:
```
$ git ls-files *.md
CODE_OF_CONDUCT.md
CONTRIBUTING.md
README.md
```

dkm: I guess this could be changed to 
```
run:  git ls-files  '*.md'  | xargs -I {} ./node_modules/.bin/remark {} -u lint -f > /dev/null
```

dkm (changes requested):

.github/workflows/remark.yml
philberty: I put that exlucsion in to avoid linting the machine descriptions in GCC which are lisp. Though it does pass if you leave it like this which is interesting.

dkm: I guess `ls-files` only list the root if not given any other args.
You would need something like `git ls-files **/**/**/*.md` to get the target md files.

philberty: I've updated this now, lets see how it works. Since this automation is looking at the root may wish to extend this if we add mark-down docs within the rust dir's at some point.

philberty:
I've acidently lost the remark changes

PR #256, opened 2021-02-26
Cleanup operator

YizhePKU:
This pull request extracts and combines the operator enums in the AST and HIR namespace. The new enums are in `operator.h` and the old `enum Operator` is removed.

Fixes issue #228.

Issue #228, opened 2021-02-11
Remove operators.h


philberty:
This should break down into 2 seperate changes.

1: Where we get rid of the HIR enums for HIR::ArithmeticOrLogicalExpr::ExprType, AST::LazyBooleanExpr::ExprType and HIR::NegationExpr::NegationType and reuse the AST version of the enum as there is no conversion required so having duplicate enums is just messy here.
2. remove operators.h and adjust rust-gcc.cc to reuse the AST mappings to covert to the GCC Tree operators.

YizhePKU:
I feel like duplicated enums are just part of a bigger problem. The `ast/` directory and `hir/` directory almost mirrors each other, e.g. `ast/rust-expr.h` and `hir/tree/rust-hir-expr.h`. A lot of code is spent converting AST types to their corresponding HIR types. Is there a better way to do this?

By the way, I'm confused that there's a `HIR:WhileLoopExpr` when [the guide](https://rustc-dev-guide.rust-lang.org/lowering.html) says that `while` loops are converted to `loop` when lowering.

SimplyTheOther:
> 
> 
> The `ast/` directory and `hir/` directory almost mirrors each other, e.g. `ast/rust-expr.h` and `hir/tree/rust-hir-expr.h`. 

I imagine this will change in the future as the HIR becomes more similar to rustc's HIR and less similar to the AST, such as by lowering the different types of loops to one type, as you mentioned. Similarly, I think `if let` expressions might be lowered to `match` expressions as well, or something like that.

YizhePKU:
An alternative would be to keep `operator.h` and remove all the other enums. This has several advantages:

1. It can be shared by AST and HIR, which removes duplication without tying HIR to AST's representation
2. It's centralized and easy to extend later

Plus, it's how the Go developers chose to represent their operators. Surely they know what they're doing:)

Of course the contents of `operator.h` needs to be modified, probably by extracting values from all the other enums. Also It'll be nice to put it inside a namespace.

philberty:
Yeah it was bootstrapped from the AST over time this will be more and more like rustc HIR. The goal here is to have an immutable IR where they have their own NodeMappings for AST NodeId, HidID, LocalDefId and DefID, CrateNum etc.

[...]

PR #453, opened 2021-05-28
reusing AST::Attribute and remove HIR::Attribute

thomasyonug:
Deleting HIR::Attribute and reusing AST::Attribute. 

Some relative reusing classes below:
1. DelimTokenTree
2. AttrInputMetaItemContainer
3. AttrInputLiteral
4. MetaItemLitExpr
5. MetaItemPathLit
6. MetaItemPath
7. MetaItemSeq
8. MetaWord
9. MetaNameValueStr
10. MetaListPath
11. MetaListNameValueStr
12. Attrbute
13. AttrInput
14. MacroParser

### DIFF
--- a/gcc/rust/backend/rust-compile-base.h
+++ b/gcc/rust/backend/rust-compile-base.h
@@ -32,13 +32,7 @@ public:
   virtual ~HIRCompileBase () {}
 
   // rust-ast.h
-  // virtual void visit(AttrInput& attr_input) {}
-  // virtual void visit(TokenTree& token_tree) {}
-  // virtual void visit(MacroMatch& macro_match) {}
   virtual void visit (HIR::Token &tok) {}
-  virtual void visit (HIR::DelimTokenTree &delim_tok_tree) {}
-  virtual void visit (HIR::AttrInputMetaItemContainer &input) {}
-  // virtual void visit(MetaItem& meta_item) {}
   // virtual void visit(Stmt& stmt) {}
   // virtual void visit(Expr& expr) {}
   virtual void visit (HIR::IdentifierExpr &ident_expr) {}
@@ -63,9 +57,6 @@ public:
 
   // rust-expr.h
   virtual void visit (HIR::LiteralExpr &expr) {}
-  virtual void visit (HIR::AttrInputLiteral &attr_input) {}
-  virtual void visit (HIR::MetaItemLitExpr &meta_item) {}
-  virtual void visit (HIR::MetaItemPathLit &meta_item) {}
   virtual void visit (HIR::BorrowExpr &expr) {}
   virtual void visit (HIR::DereferenceExpr &expr) {}
   virtual void visit (HIR::ErrorPropagationExpr &expr) {}
@@ -179,12 +170,6 @@ public:
   virtual void visit (HIR::MacroMatcher &matcher) {}
   virtual void visit (HIR::MacroRulesDefinition &rules_def) {}
   virtual void visit (HIR::MacroInvocation &macro_invoc) {}
-  virtual void visit (HIR::MetaItemPath &meta_item) {}
-  virtual void visit (HIR::MetaItemSeq &meta_item) {}
-  virtual void visit (HIR::MetaWord &meta_item) {}
-  virtual void visit (HIR::MetaNameValueStr &meta_item) {}
-  virtual void visit (HIR::MetaListPaths &meta_item) {}
-  virtual void visit (HIR::MetaListNameValueStr &meta_item) {}
 
   // rust-pattern.h
   virtual void visit (HIR::LiteralPattern &pattern) {}
--- a/gcc/rust/hir/tree/rust-hir-full-decls.h
+++ b/gcc/rust/hir/tree/rust-hir-full-decls.h
@@ -24,19 +24,10 @@
 namespace Rust {
 namespace HIR {
 // rust-ast.h
-class AttrInput;
 class TokenTree;
 class MacroMatch;
 class Token;
 struct Literal;
-class DelimTokenTree;
-class PathSegment;
-class SimplePathSegment;
-class SimplePath;
-struct Attribute;
-class MetaItemInner;
-class AttrInputMetaItemContainer;
-class MetaItem;
 class Stmt;
 class Item;
 class Expr;
@@ -76,8 +67,6 @@ class QualifiedPathInType;
 class ExprWithBlock;
 class LiteralExpr;
 class AttrInputLiteral;
-class MetaItemLitExpr;
-class MetaItemPathLit;
 class OperatorExpr;
 class BorrowExpr;
 class DereferenceExpr;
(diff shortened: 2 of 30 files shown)


philberty:
Nice work getting it all wired up! It might not have time to get around to reviewing this today but I will make sure to get back to you by Monday morning.

philberty:
I've added other reviewers in the mean time might help get any feedback sooner on this change.

thomasyonug:
Thanks @philberty . Take your time plz.

YizhePKU:
Hmm, a while ago I made a similar PR(#256) about reusing operators in AST and HIR. We discussed it in issue #228. The decision then was to put operators in the global namespace because operator "does not belong more or less to the AST or the HIR, so a shared space would be best".

YizhePKU:
Also this may be a good chance to introduce a cleanup: `using AttrVec = std::vector<AST::Attribute>`.

thomasyonug:
> Hmm, a while ago I made a similar PR(#256) about reusing operators in AST and HIR. We discussed it in issue #228. The decision then was to put operators in the global namespace because operator "does not belong more or less to the AST or the HIR, so a shared space would be best".

Yeah, about the namespace problem. I mentioned that what if we move the common class like Attribute out of AST namespace. 
But Phil said that keeping Attribute within the AST namespace is still good, as it signifies it belongs to the AST from parsing the language. 
I think maybe we can discuss it again next week when he came back.

YizhePKU:
> But Phil said that keeping Attribute within the AST namespace is still good, as it signifies it belongs to the AST from parsing the language.

That makes sense. It's not a big deal anyway.

gcc/rust/hir/tree/rust-hir-full-decls.h
```diff
@@ -29,14 +29,9 @@ class TokenTree;
 class MacroMatch;
 class Token;
 struct Literal;
-class DelimTokenTree;
 class PathSegment;
 class SimplePathSegment;
 class SimplePath;
```
YizhePKU, line 34: Do you mean to remove `SimplePath` here?

gcc/rust/hir/tree/rust-hir-full-decls.h
```diff
@@ -227,8 +220,6 @@ struct MacroTranscriber;
 struct MacroRule;
 class MacroRulesDefinition;
 class MacroInvocation;
-class MetaItemPath;
-class MetaItemSeq;
 class MetaWord;
```
YizhePKU, line 223: Do you mean to remove `Meta*` stuff here?

YizhePKU (changes requested):

gcc/rust/hir/tree/rust-hir-full-decls.h
YizhePKU: Everybody else now uses `AST::SimplePath`, correct?

thomasyonug: yeah, let me try to remove it.

thomasyonug: LOL, I found this is another miss deletion. I already remove the SimplePath definition but decl.

YizhePKU:
Suggestion: Treat force-push only as a last-resort. Force-pushing removes old files, making it harder for others to follow the conversation.

thomasyonug:
> Suggestion: Treat force-push only as a last resort. Force-pushing removes old files, making it harder for others to follow the conversation.

Sorry, not notice that.

thomasyonug:
> Beware that the first comment is used by bors when creating its own commit. It may be better to keep it more descriptive and keep other remarks as comments bellow ?

Thanks. About the other remarks do you mean #Fixes and #Address？

thomasyonug:
> Beware that the first comment is used by bors when creating its own commit. It may be better to keep it more descriptive and keep other remarks as comments bellow ?

Got it.

thomasyonug:
> Beware that the first comment is used by bors when creating its own commit. It may be better to keep it more descriptive and keep other remarks as comments bellow ?

Thanks, got it.

gcc/rust/backend/rust-compile-base.h
```diff
@@ -36,8 +36,8 @@ class HIRCompileBase : public HIR::HIRVisitor
   // virtual void visit(TokenTree& token_tree) {}
   // virtual void visit(MacroMatch& macro_match) {}
   virtual void visit (HIR::Token &tok) {}
-  virtual void visit (HIR::DelimTokenTree &delim_tok_tree) {}
-  virtual void visit (HIR::AttrInputMetaItemContainer &input) {}
+  // virtual void visit (HIR::DelimTokenTree &delim_tok_tree) {}
```
dkm, line 39: What about all these commented definitions ? Can't we simply remove them?

gcc/rust/typecheck/rust-hir-type-check-base.h
```diff
@@ -163,12 +162,12 @@ class TypeCheckBase : public HIR::HIRVisitor
   virtual void visit (HIR::MacroMatcher &) override {}
   virtual void visit (HIR::MacroRulesDefinition &) override {}
   virtual void visit (HIR::MacroInvocation &) override {}
-  virtual void visit (HIR::MetaItemPath &) override {}
-  virtual void visit (HIR::MetaItemSeq &) override {}
-  virtual void visit (HIR::MetaWord &) override {}
-  virtual void visit (HIR::MetaNameValueStr &) override {}
-  virtual void visit (HIR::MetaListPaths &) override {}
-  virtual void visit (HIR::MetaListNameValueStr &) override {}
+  // virtual void visit (HIR::MetaItemPath &) override {}
```
dkm, line 165: Same question here ?

thomasyonug: I thought that's better for consistency to comment the visit like before. But yes, I can simply remove them all.

YizhePKU:
What about putting `AttrVec` in the global namespace(`Rust` namespace actually)? It's nicer to write `AttrVec` instead of `AST::AttrVec`.

gcc/rust/typecheck/rust-hir-type-check-base.h
dkm: Maybe there's a reason for these comments (@philberty ?) but I guess they are not really needed (the visitor pattern being widely used in gccrs, it's easy to find which methods can be overridden)

thomasyonug:
> What about putting `AttrVec` in the global namespace(`Rust` namespace actually)? It's nicer to write `AttrVec` instead of `AST::AttrVec`.

It's kind of weird for me that  two type symbols are in different namespace. 

But I think it's better to do that if some other type alias also in global namespace.

philberty (approved):

philberty:
I think this looks good but before we can merge you will need to squash these commits. I see this change as a single logical commit.

philberty:
I do like having this behind the AST, since attributes are driven by the AST. It was slightly different for the PR @YizhePKU referenced where we had the logical operators enum defined in Lexer, AST, HIR, GCC. So in that case it was best to just have 1.

thomasyonug:
> I think this looks good but before we can merge you will need to squash these commits. I see this change as a single logical commit.

Sure.:)

PR #455, opened 2021-05-28
Update CONTRIBUTING.md

philberty:
We have been accepting patches via mail from developers following the normal GCC processes. Let's update our guidelines to reflect this.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,13 +6,16 @@
 
 * If you're unable to find an open issue addressing the problem, [open a new one](https://github.com/Rust-GCC/gccrs/issues/new). Be sure to include a **title and clear description**, as much relevant information as possible, and a **code sample** or an **executable test case** demonstrating the expected behavior that is not occurring.
 
-#### **Did you write a patch that fixes a bug?**
+#### **Do you want to submit a patch?**
 
 * Open a new GitHub pull request with the patch.
 
 * Ensure the PR description clearly describes the problem and solution. Include the relevant issue number if applicable.
 
 * Before submitting, GCCRS requires copyright assignment. Please read the [Contributing to GCC](https://gcc.gnu.org/contribute.html) guide to know more.
+
+* Patches sent to the [`gcc-rust` mailing list](https://gcc.gnu.org/mailman/listinfo/gcc-rust) are likewise welcome.
+These will be imported into a GitHub PR to follow the normal review process, and the link to the GitHub PR sent to the submitter.
 
 #### **Do you intend to add a new feature or change an existing one?**
 


YizhePKU (approved):

CONTRIBUTING.md
```diff
@@ -29,6 +29,8 @@
 * The PR policy: Everything has to go through a PR
   - An exception to this rule will be the merge commits of updating the repo against upstream GCC
 
+* Patches sent via email are welcome, these will be subject to the same Copyright Assignment standards. Each patch will be formed into a GitHub PR to follow the normal bos process.
```
dkm, line 32: boRs.

Maybe you should clarify where the review is happening in this case: review on the ml, and the PR is created when it has been accepted so not a real review and sent directly for merge?

flip1995 (approved):

CONTRIBUTING.md
philberty: Good point. A problem with accepting patches from the mailing list is when a patch might need changes, and requires us to update the PR.

I think doing code review on GitHub is much more powerful than via mail so we could still do the review here on GitHub via the PR. So when we say yes thanks for your patches and create the PR, we should forward the link back to the submitter to say here is your code under review. If the PR needs changes we can apply the patch updates from the mailing list but please expect some delays in this process due to the manual step involved.

"Reviews for the patches will be carried out via the PR on GitHub, this will be a publicly viewable resource. The link will be forwarded back to the submitter, as a way of viewing the state of their changes and as a way to manage any review comments. If a PR needs changes, please expect some delays in the management of updating the PR, due to the manual process involved."

thomasyonug (approved):

CONTRIBUTING.md
bjorn3: Crazy idea: What about creating a bot that automatically creates PR's based on patches on the mailing list and updates them as new patches are submitted to the mail thread? It could also forward comments both directions, so eg if you mail quoting parts of the patch, a github review gets created for the relevant lines and as you reply to them on github a mail gets sent to the mailing list. (will probably need some buffering to avoid spam on the mailing list)

flip1995: git does something like this with [gitgitgadget](https://github.com/gitgitgadget/gitgitgadget). Probably overkill for this project, since git is mainly developed on the mailing list, not on GitHub, but stuff like this already exists.

CohenArthur (approved):
I agree with Marc's suggestion, but I think it's all good! As soon as the project gets a dedicated mailing list (as it's in the works on the gcc one), it's important to add it there too for people who might want to send patches through it

CONTRIBUTING.md
ojeda: `b4`'s implementation may also prove helpful to automate stuff around patches from mailing lists.

https://git.kernel.org/pub/scm/utils/b4/b4.git/about/

dkm: Looks like it's similar to https://patchwork.ozlabs.org/ ? Not exactly sure...

dkm: Oh, I see GCC is already there : https://patchwork.ozlabs.org/project/gcc/list/

CONTRIBUTING.md
```diff
@@ -29,6 +29,8 @@
...
+
```
CohenArthur, line 33:
```suggestion
You can send those patches through the gcc-rust mailing list: gcc-rust@gcc.gnu.org

```

CONTRIBUTING.md
philberty: I really like the idea of automating this, is there a way for b4 to actually have a github user to create PRs? Or did I just miss that part.

ojeda: AFAIK, `b4` is primarily meant as a command-line tool for maintainers. I mentioned it because you may be to reuse part of its implementation to write the bot that would then contact GitHub etc.

Related: in Rust for Linux we are also trying to adapt the kernel development workflow to GitHub :-) Our bot is https://github.com/ksquirrel. I do not plan on automating the mailing list side until we are in mainline though (and even then, it will depend on the amount of patches from the mailing list).

CohenArthur (approved):

tschwinge:
I streamlined the text a bit; good to go, as far as I'm concerned.

CONTRIBUTING.md
```diff
@@ -6,14 +6,17 @@
...
-#### **Did you write a patch that fixes a bug?**
+#### **Do you want to submit a patch?**
 
 * Open a new GitHub pull request with the patch.
 
 * Ensure the PR description clearly describes the problem and solution. Include the relevant issue number if applicable.
 
 * Before submitting, GCCRS requires copyright assignment. Please read the [Contributing to GCC](https://gcc.gnu.org/contribute.html) guide to know more.
 
+* Patches sent to the [gcc-rust mailing list](https://gcc.gnu.org/mailman/listinfo/gcc-rust) are likewise welcome.
```
ojeda, line 17:
```suggestion
* Patches sent to the [`gcc-rust` mailing list](https://gcc.gnu.org/mailman/listinfo/gcc-rust) are likewise welcome.
```

CONTRIBUTING.md
dkm: What is @ksquirrel used for ? Was looking at existing solution for better integration of ml patch and github, and only find ways to create mail from PR, not the other way round.

ojeda: The goal is to have @ksquirrel automate many things within the kernel development workflow and allow us to use GitHub. Ideally, it will do all this:
  - Review commit messages for common errors (check for signatures, formatting and common practices, detect common errors such as lkml links, etc.) -- this is already implemented and I keep adding checks when I discover something new that it can check. See example at https://github.com/Rust-for-Linux/linux/pull/264 (expand the hidden comments).
  - Pick up `Acked-by`, `Reviewed-by`, `Tested-by` tags from reviewers and then appending them at merge time etc.
  - Run the CI per-commit just before the merge.
  - Collect patches from the ML and put them in GitHub automatically.

PR #458, opened 2021-05-31
stdin: Allow compilation from standard input

### LINKED ISSUE #457: Cannot compile code from standard input
I tried to compile rust code from standard input:

```sh
gccrs -x rs -
```

The compiler would try to parse what I write on standard input until EOF. This is what gccrs does with the `-x c` option in order to compile C code from stdin.

Instead, this happened:

```sh
Preparing to parse files.
Attempting to parse file: -
rust1: fatal error: cannot open filename -: No such file or directory
compilation terminated.
```

CohenArthur:
Fixes #457 

I'm not really sure how to test it with `dejagnu` or if it is even possible.

This PR adds a check for the `-` filename in the `RAIIFile` class. In that case, the inner pointer should simply be `stdin`. We also have to be careful not to `fclose(stdin)` upon deallocation.

### DIFF
--- a/gcc/rust/lex/rust-lex.h
+++ b/gcc/rust/lex/rust-lex.h
@@ -16,7 +16,14 @@ private:
   FILE *file;
 
 public:
-  RAIIFile (const char *filename) : file (fopen (filename, "r")) {}
+  RAIIFile (const char *filename)
+  {
+    if (strncmp (filename, "-", 1) == 0)
+      file = stdin;
+    else
+      file = fopen (filename, "r");
+  }
+
   RAIIFile (const RAIIFile &other) = delete;
   RAIIFile &operator= (const RAIIFile &other) = delete;
 
@@ -32,7 +39,7 @@ public:
 
   ~RAIIFile ()
   {
-    if (file != nullptr)
+    if (file != nullptr && file != stdin)
       fclose (file);
   }
 


gcc/rust/lex/rust-lex.h
```diff
@@ -16,7 +16,18 @@ struct RAIIFile
   FILE *file;
 
 public:
-  RAIIFile (const char *filename) : file (fopen (filename, "r")) {}
+  RAIIFile (const char *filename)
+  {
+    if (strncmp (filename, "-", 1) == 0)
```
dkm, line 21: I don't think the GNU coding style forces you to have all these braces for single statements if/else. You could have this a bit more compact, but that's cosmetic :)

CohenArthur: Oh right, I'm getting too used to writing Rust nowadays! I'll remove them, good catch

dkm (approved):

philberty (approved):

dkm:
Looks correct !

gcc/rust/lex/rust-lex.h
```diff
@@ -16,7 +16,14 @@ struct RAIIFile
...
```
tschwinge, line 21: It doesn't matter in practice (`error: unrecognized command-line option '-.rs'`, etc.), but why `strncmp` with a length of one, which matches every `filename` beginning with `-`?  Shouldn't this instead use `strcmp (filename, "-")` to exactly match `-`?

philberty: good point I missed that in the review.

CohenArthur: Ah, fair. I didn't think about files starting with `-`. I'll fix it right away.

PR #460, opened 2021-06-01
raiifile: Allow dash-prefixed files to be parsed

CohenArthur:
Thanks to @tschwinge for finding the bug.

I added a test case, but not a lot of programs know how to handle this file. Unless there is some specific option, `gcc` complains about no input files, and `cat` does not understand it either. So I'm not sure the test case is actually useful.

### DIFF
--- a/gcc/rust/lex/rust-lex.h
+++ b/gcc/rust/lex/rust-lex.h
@@ -18,7 +18,7 @@ private:
 public:
   RAIIFile (const char *filename)
   {
-    if (strncmp (filename, "-", 1) == 0)
+    if (strcmp (filename, "-") == 0)
       file = stdin;
     else
       file = fopen (filename, "r");


tschwinge:
@CohenArthur, yes, as I said in <https://github.com/Rust-GCC/gccrs/pull/458#pullrequestreview-672808738>: _It doesn't matter in practice (`error: unrecognized command-line option '-.rs'`, etc.)_.  But thanks for addressing this anyway.  :-)

Yes, the testcase doesn't achieve what it means to test: the DejaGnu test driver will pass a full path, so that won't begin with `-`, so I suggest you force-push with the testcase removed, and then we'll merge.  (I don't know how to easily test such a thing within the DejaGnu framework, and it doesn't seem important enough to spend more time on this.)

CohenArthur:
> Yes, the testcase doesn't achieve what it means to test: the DejaGnu test driver will pass a full path, so that won't begin with `-`, so I suggest you force-push with the testcase removed, and then we'll merge. (I don't know how to easily test such a thing within the DejaGnu framework, and it doesn't seem important enough to spend more time on this.)

That's what I was scared of. I thought the framework might cd into the directory and run the cases there, but didn't check if that was the case. I thought I'd ask the GCC experts and that it'd be faster than learning perl!

tschwinge:
bors: merge

---

> I thought the framework might [...], but didn't check [...]

That's easy enough to do: see `[build-gcc]/gcc/testsuite/rust/rust.log` for a log of the commands that it's executing.

> learning perl

Fortunately, no Perl involved here -- DejaGnu is TCL/Expect, <https://www.gnu.org/software/dejagnu/manual/What-is-DejaGnu_003f.html>.  ;-P

dkm:
> > learning perl
> 
> Fortunately, no Perl involved here -- DejaGnu is TCL/Expect, https://www.gnu.org/software/dejagnu/manual/What-is-DejaGnu_003f.html. ;-P

not sure what's best :)

PR #463, opened 2021-06-01
Always print compiler debug info on stderr

CohenArthur:
I am starting my work on cargo-gccrs, and the output from the compiler messes with what cargo expects as output. Since much of the debug prints are already done on `stderr` instead of `stdout`, this PR fixes what I believe to be the few remaining cases using `stdout`.

I used the following regexes (with ripgrep) to find all remaining instances of debug prints on `stdout`: `[^f]printf` and `fprintf \(stdout`

### DIFF
--- a/gcc/rust/rust-session-manager.cc
+++ b/gcc/rust/rust-session-manager.cc
@@ -456,7 +456,7 @@ Session::parse_files (int num_files, const char **files)
 
   for (int i = 0; i < num_files; i++)
     {
-      printf ("Attempting to parse file: %s\n", files[i]);
+      fprintf (stderr, "Attempting to parse file: %s\n", files[i]);
       parse_file (files[i]);
     }
   /* TODO: should semantic analysis be dealed with here? or per file? for now,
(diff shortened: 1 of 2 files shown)


philberty (approved):
I think this looks good but something that would be really useful for me if someone could fix this issue:  https://github.com/Rust-GCC/gccrs/issues/464

CohenArthur:
> I think this looks good but something that would be really useful for me if someone could fix this issue: #464

I could do this and replace all calls to `fprintf (stderr, ...)` with calls to `rust_debug (...)`?

philberty:
> > I think this looks good but something that would be really useful for me if someone could fix this issue: #464
> 
> I could do this and replace all calls to `fprintf (stderr, ...)` with calls to `rust_debug (...)`?

I think it makes sense in a separate PR. Just check that debug doesn't end up using stdout again lol.

dkm:
And using `rust_debug` would allow for a prefix and easier filtering

PR #465, opened 2021-06-02
Re-organize GCC/Rust testsuites

tschwinge:
... as discussed in
<https://gcc-rust.zulipchat.com/#narrow/stream/266897-general/topic/Testsuite.20results>:

> [...] best practice from what other GCC testsuites are doing), which I hope
> makes clearer the difference between the individual GCC/Rust testsuites:
> 'rust.test/compile' -> 'rust/compile/torture'
> 'rust.test/execute' -> 'rust/execute/torture'
> 'rust.test/unsupported' -> 'rust/compile/xfail'
> 'rust.test/xfail_compile' -> 'rust/compile'

... plus corresponding documentation updates in the '*.exp' files and
elsewhere.


philberty (approved):
I think I understand the naming of the other test suites better now. Thanks for doing this, I think it should be up to your or @dkm to merge this since your both more experienced than I am with this.

dkm:
It's really better like this :+1: 
bors r+

PR #466, opened 2021-06-02
Expose rust debug and use it

### LINKED ISSUE #464: Add prototype for rust_debug
Debuging is useful, we don't have the logger yet from #248 maybe I can poke GCC people about my patch. 

We do have a rust_debug wrapper, which calls inform but the prototype is not available, this needs to be moved into the header file. https://github.com/Rust-GCC/gccrs/blob/2a644a5ebb5e4b8f8dd52e5da1c87b73f2ae7f1b/gcc/rust/rust-diagnostics.cc#L203

It would also be nice if there was an overloaded version without Location and defaults Location to simply Location().

CohenArthur:
Fixes #464 

@philberty I've checked and rust_debug does output on stderr correctly :)

This PR exposes the rust_debug prototype and replaces debug printfs (`fprintf (stderr, ...)`) with calls to rust_debug. This also removes the extra newlines that were added at the end of the fprintf format strings, since calling rust_debug already adds one. Without this, the tests fail due to blank lines in the output.

I have used a template to overload the rust_debug function and pass variadic arguments to the complete version. While the code looks cleaner than using `va_args`, it's also heavier C++. I'm open to changing it and definitely not attached to this implementation.

Most of the PR was made using sed and then clang-format

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -261,10 +261,9 @@ Parser<ManagedTokenSource>::left_binding_power (const_TokenPtr token)
 	  return LBP_DOT;*/
 
     case SCOPE_RESOLUTION:
-      fprintf (
-	stderr,
+      rust_debug (
 	"possible error - looked up LBP of scope resolution operator. should "
-	"be handled elsewhere. \n");
+	"be handled elsewhere.");
       return LBP_PATH;
 
     /* Resolved by lookahead HACK that should work with current code. If next
@@ -609,10 +608,10 @@ Parser<ManagedTokenSource>::parse_simple_path ()
     {
       if (seg.is_error ())
 	{
-	  fprintf (stderr,
-		   "when parsing simple path, somehow empty path segment was "
-		   "not filtered out. Path begins with '%s' \n",
-		   segments.at (0).as_string ().c_str ());
+	  rust_debug (
+	    "when parsing simple path, somehow empty path segment was "
+	    "not filtered out. Path begins with '%s'",
+	    segments.at (0).as_string ().c_str ());
 	}
     }
 
@@ -917,11 +916,10 @@ Parser<ManagedTokenSource>::parse_delim_token_tree ()
       lexer.skip_token ();
 
       // DEBUG
-      fprintf (stderr,
-	       "finished parsing new delim token tree - peeked token is now "
-	       "'%s' while t is '%s'\n",
-	       lexer.peek_token ()->get_token_description (),
-	       t->get_token_description ());
+      rust_debug ("finished parsing new delim token tree - peeked token is now "
+		  "'%s' while t is '%s'",
+		  lexer.peek_token ()->get_token_description (),
+		  t->get_token_description ());
 
       return token_tree;
     }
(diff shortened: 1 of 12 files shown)


gcc/rust/rust-diagnostics.h
```diff
@@ -124,4 +124,13 @@ struct Error
...
 } // namespace Rust
 
+// rust_debug uses normal printf formatting, not GCC diagnostic formatting.
+void
+rust_debug (const Location location, const char *fmt, ...) ATTRIBUTE_PRINTF_2;
+
+template <typename ... Args> void
+rust_debug (const char *fmt, Args... args) {
+    rust_debug (Location (), fmt, args...);
+};
```
tschwinge, line 136: `[...]/gcc/rust/rust-diagnostics.h:134:2: warning: extra ';' [-Wpedantic]`.

gcc/rust/rust-diagnostics.cc
```diff
@@ -197,11 +197,6 @@ rust_error_at (const RichLocation location, const char *fmt, ...)
   va_end (ap);
 }
 
-// rust_debug uses normal printf formatting, not GCC diagnostic formatting.
-
-void
-rust_debug (const Location location, const char *fmt, ...) ATTRIBUTE_PRINTF_2;
-
```
tschwinge, line 204: Unless there was a reason to move it, I suggest you leave this prototype in its original place; it applies to the non-templated `rust_debug` function.

CohenArthur: The reason was in #464, I assumed this prototype needed to be available too in case there was a need for rust_debug with a given location instead of an empty one

tschwinge: Right.  I failed to see that it moved from the `*.cc` into the `*.h` file.  ;-)

gcc/rust/rust-diagnostics.h
```diff
@@ -124,4 +124,13 @@ struct Error
 };
 } // namespace Rust
 
+// rust_debug uses normal printf formatting, not GCC diagnostic formatting.
+void
+rust_debug (const Location location, const char *fmt, ...) ATTRIBUTE_PRINTF_2;
+
+template <typename ... Args> void
+rust_debug (const char *fmt, Args... args) {
```
tschwinge: Shouldn't this be marked `inline`?

tschwinge:
Yes, shouldn't simply *all* `fprintf` calls disappear from `gcc/rust/`?

CohenArthur:
> Yes, shouldn't simply _all_ `fprintf` calls disappear from `gcc/rust/`?

I think so. I believe that's where I sit right now with my uncommited fixes. I'll finish the PR later today

CohenArthur:
Sorry about the multiple CI fails. I was using clang-format 11 which did not behave exactly the same as clang-format 10

gcc/rust/rust-diagnostics.h
CohenArthur: Applied in 701b533

gcc/rust/rust-diagnostics.h
CohenArthur: Applied in a6054b0

tschwinge (changes requested):
In a few places, there's an empty line (where `stderr` use to be?):

    -             fprintf (
    -               stderr,
    +             rust_debug (
    +
                    "[...]

CohenArthur:
> In a few places, there's an empty line (where `stderr` use to be?):
> 
> ```
> -             fprintf (
> -               stderr,
> +             rust_debug (
> +
>                 "[...]
> ```

Yup, you caught it right before I could push it! Thanks! This is fixed in 5ef7783

tschwinge (changes requested):
In a few places, there's still a `\n` at the end of the `rust_debug` `fmt` string.

tschwinge:
We're getting there!  Verified that you got all `fprintf`s.  Just a few minor review items remaining.  :+1:

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -14821,7 +14796,7 @@ Parser<ManagedTokenSource>::debug_dump_lex_output (std::ostream &out)
     }
 }
 
-// Parses crate and dumps AST to stderr, recursively.
+// Parses crate and dumps AST to  recursively.
```
tschwinge, line 14799: Please fix (restore `stderr`, I suppose?) that one, and we're good to go!

CohenArthur: Ah dangit my regex-fu was too powerful. Thanks for catching that

tschwinge (approved):

tschwinge:
bors: merge

PR #479, opened 2021-06-03
Remove diagnostics pruning from 'rust_target_compile' [#301, #466]

tschwinge:
This is no longer necessary after #466 "Expose rust debug and use it", where
commit 665df329a2ad60580ab593f6cbd646aa55927a37 "rust_debug: Replace fprintf
(stderr) with rust_debug" turned all 'fprintf' to 'stderr' into standard GCC
'inform' ('note: [...]') diagnostics -- which do get pruned by default.

Adjust the testsuite: instead of 'dg-excess-errors', spell out the additional
diagnostics now seen.

### DIFF
--- a/gcc/testsuite/lib/rust.exp
+++ b/gcc/testsuite/lib/rust.exp
@@ -178,10 +178,5 @@ proc rust_target_compile { source dest type options } {
     set options [concat "$ALWAYS_RUSTFLAGS" $options]
     set options [dg-additional-files-options $options $source]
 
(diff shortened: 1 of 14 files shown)


tschwinge:
I also have a follow-up patch to hide `rust_debug` output behind a (default-off) command-line flag.

dkm (approved):

dkm:
this will make writing test a bit less confusing!

bors r+

PR #480, opened 2021-06-04
Use '-frust-debug' to control dumping of various Rust front end internals

tschwinge:
... as discussed in #479, and elsewhere.


tschwinge:
We may later want to merge `-frust-debug` and `-frust-dump-[...]`, but this seems like a good incremental step?

philberty (approved):
Brilliant I like this :)

PR #482, opened 2021-06-04
arm: Override TARGET_RUST_OS_INFO in bpabi.h

glaubitz:
Similar to PowerPC, 32-bit ARM has the issue that TARGET_RUST_OS_INFO
is defined in two separate headers, namely config/arm/bpabi.h and
config/arm/linux-elf.h. We can apply the same fix that was already
used for PowerPC.

gcc/ChangeLog:

2021-06-04  John Paul Adrian Glaubitz <glaubitz@physik.fu-berlin.de>

	* config/arm/bpabi.h: Undefine TARGET_RUST_OS_INFO before
	defining it.

### DIFF
--- a/gcc/config/arm/bpabi.h
+++ b/gcc/config/arm/bpabi.h
@@ -109,9 +109,8 @@
     /*TODO: is this even an OS? What should go here?*/ \
   } while (0)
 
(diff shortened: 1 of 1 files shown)


tschwinge:
bors: merge

PR #493, opened 2021-06-14
Path resolution fixes

philberty:
In order to support Modules, we need to do path resolution properly. The old code assumed a basic A::B but this is not generic enough in order to support modules.


bjorn3:
nit: s/untill/until in the commit message

PR #503, opened 2021-06-18
Use DW_LANG_Rust in DWARF

tromey:
This changes the GCC DWARF emitterto use DW_LANG_Rust or
DW_LANG_Rust_old as appropriate.

### DIFF
--- a/gcc/dwarf2out.c
+++ b/gcc/dwarf2out.c
@@ -24634,6 +24634,13 @@ gen_compile_unit_die (const char *filename)
     }
   else if (strcmp (language_string, "GNU F77") == 0)
     language = DW_LANG_Fortran77;
+  else if (strcmp (language_string, "GNU Rust") == 0)
+    {
+      if (dwarf_version >= 5 || !dwarf_strict)
+	language = DW_LANG_Rust;
+      else
+	language = DW_LANG_Rust_old;
+    }
   else if (dwarf_version >= 3 || !dwarf_strict)
     {
       if (strcmp (language_string, "GNU Ada") == 0)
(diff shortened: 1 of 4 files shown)


philberty (approved):

dkm:
Maybe the PR template content (`Thank you for making Rust GCC better!....`) could be removed from the first comment? It will end-up in the git log :)

tromey:
I edited the comment, no idea if that really works.

dkm:
> I edited the comment, no idea if that really works.

Yes, editing the first comment works ! Thanks !

tschwinge:
My understanding of DWARF is limited, so please bear with me.  ;-)

@tromey, I suppose that `DW_LANG_Rust`/`DW_LANG_Rust_old` have originally been created for `rustc`/LLVM proper.  Meaning that GDB (or any other consumer) will (I suppose) use this to "switch to expecting to see `rustc`/LLVM idioms" in the debug information (whatever that exactly means in practice).  Are there now any caveats that GCC/Rust has to make sure to follow (produce) the same idioms, so that GDB etc. understand GCC/Rust debug information the same way as what `rustc`/LLVM produces?

Or, are `DW_LANG_Rust`/`DW_LANG_Rust_old` agnostic of the exact producer (`rustc`/LLVM vs. GCC/Rust), meaning that as long as the producer follows the specification, GDB etc. don't care -- in theory as well as in practice?  (I further suppose that if in theory _but not_ in practice, that then points to GDB etc. bugs to be fixed.)

And no, I don't have a specific concern, I was just curious.  (Hope you understood my concern.)

tschwinge:
Aha -- what I mean are things like #514.  ;-) But I suppose that's fine -- we just have to pay attention to such _conventions_.

tromey:
I understand your concern, but there's nothing too much to be worried about.  DWARF varies between compilers but also between different versions of compilers.  Debuggers are naturally resilient to this to some degree; and GDB at least will add compiler-specific workarounds when needed (either by pattern matching or by examining the producer string).  Mentioning the language of the CU is just a good practice and it lets GDB know that it should start up in "rust mode".

PR #506, opened 2021-06-19
Fix: do not use object after it has been moved

dkm:
After std::move(mapping), mapping is unspecified


dkm:
It is working in its current state, but it's not really documented what is valid after moving a mapping, so I guess it's safer to avoid reusing it.

thomasyonug:
LGTM

thomasyonug (approved):

philberty:
Good find.

Off topic: I am kind of abusing std::move in a lot of places. When I finish Traits, I think we should open an issue to talk about memory management and the code structure in general.

philberty (approved):

PR #507, opened 2021-06-19
Add more info in HIR dump for struct

dkm:
Make struct fields more visible in dumps and add the mapping in the dump string.


philberty (approved):

philberty:
I think the IR dumps will also need to be looked, I think S-expressions will help a lot there to format them.

dkm:
> I think the IR dumps will also need to be looked, I think S-expressions will help a lot there to format them.

Yes, I was willing to give it a try as current dumps need some effort to read :)

PR #508, opened 2021-06-19
Make 3 error messages slightly more informational

dkm:
Add more context to some internal error messages (ie. not intended for user)


philberty (approved):

philberty:
I think there are a few more copy-pasted, bad error messages scattered around the place. Good find.

PR #509, opened 2021-06-19
Initial support for module

### LINKED ISSUE #432: `mod` keyword not supported.
```
mod foomod {
    struct Foo {
     f: i32,
     g: u32,
    }
}

fn test() -> foomod::Foo {
    foomod::Foo{
        f:1,
        g:3,
    }
}
```
Ends up with : 
```
gccrs/gcc/testsuite/rust.test/unsupported/struct_field_vis.rs:10:14: error: failed to resolve TypePath: foomod::Foo
gcc/testsuite/rust.test/unsupported/struct_field_vis.rs:10:14: error: unresolved type
gcc/testsuite/rust.test/unsupported/struct_field_vis.rs:11:5: error: unknown root segment in path foomod::Foo lookup foomod
```

The error could be expected because of missing `pub` in `foomod`.
But adding `pub` does not help:

```
mod foomod {
    pub struct Foo {
        pub f: i32,
        pub g: u32,
    }
}

fn test() -> foomod::Foo {
    foomod::Foo{
        f:1,
        g:3,
    }
}
```
Still gives the same error. 

Looks like all definition are currently private and that the compiler is not very friendly when trying to access a private member from outside (but I could be wrong!)

dkm:
Adds name resolution, HIR lowering and type checking. For modules
which should allow for initial multiple file parsing support.

Fixes #432

### DIFF
--- a/gcc/rust/hir/rust-ast-lower-item.h
+++ b/gcc/rust/hir/rust-ast-lower-item.h
@@ -29,6 +29,7 @@
 #include "rust-ast-lower-pattern.h"
 #include "rust-ast-lower-block.h"
 #include "rust-ast-lower-extern.h"
+#include "rust-hir-full-decls.h"
 
 namespace Rust {
 namespace HIR {
@@ -52,6 +53,43 @@ public:
     //   }
 
     return resolver.translated;
+  }
+
+  void visit (AST::Module &module) override
+  {
+    auto crate_num = mappings->get_current_crate ();
+    Analysis::NodeMapping mapping (crate_num, module.get_node_id (),
+				   mappings->get_next_hir_id (crate_num),
+				   mappings->get_next_localdef_id (crate_num));
+
+    // should be lowered from module.get_vis()
+    HIR::Visibility vis = HIR::Visibility::create_public ();
+
+    auto items = std::vector<std::unique_ptr<Item> > ();
+
+    for (auto &item : module.get_items ())
+      {
+	auto transitem = translate (item.get ());
+	items.push_back (std::unique_ptr<Item> (transitem));
+      }
+
+    // should be lowered/copied from module.get_in/outer_attrs()
+    AST::AttrVec inner_attrs;
+    AST::AttrVec outer_attrs;
+
+    translated
+      = new HIR::ModuleBodied (mapping, module.get_name (), module.get_locus (),
+			       std::move (items), std::move (vis),
+			       std::move (inner_attrs),
+			       std::move (outer_attrs));
+
+    mappings->insert_defid_mapping (mapping.get_defid (), translated);
+    mappings->insert_hir_item (mapping.get_crate_num (), mapping.get_hirid (),
+			       translated);
+    mappings->insert_module (mapping.get_crate_num (), mapping.get_hirid (),
+			     static_cast<Module *> (translated));
+    mappings->insert_location (crate_num, mapping.get_hirid (),
+			       module.get_locus ());
   }
 
   void visit (AST::TypeAlias &alias) override
--- a/gcc/rust/resolve/rust-ast-resolve-toplevel.h
+++ b/gcc/rust/resolve/rust-ast-resolve-toplevel.h
@@ -23,6 +23,7 @@
 #include "rust-ast-resolve-type.h"
 #include "rust-ast-resolve-implitem.h"
 #include "rust-ast-full.h"
+#include "rust-name-resolver.h"
 
 namespace Rust {
 namespace Resolver {
@@ -38,6 +39,26 @@ public:
     ResolveTopLevel resolver (prefix);
     item->accept_vis (resolver);
   };
+
+  void visit (AST::Module &module) override
+  {
+    auto path = prefix.append (
+      CanonicalPath::new_seg (module.get_node_id (), module.get_name ()));
+    resolver->get_name_scope ().insert (
+      path, module.get_node_id (), module.get_locus (), false,
+      [&] (const CanonicalPath &, NodeId, Location locus) -> void {
+	RichLocation r (module.get_locus ());
+	r.add_range (locus);
+	rust_error_at (r, "redefined multiple times");
+      });
+
+    resolver->insert_new_definition (module.get_node_id (),
+				     Definition{module.get_node_id (),
+						module.get_node_id ()});
+
+    for (auto &item : module.get_items ())
+      ResolveTopLevel::go (item.get (), path);
+  }
 
   void visit (AST::TypeAlias &alias) override
   {
--- a/gcc/rust/resolve/rust-ast-resolve-type.h
+++ b/gcc/rust/resolve/rust-ast-resolve-type.h
@@ -206,6 +206,9 @@ public:
 
     auto resolver = Resolver::get ();
     NodeId resolved_node = UNKNOWN_NODEID;
+
+    // We may need to change how names are resolved, like described in :
+    // https://github.com/rust-lang/rust/blob/1f94abcda6884893d4723304102089198caa0839/compiler/rustc_resolve/src/lib.rs#L1722
     if (!resolver->get_type_scope ().lookup (canonical_path, &resolved_node))
       {
 	rust_error_at (path.get_locus_slow (), "failed to resolve TypePath: %s",
--- a/gcc/rust/resolve/rust-ast-resolve.cc
+++ b/gcc/rust/resolve/rust-ast-resolve.cc
@@ -552,6 +552,10 @@ ResolvePath::resolve_path (AST::PathInExpression *expr)
       auto s = ResolvePathSegmentToCanonicalPath::resolve (seg);
       path = path.append (s);
 
+      // reset state
+      segment_is_type = false;
+      resolved_node = UNKNOWN_NODEID;
+
       if (resolver->get_name_scope ().lookup (path, &resolved_node))
 	{
 	  resolver->insert_resolved_name (seg.get_node_id (), resolved_node);
@@ -562,6 +566,7 @@ ResolvePath::resolve_path (AST::PathInExpression *expr)
       // check the type scope
       else if (resolver->get_type_scope ().lookup (path, &resolved_node))
 	{
+	  segment_is_type = true;
 	  resolver->insert_resolved_type (seg.get_node_id (), resolved_node);
 	  resolver->insert_new_definition (seg.get_node_id (),
 					   Definition{expr->get_node_id (),
@@ -610,6 +615,19 @@ ResolvePath::resolve_path (AST::PathInExpression *expr)
 	  // this
 	  return;
 	}
+    }
+
+  // its fully resolved lets mark it as such
+  if (resolved_node != UNKNOWN_NODEID)
+    {
+      if (segment_is_type)
+	resolver->insert_resolved_type (expr->get_node_id (), resolved_node);
+      else
+	resolver->insert_resolved_name (expr->get_node_id (), resolved_node);
+
+      resolver->insert_new_definition (expr->get_node_id (),
+				       Definition{expr->get_node_id (),
+						  parent});
     }
 }
 
--- a/gcc/rust/resolve/rust-name-resolver.h
+++ b/gcc/rust/resolve/rust-name-resolver.h
@@ -30,7 +30,7 @@ namespace Resolver {
 class Rib
 {
 public:
-  // Rusts uses local_def_ids assigned by def_collector on the AST
+  // Rust uses local_def_ids assigned by def_collector on the AST
   // lets use NodeId instead
   Rib (CrateNum crateNum, NodeId node_id)
     : crate_num (crateNum), node_id (node_id)
--- a/gcc/rust/typecheck/rust-hir-type-check-expr.h
+++ b/gcc/rust/typecheck/rust-hir-type-check-expr.h
@@ -936,6 +936,10 @@ public:
     if (tyseg->get_kind () == TyTy::TypeKind::ERROR)
       return;
 
+    // this is the case where the name resolver has already fully resolved the
+    // name, which means all the work is already done.
+    bool name_resolved_fully = offset >= expr.get_num_segments ();
+
     if (expr.get_num_segments () == 1)
       {
 	Location locus = expr.get_segments ().back ().get_locus ();
@@ -1052,7 +1056,7 @@ public:
       {
 	rust_assert (path_resolved_id == resolved_node_id);
       }
-    else
+    else if (!name_resolved_fully)
       {
 	resolver->insert_resolved_name (expr.get_mappings ().get_nodeid (),
 					resolved_node_id);
@@ -1214,6 +1218,8 @@ private:
     for (size_t i = 0; i < expr.get_num_segments (); i++)
       {
 	HIR::PathExprSegment &seg = expr.get_segments ().at (i);
+
+	bool have_more_segments = (expr.get_num_segments () - 1 != i);
 	bool is_root = *offset == 0;
 	NodeId ast_node_id = seg.get_mappings ().get_nodeid ();
 
@@ -1238,6 +1244,7 @@ private:
 	    resolver->lookup_resolved_type (ast_node_id, &ref_node_id);
 	  }
 
+	// ref_node_id is the NodeId that the segments refers to.
 	if (ref_node_id == UNKNOWN_NODEID)
 	  {
 	    if (is_root)
@@ -1265,27 +1272,33 @@ private:
 
 		return new TyTy::ErrorType (expr.get_mappings ().get_hirid ());
 	      }
+
 	    return root_tyty;
 	  }
 
-	// FIXME
-	// modules are not going to have an explicit TyTy.In this case we
-	// can probably do some kind of check. By looking up if the HirId ref
-	// node is a module and continue. If the path expression is single
-	// segment of module we can error with expected value but found module
-	// or something.
-	//
-	// Something like this
-	//
-	// bool seg_is_module = mappings->lookup_module (ref);
-	// if (seg_is_module)
-	//   {
-	//     if (have_more_segments)
-	//       continue;
-	//
-	//     rust_error_at (seg.get_locus (), "expected value");
-	//     return new TyTy::ErrorType (expr.get_mappings ().get_hirid ());
-	//   }
+	auto seg_is_module
+	  = (nullptr
+	     != mappings->lookup_module (expr.get_mappings ().get_crate_num (),
+					 ref));
+
+	if (seg_is_module)
+	  {
+	    // A::B::C::this_is_a_module::D::E::F
+	    //          ^^^^^^^^^^^^^^^^
+	    //          Currently handling this.
+	    if (have_more_segments)
+	      {
+		(*offset)++;
+		continue;
+	      }
+
+	    // In the case of :
+	    // A::B::C::this_is_a_module
+	    //          ^^^^^^^^^^^^^^^^
+	    // This is an error, we are not expecting a module.
+	    rust_error_at (seg.get_locus (), "expected value");
+	    return new TyTy::ErrorType (expr.get_mappings ().get_hirid ());
+	  }
 
 	TyTy::BaseType *lookup = nullptr;
 	if (!context->lookup_type (ref, &lookup))
--- a/gcc/rust/typecheck/rust-hir-type-check-toplevel.h
+++ b/gcc/rust/typecheck/rust-hir-type-check-toplevel.h
@@ -101,6 +101,12 @@ public:
     context->insert_type (struct_decl.get_mappings (), type);
   }
 
+  void visit (HIR::ModuleBodied &module) override
+  {
+    for (auto &item : module.get_items ())
+      TypeCheckTopLevel::Resolve (item.get ());
+  }
+
   void visit (HIR::StructStruct &struct_decl) override
   {
     std::vector<TyTy::SubstitutionParamMapping> substitutions;
--- a/gcc/rust/util/rust-hir-map.cc
+++ b/gcc/rust/util/rust-hir-map.cc
@@ -325,6 +325,28 @@ Mappings::lookup_hir_impl_block (CrateNum crateNum, HirId id)
 }
 
 void
+Mappings::insert_module (CrateNum crateNum, HirId id, HIR::Module *module)
+{
+  rust_assert (lookup_module (crateNum, id) == nullptr);
+
+  hirModuleMappings[crateNum][id] = module;
+}
+
+HIR::Module *
+Mappings::lookup_module (CrateNum crateNum, HirId id)
+{
+  auto it = hirModuleMappings.find (crateNum);
+  if (it == hirModuleMappings.end ())
+    return nullptr;
+
+  auto iy = it->second.find (id);
+  if (iy == it->second.end ())
+    return nullptr;
+
+  return iy->second;
+}
+
+void
 Mappings::insert_hir_implitem (CrateNum crateNum, HirId id,
 			       HirId parent_impl_id, HIR::ImplItem *item)
 {
--- a/gcc/rust/util/rust-hir-map.h
+++ b/gcc/rust/util/rust-hir-map.h
@@ -131,6 +131,9 @@ public:
 			      HIR::ImplBlock *item);
   HIR::ImplBlock *lookup_hir_impl_block (CrateNum crateNum, HirId id);
 
+  void insert_module (CrateNum crateNum, HirId id, HIR::Module *module);
+  HIR::Module *lookup_module (CrateNum crateNum, HirId id);
+
   void insert_hir_implitem (CrateNum crateNum, HirId id, HirId parent_impl_id,
 			    HIR::ImplItem *item);
   HIR::ImplItem *lookup_hir_implitem (CrateNum crateNum, HirId id,
@@ -279,6 +282,7 @@ private:
 
   std::map<DefId, HIR::Item *> defIdMappings;
   std::map<CrateNum, std::map<LocalDefId, HIR::Item *> > localDefIdMappings;
+  std::map<CrateNum, std::map<HirId, HIR::Module *> > hirModuleMappings;
   std::map<CrateNum, std::map<HirId, HIR::Item *> > hirItemMappings;
   std::map<CrateNum, std::map<HirId, HIR::Type *> > hirTypeMappings;
   std::map<CrateNum, std::map<HirId, HIR::Expr *> > hirExprMappings;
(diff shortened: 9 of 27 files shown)


dkm:
This is still on going.
There are several issues:
```
mod foomod {
    pub struct Foo {
        pub foofield: i32,
    }
}

fn main() {
    //    foomod::Foo{f:3i32}
    let t = foomod::Foo{
        foofield: 12,
    };
}
```
Is segfaulting and

```
mod foomod {
    pub fn totobob() {
    }
}
fn main() {
    foomod::totobob();
}
```
gives:
```
compile/torture/mod1.rs:12:9: error: failed to lookup function type
   12 |     pub fn totobob() {
      |         ^
```
Looks like even simple cases are not working, I must have broken something during a rebase. Still progressing slowly :)

philberty:
Can you share the segv? I think your really close to getting this right, It will take me a while to review this and the segv backtrace might give me some hints

philberty:
Awesome work though thanks for bearing with this.

As I said in the other PR, I think you can see that the structure is not that "nice" yet. So i think we should start thinking about better ways to improve the API's and interfaces in the front-end.

dkm:
Sure I can share the segfault :) I'm in the process of rebasing/cleaning my working tree (hence the other small PRs).

dkm:
Beware that I have not yet looked into it as I am more focused on the "nearly working" case.

```
rust1: internal compiler error: Segmentation fault
0xecee9f crash_signal
        ../../gcc/toplev.c:327
0x7fd82096dd5f ???
        ./signal/../sysdeps/unix/sysv/linux/x86_64/sigaction.c:0
0x9c3949 Rust::Resolver::TypeCheckExpr::visit(Rust::HIR::PathInExpression&)
        ../../gcc/rust/typecheck/rust-hir-type-check-expr.h:875
0x9bf5f5 Rust::Resolver::TypeCheckExpr::Resolve(Rust::HIR::Expr*, bool)
        ../../gcc/rust/typecheck/rust-hir-type-check-expr.h:48
0x9b9845 Rust::Resolver::TypeCheckStructExpr::visit(Rust::HIR::StructExprStructFields&)
        ../../gcc/rust/typecheck/rust-hir-type-check.cc:129
0x9c542a Rust::Resolver::TypeCheckStructExpr::Resolve(Rust::HIR::StructExprStructFields*)
        ../../gcc/rust/typecheck/rust-hir-type-check-struct-field.h:38
0x9c542a Rust::Resolver::TypeCheckExpr::visit(Rust::HIR::StructExprStructFields&)
        ../../gcc/rust/typecheck/rust-hir-type-check-expr.h:833
0x9bf5f5 Rust::Resolver::TypeCheckExpr::Resolve(Rust::HIR::Expr*, bool)
        ../../gcc/rust/typecheck/rust-hir-type-check-expr.h:48
0x9c2003 Rust::Resolver::TypeCheckStmt::visit(Rust::HIR::LetStmt&)
        ../../gcc/rust/typecheck/rust-hir-type-check-stmt.h:60
0x9b88e2 Rust::Resolver::TypeCheckStmt::Resolve(Rust::HIR::Stmt*, bool)
        ../../gcc/rust/typecheck/rust-hir-type-check-stmt.h:38
0x9b88e2 operator()
        ../../gcc/rust/typecheck/rust-hir-type-check.cc:97
0x9b88e2 __invoke_impl<bool, Rust::Resolver::TypeCheckExpr::visit(Rust::HIR::BlockExpr&)::<lambda(Rust::HIR::Stmt*)>&, Rust::HIR::Stmt*>
        /usr/include/c++/10/bits/invoke.h:60
0x9b88e2 __invoke_r<bool, Rust::Resolver::TypeCheckExpr::visit(Rust::HIR::BlockExpr&)::<lambda(Rust::HIR::Stmt*)>&, Rust::HIR::Stmt*>
        /usr/include/c++/10/bits/invoke.h:141
0x9b88e2 _M_invoke
        /usr/include/c++/10/bits/std_function.h:291
0x9b8a89 std::function<bool (Rust::HIR::Stmt*)>::operator()(Rust::HIR::Stmt*) const
        /usr/include/c++/10/bits/std_function.h:622
0x9b8a89 Rust::HIR::BlockExpr::iterate_stmts(std::function<bool (Rust::HIR::Stmt*)>)
        ../../gcc/rust/hir/tree/rust-hir-expr.h:2484
0x9b8a89 Rust::Resolver::TypeCheckExpr::visit(Rust::HIR::BlockExpr&)
        ../../gcc/rust/typecheck/rust-hir-type-check.cc:96
0x9bf5f5 Rust::Resolver::TypeCheckExpr::Resolve(Rust::HIR::Expr*, bool)
        ../../gcc/rust/typecheck/rust-hir-type-check-expr.h:48
0x9bfae5 Rust::Resolver::TypeCheckItem::visit(Rust::HIR::Function&)
        ../../gcc/rust/typecheck/rust-hir-type-check-item.h:81
0x9b904b Rust::Resolver::TypeCheckItem::Resolve(Rust::HIR::Item*)
        ../../gcc/rust/typecheck/rust-hir-type-check-item.h:40
```

gcc/rust/hir/rust-ast-lower-item.h
```diff
@@ -53,6 +54,42 @@ class ASTLoweringItem : public ASTLoweringBase
...
 
+  void visit (AST::ModuleBodied &module) override
+  {
+    auto crate_num = mappings->get_current_crate ();
+    Analysis::NodeMapping mapping (crate_num, module.get_node_id (),
+				   mappings->get_next_hir_id (crate_num),
+				   mappings->get_next_localdef_id (crate_num));
+
+    // should be lowered from module.get_vis()
+    HIR::Visibility vis = HIR::Visibility::create_public ();
```
philberty, line 65: I think there is a case to make a much simpler HIR::visibility, an enum might just work: https://doc.rust-lang.org/stable/nightly-rustc/rustc_hir/hir/enum.VisibilityKind.html

gcc/rust/resolve/rust-ast-resolve-toplevel.h
```diff
@@ -39,6 +41,29 @@ class ResolveTopLevel : public ResolverBase
     item->accept_vis (resolver);
   };
 
+  void visit (AST::ModuleBodied& module) override {
+    auto path = prefix.append (CanonicalPath(module.get_name()));
+    resolver->get_name_scope ().insert (
+      path, module.get_node_id (), module.get_locus (), false,
```
philberty, line 47: I think the name resolution stuff could be extracted into a single commit a test case like:

```rust
mod foo {
  struct A;
}

fn main() {}
```

Might work since the path expression stuff will not be invoked.

philberty (changes requested):
I think this looks really good, I am going to check out your branch and try something since I think your 99% of the way there.

I think the only change needed would be to potentially extract this into 2-3 commits.

1. name resolution
2. HIR lowering
3. Type resolution and finish PathExpression work + testcases

Nice job :)

gcc/rust/typecheck/rust-hir-type-check-toplevel.h
```diff
@@ -100,6 +101,12 @@ class TypeCheckTopLevel : public TypeCheckBase
     context->insert_type (struct_decl.get_mappings (), type);
   }
 
+  void visit (HIR::ModuleBodied &module) override
+  {
+    for (auto &item : module.get_items ())
+      TypeCheckItem::Resolve (item.get ());
```
philberty, line 107: This should be:

```c++
    for (auto &item : module.get_items ())
      TypeCheckTopLevel::Resolve (item.get ());
```

gcc/rust/typecheck/rust-hir-type-check-expr.h
```diff
@@ -1146,6 +1147,8 @@ class TypeCheckExpr : public TypeCheckBase
 		rust_error_at (seg.get_locus (), "reverse lookup failure");
 		return new TyTy::ErrorType (expr.get_mappings ().get_hirid ());
 	      }
+            // MARC: why is this not an error to have a NodeId not referering to any HIR?
+            // Is it because we may be using a to-be-completed HIR graph ?
 	    return root_tyty;
```
philberty, line 1152: The reason it is not an error is consider consider a test case:

```rust
struct <T>Foo(T);

impl Foo<i32> {
  fn test() ...
}

impl Foo<f32> {
  fn test() ...
}

fn main() {
  let a;
  a = Foo::new();  
}
```

In that path of Foo::new we must take into consideration any generic arguments to determine which new function we should be calling to. So this is where the Path Probe takes place over in: https://github.com/Rust-GCC/gccrs/blob/7d3157eec84ef8620f4648200f54b293e561b8c4/gcc/rust/typecheck/rust-hir-path-probe.h#L42

This means we take the type that has been resolved for the segments up to this point and use it as the receiver to match an impl block for that type and then look for any impl items which match the segment we want.

philberty:
I've pushed this commit onto a branch: https://github.com/Rust-GCC/gccrs/commit/c361bec2608c7e579af3a8d5ff8a283071af610d

you should be able to cherry-pick in and squash if you like.

It allows for the compilation of:

```rust
pub struct Bar {
    pub f: i32,
}

mod foomod {
    pub struct Foo {
        pub prout: i32,
    }

    pub fn totoprout() -> i32 {
        return 3i32;
    }
}

pub fn toto() {}

fn test() -> Bar {
    Bar { f: 23i32 }
}

fn main() {
    let a;
    a = foomod::Foo { prout: 3i32 };

    toto();

    let b;
    b = foomod::totoprout();
}
```

gcc/rust/typecheck/rust-hir-type-check-toplevel.h
dkm: fixed by your changes

gcc/rust/typecheck/rust-hir-type-check-expr.h
dkm: Ok, thanks, I think I understand a bit better now!

dkm:
> I think this looks really good, I am going to check out your branch and try something since I think your 99% of the way there.
> 
> I think the only change needed would be to potentially extract this into 2-3 commits.
> 
>     1. name resolution
> 
>     2. HIR lowering
> 
>     3. Type resolution and finish PathExpression work + testcases
> 
> 
> Nice job :)

Ok ! I'll first add few tests and if everything is fine, I'll split the changes in 3. I'll squash your changes and mark you as co-author if that's ok with you ?

dkm:
Haven't had much time for this :/
There is still problem with name resolution. 
This:
```
mod A {
    pub mod B {
        pub mod C {
            pub struct Foo {
            }
        }
    }
}

impl A::B::C::Foo {
    pub fn new() -> Self {
        A::B::C::Foo {
            f: 23i32,
        }
    }
}
fn main() {
    let _a = A::B::C::Foo::new();
}
```
raises this error:
```
error: failed to resolve TypePath: A::B::C::Foo
```
The error comes from the lookup method `Scope::lookup()` that iterates over Ribs but never queries the one having the name defined. I think I'm missing something to correctly route the search instead of simply going up the Rib stack.

And for other simple case like:
```
mod A {
    pub struct Foo {
        pub f: i32,
    }
}

fn main() {
    let _b = A::Foo {
        f: 33i32,
    };
}
```
GCC will ICE after getting `null` from `resolve_root_path( "A::Foo::[C: 0 Nid: 15 Hid: 26]")` in ` void visit (HIR::PathInExpression &expr)`.

gcc/rust/resolve/rust-ast-resolve-type.h
```diff
@@ -193,24 +193,52 @@ class ResolveRelativeTypePath
...
 
     CanonicalPath lookup = canonical_path;
-    if (!prefix.is_error ())
+    if (!prefix.is_empty ())
       lookup = prefix.append (canonical_path);
 
     auto resolver = Resolver::get ();
     NodeId resolved_node = UNKNOWN_NODEID;
+
+    // Lookup type in the following priority order:
```
dkm, line 210: I think here, I need to what's documented here : https://github.com/rust-lang/rust/blob/1f94abcda6884893d4723304102089198caa0839/compiler/rustc_resolve/src/lib.rs#L1722
When the type can't be found in the type NS, then we need to look for it in modules (and others element as documented). Is my understanding correct here? Currently, I'm trying to see if this would work by doing a global search in all Ribs (instead of targeting only the ones related to the module). It seems to work, but then a similar lookup is done so I need do apply the same workaround.

gcc/rust/resolve/rust-name-resolver.h
```diff
@@ -321,14 +327,25 @@ class Scope
...
     bool ok = false;
-    iterate ([&] (Rib *r) mutable -> bool {
-      if (r->decl_was_declared_here (defId))
-	{
-	  ok = true;
-	  r->append_reference_for_def (defId, refId);
-	}
-      return true;
-    });
+
```
dkm, line 330: This is where the defId is looked up only in the type NS and where it should probably use the same search behavior (ie. search in modules if not found in type NS).

philberty:
This was so close:

```
diff --git a/gcc/rust/resolve/rust-ast-resolve-item.h b/gcc/rust/resolve/rust-ast-resolve-item.h
index 923faf8b416..bc9aa4a9418 100644
--- a/gcc/rust/resolve/rust-ast-resolve-item.h
+++ b/gcc/rust/resolve/rust-ast-resolve-item.h
@@ -226,13 +226,14 @@ public:
     resolver->push_new_type_rib (resolver->get_type_scope ().peek ());
     resolver->push_new_label_rib (resolver->get_type_scope ().peek ());
 
-    auto path = CanonicalPath::new_seg (module.get_node_id(), module.get_name ());
+    auto path
+      = CanonicalPath::new_seg (module.get_node_id (), module.get_name ());
 
     for (auto &item : module.get_items ())
-      ResolveTopLevel::go (item.get (), path);
+      ResolveTopLevel::go (item.get ());
 
     for (auto &item : module.get_items ())
-      ResolveItem::go(item.get());
+      ResolveItem::go (item.get ());
 
     resolver->get_name_scope ().pop ();
     resolver->get_type_scope ().pop ();
diff --git a/gcc/rust/resolve/rust-ast-resolve-toplevel.h b/gcc/rust/resolve/rust-ast-resolve-toplevel.h
index c28a0820e52..d6ca42018e8 100644
--- a/gcc/rust/resolve/rust-ast-resolve-toplevel.h
+++ b/gcc/rust/resolve/rust-ast-resolve-toplevel.h
@@ -42,8 +42,8 @@ public:
 
   void visit (AST::Module &module) override
   {
-    auto path = prefix.append (CanonicalPath::new_seg (module.get_node_id(),
-                                                       module.get_name()));
+    auto path = prefix.append (
+      CanonicalPath::new_seg (module.get_node_id (), module.get_name ()));
     resolver->get_name_scope ().insert (
       path, module.get_node_id (), module.get_locus (), false,
       [&] (const CanonicalPath &, NodeId, Location locus) -> void {
@@ -53,15 +53,13 @@ public:
       });
 
     resolver->insert_new_definition (module.get_node_id (),
-                                     Definition{
-                                       module.get_node_id (),
-                                       module.get_node_id ()
-                                     });
+                                    Definition{module.get_node_id (),
+                                               module.get_node_id ()});
 
     // This crashes the compiler. Resolving top level is OK, but looks like it's
     // missing parts.
-    // for (auto &item : module.get_items())
-    //   ResolveTopLevel::go (item.get(), path);
+    for (auto &item : module.get_items ())
+      ResolveTopLevel::go (item.get (), path);
   }
 
   void visit (AST::TypeAlias &alias) override
```

philberty:
I think the error handling in the resolve root within typechecking is a bit unclear how it should work. It is also missing some error cases.

So when I tested the branch it was failing to resolve the path full within the typechecker. I think I will add some documentation of this, as it is also failing to handle some error cases fully which was causing the crash for you.

Because the typechecker was not resolving the segment it should have raised an error here: https://github.com/Rust-GCC/gccrs/blob/5e723d0b7a85d109f01196b264a25823e77496e1/gcc/rust/typecheck/rust-hir-type-check-expr.h#L1240

Since it was able to resolve the module which does not have a type but then went on to fail to resolve the next segment, it should handle this case and return an error but instead returned a nullptr.

Basically the name resolution pass is a bit funny where you have the module and call resolve toplevel with the path prefix which gives you all the top-level names within the root scope rib. Then we you actually call resolve item to drill down into each AST::Item to resolve those names the module case has to do two things, it must call like you already had the Resolvetoplevel again but it should not take the path prefix it needs to add in the relative names of the module for this scope rib before it can finally resolve as normal.

philberty:
Basically you found some bugs in the error handling within the resolve_root_path.

I would suggest that you apply this small diff to your code and try to cleanup/rebase your commits and we can see how the testsuite looks at that point. I think it would be nice to get this work merged soon since its very close to working.

philberty:
From reviewing this patch everything is starting to shape up really well. The changeset should break down into 4/5 commits i think:

1. Add name resolution stuff which will be all of the changes inside the resolve package
2. HIR lowering this is using the HIR::ModuleBodied hangover from the AST this is fine for now we should add a task to desugar this a bit in a separate PR but the current state of this is fine for this PR
3. Backend code generation is fine since it should simply be adding that visitor for the module.
4. Typechecking as far as I can tell you have it correct right now.
5. Testcases

It's difficult to do a more fine-grained code review at the moment I think you need to rebase some commits. I would be interested in seeing what your test suite results look like at that point. 

I can see some commits which are pulling things into cc files which I think look good along with a change to the resolve-unused pass.

Don't worry about duplicate or missing warnings or anything like this for the unused scan right now. It's part of what @thomasyonug is looking into right now.  I can see you tweaked something here in (fedd21987ef69f57d95191ac83634b26513bb9b6)

In relation to that you could create a separate PR after this to add the dead-code scanning for modules by adding a visitor for HIR::ModuleBodied into DeadCode Scan over in:

https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/lint/rust-lint-marklive.h

gcc/rust/resolve/rust-ast-resolve-unused.h
```diff
@@ -43,14 +43,15 @@ class ScanUnused
...
+    return true;
   }
 
   static void Scan ()
   {
     auto resolver = Resolver::get ();
-    resolver->iterate_name_ribs ([&] (Rib *r) -> void { ScanRib (r); });
-    resolver->iterate_type_ribs ([&] (Rib *r) -> void { ScanRib (r); });
-    resolver->iterate_label_ribs ([&] (Rib *r) -> void { ScanRib (r); });
+    resolver->iterate_name_ribs ([&] (Rib *r) -> bool { return ScanRib (r); });
```
philberty, line 52: Let's not change this as part of this PR. I think this scan unused needs some wor, in general, its causing dups for us at the moment.

gcc/rust/resolve/rust-ast-resolve.cc
```diff
@@ -608,6 +613,19 @@ ResolvePath::resolve_path (AST::PathInExpression *expr)
 	  return;
 	}
     }
+
+  // its fully resolved lets mark it as such
+  if (resolved_node != UNKNOWN_NODEID)
+    {
```
philberty, line 619: nice

gcc/rust/resolve/rust-name-resolver.cc
```diff
@@ -0,0 +1,37 @@
+#include <rust-name-resolver.h>
+
+namespace Rust {
+namespace Resolver {
+
+  void Scope::append_reference_for_def (NodeId refId, NodeId defId)
+  {
```
philberty, line 7: I like the move to pull things into implementation files. Just remember to add the GCC Copyright header too

gcc/rust/resolve/rust-name-resolver.h
```diff
@@ -80,12 +85,12 @@ class CanonicalPath
 
   static CanonicalPath create_empty () { return CanonicalPath ({}); }
 
-  bool is_error () const { return segs.size () == 0; }
+  bool is_empty () const { return segs.size () == 0; }
```
philberty, line 88: nice

gcc/rust/resolve/rust-name-resolver.h
```diff
@@ -318,19 +324,30 @@ class Scope
...
-	}
-      return true;
-    });
-    rust_assert (ok);
-  }
+  void append_reference_for_def (NodeId refId, NodeId defId);
+  // {
+  //   bool ok = false;
+
+  //   Resolver *resolver = Resolver::get ();
```
philberty, line 331: we can remove this comment now its in its own implementation file.

gcc/rust/resolve/rust-name-resolver.h
```diff
@@ -438,14 +455,15 @@ class Resolver
...
-  void iterate_type_ribs (std::function<void (Rib *)> cb)
+  void iterate_type_ribs (std::function<bool (Rib *)> cb)
   {
     for (auto it = type_ribs.begin (); it != type_ribs.end (); it++)
       {
 	if (it->first == global_type_node_id)
 	  continue;
 
-	cb (it->second);
+	if (!cb (it->second))
```
philberty, line 465: This is a good refactor change but I think it should be in a separate PR

gcc/rust/typecheck/rust-hir-type-check-expr.h
```diff
@@ -1274,17 +1286,30 @@ class TypeCheckExpr : public TypeCheckBase
...
+	    //          ^^^^^^^^^^^^^^^^
+            // This is an error, we are not expecting a module.
+	    rust_error_at (seg.get_locus (), "expected value");
+            return new TyTy::ErrorType (expr.get_mappings ().get_hirid
+                                        ());
+          }
+
+        TyTy::BaseType *lookup
+	  = nullptr;
 	if (!context->lookup_type (ref, &lookup))
```
philberty, line 1313: This all looks good here. You have a good comment about the nullptr vs error node. There is a nasty bug in here when we have bad name resolution, where you hit an assertion in the caller of resolve_root_node: when we have a path like:

module::foobar

and say for example foobar is not resolved or does not exist the result of this function is going to be a nullptr. I think the fix will be tricky to fix that as part of this PR so lets keep the changes to a minimum to handle modules for now. So long as there are no regressions to the testsuite

gcc/rust/util/rust-hir-map.cc
```diff
@@ -331,6 +331,28 @@ Mappings::lookup_hir_impl_block (CrateNum crateNum, HirId id)
...
+void
+Mappings::insert_module (CrateNum crateNum, HirId id, HIR::Module *module)
+{
+  rust_assert (lookup_module (crateNum, id) == nullptr);
+
+  hirModuleMappings[crateNum][id] = module;
+}
+
+HIR::Module *
+Mappings::lookup_module (CrateNum crateNum, HirId id)
```
philberty, line 343: This will be a nice isolated commit.

gcc/rust/util/rust-hir-map.h
```diff
@@ -25,6 +25,8 @@
 #include "rust-ast-full-decls.h"
 #include "rust-hir-full-decls.h"
 
+//#include "rust-hir-item.h"
```
philberty, line 28: nice find

gcc/rust/resolve/rust-ast-resolve-type.h
philberty: This is interesting to read, I was thinking the fixup we added yesterday here:

which will do a scan toplevel with an empty path prefix a 2nd time which was causing an assertion inside insert_new_definition which was fixed with my PR earlier today should solve this.

I think I would like to see if you can revert this change to the type path name scope resolution and see if the fixes from yesterday and today still allow your test cases to pass.

Basically what happens is when the name resolver starts at the moment we endup with all the AST::Items canonical path's within the top-level scope so for example:


```rust
mod foo {
    pub struct Foo {
        pub f: i32,
    }

    pub fn foofn() -> i32 {
        return 3i32;
    }
}

struct Bar {b:i32}
```

The first scan toplevel gives us the paths:

[foo::Foo , foo:foofn, Bar]

But when we do the ResolveItems and we hit the Module we should do the resolve toplevel on each item then resolve item on each item so we end up with:

[Foo, foofn]
[foo::Foo , foo:foofn, Bar]

This means the relative paths of each item should be resolveable within the new rib that has been created.

The solution inside rustc here is kind of clever. The name resolver inside gccrs might need some thought because, one of the cases we don't funny support properly is the case of AST:::Items within blocks which can be forward declared and constantly having to do scan toplevel is getting a bit confusing.

Lets revert this change, see how it works and maintain a comment to the piece of source code from rustc which might provide a cleaner solution down the line.

philberty (changes requested):
Nice work once again. Lets clean this up and see what the testsuite looks like.

gcc/rust/resolve/rust-ast-resolve.cc
dkm: As I don't need to touch this method, I've simply reverted the .h->.cc move here.

gcc/rust/resolve/rust-name-resolver.h
dkm: done and merged in #625

dkm:
Still need to clean the tests and double-check all the commits are ok. Don't bother reviewing it now, I'll remove the draft state when it's ready :)

gcc/rust/resolve/rust-ast-resolve-toplevel.h
dkm: Done, added in a dedicated test and checked it works with only the changes for name resolution applied.

philberty:
Nice work! The build is passing.

dkm:
I did some cleanup in tests to remove unnecessary dups. I've also added an execution test that is currently FAIL (but must be minor).

gcc/testsuite/rust/execute/torture/mod1.rs
```diff
@@ -0,0 +1,25 @@
...
+    }
+}
+
+fn main() ->i32 {
+    let a = A::B::C::Foo::new();
+    let b = A::B::C::Foo {
+        f: -23i32,
+    };
+
+    a.f - b.f
```
philberty, line 24: does that not need to be a.f + b.f since it is currently:

```
23 - (-23)
```

dkm: woops

philberty (approved):
LGTM to merge I think the commit message still references WIP.

dkm:
> LGTM to merge I think the commit message still references WIP.

Thanks for reviewing it even in draft state. Let me do a last proof reading, remove the draft state and reword first PR message :D

dkm:
Hope to have everything in order now. All commits should still be correct and should not cause failure in buildbot.

PR #512, opened 2021-06-19
Fix various Wformat-diag diagnostics

dkm:
Fix warnings mainly coming from ` being used instead of %< and %>.

Also fixing occurences of :

```
../../gcc/rust/lex/rust-token.cc:126:40: warning: unquoted sequence of 2 consecutive punctuation characters ‘',’ in format [-Wformat-diag]
  126 |        "attempted to get string for '%s', which has no string. "
      |                                        ^~
../../gcc/rust/lex/rust-token.cc:127:39: warning: spurious trailing punctuation sequence ‘.’ in format [-Wformat-diag]
  127 |        "returning empty string instead.",
      |                                       ^
../../gcc/rust/lex/rust-token.cc:126:37: warning: unterminated quote character ‘'’ in format [-Wformat-diag]
  126 |        "attempted to get string for '%s', which has no string. "
      |                                     ^
```
and various occurences of :
```
../../gcc/rust/lex/rust-lex.cc:1326:11: warning: unquoted keyword ‘char’ in format [-Wformat-diag]
 1326 |     "byte char %<%c%> out of range", byte_char);
      |           ^~~~
```
Also using full word for arguments, integer, character, declaration, variable,
floting-point instead of args, int, char, decl, var, float when applicable.

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.h
+++ b/gcc/rust/backend/rust-compile-expr.h
@@ -203,7 +203,7 @@ public:
 	      != 0)
 	    {
 	      rust_fatal_error (expr.get_locus (),
-				"bad float number in literal");
+				"bad floating-point number in literal");
 	      return;
 	    }
 
--- a/gcc/rust/backend/rust-compile-resolve-path.cc
+++ b/gcc/rust/backend/rust-compile-resolve-path.cc
@@ -117,7 +117,7 @@ ResolvePathRef::visit (HIR::PathInExpression &expr)
 	  else
 	    {
 	      rust_error_at (expr.get_locus (),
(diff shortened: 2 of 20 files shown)


dkm:
I'll update the tests when the changes in messages are fine

philberty (approved):
Really nice cleanup, my error messages have been pretty poor I hope to fix some up as I go at the moment.

dkm:
I was trying to reduce a bit the number of warnings :) I guess correct error will come later, with multilang support :)
I'll fix the test later, thanks for the review!

dkm:
```                === rust Summary ===

# of expected passes            2456
# of expected failures          15
```
Will squash now

Issue #195, opened 2021-01-27
Closure types


philberty:
https://doc.rust-lang.org/reference/types/closure.html
https://doc.rust-lang.org/rust-by-example/fn/closures.html
https://doc.rust-lang.org/nomicon/hrtb.html

I think there is stuff in GCC we can take inspiration from the C++ implementation of closures for example:

```c++
int test() {
    int capture = 456;
    auto a = [&](int a) -> int {
        return a + capture;
    };
    return a(123);
}
```

will give us

```c
int test ()
{
  int D.2400;
  int capture;
  struct ._anon_0 a;
  typedef struct ._anon_0 ._anon_0;

  try
    {
      capture = 456;
      a.__capture = &capture;
      D.2400 = test()::<lambda(int)>::operator() (&a, 123);
      return D.2400;
    }
  finally
    {
      capture = {CLOBBER};
      a = {CLOBBER};
    }
}

int test()::<lambda(int)>::operator() (const struct ._anon_0 * const __closure, int a)
{
  int D.2403;
  int & capture [value-expr: __closure->__capture];

  _1 = __closure->__capture;
  _2 = *_1;
  D.2403 = a + _2;
  return D.2403;
}
```

The task list breaks down into:

- [x] Implement ClosureType for the type system
- [x] name resolution to mark captured variables
- [x] backend code generation to generate implicit struct for this closure type and take references, all with must me TREE_ADDRESSABLE 
- [x] generate new function body and take code address of it into the lamba implicit struct
- [x] Support function lang-items https://github.com/rust-lang/rust/blob/7807a694c2f079fd3f395821bcc357eee8650071/library/core/src/ops/function.rs#L54-L71
- [x] Investigate blocking unstable stuff in rustc https://github.com/rust-lang/rust/issues/29625
- [x] rust-abi will convert variable number of arguments into a tuple 

I am currently blocked on this until we get lang-item support into the compiler from my operator overloading branch,

bjorn3:
Closures are interesting. They get a separate MIR body and `DefId` in rustc, but typechecking handles them as part of the parent function.

philberty:
Useful comment from @dkm https://github.com/Rust-GCC/gccrs/pull/523#issuecomment-867565932

bjorn3:
> Maybe similar to the GNU extension for nested function in C https://gcc.gnu.org/onlinedocs/gcc/Nested-Functions.html

[...]

PR #523, opened 2021-06-24
Add support for nested functions

philberty:
This is a missed pieces of functionality for statements in rust.

### DIFF
--- a/gcc/rust/backend/rust-compile-base.h
+++ b/gcc/rust/backend/rust-compile-base.h
@@ -210,6 +210,9 @@ protected:
   void compile_function_body (Bfunction *fndecl,
 			      std::unique_ptr<HIR::BlockExpr> &function_body,
 			      bool has_return_type);
+
+  bool compile_locals_for_block (Resolver::Rib &rib, Bfunction *fndecl,
+				 std::vector<Bvariable *> &locals);
 };
 
 } // namespace Compile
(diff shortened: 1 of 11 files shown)


CohenArthur:
It might be interesting to add a test checking for an error when using a nested function outside of its scope, just to make sure

```rust
fn test() {
   fn nested() {}
}

fn main() {
    nested(); // dg-error etc...
}
```

philberty:
Good point I will add that test case.

Another missing piece is: "error[E0434]: can't capture dynamic environment in a fn item" but it will need a little more work to add that in a general wayl.

CohenArthur:
Yeah I don't know how difficult that would be to implement, considering lambdas can use their dynamic environment etc etc

dkm:
Maybe similar to the GNU extension for nested function in C https://gcc.gnu.org/onlinedocs/gcc/Nested-Functions.html

philberty:
Yeah, I think I will likely need to wait to implement that error until closures are implemented.

philberty:
> Maybe similar to the GNU extension for nested function in C https://gcc.gnu.org/onlinedocs/gcc/Nested-Functions.html

Good point let me add this link to the closures issue #195 

I think a lot of this we get kind of for free. The abstraction we have in place allows us to say this function definition is enclosed within a parent scope to be able to reference outside variables. I think in theory closures are easy 'ish to implement in terms of GCC and this abstraction was used in gccgo so its full of them :D

PR #525, opened 2021-06-25
Support BorrowExpr, DereferenceExpr, NegationExpr in marking live symbols and corresponding test cases.

thomasyonug:
Support BorrowExpr, DereferenceExpr, NegationExpr in marking live symbols and corresponding test cases.

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.h
+++ b/gcc/rust/backend/rust-compile-expr.h
@@ -354,7 +354,7 @@ public:
   void visit (HIR::NegationExpr &expr) override
   {
     auto op = expr.get_expr_type ();
-    auto negated_expr = CompileExpr::Compile (expr.get_expr (), ctx);
+    auto negated_expr = CompileExpr::Compile (expr.get_expr ().get (), ctx);
     auto location = expr.get_locus ();
 
     translated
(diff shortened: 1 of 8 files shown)


thomasyonug:
I was wondering that this pr should be squashed or keep separate?

philberty:
I think i prefer keeping this all as separate commits since each one is an individual thing. Nice job.

gcc/rust/backend/rust-compile-expr.h
```diff
@@ -354,7 +354,7 @@ class CompileExpr : public HIRCompileBase
   void visit (HIR::NegationExpr &expr) override
   {
     auto op = expr.get_expr_type ();
-    auto negated_expr = CompileExpr::Compile (expr.get_expr (), ctx);
+    auto negated_expr = CompileExpr::Compile (expr.get_expr ().get (), ctx);
```
philberty, line 357: Some day I hope we refactor these to take a reference to the unique_ptr to enforce the ownership or take an lvalue ref to the object instead of the pointer. This might be something you can help me with later on if you have time.

philberty (approved):

gcc/rust/backend/rust-compile-expr.h
thomasyonug: yeah, sure. :)

PR #526, opened 2021-06-25
Add unit-type support for Identifier expressions

philberty:
Unit type can be reference via an identifier expression. Other type references via an identifier expression are not supported yet.

### DIFF
--- a/gcc/testsuite/rust/compile/torture/unit_type3.rs
+++ b/gcc/testsuite/rust/compile/torture/unit_type3.rs
@@ -1,0 +1,6 @@
+struct S;
+
+fn main() {
+    let s = S;
+    // { dg-warning "unused name" "" { target *-*-* } .-1 }
+}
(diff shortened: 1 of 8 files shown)


gcc/testsuite/rust/compile/torture/unit_type3.rs
```diff
@@ -0,0 +1,6 @@
+struct S;
+
+fn main() {
+    let s = S;
+    // { dg-warning "unused name" "" { target *-*-* } .-1 }
```
CohenArthur, line 5: nit: If you don't care about the warning you can also use `let _s = S;` since that's working now

CohenArthur (approved):

gcc/testsuite/rust/compile/torture/unit_type3.rs
philberty: good point will fix.

Issue #511, opened 2021-06-19
rust has two kinds of integer literal


tromey:
I tried this code:

```rust
fn first(x : (i32, i32)) -> i32 {
    0_0 + x.0_0
}

pub fn main() {
    let y = first((2, 3));
}
```

This should give an error.  It's not very well documented (maybe not at all) but Rust has two kinds of integer literal: the ordinary kind that can use `_` as a separator, and also the "tuple" kind, where a separator cannot be used.

rustc gives an error on this code:
```
error[E0609]: no field `0_0` on type `(i32, i32)`
 --> tuple.rs:2:13
  |
2 |     0_0 + x.0_0
  |             ^^^
```

philberty:
Good find I was not aware of this. I think this is a bug in the parser where we have the TupleIndexExpr:

https://github.com/Rust-GCC/gccrs/blob/f2d9faf3862a75247adbd6c233c889010902a34b/gcc/rust/ast/rust-expr.h#L1386

The TupleIndex is simply a typedef to int so I reckon the parser is reusing the normal flow for parsing an ineger which supports the 0_0 style but should not support it in this case. I think the bug is here:

https://github.com/Rust-GCC/gccrs/blob/f2d9faf3862a75247adbd6c233c889010902a34b/gcc/rust/parse/rust-parse-impl.h#L14076

Its actually doing an atoi on the string value.

SimplyTheOther:
I believe that this is caused by the lexer ignoring the underscores when tokenising integer literals. A way to solve this could be to retain the underscores in the stored string representation of the integer literal, but this may lead to issues with non-decimal integer literals (i.e. they would have to be retained as well instead of converted to decimal in the lexer).

PR #527, opened 2021-06-25
Reject non-pure decimal tuple indexes.

dkm:
Tuple indexes should be pure decimal integer literals. But the parser just sees an
integer literal. Fix this by introducing a new type hint CORETYPE_PURE_DECIMAL. This
doesn't conflict with any existing type hints since a pure decimal doesn't have a
type suffix.

Introduce a new method is_pure_decimal that the parser can use in parse_tuple_index_expr.
get_type_hint will return CORETYPE_UNKNOWN for pure decimals.

parse_decimal_int_or_float will check whether there are no execissive leading zeros.
parse_in_decimal checks the literal doesn't contain any underscores.

Add two testcases. bad_tuple_index.rs with all variants of integer literals which aren't
pure decimals. And tuple_index.rs with various correct tuple indexes.

Resolves: https://github.com/Rust-GCC/gccrs/issues/511

### DIFF
--- a/gcc/rust/lex/rust-lex.cc
+++ b/gcc/rust/lex/rust-lex.cc
@@ -942,23 +942,26 @@ Lexer::parse_in_exponent_part ()
 	}
 
       // parse another decimal number for exponent
-      auto str_length_pair = parse_in_decimal ();
-      str += str_length_pair.first;
-      additional_length_offset += str_length_pair.second;
+      auto str_length = parse_in_decimal ();
+      str += std::get<0> (str_length);
+      additional_length_offset += std::get<1> (str_length);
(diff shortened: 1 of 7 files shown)


philberty (approved):
LGTM.

I think this is a better solution than rustc see: https://godbolt.org/z/e4aTz4rec

I think we can make our version the same by reusing this piece of code. I think their TupleIndexExpr takes the Token instead of an actual int like we do, so they can check that type hint provided for that token . As their error is not a parser error its an error with finding the field.

philberty:
It could be worthwhile checking if this is the same for array index expressions as well.

PR #528, opened 2021-06-26
Marking live symbols in tuple expr, array expr, grouped expr, type cast expr, lazy boolean expr.

thomasyonug:
Marking live symbols in tuple expr, array expr, grouped expr, type cast expr, lazy boolean expr.


thomasyonug:
These commits are kind of similar, maybe squash to one commit maybe more clear?

philberty:
@thomasyonug I think this is fine :D. Also I know alot of features are mising in the compiler so if you run into issues with that let me know, we can likely leave it to be filled out later on at some point.

philberty (approved):

thomasyonug:
> @thomasyonug I think this is fine :D. Also I know alot of features are mising in the compiler so if you run into issues with that let me know, we can likely leave it to be filled out later on at some point.

Sure, I collected some unimplemented features in my kanban. 
Most of them are recorded in that feature doc, and the missed others I will create corresponding issues. :)

PR #529, opened 2021-06-26
Marking live symbol in break, return, while, whilelet, for expr

thomasyonug:
Marking live symbol in break, return, while, whilelet, for expr without test case for last two due to unimplemented.

### DIFF
--- a/gcc/rust/hir/tree/rust-hir-expr.h
+++ b/gcc/rust/hir/tree/rust-hir-expr.h
@@ -3452,6 +3452,8 @@ public:
 
   void accept_vis (HIRVisitor &vis) override;
 
+  std::unique_ptr<Expr> &get_iterator_expr () { return iterator_expr; }
+
 protected:
   /* Use covariance to implement clone function as returning this object rather
    * than base */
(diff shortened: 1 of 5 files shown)


gcc/rust/hir/tree/rust-hir-expr.h
```diff
@@ -3387,6 +3387,8 @@ class WhileLetLoopExpr : public BaseLoopExpr
 
   void accept_vis (HIRVisitor &vis) override;
 
+  std::unique_ptr<Expr> &get_cond () { return condition; }
```
philberty, line 3390: is this used?

philberty (approved):

gcc/rust/hir/tree/rust-hir-expr.h
thomasyonug: Yes, in WhileLetLoopExpr visit, I need to visit the cond.

philberty: Sorry I missed that :)

PR #532, opened 2021-06-28
Marking live symbol for struct construction.

thomasyonug:
Marking live symbol for struct construction.

### DIFF
--- a/gcc/rust/lint/rust-lint-marklive.cc
+++ b/gcc/rust/lint/rust-lint-marklive.cc
@@ -91,31 +91,39 @@ MarkLive::go (HIR::Crate &crate)
 void
 MarkLive::visit (HIR::PathInExpression &expr)
 {
-  NodeId ast_node_id = expr.get_mappings ().get_nodeid ();
-  NodeId ref_node_id = UNKNOWN_NODEID;
-  if (resolver->lookup_resolved_name (ast_node_id, &ref_node_id))
-    {
-      Resolver::Definition def;
-      if (!resolver->lookup_definition (ref_node_id, &def))
-	{
-	  rust_error_at (expr.get_locus (),
-			 "unknown reference for resolved name");
-	  return;
-	}
-      ref_node_id = def.parent;
-      HirId ref;
-      if (!mappings->lookup_node_to_hir (expr.get_mappings ().get_crate_num (),
-					 ref_node_id, &ref))
-	{
-	  rust_error_at (expr.get_locus (), "reverse lookup failure");
-	  return;
-	}
-      if (scannedSymbols.find (ref) == scannedSymbols.end ())
-	{
-	  worklist.push_back (ref);
-	}
-      liveSymbols.emplace (ref);
-    }
+  expr.iterate_path_segments ([&] (HIR::PathExprSegment &seg) -> bool {
+    NodeId ast_node_id = seg.get_mappings ().get_nodeid ();
+    NodeId ref_node_id = UNKNOWN_NODEID;
+    HirId ref;
+    if (resolver->lookup_resolved_name (ast_node_id, &ref_node_id))
+      {
+	Resolver::Definition def;
+	if (!resolver->lookup_definition (ref_node_id, &def))
+	  {
+	    rust_error_at (seg.get_locus (),
+			   "unknown reference for resolved name");
+	    return false;
+	  }
+	ref_node_id = def.parent;
+      }
+    else if (!resolver->lookup_resolved_type (ast_node_id, &ref_node_id))
+      {
+	return false;
+      }
+
+    if (!mappings->lookup_node_to_hir (seg.get_mappings ().get_crate_num (),
+				       ref_node_id, &ref))
+      {
+	rust_error_at (seg.get_locus (), "reverse lookup failure");
+	return false;
+      }
+    if (scannedSymbols.find (ref) == scannedSymbols.end ())
+      {
+	worklist.push_back (ref);
+      }
+    liveSymbols.emplace (ref);
+    return true;
+  });
 }
 
 void
--- a/gcc/rust/lint/rust-lint-scan-deadcode.h
+++ b/gcc/rust/lint/rust-lint-scan-deadcode.h
@@ -46,20 +46,34 @@ public:
   void visit (HIR::Function &function) override
   {
     HirId hirId = function.get_mappings ().get_hirid ();
-    if (live_symbols.find (hirId) == live_symbols.end ())
-      {
-	rust_warning_at (function.get_locus (), 0,
-			 "function is never used: %<%s%>",
-			 function.get_function_name ().c_str ());
-	return;
-      }
+    warning (hirId, function.get_locus (), "function",
+	     function.get_function_name (), "used");
+  }
+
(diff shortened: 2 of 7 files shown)


thomasyonug:
As for UnusedScan, I think that making it only warns for variables and labels may be the best way. Because the unused variable is another totally different analysis pass, I can't finish it quickly.

philberty:
> As for UnusedScan, I think that making it only warns for variables and labels may be the best way. Because the unused variable is another totally different analysis pass, I can't finish it quickly.

That's a good point, I wouldn't worry about removing as part of this PR right now, if its a pain. We could update that ScanUnused to be able to have some kind of enum to say what type of symbol it is, variable, label etc which will likely solve the problem for us.

Nice work btw, you are in a really good place with regards to the google summer of code proposal now.

thomasyonug:
This pr remains some test cases can not pass temporary.
case1, I don't know what's the hir structure of some codes.
```
let a: i32 = Foo::test();
```

case2, the self feature is kind of complicate.
```
impl Rectangle {
    fn from(p1: Point, p2: Point) -> Self {
        Self { p1, p2 }
    }
}
```

philberty:
Self for functions is a little bit messy at the moment, I will be doing work to help with that as part of traits so this can be something we leave to the end if we have time.

philberty:
As for PathExpressions, this is probably the most complicated piece of rust in my opinion. @dkm has been working on modules and I think he knows what I mean. If I get time I might look into Paths for you to see if I can figure it out for you.

thomasyonug:
> Self for functions is a little bit messy at the moment, I will be doing work to help with that as part of traits so this can be something we leave to the end if we have time.

After a night, I realized that it's not necessary to handle Self, cause when the analysis reaches the Self, it's definitely have reached the corresponding struct PathExpressions. like
```
let rect = Rectangle::from(p1, p2);
                 ^^^^^^^
```
So the only thing is to handle the trait method call correctly I think.

thomasyonug:
I think I got it right. :)

gcc/rust/lint/rust-lint-marklive.cc
```diff
@@ -91,31 +91,39 @@ MarkLive::go (HIR::Crate &crate)
...
-	  rust_error_at (expr.get_locus (), "reverse lookup failure");
-	  return;
-	}
-      if (scannedSymbols.find (ref) == scannedSymbols.end ())
-	{
-	  worklist.push_back (ref);
-	}
-      liveSymbols.emplace (ref);
-    }
+  expr.iterate_path_segments ([&] (HIR::PathExprSegment &seg) -> bool {
```
philberty, line 94: I think you got it right here. Baiscaly PathExpressions are:

A::B::C, but they might also contain generic arguments per segment. So when we do path resolution we iterate each segment for type resolution. When it comes to compilation to GENERIC this is simplified because:

Segment {
  mappings,
  path
}

PathExpr {
  list<Segment>
  mappings
}

So this is a simplified version of the path expression, but once type resolution runs it should be possible to simply call all this code once based on the hir mappings for the whole PathInExpression. 

I don't think you need to change anything here, lets see how this works for now.

philberty: https://github.com/Rust-GCC/gccrs/blob/bab6f6bcb299ac42096783cda32b8f11da4288ca/gcc/rust/backend/rust-compile-resolve-path.cc#L31

gcc/rust/resolve/rust-ast-resolve-unused.h
```diff
@@ -35,8 +36,17 @@ class ScanUnused
       bool ok = r->lookup_canonical_path (decl_node_id, &ident);
       rust_assert (ok);
 
+      Analysis::Mappings *map = Analysis::Mappings::get ();
+      HirId decl_hir_id;
+      HIR::Item *finded_item = nullptr;
```
philberty, line 41: I think you mean:

found_item instead of finded_item

gcc/rust/resolve/rust-ast-resolve-unused.h
```diff
@@ -48,7 +58,7 @@ class ScanUnused
   {
     auto resolver = Resolver::get ();
     resolver->iterate_name_ribs ([&] (Rib *r) -> void { ScanRib (r); });
-    resolver->iterate_type_ribs ([&] (Rib *r) -> void { ScanRib (r); });
+    // resolver->iterate_type_ribs ([&] (Rib *r) -> void { ScanRib (r); });
```
philberty, line 61: nice solution.

philberty (changes requested):
I think all you need to do is fix the name finded_item to found_item

gcc/rust/resolve/rust-ast-resolve-unused.h
thomasyonug: Oops, that's embarrassing.

gcc/rust/lint/rust-lint-scan-deadcode.h
```diff
@@ -46,20 +46,34 @@ class ScanDeadcode : public MarkLiveBase
...
 private:
   std::set<HirId> live_symbols;
-  // std::set<HirId> dead_codes;
+  Resolver::Resolver *resolver;
 
-  ScanDeadcode (std::set<HirId> &live_symbols) : live_symbols (live_symbols){};
+  ScanDeadcode (std::set<HirId> &live_symbols)
+    : live_symbols (live_symbols), resolver (Resolver::Resolver::get ()){};
+
+  void warning (HirId hirId, Location loc, std::string span, std::string name,
```
philberty, line 67: This is really minor but you couldchange this to be:

void warning (HirId hirId, Location loc, const std::string& span, const std::string& name, const std::string& participle) const

gcc/rust/lint/rust-lint-scan-deadcode.h
```diff
@@ -46,20 +46,34 @@ class ScanDeadcode : public MarkLiveBase
...
   {
     HirId hirId = function.get_mappings ().get_hirid ();
-    if (live_symbols.find (hirId) == live_symbols.end ())
-      {
-	rust_warning_at (function.get_locus (), 0,
-			 "function is never used: %<%s%>",
-			 function.get_function_name ().c_str ());
-	return;
-      }
+    warning (hirId, function.get_locus (), "function",
```
philberty, line 49: I think i would prefer a function

warning_unused (...) so you don't need to pass the argument "used"

philberty: actually ignore this comment for now let me think about it and we can change in a future PR

philberty (approved):

gcc/rust/resolve/rust-ast-resolve-unused.h
philberty: Can you put a comment here saying why this was commented out, something like.

"ScanUnused is conflicting with the dead code analysis here on types"

philberty (changes requested):
Sorry just one more change to add a comment about why the ScanUnused had to comment out the type scope I left an example comment for you.

philberty (approved):

PR #538, opened 2021-06-30
exit immediately when the type checker find an error

thomasyonug:
exit immediately when the type checker find an error

### DIFF
--- a/gcc/rust/rust-session-manager.cc
+++ b/gcc/rust/rust-session-manager.cc
@@ -572,6 +572,9 @@ Session::parse_file (const char *filename)
       dump_type_resolution (hir);
     }
 
+  if (saw_errors ())
+    return;
+
   // liveness analysis
   std::set<HirId> live_symbols = Analysis::MarkLive::Analysis (hir);
 


philberty (approved):

bjorn3:
Rustc does the exact opposite. It tries to continue compilation as long as possible. It just marks the typeck table for the function with a type error as having an error, so mir building ignores it.

philberty:
@bjorn3 I think maybe the comment is't 100% accurate this change will exit when the type checking pass completes and errors have been emited. 

So there are cases inside the pass we might get an error_node but the pass may not have necessarily emitted a rust_error_at, in which case it will continue on.

dkm:
Was wondering why the early exit. Is it because currently gccrs does not recover correctly after this ?

thomasyonug:
> Rustc does the exact opposite. It tries to continue compilation as long as possible. It just marks the typeck table for the function with a type error as having an error, so mir building ignores it.

Interesting. Gccrs exit before the next pass when any pass finds an error now. I think maybe we can do the same thing when gccrs become more stable?

thomasyonug:
> Was wondering why the early exit. Is it because currently gccrs does not recover correctly after this ?

The reason why I add return here because when gccrs raise a type error, the output will mix up with a type error and some diagnostic information

tschwinge:
I too was confused by this commit.  ;-)

Can we get a testcase added that shows the "wrong" behavior (when this commit is reverted)?

In general, GCC does the same as described above: continue compilation as much as possible, marking up/invalidating nodes that had an error emitted with `ERROR_MARK` etc., and subsequently ignoring these.

philberty:
Good point, i haven't been taking advantage of error_mark_node in the generation of GENERIC. So each pass in the compiler I have been doing if (saw_errors) return;

Maybe we are at the stage to stop doing that. There are some fatal_errors in the backend that could go away by simply using error_mark_node too.

PR #540, opened 2021-07-01
Add trait-impl support for methods

philberty:
This PR strips HIR::Method and reuses the HIR::Function with a HIR::SelfParam.
It reduces complexity in type checking and GENERIC generation.

### DIFF
--- a/gcc/testsuite/rust/compile/torture/traits3.rs
+++ b/gcc/testsuite/rust/compile/torture/traits3.rs
@@ -1,0 +1,18 @@
+pub trait Foo {
+    fn Bar(self) -> i32;
+    // { dg-warning "unused name .self." "" { target *-*-* } .-1 }
+    // { dg-warning "unused name .Bar." "" { target *-*-* } .-2 }
+}
+
+struct Baz;
+// { dg-warning "struct is never constructed: .Baz." "" { target *-*-* } .-1 }
+
+impl Foo for Baz {
+    fn Bar(self) -> i32 {
+        // { dg-warning "unused name .self." "" { target *-*-* } .-1 }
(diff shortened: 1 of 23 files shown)


gcc/testsuite/rust/compile/torture/traits3.rs
```diff
@@ -0,0 +1,18 @@
+pub trait Foo {
+    fn Bar(self) -> i32;
+    // { dg-warning "unused name .self." "" { target *-*-* } .-1 }
```
CohenArthur, line 3: I think this is outside the scope of this PR and related to @thomasyonug's work, but `gccrs` should not warn here. You are simply 'declaring' a trait's method and not 'defining' it (to speak in C/C++ terms). It should warn on the following code:

```rust
pub trait Foo {
    fn Bar(self) -> i32 {
        -1 // default implementation returns -1 for example
    }
}
```

but not in your snippet.

So there should be a distinction between a default trait method implementation and just a 'declaration' like here for unused variables

I might be misunderstanding the dejagnu directive, please forgive me in that case!

CohenArthur: Done in #541

thomasyonug: I think this warning msg is from the UnusedScan pass.

There are 3 functions here.
1. function decl and def in trait
2. function in impl trait block.
3. function in pure impl block.
And I just test it by rustc, found that rustc only warn the unused function in 3 and ignoring in 1,2.

philberty: Yeah, I think ScanUnused is a bit annoying at this point it's just a generic algorithm that it needs some refinement later on. Its simply iterating all declared names within a Rib and if any name has no references to it therefore it is unused. So it's too general at the moment.

PR #543, opened 2021-07-03
mark live symbol in if expr family except if let expr

thomasyonug:
mark live symbol in if expr family except if let expr

### DIFF
--- a/gcc/rust/lint/rust-lint-marklive.h
+++ b/gcc/rust/lint/rust-lint-marklive.h
@@ -182,6 +182,26 @@ public:
     expr.visit_rhs (*this);
   }
 
+  void visit (HIR::IfExpr &expr) override
+  {
+    expr.get_if_condition ()->accept_vis (*this);
+    expr.get_if_block ()->accept_vis (*this);
+  }
+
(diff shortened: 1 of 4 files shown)


philberty (approved):
What about TupleStructs i am not sure if I have seen an example with that.

thomasyonug:
> What about TupleStructs i am not sure if I have seen an example with that.

Yeah, this is a complicated one. I'll implement it soon.

philberty:
> > What about TupleStructs i am not sure if I have seen an example with that.
> 
> Yeah, this is a complicated one. I'll implement it soon.

Let me know if you need help with it @thomasyonug I think I see how to go about it with your other ADT example.

thomasyonug:
> > > What about TupleStructs i am not sure if I have seen an example with that.
> 
> > 
> 
> > Yeah, this is a complicated one. I'll implement it soon.
> 
> 
> 
> Let me know if you need help with it @thomasyonug I think I see how to go about it with your other ADT example.

Thanks. You are right, it's kind of similar, i'm considering how to abstracting some common function to help impl both of tuple struct and struct struct.

PR #544, opened 2021-07-04
mark live symbol support type alias and tuple struct

thomasyonug:
mark live symbol support type alias and tuple struct

### DIFF
--- a/gcc/rust/lint/rust-lint-scan-deadcode.h
+++ b/gcc/rust/lint/rust-lint-scan-deadcode.h
@@ -79,6 +79,18 @@ public:
       }
   }
 
+  void visit (HIR::TupleStruct &stct) override
+  {
+    // only warn tuple struct unconstructed, and ignoring unused field
+    HirId hirId = stct.get_mappings ().get_hirid ();
+    if (should_warn (hirId))
+      {
+	rust_warning_at (stct.get_locus (), 0, "%s is never %s: %<%s%>",
+			 "struct", "constructed",
+			 stct.get_identifier ().c_str ());
+      }
+  }
+
 private:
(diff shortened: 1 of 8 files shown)


gcc/rust/lint/rust-lint-scan-deadcode.h
```diff
@@ -79,6 +79,18 @@ class ScanDeadcode : public MarkLiveBase
...
   }
 
+  void visit (HIR::TupleStruct &stct) override
+  {
+    // only warn tuple struct unconstructed, and ignoring unused field
+    HirId hirId = stct.get_mappings ().get_hirid ();
+    if (should_warn (hirId))
+      {
+	rust_warning_at (stct.get_locus (), 0, "%s is never %s: %<%s%>",
+			 "struct", "constructed",
```
CohenArthur, line 89: I think doing this makes it harder for people to search for an error's source in the code. Why not do 
```suggestion
	rust_warning_at (stct.get_locus (), 0, "struct is never constructed: %<%s%>",
```
?

thomasyonug: good point. i'll open another cleanup issue since there are some other warning using this way in my pass.

philberty (approved):
LGTM, since your project has progressed so well I think you should spend some time thinking about the code error handling and what you can do to make the code easy to read.

I would start with #550 #545 i think we should also think about adding some comments into the code about how this all works. As I have followed your code closely I understand how it works but I think others may not.

Great work as always.

thomasyonug:
Thanks. I'll start to focus my work on readable code instead of supporting some new features.

PR #546, opened 2021-07-05
shebang handling

dkm:
Mark Wielaard:
> Shebang handling, the first line starting with #! was not done fully
> correct and it isn't necessary to keep track of the shebang line in
> the AST or HIR Crate classes.
> 
> Because an inner attribute also starts with #! the first line isn't
> regarded as a shebang line if the #! is followed by (optional)
> whitespace and comments and a [. In that case the #! is seen as the
> start of an inner attribute.
> 
> I added various testcases that hopefully show the funny things you can
> get when the first line starts with #!.

### DIFF
--- a/gcc/rust/ast/rust-ast-full-test.cc
+++ b/gcc/rust/ast/rust-ast-full-test.cc
@@ -172,12 +172,9 @@ Crate::as_string () const
   rust_debug ("beginning crate recursive as-string");
 
   std::string str ("Crate: ");
-  // add utf8bom and shebang
+  // add utf8bom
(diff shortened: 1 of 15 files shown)


bjorn3:
I presume this is from the mailinglist? Could you link to the respective archive in the future when taking a patch from the mailinglist?

dkm:
Sure! Will do for the future ones.
https://gcc.gnu.org/pipermail/gcc-rust/2021-July/000068.html

bjorn3:
Thanks!

philberty (approved):
LGTM

PR #551, opened 2021-07-05
Initial AssociatedType support

philberty:
Traits can define Type aliasses which must enforce the correct resolution
onto their respective TraitImplBlocks. This PR adds all the necessary building
blocks for the initial support.

More work is needed in checking for trait-impl-item consistency with its respective
Trait item types.

### DIFF
--- a/gcc/rust/ast/rust-path.h
+++ b/gcc/rust/ast/rust-path.h
@@ -406,11 +406,6 @@ protected:
  * ident-only segment) */
 class TypePathSegment
 {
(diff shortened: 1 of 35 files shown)


dkm:
I see some fixes in path, should I rebase #509 over this one?

philberty:
> I see some fixes in path, should I rebase #509 over this one?

I would hold off rebasing until I get this right. We have TypePaths and PathInExpression but rustc hir just has a Path I am trying to find a way to unify these for associated types, so this is all very WIP. I am just pushing the branch to switch between my laptop and desktop here.

PR #552, opened 2021-07-05
UTF-8 BOM handling

dkm:
Mark Wielaard (https://gcc.gnu.org/pipermail/gcc-rust/2021-July/000072.html ) :

> A rust source file can start with a UTF-8 BOM sequence (EF BB
> BF). This simply indicates that the file is encoded as UTF-8 (all rust
> input is interpreted as asequence of Unicode code points encoded in
> UTF-8) so can be skipped before starting real lexing.
> 
> It isn't necessary to keep track of the BOM in the AST or HIR Crate
> classes. So I removed the has_utf8bom flag.
> 
> Also included are a couple of simple tests to show we handle the BOM
> correctly now.


bjorn3:
Is `\r\n` converted to `\n` by the way and lone `\r` rejected? That is what rust requires:

https://github.com/rust-lang/rust/blob/0f6ba39fd89cff295067d974c49fc64b0b52ad18/compiler/rustc_span/src/lib.rs#L1667

It is observable in strings.

bjorn3:
I think `include_str!` requires both bom removal and newline nornalization too, though I haven't checked.

dkm:
I'll add this in a separate issue so that we can implement/check/test newline norm, thanks!

PR #555, opened 2021-07-07
use rust_assert in dead code pass instead of rust_error_at

### LINKED ISSUE #550: DeadCode pass error handling
The dead code pass is used to generate warnings for relevant code. It should not produce errors I think we could take a look at using rust_assert which should be more useful in diagnosing issues with the pass. Take for example:

https://github.com/Rust-GCC/gccrs/blob/210ae4f7b0fea9671482b8f01354fd5b9274f878/gcc/rust/lint/rust-lint-marklive.cc#L237

This function wraps lookup_node_to_hir which creates an interface to ignore errors that might provide bad errors such as:

https://github.com/Rust-GCC/gccrs/blob/210ae4f7b0fea9671482b8f01354fd5b9274f878/gcc/rust/lint/rust-lint-marklive.cc#L116

if node_id_to_hir_id fails the code will continue on to mark the node anyway regardless.

let's remove this function and call lookup_node_to_hir directly where required and we can use the error handling already provided and then use a rust_assert might be the simplest solution, since there should not be any issues with resolving nodes at this stage of the pipeline. If an error did occur it would signify something else going wrong.

thomasyonug:
use rust_assert in dead code pass instead of rust_error_at

Fixed #550

### DIFF
--- a/gcc/rust/lint/rust-lint-marklive.cc
+++ b/gcc/rust/lint/rust-lint-marklive.cc
@@ -92,7 +92,8 @@ void
 MarkLive::visit (HIR::PathInExpression &expr)
 {
   expr.iterate_path_segments ([&] (HIR::PathExprSegment &seg) -> bool {
-    return visit_path_segment (seg);
+    visit_path_segment (seg);
+    return true;
   });
 }
 
@@ -113,12 +114,13 @@ MarkLive::visit (HIR::MethodCallExpr &expr)
 
   // node back to HIR
   HirId ref;
-  node_id_to_hir_id (expr.get_mappings ().get_crate_num (), ref_node_id, ref,
-		     expr.get_locus ());
+  rust_assert (
+    mappings->lookup_node_to_hir (expr.get_mappings ().get_crate_num (),
+				  ref_node_id, &ref));
   mark_hir_id (ref);
 }
 
-bool
+void
 MarkLive::visit_path_segment (HIR::PathExprSegment seg)
 {
   NodeId ast_node_id = seg.get_mappings ().get_nodeid ();
@@ -127,23 +129,18 @@ MarkLive::visit_path_segment (HIR::PathExprSegment seg)
   if (resolver->lookup_resolved_name (ast_node_id, &ref_node_id))
     {
       Resolver::Definition def;
-      if (!resolver->lookup_definition (ref_node_id, &def))
-	{
-	  rust_error_at (seg.get_locus (),
-			 "unknown reference for resolved name");
-	  return false;
-	}
+      rust_assert (resolver->lookup_definition (ref_node_id, &def));
       ref_node_id = def.parent;
     }
   else if (!resolver->lookup_resolved_type (ast_node_id, &ref_node_id))
     {
-      return false;
+      return;
     }
   HirId ref;
-  node_id_to_hir_id (seg.get_mappings ().get_crate_num (), ref_node_id, ref,
-		     seg.get_locus ());
+  rust_assert (
+    mappings->lookup_node_to_hir (seg.get_mappings ().get_crate_num (),
+				  ref_node_id, &ref));
   mark_hir_id (ref);
-  return true;
 }
 
 void
@@ -195,8 +192,9 @@ MarkLive::visit (HIR::IdentifierExpr &expr)
 
   // node back to HIR
   HirId ref;
-  node_id_to_hir_id (expr.get_mappings ().get_crate_num (), ref_node_id, ref,
-		     expr.get_locus ());
+  rust_assert (
+    mappings->lookup_node_to_hir (expr.get_mappings ().get_crate_num (),
+				  ref_node_id, &ref));
   mark_hir_id (ref);
 }
 
@@ -207,8 +205,9 @@ MarkLive::visit (HIR::TypeAlias &alias)
   resolver->lookup_resolved_type (
     alias.get_type_aliased ()->get_mappings ().get_nodeid (), &ast_node_id);
   HirId hir_id;
-  node_id_to_hir_id (alias.get_mappings ().get_crate_num (), ast_node_id,
-		     hir_id, alias.get_locus ());
+  rust_assert (
+    mappings->lookup_node_to_hir (alias.get_mappings ().get_crate_num (),
+				  ast_node_id, &hir_id));
   mark_hir_id (hir_id);
 }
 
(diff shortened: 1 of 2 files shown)


thomasyonug:
Please let me know if I got the point entirely. :)

dkm:
Not sure this is correct. Assertion can be disabled, so any expr used in them should not do anything (no side effect) or the build will not be working with assertion disabled.

```

/* Use gcc_assert(EXPR) to test invariants.  */
#if ENABLE_ASSERT_CHECKING
#define gcc_assert(EXPR)                                                \
   ((void)(!(EXPR) ? fancy_abort (__FILE__, __LINE__, __FUNCTION__), 0 : 0))
#elif (GCC_VERSION >= 4005)
#define gcc_assert(EXPR)                                                \
  ((void)(__builtin_expect (!(EXPR), 0) ? __builtin_unreachable (), 0 : 0))
#else
/* Include EXPR, so that unused variable warnings do not occur.  */
#define gcc_assert(EXPR) ((void)(0 && (EXPR)))
#endif
```

`rust_assert` is simply a macro for `gcc_assert`

dkm:
Not sure this is correct, as assertion can be disabled and any side effect in the expr will be removed.
```
/* Use gcc_assert(EXPR) to test invariants.  */
#if ENABLE_ASSERT_CHECKING
#define gcc_assert(EXPR)                                                \
   ((void)(!(EXPR) ? fancy_abort (__FILE__, __LINE__, __FUNCTION__), 0 : 0))
#elif (GCC_VERSION >= 4005)
#define gcc_assert(EXPR)                                                \
  ((void)(__builtin_expect (!(EXPR), 0) ? __builtin_unreachable (), 0 : 0))
#else
/* Include EXPR, so that unused variable warnings do not occur.  */
#define gcc_assert(EXPR) ((void)(0 && (EXPR)))
#endif
```

gcc/rust/lint/rust-lint-marklive.cc
```diff
@@ -113,12 +114,13 @@ MarkLive::visit (HIR::MethodCallExpr &expr)
 
   // node back to HIR
   HirId ref;
-  node_id_to_hir_id (expr.get_mappings ().get_crate_num (), ref_node_id, ref,
-		     expr.get_locus ());
+  rust_assert (
+    mappings->lookup_node_to_hir (expr.get_mappings ().get_crate_num (),
+				  ref_node_id, &ref));
```
dkm, line 119: When assertion are disabled, `ref` will be uninit here

gcc/rust/lint/rust-lint-marklive.cc
```diff
@@ -127,23 +129,18 @@ MarkLive::visit_path_segment (HIR::PathExprSegment seg)
   if (resolver->lookup_resolved_name (ast_node_id, &ref_node_id))
     {
       Resolver::Definition def;
-      if (!resolver->lookup_definition (ref_node_id, &def))
-	{
-	  rust_error_at (seg.get_locus (),
-			 "unknown reference for resolved name");
-	  return false;
-	}
+      rust_assert (resolver->lookup_definition (ref_node_id, &def));
```
dkm, line 132: same here

dkm (changes requested):

dkm:
Better use something like 
```
ret = lookup(...);
rust_assert (ret);
...
```

gcc/rust/lint/rust-lint-marklive.cc
thomasyonug: Oops, I did't notice that.

dkm:
In fact, this may be correct as we are not using anything older than 4005.

thomasyonug:
> In fact, this may be correct as we are not using anything older than 4005.

Sorry, I don't know what's the 4005 means.

dkm:
> In fact, this may be correct as we are not using anything older than 4005.

I /think/ `gcc_assert` still calls `EXPR`, only `gcc_checking_assert` skips it when disabled. But I could not find any such usage in the code base, so not sure this is really used...

thomasyonug:
> > In fact, this may be correct as we are not using anything older than 4005.
> 
> I /think/ `gcc_assert` still calls `EXPR`, only `gcc_checking_assert` skips it when disabled. But I could not find any such usage in the code base, so not sure this is really used...

Ha, gotcha, thanks. I think it's better to avoid taking a risk.

dkm:
> > In fact, this may be correct as we are not using anything older than 4005.
> 
> Sorry, I don't know what's the 4005 means. Would you mind explain more details to me? Or any reference

Sorry, it's the GCC version.  `4XXX` is for GCC 4, so not a gccrs target :)

gcc/rust/lint/rust-lint-marklive.cc
```diff
@@ -113,12 +114,14 @@ MarkLive::visit (HIR::MethodCallExpr &expr)
 
   // node back to HIR
   HirId ref;
-  node_id_to_hir_id (expr.get_mappings ().get_crate_num (), ref_node_id, ref,
-		     expr.get_locus ());
+  bool ret
```
philberty, line 117: usually we do bool ok = mappings->lookup_node_to_hir

I think 'ok' reads better

gcc/rust/lint/rust-lint-marklive.h
```diff
@@ -266,10 +266,9 @@ class MarkLive : public MarkLiveBase
       tyctx (Resolver::TypeCheckContext::get ()){};
 
   void mark_hir_id (HirId);
-  bool visit_path_segment (HIR::PathExprSegment);
```
philberty, line 269: I think keep this bool visit_path_segment for now

philberty (changes requested):

gcc/rust/lint/rust-lint-marklive.h
thomasyonug: May I ask why shoud I keep this bool for now?

gcc/rust/lint/rust-lint-marklive.cc
thomasyonug: Yeah, sure it is

thomasyonug:
I'll squash the commits after approve

dkm:
I guess @philberty was asking for changing all `ret` to `ok`, not only the one for node_to_hir?

thomasyonug:
> I guess @philberty was asking for changing all `ret` to `ok`, not only the one for node_to_hir?

Yeah, I miss some of them. Thanks for pointing.

dkm (approved):
Looks fine :)

gcc/rust/lint/rust-lint-marklive.h
philberty: it may present other options for more error handling later on.

philberty (approved):

PR #561, opened 2021-07-12
Handle doc comments strings

philberty:
This adds support for inner and outer doc comments within the parser.


bjorn3:
For reference: https://gcc.gnu.org/pipermail/gcc-rust/2021-July/000076.html

PR #567, opened 2021-07-15
warn the unused associated functions

thomasyonug:
warn the unused associated functions

### DIFF
--- a/gcc/rust/lint/rust-lint-marklive.cc
+++ b/gcc/rust/lint/rust-lint-marklive.cc
@@ -106,11 +106,42 @@ MarkLive::go (HIR::Crate &crate)
 void
 MarkLive::visit (HIR::PathInExpression &expr)
 {
-  // We should iterate every path segment in order to mark the function which is
-  // called in the expression
+  // We should iterate every path segment in order to mark the struct which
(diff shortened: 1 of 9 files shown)


gcc/rust/lint/rust-lint-marklive.cc
```diff
@@ -106,11 +106,42 @@ MarkLive::go (HIR::Crate &crate)
...
+    {
+      mark_hir_id (resolved_item->get_mappings ().get_hirid ());
+    }
+  else
+    {
+      HirId parent_impl_id = UNKNOWN_HIRID;
+      HIR::ImplItem *resolved_item
+	= mappings->lookup_hir_implitem (expr.get_mappings ().get_crate_num (),
+					 ref, &parent_impl_id);
+      if (resolved_item != nullptr)
```
philberty, line 140: I wonder should we add an assertion here, or are there problems with other associated items?

philberty (approved):
LGTM

gcc/rust/lint/rust-lint-marklive.cc
thomasyonug: I think maybe a struct constructor here. But we can ignore it since it had been handled in segments iteration.

PR #578, opened 2021-07-20
Remove unused code

philberty:
Remove unused code


tschwinge:
There is still `rust-backend.c` mentioned twice in `gcc/rust/rust-object-export.c`.

There is still `gcc/rust/rust-backend.h`, which still seems to be used; should its (relevant?) content move into `gcc/rust/rust-object-export.h`?

philberty:
> There is still `rust-backend.c` mentioned twice in `gcc/rust/rust-object-export.c`.
> 
> There is still `gcc/rust/rust-backend.h`, which still seems to be used; should its (relevant?) content move into `gcc/rust/rust-object-export.h`?

Yeah this was a messy port of the gccgo interfaces. Lets make an issue to clean this up.

PR #579, opened 2021-07-20
Raw pointer support

### LINKED ISSUE #124: Raw Pointer Support
https://doc.rust-lang.org/book/raw-pointers.html

philberty:
We will need to add the remaining coercion calls for the relevant sites
to get the full support we need but this will serve as the initial support.

Fixes #124

### DIFF
--- a/gcc/rust/typecheck/rust-tyty-coercion.h
+++ b/gcc/rust/typecheck/rust-tyty-coercion.h
@@ -252,6 +252,17 @@ public:
 		   type.as_string ().c_str ());
   }
 
+  virtual void visit (PointerType &type) override
+  {
+    Location ref_locus = mappings->lookup_location (type.get_ref ());
+    Location base_locus = mappings->lookup_location (get_base ()->get_ref ());
+    RichLocation r (ref_locus);
+    r.add_range (base_locus);
+    rust_error_at (r, "expected [%s] got [%s]",
+		   get_base ()->as_string ().c_str (),
+		   type.as_string ().c_str ());
+  }
+
   virtual void visit (ParamType &type) override
   {
     Location ref_locus = mappings->lookup_location (type.get_ref ());
@@ -507,7 +518,19 @@ public:
   }
 
   void visit (ReferenceType &type) override
+  {
+    bool is_valid
+      = (base->get_infer_kind () == TyTy::InferType::InferTypeKind::GENERAL);
+    if (is_valid)
+      {
+	resolved = type.clone ();
+	return;
+      }
 
+    BaseCoercionRules::visit (type);
+  }
+
+  void visit (PointerType &type) override
   {
     bool is_valid
       = (base->get_infer_kind () == TyTy::InferType::InferTypeKind::GENERAL);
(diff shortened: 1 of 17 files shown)


gcc/testsuite/rust/compile/torture/coercion2.rs
```diff
@@ -0,0 +1,20 @@
...
+    // { dg-warning "unused name" "" { target *-*-* } .-1 }
+
+    let c: &mut i32 = &mut 123;
+    // { dg-warning "unused name" "" { target *-*-* } .-1 }
+    let d: *mut i32 = &mut 123;
+    // { dg-warning "unused name" "" { target *-*-* } .-1 }
+
+    let e: &i32 = &mut 123;
+    // { dg-warning "unused name" "" { target *-*-* } .-1 }
+    let f: *const i32 = &mut 123;
```
bjorn3, line 14: In rustc this first creates an `&i32` and then derives a `*const i32` from it.

philberty: Good point. This also implements it like this. The gimple output shows it like this:

```rust
pub fn main() {
    let a: &i32 = &123;
    let b: *const i32 = a;
}
```

```
void main ()
{
  i32 D.201;
  const i32 & const a;
  const i32 * const b;

  _1 = 123;
  D.201 = _1;
  a = &D.201;
  b = a;
}
```

bjorn3: Note that `*const i32` in Rust and `const i32 *` in C are not equivalent. `*const i32` can be casted to `*mut i32` and written to if it was originally derived from a mutable reference. This is not the case for `const i32 *`. My point was that `&mut x as *const i32` does not count as being derived from a mutable reference, but an intermediate immutable reference being created. So `*(&mut x as *const i32 as *mut i32) = 42` is UB (by being equivalent to `*(&*(&mut x) as *const i32 as *mut i32) = 42`), but `*(&mut x as *mut i32 *const i32 as *mut i32) = 42` is allowed.

philberty: Yeah, casts are different. As I found following the rust source code it's not just a simple coercion site it has slightly more subtle rules for primitive types so I am still working on that part.

philberty: GCC also supports this distinction of *const i32 vs const i32 *. That is why I shown the example with the double const.

Its not quite the same syntax as C though so when we have:

`const i32 * const` that means we have a full const variable to a const pointer. If we simply had a mutable variable to a const pointer the type would look like: `const i32 *`

bjorn3: Yeah, I mean that `const i32*` is an incorrect translation for the rust type `*const i32` in the general case. It is only correct when you are certain it was derived from an immutable reference. For example in

```rust
fn write_ptr(p: *const i32, val: i32) {
    *(p as *mut i32) = val;
}
```

`*const i32` **must** be translated as `i32 *` and not `const i32 *`.

philberty: Oh ok thats wierd that you can cast. It looks like you mean we can ignore mutability in casts of pointers?

philberty: Interesting this is not what i expected https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f320cc37d6038d991f70e70e1c008d45

I think its only matters for casts this inial implementation still makes sense to me.

philberty: Do you mean in this example:

```
fn main() {
    let a: *const i32 = &123;
    let b: *mut i32 = (a as *mut i32);
}
```

that since b casts away the const that both a and b should be mutable pointers?

bjorn3: In rust the difference between `*const T` and `*mut T` is only the fact that casting a mutable reference to `*const T` creates an intermediate immutable reference first and that writing to `*const T` requires a cast to `*mut T` first. Other than that they are interchangeable, unlike `&T` and `&mut T`.

gcc/rust/typecheck/rust-tyty-coercion.h
```diff
@@ -1087,6 +1110,75 @@ class ReferenceCoercionRules : public BaseCoercionRules
...
+    if (base_resolved == nullptr
+	|| base_resolved->get_kind () == TypeKind::ERROR)
+      {
+	BaseCoercionRules::visit (type);
+	return;
+      }
+
+    // we can allow for mutability changes here by casting down from mutability
+    // eg:  mut vs const, we cant take a mutable reference from a const
+    // eg:  const vs mut we can take a const reference from a mutable one
```
bjorn3, line 1164: In this comment: s/reference/raw pointer

gcc/testsuite/rust/compile/torture/coercion2.rs
philberty: Ah ok thanks for lets not lose this comment as part of my work into casts

gcc/rust/typecheck/rust-tyty-coercion.h
philberty: fixed

PR #600, opened 2021-07-30
Use correct lookahead to peek after 'while' token when parsing expression with block (parse_expr_with_block).

asakovets:


SimplyTheOther (approved):
PR looks great. 

Based on the code context, the original code was actually incorrect and this change shouldn't cause any regressions.

CohenArthur (approved):

PR #601, opened 2021-08-01
union support for hir type checking and gcc backend

### LINKED ISSUE #157: Unions support
https://doc.rust-lang.org/reference/items/unions.html

dkm:
From Mark Wielaard : https://gcc.gnu.org/pipermail/gcc-rust/2021-August/000107.html

> 
> Treat a union as a Struct variant like a tuple struct.  Add an
> iterator and get_identifier functions to the AST Union class.  Same
> for the HIR Union class, plus a get_generics_params method. Add a new
> ADTKind enum and adt_kind field to the ADTType to select the
> underlying abstract data type (struct struct, tuple struct or union,
> with enum as possible future variant).
> 
> An union constructor can have only one field. Add an union_index field
> to StructExprStruct which is set during type checking in the
> TypeCheckStructExpr HIR StructExprStructFields visitor.
> 
> For the Gcc_backend class rename fill_in_struct to fill_in_fields and
> use it from a new union_type method. Handle union_index in
> constructor_expression (so only one field is initialized).

Fixes #157

### DIFF
--- a/gcc/rust/typecheck/rust-tyty.h
+++ b/gcc/rust/typecheck/rust-tyty.h
@@ -855,7 +855,15 @@ protected:
 class ADTType : public BaseType, public SubstitutionRef
 {
 public:
-  ADTType (HirId ref, std::string identifier, bool is_tuple,
+  enum ADTKind
+  {
+    STRUCT_STRUCT,
+    TUPLE_STRUCT,
+    UNION,
+    // ENUM ?
+  };
+
+  ADTType (HirId ref, std::string identifier, ADTKind adt_kind,
 	   std::vector<StructFieldType *> fields,
 	   std::vector<SubstitutionParamMapping> subst_refs,
 	   SubstitutionArgumentMappings generic_arguments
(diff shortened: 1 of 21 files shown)


gcc/testsuite/rust/compile/torture/union.rs
```diff
@@ -0,0 +1,35 @@
+// { dg-do compile }
+// { dg-options "-w" }
```
philberty, line 2: what does -w do here i wonder

gcc/rust/typecheck/rust-tyty.h
```diff
@@ -855,29 +855,39 @@ class SubstitutionRef
 class ADTType : public BaseType, public SubstitutionRef
 {
 public:
-  ADTType (HirId ref, std::string identifier, bool is_tuple,
+  enum ADTKind
```
philberty, line 858: brilliant i forget c++ enum would be better than the rustc bitmask idea

gcc/testsuite/rust/compile/torture/union.rs
dkm: `       -w  Inhibit all warning messages.`
I guess it's a trick to avoid the test from failing on warnings

tschwinge: Are there really so many diagnostics to make it infeasible to spell these out individually?  The problem that I have with `-w` and similar "hammers" is that also you won't notice any bogus diagnostics appearing in the future.

philberty (approved):
LGMT great work!

gcc/testsuite/rust/compile/torture/union.rs
philberty: Yeah thats a good point I think @thomasyonug is fixing up our unused variables at the moment which will let us remove all of the duplicate warnings we have at the moment. Rustc as far as I know controls warnings via the attribute system but I am wondering if a flag system might also prove useful.

tschwinge: Aha, per <http://mid.mail-archive.com/YQciMeKNpCH+ZMsJ@wildebeest.org> the `-w` here is just an interim thing.

dkm: May I should add this new commit here then...

dkm:
I'll open a new PR for the newest patch
bors r+

philberty:
> I'll open a new PR for the newest patch
> bors r+

Thanks, @dkm  that really helps me a lot this morning.

dkm:
Looks like GHA has some connectivity issue... Let's try again?

bors r+

PR #602, opened 2021-08-02
Add locus to TupleField and pass it and union variants to HIR class

dkm:
From Mark Wielaard : https://gcc.gnu.org/pipermail/gcc-rust/2021-August/000108.html

> TupleField was missing a Location field and we dropped to locus when
> lowering Union fields to HIR.


philberty (approved):
Brilliant, there are a few cases of bad or missing location info.

I think the end closing scope of function item has bad location info off the top of my head. We should make more issues like this. I think there are plenty of cases like this.

PR #604, opened 2021-08-02
Handle 'UnsafeBlockExpr' in liveness analysis

tschwinge:
We may then remove the '{ dg-options "-w" }' as discussed in
<http://mid.mail-archive.com/YQciMeKNpCH+ZMsJ@wildebeest.org> and #601.

### DIFF
--- a/gcc/rust/lint/rust-lint-marklive.h
+++ b/gcc/rust/lint/rust-lint-marklive.h
@@ -107,6 +107,11 @@ public:
       }
   }
 
+  void visit (HIR::UnsafeBlockExpr &expr) override
+  {
+    expr.get_block_expr ()->accept_vis (*this);
+  }
+
   void visit (HIR::LoopExpr &expr) override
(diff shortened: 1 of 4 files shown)


thomasyonug (approved):
LGTM. :)

gcc/rust/lint/rust-lint-marklive.h
```diff
@@ -107,6 +107,11 @@ class MarkLive : public MarkLiveBase
       }
   }
 
+  void visit (HIR::UnsafeBlockExpr &expr) override
+  {
+    expr.get_block_expr ()->accept_vis (*this);
+  }
+
```
thomasyonug, line 114: It will be more convenient there is an auto walk visit class for HIR in order to avoid writing template walking code. I'll work on this.

philberty:
Nice work LGTM

philberty (approved):

PR #605, opened 2021-08-03
Merge both module classes in one

CohenArthur:
This PR merges both kinds of Modules (formerly `ModuleBodied` and `ModuleNoBody`) as one class with an enum. This is the [behavior used by rustc](https://github.com/rust-lang/rust/blob/2939249f294dd54a9ce78a8ee1f2922a44e7fb7c/compiler/rustc_ast/src/ast.rs#L2274), where both variants are kept in an enum with one holding a vector of items.

This change is important for multiple file parsing: An external mod (`mod foo; // defined in foo.rs or foo/mod.rs`) will see its items expanded during expansion, which occurs after parsing. This means that the previous directive will be "replaced" by `mod foo { <items> }` at the AST level. In order to achieve this, we need to be able to modify a previously parsed instance of an AST element.

In rustc, this is done [here](https://github.com/rust-lang/rust/blob/2939249f294dd54a9ce78a8ee1f2922a44e7fb7c/compiler/rustc_expand/src/expand.rs#L1427), where `mod_kind` was previously `ModKind::Unloaded` and becomes `ModKind::Loaded(parsed_items, ...)`.

### DIFF
--- a/gcc/rust/ast/rust-ast-full-test.cc
+++ b/gcc/rust/ast/rust-ast-full-test.cc
@@ -379,22 +379,20 @@ VisItem::as_string () const
 std::string
 Module::as_string () const
 {
-  std::string vis_item = VisItem::as_string ();
+  std::string str = VisItem::as_string () + "mod " + module_name;
 
-  return vis_item + "mod " + module_name;
-}
-
-std::string
-ModuleBodied::as_string () const
-{
-  // get module string for "[vis] mod [name]"
-  std::string str = Module::as_string ();
+  // Return early if we're dealing with an unloaded module as their body resides
+  // in a different file
+  if (kind == ModuleKind::UNLOADED)
+    return str + "\n no body (reference to external file)\n";
 
   // inner attributes
   str += append_attributes (inner_attrs, INNER);
 
   // items
   str += "\n items: ";
+
+  // This can still happen if the module is loaded but empty, i.e. `mod foo {}`
   if (items.empty ())
     {
       str += "none";
@@ -414,16 +412,6 @@ ModuleBodied::as_string () const
 	  str += "\n  " + item->as_string ();
 	}
     }
-
-  return str + "\n";
-}
-
-std::string
-ModuleNoBody::as_string () const
-{
-  std::string str = Module::as_string ();
-
-  str += "\n no body (reference to external file)";
 
   return str + "\n";
 }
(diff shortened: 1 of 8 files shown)


gcc/rust/ast/rust-ast-full-test.cc
```diff
@@ -379,22 +379,20 @@ VisItem::as_string () const
...
+    return str + "\n no body (reference to external file)\n";
 
   // inner attributes
   str += append_attributes (inner_attrs, INNER);
 
   // items
   str += "\n items: ";
+
+  // This can still happen if the module is loaded but empty, i.e. `mod foo {}`
   if (items.empty ())
```
philberty, line 396: It might be worth emitting a "not loaded" instead of empty for that case

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -2118,10 +2118,10 @@ Parser<ManagedTokenSource>::parse_module (AST::Visibility vis,
     case SEMICOLON:
       lexer.skip_token ();
 
-      return std::unique_ptr<AST::ModuleNoBody> (
-	new AST::ModuleNoBody (std::move (name), std::move (vis),
-			       std::move (outer_attrs),
-			       locus)); // module name?
+      return std::unique_ptr<AST::Module> (
```
philberty, line 2121: That was really neat :D

gcc/rust/ast/rust-item.h
```diff
@@ -57,7 +57,7 @@ class TypeParam : public GenericParam
 
   // bool has_type_param_bounds;
   // TypeParamBounds type_param_bounds;
-  std::vector<std::unique_ptr<TypeParamBound> >
+  std::vector<std::unique_ptr<TypeParamBound>>
```
philberty, line 60: which version of clang-format are you using?

CohenArthur: I downloaded a clang-format 10 binary that I am using specially for gccrs. I got it from the clang-format action that the CI uses. There are a lot of differences when running clang-format 11, which is what I have installed on my machine usually. Is that not the right version?

gcc/rust/ast/rust-ast-full-test.cc
CohenArthur: Yeah I was wondering about that, but since the distinction is made at parsing based on whether the next token is a curly brace or a semicolon, I don't know if it's possible. Unless I'm misunderstanding your comment?

SimplyTheOther (approved):
Seems great (apart from maybe the clang-format formatting issues).

CohenArthur:
> Seems great (apart from maybe the clang-format formatting issues).

What baffles me is that the clang-format CI still passes haha. I noticed those slight changes but thought nothing of them

philberty (approved):

philberty:
That's interesting with the clang-format here. I don't think it will block us from merging though.

Issue #603, opened 2021-08-02
ICE with 'unsafe' in 'impl'


tschwinge:
Given:

```rust
impl I {
    unsafe {
    }
}
```

`rust` diagnoses:

```
error: non-item in item list
 --> [...]/a.rs:2:5
  |
1 | impl I {
  |        - item list starts here
2 |     unsafe {
  |     ^^^^^^ non-item starts here
3 |     }
4 | }
  | - item list ends here
[...]
```

GCC/Rust Git commit b56c6fdfaad9ca1681714d288d1282cf08554462 runs into a SIGSEGV in `gcc/rust/parse/rust-parse-impl.h`:

```
   5326 template <typename ManagedTokenSource>
   5327 std::unique_ptr<AST::InherentImplItem>
   5328 Parser<ManagedTokenSource>::parse_inherent_impl_function_or_method (
   5329   AST::Visibility vis, AST::AttrVec outer_attrs)
   5330 {
   5331   Location locus = lexer.peek_token ()->get_locus ();
   5332   // parse function or method qualifiers
   5333   AST::FunctionQualifiers qualifiers = parse_function_qualifiers ();
   5334 
   5335   skip_token (FN_TOK);
   5336 
   5337   // parse function or method name
   5338   const_TokenPtr ident_tok = expect_token (IDENTIFIER);
   5339   Identifier ident = ident_tok->get_str ();
```

... because `ident_tok` is `NULL`.

PR #606, opened 2021-08-04
Always check the result of expect_token while parsing

dkm:
From Mark Wielaard : https://gcc.gnu.org/pipermail/gcc-rust/2021-August/000111.html

> When expect_token fails it produces an error and return a
> nullptr. Make sure to always check the result of expect_token so we
> don't use a nullptr token and crash.
> 
> Resolves: https://github.com/Rust-GCC/gccrs/issues/603


CohenArthur (approved):

tschwinge:
Should include test case?

SimplyTheOther (approved):

tschwinge:
I've added a more elaborate variant of my #603 test case, others still missing (if we want to add more).

philberty (approved):

dkm:
bors: merge

PR #608, opened 2021-08-05
Add `parse_items()` function from `parse_crate()`

CohenArthur:
In order to support multiple file parsing, we will need to be able to parse items from a file without specifically parsing an entire crate. This PR splits up the `parse_crate()` function in two and exposes the `parse_items()` function.

The function was previously hidden and didn't have the exact same behavior as the main loop of `parse_crate()`.

This will be especially useful when trying to expand 
```rust
mod <file>;
```
to
```rust
mod <file> {
    <items>
}
```
in the AST as we will need to parse the items present in the file.

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -388,15 +388,12 @@ Parser<ManagedTokenSource>::done_end_of_file ()
   return lexer.peek_token ()->get_id () == END_OF_FILE;
 }
 
-// Parses a crate (compilation unit) - entry point
+// Parses a sequence of items within a module or the implicit top-level module
+// in a crate
 template <typename ManagedTokenSource>
-AST::Crate
-Parser<ManagedTokenSource>::parse_crate ()
+std::vector<std::unique_ptr<AST::Item>>
+Parser<ManagedTokenSource>::parse_items ()
 {
-  // parse inner attributes
-  AST::AttrVec inner_attrs = parse_inner_attributes ();
-
-  // parse items
   std::vector<std::unique_ptr<AST::Item>> items;
 
   const_TokenPtr t = lexer.peek_token ();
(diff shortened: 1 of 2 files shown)


gcc/rust/parse/rust-parse-impl.h
```diff
@@ -388,15 +388,12 @@ Parser<ManagedTokenSource>::done_end_of_file ()
...
 }
 
-// Parses a crate (compilation unit) - entry point
+// Parses a sequence of items within a module or the implicit top-level module
+// in a crate
 template <typename ManagedTokenSource>
-AST::Crate
-Parser<ManagedTokenSource>::parse_crate ()
+std::vector<std::unique_ptr<AST::Item>>
+Parser<ManagedTokenSource>::parse_items (std::vector<Error> &error_table)
```
philberty, line 395: I think this can be simpler atfter our discussion on zulip. Lets change this to not take the error_table which should simplify things

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -407,7 +404,7 @@ Parser<ManagedTokenSource>::parse_crate ()
 	{
 	  Error error (lexer.peek_token ()->get_locus (),
 		       "failed to parse item in crate");
-	  add_error (std::move (error));
+	  error_table.push_back (std::move (error));
```
philberty, line 407: This can remain as add_error

gcc/rust/parse/rust-parse.h
```diff
@@ -585,6 +584,12 @@ template <typename ManagedTokenSource> class Parser
   // Construct parser with specified "managed" token source.
   Parser (ManagedTokenSource tokenSource) : lexer (std::move (tokenSource)) {}
 
+  // Parse items without parsing an entire crate. This function is the main
+  // parsing loop of AST::Crate::parse_crate(). It takes a mutable reference
+  // on a vector of errors as parameter in which errors will be appended.
+  std::vector<std::unique_ptr<AST::Item> >
+  parse_items (std::vector<Error> &error_table);
+
```
philberty, line 592: Add a new getter unless one already exists.

std::vector<Error> get_errors()

philberty (changes requested):

gcc/rust/parse/rust-parse.h
CohenArthur: Done in b9f4f1b

philberty (approved):

PR #610, opened 2021-08-06
More rustspec.cc cleanups

dkm:
From Mark Wielaard : https://gcc.gnu.org/pipermail/gcc-rust/2021-August/000117.html

> rustspec.cc was based on the Go frontend gospec.cc. Remove special
> handling of math and pthread libraries and profiling option. Handle
> .rs files instead of .go files. Keep support for linking with (static)
> librust which is currently commented out. Add generic static-librust
> option to common.opt.

### DIFF
--- a/gcc/rust/rustspec.cc
+++ b/gcc/rust/rustspec.cc
@@ -1,4 +1,4 @@
-/* rustspec.c -- Specific flags and argument handling of the gcc Go front end.
+/* rustspec.c -- Specific flags and argument handling of the gcc Rust front end.
    Copyright (C) 2009-2020 Free Software Foundation, Inc.
 
 This file is part of GCC.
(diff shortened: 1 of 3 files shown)


philberty (approved):

dkm:
bors: merge

dkm:
bors: cancel?

CohenArthur (approved):

gcc/rust/rustspec.cc
```diff
@@ -28,24 +28,10 @@ along with GCC; see the file COPYING3.  If not see
 
 /* This bit is set if we saw a `-xfoo' language specification.  */
 #define LANGSPEC (1 << 1)
-/* This bit is set if they did `-lm' or `-lmath'.  */
-#define MATHLIB (1 << 2)
-/* This bit is set if they did `-lpthread'.  */
-#define THREADLIB (1 << 3)
 /* This bit is set if they did `-lc'.  */
-#define WITHLIBC (1 << 4)
+#define WITHLIBC (1 << 2)
```
bjorn3, line 32: Is this necessary? Users of rustc normally don't use libc directly, but link to libc because of the `#[link(name = "c")]` inside the libc bindings rust crate (`liblibc.rlib`), which this bit won't indicate.

philberty: This mask is used to detect if the compiler was invoked with -lc so there is a variable flag saw_libc which can do some more magic with the linking options. Good catch, I had to go double-check the code myself after this comment.

I think this serves as a great start to strip the driver back which will help a lot in the future.

CohenArthur: I think a lot of this file is not compatible with how rust is expected to work in terms of linking, but that's fine because the ways used by rustc are not usable yet. For example, we shouldn't link with the standard library if a file contains `#![no_std]`. And as bjorn said, we should link with C libraries if they are mentioned in a [`link` attribute](https://doc.rust-lang.org/reference/items/external-blocks.html#the-link-attribute), which will again happen in source code and not from the compiler's invocation level. So I think a lot of the "link" options in that file will need to be reworked anyway once procedural macros are supported

CohenArthur: I mean specifically regarding to linking libc and librust, as the rustc compiler can also be invoked with `-W,l...` for linking libraries of course. But those aren't treated "differently" as those two in the rust-spec

Issue #440, opened 2021-05-21
Trait Constraints


philberty:
In order for the type system to handle traits and inference variables, we need to have the notion of trait constraints or obligations.

philberty:
this is a dup of #442 since PR #583 we have preliminary support for his now.

PR #611, opened 2021-08-06
Add check for TypeBounds on argument substitutions

philberty:
TypeParameters contain specified bounds, when we substitute them we need
to ensure that the argument satisfies those bounds otherwise... it fails
the contract.

Addreses #440

### DIFF
--- a/gcc/rust/typecheck/rust-tyty.cc
+++ b/gcc/rust/typecheck/rust-tyty.cc
@@ -49,11 +49,11 @@ BaseType::satisfies_bound (const TypeBoundPredicate &predicate) const
 	return true;
     }
 
-  std::vector<std::reference_wrapper<Resolver::TraitReference>> probed
+  std::vector<Resolver::TraitReference *> probed
     = Resolver::TypeBoundsProbe::Probe (this);
-  for (const Resolver::TraitReference &bound : probed)
+  for (const Resolver::TraitReference *bound : probed)
     {
-      bool found = bound.get_mappings ().get_defid ()
+      bool found = bound->get_mappings ().get_defid ()
 		   == query->get_mappings ().get_defid ();
       if (found)
 	return true;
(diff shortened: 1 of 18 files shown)


gcc/rust/typecheck/rust-tyty.cc
```diff
@@ -27,13 +27,41 @@
...
+
+  std::vector<std::reference_wrapper<Resolver::TraitReference>> probed
+    = Resolver::TypeBoundsProbe::Probe (this);
+  for (const Resolver::TraitReference &bound : probed)
+    {
+      bool found = bound.get_mappings ().get_defid ()
+		   == query->get_mappings ().get_defid ();
+      if (found)
+	return true;
+    }
```
CohenArthur, line 60: I'm not sure I'm understanding this code correctly. Does this mean that as soon as you find a matching bound, the function returns true? In that case, is it intended (for now) that multiple bounds are not verified correctly?

For example, the following example compiles succesfully:

```rust
trait Foo {
    fn default() -> i32;
}

trait Bar {
    fn not_default() -> i32;
}

struct Test(i32);

impl Foo for Test {
    fn default() -> i32 {
        1234
    }
}

fn type_bound_test<T: Foo + Bar>() -> i32 { // <T: Bar + Foo> also compiles
    T::default()
}

fn main() {
    let a = type_bound_test::<Test>();
}
```

philberty: Let me add that test case to verify this.

But the code is split up into 2 parts this function is about checking each bound the driver of this is actually here: 

https://github.com/Rust-GCC/gccrs/pull/611/files#diff-03fbb543dd5e74fc2332fa2e5c9c9a91ed6d46d86e7bf0310a8796fae439d26cR256-R275

So we iterate the bounds of the TypePaameter then verify that each of these are satisfied by the substitution.

CohenArthur: Ah, right thanks. I see how the code works exactly now. Glad that multiple bounds are working :D Good job!

philberty:
bors cancel

philberty:
bors merge

PR #613, opened 2021-08-07
Resolve module filename

CohenArthur:
This PR is a first attempt at resolving the filename corresponding to an external module correctly. Some of the cases are not handled yet and a lot of FIXMEs are still here, as I am looking for feedback on certain things:

* Am I correct in assuming that we have to go through directories the C way because we are using C++11 and the `filesystem` header is not available until C++17? Is there some gcc abstraction for this that I'm overlooking?
* How important is the existence of a separate SEPARATOR macro for Windows? From what I'm understanding, the OS also understands normal slashes `/` on top of the backward slashes it usually uses `\`. I don't know what happens when they are mixed and matched or how the file system handles it.
* For review simplicity, outer attributes could be accessed in a later PR. I believe they can already be accessed and looked at but haven't looked into it. I'm also unsure if this would be the right place to implement that outer_attr lookup

### DIFF
--- a/gcc/rust/ast/rust-ast-full-test.cc
+++ b/gcc/rust/ast/rust-ast-full-test.cc
@@ -17,6 +17,9 @@ You should have received a copy of the GNU General Public License
 along with GCC; see the file COPYING3.  If not see
 <http://www.gnu.org/licenses/>.  */
 
+// FIXME: This does not work on Windows
+#include <unistd.h>
+
 #include "rust-ast-full.h"
 #include "rust-diagnostics.h"
 #include "rust-ast-visitor.h"
@@ -4048,6 +4051,73 @@ Module::add_crate_name (std::vector<std::string> &names) const
 
   for (const auto &item : items)
     item->add_crate_name (names);
+}
+
+static bool
+file_exists (const std::string path)
+{
+  // Simply check if the file exists
+  // FIXME: This does not work on Windows
+  return access (path.c_str (), F_OK) != -1;
+}
+
+// FIXME: This function should also check if the module has a `path` outer
+// attribute and fetch the path from here in that case, i.e:
+// ```
+// #[path="<dir>/<subdir>/<file>.rs"]
+// mod <mod_name>;
+// ```
+std::string
+Module::get_filename ()
+{
+  rust_assert (kind == Module::ModuleKind::UNLOADED);
+
+  // This corresponds to the path of the file 'including' the module. So the
+  // file that contains the 'mod <file>;' directive
+  std::string including_fname (outer_filename);
+
+  std::string expected_file_path = module_name + ".rs";
+  std::string expected_dir_path = "mod.rs";
+
+  auto dir_slash_pos = including_fname.rfind (file_separator);
+  std::string current_directory_name;
+
+  // If we haven't found a file_separator, then we have to look for files in the
+  // current directory ('.')
+  if (dir_slash_pos == std::string::npos)
+    current_directory_name = std::string (".") + file_separator;
+  else
+    current_directory_name
+      = including_fname.substr (0, dir_slash_pos) + file_separator;
+
+  // FIXME: We also have to search for
+  // <directory>/<including_fname>/<module_name>.rs In rustc, this is done via
+  // the concept of `DirOwnernship`, which is based on whether or not the
+  // current file is titled `mod.rs`.
+
+  // First, we search for <directory>/<module_name>.rs
+  bool file_mod_found
+    = file_exists (current_directory_name + expected_file_path);
+
+  // Then, search for <directory>/<module_name>/mod.rs
+  current_directory_name += module_name + file_separator;
+  bool dir_mod_found = file_exists (current_directory_name + expected_dir_path);
+
+  bool multiple_candidates_found = file_mod_found && dir_mod_found;
+  bool no_candidates_found = !file_mod_found && !dir_mod_found;
+
+  if (multiple_candidates_found)
+    rust_error_at (locus,
+		   "two candidates found for module %s: %s.rs and %s%smod.rs",
+		   module_name.c_str (), module_name.c_str (),
+		   module_name.c_str (), file_separator);
+
+  if (no_candidates_found)
+    rust_error_at (locus, "no candidate found for module %s",
+		   module_name.c_str ());
+
+  return file_mod_found ? expected_file_path
+			: current_directory_name + expected_dir_path;
 }
 
 void
--- a/gcc/rust/ast/rust-item.h
+++ b/gcc/rust/ast/rust-item.h
@@ -1053,6 +1053,9 @@ public:
     return *this;
   }
 
+  // Search for the filename associated with an external module
+  std::string get_filename ();
+
   void accept_vis (ASTVisitor &vis) override;
 
   /* Override that runs the function recursively on all items contained within
--- a/gcc/rust/ast/rust-item.h
+++ b/gcc/rust/ast/rust-item.h
@@ -977,6 +977,8 @@ private:
   Location locus;
   ModuleKind kind;
 
+  // Name of the file including the module
+  std::string outer_filename;
   // bool has_inner_attrs;
   std::vector<Attribute> inner_attrs;
   // bool has_items;
@@ -998,10 +1000,11 @@ public:
 
   // Unloaded module constructor
   Module (Identifier module_name, Visibility visibility,
-	  std::vector<Attribute> outer_attrs, Location locus)
+	  std::vector<Attribute> outer_attrs, Location locus,
+	  std::string outer_filename)
     : VisItem (std::move (visibility), std::move (outer_attrs)),
       module_name (module_name), locus (locus), kind (ModuleKind::UNLOADED),
-      inner_attrs (std::vector<Attribute> ()),
+      outer_filename (outer_filename), inner_attrs (std::vector<Attribute> ()),
       items (std::vector<std::unique_ptr<Item>> ())
   {}
 
@@ -1013,7 +1016,8 @@ public:
 	  std::vector<Attribute> outer_attrs = std::vector<Attribute> ())
     : VisItem (std::move (visibility), std::move (outer_attrs)),
       module_name (name), locus (locus), kind (ModuleKind::LOADED),
-      inner_attrs (std::move (inner_attrs)), items (std::move (items))
+      outer_filename (std::string ()), inner_attrs (std::move (inner_attrs)),
+      items (std::move (items))
   {}
 
   // Copy constructor with vector clone
--- a/gcc/rust/expand/rust-macro-expand.cc
+++ b/gcc/rust/expand/rust-macro-expand.cc
@@ -2036,6 +2036,10 @@ public:
 	    return;
 	  }
       }
+    else
+      {
+	std::string mod_file = module.get_filename ();
+      }
 
     // strip items if required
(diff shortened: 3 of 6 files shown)


bjorn3:
> How important is the existence of a separate SEPARATOR macro for Windows? From what I'm understanding, the OS also understands normal slashes / on top of the backward slashes it usually uses \. I don't know what happens when they are mixed and matched or how the file system handles it.

`/` is only accepted when the path doesn't start with `\\?\`. In that case the win32 api dll will normalize the path for you. To allow paths longer than `MAX_PATH` (255 characters I believe), you need a path that starts with `\\?\`. This causes the win32 api dll to pass the path to the kernel without any normalization. The kernel doesn't accept `/`.

gcc/rust/ast/rust-ast-full-test.cc
```diff
@@ -4050,6 +4052,88 @@ Module::add_crate_name (std::vector<std::string> &names) const
...
+// #[path="<dir>/<subdir>/<file>.rs"]
+// mod <mod_name>;
+// ```
+std::string
+Module::get_filename ()
+{
+// TODO: Not windows compliant, at least not always. We need to have some
+// sort of #ifdef but where should it be put? Since this could also be used
+// elsewhere in the compiler
+#define SEPARATOR "/"
```
philberty, line 4087: I think a static constant might be more C++'y

gcc/rust/ast/rust-ast-full-test.cc
```diff
@@ -4050,6 +4052,88 @@ Module::add_crate_name (std::vector<std::string> &names) const
     item->add_crate_name (names);
 }
 
+static bool
+check_for_file (std::string dir_name, std::string file_name)
```
philberty, line 4056: I think you want a static bool file_exists(const std::string path) 

this function is a bit more reuseable and it will be able to just look at a path and if it exists then check to make sure it is infact a file. It avoids having to do readdir and check if paths are in fact directories etc.

philberty:
I think this PR is a good start but I would suggest taking a look at this:

https://github.com/Rust-GCC/gccrs/blob/master/libiberty/filename_cmp.c
https://github.com/Rust-GCC/gccrs/blob/master/libiberty/lrealpath.c

You should be able to include the relevant header to get access to libiberty and reuse these functions for path stuff. It might simplify things for you to avoid windows issues etc.

gcc/rust/ast/rust-ast-full-test.cc
CohenArthur: Done in e0db5ce

gcc/rust/ast/rust-ast-full-test.cc
CohenArthur: Done (I think? Not too sure about `constexpr static const`) in e484f42

CohenArthur:
> I think this PR is a good start but I would suggest taking a look at this:
> 
> https://github.com/Rust-GCC/gccrs/blob/master/libiberty/filename_cmp.c
> https://github.com/Rust-GCC/gccrs/blob/master/libiberty/lrealpath.c
> 
> You should be able to include the relevant header to get access to libiberty and reuse these functions for path stuff. It might simplify things for you to avoid windows issues etc.

So, I haven't found any function regarding the existence of a file but I might be missing something. I'm using the `access` function right now, which isn't part of Windows I believe. There seems to be an equivalent with `_access` but not sure. Maybe I haven't looked hard enough in libiberty?

gcc/rust/ast/rust-ast-full-test.cc
```diff
@@ -4050,6 +4053,82 @@ Module::add_crate_name (std::vector<std::string> &names) const
...
+{
+  // Simply check if the file exists
+  // FIXME: This does not work on Windows
+  return access (path.c_str (), F_OK) != -1;
+}
+
+// FIXME: This function should also check if the module has a `path` outer
+// attribute and fetch the path from here in that case, i.e:
+// ```
+// #[path="<dir>/<subdir>/<file>.rs"]
```
philberty, line 4067: This is a good comment. Lets not fix this within this PR.

But can you investigate how rustc does this? I am interested to know if rustc turns all attributes like this into specific classes or if there are some kind of enum on attributes or something like that to work with them.

gcc/rust/ast/rust-ast-full-test.cc
```diff
@@ -4050,6 +4053,82 @@ Module::add_crate_name (std::vector<std::string> &names) const
...
+  bool dir_mod_found = file_exists (current_directory_name + expected_dir_path);
+
+  std::string final_candidate;
+
+  if (file_mod_found)
+    final_candidate = expected_file_path;
+  if (dir_mod_found)
+    final_candidate = current_directory_name + expected_dir_path;
+
+  if (file_mod_found && dir_mod_found)
```
philberty, line 4113: do this check first before setting the final_candidate

gcc/rust/ast/rust-ast-full-test.cc
```diff
@@ -4050,6 +4053,82 @@ Module::add_crate_name (std::vector<std::string> &names) const
...
+    {
+      rust_error_at (locus,
+		     "two candidates found for module %s: %s.rs and %s%smod.rs",
+		     module_name.c_str (), module_name.c_str (),
+		     module_name.c_str (), separator);
+      final_candidate = "";
+    }
+
+  if (!file_mod_found && !dir_mod_found)
```
philberty, line 4122: sam do this before the final candidate

gcc/rust/ast/rust-ast-full-test.cc
philberty: I would also be tempted to give this check a name like:

bool multiple_candidates_found == file_mod_found && dir_mod_found;
bool no_candidates_found = !multiple_candidates_found;

gcc/rust/ast/rust-ast-full-test.cc
```diff
@@ -4050,6 +4053,82 @@ Module::add_crate_name (std::vector<std::string> &names) const
...
+
+  // Then, search for <directory>/<module_name>/mod.rs
+  current_directory_name += module_name + separator;
+  bool dir_mod_found = file_exists (current_directory_name + expected_dir_path);
+
+  std::string final_candidate;
+
+  if (file_mod_found)
+    final_candidate = expected_file_path;
+  if (dir_mod_found)
```
philberty, line 4110: shouldn't this be an if/else?

gcc/rust/ast/rust-item.h
```diff
@@ -977,6 +977,15 @@ class Module : public VisItem
   Location locus;
   ModuleKind kind;
 
+#if defined(HAVE_DOS_BASED_FILE_SYSTEM)
+  constexpr static const char *separator = "\\";
```
philberty, line 981: this might make more sense in rust-system.h?

gcc/rust/ast/rust-item.h
```diff
@@ -977,6 +977,15 @@ class Module : public VisItem
...
+#else
+  constexpr static const char *separator = "/";
+#endif /* HAVE_DOS_BASED_FILE_SYSTEM */
+
+  // Name of the file including the module
+  const char *outer_filename;
```
philberty, line 987: could this be a c++ string?

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -2101,10 +2101,11 @@ Parser<ManagedTokenSource>::parse_module (AST::Visibility vis,
     case SEMICOLON:
       lexer.skip_token ();
 
+      // Construct an external module
       return std::unique_ptr<AST::Module> (
 	new AST::Module (std::move (name), std::move (vis),
-			 std::move (outer_attrs),
-			 locus)); // module name?
+			 std::move (outer_attrs), locus,
+			 lexer.get_filename ())); // module name?
```
philberty, line 2108: i think you can remove that module name comment

gcc/rust/lex/rust-lex.h
```diff
@@ -136,8 +136,11 @@ class Lexer
   void split_current_token (TokenId new_left, TokenId new_right);
 
   Linemap *get_line_map () { return line_map; }
+  const char *get_filename () { return filename; }
 
 private:
+  // Original filename used to instantiate the lexer
+  const char *filename;
```
philberty, line 143: can this be a proxy to the RAIIFile object? It would be best to decouple this and get it from the file object.

philberty (changes requested):
Looking pretty good so far this is all style stuff. Nice work.

gcc/rust/ast/rust-ast-full-test.cc
CohenArthur: Definitely! I'd love to work on this

gcc/rust/lex/rust-lex.h
CohenArthur: In the end we still have to keep a `const char *` somewhere because we cannot get the filename from the `FILE` pointer kept in RAIIFile. But I agree that it makes more sense to keep it there instead of the lexer. Done in 06e6a4a25c8

gcc/rust/ast/rust-item.h
CohenArthur: Done in 06e6a4a

gcc/rust/expand/rust-macro-expand.cc
```diff
@@ -2036,6 +2036,10 @@ class AttrVisitor : public AST::ASTVisitor
 	    return;
 	  }
       }
+    else
+      {
+	std::string mod_file = module.get_filename ();
```
philberty, line 2041: what does this line do?

philberty (changes requested):
This looks good, but just let me know what this line does and I can approve this.

gcc/rust/ast/rust-ast-full-test.cc
```diff
@@ -4050,6 +4053,79 @@ Module::add_crate_name (std::vector<std::string> &names) const
...
+    rust_error_at (locus,
+		   "two candidates found for module %s: %s.rs and %s%smod.rs",
+		   module_name.c_str (), module_name.c_str (),
+		   module_name.c_str (), file_separator);
+
+  if (no_candidates_found)
+    rust_error_at (locus, "no candidate found for module %s",
+		   module_name.c_str ());
+
+  std::string final_candidate = "";
```
philberty, line 4119: this could be a ternary statement like:

return file_mod_found ? expected_file_path : current_directory_name + expected_dir_path;

philberty:
All other items seem to be resolved now

gcc/rust/expand/rust-macro-expand.cc
CohenArthur: Ah, this is just the first step of items loading. I added it so that the function wouldn't be unused but I can remove it for now. Basically, the flow is going to be like this:

```c++
std::string mod_file = module.get_filename ();

if (mod_file.is_empty ()) return;

RAIIFile file_wrap(mod_file);
Linemap *linemap = /* magic */;

Lexer lex(mod_file.c_str (), std::move (file_wrap), linemap);
Parser<Lexer> parser (std::move (lex));

auto items = parser.parse_items ();

this.set_items (items);
```

gcc/rust/ast/rust-ast-full-test.cc
CohenArthur: Done in a1560bd

philberty (approved):
LGTM I think you could rebase the commits a little bit into 2 commits one for lexer changes and one for module work.

PR #615, opened 2021-08-08
lex: accept zero codepoints in strings

dkm:
From Mark Wielaard : https://gcc.gnu.org/pipermail/gcc-rust/2021-August/000127.html

> 
> Zero characters (codepoints) are acceptable in strings. The current
> Lexer::parse_string skipped such zero codepoints by accidents. The
> zero codepoint was also used as error/skip indicator, but that is only
> true if the third argument of utf8_escape_pair is true (yes, it is
> called pair, but is a triple).
> 
> Add a testcase that checks the (sub)strings are separated by zero
> chars. Since we cannot slice strings yet this uses extern "C"
> functions, printf and memchr.

### DIFF
--- a/gcc/rust/lex/rust-lex.cc
+++ b/gcc/rust/lex/rust-lex.cc
@@ -1827,7 +1827,8 @@ Lexer::parse_string (Location loc)
 	  else
 	    length += std::get<1> (utf8_escape_pair);
 
-	  if (current_char32 != Codepoint (0))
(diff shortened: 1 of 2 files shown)


CohenArthur (approved):

philberty (approved):

philberty:
bors cancel

philberty:
bors merge

PR #616, opened 2021-08-08
Pass pratt parsed token to expr parser functions to fix expr locus

philberty:
Before gccrs would generate the following error message:

return.rs:3:22: error: cannot ‘break’ outside of a loop
    3 |     let x = 5 - break return (16 + 2);
      |                      ^~~~~~~~~~~~~~~~~

Now we get:

return.rs:3:17: error: cannot ‘break’ outside of a loop
    3 |     let x = 5 - break return (16 + 2);
      |                 ^

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -6586,7 +6586,7 @@ Parser<ManagedTokenSource>::parse_path_expr_segment ()
 template <typename ManagedTokenSource>
 AST::QualifiedPathInExpression
 Parser<ManagedTokenSource>::parse_qualified_path_in_expression (
-  bool pratt_parse)
+  const_TokenPtr pratt_parsed_token)
 {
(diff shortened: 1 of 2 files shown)


gcc/rust/parse/rust-parse-impl.h
```diff
@@ -6586,7 +6586,7 @@ Parser<ManagedTokenSource>::parse_path_expr_segment ()
 template <typename ManagedTokenSource>
 AST::QualifiedPathInExpression
 Parser<ManagedTokenSource>::parse_qualified_path_in_expression (
-  bool pratt_parse)
+  const_TokenPtr pratt_parsed_token)
```
CohenArthur, line 6589: @philberty I thought you said you wanted to merge the patch containing the Location and not the full Token?

philberty: Good point I got a bit confused when I raised the PR to review the patch. I will ask on the mailing list again.

PR #619, opened 2021-08-10
Rename CanonicalPath::is_error to is_empty

dkm:
`is_error` connotes that an error has been found or that the object is faulty.
It is possible to `create_empty` paths and pass them around. Having to use
`is_error` to test for valid emptiness can be misleading.


philberty (approved):

philberty:
Great work, I think you should be able to extract out the HIR lowering piece as well as the backend codegen piece as 2 commits from your PR which should help isolate the review down to name and type resolution code.

PR #623, opened 2021-08-11
Add mappings for canonical paths

philberty:
We need to be able to lookup canonical paths in order
to do proper name mangling as well as give more meaning
full names to the GENERIC IR. This is important since
the gcc IR is flat and makes it hard to distinguish which function
is which, especially when modules support will be added.


philberty:
this should help with your module work @dkm so the gimple output should contain the canonical paths.

PR #628, opened 2021-08-12
typecheck: add assert on resolve_root_path return value

dkm:
resolve_root_path can return nullptr in case of error.
Better raise an error than dereference the pointer.


dkm:
See https://github.com/Rust-GCC/gccrs/pull/509#discussion_r686222379

PR #631, opened 2021-08-14
Use builtin bool instead of creating new bool types for ComparisonExpr

dkm:
From Mark Wielaard : https://gcc.gnu.org/pipermail/gcc-rust/2021-August/000141.html 

> The TypeCheckExpr creates a new TyTy::BoolType for a
> ComparisonExpr. This new BoolType is unknown to TyTyResolveCompile
> which causes a crash when trying to compile the inferred new
> BoolType. Resolve this by looking up the builtin bool type.
> The new "bools_eq.rs" testcase uses several bools which show
> this issue.
> 
> Also the lhs and rhs types need to be compatible, but don't
> need to be bool type themselves. So don't append the reference
> to the inferred type. The existing "ifunaryexpr.rs" testcase
> will fail without this fix.

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-expr.h
+++ b/gcc/rust/typecheck/rust-hir-type-check-expr.h
@@ -630,10 +630,8 @@ public:
     if (result == nullptr || result->get_kind () == TyTy::TypeKind::ERROR)
       return;
 
-    // we expect this to be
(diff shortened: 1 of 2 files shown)


dkm:
bors try

dkm:
bors: merge

PR #637, opened 2021-08-18
Add support for const bool and const float

dkm:
From Mark Wielaard : https://gcc.gnu.org/pipermail/gcc-rust/2021-August/000143.html

> Handle BOOL and FLOAT in ConstFoldExpr::visit (HIR::LiteralExpr) to
> make it possible to create const bool, f32 and f64 constants. Add a
> new testcase "primconsts.rs". Not yet handled are const char and &str
> types.


philberty (approved):
nice one

PR #638, opened 2021-08-21
hir: Merge both module classes in one

CohenArthur:
At this point in the pipeline, all modules are either internal and
contain items, or external and have been previously expanded in the
macro-expand phase. Thus, all HIR modules will contain their final set
of items


philberty (approved):
LGTM

PR #639, opened 2021-08-21
External module expansion

CohenArthur:
Needs #638 

This PR parses the file associated with an external module and allows their compilation. The two test cases are not currently passing, and the locations will be inexact (the filename for the parsed module will just show as random characters), so the PR is still a draft.


gcc/rust/ast/rust-ast-full-test.cc
```diff
@@ -4120,6 +4122,37 @@ Module::get_filename ()
...
+    return;
+
+  RAIIFile file_wrap (mod_file.c_str ());
+  Linemap *linemap = Session::get_instance ().linemap;
+
+  if (file_wrap.get_raw () == nullptr)
+    rust_fatal_error (Location (), "cannot open module file %s: %m",
+		      mod_file.c_str ());
+
+  rust_debug ("Attempting to parse file %s", mod_file.c_str ());
```
philberty, line 4142: keep this debug line in the PR

dkm:
I wonder if there's a better way to group test files. As long as one test==one file, it's easy. But if we start having directories maybe there is a way to better isolate the test ? Like moving them in a sub dir and have the .exp compile the .rs contained in these subdir? If you think this is worth investigating, I can have a look (or maybe @tschwinge  already knows how to handle this, if needed?)

philberty:
> I wonder if there's a better way to group test files. As long as one test==one file, it's easy. But if we start having directories maybe there is a way to better isolate the test ? Like moving them in a sub dir and have the .exp compile the .rs contained in these subdir? If you think this is worth investigating, I can have a look (or maybe @tschwinge already knows how to handle this, if needed?)

That would be great, I am not sure what the convention should be here, I don't see any other equivalent type of testing in the testsuite that we could copy yet.

gcc/rust/ast/rust-ast-full-test.cc
```diff
@@ -4120,6 +4122,37 @@ Module::get_filename ()
...
+  // We will already have errored out appropriately in the get_filename ()
+  // method
+  if (mod_file.empty ())
+    return;
+
+  RAIIFile file_wrap (mod_file.c_str ());
+  Linemap *linemap = Session::get_instance ().linemap;
+
+  if (file_wrap.get_raw () == nullptr)
+    rust_fatal_error (Location (), "cannot open module file %s: %m",
```
philberty, line 4139: turn this into a rust_error_at() and a return.

gcc/rust/ast/rust-ast-full-test.cc
```diff
@@ -4120,6 +4122,37 @@ Module::get_filename ()
...
+		      mod_file.c_str ());
+
+  rust_debug ("Attempting to parse file %s", mod_file.c_str ());
+
+  Lexer lex (mod_file.c_str (), std::move (file_wrap), linemap);
+  Parser<Lexer> parser (std::move (lex));
+
+  auto items = parser.parse_items ();
```
philberty, line 4147: As we said in chat this is shadowing the field items for module.

gcc/rust/ast/rust-ast-full-test.cc
```diff
@@ -4120,6 +4122,37 @@ Module::get_filename ()
...
+
+  for (const auto &error : parser.get_errors ())
+    error.emit_error ();
+
+  items = std::move (items);
+  kind = ModuleKind::LOADED;
```
philberty, line 4153: Can you investigate to see if rustc has a ModuleKind::ERROR for the case where we fail to load? I am not sure if it actually helps with anything. Even if they do have this state lets not add it to this PR.

CohenArthur: So the Rust enum is the following:
```rust
enum ModKind {
    Load(Vec<Pointer<Item>>, Inline, Span),
    Unloaded
}
```
So they do not have an `Error` variant. The enum we use in gccrs is to try and reproduce those "enum variants" since we can't have them in C++ sadly. So we always keep a vector, which might be empty, and then the `kind` which corresponds to `Loaded` or `Unloaded`

gcc/testsuite/rust/compile/extern_mod4.rs
```diff
@@ -0,0 +1,17 @@
...
+fn main() {
+    unsafe {
+        let fmt_s = "%d\n\0";
+        let fmt_p = fmt_s as *const str;
+        let fmt_i8 = fmt_p as *const i8;
+
+        printf(fmt_i8, modules::return_12());
+    }
+}
+>>>>>>> 1ca3f3915e7 (mod: Add two test cases for external modules)
```
dkm, line 17: looks like a conflict marker has not been removed here?

CohenArthur: Oops, thanks! Yeah the PR is still very much WIP as the test cases do not pass yet

gcc/rust/ast/rust-ast-full-test.cc
```diff
@@ -4025,14 +4027,15 @@ filename_from_path_attribute (std::vector<Attribute> &outer_attrs)
   return path_attr.get_attr_input ().as_string ();
 }
 
-std::string
+void
 Module::get_filename ()
```
dkm, line 4031: nit: I wonder if a better name could be used here, as `get_*` usually implies some kind of return.

philberty (approved):
Nice work this is really exciting :D

dkm (approved):
I guess the multifile test can be refactored later :). Feel free to ignore the `get_filename` comment if you think it's already fine :)

gcc/rust/ast/rust-ast-full-test.cc
philberty: actually good point, What about changing this to module::process_file_path, of keep the name the same but actually return the string like before.

CohenArthur: You guys are completely right. I'll rename it when I get back home later

CohenArthur: Done in 13a9d29. Since we have to keep the string alive for a while, I think it's easier to keep it as a class member. Doing it this way avoids an extra std::move afterwards. But we can do that too

philberty: good job!

gcc/testsuite/rust/compile/torture/extern_mod4.rs
```diff
@@ -0,0 +1,17 @@
+// { dg-additional-options "-w" }
```
tschwinge, line 1: I suggest to avoid `-w` and instead spell out the current set of diagnostics: while `-w` does get rid of those you're not interested in at this time, it also forever prevents noticing those that may bogusly appear in the future.

tschwinge: Oh, is the problem here (and likewise in the other test cases) that we're getting emitted diagnostics not for this `*.rs` file but for the module:

```
$ build-gcc/gcc/xgcc -Bbuild-gcc/gcc/ source-gcc/gcc/testsuite/rust/compile/torture/extern_mod4.rs 
source-gcc/gcc/testsuite/rust/compile/torture/modules/mod.rs:1:5: warning: unused name 'return_12'
    1 | pub fn return_12() -> i32 {
      |     ^
```

Hmm.  Will have to think about how to deal with that...

gcc/testsuite/rust/compile/torture/extern_mod4.rs
```diff
@@ -0,0 +1,17 @@
...
+// { dg-output "12" }
```
tschwinge, line 2: As this is a `compile` test, it won't be executed, and you'll not see the `12` output.  Not sure if this is worth turning into an `execute` test?  (If yes, then also need to add the `return 0` hack at end of `fn main`.)

gcc/testsuite/rust/compile/torture/extern_mod4.rs
tschwinge: Though, this diagnostic is bogus here, so should fix the compiler to not emit it, right?  (But will need a way to deal with such cross-file diagnostics, generally.)

gcc/testsuite/rust/compile/torture/extern_mod4.rs
CohenArthur: Ah, right, thanks! I'll open a PR to fix this

gcc/testsuite/rust/compile/torture/extern_mod4.rs
CohenArthur: The problem is indeed that we get this extra warning on the external file

One of the issues is that right now, Ribs are getting duplicated in modules. This is what @philberty was explaining to me, and I'm paraphrasing, but what you get is the following structure of names/Ribs:
```
[modules::return_12]
[return_12]
```
So we actually have two Ribs pointing to the same function. And since the test uses `modules::return_12()`, well one of them is correctly marked as unused :D

PR #642, opened 2021-08-22
Reject duplicate field names in structs and unions.

dkm:
From Mark Wielaard : https://gcc.gnu.org/pipermail/gcc-rust/2021-August/000150.html

> Odd things happen if structs or unions have duplicate field names.
> Emit an error when lowering an struct/union item or declaration
> statement and a duplicate field name is detected.  A new testcase
> 'dup_fields.rs' checks an error is actually produced.


philberty (approved):
LGTM there is duplication here for when HIR::Items are contained within blocks of code which needs some thought but will be fixed with #531

PR #652, opened 2021-08-29
A few more location patches 

dkm:
From Mark Wieelard: https://gcc.gnu.org/pipermail/gcc-rust/2021-August/000160.html

> Here are a couple of little patches to improve or simplify locations a
> bit


CohenArthur (approved):

philberty (approved):
LGTM nice find there were a few missing places of location info.

PR #654, opened 2021-09-01
Cleanup some StructExpr related classes

dkm:
From Mark Wielaard: https://gcc.gnu.org/pipermail/gcc-rust/2021-September/000163.html

> There are various Structure Expressions that don't actually "exist"
> because they are syntactically equivalent to other constructs. So we
> never really construct or use these classes. But they are still listed
> in various visitors, which is somewhat confusing. Removing the AST and
> HIR variants of these classes really cleans up the code IMHO.


philberty (approved):
This is brilliant! LGTM

PR #656, opened 2021-09-05
Add mangling switch

### LINKED ISSUE #429: Need command line option for symbol mangling type
We need to support both legacy and V0 symbol mangling see #305 we need a command-line switch for this and a way to change the symbol mangling stategy.

CohenArthur:
Add option to choose mangling scheme.

Closes #429 

This PR splits the `Mangler` class in its own set of header and source and adds the base for v0 name mangling.

You are now able to specify the mangling scheme to use using `-frust-mangling=<value>`.

When inputting an invalid value, the compiler errors out using `unrecognized command-line option `-frust-mangling=<not_valid>`. Is there a better way to do this? Is there also a way to test this behaviour using dejagnu?

### DIFF
--- a/gcc/rust/backend/rust-mangle.cc
+++ b/gcc/rust/backend/rust-mangle.cc
@@ -1,0 +1,149 @@
+#include "rust-mangle.h"
+#include "fnv-hash.h"
+
+// FIXME: Rename those to legacy_*
+static const std::string kMangledSymbolPrefix = "_ZN";
+static const std::string kMangledSymbolDelim = "E";
+static const std::string kMangledGenericDelim = "$C$";
+static const std::string kMangledSubstBegin = "$LT$";
+static const std::string kMangledSubstEnd = "$GT$";
+
+namespace Rust {
+namespace Compile {
+
+static std::string
+legacy_mangle_name (const std::string &name)
+{
+  return std::to_string (name.size ()) + name;
+}
+
+static std::string
+legacy_mangle_canonical_path (const Resolver::CanonicalPath &path)
+{
+  std::string buffer;
+  path.iterate_segs ([&] (const Resolver::CanonicalPath &p) -> bool {
+    buffer += legacy_mangle_name (p.get ());
+    return true;
+  });
+  return buffer;
+}
+
+// rustc uses a sip128 hash for legacy mangling, but an fnv 128 was quicker to
+// implement for now
+static std::string
+legacy_hash (const std::string &fingerprint)
+{
+  Hash::FNV128 hasher;
+  hasher.write ((const unsigned char *) fingerprint.c_str (),
+		fingerprint.size ());
+
+  uint64_t hi, lo;
+  hasher.sum (&hi, &lo);
+
+  char hex[16 + 1];
+  memset (hex, 0, sizeof hex);
+  snprintf (hex, sizeof hex, "%08" PRIx64 "%08" PRIx64, lo, hi);
+
+  return "h" + std::string (hex, sizeof (hex) - 1);
+}
+
+static std::string
+legacy_mangle_self (const TyTy::BaseType *self)
+{
+  if (self->get_kind () != TyTy::TypeKind::ADT)
+    return legacy_mangle_name (self->get_name ());
+
+  const TyTy::ADTType *s = static_cast<const TyTy::ADTType *> (self);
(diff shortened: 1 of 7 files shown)


gcc/rust/lang.opt
```diff
@@ -43,6 +43,10 @@ frust-dump-
 Rust Joined RejectNegative
 -frust-dump-<type>	Dump Rust frontend internal information.
 
+frust-mangling=
+Rust Joined RejectNegative
+-frust-mangling=<version> Choose which version to use for name mangling (legacy, v0)
```
dkm, line 48: Maybe this could be a good case for TargetVariable + Enum? So that the possible version are known/handled automatically?

gcc/rust/backend/rust-mangle.h
```diff
@@ -0,0 +1,59 @@
...
+  std::string mangle_item (const TyTy::BaseType *ty,
+			   const Resolver::CanonicalPath &path,
+			   const std::string &crate_name) const;
+
+  std::string mangle_impl_item (const TyTy::BaseType *self,
+				const TyTy::BaseType *ty,
+				const std::string &name,
+				const std::string &crate_name) const;
+
+  static bool choose_mangling (std::string arg)
```
dkm, line 43: Using Enum from https://gcc.gnu.org/onlinedocs/gccint/Option-properties.html would help here and save you from doing this string->enum variant by hand

dkm (changes requested):

gcc/rust/backend/rust-mangle.h
CohenArthur: Done in 37f3b02

gcc/rust/lang.opt
CohenArthur: Done in 37f3b02

gcc/rust/backend/rust-mangle.h
CohenArthur: I used an int for simplicity, with a static cast. I don't know how ugly that is in C++, but it was easier and simpler to read than declaring the enum somewhere other than in `rust-mangle.h`

CohenArthur:
I think this can become a single commit, as handling the option is not separated from the `Mangler` class?

gcc/rust/backend/rust-mangle.cc
```diff
@@ -0,0 +1,152 @@
...
+{
+  switch (version)
+    {
+    case Mangler::MangleVersion::LEGACY:
+      return legacy_mangle_impl_item (self, ty, name, crate_name);
+    case Mangler::MangleVersion::V0:
+      gcc_unreachable ();
+    default:
+      gcc_unreachable ();
+    }
```
philberty, line 148: Pet nit not sure what others thinl but I like to add a default return of "" but have a gcc_unreachable, since c++ does not have block expressions.

philberty (approved):
LGTM the only thing i noticed are very minor nits. What do you think @dkm and @tschwinge about the enum?

gcc/rust/backend/rust-mangle.h
```diff
@@ -0,0 +1,54 @@
...
+			   const std::string &crate_name) const;
+
+  std::string mangle_impl_item (const TyTy::BaseType *self,
+				const TyTy::BaseType *ty,
+				const std::string &name,
+				const std::string &crate_name) const;
+
+  static void set_mangling (int frust_mangling_value)
+  {
+    version = static_cast<MangleVersion> (frust_mangling_value);
```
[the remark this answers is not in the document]
philberty: Maybe an assertion that the value is a value == V0 || value == legacy might be the best way to do it.

dkm (commented):
> LGTM the only thing i noticed are very minor nits. What do you think @dkm and @tschwinge about the enum?

Not really sure, I've always used this "the easy way". Having the option mechanism define an enum that's in fact never used and redefined by hand somewhere else seems odd, but I don't know if it can handle the C++ namespace/classe (guess not).

CohenArthur:
> Not really sure, I've always used this "the easy way". Having the option mechanism define an enum that's in fact never used and redefined by hand somewhere else seems odd, but I don't know if it can handle the C++ namespace/classe (guess not).

This is what I tried at first and it can't, sadly. This is why I chose to use an integer instead of a C enum lost somewhere in the options.c file

dkm (approved):

Issue #658, opened 2021-09-08
Compiler Intrinsics


philberty:
We need to support the extern "rust-intrinsics" GCC seems to have something at least pretty similar to a lot of these not 100% about all of them, things like transmute look more custom to rust.

https://github.com/rust-lang/rust/blob/master/compiler/rustc_codegen_llvm/src/intrinsic.rs

- [ ] pub fn atomic_cxchg<T: Copy>(dst: *mut T, old: T, src: T) -> (T, bool);
- [ ] pub fn atomic_cxchg_acq<T: Copy>(dst: *mut T, old: T, src: T) -> (T, bool);
- [ ] pub fn atomic_cxchg_rel<T: Copy>(dst: *mut T, old: T, src: T) -> (T, bool);
- [ ] pub fn atomic_cxchg_acqrel<T: Copy>(dst: *mut T, old: T, src: T) -> (T, bool);
- [ ] pub fn atomic_cxchg_relaxed<T: Copy>(dst: *mut T, old: T, src: T) -> (T, bool);
- [ ] pub fn atomic_cxchg_failrelaxed<T: Copy>(dst: *mut T, old: T, src: T) -> (T, bool);
- [ ] pub fn atomic_cxchg_failacq<T: Copy>(dst: *mut T, old: T, src: T) -> (T, bool);
- [ ] pub fn atomic_cxchg_acq_failrelaxed<T: Copy>(dst: *mut T, old: T, src: T) -> (T, bool);
- [ ] pub fn atomic_cxchg_acqrel_failrelaxed<T: Copy>(dst: *mut T, old: T, src: T) -> (T, bool);
- [ ] pub fn atomic_cxchgweak<T: Copy>(dst: *mut T, old: T, src: T) -> (T, bool);
- [ ] pub fn atomic_cxchgweak_acq<T: Copy>(dst: *mut T, old: T, src: T) -> (T, bool);
- [ ] pub fn atomic_cxchgweak_rel<T: Copy>(dst: *mut T, old: T, src: T) -> (T, bool);
- [ ] pub fn atomic_cxchgweak_acqrel<T: Copy>(dst: *mut T, old: T, src: T) -> (T, bool);
- [ ] pub fn atomic_cxchgweak_relaxed<T: Copy>(dst: *mut T, old: T, src: T) -> (T, bool);
- [ ] pub fn atomic_cxchgweak_failrelaxed<T: Copy>(dst: *mut T, old: T, src: T) -> (T, bool);
[...]

PR #659, opened 2021-09-08
Initial intrinsics builtin block

philberty:
This is the initial piece to get the simple intrinsic's mapped over to GCC ones. The
GCC wrapper contains a mapping system of rust names over to the builtin gcc names
as far as I can tell gcc will allow for fallback onto linking against -lm.

I think this will be a nice piece of work for new contributors, given the number of intrinsics

Addresses #658


gcc/rust/typecheck/rust-tyty.h
```diff
@@ -1071,35 +1071,74 @@ class ADTType : public BaseType, public SubstitutionRef
...
+	return "rust";
+      case ABI::INTRINSIC:
+	return "rust-intrinsic";
+      case ABI::C:
+	return "C";
+
+      case ABI::UNKNOWN:
+	return "unknown";
+      }
+    return "unknown";
```
CohenArthur, line 1115:
```suggestion
      default:
	return "unknown";
      }
```

CohenArthur: Or are there any other ABI types for the future?

gcc/rust/backend/rust-compile-intrinsic.cc
```diff
@@ -0,0 +1,105 @@
...
+{
+  rust_assert (fntype->get_abi () == TyTy::FnType::ABI::INTRINSIC);
+
+  // https://github.com/rust-lang/rust/blob/master/compiler/rustc_codegen_llvm/src/intrinsic.rs
+  // https://github.com/Rust-GCC/gccrs/issues/658
+
+  std::map<std::string, std::string> mappings;
+  mappings["sqrtf32"] = "__builtin_sqrtf";
+  // mappings["sqrtf64"] = "__builtin_sqrtl";
+  mappings["sinf32"] = "__builtin_sinf";
```
CohenArthur, line 35: I think this is a nice way of doing this. Could it be refactored with...

gcc/rust/rust-gcc.cc
```diff
@@ -744,11 +744,11 @@ Gcc_backend::Gcc_backend ()
 			math_function_type, builtin_const);
   this->define_builtin (BUILT_IN_LOG2L, "__builtin_log2l", "log2l",
 			math_function_type_long, builtin_const);
-  this->define_builtin (BUILT_IN_SIN, "__builtin_sin", "sin",
+  this->define_builtin (BUILT_IN_SINF, "__builtin_sinf", "sinf",
 			math_function_type, builtin_const);
   this->define_builtin (BUILT_IN_SINL, "__builtin_sinl", "sinl",
 			math_function_type_long, builtin_const);
-  this->define_builtin (BUILT_IN_SQRT, "__builtin_sqrt", "sqrt",
+  this->define_builtin (BUILT_IN_SQRTF, "__builtin_sqrtf", "sqrtf",
```
CohenArthur, line 751: This? In some way or another?

gcc/testsuite/rust/execute/torture/sinf32.rs
```diff
@@ -0,0 +1,22 @@
+/* { dg-output "0.893996" }*/
```
CohenArthur, line 1: I agree that this is probably a little sketch on some architectures :D We could lower the precision, because what matters in the end is that the intrinsic gets called. So instead we could have something like the following for the format string:
```rust
let a  = "%.3f\n\0";
```

gcc/rust/typecheck/rust-tyty.h
bjorn3: There will be others in the future like rust-call, efiapi and architecture specific ones.

philberty: yeah, we will be adding more here over time, rust-call is definitely the next one on the list. I am unsure how rustc handles the other architecture-specific ones, is this an LLVM feature?

gcc/rust/backend/rust-compile-intrinsic.cc
philberty: Yeah i think you are right I have changed it do do this now, which has also solved the need to link against -lm since it seemed to be falling back instead of using the buiiltin.

gcc/rust/rust-gcc.cc
philberty: done

gcc/testsuite/rust/execute/torture/sinf32.rs
philberty: I changed the dg-output to just match the 3 decimals seems to work.

philberty:
No longer seem to need to link against -lm since I have updated the PR.

gcc/rust/typecheck/rust-tyty.h
bjorn3: Rust has it's own set of abi's. Rustc maps some of them directly to LLVM abi's and has extra code for others. Most rust specific ones turn into the C abi at the LLVM side, but for example efiapi depends is either windows fastcall or the C abi depending on the architecture.

gcc/testsuite/rust/execute/torture/sinf32.rs
tschwinge: I suggest to avoid adding all such execution tests -- won't this quickly "proliferate"?

Such builtins typically are mapped to regular function calls (which we already do have confidence that they work correctly), so it ought to suffice to just scan '-fdump-tree-original' (see #660).  I've prepared a commit to "Turn 'rust/execute/torture/sinf32.rs', 'rust/execute/torture/sqrtf32.rs' into 'rust/compile/torture/intrinsics-1.rs'" -- if you agree, want me to push that to this branch here?

philberty: That sounds like a much better plan @tschwinge lets do it!

philberty: this looks much better lets merge it :D

PR #663, opened 2021-09-10
Move module output test case from compile/ to execute/

CohenArthur:
Fixes [this issue](https://github.com/Rust-GCC/gccrs/pull/639#discussion_r705922519) by @tschwinge 

I think this shows that we need to figure out a better way to keep test files and their modules, as pointed out by @dkm [here](https://github.com/Rust-GCC/gccrs/pull/639#issuecomment-905457557). Here, the modules/ directory is a straight up duplicate of the other one in compile/ which I'm not a fan of.

I have no experience with having a separate .exp in the subdir but I can try and I think that it would be better. Maybe it can also be simpler if the regex to match *.rs files only looks inside one level of subdirectories? But I can think of a few places where this might be an issue


philberty (approved):
LGTM

PR #664, opened 2021-09-10
Fix usage of path attributes and module filename processing

CohenArthur:
`#[path]` attributes were used wrongly in two ways:
- The attribute's value would still contain extra data such as an equal
sign and whitespaces, which need to be discarded in order to fetch the
proper file path.
- `#[path]`s are relative to the directory where the current source file
is located. So we actually need to compute them later on down the line,
once the including directory has been figured out.

Another issue was that in the case that no candidates were found, or too
many were founds, the function `process_file_path` would still return
"successfully" despite erroring out. This causes the function to return
early and thus not assign any string to the `module_file` member.

We also have the ability to emit an error for the module's location,
which looks better for the user, instead of using a new `Location()`
like the session manager does (because it cannot do otherwise, as it
is starting to parse a file and thus has no file location information)

### DIFF
--- a/gcc/rust/ast/rust-ast-full-test.cc
+++ b/gcc/rust/ast/rust-ast-full-test.cc
@@ -4010,7 +4010,15 @@ filename_from_path_attribute (std::vector<Attribute> &outer_attrs)
       return "";
     }
 
-  auto path = path_attr.get_attr_input ().as_string ();
+  auto path_value = path_attr.get_attr_input ().as_string ();
+
+  // At this point, the 'path' is of the following format: '= "<file.rs>"'
+  // We need to remove the equal sign and only keep the actual filename.
+  // In order to do this, we can simply go through the string until we find
+  // a character that is not an equal sign or whitespace
+  auto filename_begin = path_value.find_first_not_of ("=\t ");
+
+  auto path = path_value.substr (filename_begin);
 
   // On windows, the path might mix '/' and '\' separators. Replace the
   // UNIX-like separators by MSDOS separators to make sure the path will resolve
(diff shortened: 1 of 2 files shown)


gcc/rust/ast/rust-ast-full-test.cc
```diff
@@ -4010,7 +4010,15 @@ filename_from_path_attribute (std::vector<Attribute> &outer_attrs)
...
     }
 
-  auto path = path_attr.get_attr_input ().as_string ();
+  auto path_value = path_attr.get_attr_input ().as_string ();
+
+  // At this point, the 'path' is of the following format: '= "<file.rs>"'
+  // We need to remove the equal sign and only keep the actual filename.
+  // In order to do this, we can simply go through the string until we find
+  // a character that is not an equal sign or whitespace
+  auto filename_begin = path_value.find_first_not_of ("=\t ");
```
philberty, line 4019: I wonder is it worth checking in Rustc to see if their Attributes structure cleans this up for you in the parser so you get a structure like:

string attr_name, bool has_rhs; string get_rhs;

philberty (approved):
LGTM

gcc/rust/ast/rust-ast-full-test.cc
philberty: I think we will need some type of code like that for attributes in general so lets make an issue. Don't change it for this PR.

philberty: https://github.com/Rust-GCC/gccrs/issues/665

PR #685, opened 2021-09-18
Add v0 type mangling prefixing for simple types

CohenArthur:
This PR adds the generation of type prefixes for simple types, which are numeric types, booleans, chars, strings, empty tuples/unit types and placeholder types.

I'm unsure as to how to test this, even in the long run. There might be some shenanigans we can pull using an elf reader and regexes in order to compare ABI names with rustc.

The entire implementation of v0 name mangling is very large, so I thought I'd split it up in multiple PRs.

### DIFF
--- a/gcc/rust/backend/rust-mangle.cc
+++ b/gcc/rust/backend/rust-mangle.cc
@@ -80,6 +80,92 @@ legacy_mangle_self (const TyTy::BaseType *self)
 }
 
 static std::string
+v0_tuple_prefix (const TyTy::BaseType *ty)
+{
+  if (ty->is_unit ())
+    return "u";
+
+  // FIXME: ARTHUR: Add rest of algorithm
+  return "";
+}
+
+static std::string
+v0_numeric_prefix (const TyTy::BaseType *ty)
+{
+  static const std::map<std::string, std::string> num_prefixes = {
+    {"[i8]", "a"},    {"[u8]", "h"},	{"[i16]", "s"}, {"[u16]", "t"},
+    {"[i32]", "l"},   {"[u32]", "m"},	{"[i64]", "x"}, {"[u64]", "y"},
+    {"[isize]", "i"}, {"[usize]", "j"}, {"[f32]", "f"}, {"[f64]", "d"},
+  };
+
+  auto ty_kind = ty->get_kind ();
+  auto ty_str = ty->as_string ();
+  auto numeric_iter = num_prefixes.end ();
+
+  // Special numeric types
+  if (ty_kind == TyTy::TypeKind::ISIZE)
+    return "i";
+  else if (ty_kind == TyTy::TypeKind::USIZE)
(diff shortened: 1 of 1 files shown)


dkm:
I guess you should use the `scan-assembler` (see https://gcc.gnu.org/onlinedocs/gccint/Final-Actions.html#Final-Actions ) to look for a specific symbol name somewhere ?

CohenArthur:
Fantastic Marc, thanks a lot!!!! I didn't know this existed. I guess we can think about comparing the output to rustc later on but it would be quite easy with this

dkm:
Yes. You can already check that you end up with the expected names in the assembly :).

philberty (approved):
LGTM

gcc/rust/backend/rust-mangle.cc
```diff
@@ -79,6 +79,92 @@ legacy_mangle_self (const TyTy::BaseType *self)
...
+    return "u";
+
+  // FIXME: ARTHUR: Add rest of algorithm
+  return "";
+}
+
+static std::string
+v0_numeric_prefix (const TyTy::BaseType *ty)
+{
+  static const std::map<std::string, std::string> num_prefixes = {
```
philberty, line 95: This is neat :)

gcc/rust/backend/rust-mangle.cc
```diff
@@ -79,6 +79,92 @@ legacy_mangle_self (const TyTy::BaseType *self)
...
+{
+  if (ty->is_unit ())
+    return "u";
+
+  // FIXME: ARTHUR: Add rest of algorithm
+  return "";
+}
+
+static std::string
+v0_numeric_prefix (const TyTy::BaseType *ty)
```
philberty, line 93: I think in later PR's you might want to make this more error safe by doing something like:

```
static bool v0_numeric_prefix(const TyTy::BaseType* ty, std::string& prefix)
```

So in future changes you can error by returning false, so it becomes:

```
std::string prefix;
bool ok = v0_numeric_prefix(ty, prefix);
rust_asser(ok);
```

gcc/rust/backend/rust-mangle.cc
```diff
@@ -79,6 +79,92 @@ legacy_mangle_self (const TyTy::BaseType *self)
...
+      return "e";
+    case TyTy::TypeKind::NEVER:
+      return "z";
+
+      // Placeholder types
+    case TyTy::TypeKind::ERROR:	      // Fallthrough
+    case TyTy::TypeKind::INFER:	      // Fallthrough
+    case TyTy::TypeKind::PLACEHOLDER: // Fallthrough
+    case TyTy::TypeKind::PARAM:
+      // FIXME: TyTy::TypeKind::BOUND is also a valid variant in rustc
```
philberty, line 137: Good to know thanks for the heads up

Issue #247, opened 2021-02-23
Rebasing against GCC


philberty:
Last time the front-end was rebased against GCC Master was on:

https://github.com/Rust-GCC/gccrs/commit/b8ae0812ee567fa380f0a8c7f30c7a660a5069c3

We need a strategy for when we rebase the front-end as this requires a force push back to master

tschwinge:
@philberty, do you really intend to `git rebase`, or rather `git merge`?  The latter would be much simpler to do (and I shall offer to do that every once in a while; I'm loosely tracking GCC master branch for work purposes anyway).  Doing it this way also avoids the _non-fast-forward issue_ (Git history rewriting).  What, for this project here, would be the benefit of a (periodical) real `git rebase`?

philberty:
Good point, I have been assuming that in the event of going upstream to GCC this would require a rebase against GCC so all the front-end work is ontop at that point but if merges are ok then let's do that.

tschwinge:
Eventually integrating this work into GCC upstream would -- per current upstream rules; I didn't invent these ;-) -- not be done as a `git merge`, but also would not preserve the individual development commits (of this repository here), but would rather squash all development commits into one big commit.  So, some kind of _rebase_ will be required at that point, but that's then the simple matter of (logically) rebasing/squashing the whole development history into one commit, on top of upstream master branch.  (And that's not something we have to worry about right now, so no point spending any effort on that, from my point of view.)

tschwinge:
..., and of course, any changes to existing GCC code (bug fixes, enhancements, etc.) shall be contributed individually, at any point in time, outside of the big "Add Rust front end" commit.  Also, adding run-time libraries, etc. is often done in separate commits from the one adding the front end proper.  We'll sort that out once we get to that point.  ;-)

NalaGinrut:
I think we've been using `git rebase` for a long time since this project was awake last year. Why not just keep it?

tschwinge:
@NalaGinrut: a `git rebase` rewrites Git history (making invalid existing refernces to Git commits), requires force-push, and thus requires that any open development branches also be rebased, too.  And on the other hand, I don't see any benefit of a `git rebase` over a simple `git merge`; the latter would seem simpler.  Would you prefer `git rebase`?

NalaGinrut:
[...]

PR #689, opened 2021-09-23
x86: Instead of 'TARGET_ISA_[...]', 'TARGET_ISA2_[...]', use 'TARGET_[...]' [#247]

tschwinge:
... in preparation for a merge from GCC upstream, where the former disappear.


philberty (approved):
LGTM

PR #695, opened 2021-09-29
Fix raw byte string parsing of zero and out of range bytes

CohenArthur:
Allow \0 escape in raw byte string and reject non-ascii byte
values. Change parse_partial_hex_escapes to not skip bad characters to
provide better error messages.

Add rawbytestring.rs testcase to check string, raw string, byte string
and raw byte string parsing.

Patch: https://code.wildebeest.org/git/user/mjw/gccrs/commit/?h=parse-raw-byte-string
Email: https://gcc.gnu.org/pipermail/gcc-rust/2021-September/000227.html


philberty (approved):
LGTM I love how github thinks the test case is a binary file.

PR #698, opened 2021-09-30
Implement Byte Strings

### LINKED ISSUE #697: Byte strings are not str's
I tried this code:

```rust
pub fn test() {
    let a:&str = b"hello world";
}
```

https://godbolt.org/z/7qaK8n8bx

```
expected `str`, found array `[u8; 11]`
```

Instead, this happened: it compiled and treated the byte string as a normal str.

### Meta

- What version of Rust GCC were you using, git sha if possible. 28f527c9598339cf834a30b5ee1f14258b8ecbb2

philberty:
Byte strings are not str's they are arrays of [u8; capacity], this
preserves their type guarantees as a byte string.

This patch merges work from Mark to implement the correct typing, the
missing piece was that each implicit type needed its own implicit id, other
wise their is a loop in looking up the covariant types.

Fixes #697

Co-authored-by: Mark Wielaard <mark@klomp.org>

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.h
+++ b/gcc/rust/backend/rust-compile-expr.h
@@ -304,12 +304,51 @@ public:
 	}
 	return;
 
-      case HIR::Literal::STRING:
-	case HIR::Literal::BYTE_STRING: {
+	case HIR::Literal::STRING: {
 	  auto base = ctx->get_backend ()->string_constant_expression (
 	    literal_value->as_string ());
 	  translated
 	    = ctx->get_backend ()->address_expression (base, expr.get_locus ());
+	}
+	return;
+
+	case HIR::Literal::BYTE_STRING: {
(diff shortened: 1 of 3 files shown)


CohenArthur (approved):

gcc/rust/backend/rust-compile-expr.h
```diff
@@ -304,15 +304,54 @@ class CompileExpr : public HIRCompileBase
...
 	    = ctx->get_backend ()->address_expression (base, expr.get_locus ());
 	}
 	return;
 
+	case HIR::Literal::BYTE_STRING: {
+	  TyTy::BaseType *tyty = nullptr;
+	  if (!ctx->get_tyctx ()->lookup_type (
+		expr.get_mappings ().get_hirid (), &tyty))
+	    {
+	      rust_fatal_error (expr.get_locus (),
```
philberty, line 320: this should be a different error message

PR #725, opened 2021-10-07
Add array constant folding

### LINKED ISSUE #681: Array Constant Folding
I tried this code:

```rust
const TEST: [i32; 16] = [2, 6, 3, 10, 7, 0, 4, 13, 1, 11, 12, 5, 9, 14, 15, 8];
```

I expected to see this happen: compiles without error

Instead, this happened:
```
<source>:1:25: error: non const value
    1 | const TEST: [i32; 16] = [2, 6, 3, 10, 7, 0, 4, 13, 1, 11, 12, 5, 9, 14, 15, 8];
      |                         ^
<source>:1:1: error: non const value
    1 | const TEST: [i32; 16] = [2, 6, 3, 10, 7, 0, 4, 13, 1, 11, 12, 5, 9, 14, 15, 8];
      | ^
```

### Meta

- What version of Rust GCC were you using, git sha if possible. c644ee4c4351e3590f5396e94ec24ad7b828a954

## Fix Guide

The constant folder implementation in Rust is done as part of type checking,  So for example Array Types contain a GCC backend Bexpression* which is the compile-time verified constant.

- [ ] Add missing ArrayType compiler https://github.com/Rust-GCC/gccrs/blob/28f527c9598339cf834a30b5ee1f14258b8ecbb2/gcc/rust/typecheck/rust-hir-const-fold.h#L44 implementation from here should work by updating it to call ConstFoldType instead of the backend one https://github.com/Rust-GCC/gccrs/blob/28f527c9598339cf834a30b5ee1f14258b8ecbb2/gcc/rust/backend/rust-compile-context.h#L507-L513
- [ ] add missing visitors for https://github.com/Rust-GCC/gccrs/blob/28f527c9598339cf834a30b5ee1f14258b8ecbb2/gcc/rust/backend/rust-compile-expr.h#L345-L394
- [ ] copy them here https://github.com/Rust-GCC/gccrs/blob/28f527c9598339cf834a30b5ee1f14258b8ecbb2/gcc/rust/typecheck/rust-hir-const-fold.h#L392
- [ ] The visitors could be copied but for each element in the array expr you will need to call ConstFoldExpr for each to get their Expression

rodrigovalle:
Fixes #681 

I created a new ConstFoldArrayElems folder to visit ArrayElemsCopied and ArrayElemsValues while holding onto the original ArrayExpr object that we recurse into.


dkm:
Also, there's still the copyright assignment / DCO to be checked. What are you planning on using @rodrigovalle  ?

philberty:
bors try

rodrigovalle:
Ok thanks! I've removed the nullptr check and the first commit so it should merge nicely onto master. I also have a patch to fold ArrayElemsCopied but I will PR that separately.

rodrigovalle:
This second push was to remove the nullptr check after we call ConstFoldArrayElems::fold since we use error_expression as the default return value.

philberty:
@rodrigovalle this is fantastic thanks for rebasing this :D

Issue #703, opened 2021-10-05
Get rid of lambda within HIR::ArrayElemsValues


philberty:
https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/hir/tree/rust-hir-expr.h#L799-L806

Lambda's like this need to be removed they make working with the IR's more complex for static analysis. To fix this issue you should be able to remove this lambda helper function and provide a method to return a reference to the values like:

```c++
std::vector<std::unique_ptr<Expr> >& get_values() { return values' }
```

This can then be used in a for loop:

```
for (auto& v : expr.get_values()) { ... }
```

Issue #707, opened 2021-10-05
Get rid of lambda within HIR::TupleStruct


philberty:
https://github.com/Rust-GCC/gccrs/blob/99c28309d3553346d4f0337dbae49f4a8e48da01/gcc/rust/hir/tree/rust-hir-item.h#L1613-L1620

Lambda's like this need to be removed they make working with the IR's more complex for static analysis. To fix this issue you should be able to remove this lambda helper function and provide a method to return a reference to the values like

Create a new helper to access the elements:

```c++
std::vector<TupleField>& get_fields() { return fields; }
```

Issue #705, opened 2021-10-05
Get rid of lambda HIR::StructExprStructFields


philberty:
https://github.com/Rust-GCC/gccrs/blob/99c28309d3553346d4f0337dbae49f4a8e48da01/gcc/rust/hir/tree/rust-hir-expr.h#L1494-L1501

Lambda's like this need to be removed they make working with the IR's more complex for static analysis. To fix this issue you should be able to remove this lambda helper function and provide a method to return a reference to the values like

This class already provides a helper to access the elements:

https://github.com/Rust-GCC/gccrs/blob/99c28309d3553346d4f0337dbae49f4a8e48da01/gcc/rust/hir/tree/rust-hir-expr.h#L1503

Also get rid of the get_fields_as_owner, HIR is designed to be immutable:

https://github.com/Rust-GCC/gccrs/blob/99c28309d3553346d4f0337dbae49f4a8e48da01/gcc/rust/hir/tree/rust-hir-expr.h#L1513-L1522

Issue #704, opened 2021-10-05
Get rid of lambda within HIR::TupleExpr


philberty:
https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/hir/tree/rust-hir-expr.h#L1073-L1080

Lambda's like this need to be removed they make working with the IR's more complex for static analysis. To fix this issue you should be able to remove this lambda helper function and provide a method to return a reference to the values like

This class already provides a helper to access the elements:

https://github.com/Rust-GCC/gccrs/blob/99c28309d3553346d4f0337dbae49f4a8e48da01/gcc/rust/hir/tree/rust-hir-expr.h#L1066

Use this instead of the lambda.

PR #726, opened 2021-10-08
Remove lambda iterators in various HIR classes

dkm:
From David Faust : https://gcc.gnu.org/pipermail/gcc-rust/2021-October/000233.html

This patch removes the lambda iterators used in various HIR objects.
These iterators make interacting with the IR for static analysis more
difficult. Instead, get_X () helpers are added for accessing elements,
and uses of the iterators replaced with for loops.

The following objects are adjusted in this patch:
- HIR::TupleExpr
- HIR::StructExprField
- HIR::StructStruct
- HIR::TupleStruct

Fixes: #703, #704, #705, #706, #707

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.h
+++ b/gcc/rust/backend/rust-compile-expr.h
@@ -412,11 +412,11 @@ public:
 
   void visit (HIR::ArrayElemsValues &elems) override
   {
-    elems.iterate ([&] (HIR::Expr *e) mutable -> bool {
-      Bexpression *translated_expr = CompileExpr::Compile (e, ctx);
-      constructor.push_back (translated_expr);
-      return true;
-    });
+    for (auto &elem : elems.get_values ())
+      {
+	Bexpression *translated_expr = CompileExpr::Compile (elem.get (), ctx);
+	constructor.push_back (translated_expr);
+      }
   }
 
   void visit (HIR::ArrayElemsCopied &elems) override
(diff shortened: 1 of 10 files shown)


dkm:
@dafaust , it seems that the errors reported by the linter are introduced by your patch. See for example:
```
$ gcc/rust/hir/tree$ git reset --hard 99c28309d3553346d4f0337dbae49f4a8e48da01
$ gcc/rust/hir/tree$ clang-format -i *
$ git status 
<nothing changed>
$ git reset --hard dkm/pr/dfaust/remove_lambda
$ gcc/rust/hir/tree$ clang-format -i *
$ git diff
 diff --git a/gcc/rust/hir/tree/rust-hir-expr.h b/gcc/rust/hir/tree/rust-hir-expr.h
index d9958a153be..5457ed7d993 100644
--- a/gcc/rust/hir/tree/rust-hir-expr.h
+++ b/gcc/rust/hir/tree/rust-hir-expr.h
@@ -759,12 +759,12 @@ protected:
 // Value array elements
 class ArrayElemsValues : public ArrayElems
 {
-  std::vector<std::unique_ptr<Expr> > values;
+  std::vector<std::unique_ptr<Expr>> values;

   // TODO: should this store location data?

 public:
-  ArrayElemsValues (std::vector<std::unique_ptr<Expr> > elems)
+  ArrayElemsValues (std::vector<std::unique_ptr<Expr>> elems)
     : values (std::move (elems))
   {}
....
```

I guess you could simply apply clang format on your patch. I can do it if you want, as I'm merely proxying your changes here, just me know (I can also be wrong :) )

dafaust:
Ahh I see, looks like you are right @dkm, matching to the existing format like
`std::vector<std::unique_ptr<Expr> > &get_values () `
pacifies clang-format (at least locally, we shall see what github thinks... :P )

PR #728, opened 2021-10-12
Remove `AST::BlockExpr` lambda and add `get_statements`

### LINKED ISSUE #724: Get rid of lambda within AST::BlockExpr
https://github.com/Rust-GCC/gccrs/blob/99c28309d3553346d4f0337dbae49f4a8e48da01/gcc/rust/ast/rust-expr.h#L2441-L2448
Lambda's like this need to be removed they make working with the IR's more complex for static analysis. To fix this issue you should be able to remove this lambda helper function and provide a method to return a reference to the values like

```c++
std::vector<Stmt>& get_statements() { return statements; }
```

kazuki-hanai:
This PR Fixes #724.

This patch removes lambda iterators in `AST::BlockExpr` and replace these with `get_statements`. These lambda iterators need to be removed they make working with the IR's more complex for static analysis.

### DIFF
--- a/gcc/rust/ast/rust-expr.h
+++ b/gcc/rust/ast/rust-expr.h
@@ -2438,15 +2438,6 @@ public:
   void mark_for_strip () override { marked_for_strip = true; }
   bool is_marked_for_strip () const override { return marked_for_strip; }
 
-  void iterate_stmts (std::function<bool (Stmt *)> cb)
-  {
-    for (auto it = statements.begin (); it != statements.end (); it++)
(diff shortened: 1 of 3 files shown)


wan-nyan-wan:
This is the first time to contribute to GCC. So I haven't known about the copyright assignment / DCO. I have no problem to agree on these rules. But I don't know how to check the DCO, so I'm googling about it.

wan-nyan-wan:
I rebased needless commits.

Issue #731, opened 2021-10-13
Use new TypeMutability in HIR


philberty:
With the new mutability enum lets make this useable from HIR onwards. https://github.com/Rust-GCC/gccrs/blob/7a341772dcf51445af08071eeb0ed0ea93cd60a1/gcc/rust/typecheck/rust-tyty.h#L38

- [ ] Rename to Mutability so it is generic from Types
- [ ] Move to a common header like gcc/rust/util/rust-common.h
- [ ] Update HIR::IdentifierPattern https://github.com/Rust-GCC/gccrs/blob/7a341772dcf51445af08071eeb0ed0ea93cd60a1/gcc/rust/hir/tree/rust-hir-pattern.h#L73
- [ ] Update ReferencePattern https://github.com/Rust-GCC/gccrs/blob/7a341772dcf51445af08071eeb0ed0ea93cd60a1/gcc/rust/hir/tree/rust-hir-pattern.h#L330
- [ ] Update BorrowExr https://github.com/Rust-GCC/gccrs/blob/7a341772dcf51445af08071eeb0ed0ea93cd60a1/gcc/rust/hir/tree/rust-hir-expr.h#L174
- [ ] Update RawPointerType https://github.com/Rust-GCC/gccrs/blob/7a341772dcf51445af08071eeb0ed0ea93cd60a1/gcc/rust/hir/tree/rust-hir-type.h#L451
- [ ] Update ReferenceType https://github.com/Rust-GCC/gccrs/blob/7a341772dcf51445af08071eeb0ed0ea93cd60a1/gcc/rust/hir/tree/rust-hir-type.h#L519
- [ ] Update StaticItem https://github.com/Rust-GCC/gccrs/blob/7a341772dcf51445af08071eeb0ed0ea93cd60a1/gcc/rust/hir/tree/rust-hir-item.h#L2036

PR #738, opened 2021-10-13
Refactor HIR to use new Mutability enum

dafaust:
Introduce a new header rust/util/rust-common.h and move the enum
previously Rust::TyTy::TypeMutability there as Rust::Mutability.

Update the following objects to use Mutability enum rather than a bool:
- HIR::IdentifierPattern
- HIR::ReferencePattern
- HIR::StructPatternFieldIdent
- HIR::BorrowExpr
- HIR::RawPointerType
- HIR::ReferenceType
- HIR::StaticItem
- HIR::ExternalStaticItem

Also add a HIR::SelfParam::get_mut () helper, mapping its internal
custom mutability to the common Rust::Mutability.

Fixes: #731

### DIFF
--- a/gcc/rust/backend/rust-compile-fnparam.h
+++ b/gcc/rust/backend/rust-compile-fnparam.h
@@ -40,7 +40,7 @@ public:
 
   void visit (HIR::IdentifierPattern &pattern) override
   {
-    if (!pattern.is_mut)
+    if (!pattern.is_mut ())
       decl_type = ctx->get_backend ()->immutable_type (decl_type);
 
     translated
--- a/gcc/rust/backend/rust-compile-var-decl.h
+++ b/gcc/rust/backend/rust-compile-var-decl.h
@@ -54,7 +54,7 @@ public:
 
   void visit (HIR::IdentifierPattern &pattern) override
   {
-    if (!pattern.is_mut)
+    if (!pattern.is_mut ())
       translated_type = ctx->get_backend ()->immutable_type (translated_type);
 
     translated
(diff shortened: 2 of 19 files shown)


philberty:
> One thing I only partially addressed here -
> 
> `HIR::SelfParam` is a little odd with this `ImplicitSelfKind` that may be None
> 
> https://github.com/Rust-GCC/gccrs/blob/3057d98a4e3fc777b06e51cd678723f5c5b4c898/gcc/rust/hir/tree/rust-hir-item.h#L308-L318
> 
> The constructor has a ternary which maps a boolean is_mut from AST into the appropriate one of these, otherwise it would be done e.g. here:
> 
> https://github.com/Rust-GCC/gccrs/blob/3057d98a4e3fc777b06e51cd678723f5c5b4c898/gcc/rust/hir/rust-ast-lower.cc#L414-L416
> 
> I wonder if we can get rid of this `ImplicitSelfKind` enum in favor of using Mutability - looks like None is only used for error cases?

Its interesting that in rustc they choose to leave it as is: https://doc.rust-lang.org/beta/nightly-rustc/rustc_hir/hir/enum.ImplicitSelfKind.html

I don't have strong opinions on this it might make sense to use this new enum and have a new is reference flag but if you do it make it in a separate PR.

dafaust:
> Its interesting that in rustc they choose to leave it as is: https://doc.rust-lang.org/beta/nightly-rustc/rustc_hir/hir/enum.ImplicitSelfKind.html
> 
> I don't have strong opinions on this it might make sense to use this new enum and have a new is reference flag but if you do it make it in a separate PR.

OK I will think about this some more, agreed on a separate PR.

PR #748, opened 2021-10-18
Const fold ArrayElemsCopied

rodrigovalle:
Added support for const fold inside of an ArrayElemsCopied expression, e.g. `const array: [i32; 10] = [0; 10];`. I also removed a .swp file that I accidentally merged in a previous commit and added a test for this new functionality.


philberty:
> I also ran clang-format and added DCO. Maybe it'd be possible to make DCO be one of the checks in github actions? But it would probably also have to know about the MAINTAINERS file.

Thats a clever idea, what we could do is have a check if the person has added the DCO sign off to their commit or check that they have done copyright assignment by maintaining a file of emails separate to the maintainers file. Lets open an issue for this.

PR #751, opened 2021-10-19
Enable the usage of selftests

CohenArthur:
From what I can understand, self tests seem to only be used in the C family of languages and in no other gcc frontend. And for some reason one of the self test for the gccrs frontend seem to be the parsing of C code and C types, which fails, which is why this PR is a draft. Once this is fixed, selftests should be enabled for the frontend and we should be able to add some to the project.

### DIFF
--- a/gcc/c-family/c-common.h
+++ b/gcc/c-family/c-common.h
@@ -1512,8 +1512,10 @@ extern tree braced_lists_to_strings (tree, tree);
 namespace selftest {
   /* Declarations for specific families of tests within c-family,
      by source file, in alphabetical order.  */
+  extern void c_diagnostic_c_tests (void);
   extern void c_format_c_tests (void);
   extern void c_indentation_c_tests (void);
+  extern void c_opt_problem_cc_tests (void);
   extern void c_pretty_print_c_tests (void);
   extern void c_spellcheck_cc_tests (void);
 
--- a/gcc/rust/rust-lang.cc
+++ b/gcc/rust/rust-lang.cc
@@ -32,6 +32,7 @@
 #include "convert.h"
 #include "langhooks.h"
 #include "langhooks-def.h"
+#include "selftest.h"
 
 #include <mpfr.h>
 // note: header files must be in this order or else forward declarations don't
@@ -433,6 +434,29 @@ rust_localize_identifier (const char *ident)
 #define LANG_HOOKS_GETDECLS grs_langhook_getdecls
 #define LANG_HOOKS_GIMPLIFY_EXPR grs_langhook_gimplify_expr
 #define LANG_HOOKS_EH_PERSONALITY grs_langhook_eh_personality
+
+#if CHECKING_P
+
+#undef LANG_HOOKS_RUN_LANG_SELFTESTS
+#define LANG_HOOKS_RUN_LANG_SELFTESTS selftest::run_rust_tests
+
+namespace selftest {
+
+static void
+simple_assert ()
+{
+  ASSERT_TRUE (true);
+}
+
+void
+run_rust_tests ()
+{
+  // Call tests for the rust frontend here
+  simple_assert ();
+}
+} // namespace selftest
+
+#endif /* !CHECKING_P */
 
 // Expands all LANG_HOOKS_x of GCC
 struct lang_hooks lang_hooks = LANG_HOOKS_INITIALIZER;
--- a/gcc/c-family/c-common.cc
+++ b/gcc/c-family/c-common.cc
@@ -9111,6 +9111,8 @@ c_family_tests (void)
   c_indentation_c_tests ();
   c_pretty_print_c_tests ();
   c_spellcheck_cc_tests ();
+  c_diagnostic_c_tests ();
+  c_opt_problem_cc_tests ();
 }
 
 } // namespace selftest
--- a/gcc/diagnostic.cc
+++ b/gcc/diagnostic.cc
@@ -2466,7 +2466,7 @@ test_num_digits ()
 /* Run all of the selftests within this file.  */
 
 void
-diagnostic_c_tests ()
+c_diagnostic_c_tests ()
 {
   test_print_escaped_string ();
   test_print_parseable_fixits_none ();
--- a/gcc/opt-problem.cc
+++ b/gcc/opt-problem.cc
@@ -324,7 +324,7 @@ test_opt_result_failure_at (const line_table_case &case_)
 /* Run all of the selftests within this file.  */
 
 void
-opt_problem_cc_tests ()
+c_opt_problem_cc_tests ()
 {
   test_opt_result_success ();
   for_each_line_table_case (test_opt_result_failure_at);
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -209,8 +209,16 @@ rust.uninstall:
 	-rm -f $(RUST_ALL_OBJS)
 # ^those two are a maybe
 
-# No rust-specific selftests
-selftest-rust:
+# Enable selftests for the rust frontend
+selftest-rust: s-selftest-rust
+
+RUST_SELFTEST_FLAGS = -xrs $(SELFTEST_FLAGS)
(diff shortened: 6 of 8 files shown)


CohenArthur:
@philberty @tschwinge I ended up returning early from the "C tests" that are in gcc files common to all languages in 606250b. This is definitely not a good solution and I will also ask for advice on the mailing list as to how to handle the situation. I think these tests could be moved, or maybe we might need to add some extra flags to not run them if we are running the rust selftests.

CohenArthur:
But we can now have selftests and their failure is caught by the CI :)

tschwinge:
@CohenArthur, given the replies to your <http://mid.mail-archive.com/20211105093847.urtlztspddvlmahz@oneshade> "C-family selftests in language-independant source files", do you know how to proceed or would like me to look into it at this point?  (No worries, of course, if you've just not yet gotten around to continuing that.)

CohenArthur:
Hi @tschwinge ! Thanks for offering your help :D I think I know how to proceed, or at least where to start, but I haven't gotten around to it. Thanks for the review, I'll fix up that commit

CohenArthur:
I think it should be good now @tschwinge @philberty :) I'll add some unit tests for the base 62 code as a starting point

philberty:
nice work @CohenArthur this is something I am really excited about! :D

gcc/rust/rust-lang.cc
```diff
@@ -434,6 +435,31 @@ rust_localize_identifier (const char *ident)
 #define LANG_HOOKS_GIMPLIFY_EXPR grs_langhook_gimplify_expr
 #define LANG_HOOKS_EH_PERSONALITY grs_langhook_eh_personality
 
+#if CHECKING_P
```
philberty, line 438: can this or the tests be moved to a rust-base62-test.h?

tschwinge:
Great; I'll have a look.

gcc/rust/rust-lang.cc
CohenArthur: It should defintely be possible as the C frontend is doing it. I was in a rush when finishing this yesterday. I was planning on giving it a look in a next PR where I'd add actual assertions for the base62 functions, but I can also do it here if you prefer

CohenArthur:
> As far as i can tell this looks good but I am not sure about the final commit [777d1d2](https://github.com/Rust-GCC/gccrs/commit/777d1d2bc7a65ac8b6bd874f5a2799d57c52a11a) do we need to touch this or can we drop it?

The last commit is "extremely important" as it prevents our selftests from running some C specific selftests which we obviously fail. This is the commit that fixes what David Malcolm was suggesting over on the gcc mailing list

dkm:
> > As far as i can tell this looks good but I am not sure about the final commit [777d1d2](https://github.com/Rust-GCC/gccrs/commit/777d1d2bc7a65ac8b6bd874f5a2799d57c52a11a) do we need to touch this or can we drop it?
> 
> The last commit is "extremely important" as it prevents our selftests from running some C specific selftests which we obviously fail. This is the commit that fixes what David Malcolm was suggesting over on the gcc mailing list

Should this be sent upstream instead ? If that was suggested by David, that should be easy to have it merged ?

gcc/rust/rust-lang.cc
tschwinge: The main `LANG_HOOKS_RUN_LANG_SELFTESTS` should indeed be defined here, including the `selftest::rust_tests` function (or `selftest::run_rust_tests`, to keep it similar to C, C++?).  The individual tests (like for base62) should then reside in the respective files.  The `simple_assert` test here should eventually go away; but OK to leave it for now, if you'd like.

gcc/rust/util/rust-base62.cc
```diff
@@ -17,6 +17,7 @@
 // <http://www.gnu.org/licenses/>.
 
 #include "rust-base62.h"
+#include "selftest.h"
```
tschwinge, line 20: Maybe defer these changes here until we actually need them?

.github/workflows/ccpp.yml
```diff
@@ -41,7 +41,8 @@ jobs:
            ../configure \
                --enable-languages=rust \
                --disable-bootstrap \
-               --enable-multilib
+               --enable-multilib \
+               --enable-selftest
```
tschwinge, line 45: As far as I can tell, there is no such option in GCC, so that is a no-op, and should be removed.

gcc/c/c-lang.c
```diff
@@ -60,6 +60,8 @@ run_c_tests (void)
 {
   /* Run selftests shared within the C family.  */
   c_family_tests ();
+  diagnostic_c_tests ();
+  opt_problem_cc_tests ();
```
tschwinge, line 64: I suppose, these two should rather move into `gcc/c-family/c-common.c:selftest::c_family_tests`, to make (or rather: keep) them effective for both C and C++ front ends?  Also, should (eventually, need not be now) rename to `c_family_[...]` (or similar) to make clear their context, or split up into generic vs. c-family parts, or similar.  As the discussion has been started already, this needs to be sorted out in GCC upstream, once GCC again is in development stage 1 (2022-04, I suppose).  I'm able to help, as necessary.

gcc/rust/rust-lang.cc
```diff
@@ -434,6 +435,31 @@ rust_localize_identifier (const char *ident)
...
+
+#undef LANG_HOOKS_RUN_LANG_SELFTESTS
+#define LANG_HOOKS_RUN_LANG_SELFTESTS selftest::rust_tests
+
+#include "rust-base62.h"
```
tschwinge, line 443: Please remove that `#include "rust-base62.h"` here.

gcc/c-family/c-common.h
```diff
@@ -1519,6 +1519,8 @@ namespace selftest {
 
   /* The entrypoint for running all of the above tests.  */
   extern void c_family_tests (void);
+  extern void diagnostic_c_tests (void);
+  extern void opt_problem_cc_tests (void);
```
tschwinge, line 1523: Shouldn't these (remain to) be declared elsewhere?

tschwinge (changes requested):
We're getting there!  :-)

.github/workflows/ccpp.yml
CohenArthur: Done! Thanks for the keen eye, I don't know why I thought this was important

gcc/c/c-lang.c
CohenArthur: I've moved and renamed the tests. Thanks

gcc/c-family/c-common.h
CohenArthur: I declared them alongside the other C family tests, and I think it makes more sense now that they're renamed. Please let me know if you were expecting another location :)

CohenArthur:
@tschwinge Thanks a lot for the review and offer to help getting one commit upstream. @dkm I think this commit should definitely be sent upstream and I'll start looking into it!

CohenArthur:
Commits have been fixed-up as talked about during the jitsi call :)

tschwinge (approved):
... and my response simply is:

bors: merge

tschwinge:
> > > As far as i can tell this looks good but I am not sure about the final commit [777d1d2](https://github.com/Rust-GCC/gccrs/commit/777d1d2bc7a65ac8b6bd874f5a2799d57c52a11a) do we need to touch this or can we drop it?
> > 
> > 
> > The last commit is "extremely important" as it prevents our selftests from running some C specific selftests which we obviously fail. This is the commit that fixes what David Malcolm was suggesting over on the gcc mailing list
> 
> Should this be sent upstream instead ? If that was suggested by David, that should be easy to have it merged ?

:tada: This has recently been pushed (commit 981524428b21bf6935f8f9a3b8394d5504f4f0fc "selftest: Move C-specific tests to c_family" -- @CohenArthur's first in GCC upstream!), and I'm working on another #247 merge, and then also resolve any local divergence we may have in these selftest files.  (..., but first need to resolve some other GCC/Rust local breakage re `*.opt` files; stay tuned...)

PR #777, opened 2021-10-28
Tag 'rust_fatal_error' as 'noreturn'

tschwinge:
... like GCC's underlying 'fatal_error'.

By definition, we cannot continue, upon having run into a fatal error.

### DIFF
--- a/gcc/rust/rust-diagnostics.h
+++ b/gcc/rust/rust-diagnostics.h
@@ -57,7 +57,8 @@ rust_warning_at (const Location, int opt, const char *fmt, ...)
   RUST_ATTRIBUTE_GCC_DIAG (3, 4);
 extern void
 rust_fatal_error (const Location, const char *fmt, ...)
-  RUST_ATTRIBUTE_GCC_DIAG (2, 3);
+  RUST_ATTRIBUTE_GCC_DIAG (2, 3)
+  RUST_ATTRIBUTE_NORETURN;
 extern void
 rust_inform (const Location, const char *fmt, ...)
   RUST_ATTRIBUTE_GCC_DIAG (2, 3);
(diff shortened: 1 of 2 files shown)


tschwinge:
Don't like how clang-format thinks this ought to be formatted:  ;-P

bors: merge

philberty:
> Don't like how clang-format thinks this ought to be formatted: ;-P
> 
> bors: merge

No problem lets remove the clang-format check from bors then we can do this if we remove it from here: https://github.com/Rust-GCC/gccrs/blob/8e992e371bbd896c8b605e1ebade5cad80b1b411/.github/bors.toml#L2

philberty:
> > Don't like how clang-format thinks this ought to be formatted: ;-P
> > bors: merge
> 
> No problem lets remove the clang-format check from bors then we can do this if we remove it from here:
> 
> https://github.com/Rust-GCC/gccrs/blob/8e992e371bbd896c8b605e1ebade5cad80b1b411/.github/bors.toml#L2

I don't mind if you want to just go ahead and push that change

CohenArthur:
@philberty @tschwinge isn't this going to cause CI failures for all future pull requests?

tschwinge:
@CohenArthur, per my understanding, clang-format just looks at the local diff.  (After all, the large body of GCC code that we import doesn't generally adhere to GCC/Rust clang-format rules either.)

tschwinge:
@CohenArthur, well that's actually not true... having reverse-engineered what the _Clang Format Lint_ (`.github/workflows/clang-format.yml`) is doing; see #795.

So we'll need a different solution.  For example, configure `clang-format` to allow for the formatting that I'd like to use.

mathstuf:
> So we'll need a different solution. For example, configure clang-format to allow for the formatting that I'd like to use.

Alas, it's like "no warnings" enforcement. Having just one poorly formatted bit makes the viability of doing any kind of enforcement much more difficult. I've learned to live with some of `clang-format`'s…idiosyncrasies if it brings the benefit of being able to end all style guideline wars.

Note that there is the sledgehammer of:

```c++
// clang-format off
some_code_clang_format_mangles ();
// clang-format on
```

Note that `clang-format` will still track `;` usage there and affect things like indentation afterwards in some cases (usually macros not ending with a `;`).

philberty:
I am going to directly merge this outside of bors and raise an issue about the clang-format

PR #779, opened 2021-10-29
'rust_internal_error_at'

tschwinge:
... to be used for internal sanity checks, where in case of failure you'd like to provide/print more information than a plain 'rust_assert'.


philberty (approved):
nice one didn't know this existed

philberty:
@tschwinge are you able to fix up the clang-format error ok?

tschwinge:
Don't like how clang-format thinks this ought to be formatted:  ;-P

bors: merge

PR #780, opened 2021-10-29
No side effects in 'assert' expressions

tschwinge:
Usually, if 'assert'ions are disabled, 'assert' expressions are not evaluated,
so in that case won't effect any side effects.

Via spurious ICEs/test suite FAILs, this may be observed in GCC/Rust, for
example, if configuring with '--enable-checking=no' and disabling a "more
forgiving" 'gcc/system.h:gcc_assert' definition, so that '0 && (EXPR)' gets
used:

     /* Use gcc_assert(EXPR) to test invariants.  */
     #if ENABLE_ASSERT_CHECKING
     #define gcc_assert(EXPR)                                               \
        ((void)(!(EXPR) ? fancy_abort (__FILE__, __LINE__, __FUNCTION__), 0 : 0))
    -#elif (GCC_VERSION >= 4005)
    +#elif (0) //GCC_VERSION >= 4005)
     #define gcc_assert(EXPR)                                               \
       ((void)(__builtin_expect (!(EXPR), 0) ? __builtin_unreachable (), 0 : 0))
     #else
     /* Include EXPR, so that unused variable warnings do not occur.  */
     #define gcc_assert(EXPR) ((void)(0 && (EXPR)))
     #endif

As that one does cause some issues in GCC proper (that I shall fix separately),
may use this change to 'gcc/rust/rust-system.h:rust_assert' instead:

    +#if 0
     #define rust_assert(EXPR) gcc_assert (EXPR)
    +#else
    +#define rust_assert(EXPR) ((void) (0 && (EXPR)))
    +#endif

To fix these, use the same pattern as is already used in a lot of existing
GCC/Rust code:

    bool ok = [expression with side effects];
    rust_assert (ok);

I've only done a quick manual review; maybe there is a tool for doing this?

### DIFF
--- a/gcc/rust/backend/rust-compile-context.h
+++ b/gcc/rust/backend/rust-compile-context.h
@@ -53,11 +53,12 @@ public:
     for (auto it = builtins.begin (); it != builtins.end (); it++)
       {
 	HirId ref;
-	rust_assert (
-	  tyctx->lookup_type_by_node_id ((*it)->get_node_id (), &ref));
+	bool ok = tyctx->lookup_type_by_node_id ((*it)->get_node_id (), &ref);
+	rust_assert (ok);
 
 	TyTy::BaseType *lookup;
-	rust_assert (tyctx->lookup_type (ref, &lookup));
+	ok = tyctx->lookup_type (ref, &lookup);
+	rust_assert (ok);
 
 	Btype *compiled = TyTyCompile::compile (backend, lookup);
 	compiled_type_map.insert (std::pair<HirId, Btype *> (ref, compiled));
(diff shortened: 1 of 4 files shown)


dkm:
Reminds me of https://github.com/Rust-GCC/gccrs/pull/555 :)

philberty (approved):
nice cleanup LGTM

CohenArthur:
@tschwinge I think a tool like [Coccinelle](https://coccinelle.gitlabpages.inria.fr/website/) would be pretty useful here (and might also help you with the gcc_assert macro).

Issue #718, opened 2021-10-06
Get rid of lambda within AST::TypePath


CohenArthur:
https://github.com/Rust-GCC/gccrs/blob/99c28309d3553346d4f0337dbae49f4a8e48da01/gcc/rust/ast/rust-path.h#L758-L766
Lambda's like this need to be removed they make working with the IR's more complex for static analysis. To fix this issue you should be able to remove this lambda helper function and provide a method to return a reference to the values like

```c++
std::vector<TypePathSegment>& get_segments() { return segments; }
```

PR #783, opened 2021-10-31
Get rid of lambda within `AST::TypePath` and provide a method to return a reference

diohabara:
## Related issue

This PR will fix <https://github.com/Rust-GCC/gccrs/issues/718>

- \[x] Run `clang-format`

order to get feedback.

### DIFF
--- a/gcc/rust/hir/rust-ast-lower-type.h
+++ b/gcc/rust/hir/rust-ast-lower-type.h
@@ -62,20 +62,18 @@ public:
   {
     std::vector<std::unique_ptr<HIR::TypePathSegment>> translated_segments;
 
-    path.iterate_segments ([&] (AST::TypePathSegment *seg) mutable -> bool {
-      translated_segment = nullptr;
-      seg->accept_vis (*this);
-      if (translated_segment == nullptr)
-	{
-	  rust_fatal_error (seg->get_locus (),
-			    "failed to translate AST TypePathSegment");
-	  return false;
-	}
-
-      translated_segments.push_back (
-	std::unique_ptr<HIR::TypePathSegment> (translated_segment));
-      return true;
-    });
+    for (auto &seg : path.get_segments ())
+      {
+	translated_segment = nullptr;
+	seg->accept_vis (*this);
+	if (translated_segment == nullptr)
+	  {
+	    rust_fatal_error (seg->get_locus (),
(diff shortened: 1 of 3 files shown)


gcc/rust/hir/rust-ast-lower-type.h
```diff
@@ -62,20 +62,18 @@ class ASTLowerTypePath : public ASTLoweringBase
...
-	std::unique_ptr<HIR::TypePathSegment> (translated_segment));
-      return true;
-    });
+    for (auto &seg : path.get_segments ())
+      {
+	translated_segment = nullptr;
+	seg->accept_vis (*this);
+	if (translated_segment == nullptr)
+	  {
+	    rust_fatal_error (seg->get_locus (),
```
philberty, line 71: This is not related to your PR but lets raise an issue about this fatal_error it should be an assertion

gcc/rust/hir/rust-ast-lower.cc
```diff
@@ -490,20 +490,18 @@ ASTLowerQualifiedPathInType::visit (AST::QualifiedPathInType &path)
...
-      std::unique_ptr<HIR::TypePathSegment> (translated_segment));
-    return true;
-  });
+  for (auto &seg : path.get_segments ())
+    {
+      translated_segment = nullptr;
+      seg->accept_vis (*this);
+      if (translated_segment == nullptr)
+	{
+	  rust_fatal_error (seg->get_locus (),
```
philberty, line 499: Same here

philberty (changes requested):
LGTM you just need to squash these fixup commits:

$ git rebase -i HEAD~3

Is what i use to do it.

gcc/rust/hir/rust-ast-lower-type.h
diohabara: Thanks for you quick review!

Are you saying I should replace `rust_fatal_error` with another function or another procedure?

philberty: Not as part of this pr. I raised a separate issue for that.

tschwinge (changes requested):
Please again force-push after updating the commit log (`git commit --amend`) into standard format (see `git log`), and then we're good to go.

philberty (approved):
LGTM thanks again! :D nice job

Issue #645, opened 2021-08-24
Modules without module files


flip1995:
When looking at this [rustc test case](https://github.com/rust-lang/rust/tree/master/src/test/ui/invalid-module-declaration) I learned that modules can exist and be used without a `mod.rs` file.

Consider this crate structure:

```
├── aux
│   └── foo
│       └── mod.rs
└── main.rs
```
_NOTE: There is no `aux/mod.rs` file!_

and this file content:

```rs
// aux/foo/mod.rs
pub fn f() -> i32 { 42 }
```

```rs
// main.rs
mod aux {
    pub mod foo;
}

fn main() {
    println!("{}", aux::foo::f());
}
```

When run this with `cargo run` this will compile and output `42` (as expected!? :upside_down_face:)

If you want to try it out, just checkout this repo: https://github.com/flip1995/mod_without_mod

cc @CohenArthur  since you're working on compiling multiple files.

CohenArthur:
Wow, thanks a lot for finding that out. I'll look into rustc to see how this works. This is definitely an interesting test case I think

mathstuf:
So I started looking into this. My hacks involved tracking module scopes in the `Lexer` (almost certainly not the right place) and `Module` itself. However, it is unknown how to strip enough components from the including path. Now that I've written this, I think I may have another idea…testing that out.

In any case, other interesting cases to consider:

```rust
mod dont_be_lazy {
    mod middle {
        mod inner; // naively searching by trying to incrementally add the scopes and search will find…
    }
}

mod inner; // …this module file
```

```rust
// Referred to via #![path = "some_other_directory/file.cxx"]
mod middle {
    mod inner; // Where is this looked up (need to check with `rustc`)?
}
```

Issue #676, opened 2021-09-15
Bad warnings


philberty:
I tried this code: https://godbolt.org/z/nTj8qfqTx

```rust
pub struct S(u64, u64);

impl S {
    pub fn sum(&self) -> u64 {
        self.0 + self.1
    }
}

pub fn f() -> u64 {
    S(123, 42).sum()
}

```

I expected to see this happen: no warnings produced

Instead, this happened:

```
<source>:1:5: warning: struct is never constructed: 'S'
    1 | pub struct S(u64, u64);
      |     ^
<source>:4:9: warning: associated function is never used: 'sum'
    4 |     pub fn sum(&self) -> u64 {
      |         ^
<source>:9:5: warning: function is never used: 'f'
    9 | pub fn f() -> u64 {
      |     ^
<source>:9:5: warning: unused name 'f'
Compiler returned: 0
```

### Meta

- What version of Rust GCC were you using, git sha if possible. c644ee4c4351e3590f5396e94ec24ad7b828a954

Issue #665, opened 2021-09-10
Attributes Structure


philberty:
I wonder is it worth checking in Rustc to see if their Attributes structure cleans this up for you in the parser so you get a structure like:

string attr_name
bool has_rhs
string get_rhs

_Originally posted by @philberty in https://github.com/Rust-GCC/gccrs/pull/664#discussion_r706023945_

CohenArthur:
I think this part should live in its own header + source files. We can then add attribute-specific functions such as the one to handle #[path] as written by @mathstuf in #785

mathstuf:
I'll note that it seems that `rustc` validates built-in attributes even if they aren't used. For example:

```rust
#[path = "foo"]
mod blah {
    #![path] // `rustc` errors even though it doesn't matter (the outer one "wins")
}
```

philberty:
this seems too old to be useful now

Issue #635, opened 2021-08-16
Out of range float casts miscompilation


bjorn3:
I tried this code:

```rust
pub fn foo() -> f32 {
    0xffff_ffff_ffff_ffff_ffff_ffff_ffff_ffffu128 as f32
}
```

I expected to see this happen: Saturating to `f32::INFINITY`

Instead, this happened: 9223372000000000000.0 is returned

Another case is

```rust
pub fn foo() -> f32 {
    1<an absolutely huge amount of zeros>u128 as f32
}
```

which should give an out of range error, but returns 0.0.

### Meta

- What version of Rust GCC were you using, git sha if possible: The current version at godbolt.

philberty:
I think the code to check for this is in the GCC TREE_OVERFLOW flags

philberty:
@bjorn3 I have a branch of code to fix this but the part about Saturating to f32::INFINITY I might leave out for the first PR, I have code to catch this case but I wonder is there documentation that I can reference on this saturating to infinity?

bjorn3:
This is the PR which originallt changed it from UB to saturating: https://github.com/rust-lang/rust/pull/71269

PR #785, opened 2021-11-01
rust: track inline module scopes for module file resolution

mathstuf:
The set of inline modules is required to find the expected location of a
module file. Track this information with an RAII object
(`InlineModuleStackScope`) and pass it down to any out-of-line modules
so that, when requested, the set of inline modules can be added to the
search path.

---
Note that this does not include a test case because I have no idea how to mark up this for all the warnings that come out (#676):

```diff
diff --git a/gcc/testsuite/rust/compile/missing_middle/sub/mod.rs b/gcc/testsuite/rust/compile/missing_middle/sub/mod.rs
new file mode 100644
index 00000000000..f099d61e04a
--- /dev/null
+++ b/gcc/testsuite/rust/compile/missing_middle/sub/mod.rs
@@ -0,0 +1,3 @@
+pub fn f() -> u32 {
+    1
+}
diff --git a/gcc/testsuite/rust/compile/mod_missing_middle.rs b/gcc/testsuite/rust/compile/mod_missing_middle.rs
new file mode 100644
index 00000000000..d9cdf0a54f1
--- /dev/null
+++ b/gcc/testsuite/rust/compile/mod_missing_middle.rs
@@ -0,0 +1,11 @@
+pub mod missing_middle {
+    pub mod sub;
+}
```

Observed warnings:

```
/home/boeckb/code/depot/group-compilers/gcc/src-gccrs/gcc/testsuite/rust/compile/missing_middle/sub/mod.rs:1:5: warning: unused name 'f'
/home/boeckb/code/depot/group-compilers/gcc/src-gccrs/gcc/testsuite/rust/compile/mod_missing_middle.rs:2:9: warning: unused name 'sub'
/home/boeckb/code/depot/group-compilers/gcc/src-gccrs/gcc/testsuite/rust/compile/missing_middle/sub/mod.rs:1:5: warning: unused name 'sub::f'
/home/boeckb/code/depot/group-compilers/gcc/src-gccrs/gcc/testsuite/rust/compile/mod_missing_middle.rs:2:9: warning: unused name 'missing_middle::sub'
/home/boeckb/code/depot/group-compilers/gcc/src-gccrs/gcc/testsuite/rust/compile/mod_missing_middle.rs:1:5: warning: unused name 'missing_middle'
/home/boeckb/code/depot/group-compilers/gcc/src-gccrs/gcc/testsuite/rust/compile/missing_middle/sub/mod.rs:1:5: warning: unused name 'missing_middle::sub::f'
```

Fixes: #645 

- \[x] Run `clang-format` (leaving to CI)
- \[x] Added any relevant test cases to `gcc/testsuite/rust/` (see above)


gcc/rust/ast/rust-item.h
```diff
@@ -1013,11 +1015,12 @@ class Module : public VisItem
...
   Module (Identifier module_name, Visibility visibility,
 	  std::vector<Attribute> outer_attrs, Location locus,
-	  std::string outer_filename)
+	  std::string outer_filename, std::vector<std::string> module_scope)
     : VisItem (std::move (visibility), std::move (outer_attrs)),
       module_name (module_name), locus (locus), kind (ModuleKind::UNLOADED),
       outer_filename (outer_filename), inner_attrs (std::vector<Attribute> ()),
-      items (std::vector<std::unique_ptr<Item>> ())
+      items (std::vector<std::unique_ptr<Item>> ()),
+      module_scope (std::move (module_scope))
```
CohenArthur, line 1023: Isn't it undefined behavior to move the parser's `inline_module_stack` vector every time we create a new Module? I'm not super clear on that aspect of C++, but I don't know what moving the vector and then emplacing_back a new name will produce.

I think it would be really nice to have a test proceeding to multiple module scope inclusions for this. Something like
```rust
mod foo {
    mod bar {
        mod baz;
    }
}

mod fooz {
    mod barz;
}
```
and see what asan/valgrind tells us. 

I might be worrying over nothing and it's quite late so feel free to disregard :)

CohenArthur (approved):
Thanks :D I think the PR needs some tests before it gets merged, I can take a look at the warning thingies

CohenArthur:
Ah, @mathstuf: You can add the following to the test cases for now to remove (unused) warnings:
```rust
mod modules;

fn main() {
    let twelve = modules::return_12();
}
```
the issues you are seeing are because of name resolution. I faced them too when implementing the little bits for module inclusion

gcc/rust/ast/rust-item.h
mathstuf: It is copied in at the call site (as needed). It is needed here, so it is not moved then. I'll add this test case (as well as some `#[path]` ones).

CohenArthur: Oh gotcha. Thanks!

mathstuf:
OK, so something isn't quite right with `#[path]` yet, but I'll look at that later.

mathstuf:
Gah. Mutable local variables bite again. `#[path]` issues resolved.

gcc/rust/ast/rust-ast-full-test.cc
```diff
@@ -4093,8 +4101,7 @@ Module::process_file_path ()
   if (no_candidates_found || multiple_candidates_found)
     return;
 
-  module_file = file_mod_found ? expected_file_path
-			       : current_directory_name + expected_dir_path;
+  module_file = std::move (file_mod_found ? file_mod_path : dir_mod_path);
```
mathstuf, line 4104: Now this more accurately represents what was actually tested above with `file_exists` :) .

philberty:
I don't think this PR relates to the #635 I think its this one you mean https://github.com/Rust-GCC/gccrs/issues/645

gcc/rust/parse/rust-parse.h
```diff
@@ -622,6 +622,21 @@ template <typename ManagedTokenSource> class Parser
...
+
+  public:
+    InlineModuleStackScope (Parser &parser, std::string name) : parser (parser)
+    {
+      parser.inline_module_stack.emplace_back (std::move (name));
+    }
+    ~InlineModuleStackScope () { parser.inline_module_stack.pop_back (); }
+  };
+  // The names of inline modules while parsing.
+  std::vector<std::string> inline_module_stack;
```
philberty, line 639: can we make a private field?

philberty (changes requested):
Nice work, I think it looks good but I would like to see any new classes have private fields if possible its not always the case so let me know.

gcc/rust/parse/rust-parse.h
mathstuf: This is already a private field.

CohenArthur (approved):

gcc/rust/parse/rust-parse.h
philberty: Ah ok sorry, I was misreading this didn't know the '};', can we keep those private fields together and the private class below? It makes it easier to read for me.

mathstuf:
OK, so I've updated to support `#[path = "…"] mod inline {}` as well. The test suite needed an update.

I'll also note that it seems that inline modules don't support inner attributes.

gcc/testsuite/rust/compile/mod_missing_middle.rs
```diff
@@ -0,0 +1,30 @@
...
+
+#[path = "missing_middle"]
+mod with_outer_path_attr {
+    #[path = "outer_path.rs"]
+    mod inner;
+}
+
+// FIXME: the parser doesn't accept inner attributes for inline modules.
+// mod with_inner_path_attr {
+//     #![path = "missing_middle"]
```
philberty, line 18: Can you raise an issue for this?

philberty (approved):
LGTM

gcc/testsuite/rust/compile/mod_missing_middle.rs
mathstuf: #790

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -2078,6 +2078,79 @@ Parser<ManagedTokenSource>::parse_visibility ()
     }
 }
 
+static std::string
+extract_module_path (const AST::AttrVec &inner_attrs,
+		     const AST::AttrVec &outer_attrs, const std::string &name)
+{
```
CohenArthur, line 2084: The current implementation of this function is in `gcc/rust/ast/rust-ast-full-test.cc`. Any reason for copying it here but no removing/renaming it from the other source?

CohenArthur (changes requested):
Apart from the copied function I think this is all good :D Thanks a lot!

gcc/rust/parse/rust-parse-impl.h
mathstuf: This one deals with inner attributes as well. The other one only deals with out-of-line modules, so there's no chance of an inner attribute affecting its path (since we're already reading the file).

mathstuf: Also, the return value is different. This is "the name to use for the directory to look up submodules in". The other returns (effectively) `Option<String>` for the path to use for the module itself (with an empty string being `None`).

CohenArthur: Mmh. I understand the idea but there is a lot of code present in both functions. In fact, I can't really review what changed in this function versus the previously written one. Do you think you could factor them together? I'm not a fan of how a lot of the logic is repeated. We also have issues regarding cleaning up attribute handling such as https://github.com/Rust-GCC/gccrs/issues/665 so I think it's fine if factoring them together isn't perfect as it will need cleaning up anyway

CohenArthur (approved):
Thanks! LGTM!

Issue #634, opened 2021-08-16
Save.log file for every testsuite run in GHA


dkm:
Having the log file for GHA testsuite run would help when it fails.

PR #787, opened 2021-11-01
gha: save `check-rust` results as artifacts

mathstuf:
They will stick around for 90 days (as per GitHub default limits).

     https://docs.github.com/en/actions/advanced-guides/storing-workflow-data-as-artifacts

Fixes: #634

### DIFF
--- a/.github/workflows/bootstrap.yml
+++ b/.github/workflows/bootstrap.yml
@@ -47,6 +47,12 @@ jobs:
       run: |
            cd gccrs-build; \
            make check-rust
+    - name: Archive check-rust results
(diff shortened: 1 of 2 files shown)


mathstuf:
Artifacts are visible on [this page](https://github.com/Rust-GCC/gccrs/actions/runs/1409994782). It can be found by clicking on the "Summary" in the Check Suite page and scrolling to the bottom. Not sure how to fix the name of the archive yet…

mathstuf:
Found it; was some copy pasta.

philberty (approved):
I'm not sure we need to archive results for bootstrap builds but lets do it anyway :D

Issue #627, opened 2021-08-12
Need option to specify crate name


philberty:
Currently, the crate name is hardcoded over here:

https://github.com/Rust-GCC/gccrs/blob/7be0232c686a75f98b2ca3c27f7de3139b8999c6/gcc/rust/rust-session-manager.cc#L75

I think there should be some kind of default name if it is not specified. But you can add new command line options by adding to this file:

https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/lang.opt

And you can handle the option over in: https://github.com/Rust-GCC/gccrs/blob/7be0232c686a75f98b2ca3c27f7de3139b8999c6/gcc/rust/rust-session-manager.cc#L346

PR #788, opened 2021-11-02
gccrs: add the `-frust-crate=` option to set the crate name

mathstuf:

Fixes: #627

### DIFF
--- a/gcc/rust/lang.opt
+++ b/gcc/rust/lang.opt
@@ -34,6 +34,10 @@ L
 Rust Joined Separate
 ; Not documented
 
+frust-crate=
+Rust Joined RejectNegative
+-frust-crate=<name>             Set the crate name for the compilation
+
 frust-debug
 Rust Var(flag_rust_debug)
 Dump various Rust front end internals.
--- a/gcc/rust/rust-session-manager.cc
+++ b/gcc/rust/rust-session-manager.cc
@@ -70,8 +70,6 @@ const char *kHIRDumpFile = "gccrs.hir.dump";
 const char *kHIRTypeResolutionDumpFile = "gccrs.type-resolution.dump";
 const char *kTargetOptionsDumpFile = "gccrs.target-options.dump";
 
-// FIME in the imports/visibility milestone - this needs to be command line
-// option
 const std::string kDefaultCrateName = "TestCrate";
 
 // Implicitly enable a target_feature (and recursively enable dependencies).
@@ -330,6 +328,9 @@ Session::init ()
   // setup backend to GCC GIMPLE
   backend = rust_get_backend ();
 
+  // set the default crate name
+  options.set_crate_name (kDefaultCrateName);
+
   // the constant folder uses gcc
   ConstFold::Context::init (backend);
 }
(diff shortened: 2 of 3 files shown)


gcc/rust/rust-session-manager.cc
```diff
@@ -330,6 +328,9 @@ Session::init ()
   // setup backend to GCC GIMPLE
   backend = rust_get_backend ();
 
+  // set the default crate name
+  options.set_crate_name (kDefaultCrateName);
```
bjorn3, line 332: The default in rustc is the source filename without `.rs` or `rust_out` if the source is stdin.

philberty: Lets raise an issue to fix that default crate name, I think it makes sense to fix that outside of this PR

gcc/rust/lang.opt
```diff
@@ -34,6 +34,10 @@ L
 Rust Joined Separate
 ; Not documented
 
+frust-crate=
+Rust Joined RejectNegative
+-frust-crate=<naem>             Set the crate name for the compilation
```
philberty, line 39: typo 'name'

gcc/rust/rust-session-manager.h
```diff
@@ -203,6 +204,13 @@ struct CompileOptions
     enable_dump_option (DumpOption::HIR_DUMP);
     enable_dump_option (DumpOption::TYPE_RESOLUTION_DUMP);
   }
+
+  bool set_crate_name (std::string name)
```
philberty, line 208: is there a getter for this? Not sure if we need a move here. I would have either done:

```
bool set_crate_name (std::string&& name) { crate_name = std::move(name); }
```

Or copy

```
void set_crate_name(const std::string& name) { crate_name.assign(name); }
```

philberty (changes requested):
I think after fixing that typo its good to go

philberty:
Also as you pointer out until we get the unit-test framework in there are no tests to add.

gcc/rust/rust-session-manager.h
mathstuf: It is public already (`CompileOptions` is a `struct`).

gcc/rust/rust-session-manager.cc
mathstuf: Another question: what if the source doesn't have a `.rs` extension?

bjorn3: https://github.com/rust-lang/rust/blob/758c00ea4088e1ca2714890f00c791c24f001536/compiler/rustc_session/src/output.rs#L48

Rustc first tries `--crate-name` and `#![crate_name]` (if both are used they must match), then tries to take the filename, take the file stem (filename with last extension removed if any. `a.tar.gz` would result in `a.tar` for the file stem, which is not a valid crate name) and replace `-` with `_`. And finally falls back to `rust_out` if the input is not a file. The file extension doesn't matter.

philberty (approved):

gcc/rust/rust-session-manager.cc
philberty: We still need to figure out how we can handle ```#![crate_name]``` as we don't have a way to feed this back to the compiler driver yet to auto set the output name.

mathstuf: See #789.

PR #791, opened 2021-11-02
Replace TyTy::TupleType::iterate_fields with a new function get_fields

### LINKED ISSUE #735: Get rid of iterate_fields lambda in TyTy::TupleType
Just like the other lambda's, this is a bad way to do static analysis on the parameters.  https://github.com/Rust-GCC/gccrs/blob/3057d98a4e3fc777b06e51cd678723f5c5b4c898/gcc/rust/typecheck/rust-tyty.h#L590

The problem here is that for function pointers there can be type indirection so we wrap up the parameters within TyVar's so it forces a lookup based on the context. To fix this issue:

- [ ] Remove iterate_fields function
- [ ] Create new get_fields function

This get params function should either:

1. Simply return the std::vector<TyVar>& fields
2. Or we could create a new std::vector or TyTy::BaseType* and construct it with field[idx].get_tyty().

I am not sure which is the cleanest solution right now I think the 2nd one fits in best.

cls:
This PR replaces TyTy::TupleType::iterate_fields with a new function get_fields.

In #735, @philberty mentions two possible implementations of get_fields, one that returns `std::vector<TyVar>&` and one that returns a new `std::vector<BaseType *>`. I've included two commits, the first doing the former and second going on to do the latter. I'm happy with one or both being merged.

This is my first PR, so please let me know if I've missed anything! Thanks.

Fixes #735


philberty (changes requested):
Nice work lets go with the first approach returning the TvVar wrapper it looks tidy and offloads the responsibility for error handling onto the caller. As you using the DCO to sign off your commit because you will need to add the appropriate signed-off message into your commit.

Nice work.

philberty (approved):
nice one :D Thanks for doing the experimentation for me by doing both approaches. Sometimes it is hard to tell which is the right path until you see both.

PR #792, opened 2021-11-03
Change DefId type from uint64_t to be struct

### LINKED ISSUE #439: Change DefId to be a struct
Haven't look at all involved types, but would it make sense to have a struct/union for this?
If both cratenum and localdefid are uint32 this change looks correct :)
_Originally posted by @dkm in https://github.com/Rust-GCC/gccrs/issues/438#issuecomment-845937719_

See Rustc: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/def_id/struct.DefId.html

- [ ] Create Struct for DefID over here: https://github.com/Rust-GCC/gccrs/blob/28f527c9598339cf834a30b5ee1f14258b8ecbb2/gcc/rust/util/rust-mapping-common.h#L35
- [ ] Remove bitmask code https://github.com/Rust-GCC/gccrs/blob/28f527c9598339cf834a30b5ee1f14258b8ecbb2/gcc/rust/util/rust-mapping-common.h#L37
- [ ] Update the mappings code bitmask to return a struct https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/util/rust-hir-map.cc#L57-L67
- [ ] Update fntype code to assert using struct instead of bitmask https://github.com/Rust-GCC/gccrs/blob/98359f20cd2d9268df582ea1ee289e0ea69efcb5/gcc/rust/typecheck/rust-tyty.h#L1116

npate012:
DefId was uint64_t previously but it has been changed to be a struct.
In order to reduce code breakage, ==, !=, and < operators have been implemented
for DefId. Since DefId is now a proper struct, bit manipulation code has been
removed with member accesses.

Fixes #439

Signed-off-by: Nirmal Patel <npate012@gmail.com>

### DIFF
--- a/gcc/rust/typecheck/rust-tyty.h
+++ b/gcc/rust/typecheck/rust-tyty.h
@@ -1297,7 +1297,7 @@ public:
       params (std::move (params)), type (type), flags (flags),
       identifier (identifier), id (id), abi (abi)
   {
-    LocalDefId local_def_id = id & DEF_ID_LOCAL_DEF_MASK;
+    LocalDefId local_def_id = id.localDefId;
     rust_assert (local_def_id != UNKNOWN_LOCAL_DEFID);
   }
 
@@ -1312,7 +1312,7 @@ public:
       params (params), type (type), flags (flags), identifier (identifier),
       id (id), abi (abi)
   {
-    LocalDefId local_def_id = id & DEF_ID_LOCAL_DEF_MASK;
+    LocalDefId local_def_id = id.localDefId;
     rust_assert (local_def_id != UNKNOWN_LOCAL_DEFID);
   }
 
@@ -1471,7 +1471,7 @@ public:
       parameter_types (std::move (parameter_types)),
       result_type (std::move (result_type)), id (id)
   {
-    LocalDefId local_def_id = id & DEF_ID_LOCAL_DEF_MASK;
+    LocalDefId local_def_id = id.localDefId;
     rust_assert (local_def_id != UNKNOWN_LOCAL_DEFID);
   }
 
@@ -1485,7 +1485,7 @@ public:
       parameter_types (std::move (parameter_types)),
       result_type (std::move (result_type)), id (id)
   {
-    LocalDefId local_def_id = id & DEF_ID_LOCAL_DEF_MASK;
+    LocalDefId local_def_id = id.localDefId;
     rust_assert (local_def_id != UNKNOWN_LOCAL_DEFID);
   }
 
--- a/gcc/rust/util/rust-mapping-common.h
+++ b/gcc/rust/util/rust-mapping-common.h
@@ -31,17 +31,32 @@ typedef uint32_t NodeId;
 typedef uint32_t HirId;
 // refers to any top-level decl in HIR
 typedef uint32_t LocalDefId;
-// refers to <Crate><DefId>
-typedef uint64_t DefId;
 
-#define DEF_ID_CRATE_MASK 0xFFFFFFFF00000000
-#define DEF_ID_LOCAL_DEF_MASK 0x00000000FFFFFFFF
+struct DefId
+{
+  CrateNum crateNum;
+  LocalDefId localDefId;
+
+  bool operator== (const DefId &other) const
+  {
+    return this->crateNum == other.crateNum
+	   && this->localDefId == other.localDefId;
+  }
+
+  bool operator!= (const DefId &other) const { return !(*this == other); }
+
+  bool operator< (const DefId &other) const
+  {
+    return ((uint64_t) this->crateNum << 32 | this->localDefId)
+	   < ((uint64_t) other.crateNum << 32 | other.localDefId);
+  }
+};
 
 #define UNKNOWN_CREATENUM ((uint32_t) (0))
 #define UNKNOWN_NODEID ((uint32_t) (0))
 #define UNKNOWN_HIRID ((uint32_t) (0))
 #define UNKNOWN_LOCAL_DEFID ((uint32_t) (0))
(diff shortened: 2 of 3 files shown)


gcc/rust/util/rust-mapping-common.h
```diff
@@ -32,16 +32,31 @@ typedef uint32_t HirId;
...
+	   && this->localDefId == other.localDefId;
+  }
+
+  bool operator!= (const DefId &other) const { return !(*this == other); }
+
+  bool operator< (const DefId &other) const
+  {
+    return ((uint64_t) this->crateNum << 32 | this->localDefId)
+	   < ((uint64_t) other.crateNum << 32 | other.localDefId);
+  }
```
CohenArthur, line 52: I personally do not like operator overloading as I find it hinders code clarity. I'm not a C++ expert so if everyone else agrees with the change, I'm fine with it. I agree that removing the bit shifts and masks is a good thing, but that could also be solved by adding methods

philberty: I know what you mean @CohenArthur I think we need it for some classes like this one and some others in the name-resolver iirc, since the code at the moment uses a uin64_t and we are using a lot of != and == on it it would be a fairly big refactor I think to change that over.

philberty: OVerall I think this is ok to use operator overloading at least for now.

npate012: @CohenArthur and @philberty  Thank you for the feedback. I just realized that I can get rid of the < operator if I use a lambda comparator when declaring the map like map<_, _, decltype(comparator)> in the following line:

https://github.com/Rust-GCC/gccrs/blob/5f0df4812c37fc428b5508e019e9fb7f8a7b77b1/gcc/rust/backend/rust-compile-context.h#L334

I am also attempt getting rid of other operators and do refactor if that is a good approach. I can certainly work on this if it makes sense.

philberty: Ah ok in that case I would prefer that this is defined in the class or you can define the comparator operators in a separate class instead of another lambda.

npate012: @philberty and @CohenArthur I have removed operator overloading and added a new class called DefIdComparator. This class has static methods to compare DefId structs. I also refactored code so it does not break.

Not sure if you think the code looks better. I kinda feel the code is a little more complicated now. So I can add operator overloading back if you would like or we can keep the code the way it is.

Once you like the code, I will squash the commits to 1 commit.

Thank you.

gcc/rust/typecheck/rust-tyty.cc
```diff
@@ -53,8 +55,9 @@ BaseType::satisfies_bound (const TypeBoundPredicate &predicate) const
   for (auto &b : probed)
     {
       const Resolver::TraitReference *bound = b.first;
-      bool found = bound->get_mappings ().get_defid ()
-		   == query->get_mappings ().get_defid ();
+      bool found
+	= DefIdComparator::equals (bound->get_mappings ().get_defid (),
+				   query->get_mappings ().get_defid ());
```
mathstuf, line 60: Is the indentation here bad or is it just GitHub not knowing how to render tabulator characters properly? Assuming the latter given the rendering of `return true;` below.

gcc/rust/typecheck/rust-tyty.h
```diff
@@ -1485,7 +1485,7 @@ class ClosureType : public BaseType, public SubstitutionRef
       parameter_types (std::move (parameter_types)),
       result_type (std::move (result_type)), id (id)
   {
-    LocalDefId local_def_id = id & DEF_ID_LOCAL_DEF_MASK;
+    LocalDefId local_def_id = id.localDefId;
     rust_assert (local_def_id != UNKNOWN_LOCAL_DEFID);
```
mathstuf, line 1489: This is more likely a separate PR for the future (as I suspect other constants could use similar help), but having this be a `LocalDefId::Unknown` class-level constant rather than a `#define` or `enum` would help with readability IMO.

mathstuf (approved):

gcc/rust/typecheck/rust-tyty.cc
npate012: I think the indentation is ok and clang-format did not show any issues. Most likely the tab character rendering is weird. There are 4 space in the beginning of line 58 but 1 tab in the beginning of line 59.

gcc/rust/typecheck/rust-tyty.h
philberty: yep this is a good-first-pr i have been meaning to create

philberty (changes requested):
Thanks for trying this out @npate012 but lets stick with your original implementation.

npate012:
@philberty I removed the second commit so the pull request is back to the original implementation. I am going to change its status from draft to ready for review. Thank you.

gcc/rust/util/rust-mapping-common.h
```diff
@@ -32,16 +32,31 @@ typedef uint32_t HirId;
 // refers to any top-level decl in HIR
 typedef uint32_t LocalDefId;
 // refers to <Crate><DefId>
```
mathstuf, line 34: Is this comment still accurate?

npate012: Thanks. Removed line 34 comment. I am assuming line 32 comment is good because I did not make any changes to that part.

CohenArthur (approved):

mathstuf (approved):

CohenArthur:
Woops, forgot about bors... Sorry everyone

dkm:
Did the same once... :D

PR #514, opened 2021-06-21
Rename tuple field names

tromey:
Tuples currently use plain integer field names in the DWARF.  However,
rustc prefixes these with "__", and GDB understands this convention.

This patch changes gccrs to match, and adds an explanatory comment.

PR #795, opened 2021-11-05
Running `clang-format` locally [#242]

tschwinge:
Apparently everyone but me knows how to do that; I'm now documenting it for my
own future use.

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -13,6 +13,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    # If updating these steps, please also correspondingly update '../../CONTRIBUTING.md', "Running `clang-format` locally".
     - uses: actions/checkout@v2
     - name: Copy .clang-format file
       run: cp contrib/clang-format .clang-format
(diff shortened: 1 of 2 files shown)


mathstuf (approved):
Changes look good. Not happy with this way of enforcing formatting, but I'll bring that up elsewhere :) .

philberty (approved):

philberty:
I think @dkm uses git clang-format is that right?

dkm:
November 5, 2021 1:18 PM, "Philip Herron" ***@***.***> wrote:

> I think @dkm uses git clang-format is that right?
>

Yes, and I also said I would document this... So, let me commit to that and add some more details here. You can still merge that and I'll add more change later this weekend :)

tschwinge:
Yes, that's what @dkm said in <https://github.com/Rust-GCC/gccrs/pull/514#issuecomment-865681557>, <https://github.com/Rust-GCC/gccrs/pull/725#issuecomment-938032551>, for example.

`git clang-format` (as well as the manual-use instructions in GCC `contrib/clang-format`) work on patches; GCC/Rust `.github/workflows/clang-format.yml` (and my new manual-use instructions) works on the actual `gcc/rust/` files.

tschwinge:
And then we also have @philberty's Emacs magic: <https://github.com/Rust-GCC/gccrs/pull/514#issuecomment-865323008>, which may also be worth documenting.

tschwinge:
Relatedly, see also discussion in <https://github.com/Rust-GCC/gccrs/pull/777#issuecomment-961745971> and following, and <https://gcc-rust.zulipchat.com/#narrow/stream/266897-general/topic/clang-format.20strategy>.

PR #800, opened 2021-11-06
Added Remark markdown automation 

### LINKED ISSUE #487: Add remark markdown automation
It will be good practice down the line to have some mark-down documentation within the repo for specific areas within the compiler so we are not locked into things like github. Lets add remark lint automation for our PR's  we can use this as an example: https://github.com/rust-lang/rust-clippy/blob/master/.github/workflows/remark.yml

mvvsmk:
Fixes #487 
Remark.yml contains the workflow, and .remarkrc is the linter's config file. The configurations follow the same ones as shown in the reference repo specified by @philberty in the issue.

The linter works great, and the Test is shown as unsuccessful if a check fails. I also created a PR ( [here](https://github.com/mvvsmk/gccrs/pull/1) )in my fork to test this.  You can check the results out [here.](https://github.com/mvvsmk/gccrs/runs/4126281781?check_suite_focus=true) The test file's name is TEST.md .

Please note that the run mentioned above also shows errors in CONTRIBUTING.md and README.md, which were not part of the commit, but the linter goes over them regardless, as they are part of the Repository.

### The issues that still remain are :

- There are two files CONTRIBUTING.md and README.md which have a few errors, I could configure the linter to skip them, but it would be much easier to correct them.
- The trigger branch for the workflow was not specified. Hence I use master.
- Custom configurations/changes you would like me to add.

This is my first PR, so please let me know if I've missed anything! Thanks.

See: #487 

Signed-off-by: M V V S Manoj Kumar <mvvsmanojkumar@gmail.com>

### DIFF
--- a/.github/workflows/Remark.yml
+++ b/.github/workflows/Remark.yml
@@ -1,0 +1,31 @@
+name: Remark
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches: 
+      - master
+
+jobs:
+  remark:
+    runs-on: ubuntu-latest
+
+    steps:
+    # Setup
+    - name: Checkout
+      uses: actions/checkout@v2.3.3
+
+    - name: Setup Node.js
+      uses: actions/setup-node@v1.4.4
+      with:
+        node-version: '12.x'
+
+    - name: Install remark
+      run: npm install remark-cli remark-lint remark-lint-maximum-line-length remark-preset-lint-recommended remark-gfm
+
+    # Run
+    - name: Check *.md files
+      run: git ls-files -z *.md | xargs -0 -n 1 npx remark -u lint -f > /dev/null
+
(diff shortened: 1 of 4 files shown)


CohenArthur (approved):
Wow, that's super cool. I didn't know markdown linters existed. Thanks for working on those. I guess we could fix the issues in a later PR or in this one, I don't know how everyone feels about it

philberty (changes requested):
Thanks for doing this if you rebase your PR against master it will make the clang-format errors go away. It would also be nice if his PR could include fixes for the mark-down lint errors too to keep everything green upon merging.

mvvsmk:
Sure, I'll get on the changes ASAP, and eddit the .md files so that the linter doesn't show any errors when I make the pr.

philberty (approved):
Awesome work Thanks. LGTM

PR #803, opened 2021-11-09
Partially disable clang-format in rust-diagnostics.h

dkm:
clang-format is confused by attribute macros and does a very bad job at
indenting the code.

Changes merged in #779 make all clang-format checks to fail.

### DIFF
--- a/gcc/rust/rust-diagnostics.h
+++ b/gcc/rust/rust-diagnostics.h
@@ -48,6 +48,7 @@
 // All other format specifiers are as defined by 'sprintf'. The final resulting
 // message is then sent to the back end via rust_be_error_at/rust_be_warning_at.
 
+// clang-format off
 // simple location
 extern void
 rust_internal_error_at (const Location, const char *fmt, ...)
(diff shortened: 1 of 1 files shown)


CohenArthur (approved):
I think this is a nice way of handling the issue, at least for now. I don't know how these comments are gonna work when upstreaming however

philberty:
Nice job for this. I am really torn on this issue because I prefer just enforcing clang-format but want to allow for experimentation.

philberty (approved):

dkm:
> Nice job for this. I am really torn on this issue because I prefer just enforcing clang-format but want to allow for experimentation.

Yes, but as we've merged code that is rejected by `clang-format`, it always fail... This change does not prevent `clang-format` to be blocking again :)

dkm:
bors:merge

dkm:
bors r+ 
?

tschwinge:
Well, I'm not particularly excited about such noisy, manual markup -- but I also won't lose any sleep over it.  I can't tell if I dislike more the manual markup vs. just letting `clang-format` mis-format the code.  ;-)  But, at least, we're now again `clang-format`-clean.  ;-)

dkm:
If we can make sure that the CI run only checks the patch (and correctly) we can remove these ugly comment.
At least now, new PR will have some hints on the indentation (and not always fail). It's a temporary workaround until something is really decided... At least that's how I see it. We could also simply remove clang-format from the GHA, or keep the failed status and go check if there are other errors.

Issue #412, opened 2021-05-07
GCC Abstractions


philberty:
The front-end was bootstrapped from GCCGO which abstracts GCC via backend.c. This really helped bootstrap the front-end but we may find maintaining the levels of abstraction to become very difficult over time. The goal for the project should really be to compile Rust for GCC first.

Personally I quite like many of the functions this abstraction has provided.

As a way to get a middle ground here what if we were to get rid of the abstract classes wrapping up gcc tree into Bexpression, Bstmt, etc, so that they simply return tree.

tschwinge:
@philberty, I think we should do a proper decision here.  ;-)

Given #805, we're going to use the core-GCC-specific `tree` in the GCC/Rust front end -- thus forfeiting your original stated goal of keeping the front end independent of GCC specifics (via the abstractions/wrappers that #805 is now removing).  So are we now going all-GCC?  Meaning, we can also get rid of other such abstractions/wrappers?  (I'm fine with that for what it's worth.)

tschwinge:
Here is one other thought.  This came up outside of GCC/Rust, but now suddenly would be relevant here.  ;-)

A few weeks ago I had the idea (which also has occasionally been discussed in GCC upstream, for at least two decades or so) that we may add some GCC-compile-time static typing to `tree` (which is just a `union tree_node *`), that is, replace the dynamically typed `union [GTY] tree_node` with a proper statically-typed C++ inheritance hierarchy, like we have for `gimple`, for example.

<https://github.com/Rust-GCC/gccrs/blob/dcd758595f646a480947265ccc9833fdd3976b75/gcc/coretypes.h#L97>
<https://github.com/Rust-GCC/gccrs/blob/dcd758595f646a480947265ccc9833fdd3976b75/gcc/tree-core.h#L2021>
<https://github.com/Rust-GCC/gccrs/blob/dcd758595f646a480947265ccc9833fdd3976b75/gcc/gimple.h#L219>

I did start looking into that, but it'll be slow, very slow.  Per my assessment, given the expected volume of the changes, this needs to be done non-disruptively (and thus incrementally), directly is GCC master branch, with both the "old" and the "new" way co-existing for "a while".  (... and first needs to be accepted by GCC upstream, of course...)

[...]

Issue #804, opened 2021-11-15
TREE_ADDRESSABLE is not being set when required


philberty:
I tried this code:

```rust
extern "C" {
    fn printf(s: *const i8, ...);
}

#[lang = "add_assign"]
pub trait AddAssign<Rhs = Self> {
    fn add_assign(&mut self, rhs: Rhs);
}

impl AddAssign for i32 {
    fn add_assign(&mut self, other: i32) {
        unsafe {
            let a = "add_assign\n\0";
            let b = a as *const str;
            let c = b as *const i8;

            printf(c);
        }
        *self += other
    }
}

fn main() -> i32 {
    let mut a = 1;
    a += 2;

    0
}

```

I expected to see this happen: The address is taken on the VAR_DECL not an implicit temporary

Instead, this happened: TREE_ADDRESSABLE is false on 'a' and is required to stop the implicit copy

```c
void <i32 as AddAssign>::add_assign (i32 & self, const i32 other)
{
  {
    unsigned char * const & const a;
    unsigned char * const * const b;
    const i8 * const c;

    a = "add_assign\n";
    b = a;
    c = b;
    printf (c);
  }
  _1 = *self;
  _2 = other + _1;
  *self = _2;
}


i32 main ()
{
  i32 a.1;
  i32 D.86;
  i32 a;

  a = 1;
  a.1 = a; // this is wrong
  <i32 as AddAssign>::add_assign (&a.1, 2);
  D.86 = 0;
  return D.86;
}

```

### Meta

- What version of Rust GCC were you using, git sha if possible. This was found during dev of the operator overloading branch but exists on:  dcd758595f646a480947265ccc9833fdd3976b75

PR #805, opened 2021-11-15
Remove Btype, Bexpression, etc. abstractions over gcc trees

dafaust:
As discussed in #412, the Rust front end inherits an abstraction over gcc from the initial bootstrap via GCCGO. 
This is a cool idea, but adds overhead for the primary goal of gcc compiling Rust. It's not clear that the benefits of maintaining the abstraction are worth the potential extra headaches of maintaining it.

I figured for the sake of discussion, I'd draft an initial step towards removing this abstraction. 

The implementations of classes `Bytpe`, `Bexpression`, `Bstatement`, `Bfunction`, and `Bblock` are only wrappers around gcc's GENERIC `tree` structure, with no added functionality. This PR removes them, and changes all the functions for creating and manipulating these abstract types to just use/return `tree` instead. I also deleted a few functions that are vestiges from GCCGO port and aren't used.

Personally, I think the abstraction should be removed in favor of using `tree`s. This is more in line with the other gcc front ends and makes the interface between the frontend and the rest of gcc simpler.

I'm curious to hear other opinions on the matter :)

Addresses: #412

### DIFF
--- a/gcc/rust/backend/rust-compile-base.h
+++ b/gcc/rust/backend/rust-compile-base.h
@@ -200,16 +200,14 @@ protected:
   bool compile_locals_for_block (Resolver::Rib &rib, Bfunction *fndecl,
 				 std::vector<Bvariable *> &locals);
 
-  Bexpression *coercion_site (Bexpression *compiled_ref, TyTy::BaseType *actual,
-			      TyTy::BaseType *expected, Location locus);
+  tree coercion_site (tree compiled_ref, TyTy::BaseType *actual,
+		      TyTy::BaseType *expected, Location locus);
 
-  Bexpression *coerce_to_dyn_object (Bexpression *compiled_ref,
-				     const TyTy::BaseType *actual,
-				     const TyTy::BaseType *expected,
-				     const TyTy::DynamicObjectType *ty,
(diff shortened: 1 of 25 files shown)


philberty:
I think this is a great first step to vastly cleanup our code. GCC's GENERIC tree's are very similar to Rust's MIR IR this will make doing further static analysis much simpler and make the code much easier to follow. As you mentioned on irc the goal should probably be to remove the backend class over time but I guess we will still need a bunch of similar functions but I am not quite sure about what that looks like yet maybe @dkm or @tschwinge or mjw will help here to figure that out. I support this change.

dkm (approved):
I also like this change, makes the code more easy to read !

CohenArthur (approved):

philberty:
Are you ready to merge this @dafaust

philberty:
I think this will make fixing #804 quite easy instead of setting up new tables to track vars/params if they need to be marked as TREE_ADDRESSABLE we have access to the tree now to just set it at the GCC Generic level instead.

dafaust:
Yes I'm ready for this to be merged if everyone else is happy with it.

PR #820, opened 2021-11-25
Bug fix mutability checks in can_eq for autoderef

philberty:
Rust is permissive about mutablity in type checking for example, if we have
a function:

  fn foo(a:&bar) { ... }

  fn caller() {
    let a:&mut bar = ...;
    foo(a);
  }

This is valid since the mutable reference to bar is valid to be turned into
an immutable reference without any conversion. Like in C a non-const
pointer is valid to be passed to a const pointer inferface.


philberty:
bors cancel

philberty:
bors ping

PR #828, opened 2021-12-09
Add script to generate gource

philberty:
This script tries to isolate our changes to GCC minus the changes to
gcc/config so that we can visualize our development history.

I will be uploading a video output to youtube shortly, it also includes our cargo-integration project too :D

### DIFF
--- a/gcc/rust/gource-gccrs.sh
+++ b/gcc/rust/gource-gccrs.sh
@@ -1,0 +1,129 @@
+#!/usr/bin/env bash
+set -e
+
+usage() { echo "Usage: $0 [-o output.mp4] [-r 1600x1080]" 1>&2; }
+
+RESOLUTION="1600x1080"
+output=""
+while getopts ho:r: flag; do
+    case "${flag}" in
+        o)
+            output=${OPTARG}
+            ;;
+        r)
+            RESOLUTION=${OPTARG}
+            ;;
+        h)
+            usage
+            exit 0
+            ;;
+        *)
+            usage
+            exit 1
+            ;;
+    esac
+done
+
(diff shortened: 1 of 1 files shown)


flip1995 (commented):
I recommend running shellcheck over this. I did and the output was pretty big.

gcc/rust/gource-gccrs.sh
```diff
@@ -0,0 +1,108 @@
...
+    *.md \
+    gcc/rust \
+    gcc/testsuite/rust \
+    gcc/testsuite/rust.test > gccrs-raw.log
+cat gccrs-raw.log | gource --log-format git --output-custom-log gccrs.log -
+rm -f gccrs-raw.log
+sed -i -E "s#(.+)\|#\1|/compiler#" cargo.log
+
+# combine them
+combined_log="$(mktemp /tmp/gource.XXXXXX)"
```
dkm, line 72: Maybe remove the temp file after its use or using `trap .... EXIT` ?

philberty:
> I recommend running shellcheck over this. I did and the output was pretty big.

Thanks for the tip I was not aware of shellcheck, I've fixed all but the awk line warning not sure how I can fix it yet.

gcc/rust/gource-gccrs.sh
philberty: Added a cleanup step instead of a trap because sometimes killing to keep the tmp files can be useful.

philberty:
Here is the youtube! https://www.youtube.com/watch?v=QkdGY9VljTc

CohenArthur (approved):
Very cute present :)

PR #829, opened 2021-12-11
Constant folder now returns error_mark_node instead of nullptr

### LINKED ISSUE #692: Constant folder should use error_mark_node
The constant folder implementation should take advantage of error nodes.

- [ ] return error expression https://github.com/Rust-GCC/gccrs/blob/28f527c9598339cf834a30b5ee1f14258b8ecbb2/gcc/rust/typecheck/rust-hir-const-fold.h#L217 the default value of the folded result is already an error expression https://github.com/Rust-GCC/gccrs/blob/28f527c9598339cf834a30b5ee1f14258b8ecbb2/gcc/rust/typecheck/rust-hir-const-fold.h#L395
- [ ] update error handling code now that the result is at least error_mark_node this no longer needs a nullptr check https://github.com/Rust-GCC/gccrs/blob/28f527c9598339cf834a30b5ee1f14258b8ecbb2/gcc/rust/typecheck/rust-hir-type-check.cc#L395-L401

npate012:
Removed nullptr checking on results from constant folder because when the
result is already error_mark_node, we no longer need to check if the result
is nullptr.

Fixes #692

Signed-off-by: Nirmal Patel <npate012@gmail.com>

### DIFF
--- a/gcc/rust/typecheck/rust-hir-const-fold.cc
+++ b/gcc/rust/typecheck/rust-hir-const-fold.cc
@@ -64,8 +64,6 @@ void
 ConstFoldItem::visit (HIR::ConstantItem &item)
 {
   auto folded_expr = ConstFoldExpr::fold (item.get_expr ());
-  if (folded_expr == nullptr)
-    return;
 
   folded = folded_expr;
 }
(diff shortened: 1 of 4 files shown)


philberty (approved):
LGTM nice one this is exactly the idea, the more error nodes we use the more nullptr checks we can remove :D

Issue #825, opened 2021-12-08
HIR Visitor Refactoring


philberty:
# Background

Our HIR implementation is based off our AST slowly but surely we are cleaning this up to make the HIR much slimer such as no macros and methods are functions, all loops are loop expressions etc.

As part of this bootstrapping from our AST we only have a single Visitor interface that contains abstract methods for every type see: https://github.com/Rust-GCC/gccrs/blob/0024bc2f028369b871a65ceb11b2fddfb0f9c3aa/gcc/rust/hir/tree/rust-hir-visitor.h

The problem here is that when we are working with the full tree of the HIR we are forced to create visitor classes which contain all of the potential types even though the flow of the tree is always:

  Crate -> Items / types -> Stmts/Items/types -> Expressions / types

So even though we have a fully and strongly typed Intermediate representation to work with, we are not getting the full advantages of this yet. The major issue we have is that to avoid having to create overblown classes implementing every node in the graph we have these base classes: https://github.com/Rust-GCC/gccrs/blob/0024bc2f028369b871a65ceb11b2fddfb0f9c3aa/gcc/rust/backend/rust-compile-base.h

Which are starting to make the code quite messy to work with, the other problem is that every method has an empty block default we have cases in the code that do not have implementation such as the recent issue with ReferenceTypes on ImplBlocks such as:

```rust
impl<T> Foo for &T {}
```

This was due to the fact that impl blocks act on any HIR::TypeNoBounds and this visitor was missing an implementation for reference types:

https://github.com/Rust-GCC/gccrs/blob/0024bc2f028369b871a65ceb11b2fddfb0f9c3aa/gcc/rust/resolve/rust-ast-resolve-type.h#L125

Although that example is in the AST it shows how this visitor was missing making it very easy to miss cases in the full IR graph. If we have many smaller visitors we can do away with these base classes and use the C++ compiler to enforce the missing methods through using visitor abstract base classes. This will make it less confusing what methods we need to implement.

Then to take advantage of these new visitors let's focus this on the backend module (this should also be renamed to generic or something but that's a separate issue, backend is a bit of a weird name). If we focus usage of these new HIR visitors on the backend we limit the impact in finding regressions in the type-checking and lint passes until we are happy with this and the backend pass is a simpler case to start with.

https://github.com/Rust-GCC/gccrs/blob/0024bc2f028369b871a65ceb11b2fddfb0f9c3aa/gcc/rust/hir/tree/rust-hir-item.h

[...]

PR #846, opened 2021-12-22
HIR visitor refactor

dkm:
This is the first step of a bigger refactor (see #825).

This patch introduces a new HIRFullVisitor with default implementations for all
visitor methods. This visitor is used as base classe for several existing HIR
visitors that also currently define empty behavior (eg. TypeCheckBase,
ConstFoldBase, ...).


philberty (approved):
Nice one this already has removed a lot of duplication.

PR #847, opened 2022-01-03
HIR::ImplBlock items should mangle based from their canonical path's

### LINKED ISSUE #845: Duplicate symbol name generated
I tried this code:

```rust
extern "C" {                                                                          
    fn printf(s: *const i8, ...);                                                                                                                                            
} 

struct Foo {}

trait Bar {
  fn bar(self) {
      unsafe {                                                                                                                                                                 
        let _a = "Bar::bar\n\0"; 
        let _b = _a as *const str;                                                                                                                                           
        let _c = _b as *const i8;                                                     
        printf(_c);                                                                
    }  
  }
}

impl Foo {
  fn bar(self) {
    unsafe {                                                                                                                                                                 
        let _a = "Foo::bar\n\0"; 
        let _b = _a as *const str;                                                                                                                                           
        let _c = _b as *const i8;                                                     
        printf(_c);                                                                
    }   
  }
}

impl Bar for Foo {
  fn bar(self) {
    unsafe {                                                                                                                                                                 
        let _a = "<Bar as Foo>::bar\n\0"; 
        let _b = _a as *const str;                                                                                                                                           
        let _c = _b as *const i8;                                                     
        printf(_c);                                                                
    }  
  }
}

fn main() {
  let mut f = Foo{};
  f.bar();
}
```

Instead, this happened: 

```
$ /opt/gccrs-dev/bin/gccrs -g -O0 test.rs -o test
/tmp/ccefeQA4.s: Assembler messages:
/tmp/ccefeQA4.s:51: Error: symbol `_ZN9TestCrate3Foo3bar17h75f61ccaa11b25a4E' is already defined
```

The gimple looks good but the symbol that is being mangled must be bad:

```c
void Foo::bar (const struct  self)
{    
  {
    unsigned char * const & const _a;
   unsigned char * const * const _b;
    const i8 * const _c;

    _a = "Foo::bar\n";
    _b = _a;
    _c = _b;
    printf (_c);
  }
}


void <Foo as Bar>::bar (const struct  self)
{
  {
    unsigned char * const & const _a;
    unsigned char * const * const _b;
    const i8 * const _c;

    _a = "<Bar as Foo>::bar\n";
    _b = _a;
    _c = _b;
    printf (_c);
  }
}


void main ()
{
  struct  f;

  try
    {
      Foo::bar (f);
    }
  finally
    {
      f = {CLOBBER};
    }
}

```

### Meta

- What version of Rust GCC were you using, git sha if possible. 310968874db1902084012cf767ad0b6e93f028c2

philberty:
Legacy mangling converts the '<', '>' from their canonical paths into '..'
this means we can simply reuse our normal mangle_item code for all types
of symbols. So we can now remove the mangle_impl_item code prior to having
the canonical path code in place.

Fixes #845

### DIFF
--- a/gcc/rust/backend/rust-compile-context.h
+++ b/gcc/rust/backend/rust-compile-context.h
@@ -310,14 +310,6 @@ public:
     return mangler.mangle_item (ty, path, mappings->get_current_crate_name ());
   }
 
-  std::string mangle_impl_item (const TyTy::BaseType *self,
-				const TyTy::BaseType *ty,
-				const std::string &name) const
-  {
-    return mangler.mangle_impl_item (self, ty, name,
-				     mappings->get_current_crate_name ());
-  }
-
 private:
   ::Backend *backend;
   Resolver::Resolver *resolver;
--- a/gcc/rust/backend/rust-compile-implitem.h
+++ b/gcc/rust/backend/rust-compile-implitem.h
@@ -153,8 +153,7 @@ public:
 
     std::string ir_symbol_name
       = canonical_path->get () + fntype->subst_as_string ();
-    std::string asm_name
-      = ctx->mangle_impl_item (self, fntype, function.get_function_name ());
+    std::string asm_name = ctx->mangle_item (fntype, *canonical_path);
 
     tree fndecl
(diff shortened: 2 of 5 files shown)


CohenArthur (approved):
LGTM

PR #870, opened 2022-01-13
Add constant folding to const functions

### LINKED ISSUE #799: add constant folding to blocks and const fn
I tried this code:

```rust
const A:i32 = 1;
const B:i32 = { A+ 2};

const fn test(a:i32) -> i32 {
    a
}

const C:i32 = { const a:i32 = 123; A + a + test(456)};

pub fn cake() -> i32{
    C
}
```

Instead, this happened: errored with non const value

### Meta

- What version of Rust GCC were you using, git sha if possible.e28c43f2f40cf405e89b3892aa65f6a06fa1c802

philberty:
In Rust the ArrayType has a constant capacity constraint, which means it
allows for bounds checking at compile time as no variable-length arrays are
allowed. In order to typecheck this case we had a constant folding pass as
part of the type checking system which generated gcc tree's for the IR
and enforced the constant checking along the way.

GCC with optimizations turned on is capable of constant folding/propogating
the compilation unit fully, but we need a method that works regardless of
middlle-end optimizations to fold constant expressions at the front-end,
turns out the CPP front-end already does this via its constexpr
mechanism to ensure that these _do_ fold correctly. Another major reason
to do this change is that the original const fold pass was a striped down
copy of what the backend is _already_ doing which is creating a duplication
of the code generation pass. With this all unified into the code generation
pass all we need to do is port over gcc/cp/constexpr.c to enforce the const
rules fully but at the GCC tree level not at the typed HIR level.

Now that we have unified the pass when we hit a const function we can
simply emit a normal GCC function and outside of const expressions GCC
will simply emit a normal CallExpr and depending on optimization level
fully optimize this. If we are in a ConstDecl we will follow the
rust-constexpr.cc and fold the values or error_mark_node with an apropriate
error. By reusing the CPP constexpr code we _know_ it works so reusing it
as much as possible is a good idea in general for this front-end.

Fixes #799


philberty:
bors try

Issue #877, opened 2022-01-20
GIMPLE record and union types have empty struct names


philberty:
Our gimple trees when dumped have an empty name the debug_tree shows the names in their respective TYPE_DECL. see -fdump-tree-gimple

dafaust:
tl;dr -  looks like DECL_SOURCE_LOCATION (node) is 0 for our structs and unions. Looking into it.

We're hitting this in tree-pretty-print.c when printing the decls at the beginning of the gimple bind, for structs and unions:
https://github.com/Rust-GCC/gccrs/blob/b02824c6a798a78657568e7d831bd10529d63e37/gcc/tree-pretty-print.c#L2284-L2289

The gimplified TREE nodes themselves have the names as expected.

Why is DECL_IS_UNDECLARED_BUILTIN  true for our records and unions?
https://github.com/Rust-GCC/gccrs/blob/b02824c6a798a78657568e7d831bd10529d63e37/gcc/tree.h#L2528-L2533

Uh oh, looks like we are missing source location info:
```
(gdb) pt node
 <type_decl 0x7ffff74672f8 Header
    type <record_type 0x7ffff7466690 Header sizes-gimplified BLK
 ...
(gdb) p node->decl_minimal.locus
$4 = 0
```

<details><summary> gdb + backtrace</summary>
<p>

```
Breakpoint 6, dump_generic_node (pp=0x7fffffffd9e0, node=<type_decl 0x7ffff74672f8 Header>, spc=0, 
    flags=100663296, is_stmt=false) at ../../gccrs/gcc/tree-pretty-print.c:2285
2285	      if (DECL_IS_UNDECLARED_BUILTIN (node))
(gdb) p node
$2 = <type_decl 0x7ffff74672f8 Header>
(gdb) pt node
 <type_decl 0x7ffff74672f8 Header
    type <record_type 0x7ffff7466690 Header sizes-gimplified BLK
        size <integer_cst 0x7ffff7459648 constant 160>
        unit-size <integer_cst 0x7ffff7459618 constant 20>
        align:32 warn_if_not_align:0 symtab:0 alias-set -1 structural-equality
        fields <field_decl 0x7ffff745be40 magic type <integer_type 0x7ffff7452540 u16>
            unsigned HI (null):0:0
            size <integer_cst 0x7ffff743bdb0 constant 16>
            unit-size <integer_cst 0x7ffff743bdc8 constant 2>
            align:16 warn_if_not_align:0 offset_align 128
[...]

PR #879, opened 2022-01-20
Record correct location when compiling ADT types

dafaust:
When compiling struct and union types, we were using the wrong HirId
to lookup location information. The resulting GIMPLE nodes therefore
had no source location information, which made them indistinguishable
as user-declarations versus decls created by the compiler. As a result,
the type names were not shown in GIMPLE dumps, e.g. using
-fdump-tree-gimple.

Fix the location lookup, so these types are properly printed, and add a
simple test checking as much.

Fixes: #877


philberty (approved):
nice one

PR #909, opened 2022-02-04
Add support for key = value pairs on -frust-cfg=

### LINKED ISSUE #889: Add --cfg option to specify config options
https://doc.rust-lang.org/rustc/command-line-arguments.html#--cfg-configure-the-compilation-environment

This is kind of similar to GCC -D to add preprocessor defines but this added new specifications to cfg expansion.

philberty:
This adds in a basic parser to parse out key value pairs for the config
option it needs to be tested poperly once the self-test framework is
merged in.

Fixes #889

### DIFF
--- a/gcc/testsuite/rust/compile/cfg5.rs
+++ b/gcc/testsuite/rust/compile/cfg5.rs
@@ -1,0 +1,11 @@
+// { dg-additional-options "-w -frust-cfg=A=B" }
+struct Foo;
+impl Foo {
+    #[cfg(A = "B")]
+    fn test(&self) {}
+}
+
+fn main() {
+    let a = Foo;
+    a.test();
+}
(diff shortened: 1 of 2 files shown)


gcc/testsuite/rust/compile/cfg5.rs
```diff
@@ -0,0 +1,11 @@
+// { dg-additional-options "-w -frust-cfg=A=B" }
```
bjorn3, line 1: Rustc expects `B` to be between double quotes.

philberty: should we error our in this case then?

bjorn3: Maybe?

philberty: I've raised an issue from your comment we probably should do the same.

PR #922, opened 2022-02-13
Support inline never and always options

### LINKED ISSUE #921: rust inline attribute supports options
it supports never and always which seem to map nicely over to DECL_UNINLINEABLE and the always_inline attribute.

philberty:
This maps over to DECL_UNINLINEABLE and to use the GCC attribute
always_inline.

Fixes #921

### DIFF
--- a/gcc/testsuite/rust/compile/inline_1.rs
+++ b/gcc/testsuite/rust/compile/inline_1.rs
@@ -1,0 +1,15 @@
+#[inline]
+fn test_a() {}
+
+// { dg-final { scan-tree-dump-times {always_inline} 1 gimple } }
+#[inline(always)]
+fn test_b() {}
+
+#[inline(never)]
+fn test_c() {}
+
+fn main() {
+    test_a();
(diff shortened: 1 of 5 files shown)


gcc/testsuite/rust/compile/inline_1.rs
```diff
@@ -0,0 +1,15 @@
+#[inline]
+fn test_a() {}
+
+// { dg-final { scan-tree-dump-times {always_inline} 1 gimple } }
```
tschwinge, line 4: This produces:

```
PASS: rust/compile/inline_1.rs (test for excess errors)
UNRESOLVED: rust/compile/inline_1.rs scan-tree-dump-times gimple "always_inline" 1
```

```
                === rust Summary for unix ===

# of expected passes            5746
# of expected failures          21
# of unresolved testcases       1
```

1. Our automated testing shouldn't have let that pass.
2. Add the missing `-fdump-tree-gimple`.

PR #926, opened 2022-02-15
macros: Add multiple macro syntax tests

CohenArthur:
This PR adds test cases for macros, including parsing and execution.

I am unsure on how to check for proper execution: The solution I have chosen so far is to make sure that a correct amount of lines is printed, which I'm not entirely satisfied with. 

Another solution would be to increase a global integer to use when exiting, which we can then assert on using dejagnu, which is cleaner but relies on unsafe rust code.


philberty (approved):
Awesome

PR #935, opened 2022-02-16
frust-cfg: Only allow double quoted values

### LINKED ISSUE #910: Rustc expects `B` to be between double quotes.
Rustc expects `B` to be between double quotes.

_Originally posted by @bjorn3 in https://github.com/Rust-GCC/gccrs/pull/909#discussion_r799681447_

CohenArthur:
Closes #910 

This PR separates the `handle_cfg_option()` function in two, separating the parsing logic from the session logic. The parsing logic is able to be unit tested, and now only allows quoted values.

What remains to be done is to only allow `key` and `value` to be proper rust identifiers. We need to figure out if we'd like to spawn a parser here and parse identifiers, or simply sanitize both strings to make sure they do not contain invalid characters.

### DIFF
--- a/gcc/rust/rust-session-manager.cc
+++ b/gcc/rust/rust-session-manager.cc
@@ -29,6 +29,7 @@
 #include "rust-tycheck-dump.h"
 #include "rust-ast-resolve-unused.h"
 #include "rust-compile.h"
+#include "rust-cfg-parser.h"
 
 #include "diagnostic.h"
 #include "input.h"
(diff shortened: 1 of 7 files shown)


CohenArthur:
[This](https://github.com/rust-lang/rust/blob/a240ccd81c74c105b6f5fe84c46f8d36edb7e306/compiler/rustc_interface/src/interface.rs#L105-L108) is the part where rustc validates the `key` as well as the quoting. I'll open up an issue for this

gcc/rust/rust-session-manager.cc
```diff
@@ -1177,3 +1150,39 @@ TargetOptions::enable_implicit_feature_reqs (std::string feature)
  *  - code generation
  *  - link */
 } // namespace Rust
+
+#include "selftest.h"
```
philberty, line 1154: is it possible this can be in a bla_test.cc file? It might be that we pull out a rust-cfg-parser.cc and have a rust-cfg-parser-test.cc. Is that possible?

CohenArthur: Sure thing! From what I'm seeing, selftests usually live in the same file as the implementation, but we can definitely move them. Either have a `rust-cfg-parser.cc` which contains the `parse_cfg_option` function and the tests associated, or two different files. Let me know

philberty (approved):
LGTM

gcc/rust/rust-session-manager.cc
tschwinge: Yay for unit testing!

tschwinge: I think in GCC, often the unit tests are in the same file as the implementation they're testing, as otherwise you couldn't easily access `static` functions, for example.

Issue #573, opened 2021-07-17
Declarative macro expansion


SimplyTheOther:
This issue is designed to document the planned macro expansion system for declarative macros and to allow discussion regarding potential improvements to it.

All macro invocations are parsed in the Parser into either `MacroInvocation` or `MacroInvocationSemi` objects, depending on context. `MacroInvocationSemi`s can be `Item`s, associate items, or statements - basically anywhere where a semicolon would be required (though curly bracket macros do not require one). `MacroInvocation`s can be types, patterns or expressions - basically anywhere where a semicolon should never be used. This was done largely due to the Reference's separation into these two forms, and has its advantages and disadvantages.

However, in order to simplify expansion, both `MacroInvocation` and `MacroInvocationSemi` classes have a `MacroInvocData` member variable, which stores the actual "data" of the macro (including name of macro and all tokens used inside). This was intended to allow the MacroExpander to operate by expanding instances of the data class rather than having to write two virtually identical methods for each expand method.

The basic idea for macro expansion is for the `MacroExpander` to basically scan through the entire AST, storing any macro definitions found (`MacroRulesDefinition` for declarative macros) and then attempting to expand any macro invocations found. The result of the expanded macro invocation (if it is possible to expand it currently) will then replace the macro invocation in the AST.

To actually expand declarative macros, it is planned for the `expand_decl_macro` method in `MacroExpander` be used. This takes the invocation data and the rules definition, and should return an `ASTFragment`. This sounds simple enough, but the actual implementation required is quite complicated and has multiple potential approaches, as detailed in #17. In short, Rust macro definitions can (and frequently do) have several potential "match patterns", and if one does not match, then the next should be tried. The basic consensus at the time was to follow this "backtracking" approach, rather than attempt a predictive one or use a generated parser.

With the backtracking approach, the macro parser would call specific methods of a new `Parser` instance on the token stream stored in the macro - e.g. `parse_expr` when an expression non-terminal was found as the next token in the match pattern. If there was a parse error, the macro parser would abort parsing that match pattern and attempt the next. If the macro parser successfully matched the entire "match pattern" with no parse errors, then the "replacement pattern" specified by that match-replacement pair would be used to generate the AST fragment.

[...]

Issue #22, opened 2020-05-16
Macro invocation with curly-braces as a statement


NalaGinrut:
This issue is related to [0378-expr-macros ](https://github.com/rust-lang/rfcs/blob/master/text/0378-expr-macros.md) 

- Treat all macro invocations with parentheses, (), or square brackets, [], as expressions.
- Curly braces or a semicolon following the invocation to invoke a macro as a statement

However, it seems that `macro_invoce!{x}` doesn't work for nightly rustc:
```rust
macro_rules! maybe_return { ($e:expr) => ($e); }

fn frob(x: i32) -> i32{
    maybe_return! {x}
    // should return -1 
    -1
}
```

The result is:

```rust
error[E0308]: mismatched types
 --> ma.rs:4:20
  |
4 |     maybe_return! {x}
  |                    ^- help: consider using a semicolon here
  |                    |
  |                    expected `()`, found `i32`
```

I may incorrectly understand the RFC description, anyway, I leave it here for a placeholder.

Our curent parser doesn't recogonize it too:
```rust
i32 frob(x : i32)
BlockExpr:
{
 outer attributes: none
 inner attributes: none
 statements: none
 final expression: 
ArithmeticOrLogicalExpr: MacroInvocation: maybe_return!{x} - 1
}
```

SimplyTheOther:
The RFC actually seems to be a bit ambiguous on what it really means, but the name of the RFC (expr-macros) and the name of the issue for the RFC ("Implement parse macro!()/macro![] as expr) implies to me that the primary point of the RFC is to force parsing macro invocations with parentheses or square brackets as an expression. As such, macro invocations with curly brackets _can_ be parsed as an expression, but do not have to be. 

That's just my interpretation of it, but it seems like it probably works with how rustc implements it.

TL;DR:

- Treat all macro invocations with parentheses, (), or square brackets, [], as expressions. 
- If the macro invocation has curly brackets, {}, it may be parsed as a statement depending on the context.
- If the macro invocation has a semicolon at the end, it must be parsed as a statement (either via ExpressionStatement or MacroInvocationWithSemi)

NalaGinrut:
@SimplyTheOther Yes, I realized that something in the RFC doesn't actually hold in the implementation, I'll try to point them out if I found them. At least for now, I think we should do what `rustc` does, and enhance them according to the ideas in RFC when the time is proper. After all, following the RFC could be one of our selling points.

cc @philberty

[...]

PR #938, opened 2022-02-16
First pass at declarative macro expansion 

### LINKED ISSUE #17: Macro expansion

philberty:
This does not support repetition matchers but it supports simple
declarative macros and transcribes them. The approach taken here is that
we reuse our existing parser to call the apropriate functions as specified
as part of the MacroFragmentType enum if the parser does not have errors
parsing that item then it must be a match.
    
Then once we match a rule we have a map of the token begin/end offsets
for each fragment match, this is then used to adjust and create a new token
stream for the macro rule definition so that when we feed it to the parser
the tokens are already substituted. The resulting expression or item is
then attached to the respective macro invocation and this is then name
resolved and used for hir lowering.
    
Fixes #17 #22
Addresses #573


gcc/rust/ast/rust-ast.h
```diff
@@ -786,6 +773,41 @@ class DelimTokenTree : public TokenTree, public AttrInput
...
+    switch (delim_type)
+      {
+      case PARENS:
+	return LEFT_PAREN;
+      case SQUARE:
+	return LEFT_SQUARE;
+      case CURLY:
+	return LEFT_CURLY;
+      }
+    return LEFT_PAREN;
```
CohenArthur, line 795: I wonder if here it wouldn't be more "future-proof" to do an exhaustive switch statement and have a gcc_unreachable as the default case. This way, if the rust language decides to add new delimiters we'd find ourselves in a crash situation rather than a situation where the compiler tells us it's fine and assures us it's seen parentheses
```suggestion
    switch (delim_type)
      {
      case PARENS:
	return LEFT_PAREN;
      case SQUARE:
	return LEFT_SQUARE;
      case CURLY:
	return LEFT_CURLY;
      default:
        gcc_unreachable();
      }
```

gcc/rust/parse/rust-parse.h
```diff
@@ -86,7 +86,7 @@ struct ParseRestrictions
 // TODO: if updated to C++20, ManagedTokenSource would be useful as a concept
 template <typename ManagedTokenSource> class Parser
 {
-private:
+public:
```
CohenArthur, line 89: Maybe keep this as private and move `peek_current_token()` in public members? I don't know how much it affects the rest of the codebase to leak these private functions

gcc/testsuite/rust/execute/torture/macros2.rs
```diff
@@ -0,0 +1,40 @@
+// { dg-output "arg\narg\n" }
```
CohenArthur, line 1:
```suggestion
// { dg-output "arg\narg\narg\n" }
```

gcc/rust/ast/rust-ast.h
philberty: sounds good

gcc/rust/parse/rust-parse.h
philberty: yeah lets do that, i will need to do this fix manually to make public the parse functions we want

gcc/testsuite/rust/execute/torture/macros2.rs
philberty: good find

gcc/rust/parse/rust-parse.h
philberty: fixed

gcc/rust/ast/rust-ast.h
```diff
@@ -788,6 +771,42 @@ class DelimTokenTree : public TokenTree, public AttrInput
...
+    switch (delim_type)
+      {
+      case PARENS:
+	return RIGHT_PAREN;
+      case SQUARE:
+	return RIGHT_SQUARE;
+      case CURLY:
+	return RIGHT_CURLY;
+      }
+    return RIGHT_PAREN;
```
CohenArthur, line 808: Oops, same here. I didn't put a suggestion here in my first review
```suggestion
    switch (delim_type)
      {
      case PARENS:
	return RIGHT_PAREN;
      case SQUARE:
	return RIGHT_SQUARE;
      case CURLY:
	return RIGHT_CURLY;
      default:
        gcc_unreachable ();
      }
```

CohenArthur: Otherwise nice :D I'm looking forward to playing with macros in gccrs

CohenArthur (approved):

PR #944, opened 2022-02-18
parser: Add `clear_errors()` method

CohenArthur:
Clears all errors from the error table so we can reuse the parser in later situations. I'm unsure whether or not the method should also emit all of the errors if present? In that case, we might want to rename it or add another wrapper that emits then clears


philberty (approved):
LGTM

Issue #849, opened 2022-01-04
Add support for Slices


philberty:
We need to support slices to be able to compile the goal test case #682 

see:

https://doc.rust-lang.org/book/ch04-03-slices.html
https://github.com/bminor/binutils-gdb/blob/79541a6d9220e800ebc5278594105982d6e1d80c/gdb/rust-lang.c#L157-L166
https://github.com/bminor/binutils-gdb/blob/79541a6d9220e800ebc5278594105982d6e1d80c/gdb/rust-lang.c#L299-L312

```rust
let a = [1, 2, 3, 4, 5];

let slice = &a[1..3];
```

- [x] range lang items #974
- [x] index lang items #974
- [x] offset intrinsics #1003
- [x] Slice resolution and type code #951
- [x] #1005 #1007
- [x] #1008
- [x] #1009
- [x] Add backend code-generation for the slice type and respect the layout of the FatPtr in libcore #1015 
- [x] hir lowering #1016 
- [x] #1019 #1030 
- [x] #1031 #1033
- [x] #1034
- [x] unsized method resolution #1045 
- [x] #1075
- [ ] #1097 
- [ ] #1098
- [ ] #1099
- [ ] #1086

PR #951, opened 2022-02-18
Add name resolution to slices

philberty:
This is the first part to support slices where we complete the name
resolution of the SliceType with its element as well as the range
expressions used in construction.

This patch also includes separation of the implementation from the
headers to try and incrementally improve build speed.

Addresses #849


CohenArthur (approved):
LGTM :) I'm glad you split up the commits nicely, I was a bit scared to try and find changes among the big file refactors haha

PR #955, opened 2022-02-21
matched_fragment: Track and set fragment match amount

CohenArthur:


gcc/rust/expand/rust-macro-expand.cc
```diff
@@ -3669,10 +3677,23 @@ MacroExpander::match_n_matches (
...
+			 : match_amount >= lo_bound;
+
+  // We can now set the amount to each fragment we matched in the substack
+  auto &stack_map = sub_stack.peek ();
+  for (auto &fragment_id : fragment_identifiers)
+    {
+      auto it = stack_map.find (fragment_id);
+
+      // We literally just added the fragment to the sub_stack at this point...
+      if (it == stack_map.end ())
```
philberty, line 3690: we usually do this check via rust_assert(it != stack_map.end())

philberty (approved):
LGTM

PR #956, opened 2022-02-21
Substitute repetitions

### LINKED ISSUE #960: Expanding needs to reference to the proper subfragments
When expanding the following function:
```rust
macro_rules! add_exprs {
    ($($e:expr)*) => (0 $(+ $e)*)
}

fn main() {
    // 1 + 2 + 15 => 18
    print_int(add_exprs!(1 2 15));
}
```
the following token stream would get produced once #956 is merged: `0 + 1 + 1 + 1`

We must be able to reference "sub-fragments" in the macro expansion. We could do this by giving them a new name, or fixing the offsets in the `MatchedFragment` struct

### LINKED ISSUE #961: Expanding fragment matched zero times should succeed
```rust
macro_rules! add_exprs {
    ($($e:expr)?) => ( 0 $(+ $e)?)
}
```
should expand properly to `0` if no argument is given to `add_exprs!`, and to `0 + $e` if one is passed

CohenArthur:
Needs #955 

This PR splits up the `substitute_tokens` function into multiple smaller functions. Still a draft until I can get repetitions working.

Closes #960 
Closes #961

### DIFF
--- a/gcc/rust/expand/rust-macro-expand.cc
+++ b/gcc/rust/expand/rust-macro-expand.cc
@@ -3882,7 +3882,6 @@ MacroExpander::substitute_metavar (
 {
   auto metavar_name = metavar->get_str ();
 
-  rust_debug ("expanding metavar: %s", metavar_name.c_str ());
   std::vector<std::unique_ptr<AST::Token>> expanded;
   auto it = fragments.find (metavar_name);
   if (it == fragments.end ())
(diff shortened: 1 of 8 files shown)


CohenArthur:
We can now substitute simple patterns such as these
```rust
macro_rules! add_exprs {
    ($($e:expr)+) => ( 0 + $($e +) )
}

let a0 = add_exprs!(2); // expands to 0 + 2
let a1 = add_exprs!(a0); // expands to 0 + a0
```

What remains is fixing #960 and #961 to pass the added tests

gcc/rust/expand/rust-macro-expand.cc
```diff
@@ -3677,21 +3672,8 @@ MacroExpander::match_n_matches (
...
-    {
-      auto it = stack_map.find (fragment_id);
-
-      rust_assert (it != stack_map.end ());
-
-      it->second.set_match_amount (match_amount);
-    }
-
-  return result;
+  return hi_bound ? match_amount >= lo_bound && match_amount <= hi_bound
```
philberty, line 3675: This is just a style thing when it comes to a long ternery statement like this I would suggest change it into this:

```c
bool did_meet_hi_bound = hi_bound ? match_amount <= hi_bound : true;
bool did_meet_lo_bound = match_amound >= lo_bound;

return did_meet_hi_bound && did_meet_lo_bound
```

gcc/rust/expand/rust-macro-expand.h
```diff
@@ -57,11 +57,21 @@ struct MatchedFragment
   size_t match_amount;
 
   MatchedFragment (std::string identifier, size_t token_offset_begin,
-		   size_t token_offset_end, size_t match_amount = 0)
+		   size_t token_offset_end, size_t match_amount = 1)
```
philberty, line 60: is this change meant to be here?

philberty (approved):

gcc/rust/expand/rust-macro-expand.h
CohenArthur: Yes! The default behavior should be that there is a single match when we create it. It's only when we want to specifically create an empty match that we set it to zero. The previous implementation was slightly flawed

PR #969, opened 2022-02-23
Add builtin macros framework

CohenArthur:
This PR adds bases to define new builtin macro functions.

Since we operate at the `insert_macro_def` level, this requires builtin macros to be defined, as is the case in the rust standard library:

```rust
    macro_rules! assert {
        ($cond:expr $(,)?) => {{ /* compiler built-in */ }};
        ($cond:expr, $($arg:tt)+) => {{ /* compiler built-in */ }};
    }
```

### DIFF
--- a/gcc/rust/ast/rust-macro.h
+++ b/gcc/rust/ast/rust-macro.h
@@ -361,8 +361,25 @@ class MacroRulesDefinition : public MacroItem
   DelimType delim_type;
   // MacroRules rules;
   std::vector<MacroRule> rules; // inlined form
-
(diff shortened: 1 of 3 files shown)


gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -0,0 +1,12 @@
+#include "rust-macro-builtins.h"
```
philberty, line 1: missing copyright header

philberty (approved):
Looks good just missing the header

PR #970, opened 2022-02-23
Add file!() builtin

CohenArthur:


gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -0,0 +1,27 @@
...
+#include "rust-expr.h"
+#include "rust-session-manager.h"
+
+namespace Rust {
+AST::ASTFragment
+MacroBuiltin::assert (Location invoc_locus, AST::MacroInvocData &invoc)
+{
+  rust_debug ("assert!() called");
+
+  return AST::ASTFragment::create_empty ();
```
philberty, line 12: this could call GCC's builtin assert but we need a way to preserve that information into the HIR that its a builtin call. Maybe we could introduce a builtin call HIR node and handle it that way.

CohenArthur: Yeah I agree. This was just to make sure that the builtin was working. I'm working on actually implementing assert in my `add-assert-macro` branch [here](https://github.com/CohenArthur/gccrs/blob/add-assert-macro/gcc/rust/expand/rust-macro-builtins.cc#L56). I don't know how easy it would be to insert a call to __builtin_assert at the AST level?

gcc/testsuite/rust/execute/torture/builtin_macros1.rs
```diff
@@ -0,0 +1,18 @@
+// { dg-output "rust/execute/torture/builtin_macros1.rs" }
+macro_rules! file {
```
philberty, line 2: does this need to be defined if its a builtin macro?

philberty: Ah that's interesting we could use that attribute as a mechanism to mark the NodeId over to a built-in definition. This will make more sense when we start working on imports and visibility though.

bjorn3:
Are you remapping paths in `file!()` if they are remapped for debuginfo? That is necessary for reproducible builds. (AFAIK rustc does this. Haven't checked though.)

CohenArthur:
> Are you remapping paths in `file!()` if they are remapped for debuginfo? That is necessary for reproducible builds. (AFAIK rustc does this. Haven't checked though.)

Not yet, no. I saw mentions of this behavior and thought we could skip it for now. I do need to open up an issue regarding this.

I saw that the remapping was "namely [for] paths into libstd", which we cannot compile yet, so I thought I'd add a base `file!()` macro in the mean time, to help with unit testing and other nice, rust-side assertions that we might wanna implement :)

PR #981, opened 2022-02-25
macro-expand: Add SubstitutionCtx class in its own file

### LINKED ISSUE #957: Refactor MacroExpander::substitute_* functions
Add a new `MacroSubstitutionContext` class and move `MacroExpander::substitute_*` static functions there

CohenArthur:
The `MacroExpander` class had multiple static functions which were constantly passing the same parameters around for expansion. This refactor adds a new `SubstituteCtx` class which keeps track of the three common arguments given to the substitute functions, and offers these implementations in a new source file to keep the original expander light.

Closes #957


philberty (approved):
nice work

PR #983, opened 2022-02-26
Parse proper cfg values

### LINKED ISSUE #936: Validate key and value for `-frust-cfg` key-value pairs
`rustc` expects the key and possible value given to the `--cfg` argument to be valid identifiers:

> The value can either be a single identifier or two identifiers separated by =.

 ([source](https://doc.rust-lang.org/rustc/command-line-arguments.html#--cfg-configure-the-compilation-environment))
 
 We must thus proceed to the parsing and checking of the various key-value pairs given to the compiler through the `-frust-cfg` argument.
 
 rustc validates the key [like so](https://github.com/rust-lang/rust/blob/a240ccd81c74c105b6f5fe84c46f8d36edb7e306/compiler/rustc_interface/src/interface.rs#L105-L108), making sure that it only corresponds to one parsed item. Just below, the compiler proceeds to more validation and also checks that the value is surrounded by double quotes, as pointed out in #910.

CohenArthur:
Closes #936

### DIFF
--- a/gcc/rust/lex/rust-lex.h
+++ b/gcc/rust/lex/rust-lex.h
@@ -144,7 +144,7 @@ public:
   /**
    * Lex the contents of a string instead of a file
    */
-  static Lexer lex_string (std::string &input)
+  static Lexer lex_string (std::string input)
   {
     // We can perform this ugly cast to a non-const char* since we're only
     // *reading* the string. This would not be valid if we were doing any
--- a/gcc/rust/lex/rust-lex.h
+++ b/gcc/rust/lex/rust-lex.h
@@ -23,6 +23,7 @@
 #include "rust-buffered-queue.h"
 #include "rust-token.h"
 
+#include <cstdio>
 #include <utility>
 #include <tuple>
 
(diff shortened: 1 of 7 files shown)


CohenArthur:
Not sure why there are so many formatting changes in rust-session-manager... I can remove them if necessary but it doesn't fail the CI

gcc/rust/lex/rust-lex.h
```diff
@@ -132,6 +141,19 @@ class Lexer
   Lexer (const char *filename, RAIIFile input, Linemap *linemap);
   ~Lexer ();
 
+  /**
+   * Lex the contents of a string instead of a file
+   */
+  static Lexer lex_string (std::string &input)
```
philberty, line 147: To make this a little safer I think you should remove the '&'  on the parameter. Since this is no longer const we can no longer guarantee that there will be no side effects on that parameter anymore.

philberty (approved):
Minor nit in the argument of lex_string but otherwise nice work! :D I will leave it up to you wether you want to change this.

The only other option is if we change the lex_string to keep the const std::string& but then take a strdup of the contents which then gets free'd in the destructor.

PR #984, opened 2022-02-27
Implimented Soluion 1 and solution 2 for issue_734

### LINKED ISSUE #734: Get rid of lambda in TyTy::FnPtr iterate_params
Just like the other lambda's, this is a bad way to do static analysis on the parameters.  https://github.com/Rust-GCC/gccrs/blob/3057d98a4e3fc777b06e51cd678723f5c5b4c898/gcc/rust/typecheck/rust-tyty.h#L1267

The problem here is that for function pointers there can be type indirection so we wrap up the parameters within TyVar's so it forces a lookup based on the context. To fix this issue:

- [ ] Remove iterate_params function
- [ ] Create new get_params function

This get params function should either:

1. Simply return the std::vector<TyVar>& params
2. Or we could create a new std::vector or TyTy::BaseType* and construct it with param[idx].get_tyty().

I am not sure which is the cleanest solution right now I think the 2nd one fits in best.

mvvsmk:
Fixes #734 
Done :
- [x]  Remove iterate_params function
- [x] Create new get_params function

Solution 1
1) Created a new get_params function which returns the parameters.
2) Changed the references of the iterate_params to use get_params.

Solution 2
1) Added get_params2 which returns `std::vector<TyTy::BaseType*>`
2) Changed the references of the iterate_params to use get_params.

Status :  
Currently I have implemented the first solution. 

Signed-off-by : M V V S Manoj Kumar <mvvsmanojkumar@gmail.com>

### DIFF
--- a/gcc/rust/backend/rust-compile-type.cc
+++ b/gcc/rust/backend/rust-compile-type.cc
@@ -148,11 +148,13 @@ TyTyResolveCompile::visit (const TyTy::FnPtr &type)
   tree result_type = TyTyResolveCompile::compile (ctx, type.get_return_type ());
 
   std::vector<tree> parameters;
-  type.iterate_params ([&] (TyTy::BaseType *p) mutable -> bool {
-    tree pty = TyTyResolveCompile::compile (ctx, p);
-    parameters.push_back (pty);
-    return true;
-  });
+
+  auto &params = type.get_params ();
+  for (auto &p : params)
+    {
+      tree pty = TyTyResolveCompile::compile (ctx, p.get_tyty ());
+      parameters.push_back (pty);
+    }
 
   translated = ctx->get_backend ()->function_ptr_type (result_type, parameters,
 						       type.get_ident ().locus);
--- a/gcc/rust/typecheck/rust-tyty.cc
+++ b/gcc/rust/typecheck/rust-tyty.cc
@@ -1294,10 +1294,13 @@ std::string
 FnPtr::as_string () const
 {
   std::string params_str;
-  iterate_params ([&] (BaseType *p) mutable -> bool {
-    params_str += p->as_string () + " ,";
-    return true;
-  });
+
+  auto &params = get_params ();
+  for (auto &p : params)
+    {
+      params_str += p.get_tyty ()->as_string () + " ,";
+    }
+
   return "fnptr (" + params_str + ") -> " + get_return_type ()->as_string ();
 }
 
--- a/gcc/rust/typecheck/rust-tyty.h
+++ b/gcc/rust/typecheck/rust-tyty.h
@@ -1526,6 +1526,9 @@ public:
       }
   }
 
+  std::vector<TyVar> &get_params () { return params; }
+  const std::vector<TyVar> &get_params () const { return params; }
+
   bool is_concrete () const override final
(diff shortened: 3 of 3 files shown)


gcc/rust/backend/rust-compile-type.cc
```diff
@@ -148,11 +148,23 @@ TyTyResolveCompile::visit (const TyTy::FnPtr &type)
...
+  //     tree pty = TyTyResolveCompile::compile (ctx, p.get_tyty ());
+  //     parameters.push_back (pty);
+  //   }
+
+  // Solution 2
+
+  std::vector<TyTy::BaseType *> p = type.get_params ();
+  for (long unsigned int i = 0; i < p.size (); i++)
+    {
+      tree pty = TyTyResolveCompile::compile (ctx, p[i]);
```
CohenArthur, line 165:
```suggestion
  auto params = type.get_params ();
  for (auto &param: params)
    {
      tree pty = TyTyResolveCompile::compile (ctx, param);
```

I think this is more of a style thing, but we have all of these facilities in C++11 already so why not use them :)

gcc/rust/typecheck/rust-tyty.cc
```diff
@@ -1294,10 +1294,21 @@ std::string
...
+  //    {
+  //      params_str += p.get_tyty ()->as_string () + " ,";
+  //    }
+
+  // solution 2
+  auto p = this->get_params ();
+  for (long unsigned int i = 0; i < p.size (); i++)
+    {
+      params_str += p[i]->as_string () + " ,";
+    }
```
CohenArthur, line 1310:
```suggestion
  auto params = this->get_params ();
  for (auto &param: params)
      params_str += param->as_string () + " ,";
```

gcc/rust/typecheck/rust-tyty.h
```diff
@@ -1526,6 +1526,22 @@ class FnPtr : public BaseType
...
+  // const std::vector<TyVar> &get_params () const { return params; }
+
+  // Solution 2
+  std::vector<TyTy::BaseType *> get_params () const
+  {
+    std::vector<TyTy::BaseType *> p;
+    for (long unsigned int i = 0; i < params.size (); i++)
+      {
+	p.push_back (params[i].get_tyty ());
+      }
```
CohenArthur, line 1541:
```suggestion
    std::vector<TyTy::BaseType *> p;
    for (auto &param: params)
	p.push_back (param.get_tyty ());
```

CohenArthur (commented):
Minor style nitpickings :D I think once you discuss with Philip which solution you both like best and remove the commented-out algorithms, this is great! Thanks a lot!

mvvsmk:
Hey @CohenArthur , I'll incorporate these changes :) , as soon as Phllip says which solution he prefers.

philberty:
@mvvsmk nice work. I think you might be right that solution 1 is best because in solution two we hide any possible error handling over cases where the TyVar might fail which will be important in a few months from now when we focus on improved error handling.

Thanks so much for doing both solutions so we can compare nice work.

CohenArthur (approved):
Thank you! LGTM

philberty (approved):

PR #985, opened 2022-03-01
Parse macro!(); as MacroInvocation with semicolon

### LINKED ISSUE #941: Macro invocation statements with semi colon are not being parsed as AST::MacroInvocationSemi
I tried this code:

```rust
macro_rules! Test {
    ($a:ident, $b:ty) => {
        struct $a($b);
    };
}

fn main() -> i32 {
    Test!(Foo, i32);

    let a = Foo(123);
    a.0 - 123
}

```

I expected to see this happen: Macro invocation is parsed as AST::MacroInvocationSemi

Instead, this happened: it is currently parsed as AST::MacroInvocation

```
Crate: 
 inner attributes: none
 items: 
  
 outer attributes: nonemacro_rules!Test
 Macro rules: 
  Macro rule: 
 Matcher: 
  Macro matcher: 
 Delim type: parentheses
 Matches: 
  $a: ident
  ,
  $b: ty
 Transcriber: 
  {{struct$a(($b));}}
 Delim type: curly
  
i32 main()
BlockExpr:

  outer attributes: none
  inner attributes: none
 statements: 
 ExprStmtWithoutBlock:
  MacroInvocation: 
   outer attributes: none
 Test!((Foo,i32))
 
  outer attributes: none
 let a = CallExpr: 
  outer attributes: none
 Function expr: Foo
 Call params:
  123
 final expression: 
ArithmeticOrLogicalExpr: a.0 - 123
```

### Meta

- What version of Rust GCC were you using, git sha if possible. 37415eec77438bba2fc61df3e9a396c1e2cbaca8

CohenArthur:
When parsing a macro invocation as a statement, the parser would parse
an expression and then try parsing a semicolon. Since no actual
lookahead was done (which is a good thing), we couldn't convert a
`MacroInvocation` to a `MacroInvocationSemi` after the fact.

Since, unlike function calls, macro invocations can act differently
based on whether or not they are followed by a semicolon, we actually
need to differentiate between the two up until expansion.

This commits adds a new virtual method for ExprWithoutBlock when
converting to ExprStmtWithoutBlock so that classes inheriting
ExprWithoutBlock can specify a new behavior. In the case of our
MacroInvocation class, it simply means toggling a boolean: If we're
converting a macro from an expression to a statement, it must mean that
it should contain a semicolon.

Closes #941

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -861,7 +861,9 @@ public:
   add_crate_name (std::vector<std::string> &names ATTRIBUTE_UNUSED) const
   {}
 
-  bool is_item () const override final { return true; }
+  // FIXME: ARTHUR: Is it okay to have removed that final? Is it *required*
+  // behavior that we have items that can also be expressions?
+  bool is_item () const override { return true; }
 
 protected:
   // Clone function implementation as pure virtual method
@@ -957,6 +959,8 @@ public:
   {
     return clone_expr_without_block_impl ();
   }
+
+  virtual ExprWithoutBlock *to_stmt () const { return clone_expr_impl (); }
 };
 
 /* HACK: IdentifierExpr, delete when figure out identifier vs expr problem in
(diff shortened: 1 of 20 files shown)


gcc/rust/ast/rust-ast.h
```diff
@@ -861,7 +861,9 @@ class Item : public Stmt
   add_crate_name (std::vector<std::string> &names ATTRIBUTE_UNUSED) const
   {}
 
-  bool is_item () const override final { return true; }
+  // FIXME: ARTHUR: Is it okay to have removed that final? Is it *required*
+  // behavior that we have items that can also be expressions?
+  bool is_item () const override { return true; }
```
philberty, line 866: It might be worth putting the final back in to see if it still works as this will enforce that is_item is not overridden anymore though I am not too worried about it.

philberty (approved):
Very neat solution the only question I have is that we should check does rustc keep a separate structure for Semicolon macro invocations? I think it makes sense to just turn it into a flag.

gcc/rust/ast/rust-ast.h
CohenArthur: The test suite doesn't pass properly without it sadly. The checks in the various visitors are not enough, so I think it's necessary to keep it. I can change the comment for one meaning we *have* to look at fixing this at some point

CohenArthur:
> Very neat solution the only question I have is that we should check does rustc keep a separate structure for Semicolon macro invocations? I think it makes sense to just turn it into a flag.

rustc actually parses it the other way around from us from what I can figure. The code is as follows:
```rust
154:    fn parse_stmt_mac(&mut self, lo: Span, attrs: AttrVec, path: ast::Path) -> PResult<'a, Stmt> {
155-        let args = self.parse_mac_args()?;
156-        let delim = args.delim();
157-        let hi = self.prev_token.span;
158-
159-        let style =
160-            if delim == token::Brace { MacStmtStyle::Braces } else { MacStmtStyle::NoBraces };
161-
162-        let mac = MacCall { path, args, prior_type_ascription: self.last_type_ascription };
163-
164-        let kind =
165-            if (delim == token::Brace && self.token != token::Dot && self.token != token::Question)
166-                || self.token == token::Semi
167-                || self.token == token::Eof
168-            {
169-                StmtKind::MacCall(P(MacCallStmt { mac, style, attrs, tokens: None }))
170-            } else {
171-                // Since none of the above applied, this is an expression statement macro.
172-                let e = self.mk_expr(lo.to(hi), ExprKind::MacCall(mac), AttrVec::new());
173-                let e = self.maybe_recover_from_bad_qpath(e, true)?;
174-                let e = self.parse_dot_or_call_expr_with(e, lo, attrs.into())?;
175-                let e = self.parse_assoc_expr_with(0, LhsExpr::AlreadyParsed(e))?;
176-                StmtKind::Expr(e)
177-            };
178-        Ok(self.mk_stmt(lo.to(hi), kind))
179-    }
```
where they can look for a semicolon before actually creating the `Statement`. In the case that a semicolon (or other valid delimiters, but for this specific case that we're fixing, a semicolon) isn't found, an expression is built and returned. A bit like our `ExprOrStmt` class, except here it's a `StmtOrExpr` in a way! It's an enum/sum type instead of a class with two pointers though.

philberty (approved):

PR #990, opened 2022-03-03
Add must use attribute support

### LINKED ISSUE #856: must_use attribute support
see: https://github.com/rust-lang/rust/blob/2b681ac06b1a6b7ea39525e59363ffee0d1a68e5/library/core/src/result.rs#L537-L543

This attribute is very important in rust to provide an error when the result is not used. This seems like a very good candidate to be implemented at the GCC IR level using TREE_USED to track it possibly with a custom ATTRIBUTE on the tree.

philberty:
This is a port of the CPP front-end nodiscard attribute to be used for
must_use. It contains a patch to clean up how we handle expressions vs
statements and removes more of the GCC abstraction. Its my hope that we
can leverage more and more existing code to get the static analysis where
we want it.

Fixes #856

### DIFF
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -111,6 +111,7 @@ GRS_OBJS = \
     rust/rust-compile-type.o \
     rust/rust-constexpr.o \
     rust/rust-compile-base.o \
+    rust/rust-tree.o \
     $(END)
 # removed object files from here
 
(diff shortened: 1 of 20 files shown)


CohenArthur (approved):
Great :D I tried it and the output is *really cool*. Good job!!! I'm excited to have this in and working for types too :D

PR #992, opened 2022-03-04
Cleanup bad unused code warnings

### LINKED ISSUE #676: Bad warnings

philberty:
This patchset contains 4 distinct fixes:

When a constant is declared after where it is used the code-generation pass falls
back to a query compilation of the HIR::Item this did not contain a check to verify
if it was already compiled and results in duplicate CONST_DECLS being generated
if query compilation was used.

We were using a zero precision integer to contain unit-type expressions this results
in VAR_DECLS being lost in the GENERIC graph which does not allow us to perform
any static analysis upon the DECL. This changes the unit type to use an empty struct
and for initialization of a VAR_DECL we can simply pass an empty constructor and let 
GCC optimize this code for us.

Update our DEAD_CODE scan to take into account modules of items and also respect
if structures are prefixed with an underscore we can ignore generating an unused warning.

Remove our AST scan for unused code and reuse GCC TREE_USED to track wether
VAR_DECL, PARM_DECL, CONST_DECL are actually used or not. We reuse the GCC
walk_tree functions to have this as nice separate lint.

Fixes #676


gcc/testsuite/rust/compile/torture/all_doc_comment_line_blocks_crlf.rs
```diff
@@ -9,40 +9,37 @@
...
-    /*!*/
-    //
-
-    ///
-    mod doc { }    // { dg-warning "unused name" }
-
-    /**/
-    /***/
-  }
+pub mod module {
```
dkm, line 12: minor comment, but looks like you are reindenting automatically here.

CohenArthur (approved):
Nice, this is clean. I guess we'll need to think about the default behavior/warnings we want to enable regarding unused variables, either on gccrs or over on cargo-gccrs

gcc/testsuite/rust/compile/torture/all_doc_comment_line_blocks_crlf.rs
philberty: good find, those edits negated the test case. I have reverted my changes and simply removed the dg-warnings properly.

gcc/rust/lang.opt
```diff
@@ -38,6 +38,18 @@ Wall
...
+Rust Var(warn_unused_variable) Init(1) Warning
+; documented in common.opt
+
+Wunused-const-variable
+Rust Warning Alias(Wunused-const-variable=, 2, 0)
+Warn when a const variable is unused.
+
+Wunused-const-variable=
+Rust Joined RejectNegative UInteger Var(warn_unused_const_variable) Init(1) Warning LangEnabledBy(Rust,Wunused-variable, 1, 0) IntegerRange(0, 2)
+Warn when a const variable is unused.
```
tschwinge, line 51: I'm back to working on resolving <https://gcc.gnu.org/PR105914> "gccrs setting warn_unused_variable breaks thousands of non-Rust tests".  I do understand what the problem is (details to be shared later), and how to resolve it.  And, I do have a set of patches for GCC options machinery, to be upstreamed separately, to make it less likely that the same issue appears again.

I do understand that you'd like `-Wunused-variable`, `-Wunused-const-variable=1` enabled by default.  (As implemented here, via `Init(1)`.)  Now I need to know: (1) should that be enforced unconditionally (that is, no way to disable those diagnostics), or (2) should it be possible to specify `-Wno-unused-variable` etc. to disable the diagnostics?  (Natural for GCC would be the latter, (2).)

PR #994, opened 2022-03-07
Parse macro patterns properly in repetitions

### LINKED ISSUE #966: Parse macro pattern properly in substitute_repetitions
Instead of simply consuming tokens until a closing parenthesis, we need to parse repetition patterns properly and add test cases for edge cases

CohenArthur:
Closes #966 

We actually cannot reuse functions from the parser since we're expanding a macro transcriber. This is fine as the "algorithm" is extremely simple


gcc/rust/expand/rust-macro-substitute-ctx.cc
```diff
@@ -154,19 +151,56 @@ SubstituteCtx::substitute_token (size_t token_idx)
...
-	       && macro.at (pattern_end)->get_id () != RIGHT_PAREN;
-	     pattern_end++)
-	  ;
+	auto parentheses_stack = 0;
+	for (size_t idx = pattern_start; idx < macro.size (); idx++)
+	  {
+	    if (macro.at (idx)->get_id () == LEFT_PAREN)
+	      parentheses_stack++;
+
+	    if (macro.at (idx)->get_id () == RIGHT_PAREN)
```
philberty, line 183: should this be an else if?

philberty (approved):
LGTM great to see you are always progressing forward

gcc/rust/expand/rust-macro-substitute-ctx.cc
CohenArthur: Right! Fixed in [08b7516](https://github.com/Rust-GCC/gccrs/pull/994/commits/08b7516191a62e53819e24d08f2a413c84cf4925)

philberty (approved):

PR #997, opened 2022-03-08
macros: Allow any delimiters for invocation

### LINKED ISSUE #946: All outer delimiters are valid for macro matching
I tried this code:

```rust
macro_rules! delimiters {
    (()) => { }
}

fn main() {
    delimiters!(());
    delimiters!{()};
    delimiters![()];
}

```

I expected to see this happen: no errors are emitted.

Source:
> Tokens that aren't part of such an invocation are matched and transcribed literally, with one exception. The exception is that the outer delimiters for the matcher will match any pair of delimiters. Thus, for instance, the matcher (()) will match {()} but not {{}}. The character $ cannot be matched or transcribed literally.

From https://doc.rust-lang.org/reference/macros-by-example.html

Instead, this happened: 

Two kinds of errors:
```rust
test.rs:7:5: error: Failed to match any rule within macro
    1 | macro_rules! delimiters {
      | ~
......
    7 |     delimiters!{()};
      |     ^
test.rs:8:5: error: Failed to match any rule within macro
    1 | macro_rules! delimiters {
      | ~
......
    8 |     delimiters![()];
      |     ^
```
meaning that not all delimiters are accepted, regardless of the macro definition.

And
```rust
test.rs:2:15: error: found unexpected token ‘}’ in null denotation
    2 |     (()) => { }
      |      
```
on the first invocation

commit: 9fb06d66cef

CohenArthur:
Closes #946

### DIFF
--- a/gcc/rust/expand/rust-macro-expand.cc
+++ b/gcc/rust/expand/rust-macro-expand.cc
@@ -3493,26 +3493,30 @@ MacroExpander::match_matcher (Parser<MacroInvocLexer> &parser,
       return false;
     }
 
+  auto delimiter = parser.peek_current_token ();
+
   // this is used so we can check that we delimit the stream correctly.
-  switch (matcher.get_delim_type ())
+  switch (delimiter->get_id ())
     {
-      case AST::DelimType::PARENS: {
+      case LEFT_PAREN: {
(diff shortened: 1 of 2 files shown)


philberty (approved):
This confused me when i started out but I remember thinking this seemed more correct

PR #1003, opened 2022-03-09
Add more intrinsics and refactor how we implement them

philberty:
This patch series implements:

1. offset
2. size_of
3. unreachable
4. abort

It removes the GCC wrapper mappings to make them much easier to implement. It also demonstrates in single commits
the implementation of each of these intrinsic to make it easy to follow in how we implement them.

Addresses #658 #849

### DIFF
--- a/gcc/rust/backend/rust-compile-intrinsic.cc
+++ b/gcc/rust/backend/rust-compile-intrinsic.cc
@@ -113,6 +113,10 @@ private:
 		    "__builtin_unreachable", NULL,
 		    build_function_type (void_type_node, void_list_node),
 		    builtin_const | builtin_noreturn);
+
+    define_builtin ("abort", BUILT_IN_ABORT, "__builtin_abort", "abort",
+		    build_function_type (void_type_node, void_list_node),
+		    builtin_const | builtin_noreturn);
   }
 
   // Define a builtin function.  BCODE is the builtin function code
--- a/gcc/rust/backend/rust-compile-intrinsic.cc
+++ b/gcc/rust/backend/rust-compile-intrinsic.cc
@@ -15,12 +15,151 @@
 // <http://www.gnu.org/licenses/>.
 
 #include "rust-compile-intrinsic.h"
+#include "langhooks.h"
 #include "rust-compile-type.h"
 #include "rust-compile-fnparam.h"
 #include "rust-tree.h"
 
 namespace Rust {
 namespace Compile {
+
+// https://github.com/rust-lang/rust/blob/master/library/core/src/intrinsics.rs
+// https://github.com/rust-lang/rust/blob/master/compiler/rustc_codegen_llvm/src/intrinsic.rs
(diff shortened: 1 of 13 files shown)


philberty:
I would appreciate any GCC folks looking at this. This is part of my set of patches to get slices working, just so you know for context. It would be easier to look at each individual commit rather than the full diff for reviewing purposes.

@tschwinge @dkm @dafaust

gcc/rust/backend/rust-compile-intrinsic.cc
```diff
@@ -15,10 +15,29 @@
...
+{
+  std::string identifier;
+  generic_intrinsic_handler handler;
+};
+static const BuiltinGenericIntrinsic generic_intrinsics[]
+  = {{"offset", &offset_intrinsic_handler}};
+
+#define GENERIC_INTRINSIC_SIZE                                                 \
+  sizeof (generic_intrinsics) / sizeof (BuiltinGenericIntrinsic)
+
```
CohenArthur, line 192: I think you can make this a little more "Cplusplussy"

```suggestion
static tree
offset_intrinsic_handler (Context *ctx, TyTy::BaseType *fntype);

static const std::map<std::string, std::function<tree (Context *, TyTy::BaseType *)>> generic_intrinsics[]
  = {{"offset", offset_intrinsic_handler}};
```
Same exact semantics for defining them but shorter + you get faster lookup + a `.size()` method to avoid the macro

gcc/rust/backend/rust-compile-intrinsic.cc
```diff
@@ -82,11 +101,128 @@ Intrinsics::compile (TyTy::FnType *fntype)
...
 
+  // is it an generic builtin?
+  for (size_t i = 0; i < GENERIC_INTRINSIC_SIZE; i++)
+    {
+      const BuiltinGenericIntrinsic &intrinsic = generic_intrinsics[i];
+      bool found
+	= intrinsic.identifier.compare (fntype->get_identifier ()) == 0;
+      if (found)
+	return intrinsic.handler (ctx, fntype);
+    }
```
CohenArthur, line 214:
```suggestion
// is it a generic builtin?
auto it = generic_intrinsics_map.find (fntype->get_identifier ());

if (it != generic_intrinsics_map.end ())
   return it->second (ctx, fntype);
```

philberty: yeah this sounds good i forget you can statically initialize a std::map

gcc/rust/backend/rust-compile-intrinsic.cc
philberty: fixed

gcc/rust/backend/rust-compile-intrinsic.cc
philberty: fixed

CohenArthur (approved):
Looks good to me! Waiting for the experts' opinions :)

PR #1015, opened 2022-03-11
Add code generation for the slice type

philberty:
This type must respect the layout of the FatPtr type in libcore. Rust
implements slices using Rustc types in libcore and uses a neat trick.

Addresses #849


CohenArthur (approved):
This is cool!

PR #1017, opened 2022-03-11
attr-visitor: Split in its own source and header

CohenArthur:
Split up the 4000 lines rust-macro-expand.cc file containing the
AttrVisitor class and the macro expander implementation


philberty (approved):
Brilliant we need to keep doing this, it all helps to improve incrmental compilation :D

PR #1020, opened 2022-03-14
macros: Lower stmt macro definitions to empty statements

### LINKED ISSUE #1001: ICE when lowering macro definition in block
I tried this code:

```rust
fn main() {
    macro_rules! create_type {
        ($s:ident) => {
            struct $s;
        };
    }
}
```

I expected to see this happen: compilation succeeds

Instead, this happened: ICE when lowering block statements

Since the block contains a macro definition, it tries lowering it using the `ASTLoweringStmt` visitor. And since macro definitions currently do not lower to anything, this fails.

The temporary fix is to move the macro definition outside of the block, so it becomes a crate item. I'm unsure if this is a trail for fixing this

gccrs sha: 865b6090a8f8981cdfc050ea2ee44abbe92de141

CohenArthur:
This commit simply removes the ICE happening when lowering a macro
defined inside a block by lowering it to an empty statement. We do need
to figure out how to lower them properly to the corresponding .rlib and
.rmeta files later on

Fixes #1001

### DIFF
--- a/gcc/rust/hir/rust-ast-lower-stmt.h
+++ b/gcc/rust/hir/rust-ast-lower-stmt.h
@@ -445,6 +445,19 @@ public:
     translated = fn;
   }
 
+  void visit (AST::MacroRulesDefinition &macro_def) override
+  {
+    // FIXME: This lowers macro definitions as statements (i.e. macros defined
+    // in blocks) as an empty statement: We actually need to perform
+    // lowering of the macro and figure out how to store it using rust's
+    // metadata formats: .rlib and .rmeta files
+    auto crate_num = mappings->get_current_crate ();
+    Analysis::NodeMapping mapping (crate_num, macro_def.get_node_id (),
+				   mappings->get_next_hir_id (crate_num),
+				   UNKNOWN_LOCAL_DEFID);
+    translated = new HIR::EmptyStmt (mapping, macro_def.get_locus ());
+  }
+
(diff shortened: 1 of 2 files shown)


gcc/rust/hir/rust-ast-lower-stmt.h
```diff
@@ -445,6 +445,19 @@ class ASTLoweringStmt : public ASTLoweringBase
     translated = fn;
   }
 
+  void visit (AST::MacroRulesDefinition &macro_def) override
+  {
+    // FIXME: This lowers macro definitions as statements (i.e. macros defined
```
philberty, line 450: good comment. Do we need this in the item scope? I guess it must be more permissive there.

philberty (approved):
LGTM

philberty:
Actually I just left a comment over in https://github.com/Rust-GCC/gccrs/issues/1001 we should probably fix this slightly differently.

gcc/rust/hir/rust-ast-lower-stmt.h
CohenArthur: It is more permissive in the item scope, the crate visitor does not check if an item has "not" been lowered to HIR. Block on the other hand will error out if one of their statements gets lowered to a null pointer.

CohenArthur:
Closing in favor of #1032

PR #1026, opened 2022-03-14
Enable -Werror

### LINKED ISSUE #694: Update the CI to use -Werror
We are using github automations for our PR's. Now that we have working bootstrap build lets avoid regressions here by adding -Werror to our CI

This is how we configure our build for CI automation in PR's

https://github.com/Rust-GCC/gccrs/blob/28f527c9598339cf834a30b5ee1f14258b8ecbb2/.github/workflows/ccpp.yml#L37-L44

Let's update this to add --enable-werror to catch warnings breaking the build again.

CastilloDel:
Fixes #694 

- \[ ] Run `clang-format`

The last three ones shouldn't be necessary for this change.

---

Update the CI to use the bootstrap build process and enable -Werror

Signed-off-by: Daniel del Castillo <delcastillodelarosadaniel@gmail.com>


CastilloDel:
But didn't @/tschwinge said gcc is only supposed to be warning free when using the bootstrap process? I'm talking about [this comment](https://github.com/Rust-GCC/gccrs/issues/694#issuecomment-929347162)

dkm:
Bootstrap is not wanted for every PR (waste of time/resources), and `--enable-werror` is only meaningful when bootstrap is enabled, if the doc is correct (from your link in the related discussion). So, not really sure this would be useful... ?

CastilloDel:
I don't really know, I just wanted to fix the issue. But maybe it could be useful in `bootstrap.yml` which doesn't seem to run as often as `ccpp.yml`, but does have bootstrap enabled.

CohenArthur:
@CastilloDel We want to avoid regressions by making sure that our bootstrapped build does not stop working silently at some point. We can do this two ways:

1. by adding `-Werror` to our build flags in ccpp.yml. We cannot use `--enable-werror` here as, as you pointed out, it is only enabled when bootstrapping is also enabled. We don't want to bootstrap the compiler every time we create a pull request, as this takes an extremely long time (and our CI already takes something like an hour to run). We can add the -Werror flag otherwise, by specifying `CXXFLAGS='-Werror'` in the configure and building step for example. I haven't looked into how to do it here but that's typically what you would do.
2. by adding `--enable-werror` to the bootstrap workflow (bootstrap.yml). This workflow only triggers when changes are pushed to master, so after a PR has been reviewed and approved and merged. This has the disadvantage of not preventing regressions but simply showing that master's build failed when arriving on the repository.

I personally think the first option would be preferable. To show you that you can't use `--enable-werror` I've added a warning on my clone, by changing `feature.c_str()` to `feature` (effectively giving a C++ `std::string` where a C-equivalent `const char *` is expected)
```c++
../../gcc/rust/rust-session-manager.cc:1011:19: warning: format ‘%s’ expects argument of type ‘char*’, but argument 3 has type ‘std::string’ {aka ‘std::__cxx11::basic_string<char>’} [-Wformat=]
 1011 |       rust_debug ("had to implicitly enable feature '%s'!", feature);
      |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../gcc/rust/rust-diagnostics.h:143:54: note: in definition of macro ‘rust_debug’
  143 | #define rust_debug(...) rust_debug_loc (Location (), __VA_ARGS__)
      |                                                      ^~~~~~~~~~~
../../gcc/rust/rust-session-manager.cc:1011:55: note: format string is defined here
 1011 |       rust_debug ("had to implicitly enable feature '%s'!", feature);
      |                                                      ~^
      |                                                       |
      |                                                       char*
```
I configured the build directory like so:
`./configure --enable-werror --disable-bootstrap --disable-multilib --enable-languages=c,rust`
And the build still succeeded, emitting a warning but not treating is as an error. So your solution does not do what you want

CastilloDel:
I did the first option, but now it doesn't compile :(. Doesn't this have to do with what @/tschwinge said in the issue and I referenced before? I understood that the build process isn't warning free when the bootstrap is disabled.

CohenArthur:
I'd be fine with the second solution too. The only thing I don't like is that it can fail silently. You can also maybe look into how to add an extra step for our merging bot, bors, so that it runs an extra file before merging to master. This way contributors that open PRs still have reasonably fast feedback where bootstrapping is disabled, but bors performs a full check before merging it to master?

Daniel-del-Castillo:
I think that makes sense. I'll have a look

tschwinge:
I think all that has been said here is correct...  ;'-)

I thought I'd seen this suggested somewhere, but now I can't find it anymore: to just enable `-Werror` for files in `gcc/rust/` -- that doesn't work, because these `#include` generic GCC files, and warning may be coming from those.

---

Just to throw in yet another option (but I'm not convinced that I like it...): bors is running in a mostly stable setup (meaning, here: known GCC version used to build GCC/Rust).  We could generate a *baseline expected warnings file* containing `grep 'warning: ' | sort` of the build log, to document the baseline set of warning that we get/expect with the GCC version that bors uses to build GCC/Rust.  Now, after a successful build, the ongoing bors CI could again do a `grep 'warning: ' | sort` of the *current* build log, and `diff -U0` that one against the *baseline expected warnings file*.  If that's clean, no new warnings have been introduced, otherwise, report the difference to the user, like a build failure.

That way, we don't get any noticeable overhead for the bors CI runs, and the only downside is that we have to update the *baseline expected warnings file* if the bors GCC version changes, and when doing merges from GCC upstream.  (... which seems reasonable.)

CohenArthur:
@CastilloDel to debug the CI failure I can recommend the tmate github action! I think in your case it might just be because of the missing newline at the end of the `expected_warnings` file and an extra libtool warning

.github/workflows/ccpp.yml
```diff
@@ -46,7 +46,10 @@ jobs:
     - name: Build
       run: |
            cd gccrs-build; \
-           make -j $(nproc)
+           make -j $(nproc) > log 2>&1;
+           cat log;
+           cat log | grep 'warning:' > warnings;
+           diff warnings ../expected_warnings
```
tschwinge, line 52: As the warnings may appear in random order, I suppose we really need to `sort` them, as I had mentioned.  (Then also `sort` the baseline file.)

May want to address <https://en.wikipedia.org/wiki/Cat_(Unix)#Useless_use_of_cat>.

I suggest to rename the `warnings` file to `log-warnings` or similar.

I suggest to rename the `expected_warnings` file to `bors-log-expected_warnings` or similar, so as to give more context what this file is about -- as it'll live in the GCC/Rust repository.

I suggest to swap the `diff` arguments: convention is to show old vs. new, here: expected vs. actual.

Instead of plain `diff` I suggest `diff -U0`, so that we get the standard uni-diff header printed (so that users recognize what the following output is about), but without any uni-diff context (as that's not useful here). 

As the `make` now no longer is the last command, we'll loose its exit status (relevant in case of build failure), so we'll have to tune the shell scripting to preserve that.

tschwinge (changes requested):
@CastilloDel, aha, so you're indeed experimenting with my `diff` suggestion, nice.  :-)

I think I convinced myself to "like" it sufficiently.  ;-)

tschwinge:
> extra libtool warning

Now that's of course a (solvable) problem if the warnings are not deterministic...

.github/workflows/ccpp.yml
tschwinge: @CastilloDel, please let us know if you'd like help with the shell scripting -- not everyone's favorite, I heard.  ;-)

dkm: If using `sort` you may need to force the locale to have a stable sort across the CI and dev's hosts?

tschwinge: @dkm, I was expecting this to only run with bors "in the could", which I assume to have deterministic setup/behavior.  But yeah, for local use (and also local updating of the `bors-log-expected_warnings`!), doing this with `LC_ALL=C` or similar would make a lot of sense, yes.

tschwinge: > As the `make` now no longer is the last command, we'll loose its exit status (relevant in case of build failure), so we'll have to tune the shell scripting to preserve that.

Actually, not sure about that -- I don't know how these workflow files are interpreted, exactly.

What should work, is to simply move the `cat log` into its own *step*, and move the `grep` and `diff` into another *step*?

dkm: > @dkm, I was expecting this to only run with bors "in the could", which I assume to have deterministic setup/behavior. But yeah, for local use (and also local updating of the `bors-log-expected_warnings`!), doing this with `LC_ALL=C` or similar would make a lot of sense, yes.

Yes, I was thinking about what would happen when someone needs to update the refs. A bit like what happened with clang-format behaving differently btw CI and local.

CastilloDel:
Okay. I read all the comments (you guys have been really helpful!) and now I think I didn't miss anything and it should work.

CohenArthur (changes requested):
This looks nice! I think it would be better to keep the `bors_log_expected_warnings` file inside the `.github/` directory: When we try to upstream the project, this a directory we're sure to skip, and this file only has a purpose for github actions, meaning it shouldn't get upstreamed either

CastilloDel:
Done!

CohenArthur (approved):
LGTM

tschwinge:
@CastilloDel , I'm looking into this now, thanks!

Still would like to tune a few things, documentation, and try to find a way to make this usable for local "workflows" (standard `configure && make`, etc.).

Otherwise, I'll now have to route any merges from upstream GCC (#247) through bors/CI.  (Which would be fine, too.)

First I've now got some learning to do about <https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions>, etc.  ;-)

I see "Add more commits by pushing to the **[master](https://github.com/CastilloDel/gccrs)** branch on **[CastilloDel/gccrs](https://github.com/CastilloDel/gccrs)**", so for simplicity, I'll use that branch and this PR #1026 for my work.  (Hope you don't mind...)  ;-D

.github/workflows/ccpp.yml
```diff
@@ -46,7 +46,15 @@ jobs:
     - name: Build
       run: |
            cd gccrs-build; \
-           make -j $(nproc)
+           LC_ALL=C make -j $(nproc) > log 2>&1;
+           
+    - name: Build logs
+      run: cat gccrs-build/log
```
tschwinge, line 52: May have to tune this: I suppose (but have not verified!) that in case that `make` fails, the whole `jobs.build-and-check` stops, and *Build logs* isn't executed, and thus there's no indication in the GitHub UI why `make` failed.

CastilloDel: Yes, that's true. I didn't notice :( 

Maybe we could do something like `LC_ALL=C make -j $(nproc) > log 2>&1 && cat gccrs-build/log` and then remove the separate task. If `make` fails the whole command should fail.

tschwinge: Right, working on that, will push in a moment -- thanks for confirming!

.github/workflows/ccpp.yml
```diff
@@ -46,7 +50,15 @@ jobs:
...
            cd gccrs-build; \
-           make -j $(nproc)
+           make -j $(nproc) > log 2>&1
+           
+    - name: Build logs
+      run: cat gccrs-build/log
+
+    - name: Check for new warnings
+      run: |
+           grep 'warning:' gccrs-build/log | sort > log_warnings;
```
tschwinge, line 60: This *step* will fail in case that `grep` doesn't find a single *warning*.  We shall defer that latent problem until relevant...  ;-)

dkm:
> @dkm, ah good point. Indeed that's currently not implemented, and `.github/bors_log_expected_warnings` needs to be updated manually, according to the `diff -U0` output in the CI log/web page. (But I suppose feeding that into a local `patch -R .github/bors_log_expected_warnings` should in fact work?) I don't know how to get data out of CI.

I guess if we can copy/paste and use `patch` that's already good enough :). That should be very occasional, so no need to dig it further!

.github/workflows/ccpp.yml
```diff
@@ -52,11 +52,17 @@ jobs:
...
     - name: Check for new warnings
       run: |
-           grep 'warning:' gccrs-build/log | sort > log_warnings;
-           diff -U0 .github/bors_log_expected_warnings log_warnings 
+           cd gccrs-build
+           < log grep 'warning: ' | sort > log_warnings
+           if diff -U0 ../.github/bors_log_expected_warnings log_warnings; then
+               :
+           else
+               echo 'See <https://github.com/Rust-GCC/gccrs/pull/1026>.'
```
CastilloDel, line 63: I really like this! This way it will be clear why it failed

tschwinge:
bors: merge

philberty (approved):
This is amazing work guys!

tschwinge:
bors: merge

... once again.

PR #1028, opened 2022-03-16
Add remaining contexts for macro expansion

CohenArthur:

PR #1029, opened 2022-03-16
Macro in trait impl

CohenArthur:
Needs #1028 

You can just review the last commit to avoid reviewing twice. Sorry about that!

### DIFF
--- a/gcc/rust/expand/rust-attribute-visitor.h
+++ b/gcc/rust/expand/rust-attribute-visitor.h
@@ -40,6 +40,54 @@ public:
   void expand_trait_function_decl (AST::TraitFunctionDecl &decl);
   void expand_trait_method_decl (AST::TraitMethodDecl &decl);
 
+  /**
+   * Expand a set of values, erasing them if they are marked for strip, and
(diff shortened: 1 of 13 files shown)


gcc/rust/expand/rust-attribute-visitor.h
```diff
@@ -40,19 +40,70 @@ class AttrVisitor : public AST::ASTVisitor
...
+   * @param extractor Function to call when replacing values with the content
+   * 		of an expanded AST node
+   */
+  template <typename T, typename U>
+  void expand_macro_children (MacroExpander::ContextType ctx, T &values,
+			      std::function<U (AST::SingleASTNode)> extractor)
   {
+    expander.push_context (ctx);
+
     for (auto it = values.begin (); it != values.end ();)
```
philberty, line 61: is there a way we could rewrite this for loop so it always continues to the  end even if we miss incremenentations in the body?

If you hit a case in the fort loop where you would want the iteratator not to move forward you could always issue it--;

CohenArthur:
Retrying after outage?

bors r+

PR #1030, opened 2022-03-16
Rewrite our unconstrained type-param error checking

### LINKED ISSUE #1019: Bad unconstrained type parameter error
I tried this code:

```rust
trait A<T> {                                                                                                         
    type Output;                                                                                                                         
                                                                        
    fn test(self, a: &T) -> &Self::Output;
}

struct Foo<T>{
    start: T,
    end: T,
}

impl<T> A<T> for Foo<usize> {
    type Output = T;

    fn test(self, a: &T) -> &Self::Output {
        a
    }
}
```

Instead, this happened:

```
<source>:12:6: error: unconstrained type parameter
   12 | impl<T> A<T> for Foo<usize> {
      |      ^
```

see: https://godbolt.org/z/aK4T4xEGq

### Meta

- What version of Rust GCC were you using, git sha if possible. b82408fd6a17762fcb6fc7b5efe9307a3e59f8b7

philberty:
This is a series of patches that were all required to fix this issue. We
now take advantage of our substitutions abstractions and traits
so that our TypeBoundPredicate's which form the basis of our HRTB code
I think this class is almost akin to rustc existential-trait-references. This now
reuses the same code path to give us the same error checking for generics
as we get with ADT's, functions etc.

With this refactoring in place we can then reuse the abstractions to map the
ID's from the used arguments in the type-bound-predicate, the impl block type
substation mappings and the self type itself.

There are quite a few cases to handle and our testsuite picked up all the regressions
so no behaviour of our existing test-cases have changed now. See each commit for
more detailed information.

Fixes #1019 
Addresses #849

### DIFF
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -99,6 +99,7 @@ GRS_OBJS = \
     rust/rust-hir-type-check-pattern.o \
     rust/rust-hir-type-check-expr.o \
     rust/rust-hir-dot-operator.o \
+    rust/rust-hir-type-check-base.o \
     rust/rust-autoderef.o \
     rust/rust-substitution-mapper.o \
     rust/rust-lint-marklive.o \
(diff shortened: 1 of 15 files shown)


CohenArthur (approved):
LGTM!

PR #1032, opened 2022-03-17
Add AST kind information

### LINKED ISSUE #1001: ICE when lowering macro definition in block

CohenArthur:
Closes #1001 

This PR adds a base for adding node information to our AST types. It can be used when requiring to differentiate between multiple kinds of nodes, while not necessarily wanting to do a full static cast. This will open up a lot of cleanup issues and good first issues for Project Pineapple

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -41,6 +41,7 @@ enum Kind
 {
   UNKNOWN,
   MACRO_RULES_DEFINITION,
+  MACRO_INVOCATION,
 };
 
 // Abstract base class for all AST elements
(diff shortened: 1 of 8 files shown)


philberty (approved):
LGTM

PR #1035, opened 2022-03-17
Handle -fsyntax-only

CohenArthur:
Handle the -fsyntax-only properly from the rust frontend. This flag
allows checking for syntax and stopping after that, skipping further
passes of the pipeline.
The flag was accepted by the frontend, but was not used anywhere.

### DIFF
--- a/gcc/rust/rust-session-manager.cc
+++ b/gcc/rust/rust-session-manager.cc
@@ -542,6 +542,11 @@ Session::parse_file (const char *filename)
 
   rust_debug ("\033[0;31mSUCCESSFULLY PARSED CRATE \033[0m");
 
+  // If -fsyntax-only was passed, we can just skip the remaining passes.
+  // Parsing errors are already emitted in `parse_crate()`
+  if (flag_syntax_only)
+    return;
+
   // register plugins pipeline stage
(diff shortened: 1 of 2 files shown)


gcc/rust/rust-session-manager.cc
```diff
@@ -506,6 +506,11 @@ Session::parse_file (const char *filename)
   // generate crate from parser
   auto parsed_crate = parser.parse_crate ();
 
+  // If -fsyntax-only was passed, we can just skip the remaining passes.
+  // Parsing errors are already emitted in `parse_crate()`
```
philberty, line 510: it might be worth moving this beflow the point where we dump the last so its still possible to get an ast dump

philberty (changes requested):

gcc/rust/rust-session-manager.cc
CohenArthur: Done in [e30d07d](https://github.com/Rust-GCC/gccrs/pull/1035/commits/e30d07de5c89e0f8a4d8304846e21f8e775c9f79)

philberty (commented):
nice one

philberty (approved):

PR #1037, opened 2022-03-17
Support placeholders becoming slices

### LINKED ISSUE #1034: Bad placeholder type handling
I tried this code:

```rust
trait Foo<T> {
    type Output;

    fn test(self, slice: &T) -> &Self::Output;
}

struct Bar<T>(T);
impl<T> Foo<[T]> for Bar<usize> {
    type Output = [T];

    fn test(self, slice: &[T]) -> &[T] {
        slice
    }
}
```

Instead, this happened:

```
test.rs:9:19: error: expected [<placeholder:>] got [[T=T REF: 42]]
    2 |     type Output;
      |     ~              
......
    9 |     type Output = [T];
      |                   ^
test.rs:9:5: error: type alias ‘Output’ has an incompatible type for trait ‘Foo’
    2 |     type Output;
      |     ~
......
    9 |     type Output = [T];
      |     ^
```

### Meta

- What version of Rust GCC were you using, git sha if possible. 0977ef967a25d373a6693231cbcc2b69836dda83

philberty:
When we setup trait-impls the type-alias are allowed to become any type
this interface was missing a visitor. We also need to support constraining
type-parameters behind slices.

The get_root interface is currently unsafe, it needs a flag for allowing
unsized and for keeping a map of adjustments along the way. This will
be added down the line when we support unsized method resolution.

Fixes #1034
Addresses #849

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-expr.h
+++ b/gcc/rust/typecheck/rust-hir-type-check-expr.h
@@ -258,7 +258,6 @@ public:
     // Get the adjusted self
     Adjuster adj (receiver_tyty);
     TyTy::BaseType *adjusted_self = adj.adjust_type (candidate.adjustments);
-    adjusted_self->debug ();
 
     // store the adjustments for code-generation to know what to do
     context->insert_autoderef_mappings (expr.get_mappings ().get_hirid (),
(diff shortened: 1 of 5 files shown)


CohenArthur (approved):
I guess we should probably open up issues about adding the new visitor so it doesn't get forgotten

philberty:
I've created #1038

PR #1040, opened 2022-03-18
Do not propagate parse errors in match repetitions

### LINKED ISSUE #958: Valid macro fails rule matching
```rust
extern "C" {
macro_rules! any {
    ($($a:literal)* $b:ident) => {
        /* snip */
    };
}

fn main() {
    any!(14 15 "hey" b);
}
```
errors out with
```rust
test.rs:25:5: error: Failed to match any rule within macro
    7 | macro_rules! any {
      | ~
......
   25 |     any!(14 15 "hey" b);
      |     ^
```
while rustc matches it properly

CohenArthur:
Since parsing repetitions is very eager, the parser might accumulate
bogus errors by trying to match more repetitions than there are. We can
avoid this by clearing the parsing errors if parsing repetitions
returned a valid result. This should not be an issue for previous
matchers erroring out, as they would immediately return upon failure and
not reach inside other match functions.

We need to figure out the best way to emit parser errors, as we do not always want to emit them in `match_fragment`. I think for now the easiest is to just *not* emit parse errors and simply error out with "failed to match macro rule". We will need to think about adding a bunch of hints too in order to make using macros easier.

Fixes #958

### DIFF
--- a/gcc/rust/expand/rust-macro-expand.cc
+++ b/gcc/rust/expand/rust-macro-expand.cc
@@ -505,9 +505,6 @@ MacroExpander::match_fragment (Parser<MacroInvocLexer> &parser,
       return false;
     }
 
-  for (const auto &error : parser.get_errors ())
-    error.emit_error ();
-
   // it matches if the parser did not produce errors trying to parse that type
   // of item
   return !parser.has_errors ();
@@ -714,7 +711,13 @@ MacroExpander::match_n_matches (Parser<MacroInvocLexer> &parser,
   bool did_meet_lo_bound = match_amount >= lo_bound;
   bool did_meet_hi_bound = hi_bound ? match_amount <= hi_bound : true;
 
-  return did_meet_lo_bound && did_meet_hi_bound;
+  // If the end-result is valid, then we can clear the parse errors: Since
+  // repetitions are parsed eagerly, it is okay to fail in some cases
+  auto res = did_meet_lo_bound && did_meet_hi_bound;
+  if (res)
+    parser.clear_errors ();
+
+  return res;
 }
 
 bool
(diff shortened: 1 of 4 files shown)


gcc/rust/expand/rust-macro-expand.cc
```diff
@@ -714,7 +711,13 @@ MacroExpander::match_n_matches (Parser<MacroInvocLexer> &parser,
   bool did_meet_lo_bound = match_amount >= lo_bound;
   bool did_meet_hi_bound = hi_bound ? match_amount <= hi_bound : true;
 
-  return did_meet_lo_bound && did_meet_hi_bound;
+  // If the end-result is valid, then we can clear the parse errors: Since
+  // repetitions are parsed eagerly, it is okay to fail in some cases
+  auto res = did_meet_lo_bound && did_meet_hi_bound;
+  if (res)
+    parser.clear_errors ();
+
```
philberty, line 719: do we need an else case to emit errors here now?

philberty (approved):

gcc/rust/expand/rust-macro-expand.cc
CohenArthur: I'm not sure if we want to emit errors here: Since matching the repetition failed, it might simply mean that we're matching on the wrong Macro rule:
```rust
macro_rules! m {
    ($($a:literal),*) => {};
    ($($a:literal)tok*) => {};
}

m!(15 tok 14 tok 13);
```
would fail the first rule, but match the second one. And we don't want parse errors to be emitted, we simply want to silently return false and check if the next rule matches

philberty: ah ok i see what you mean now when I read the code again.

philberty (approved):

PR #1041, opened 2022-03-18
macros: Only expand merged repetitions if they contain the same amount

### LINKED ISSUE #948: Merged repetitions must refer to the same amount of fragments
```rust
macro_rules! repeat {
    ( $( $i:literal ),* ; $( $j:literal ),* ) => (( $( ($i,$j) ),* ))
}

fn main() {
    let a = repeat!(1, 2, 3; 2, 3);
}
```
should error out since `$i` repeats 3 times, but `$j` repeats only twice.

> Each repetition in the transcriber must contain at least one metavariable to decide how many times to expand it. If multiple metavariables appear in the same repetition, they must be bound to the same number of fragments

Taken directly from https://doc.rust-lang.org/reference/macros-by-example.html

CohenArthur:
Depends on #1040 
Fixes #948

### DIFF
--- a/gcc/rust/expand/rust-macro-substitute-ctx.cc
+++ b/gcc/rust/expand/rust-macro-substitute-ctx.cc
@@ -30,18 +30,14 @@ SubstituteCtx::substitute_metavar (std::unique_ptr<AST::Token> &metavar)
   return expanded;
 }
 
-std::vector<std::unique_ptr<AST::Token>>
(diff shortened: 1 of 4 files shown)


CohenArthur:
The branch needs to be rebased on #1040 for the tests to pass, but I think this makes for easier reviewing?

philberty (approved):
this looks good

PR #1042, opened 2022-03-18
Parse reserved keywords as valid fragments identifiers

### LINKED ISSUE #1013: Metavar identifiers are not parsed properly
I tried this code:

```rust
macro_rules! add0 {
    ($a:literal) => { $a };
    ($a:literal $($as:literal)+) => { $a + add0!($($as)*) }
}

fn main() {
    let a = add0!(1 2 3 4 67); // no separator
}
```

I expected to see this happen: successful compilation

Instead, this happened: `unrecognised token combination $as`

This is probably due to `as` being a reserved keyword. However, rustc accepts the above code

### Meta

- What version of Rust GCC were you using: 51c7cf4b974

CohenArthur:
Per the reference, macro fragments actually accept all identifiers, not
NON_KEYWORD_IDENTIFIERS

Fixes #1013

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -1832,6 +1832,56 @@ Parser<ManagedTokenSource>::parse_macro_match ()
 	const_TokenPtr t2 = lexer.peek_token (1);
 	switch (t2->get_id ())
 	  {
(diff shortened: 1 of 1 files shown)


philberty (approved):
thats a big list good find

Issue #927, opened 2022-02-15
Builtin macros


CohenArthur:
- [ ] asm
- [ ] global_asm
- [ ] assert
- [ ] cfg_accessible
- [ ] cfg_eval
- [ ] cfg
- [ ] cmdline_attrs
- [ ] compile_error
- [ ] concat_bytes
- [ ] concat_idents
- [ ] concat
- [ ] derive
- [ ] deriving
- [ ] panic
- [ ] unreachable
- [ ] edition_panic
- [ ] env
- [ ] option_env
- [ ] format_args
- [ ] format_args_nl
- [ ] const_format_args
- [ ] format
- [ ] global_allocator
- [ ] log_syntax
- [ ] proc_macro_harness
- [ ] source_util
- [ ] standard_library_imports
- [ ] test_harness
- [ ] test
- [ ] bench
- [ ] trace_macros
- [ ] quote
- [ ] stringify
- [ ] include
- [ ] include_str
- [ ] include_bytes
- [ ] module_path

## File system macros

- [ ] line
- [ ] column
- [ ] file

## Derive macros

- [ ] Copy
- [ ] Clone
- [ ] PartialEq
- [ ] Eq
- [ ] PartialOrd
- [ ] Ord
- [ ] Hash
- [ ] Default

## Deprecated Derive macros

- [ ] RustcEncodable
- [ ] RustcDecodable

bjorn3:
> Deprecated Derive macros

There seems to be a push to destabilize them in rustc. Skipping them in gccrs should be fine I think.

> cmdline_attrs

This is not a macro, but the module implementing `-Zcrate-attr`.

> deriving

This is the module containing all builtin `#[derive]` macros.

> proc_macro_harness

This module is for the proc macro implementation.

> source_util

Haven't checked, but also likely not a macro.

There are a couple other items on the list that I don't think are actual macros.

CohenArthur:
@bjorn3 this was mostly created by looking [here](https://github.com/rust-lang/rust/tree/master/compiler/rustc_builtin_macros/src) and by grepping for `compiler built-in` in the rust repo. I do think it needs some adjustments, so thanks a lot! I'll double check it

liushuyu:
This table probably needs an update.

Since the last update of this issue, the following macros have been implemented:

- compiler_error
- include
- line
- cfg (partially?)

Most remainders require the `format_args!` macro to be implemented (`print!` and `println!` family of macros also depend on this).

[...]

PR #1043, opened 2022-03-18
implement include_bytes! and include_str! macros

dafaust:
Implement the include_bytes! and include_str! builtin macros.

Addresses:  #927

### DIFF
--- a/gcc/rust/expand/rust-macro-builtins.cc
+++ b/gcc/rust/expand/rust-macro-builtins.cc
@@ -20,6 +20,9 @@
 #include "rust-diagnostics.h"
 #include "rust-expr.h"
 #include "rust-session-manager.h"
+#include "rust-macro-invoc-lexer.h"
+#include "rust-lex.h"
+#include "rust-parse.h"
 
 namespace Rust {
 namespace {
@@ -29,6 +32,107 @@ make_string (Location locus, std::string value)
   return std::unique_ptr<AST::Expr> (
     new AST::LiteralExpr (value, AST::Literal::STRING,
 			  PrimitiveCoreType::CORETYPE_STR, {}, locus));
+}
+
+/* Parse a single string literal from the given delimited token tree,
+   and return the LiteralExpr for it. Allow for an optional trailing comma,
+   but otherwise enforce that these are the only tokens.  */
+
+std::unique_ptr<AST::LiteralExpr>
+parse_single_string_literal (AST::DelimTokenTree &invoc_token_tree,
+			     Location invoc_locus)
+{
+  MacroInvocLexer lex (invoc_token_tree.to_token_stream ());
+  Parser<MacroInvocLexer> parser (std::move (lex));
+
+  auto last_token_id = TokenId::RIGHT_CURLY;
+  switch (invoc_token_tree.get_delim_type ())
+    {
+    case AST::DelimType::PARENS:
+      last_token_id = TokenId::RIGHT_PAREN;
+      rust_assert (parser.skip_token (LEFT_PAREN));
+      break;
+
+    case AST::DelimType::CURLY:
+      rust_assert (parser.skip_token (LEFT_CURLY));
+      break;
+
+    case AST::DelimType::SQUARE:
+      last_token_id = TokenId::RIGHT_SQUARE;
+      rust_assert (parser.skip_token (LEFT_SQUARE));
+      break;
+    }
+
+  std::unique_ptr<AST::LiteralExpr> lit_expr = nullptr;
+
+  if (parser.peek_current_token ()->get_id () == STRING_LITERAL)
+    {
+      lit_expr = parser.parse_literal_expr ();
+      parser.maybe_skip_token (COMMA);
+      if (parser.peek_current_token ()->get_id () != last_token_id)
+	{
+	  lit_expr = nullptr;
+	  rust_error_at (invoc_locus, "macro takes 1 argument");
+	}
+    }
+  else if (parser.peek_current_token ()->get_id () == last_token_id)
+    rust_error_at (invoc_locus, "macro takes 1 argument");
+  else
+    rust_error_at (invoc_locus, "argument must be a string literal");
+
+  parser.skip_token (last_token_id);
+
+  return lit_expr;
+}
+
+/* Treat PATH as a path relative to the source file currently being
+   compiled, and return the absolute path for it.  */
+
+std::string
+source_relative_path (std::string path, Location locus)
+{
+  std::string compile_fname
+    = Session::get_instance ().linemap->location_file (locus);
+
+  auto dir_separator_pos = compile_fname.rfind (file_separator);
+
+  /* If there is no file_separator in the path, use current dir ('.').  */
+  std::string dirname;
+  if (dir_separator_pos == std::string::npos)
+    dirname = std::string (".") + file_separator;
+  else
+    dirname = compile_fname.substr (0, dir_separator_pos) + file_separator;
+
+  return dirname + path;
+}
+
+/* Read the full contents of the file FILENAME and return them in a vector.
+   FIXME: platform specific.  */
+
+std::vector<uint8_t>
+load_file_bytes (const char *filename)
+{
+  RAIIFile file_wrap (filename);
+  if (file_wrap.get_raw () == nullptr)
+    {
+      rust_error_at (Location (), "cannot open filename %s: %m", filename);
+      return std::vector<uint8_t> ();
+    }
+
+  FILE *f = file_wrap.get_raw ();
+  fseek (f, 0L, SEEK_END);
+  long fsize = ftell (f);
+  fseek (f, 0L, SEEK_SET);
+
+  std::vector<uint8_t> buf (fsize);
+
+  if (fread (&buf[0], fsize, 1, f) != 1)
+    {
+      rust_error_at (Location (), "error reading file %s: %m", filename);
+      return std::vector<uint8_t> ();
+    }
+
+  return buf;
 }
 } // namespace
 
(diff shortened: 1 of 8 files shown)


dafaust:
Draft for now because these TODOs:
```C++
  /* TODO: filename should be treated as a path relative to the source file.  */
  std::vector<uint8_t> bytes = load_file_bytes (filename);
```
make it hard to add proper tests - the filename is used as relative to the working directory from which gcc is invoked.

CohenArthur:
> These TODOs:
> 
> ```c++
>   /* TODO: filename should be treated as a path relative to the source file.  */
>   std::vector<uint8_t> bytes = load_file_bytes (filename);
> ```
> 
> make it hard to add proper tests - the filename is used as relative to the working directory from which gcc is invoked. Once I work that out I will add test cases. The macros themselves seem to work pretty well.

For that problem specifically I think you should be able to reuse some code from the module loading code, especially [this part](https://github.com/Rust-GCC/gccrs/blob/1bb9a29688ab4ddfec7f8d36ca2cee63c5f258d2/gcc/rust/ast/rust-ast-full-test.cc#L3986-L4000). Thanks a lot for the PR! I'm reviewing it right now

CohenArthur (commented):
Thanks!! Waiting for the tests but this is going to be really useful

dafaust:
> For that problem specifically I think you should be able to reuse some code from the module loading code, especially [this part](https://github.com/Rust-GCC/gccrs/blob/1bb9a29688ab4ddfec7f8d36ca2cee63c5f258d2/gcc/rust/ast/rust-ast-full-test.cc#L3986-L4000). Thanks a lot for the PR! I'm reviewing it right now

Perfect thanks for the pointer :)

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -30,6 +33,70 @@ make_string (Location locus, std::string value)
     new AST::LiteralExpr (value, AST::Literal::STRING,
 			  PrimitiveCoreType::CORETYPE_STR, {}, locus));
 }
+
+/* Parse a single string literal from the given delimited token tree,
+   and return the LiteralExpr for it. Note that this function enforces
+   that the given token tree contains exactly one string literal and
```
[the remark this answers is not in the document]
dafaust: Ah good catch, thanks.

dafaust:
Updated to fix the issues @CohenArthur noted and add tests

P.S. I think this is known but it definitely threw me for a loop :P
```
(gdb) p bytes
$2 = (u8 (&)[16]) @0x7fffffffe060: [104, 101, 108, 108, 111, 44, 32, 105, 110, 99, 108, 117, 100, 101, 33, 
  10]
(gdb) p the_bytes
$3 = (u8 (&)[16]) @0x7fffffffe070: [104, 101, 108, 108, 111, 44, 32, 105, 110, 99, 108, 117, 100, 101, 33, 
  10]
(gdb) p bytes == the_bytes
$4 = false
```

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -30,6 +33,90 @@ make_string (Location locus, std::string value)
...
+
+/* Read the full contents of the file FILENAME and return them in a vector.
+   FIXME: platform specific.  */
+
+std::vector<uint8_t>
+load_file_bytes (const char *filename)
+{
+  RAIIFile file_wrap (filename);
+  if (file_wrap.get_raw () == nullptr)
+    rust_fatal_error (Location (), "cannot open filename %s: %m", filename);
```
CohenArthur, line 103: Any reason for doing a fatal error instead of just an error? I think we could just error out and return an empty vector of bytes

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -30,6 +33,90 @@ make_string (Location locus, std::string value)
...
+  fseek (f, 0L, SEEK_END);
+  long fsize = ftell (f);
+  fseek (f, 0L, SEEK_SET);
+
+  std::vector<uint8_t> buf (fsize);
+
+  fread (&buf[0], 1, fsize, f);
+  int err;
+  if ((err = ferror (f)) != 0)
+    rust_fatal_error (Location (), "error reading file %s: %d: %m", filename,
```
CohenArthur, line 115: Same here imo (but I might be wrong!)

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -63,4 +150,67 @@ MacroBuiltin::column (Location invoc_locus, AST::MacroInvocData &invoc)
...
+  /* Get target filename from the macro invocation, which is treated as a path
+     relative to the include!-ing file (currently being compiled).  */
+  std::string include_path
+    = parse_single_string_literal (invoc.get_delim_tok_tree ())->as_string ();
+  std::string target_filename
+    = source_relative_path (include_path, invoc_locus);
+
+  std::vector<uint8_t> bytes = load_file_bytes (target_filename.c_str ());
+
+  /* TODO: Enforce that the file contents are valid UTF-8?  */
```
CohenArthur, line 209: Yes! I think this is fine for now, but we should definitely open up an issue (and maybe turn this into a FIXME instead). rustc errors out if the included string is not valid UTF-8:

```rust
arthur@platypus ~/G/gccrs (laxer-parser-on-stmt-fragment) [1]> wget https://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-test.txt -O incorrect.utf # snip
arthur@platypus ~/G/gccrs (laxer-parser-on-stmt-fragment)> cat test.rs
fn main() {
    println!("{}", include_str!("incorrect.utf"));
}
arthur@platypus ~/G/gccrs (laxer-parser-on-stmt-fragment)> rustc test.rs
error: incorrect.utf wasn't a utf-8 file
 --> test.rs:2:20
  |
2 |     println!("{}", include_str!("incorrect.utf"));
  |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: this error originates in the macro `include_str` (in Nightly builds, run with -Z macro-backtrace for more info)

```

gcc/testsuite/rust/execute/torture/builtin_macro_include_str.rs
```diff
@@ -0,0 +1,23 @@
...
+}
+
+fn print(s: &str) {
+  printf("%s" as *const str as *const i8, s as *const str as *const i8);
+}
+
+
+fn main() -> i32 {
+  // include_str! (and include_bytes!) allow for an optional trailing comma.
+  let my_str = include_str! ("include.txt",);
```
CohenArthur, line 18: Could you add a test that tries giving an invalid set of arguments to the macro to make sure we don't introduce regressions down the line? Something like
```rust
let file = "file.txt";
include_str!(file); // not a string literal - invalid
include_str!(); // invalid amount of arguments
include_str!("file1.txt", "file2.txt"); // same
```

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -30,6 +33,90 @@ make_string (Location locus, std::string value)
...
+      last_token_id = TokenId::RIGHT_SQUARE;
+      rust_assert (parser.skip_token (LEFT_SQUARE));
+      break;
+    }
+
+  rust_assert (parser.peek_current_token ()->get_id () == STRING_LITERAL);
+  std::unique_ptr<AST::LiteralExpr> lit_expr = parser.parse_literal_expr ();
+
+  parser.maybe_skip_token (COMMA);
+  rust_assert (parser.skip_token (last_token_id));
```
CohenArthur, line 69: If the user passes an incorrect amount of arguments, this will ICE while we could have a nicer formatting imo

CohenArthur (commented):
Thanks a lot for working on this!

CohenArthur:
> Updated to fix the issues @CohenArthur noted and add tests
> 
> P.S. I think this is known but it definitely threw me for a loop :P
> 
> ```
> (gdb) p bytes
> $2 = (u8 (&)[16]) @0x7fffffffe060: [104, 101, 108, 108, 111, 44, 32, 105, 110, 99, 108, 117, 100, 101, 33, 
>   10]
> (gdb) p the_bytes
> $3 = (u8 (&)[16]) @0x7fffffffe070: [104, 101, 108, 108, 111, 44, 32, 105, 110, 99, 108, 117, 100, 101, 33, 
>   10]
> (gdb) p bytes == the_bytes
> $4 = false
> ```

Ah :D I think this is worth raising an issue, it might be simply because we're not compiling libcore yet

gcc/rust/expand/rust-macro-builtins.cc
dafaust: I just followed Session::parse_file:
https://github.com/Rust-GCC/gccrs/blob/cc6e405912c83aee41efd3015d9157cdbe9134fe/gcc/rust/rust-session-manager.cc#L493-L498
without thinking much about it. But you're right, no need to fatal here. Fixed.

gcc/rust/expand/rust-macro-builtins.cc
dafaust: Also fixed.

gcc/rust/expand/rust-macro-builtins.cc
dafaust: OK. Marked as FIXME, will open up an issue pointing at it once this PR is closed.

gcc/testsuite/rust/execute/torture/builtin_macro_include_str.rs
dafaust: Good call. Done.

gcc/rust/expand/rust-macro-builtins.cc
dafaust: Yes, this was silly. We shouldn't ICE. I cleaned up the parsing and error handling here so now we get clear errors instead.

philberty:
> > Updated to fix the issues @CohenArthur noted and add tests
> > P.S. I think this is known but it definitely threw me for a loop :P
> > ```
> > (gdb) p bytes
> > $2 = (u8 (&)[16]) @0x7fffffffe060: [104, 101, 108, 108, 111, 44, 32, 105, 110, 99, 108, 117, 100, 101, 33, 
> >   10]
> > (gdb) p the_bytes
> > $3 = (u8 (&)[16]) @0x7fffffffe070: [104, 101, 108, 108, 111, 44, 32, 105, 110, 99, 108, 117, 100, 101, 33, 
> >   10]
> > (gdb) p bytes == the_bytes
> > $4 = false
> > ```
> 
> Ah :D I think this is worth raising an issue, it might be simply because we're not compiling libcore yet

Thanks for raising this one, I think its because we need to support some of the comparator stuff in libcore. There is a special impl block for slices and traits for comparators so I bet it relates to that. Nice work btw!

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -30,6 +33,110 @@ make_string (Location locus, std::string value)
...
+      return std::vector<uint8_t> ();
+    }
+
+  FILE *f = file_wrap.get_raw ();
+  fseek (f, 0L, SEEK_END);
+  long fsize = ftell (f);
+  fseek (f, 0L, SEEK_SET);
+
+  std::vector<uint8_t> buf (fsize);
+
```
philberty, line 128: It might be a good idea to look up if there is any size limit here in rustc. @CohenArthur found a test case in the rustc testsuite where it tries to create a insane sized array and gccrs tries to do it and runs out of memory lol. But don't do it as part of this PR I think this PR is good to go.

philberty (approved):

philberty:
I think you should have bors rights @dafaust so feel free to merge when @CohenArthur is happy :)

CohenArthur (approved):
Great job!! Thanks a lot! I agree with Philip, tell bors to merge whenever you're ready

gcc/rust/expand/rust-macro-builtins.cc
dafaust: rustc ends up ultimately calling fs::read (), which can error with io::ErrorKind::OutOfMemory. Otherwise I don't don't believe there is any checking on the file size.

https://github.com/rust-lang/rust/blob/c99b42cf14014987a147918fac2ad3d3290b257e/compiler/rustc_builtin_macros/src/source_util.rs#L215

https://github.com/rust-lang/rust/blob/c99b42cf14014987a147918fac2ad3d3290b257e/compiler/rustc_span/src/source_map.rs#L224

CohenArthur:
Ah, maybe not then, or github actions are down again. I'll ask Philip about it if it's a bors-rights issue

bors r+

dafaust:
bors complained because the return value of `fread` was ignored (I used `ferror` to check for error instead). Changed to be more like the other uses of `fread` in gcc and pacify `-Wunused-result`.

Issue #947, opened 2022-02-18
Handle repetition separators properly


CohenArthur:
When repeating tokens using `?+*`, one can add separators to the invocation, i.e:
```rust
macro_rules! repeat {
    ($($any:literal),*) => { ... }
}
```
will match the following: `repeat!(1), repeat!(1, 2, 3), repeat!(1  , 2      , 46)`

- [x] Handle separators properly once repetitions are implemented
- [x] Restrict separators on some fragments
    - [x] `expr` and `stmt` may only be followed by one of: `=>`, `,`, or `;`. #1044 
    - [x] `pat` and `pat_param` may only be followed by one of: `=>`, `,`, `=`, `|`, `if`, or `in`. #1051 
    - [x] `path` and `ty` may only be followed by one of: `=>`, `,`, `=`, `|`, `;`, `:`, `>`, `>>`, `[`, `{`, `as`, `where`, or a macro variable of block fragment specifier. #1051 
    - [x] `vis` may only be followed by one of: `,`, an identifier other than a non-raw priv, any token that can begin a type, or a metavariable with a `ident`, `ty`, or `path` fragment specifier. #1051 

```rust
error: `$a:expr` is followed by `sep`, which is not allowed for `expr` fragments
 --> gcc/testsuite/rust/compile/xfail/macro8.rs:2:16
  |
2 |     ($($a:expr)sep*) => {
  |                ^^^ not allowed after `expr` fragments
  |
  = note: allowed there are: `=>`, `,` or `;`

error: aborting due to previous error
```

From https://doc.rust-lang.org/reference/macros-by-example.html#follow-set-ambiguity-restrictions

- [ ] #1055 

If a repetition can occur zero times, then the set of follow-up restrictions must also be checked for the match arriving afterwards:
```rust
macro_rules! add_with_reserved_kw {
    ($a:expr $(,)? tok) => {{}};
}
```
errors out with
```rust
error: `$a:expr` may be followed by `tok`, which is not allowed for `expr` fragments
 --> test.rs:2:20
  |
2 |     ($a:expr $(,)? tok) => {{}};
  |                    ^^^ not allowed after `expr` fragments
  |
  = note: allowed there are: `=>`, `,` or `;`
```

PR #1044, opened 2022-03-21
Restrict follow-up tokens on `expr` and `stmt`

CohenArthur:
This adds a base for respecting the [Macro Follow-Set Ambiguity specification](https://doc.rust-lang.org/reference/macro-ambiguity.html).

If the design is validated, adding more restrictions on other fragment specifiers should not be difficult

Addresses #947

### DIFF
--- a/gcc/rust/parse/rust-parse.cc
+++ b/gcc/rust/parse/rust-parse.cc
@@ -92,4 +92,145 @@ extract_module_path (const AST::AttrVec &inner_attrs,
 
   return path;
 }
+
+static bool
+peculiar_fragment_match_compatible (AST::MacroMatchFragment &last_match,
+				    AST::MacroMatch &match)
+{
+  static std::unordered_map<AST::MacroFragSpec::Kind, std::vector<TokenId>>
+    follow_set = {
+      {AST::MacroFragSpec::EXPR, {MATCH_ARROW, COMMA, SEMICOLON}},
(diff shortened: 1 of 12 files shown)


gcc/rust/parse/rust-parse.cc
```diff
@@ -92,4 +92,143 @@ extract_module_path (const AST::AttrVec &inner_attrs,
...
+  if (values.empty ())
+    return nullptr;
+
+  return values.front ().get ();
+}
+
+bool
+is_match_compatible (AST::MacroMatch &last_match, AST::MacroMatch &match)
+{
+  AST::MacroMatch *new_last;
```
philberty, line 183: i hate how C++ doesn't support block-expressions. It might be idiomatic here to set this to nullptr and assert if you need to be sure its not null at the end.

philberty (approved):

CohenArthur:
rustc does not perform this analysis during parsing, but afterwards during expansion. I'm wondering if there is a reason for it  @bjorn3 @flip1995 ? I think this is something that can be tackled afterwards if necessary and moved in the macro expansion phase.

flip1995:
> rustc does not perform this analysis during parsing, but afterwards during expansion. I'm wondering if there is a reason for it?

Ambiguity doesn't / might not (?) come from the macro definition, but from the macro usage. Here's an example from the linked document: [godbolt](https://godbolt.org/z/hs1s64PTn).

PR #1045, opened 2022-03-21
Add initial support for unsized method resolution

philberty:
In order to support slices, we end up with an operator overload call of:

```
impl<T, I> Index<I> for [T]
where
    I: SliceIndex<[T]>,
{
    type Output = I::Output;

    fn index(&self, index: I) -> &I::Output {
        index.index(self)
    }
}
```

So this means the self, in this case, is an array[T,capacity] and the index parameter is of type Range<usize>. In order to actually call this method
which has a self parameter of [T] we need to be able to 'unsize' the array
into a slice.

Addresses #849

### DIFF
--- a/gcc/rust/backend/rust-compile-base.h
+++ b/gcc/rust/backend/rust-compile-base.h
@@ -72,6 +72,9 @@ protected:
   tree resolve_indirection_adjustment (Resolver::Adjustment &adjustment,
 				       tree expression, Location locus);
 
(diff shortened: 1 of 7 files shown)


CohenArthur (approved):
This looks good and is readable, but I guess I am misunderstanding its usage. I thought this could help in allowing the following kind of code to work, but I gues not?
```rust
fn takes_slice(a: &[u8]) {}

fn main() {
    let slice = &[1u8, 2u8];
    takes_slice(slice);
}
```

philberty:
bors try

PR #1054, opened 2022-03-23
Fix overzealous follow set ambiguity

### LINKED ISSUE #1053: Invalid follow-set ambiguity error in repetition following :expr
I tried this code:

```rust
macro_rules! p {
    ($e:expr $(,)*) => {};
}
```

I expected to see this happen: no error, an :expr fragment can be followed by a comma

Instead, this happened: gccrs complains about the follow-set ambiguity

### Meta

- What version of Rust GCC were you using: eef1ee2638c

CohenArthur:
When checking if a follow-up is valid, we previously always returned
false when comparing with MacroMatchRepetitions. This is however
invalid, as we should be comparing with the first match of the
repetition to be sure.

Closes #1053 
Addresses #947

### DIFF
--- a/gcc/rust/parse/rust-parse.cc
+++ b/gcc/rust/parse/rust-parse.cc
@@ -100,6 +100,30 @@ contains (std::vector<T> &vec, T elm)
   return std::find (vec.begin (), vec.end (), elm) != vec.end ();
 }
 
+/**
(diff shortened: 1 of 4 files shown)


CohenArthur:
bors try

CohenArthur:
bors try

PR #1056, opened 2022-03-23
Fix '#include <algorithm>' [#159]

tschwinge:
... recently introduced in #1044 commit 35ca685200830626e5abd623f65a850649beace2
"macros: Add base functions to check for follow-set ambiguities".

GCC doesn't like that:

    In file included from [...]
    ./mm_malloc.h:42:12: error: attempt to use poisoned "malloc"
         return malloc (__size);
                ^

See commit e7b3f654f2ab0400c95c0517387a9ad645a5c4cd, for example.

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -20,9 +20,9 @@ along with GCC; see the file COPYING3.  If not see
 /* DO NOT INCLUDE ANYWHERE - this is automatically included with rust-parse.h
  * This is also the reason why there are no include guards. */
 
+#define INCLUDE_ALGORITHM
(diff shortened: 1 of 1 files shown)


CohenArthur (approved):
Oh, I had no idea. Sorry about that! No warnings on my end or on bors's, or even on the buildbot. I'll keep it in mind!

tschwinge:
No worries.  Actually, I also don't know the reason for this.

I had seen that there's more cleanup to be done in that area.  For example, see in `gcc/rust/rust-system.h`: _These must be included before the #poison declarations in system.h._  These also ought to be handled differently (per GCC standards).  But that's for another day.

---

bors: merge

PR #1066, opened 2022-03-27
Fix warning of uninitialized delim_id variable

CohenArthur:


tschwinge (commented):
Is that really the correct logic; or should this instead simply get a `default:` with `gcc_unreachable ()`?

CohenArthur:
> Is that really the correct logic; or should this instead simply get a `default:` with `gcc_unreachable ()`?

You're right... Thanks!

PR #1069, opened 2022-03-28
Handle macro invocations in type contexts

### LINKED ISSUE #1067: Macro invocations do not get handled properly as types
Thanks to @SanderJSA for figuring this out. I've added a minimum reproducible example

I tried this code:

```rust
macro_rules! int {
    () => { i32 }
}

fn f(input: int!()) -> int!() {
    input + 2
}
```

I expected to see this happen: Compilation succeeds, as macro invocations can be done in type positions

https://doc.rust-lang.org/reference/macros.html

Instead, this happened: gccrs is looping infinitely

### Meta

- What version of Rust GCC were you using: 5a15694ee22

CohenArthur:
Closes #1067 

This highlighted two issues where parsing types is not entirely correct, which I'll raise. The code necessary to handle macro invocations in these two places should already be implemented.

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -1800,7 +1800,7 @@ public:
       case IMPL:
 	return "Impl Item: " + impl_item->as_string ();
       case TRAIT_IMPL:
-	return "Trait Impl Item: " + impl_item->as_string ();
+	return "Trait Impl Item: " + trait_impl_item->as_string ();
       case TYPE:
 	return "Type: " + type->as_string ();
       }
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -1827,6 +1827,18 @@ private:
   std::vector<SingleASTNode> nodes;
   bool fragment_is_error;
 
+  /**
+   * We need to make a special case for Expression and Type fragments as only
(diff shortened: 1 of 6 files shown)


gcc/rust/ast/rust-ast.h
```diff
@@ -1840,22 +1868,27 @@ class ASTFragment
...
 
   /**
-   * We need to make a special case for Expression fragments as only one
-   * Node will be extracted from the `nodes` vector
+   * We need to make a special case for Expression and Type fragments as only
+   * one Node will be extracted from the `nodes` vector
    */
 
-  bool is_expression_fragment () const
+  bool is_single_fragment (SingleASTNode::NodeType kind) const
```
philberty, line 1875: Not sure i like this interface.

I think it would be better to have a few helpers here:

```c++
bool is_single_fragment() const { return nodes.size() == 1; }

bool is_single_fragment_kind(SingleASTNode::NodeType kind) const { return is_single_fragment() && nodes.at(0).get_kind() == kind; }
```

philberty (approved):

gcc/rust/ast/rust-ast.h
CohenArthur: I think this function should actually be private, but I agree with you. I'll change the interface

PR #1071, opened 2022-03-29
Allow transcribing of zero nodes in certain cases

CohenArthur:
When expanding AST fragments containing multiple nodes, we must be aware
that some cases allow expanding zero or more nodes. Any macro
transcription that gets parsed as many nodes (ie any transcriber function that calls `parse_many`) needs to be able to parse zero of those nodes and still get expanded properly (basically, removed).

Previously, this would cause a failure to lower the macro invocation which would remain as a child instead of getting stripped/erased.

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -1871,13 +1871,12 @@ public:
     return *this;
   }
 
-  static ASTFragment create_empty () { return ASTFragment ({}); }
   static ASTFragment create_error () { return ASTFragment ({}, true); }
 
   std::vector<SingleASTNode> &get_nodes () { return nodes; }
   bool is_error () const { return fragment_is_error; }
 
-  bool should_expand () const { return !is_error () && !nodes.empty (); }
+  bool should_expand () const { return !is_error (); }
 
   std::unique_ptr<Expr> take_expression_fragment ()
   {
--- a/gcc/rust/ast/rust-macro.h
+++ b/gcc/rust/ast/rust-macro.h
@@ -469,7 +469,7 @@ class MacroRulesDefinition : public MacroItem
   static ASTFragment dummy_builtin (Location, MacroInvocData &)
   {
     gcc_unreachable ();
-    return ASTFragment::create_empty ();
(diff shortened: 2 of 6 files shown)


gcc/rust/ast/rust-ast.h
```diff
@@ -1831,13 +1835,34 @@ class ASTFragment
...
+  {
+    auto fragment = ASTFragment ({node}, false);
+    fragment.fragment_is_one_node = true;
+
+    return fragment;
+  }
+
   std::vector<SingleASTNode> &get_nodes () { return nodes; }
   bool is_error () const { return fragment_is_error; }
+  bool is_one_node () const { return fragment_is_one_node; }
```
philberty, line 1850: does this need to be a flag? What about nodes.size() == 1?

gcc/rust/ast/rust-ast.h
```diff
@@ -1831,13 +1835,34 @@ class ASTFragment
...
+  {
+    // If the fragment is an error fragment, it should not expand
+    if (is_error ())
+      return false;
+
+    // If the fragment contains zero nodes, it might be a fragment we want to
+    // expand! In the case of statements, items, trait impl items and so on, it
+    // is perfectly valid to transcribe zero of those things.
+    if (is_one_node ())
+      return nodes.size () == 1;
```
philberty, line 1862: this seems confusing to me. If we can remove the flag then I think this is fine. But I might be missing something going on. Is this a flag to say its an empty macro invocation but we need to still transcribe? if ```!nodes.empty()```

philberty (changes requested):
I think the is_one_node flag needs looked at here.

If this is to handle a case where its an empty macro invocation but still needs to transcribe it could be a special flag like is_empty_macro because its too confusing with nodes.size() == 1 here.

Or as a last resort a comment surrounding the flag field. Though I still think we need to change that name.

CohenArthur:
@philberty you're right about the flag, it was not necessary. I was trying to still have the possibility to create empty AST fragments and have them be an error sometimes, but it's not necessary. So simply removing the flag and the extra code to handle this special case works perfectly. Sorry about that!

PR #1080, opened 2022-04-03
macros: add compile_error! macro

liushuyu:
- Added `compile_error` macro

### DIFF
--- a/gcc/rust/expand/rust-macro-builtins.cc
+++ b/gcc/rust/expand/rust-macro-builtins.cc
@@ -236,4 +236,20 @@ MacroBuiltin::include_str (Location invoc_locus, AST::MacroInvocData &invoc)
   return AST::ASTFragment ({node});
 }
 
+/* Expand builtin macro compile_error!("error"), which forces a compile error
+   during the compile time. */
+AST::ASTFragment
+MacroBuiltin::compile_error (Location invoc_locus, AST::MacroInvocData &invoc)
+{
+  auto lit_expr
+    = parse_single_string_literal (invoc.get_delim_tok_tree (), invoc_locus);
+  if (lit_expr == nullptr)
+    return AST::ASTFragment::create_error ();
+
+  std::string error_string = lit_expr->as_string ();
+  rust_error_at (invoc_locus, "%s", error_string.c_str ());
+
+  return AST::ASTFragment::create_error ();
+}
+
 } // namespace Rust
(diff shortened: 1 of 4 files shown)


CohenArthur (approved):
This looks great, thank you so much!

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -236,4 +236,20 @@ MacroBuiltin::include_str (Location invoc_locus, AST::MacroInvocData &invoc)
...
+AST::ASTFragment
+MacroBuiltin::compile_error (Location invoc_locus, AST::MacroInvocData &invoc)
+{
+  auto lit_expr
+    = parse_single_string_literal (invoc.get_delim_tok_tree (), invoc_locus);
+  if (lit_expr == nullptr)
+    return AST::ASTFragment::create_error ();
+
+  std::string error_string = lit_expr->as_string ();
+  rust_error_at (invoc_locus, "%s", error_string.c_str ());
```
bjorn3, line 250: Which token does rustc get the error span from? Is it the `compile_error` identifier, the !, the brackets or the string argument?

liushuyu: > Which token does rustc get the error span from? Is it the `compile_error` identifier, the !, the brackets or the string argument?

From what I could understand, `rustc` outputs the following:

```
    Checking test-error v0.1.0 (/tmp/test-error)
error: Hello, world!
 --> src/main.rs:2:5
  |
2 |     compile_error!("Hello, world!");
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `test-error` due to previous error
```

So I guess it got the span from the whole macro invokation?

bjorn3: I think so too: https://github.com/rust-lang/rust/blob/45e2c2881d11324d610815bfff097e25c412199e/compiler/rustc_builtin_macros/src/compile_error.rs#L16

CohenArthur: Yes, from my understanding the given span is that of the invocation, with the token stream being the invocation token stream. We pass a reference to the entire `MacroInvocData` for simplicity. It contains a lot of useless (in this situation) info such as the delimiters used for the invocation. I think it's fine for now but we could be changing it in the future

PR #1082, opened 2022-04-05
Add base for privacy visitor

CohenArthur:
This PR is extremely early and implements some building blocks for privacy visitors. I'd like to get some feedback on the architecture and, if satisfactory, merge this first "visitor" which only takes care of visiting HIR struct definitions, to make reviewing easier. We could also merge it to a different branch for now, in order to not add an incomplete pass to the compiler.

Thanks!

### DIFF
--- a/gcc/rust/hir/tree/rust-hir.h
+++ b/gcc/rust/hir/tree/rust-hir.h
@@ -45,6 +45,41 @@ class HIRTypeVisitor;
 // forward decl for use in token tree method
 class Token;
 
+// Kind for downcasting various HIR nodes to other base classes when visiting
+// them
+enum BaseKind
+{
+  /* class ExternalItem */
+  EXTERNAL,
+  /* class TraitItem */
+  TRAIT_ITEM,
+  /* class VisItem */
+  VIS_ITEM,
+  /* class Item */
+  ITEM,
+  /* class ImplItem */
+  IMPL,
+  /* class Type */
+  TYPE,
+  /* class Stmt */
+  STMT,
+  /* class Expr */
+  EXPR,
+  /* class Pattern */
+  PATTERN,
+};
+
+class Node
+{
+public:
+  /**
+   * Get the kind of HIR node we are dealing with. This is useful for
+   * downcasting to more precise types when necessary, i.e going from an `Item*`
+   * to a `VisItem*`
+   */
+  virtual BaseKind get_hir_kind () = 0;
+};
+
 // A literal - value with a type. Used in LiteralExpr and LiteralPattern.
 struct Literal
 {
@@ -91,7 +126,7 @@ public:
 
 /* Base statement abstract class. Note that most "statements" are not allowed in
  * top-level module scope - only a subclass of statements called "items" are. */
-class Stmt
+class Stmt : public Node
 {
 public:
   // Unique pointer custom clone function
@@ -99,6 +134,8 @@ public:
   {
     return std::unique_ptr<Stmt> (clone_stmt_impl ());
   }
+
+  BaseKind get_hir_kind () override { return STMT; }
 
   virtual ~Stmt () {}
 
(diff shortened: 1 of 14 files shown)


CohenArthur:
@tschwinge @CastilloDel I think there might be an issue with the `Check for new warnings` part of the CI because this PR adds a loooooot of unused arguments warnings!

tschwinge:
> @tschwinge @CastilloDel I think there might be an issue with the `Check for new warnings` part of the CI because this PR adds a loooooot of unused arguments warnings!

I've just run this locally, and I'm not seeing any new warnings.

Then, I downloaded "old" and "new" (here) logs from the CI system, and compared these, and I'm also not seeing any new warnings.

So, which exactly are you seeing?

CohenArthur:
Oh, really sorry about that @tschwinge it seems to just be my linter going crazy. Sorry for bothering you!

CastilloDel:
Nice! I was also trying to check, but didn't found anything

tschwinge:
@CohenArthur, no worries -- could easily have been an actual issue.  Still happy to have a look re your _linter going crazy_, if you'd like.

gcc/rust/hir/tree/rust-hir.h
```diff
@@ -45,6 +45,41 @@ class HIRTypeVisitor;
 // forward decl for use in token tree method
 class Token;
 
+// Kind for downcasting various HIR nodes to other base classes when visiting
+// them
+enum BaseKind
```
philberty, line 50: i like this alot!

philberty: Usually we put this under the Node class but its up to your wether you want to do that or not it might be annoying

gcc/rust/privacy/rust-privacy-ctx.cc
```diff
@@ -0,0 +1,91 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#include "rust-privacy-ctx.h"
+#include "selftest.h"
+
+namespace Rust {
+namespace Privacy {
```
philberty, line 23: space :)

gcc/rust/privacy/rust-reachability.cc
```diff
@@ -0,0 +1,137 @@
...
+  current_level = struct_reach;
+
+  if (struct_reach != ReachLevel::Unreachable)
+    {
+      for (auto &field : struct_item.get_fields ())
+	if (field.get_visibility ().is_public ())
+	  ctx.update_reachability (field.get_mappings (), struct_reach);
+
+      // for (auto &generic : struct_item.get_generic_params ())
+      // FIXME: How do we visit these generics's predicates with the
```
philberty, line 84: Do you mean to visit the generic parameters for the structure? like struct Foo<T>(T)?

gcc/rust/privacy/rust-reachability.cc
```diff
@@ -0,0 +1,137 @@
...
+      // reachability visitor?
+
+      // FIXME: How do we get each generic's predicates from here?
+      // TypeContext?
+
+      // for (auto &field : struct_item.get_fields ())
+      // if (field.get_visibility ().is_public ())
+      // FIXME: How do we visit these fields with the reachability
+      // visitor?
```
philberty, line 93: HIR::StructField is just a struct here at this point so it has visibility directly. So you only need to visit if you need to visit the types they specify in which case we will need a new visitor for types which you could pass the privacy context down into that

gcc/rust/hir/tree/rust-hir.h
CohenArthur: Done in 1df917c0723

gcc/rust/privacy/rust-reachability.cc
CohenArthur: I believe this is to visit and see what trait bounds are being used. So for the following struct:
```rust
struct Foo<T: SomeTrait> {
    foo: T,
}
```
We want to mark `SomeTrait` as reachable

gcc/rust/privacy/rust-reachability.cc
CohenArthur: Yup, that's exactly it. Visit the type each field specify to mark it as reachable

Issue #789, opened 2021-11-02
Default Crate Name is wrong


philberty:
The default in rustc is the source filename without `.rs` or `rust_out` if the source is stdin.

_Originally posted by @bjorn3 in https://github.com/Rust-GCC/gccrs/pull/788#discussion_r740758707_

PR #1083, opened 2022-04-05
bugfix: fix several minor issues

liushuyu:
- Fixed `-frust-crate= option` got incorrectly overridden by a default value (`example`)
- Fix a minor typo in `gcc/rust/ast/rust-ast-full-test.cc`

### DIFF
--- a/gcc/rust/ast/rust-ast-full-test.cc
+++ b/gcc/rust/ast/rust-ast-full-test.cc
@@ -311,7 +311,7 @@ Visibility::as_string () const
     case NONE:
       return std::string ("pub");
     case CRATE:
-      return std::string ("ub(crate)");
+      return std::string ("pub(crate)");
     case SELF:
       return std::string ("pub(self)");
     case SUPER:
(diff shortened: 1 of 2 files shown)


CohenArthur (approved):
Thanks! I guess this ties in to [this issue](https://github.com/Rust-GCC/gccrs/issues/789) which is also related to the crate name. LGTM :)

liushuyu:
> Thanks! I guess this ties in to [this issue](https://github.com/Rust-GCC/gccrs/issues/789) which is also related to the crate name. LGTM :)

You are right! I guess the default crate name should be the file stem of the source file name.

philberty:
Great work @liushuyu i think we just need you to remove auto generated stuff in the PR comment. "Thank you for making Rust GCC better!" otherwise this will end up in the git history we just need the comment relating to these changes. Then we can merge this.

philberty (approved):

liushuyu:
> Great work @liushuyu i think we just need you to remove auto generated stuff in the PR comment. "Thank you for making Rust GCC better!" otherwise this will end up in the git history we just need the comment relating to these changes. Then we can merge this.

Oh, I did not know that. I have removed the extra contents now, and for the record, I will move the note I added to this comment.

---

### Considerations

- In `rustc`, when the crate name is not supplied, `rustc` will substitute the name `main` into the crate name when emitting symbols (observed behavior, I have yet to find where this logic is in `rustc`).

Issue #914, opened 2022-02-09
Dynamic objects support inheritance 


philberty:
I tried this code:

```rust
extern "C" {
    fn printf(s: *const i8, ...);
}

struct Foo(i32);
trait Bar {
    fn baz(&self);
}

trait Baz : Bar {
    fn qux(&self);
}

impl Bar for Foo {
    fn baz(&self) {
        unsafe {
            let a = "baz %i\n\0";
            let b = a as *const str;
            let c = b as *const i8;

            printf(c, self.0);
        }
    }
}

impl Baz for Foo {
    fn qux(&self) {
        unsafe {
            let a = "baz %i\n\0";
            let b = a as *const str;
            let c = b as *const i8;

            printf(c, self.0);
        }
    }
}


fn dynamic_dispatch(t: &dyn Baz) {
    t.baz();
    t.qux();
}

pub fn main() {
    let a;
    a = &Foo(123);
    dynamic_dispatch(a);
}
```

I expected to see this happen: compile without error

Instead, this happened: 

```
<source>:40:7: error: failed to resolve method for 'baz'
   40 |     t.baz();
      |       ^
<source>:40:5: error: failed to type resolve expression
   40 |     t.baz();
      |     ^
rust1: error: bounds not satisfied for dyn [example::Baz] 'example::Bar' is not satisfied
```

### Meta

- What version of Rust GCC were you using, git sha if possible. 83bfbf0746c87b641754697a3c8e9f7a7cb08aa9

bjorn3:
You can't get from `&dyn Baz` to `&dyn Bar` on stable rustc. What happens here is that the vtable contains all methods of all super traits. On nightly rustc it also contains pointers to vtables of super traits as necessary to allow upcasting trait objects to super traits.

liamnaddell:
New behavior: ICE!

```
crab1: internal compiler error: in compute_address_for_trait_item, at rust/backend/rust-compile.cc:362
0x10afce3 Rust::Compile::HIRCompileBase::compute_address_for_trait_item(Rust::Resolver::TraitItemReference const*, Rust::TyTy::TypeBoundPredicate const*, std::vector<std::pair<Rust::Resolver::TraitReference*, Rust::HIR::ImplBlock*>, std::allocator<std::pair<Rust::Resolver::TraitReference*, Rust::HIR::ImplBlock*> > >&, Rust::TyTy::BaseType const*, Rust::TyTy::BaseType const*, unsigned int)
[...]

PR #1086, opened 2022-04-06
Slice support

### LINKED ISSUE #849: Add support for Slices

philberty:
Please see the commit a8de089969cb45199008027cd8d1b80dff25746f for
a long explanation of what's going on in the patch. Unfortunately, I have not been
able to split this patch up anymore since supporting slices exposed many bugs
in the implementation of generics in general never main the missing support for
generic associated types.

Fixes #849

### DIFF
--- a/gcc/rust/typecheck/rust-hir-trait-ref.h
+++ b/gcc/rust/typecheck/rust-hir-trait-ref.h
@@ -262,6 +262,8 @@ public:
     return hir_trait_ref->get_mappings ();
   }
 
+  DefId get_defid () const { return get_mappings ().get_defid (); }
+
   bool lookup_hir_trait_item (const HIR::TraitItem &item,
 			      TraitItemReference **ref)
   {
@@ -435,6 +437,9 @@ public:
   TyTy::BaseType *get_self () { return self; }
 
   void setup_associated_types ();
+
+  void setup_associated_types2 (const TyTy::BaseType *self,
+				const TyTy::TypeBoundPredicate &bound);
 
   void reset_associated_types ();
 
--- a/gcc/rust/typecheck/rust-hir-trait-resolve.cc
+++ b/gcc/rust/typecheck/rust-hir-trait-resolve.cc
@@ -137,7 +137,7 @@ TraitItemReference::associated_type_set (TyTy::BaseType *ty)
   TyTy::PlaceholderType *placeholder
     = static_cast<TyTy::PlaceholderType *> (item_ty);
 
-  placeholder->set_associated_type (ty->get_ref ());
+  placeholder->set_associated_type (ty->get_ty_ref ());
 }
 
 void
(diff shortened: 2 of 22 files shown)


gcc/testsuite/rust/execute/torture/slice-magic.rs
```diff
@@ -0,0 +1,106 @@
...
+}
+
+#[lang = "Range"]
+pub struct Range<Idx> {
+    pub start: Idx,
+    pub end: Idx,
+}
+
+#[lang = "const_slice_ptr"]
+impl<A> *const [A] {
```
bjorn3, line 24: Just as a heads up https://github.com/rust-lang/rust/pull/94963 changed the way this kind of impls on builtin and foreign types are handled. Libcore now uses `#![rustc_coherence_is_core]` to allow impls on any builtin type without `#[lang = "..."]` and other sysroot crates use `#[rustc_allow_incoherent_impl]` on all impl items for non-local impls.

gcc/rust/typecheck/rust-hir-trait-ref.h
```diff
@@ -436,6 +438,9 @@ class AssociatedImplTrait
 
   void setup_associated_types ();
 
+  void setup_associated_types2 (const TyTy::BaseType *self,
```
CohenArthur, line 441: I think the name might be a little unclear. Can we overload the previous `setup_associated_types` or maybe give it a name reflecting the fact that it performs a different operation?

gcc/rust/typecheck/rust-hir-trait-resolve.cc
```diff
@@ -175,6 +175,172 @@ AssociatedImplTrait::setup_associated_types ()
...
+      hrtb_bound_arguments.push_back (r);
+    }
+
+  // FIXME I wish we had a zip function
+  rust_assert (impl_trait_predicate_args.size ()
+	       == hrtb_bound_arguments.size ());
+  for (size_t i = 0; i < impl_trait_predicate_args.size (); i++)
+    {
+      TyTy::BaseType *a = impl_trait_predicate_args.at (0);
+      TyTy::BaseType *b = hrtb_bound_arguments.at (0);
```
CohenArthur, line 296:
```suggestion
      TyTy::BaseType *a = impl_trait_predicate_args.at (i);
      TyTy::BaseType *b = hrtb_bound_arguments.at (i);
```
Typo? Or is this the expected behavior?

gcc/rust/typecheck/rust-tyty-bounds.cc
```diff
@@ -103,46 +106,76 @@ TypeBoundPredicate::TypeBoundPredicate (
...
+    substitutions.push_back (p.clone ());
+
+  std::vector<SubstitutionArg> mappings;
+  for (size_t i = 0; i < other.used_arguments.get_mappings ().size (); i++)
     {
-      for (const auto &p : other.get_substs ())
-	substitutions.push_back (p.clone ());
+      const SubstitutionArg &oa = other.used_arguments.get_mappings ().at (i);
+      TyTy::BaseType *argument
+	= oa.get_tyty () == nullptr ? nullptr : oa.get_tyty ()->clone ();
```
CohenArthur, line 170: This is the second time this ternary is used, I think it might be worth it to refactor it in a function or method so that it's clearer to use and easier to understand?

CohenArthur (commented):
Great progress :DD Excited to see this land!

gcc/rust/typecheck/rust-hir-trait-resolve.cc
philberty: this is a typo i thought i fixed this but I must have missed it, good find.

gcc/rust/typecheck/rust-hir-trait-ref.h
philberty: This function is designed to overtake the other setup_associated_types the other function is still used in one path for dyn traits which I need to refactor and update for #914 so I would rather fix this then since this PR is already so big.

gcc/rust/typecheck/rust-tyty-bounds.cc
philberty: yeah, I might be able to update the copy constructor on the SubstitutionArg to make this safer. There is an awkward case as Traits have an implicit Self type-parameter in order to do proper compatibility checking when a trait is implemented.

philberty: fixed

gcc/rust/typecheck/rust-hir-trait-resolve.cc
philberty: fixed

gcc/rust/typecheck/rust-hir-trait-ref.h
philberty: I've raised https://github.com/Rust-GCC/gccrs/issues/1105 to track this

Issue #1068, opened 2022-03-28
gccrs crashes when trying to allocate extremely large array


CohenArthur:
Two test cases expose the issue:

- https://github.com/rust-lang/rust/blob/0e4524e5b4a9c5656fef69f532f96eb9959a1803/src/test/ui/consts/issue-66342.rs
- https://github.com/rust-lang/rust/blob/0e4524e5b4a9c5656fef69f532f96eb9959a1803/src/test/ui/consts/issue-66397.rs

The tests make sure that rustc does not actually try to allocate 4 terabytes of memory during compilation and crashes

philberty:
Nice find @dafaust that looks like the correct solution for correctness. There is one caveat but maybe GCC will handle it for us.

```rust
const A:[u8; 4 * 1024 * 1024 * 1024 * 1024] = [0; 4 * 1024 * 1024 * 1024 * 1024];
```

Rustc also fails to handle this case https://godbolt.org/z/TWznxnfon

I think in a const and possibly even in the static context, a memset may not be possible. Though we could probably figure something out but if rustc fails on a const context then we can fail too lol.

bjorn3:
`std::ptr::write_bytes` which lowers to memset for u8/i8 is const unstable.

PR #1087, opened 2022-04-06
Use loop to initialize repeat arrays

dafaust:
This PR changes how we compile initializers for arrays of repeating elements. I use the same approach outlined in the comments of the linked issue, with some tweaks. It is very similar to how the D language front-end compiles, the new function `Gcc_backend::array_initializer` is heavily inspired by the D front-end's `build_array_set`

This fixes the issue where the compiler tries to allocate a vec containing all elements of the array to be constructed, and therefore explodes on huge constructions (e.g. `let x = [0u8; 4 * 1024 * 1024 * 1024 * 1024]`)

However, we can only initialize non-const arrays in this way. For arrays in const contexts we must initialize them at compile time, and therefore continue using the old method.

Fixes: #1068

### DIFF
--- a/gcc/rust/backend/rust-compile-context.h
+++ b/gcc/rust/backend/rust-compile-context.h
@@ -250,6 +250,10 @@ public:
     fn_stack.push_back (fncontext{fn, ret_addr});
   }
   void pop_fn () { fn_stack.pop_back (); }
+
+  bool in_fn () { return fn_stack.size () != 0; }
+
+  // Note: it is undefined behavior to call peek_fn () if fn_stack is empty.
   fncontext peek_fn () { return fn_stack.back (); }
 
   void push_type (tree t) { type_decls.push_back (t); }
@@ -301,6 +305,14 @@ public:
     return pop;
   }
 
+  void push_const_context (void) { const_context++; }
+  void pop_const_context (void)
+  {
+    if (const_context > 0)
+      const_context--;
+  }
+  bool const_context_p (void) { return (const_context > 0); }
+
   std::string mangle_item (const TyTy::BaseType *ty,
 			   const Resolver::CanonicalPath &path) const
   {
@@ -341,6 +353,9 @@ private:
   std::vector<::Bvariable *> var_decls;
   std::vector<tree> const_decls;
   std::vector<tree> func_decls;
+
+  // Nonzero iff we are currently compiling something inside a constant context.
+  unsigned int const_context = 0;
 };
 
 } // namespace Compile
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -1163,19 +1163,54 @@ CompileExpr::array_copied_expr (Location expr_locus,
   unsigned HOST_WIDE_INT len
     = wi::ext (max - min + 1, precision, sign).to_uhwi ();
 
-  // create the constructor
-  size_t idx = 0;
-  std::vector<unsigned long> indexes;
-  std::vector<tree> constructor;
-  for (unsigned HOST_WIDE_INT i = 0; i < len; i++)
+  // In a const context we must initialize the entire array, which entails
+  // allocating for each element. If the user wants a huge array, we will OOM
+  // and die horribly.
+  if (ctx->const_context_p ())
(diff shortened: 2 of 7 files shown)


gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -1163,6 +1163,36 @@ CompileExpr::array_copied_expr (Location expr_locus,
   unsigned HOST_WIDE_INT len
     = wi::ext (max - min + 1, precision, sign).to_uhwi ();
 
+#if 1
+  // TODO: Somehow check whether we are compiling in a const context, since we
+  //       obviously can't make a loop to init in such cases, and will need to
+  //       allocate/init the array (however large) at compile time...
```
dafaust, line 1169: What would be the best way to check whether we are compiling an array in a `const` item?

CohenArthur: I assume it would probably be by using some form of context information: If we are compiling a constant assignment, `push_context(CONST)`, visit the initializer and then `pop_context()`. I've never touched this part of the codebase so unsure on how all of that maps! This is what we had to do for macro expansion, where you do want specific behavior based on the "parent" of the thing you're currently handling. In that case, you want information about the "parent" which will receive/contain the array you're compiling, in a way

dafaust:
Saving the draft just in case because idk how github works

> This draft PR changes how we compile initializers for arrays of repeating elements. I use the same approach outlined in the comments of the linked issue, with some tweaks. It is very similar to how the D language front-end compiles, the new function `Gcc_backend::array_initializer` is heavily inspired by the D front-end's `build_array_set`
> 
> This fixes the issue where the compiler tries to allocate a vec containing all elements of the array to be constructed, and therefore explodes on huge constructions (e.g. `let x = [0u8; 4 * 1024 * 1024 * 1024 * 1024]`)
> 
> Draft for feedback and because there is one issue yet to resolve - this breaks const eval arrays. In particular, this patch will cause a SEGV when compiling `testsuite/rust/compile/torture/array_const_fold_2.rs`, this can easily be made to not SEGV but the behavior would not be valid for a const expression - we'd emit the same loop initializer to be evaluated at runtime. For reference, the test case in question:
> 
> ```
> const SIZE: usize = 14 + 2;
> const TEST: [i32; SIZE] = [2; SIZE];
> // { dg-warning "unused name" "" { target *-*-* } .-1 }
> ```
> 
> I left the old initializer code in `rust-compile-expr.cc` `#if 0`'d out, I think we will need to use it for const expressions (since it does work for those cases), switching somehow on the context in which we are constructing the array, and bite the bullet that if the user wants a 4TiB array initialized at compile time we will fall over.
> 
> Thoughts/comments/suggestions welcome as always smiley
> 
> Addresses: #1068

gcc/rust/rust-gcc.cc
```diff
@@ -1694,6 +1696,77 @@ Gcc_backend::array_constructor_expression (
   return ret;
 }
 
+// Build insns to create an array, initialize all elements of the array to
+// value, and return it
+tree
+Gcc_backend::array_initializer (tree fndecl, tree block, tree array_type,
```
philberty, line 1702: Can we make an issue to pull this out into rust-tree.h at some point it would be nice to eventually get rid of this abstraction since you have done most of the hard work for it :D

gcc/rust/backend/rust-compile-context.h
```diff
@@ -341,6 +353,9 @@ class Context
   std::vector<::Bvariable *> var_decls;
   std::vector<tree> const_decls;
   std::vector<tree> func_decls;
+
+  // Nonzero iff we are currently compiling something inside a constant context.
+  unsigned int const_context = 0;
```
philberty, line 358: This is great we need to port over from CPP DECL_DECLARED_CONSTEXPR_P which sets a lang_flag bit on all relevant tree's. We might call it the same name but when we are compiling any constants we could set this as well. thanks for doing it

philberty (approved):
LGTM thanks for doing this one. You have also setup the piece we need for DECL_DECLARED_CONSTEXPR_P. We might have to change it to be a stack tracking the current context at some point but that needs some thought.

PR #1088, opened 2022-04-07
rust-session-manager: better crate name handling logic

liushuyu:
- rust-session-manager: set and validate crate name properly
- testsuite/rust: fix the testcases and add more testcases for testing crate name handling


gcc/rust/rust-session-manager.cc
```diff
@@ -474,6 +498,26 @@ Session::enable_dump (std::string arg)
 void
 Session::parse_files (int num_files, const char **files)
 {
+  rust_assert (num_files > 0);
+
+  if (options.crate_name.empty ())
+    {
+      /* HACK: We use the first file to infer the crate name, which might be
+       * incorrect: since rustc only allows one file to be supplied in the
+       * command-line */
```
liushuyu, line 507: I actually would like to add the following changes so that this assumption will always be true:

```diff
diff --git a/gcc/rust/rust-lang.cc b/gcc/rust/rust-lang.cc
index 5ecd79b15..5f06862eb 100644
--- a/gcc/rust/rust-lang.cc
+++ b/gcc/rust/rust-lang.cc
@@ -168,6 +168,12 @@ grs_langhook_parse_file (void)
 {
   rust_debug ("Preparing to parse files. ");
 
+  /* Rust only allows the "root file" to be supplied */
+  if (num_in_fnames > 1) {
+    rust_error_at(Location(), "multiple input filenames provided");
+    return;
+  }
+
   Rust::Session::get_instance ().parse_files (num_in_fnames, in_fnames);
 }
 
```

gcc/testsuite/rust/compile/bad-crate-name.rs
```diff
@@ -0,0 +1,4 @@
+// { dg-additional-options "-frust-crate=bad+name" }
+// { dg-excess-errors "invalid crate name: ...." }
+// { dg-excess-errors "unrecognized command-line option ...." }
```
liushuyu, line 3: Is there a better way to write those directives? It looks like error messages w/o location information can't be matched by `dg-error`.

gcc/rust/rust-session-manager.cc
```diff
@@ -474,6 +498,26 @@ Session::enable_dump (std::string arg)
 void
 Session::parse_files (int num_files, const char **files)
 {
+  rust_assert (num_files > 0);
```
liushuyu, line 501: Is this `assert` necessary? I am not quite sure.

gcc/rust/rust-session-manager.cc
```diff
@@ -474,6 +498,26 @@ Session::enable_dump (std::string arg)
...
+      /* HACK: We use the first file to infer the crate name, which might be
+       * incorrect: since rustc only allows one file to be supplied in the
+       * command-line */
+      auto crate_name = infer_crate_name (files[0]);
+      rust_debug_loc (Location (), "inferred crate name: %s",
+		      crate_name.c_str ());
+      if (!options.set_crate_name (crate_name))
+	{
+	  rust_error_at (
+	    Location (),
```
liushuyu, line 514: I actually want to fake location information here so that it points to the `<input-file>:1` or something like that.

So that `(inferred from the input file %<%s%>)` could be omitted.

gcc/rust/rust-session-manager.h
```diff
@@ -205,9 +208,31 @@ struct CompileOptions
     enable_dump_option (DumpOption::TYPE_RESOLUTION_DUMP);
   }
 
+  /* Validate the crate name using the ASCII rules
+   TODO: Support Unicode version of the rules */
+  bool validate_crate_name (const std::string crate_name)
```
CohenArthur, line 213:
```suggestion
  bool validate_crate_name (const std::string &crate_name)
```

gcc/rust/rust-session-manager.h
```diff
@@ -205,9 +208,31 @@ struct CompileOptions
...
+  {
+    if (crate_name.length () > kMaxNameLength || crate_name.empty ())
+      return false;
+    for (auto &c : crate_name)
+      {
+	if (!(ISALNUM (c) || c == '_' || c == '-'))
+	  return false;
+      }
+    return true;
+  }
```
CohenArthur, line 223: This is something we can easily add unit tests for! If you think it would be useful, it would be great to have a base in place for when we implement UTF-8 crate names as you mention in the TODO above

gcc/rust/rust-session-manager.h
```diff
@@ -205,9 +208,31 @@ struct CompileOptions
...
-    // TODO: validate the crate name?
+    if (!validate_crate_name (name))
+      return false;
+
+    /* Replace all the '-' symbols with '_' per Rust rules */
+    for (size_t i = 0; i < name.size (); i++)
+      {
+	if (name[i] == '-')
+	  name[i] = '_';
+      }
```
CohenArthur, line 235:
```suggestion
    for (auto &c: name)
	if (c == '-')
	  c = '_';
```
Any reason why you use the range based for loop above and not here?

gcc/rust/rust-session-manager.h
```diff
@@ -205,9 +208,31 @@ struct CompileOptions
...
+  /* Validate the crate name using the ASCII rules
+   TODO: Support Unicode version of the rules */
+  bool validate_crate_name (const std::string crate_name)
+  {
+    if (crate_name.length () > kMaxNameLength || crate_name.empty ())
+      return false;
+    for (auto &c : crate_name)
+      {
+	if (!(ISALNUM (c) || c == '_' || c == '-'))
+	  return false;
```
CohenArthur, line 220: I think it would be good to emit the error here rather than when checking for the return value of `set_crate_name`? This way, we could have errors saying ("crate name cannot be empty", "crate name cannot be longer than 64 characters", "invalid character in crate name: @") and so on

CohenArthur (changes requested):

gcc/rust/rust-session-manager.h
liushuyu: You are right! I will change the implementation to emit an error here.

gcc/rust/rust-session-manager.cc
```diff
@@ -53,7 +53,32 @@ const char *kHIRDumpFile = "gccrs.hir.dump";
...
+{
+  if (filename == "-")
+    return kDefaultCrateName;
+
+  std::string crate = std::string (filename);
+#ifdef __WIN32
+  size_t path_sep = crate.find_last_of ('\\');
+#else
+  size_t path_sep = crate.find_last_of ('/');
+#endif
```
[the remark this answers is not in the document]
liushuyu: > There's a `file_separator` constant somewhere in the parser :D I think it would help here. It's located [here in rust-system.h](https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/rust-system.h#L64-L70)

Ah, thank you for the pointer! I will use that instead.

CohenArthur:
Sorry about all the nitpicking @liushuyu :D This looks great! Thank you for working on it!

gcc/rust/rust-session-manager.h
liushuyu: > This is something we can easily add unit tests for! If you think it would be useful, it would be great to have a base in place for when we implement UTF-8 crate names as you mention in the TODO above

I will add some unit-tests for this.

liushuyu:
> Sorry about all the nitpicking @liushuyu :D This looks great! Thank you for working on it!

Don't worry about the "nitpicking", I can't learn anything from contributing to this project if nobody points out the issues, isn't it 😄

gcc/rust/rust-session-manager.h
liushuyu: > Any reason why you use the range based for loop above and not here?

I had an incorrect intuition that C++ for-range loops can't mutate the elements, so I used the C-style for loop here.

I will switch to the for-range loop here

CohenArthur: I'm pretty sure the rules are as follow (I'm not well-versed in C++ by any means):

1. `auto &elt` is just a regular reference, so it is mutable by default
2. `const auto &elt` is a const reference, immutable
3. `auto elt` creates a copy of the element, so the copy is mutable but will not modify the underlying iterator
4. `const auto elt` also create a copy, but an immutable one obviously

liushuyu:
I have changed the display style of the error when inferred crate name is invalid to this:

```
./gcc/rust1 -flto -frust-dump-parse -Warray-bounds -fdump-tree-gimple ../gcc/testsuite/rust/compile/bad_file_name.txt.rs -o t.s              04:43:42
rust1: error: invalid character ‘.’ in crate name: ‘bad_file_name.txt’
../gcc/testsuite/rust/compile/bad_file_name.txt.rs: note: crate name inferred from this file
```

gcc/rust/rust-session-manager.h
liushuyu: Unit-tests added. But I am not sure if I am doing this right, since the only unit test framework I could find was the `selftest` module.

CohenArthur: selftest is the correct framework to use :) Good job

gcc/rust/rust-session-manager.cc
```diff
@@ -1113,3 +1186,32 @@ TargetOptions::enable_implicit_feature_reqs (std::string feature)
...
+
+#if CHECKING_P
+namespace selftest {
+void
+rust_crate_name_validation_test (void)
+{
+  ASSERT_TRUE (Rust::CompileOptions::validate_crate_name ("example"));
+  ASSERT_TRUE (Rust::CompileOptions::validate_crate_name ("abcdefg_1234"));
+  ASSERT_TRUE (Rust::CompileOptions::validate_crate_name ("1"));
+  // The next test does not pass as of current implementation
```
CohenArthur, line 1198: We should add a `TODO` or a `FIXME` here

gcc/rust/rust-session-manager.cc
```diff
@@ -1113,3 +1186,32 @@ TargetOptions::enable_implicit_feature_reqs (std::string feature)
  *  - code generation
  *  - link */
 } // namespace Rust
+
+#if CHECKING_P
+namespace selftest {
+void
+rust_crate_name_validation_test (void)
```
CohenArthur, line 1193: You need to call this function from the `rust-lang.cc` file. It contains the hook and entrypoint of the unit tests for our frontend. Simply add a call to `rust_crate_name_validation_test ()` in the `run_rust_tests()` function

gcc/rust/rust-session-manager.cc
CohenArthur: I think this should probably be done in a separate PR to help discussion and avoid cluttering. I personally think it would be a good idea

gcc/rust/rust-session-manager.cc
liushuyu: Function call added.

gcc/rust/rust-session-manager.cc
```diff
@@ -1113,3 +1187,33 @@ TargetOptions::enable_implicit_feature_reqs (std::string feature)
...
+  ASSERT_TRUE (Rust::CompileOptions::validate_crate_name ("example"));
+  ASSERT_TRUE (Rust::CompileOptions::validate_crate_name ("abcdefg_1234"));
+  ASSERT_TRUE (Rust::CompileOptions::validate_crate_name ("1"));
+  // FIXME: The next test does not pass as of current implementation
+  // ASSERT_TRUE (Rust::CompileOptions::validate_crate_name ("惊吓"));
+  // NOTE: - is not allowed in the crate name ...
+  /*
+  ASSERT_FALSE (Rust::CompileOptions::validate_crate_name ("abcdefg-1234"));
+  ASSERT_FALSE (Rust::CompileOptions::validate_crate_name ("a+b"));
+  ASSERT_FALSE (Rust::CompileOptions::validate_crate_name ("/a+b/")); */
```
liushuyu, line 1205: I am not sure how to correctly handle the negative tests, since `error_at` itself will just fail the test.

gcc/testsuite/rust/compile/bad-crate-name.rs
philberty: @dkm ^^ is this the best way? I think it is i cant think of how we handle error messages without location info.

gcc/rust/rust-session-manager.cc
philberty: yeah this will cause an assertion if someone invokes the compiler directly without any inputs which is valid to produce an empty compilation unit. So we need to remove that

gcc/rust/rust-session-manager.cc
```diff
@@ -474,6 +526,28 @@ Session::enable_dump (std::string arg)
...
 {
+  rust_assert (num_files > 0);
+
+  if (options.crate_name.empty ())
+    {
+      /* HACK: We use the first file to infer the crate name, which might be
+       * incorrect: since rustc only allows one file to be supplied in the
+       * command-line */
+      auto crate_name = infer_crate_name (files[0]);
+      rust_debug_loc (Location (), "inferred crate name: %s",
```
philberty, line 537: you can call rust_debug instead of rust_debug_loc when you dont have a location

liushuyu: > you can call rust_debug instead of rust_debug_loc when you dont have a location

I see. I will fix this.

gcc/rust/rust-session-manager.cc
```diff
@@ -474,6 +526,28 @@ Session::enable_dump (std::string arg)
...
+		      crate_name.c_str ());
+      if (!options.set_crate_name (crate_name))
+	{
+	  // fake a linemapping so that we can show the filename
+	  linemap->start_file (files[0], 0);
+	  linemap->start_line (0, 1);
+	  rust_inform (linemap->get_location (0),
+		       "crate name inferred from this file");
+	  linemap->stop ();
```
philberty, line 547: we don't need this linemap, since the options.set_crate_name() calls the validate function errors are already emited.

gcc/rust/rust-session-manager.cc
liushuyu: > yeah this will cause an assertion if someone invokes the compiler directly without any inputs which is valid to produce an empty compilation unit. So we need to remove that

Hmm. I will provide a fallback path for this case.

philberty (changes requested):
This is not your fault, but the bool set_crate_name I don't think I like.

It would be cleaner if we turned this into two functions. let's keep your

```bool validate_crate_name (const std::string &crate_name)```

And lets change the

```bool set_crate_name (std::string name)```

into

```void set_crate_name (std::string name)``` and just add an assertion of:

```
rust_assert(!name.empty());
```
This means we can change the set_crate_name functions to be a bit more functional so we call your validate_crate_name and then depending on the outcome of that we then set the crate name.

If your unit-tests are failing because of the error messages thats annoying because we want this to work as a stop gap measure what we could do is add a flag 

```bool validate_crate_name (const std::string &crate_name, bool emit_errors=true)```

But I don't like that, what we could do is investigate why emitting error messages are causing a failure, it would be very annoying to have to pass error messages back up to the caller.

File: gcc/rust/rust-session-manager.cc (hunk 14 above)
Comment (liushuyu) [reply gcc/rust/rust-session-manager.cc:547]: > we don't need this linemap, since the options.set_crate_name() calls the validate function errors are already emited.

I added this due to UX consideration, so the user will know which file caused this issue (the crate name is derived from the filename after all).

File: gcc/rust/rust-session-manager.cc (hunk 12 above)
Comment (CohenArthur) [reply gcc/rust/rust-session-manager.cc:1206]: I'd say a way to do this is to have the `validate_crate_name` function signature look like this:
```c++
bool validate_crate_name(const std::string &name, Error &error);
```
If the input is invalid, return false and store the error inside the given reference. Then, the caller emits it
```c++
Error error;
if (!validate_crate_name(name, error)) {
    error.emit_error();
    return false;
}
```

liushuyu: > I'd say a way to do this is to have the `validate_crate_name` function signature look like this:
> 
> ```c++
> bool validate_crate_name(const std::string &name, Error &error);
> ```
> 
> If the input is invalid, return false and store the error inside the given reference. Then, the caller emits it
> 
> ```c++
> Error error;
> if (!validate_crate_name(name, error)) {
>     error.emit_error();
>     return false;
> }
> ```

I think this is also valid. I am not sure which one is better? (1) clear the error flag somewhere in GCC; (2) pass the `Error` back to the caller.

CohenArthur:
> But I don't like that, what we could do is investigate why emitting error messages are causing a failure, it would be very annoying to have to pass error messages back up to the caller.

Emitting error does not cause a selftest failure, but it causes a general failure of the `make` process since it sets the global exit code to 1.

liushuyu:
I took @CohenArthur 's suggestion to fix `selftest` issues. However, it appears that the code is not so pretty since the `Rust::Error` type does not have a constructor that accepts zero arguments.

Since we are limited to C++11, we can't use fancy stuff like `std::optional`. I guess another option is to make `validate_crate_name` to return `std::unique_ptr<Error>`? It sounds stupid, but probably a bit cleaner in code?

CohenArthur:
> I took @CohenArthur 's suggestion to fix `selftest` issues. However, it appears that the code is not so pretty since the `Rust::Error` type does not have a constructor that accepts zero arguments.
> 
> Since we are limited to C++11, we can't use fancy stuff like `std::optional`. I guess another option is to make `validate_crate_name` to return `std::unique_ptr<Error>`? It sounds stupid, but probably a bit cleaner in code?

Returning a unique_ptr has more ergonomic hit imo than filling up a reference + it's an extra allocation (granted we're not a in codepath that's very popular, but still).

We can add an empty constructor to `Error` if @philberty thinks it's a good idea. Something that would just create en error with an empty location and empty error message

gcc/rust/rust-session-manager.cc
```diff
@@ -346,7 +401,13 @@ Session::handle_option (
     case OPT_frust_crate_:
       // set the crate name
       if (arg != nullptr)
-	ret = options.set_crate_name (arg);
+	{
+	  Error error ((Location ()), std::string ());
```
CohenArthur, line 405:
```suggestion
	  Error error (Location (), "");
```

gcc/rust/rust-session-manager.cc
```diff
@@ -346,7 +401,13 @@ Session::handle_option (
...
+	  if ((ret = validate_crate_name (arg, error)))
+	    options.set_crate_name (arg);
+	  else
+	    error.emit_error ();
```
CohenArthur, line 409:
```suggestion
	  else {
	    rust_assert (!error.message.empty ());
	    error.emit_error ();
	  }
```
To make sure we filled the reference properly

CohenArthur (commented):
Great! I think this looks good and we're getting there. Good job :)

gcc/rust/rust-session-manager.cc
liushuyu: I tried this, and CI/bors did not like that. See https://github.com/Rust-GCC/gccrs/runs/5883397729?check_suite_focus=true#step:6:17.

CohenArthur: Oh okay, fair! I also suggested removing the extra parentheses around the `Location ()` constructor. By the way, if you're having trouble building the project or run the test suite on your machine let us know on Zulip :) We'll be happy to help

liushuyu: > Oh okay, fair! I also suggested removing the extra parentheses around the `Location ()` constructor. By the way, if you're having trouble building the project or run the test suite on your machine let us know on Zulip :) We'll be happy to help

I can run the test suites on my local machine. However, I discovered that sometimes CI is pickier about this, probably due to a newer GCC version.

gcc/rust/rust-session-manager.cc
```diff
@@ -53,7 +54,65 @@ const char *kHIRDumpFile = "gccrs.hir.dump";
 const char *kHIRTypeResolutionDumpFile = "gccrs.type-resolution.dump";
 const char *kTargetOptionsDumpFile = "gccrs.target-options.dump";
 
-const std::string kDefaultCrateName = "example";
+const std::string kDefaultCrateName = "rust_out";
```
philberty, line 57: is this the default crate name in Rustc? @flip1995 ?

flip1995: [Yes](https://github.com/rust-lang/rust/blob/2a83fbc42a9bb6bfdb8d3fb4ecce83fb410d7642/compiler/rustc_session/src/output.rs#L91) if no `--crate-name` was given AND no `#![crate_name]` attr was found AND there is no filename.

gcc/rust/rust-session-manager.cc
liushuyu: Hmm, my GCC 11.1 on my system does not like removing the parenthesis as well:

```
../../gcc/rust/rust-session-manager.cc:406:23: warning: parentheses were disambiguated as a function declaration [-Wvexing-parse]
  406 |           Error error (Location (), std::string ());
      |                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

CohenArthur: Oh that's interesting! First time I see the [most vexing parse](https://en.wikipedia.org/wiki/Most_vexing_parse) in the wild :D You can either keep the parentheses or turn that line into
```suggestion
	  auto error = Error (Location (), std::string ());
```
Whichever you prefer. I like the `auto` keyword personally :D

gcc/rust/rust-session-manager.cc
```diff
@@ -479,6 +543,32 @@ Session::enable_dump (std::string arg)
...
+    {
+      /* HACK: We use the first file to infer the crate name, which might be
+       * incorrect: since rustc only allows one file to be supplied in the
+       * command-line */
+      auto filename = "-";
+      if (num_files > 0)
+	filename = files[0];
+
+      auto crate_name = infer_crate_name (filename);
+      Error error ((Location ()), std::string ());
```
CohenArthur, line 556: Same nitpick here, feel free to ignore
```suggestion
      auto error = Error (Location (), std::string ());
```

CohenArthur (approved):
Thank you for all your work, I think this is in a really good state now. We can create issues out of a few comments which I'll do later. Great job!

gcc/rust/rust-session-manager.cc
philberty: Ah ok let's raise an issue for this then, I used this to default to example so that on compiler explorer we can compare the outputs and assumed the default crate name was an example for rustc.

philberty (approved):
LGTM, lets figure out a way to add a default argument to gccrs to update compiler explorer for our default crate name to match rustc @dkm might be able to help us here.

Issue #1085, opened 2022-04-06
macros: Built-in macros do not get expanded recursively


liushuyu:
I tried this code:

```rust
macro_rules! include_str {
    () => {{}};
}

fn main() -> i32 {
    let _ = include_str!(file!());

    0
}
```

I expected to see this happen: The program compiles successfully.
Instead, this happened: Rust GCC reported:

```
rust1: note: SUCCESSFULLY FINISHED INJECTION 
rust1: note: started expansion
macro11.rs:6:13: error: argument must be a string literal
    6 |     let _ = include_str!(file!());
      |             ^
rust1: note: finished expansion
rust1: note: SUCCESSFULLY FINISHED EXPANSION
```

### Meta

- Rust GCC version: 9011184f38a04f81ba3194b826bec3f30a11c07b

CohenArthur:
Thanks :D I think this is basically the same issue as #1084. Having macros as return values is not an issue, but having macro invocations as macro arguments does not work.

liushuyu:
> Thanks :D I think this is basically the same issue as #1084. Having macros as return values is not an issue, but having macro invocations as macro arguments does not work.

Oh, I see. I thought this was due to the way `MacroInvocData` was passed to the built-in macros. Because I tried expanding the following code and it did work:

```rust
macro_rules! add {
    ($e:expr, $($es:expr),*) => {
        $e + add!($($es),*)
    };
    ($e:expr) => {
        $e
    };
}

fn main() -> i32 {
    let b = add!(add!(15, 20), add!(3,4));

    0
}
```

So I guess two different parsing logic were used in these two cases? The `MacroExpander::expand_invoc_semi` and `MacroExpander::expand_invoc` are a bit confusing when I was trying to understand what the actual issue was.

CohenArthur:
Ah, that's really interesting. I'm investigating right now, thank you for the precisions :)

liamnaddell:
This example code doesn't seem to compile with rustc: https://godbolt.org/z/ehvh69WrG

liamnaddell:
```
liam@gentoo ~/expand $ rustc main.rs
error: no rules expected the token `file`
 --> main.rs:6:26
  |
1 | macro_rules! include_str {
  | ------------------------ when calling this macro
...
6 |     let _ = include_str!(file!());
  |                          ^^^^ no rules expected this token in macro call
  |
  = note: while trying to match end of macro

error: aborting due to 1 previous error

[...]

PR #1090, opened 2022-04-07
macros: add concat! macro

liushuyu:
- extracts parenthesis-matching logic into a function
- adds `concat!` macro

### DIFF
--- a/gcc/rust/expand/rust-macro-builtins.cc
+++ b/gcc/rust/expand/rust-macro-builtins.cc
@@ -34,17 +34,12 @@ make_string (Location locus, std::string value)
 			  PrimitiveCoreType::CORETYPE_STR, {}, locus));
 }
 
-/* Parse a single string literal from the given delimited token tree,
-   and return the LiteralExpr for it. Allow for an optional trailing comma,
-   but otherwise enforce that these are the only tokens.  */
+/* Match the end token of a macro given the start delimiter of the macro */
 
-std::unique_ptr<AST::LiteralExpr>
-parse_single_string_literal (AST::DelimTokenTree &invoc_token_tree,
-			     Location invoc_locus)
+static inline TokenId
+macro_end_token (AST::DelimTokenTree &invoc_token_tree,
+		 Parser<MacroInvocLexer> &parser)
 {
-  MacroInvocLexer lex (invoc_token_tree.to_token_stream ());
-  Parser<MacroInvocLexer> parser (std::move (lex));
-
   auto last_token_id = TokenId::RIGHT_CURLY;
   switch (invoc_token_tree.get_delim_type ())
     {
@@ -62,6 +57,22 @@ parse_single_string_literal (AST::DelimTokenTree &invoc_token_tree,
       rust_assert (parser.skip_token (LEFT_SQUARE));
       break;
     }
+
+  return last_token_id;
+}
+
+/* Parse a single string literal from the given delimited token tree,
+   and return the LiteralExpr for it. Allow for an optional trailing comma,
+   but otherwise enforce that these are the only tokens.  */
+
+std::unique_ptr<AST::LiteralExpr>
+parse_single_string_literal (AST::DelimTokenTree &invoc_token_tree,
+			     Location invoc_locus)
+{
+  MacroInvocLexer lex (invoc_token_tree.to_token_stream ());
+  Parser<MacroInvocLexer> parser (std::move (lex));
+
+  auto last_token_id = macro_end_token (invoc_token_tree, parser);
 
   std::unique_ptr<AST::LiteralExpr> lit_expr = nullptr;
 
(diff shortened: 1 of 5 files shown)


gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -252,4 +263,42 @@ MacroBuiltin::compile_error (Location invoc_locus, AST::MacroInvocData &invoc)
...
+	  auto lit_expr = parser.parse_literal_expr ();
+	  str += lit_expr->as_string ();
+	}
+      else
+	{
+	  rust_error_at (parser.peek_current_token ()->get_locus (),
+			 "argument must be a constant literal");
+	  return AST::ASTFragment::create_error ();
+	}
+      parser.maybe_skip_token (COMMA);
```
CohenArthur, line 295:
```suggestion
      auto literal = parser.parse_literal_expr();
      if (!literal) // or literal == nullptr
      {
          rust_error_at (parser.peek_current_token ()->get_locus (), "argument to %<concat%> must be a constant literal");
          return AST::ASTFragment::create_error ();
      }
     
      parser.maybe_skip_token (COMMA);
```
I think you might be able to simplify the logic here?

CohenArthur (commented):
Also I would suggest not using your fork's master for PRs :D Thank you for the contribution!

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -252,4 +263,42 @@ MacroBuiltin::compile_error (Location invoc_locus, AST::MacroInvocData &invoc)
...
+      if (token->is_literal () || token->get_id () == TRUE_LITERAL
+	  || token->get_id () == FALSE_LITERAL)
+	{
+	  auto lit_expr = parser.parse_literal_expr ();
+	  str += lit_expr->as_string ();
+	}
+      else
+	{
+	  rust_error_at (parser.peek_current_token ()->get_locus (),
+			 "argument must be a constant literal");
```
bjorn3, line 292: Rustc also accepts a macro invocation here which should expand to a string.

liushuyu: I think this is related to #1085. All the built-in macros in Rust GCC currently do not support nesting macro expressions.

liushuyu:
> Also I would suggest not using your fork's master for PRs :D Thank you for the contribution!

Ah, I will remember to use the "feature-name branch" method next time.

gcc/rust/expand/rust-macro-builtins.cc
liushuyu: Thank you. However, should we apply the following changes to `gcc/rust/lex/rust-token.h`?

```diff
diff --git a/gcc/rust/lex/rust-token.h b/gcc/rust/lex/rust-token.h
index bbc6d5ddf..3a7c63d4e 100644
--- a/gcc/rust/lex/rust-token.h
+++ b/gcc/rust/lex/rust-token.h
@@ -434,6 +434,8 @@ return *str;
       case STRING_LITERAL:
       case BYTE_CHAR_LITERAL:
       case BYTE_STRING_LITERAL:
+      case TRUE_LITERAL:
+      case FALSE_LITERAL:
        return true;
       default:
        return false;
```

philberty: Yeah that could be pretty useful in other passes

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -252,4 +263,40 @@ MacroBuiltin::compile_error (Location invoc_locus, AST::MacroInvocData &invoc)
...
+      auto lit_expr = parser.parse_literal_expr ();
+      if (lit_expr)
+	{
+	  str += lit_expr->as_string ();
+	}
+      else
+	{
+	  rust_error_at (parser.peek_current_token ()->get_locus (),
+			 "argument must be a constant literal");
+	  return AST::ASTFragment::create_error ();
```
CohenArthur, line 291: This is definitely a nitpick, but: Instead of returning early here, we could emit errors as we go, going through the whole loop. This would help in reporting to the user that there might be multiple places in which they're using non-literals in their invocation to `concat!()`. This could definitely be done in a separate PR so no worries about that, but if you want we can open up an issue for it. Then, after the loop, if an error has been seen, return an error `ASTFragment`

CohenArthur (approved):
I think this is definitely in a good place, thanks a lot @liushuyu ! I think we should wait a little for @dafaust to review this since he recently implemented a builtin macro, but otherwise I am good for this to be merged

gcc/rust/expand/rust-macro-builtins.cc
liushuyu: > This is definitely a nitpick, but: Instead of returning early here, we could emit errors as we go, going through the whole loop. This would help in reporting to the user that there might be multiple places in which they're using non-literals in their invocation to `concat!()`. This could definitely be done in a separate PR so no worries about that, but if you want we can open up an issue for it. Then, after the loop, if an error has been seen, return an error `ASTFragment`

Acknowledged. I thought we need to exit early to avoid reporting false-positives.

philberty: Yeah this is always a balance i think for now lets try and produce as many errors as possible then we can examine if that becomes a problem down the line.

philberty (approved):
LGTM

PR #1092, opened 2022-04-08
gcc/rust/Make-lang.in: add missing rust compiler driver

RomainNaour:
When building gccrs with Buildroot toolchain infrastructure, the gccrs
compiler driver is missing when intalling gcc.

This is due to missing depedency on gccrs$(exeext) in rust.all.cross
target.

With that fixed, the gcc toolchain with Rust support is correctly installed into Buildroot:

$ ./test/gccrs/host/bin/aarch64-linux-gccrs --version
aarch64-linux-gccrs (Buildroot 2022.02-442-g54d638fbd1-dirty) 12.0.1 20220118 (experimental)

Note: We probably needs gccrs-cross target like other supported languages.

Copyright assignment signed between Smile and the FSF to contribute to GNU tools.

### DIFF
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -140,7 +140,7 @@ check_rust_parallelize = 10
 # was specified as a configure option.
 rust.srcextra:
 
-rust.all.cross:
+rust.all.cross: gccrs$(exeext)
 
 # idk what this does but someone used it
 rust.start.encap: gccrs$(exeext)


CohenArthur (approved):
Thank you for the investigation and effort :D I would say it's still a little early to ship gccrs in BSPs haha

RomainNaour:
> Thank you for the investigation and effort :D I would say it's still a little early to ship gccrs in BSPs haha

Indeed :)
I'm mentoring an intern at Smile to try Rust tools with Buildroot and Yocto buildsystem.
gccrs is one of them but we need gccrs-cargo wrapper at some point since all rust based projects are using cargo.
Without cargo, we can't do much with gccrs and usual buildsystem (autotools, cmake, meson etc...), only hand written Makefiles is currently possible.

Note: It would be interesting if gccrs-cargo can be build without cargo to avoid dependency with rustc.

philberty:
> needs gccrs-cross target like

Yeah we should raise an issue for that, our makefile was the bare minimum to get the project setup but I would love to know if cross-compilation works for our front-end. We do have a build farm from mjw that building and testing directly on the host is working https://builder.wildebeest.org/buildbot/#/builders

we have a few failures on i386 but the 64 bit targets are working well :)

philberty (approved):

PR #1096, opened 2022-04-09
macros: add include! macro

liushuyu:
- add `include!` macro

### DIFF
--- a/gcc/rust/expand/rust-macro-builtins.cc
+++ b/gcc/rust/expand/rust-macro-builtins.cc
@@ -425,22 +425,24 @@ MacroBuiltin::include (Location invoc_locus, AST::MacroInvocData &invoc)
   if (lit_expr == nullptr)
     return AST::ASTFragment::create_error ();
 
-  std::string target_filename
+  std::string filename
     = source_relative_path (lit_expr->as_string (), invoc_locus);
+  auto target_filename
+    = Rust::Session::get_instance ().include_extra_file (std::move (filename));
 
-  RAIIFile target_file (target_filename.c_str ());
+  RAIIFile target_file (target_filename);
   Linemap *linemap = Session::get_instance ().linemap;
 
   if (target_file.get_raw () == nullptr)
     {
-      rust_error_at (lit_expr->get_locus (), "cannot open included file %s: %m",
-		     target_filename.c_str ());
+      rust_error_at (lit_expr->get_locus (),
+		     "cannot open included file %qs: %m", target_filename);
       return AST::ASTFragment::create_error ();
     }
 
-  rust_debug ("Attempting to parse included file %s", target_filename.c_str ());
+  rust_debug ("Attempting to parse included file %s", target_filename);
 
-  Lexer lex (target_filename.c_str (), std::move (target_file), linemap);
+  Lexer lex (target_filename, std::move (target_file), linemap);
   Parser<Lexer> parser (std::move (lex));
 
   auto parsed_items = parser.parse_items ();
--- a/gcc/rust/expand/rust-macro-builtins.cc
+++ b/gcc/rust/expand/rust-macro-builtins.cc
@@ -412,4 +412,58 @@ MacroBuiltin::cfg (Location invoc_locus, AST::MacroInvocData &invoc)
   return AST::ASTFragment ({literal_exp});
 }
 
+/* Expand builtin macro include!(), which includes a source file at the current
+ scope compile time. */
+
+AST::ASTFragment
+MacroBuiltin::include (Location invoc_locus, AST::MacroInvocData &invoc)
+{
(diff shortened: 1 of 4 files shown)


gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -252,4 +252,55 @@ MacroBuiltin::compile_error (Location invoc_locus, AST::MacroInvocData &invoc)
...
+  if (target_file.get_raw () == nullptr)
+    {
+      rust_error_at (lit_expr->get_locus (), "cannot open included file %s: %m",
+		     target_filename.c_str ());
+      return AST::ASTFragment::create_error ();
+    }
+
+  rust_debug ("Attempting to parse included file %s", target_filename.c_str ());
+
+  Lexer lex (target_filename.c_str (), std::move (target_file), linemap);
```
liushuyu, line 280: `target_filename.c_str ()`

I did not find a way to correctly move `target_filename`'s ownership to either `Lexer` or `Parser`.

This creates an issue where if there are warnings in an `include`d file, the diagnostic information will be garbage data.

CohenArthur (commented):
Thank you! I think if you figure out the lifetime/ownership issue for the string this is good to go :)

gcc/rust/expand/rust-macro-builtins.cc
CohenArthur: Yes, this is an issue we also faced when started fetching items from modules. The solution was to keep the string as a member of the `Module` class, which we cannot easily do here. If you can't figure it out, let me know and we'll try to work something out together (for example by adding a `private_data` field to the macro-invocation or something, even though I'm not a big fan of that haha).

liushuyu: > Yes, this is an issue we also faced when started fetching items from modules. The solution was to keep the string as a member of the `Module` class, which we cannot easily do here. 

I saw that, that's why I am kind of in a dilemma at the moment.

> If you can't figure it out, let me know and we'll try to work something out together (for example by adding a `private_data` field to the macro-invocation or something, even though I'm not a big fan of that haha).

Yes, that would make the type conversation a bit unsafe when casting data types from the `private_data` pointer.

gcc/rust/rust-session-manager.h
```diff
@@ -267,6 +271,12 @@ struct Session
   void parse_files (int num_files, const char **files);
   void init_options ();
 
+  inline const char *include_extra_file (std::string filename)
+  {
+    extra_files.push_back (std::move (filename));
+    return extra_files.back ().c_str ();
+  }
```
CohenArthur, line 278: I think this is a great idea. The API might be a little weird but I am unsure about what would be a better solution. We should open up an issue to also use this for module loading.

For now, I think it would be good if this function had a little more documentation. Something explaining that it returns a raw pointer to data stored in the session manager instance for example

CohenArthur (commented):
Great workaround :)

gcc/rust/rust-session-manager.h
liushuyu: > The API might be a little weird but I am unsure about what would be a better solution.

This is due to an issue of unknown cause I discovered when I was implementing this workaround. If you return a `std::string&`, then when you pass its `.c_str()` pointer later to `Lexer`, the pointer will point to an invalid stack address for unknown reasons. If you return a `const char*` pointer directly, it will return a heap address pointing directly to the target string.

I tried `asan` and `valgrind`, and surprisingly, they were unable to find anything.

CohenArthur: I don't know what your code was like, but if the reference you were returning referred to the moved value then it would make sense since the string would now be in an erroneous state? Or were you doing something like `return extra_files.back ()` too?

Regardless, I don't think it's especially bad that this returns a raw C pointer. But I think it should be documented so that we know the actual data is stored inside the session manager and someone using `include_extra_file` does not have to free the pointer they get or something like that.

liushuyu: > I don't know what your code was like, but if the reference you were returning referred to the moved value then it would make sense since the string would now be in an erroneous state? Or were you doing something like `return extra_files.back ()` too?

Yes, I was using `return extra_files.back ()` because obviously `move`d data can't be used anymore.

> 
> Regardless, I don't think it's especially bad that this returns a raw C pointer. But I think it should be documented so that we know the actual data is stored inside the session manager and someone using `include_extra_file` does not have to free the pointer they get or something like that.

Comments added.

liushuyu:
Sidenote:

- This implementation is naive and currently does not implement recursion limits.
- Due to macro hygiene issues, `include`d content may spill into other scopes.

CohenArthur (approved):
Great job, thanks a lot :D

PR #1099, opened 2022-04-09
Fix bad inherent overlap error

### LINKED ISSUE #1075: Bad inherent overlap error
I tried this code:

```rust
extern "rust-intrinsic" {
}

struct FatPtr<T> {
    data: *const T,
    len: usize,
}

union Repr<T> {
    rust: *const [T],
    rust_mut: *mut [T],
    raw: FatPtr<T>,
}

impl<T> *const [T] {
        // SAFETY: this is safe because `*const [T]` and `FatPtr<T>` have the same layout.
        // Only `std` can make this guarantee.
        let a = unsafe { Repr { rust: self }.raw };
        a.len
    }

    pub const fn as_ptr(self) -> *const T {
        self as *const T
    }
}

impl<T> *const T {
    pub const unsafe fn offset(self, count: isize) -> *const T {
        unsafe { offset(self, count) }
    }

    }

    pub const fn as_ptr(self) -> *const T {
        self as *const T
    }
}

```

Instead, this happened: 

```
<source>:24:9: error: duplicate definitions with name as_ptr
   24 |     pub const fn as_ptr(self) -> *const T {
      |         ^
......
   38 |     pub const fn as_ptr(self) -> *const T {
      |         ~
<source>:38:9: error: duplicate definitions with name as_ptr
   24 |     pub const fn as_ptr(self) -> *const T {
      |         ~
......
   38 |     pub const fn as_ptr(self) -> *const T {
      |         ^
```

### Meta

- What version of Rust GCC were you using, git sha if possible. e8b9587d3a0615f497cfe9c66995c1f21e42a536

philberty:
When we examine HIR::ImplBlock's we determine if an impl might overlap
another impl based on the Self type. So for example you might have a
generic structure Foo<T>(T), and an associated impl block for Foo<i32>, but
then go on to define an associated impl of Foo<T> the generic one will
overlap any associated impl hiding the generic implementation.

In this case we have two generic impl blocks

  *const [T]
  *const T

This means the *const T might overlap with the slice one since it is
generic. As bjorn3 pointed out in #1075 , the correct implementation is to
observe that [T] is constrained by size but untill we have the auto trait
of Sized we must example the two generic impls and just determine that
they are not-equal so for now this is the best implementation we can do.

Fixes #1075

### DIFF
--- a/gcc/rust/typecheck/rust-hir-inherent-impl-overlap.h
+++ b/gcc/rust/typecheck/rust-hir-inherent-impl-overlap.h
@@ -123,7 +123,26 @@ public:
 	      continue;
 
 	    if (query->can_eq (candidate, false))
-	      possible_collision (it->second, iy->second);
+	      {
+		// we might be in the case that we have:
+		//
+		// *const T vs *const [T]
+		//
+		// so lets use an equality check when the
+		// candidates are both generic to be sure we dont emit a false
+		// positive
+
+		bool a = query->is_concrete ();
+		bool b = candidate->is_concrete ();
+		bool both_generic = !a && !b;
+		if (both_generic)
+		  {
+		    if (!query->is_equal (*candidate))
+		      continue;
+		  }
+
+		possible_collision (it->second, iy->second);
+	      }
 	  }
       }
   }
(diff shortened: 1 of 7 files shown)


CohenArthur (approved):
LGTM! I think the issue number in the PR's comment might be wrong, as it points to one of Thomas' PRs regarding old libstdc++ issues :D

philberty:
thanks for noticing that it was actually #1075 not 1057 lol

PR #1101, opened 2022-04-09
Add helper as_string for DefIds

philberty:
This just adds a useful helper to as_string DefId's directly

### DIFF
--- a/gcc/rust/util/rust-mapping-common.h
+++ b/gcc/rust/util/rust-mapping-common.h
@@ -50,6 +50,14 @@ struct DefId
     return ((uint64_t) this->crateNum << 32 | this->localDefId)
 	   < ((uint64_t) other.crateNum << 32 | other.localDefId);
   }
+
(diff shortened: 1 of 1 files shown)


gcc/rust/util/rust-mapping-common.h
```diff
@@ -50,6 +50,14 @@ struct DefId
     return ((uint64_t) this->crateNum << 32 | this->localDefId)
 	   < ((uint64_t) other.crateNum << 32 | other.localDefId);
   }
+
+  std::string as_string () const
+  {
+    std::string buf;
+    buf += std::to_string (crateNum);
+    buf += std::to_string (localDefId);
```
CohenArthur, line 58: I think we might want to add a separator here so that the numbers aren't right next to each other?
```suggestion
    buf += std::to_string (crateNum);
    buf += " "; // or anything else
    buf += std::to_string (localDefId);
```

PR #1106, opened 2022-04-11
macros: fix an infinite loop in `concat!` macro parser

liushuyu:
- Fix concat macro parser issue which caused an infinite loop when invalid token is encountered

### DIFF
--- a/gcc/rust/expand/rust-macro-builtins.cc
+++ b/gcc/rust/expand/rust-macro-builtins.cc
@@ -287,9 +287,12 @@ MacroBuiltin::concat (Location invoc_locus, AST::MacroInvocData &invoc)
 	}
       else
 	{
-	  rust_error_at (parser.peek_current_token ()->get_locus (),
+	  auto current_token = parser.peek_current_token ();
+	  rust_error_at (current_token->get_locus (),
 			 "argument must be a constant literal");
 	  has_error = true;
+	  // Just crash if the current token can't be skipped
+	  rust_assert (parser.skip_token (current_token->get_id ()));
 	}
       parser.maybe_skip_token (COMMA);
     }
--- a/gcc/testsuite/rust/compile/builtin_macro_concat.rs
+++ b/gcc/testsuite/rust/compile/builtin_macro_concat.rs
@@ -3,13 +3,14 @@ macro_rules! concat {
 }
 
 fn main() {
-    // let not_literal = "identifier";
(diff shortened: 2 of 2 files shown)


gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -287,9 +287,16 @@ MacroBuiltin::concat (Location invoc_locus, AST::MacroInvocData &invoc)
 	}
       else
 	{
-	  rust_error_at (parser.peek_current_token ()->get_locus (),
+	  auto current_token = parser.peek_current_token ();
+	  rust_error_at (current_token->get_locus (),
 			 "argument must be a constant literal");
 	  has_error = true;
+	  if (!parser.skip_token (current_token->get_id ()))
```
CohenArthur, line 294: Oh my, that would be quite the race condition if our parser could not skip the token it's currently looking at! I think a shorter way to do this would be to add an assertion? Because imo we clearly want the compiler to ICE if we ever get into such a situation.
```suggestion
	  rust_assert (parser.skip_token (current_token->get_id ()));
```

If anyone has other input on whether or not we should crash here I'll gladly read it :)

CohenArthur (commented):
@liushuyu don't worry about the test, it's not your fault. This is definitely something that our CI should have reported, and is caused by `make check-rust` not reporting an error code when failing tests, meaning the verificiation step is us `grep`ping through the logs to find whether or not failures occured.

gcc/rust/expand/rust-macro-builtins.cc
liushuyu: > Oh my, that would be quite the race condition if our parser could not skip the token it's currently looking at! I think a shorter way to do this would be to add an assertion? Because imo we clearly want the compiler to ICE if we ever get into such a situation.
> 
> If anyone has other input on whether or not we should crash here I'll gladly read it :)

You are right. In which case, the compiler probably has bigger trouble than I could imagine. Just crashing could reduce the undefined behavior in this situation.

philberty:
Yep please don't feel bad about the PR. It's also something we should have found I the code review but also missed. We have not released the compiler yet so there is no harm done this is the point of development to find fix and find more issue again :)

philberty (commented):
This looks good, I like the assertion which will ensure we definitely make sure the stream is moving forward.

CohenArthur (approved):
LGTM! Thank you!

Issue #1102, opened 2022-04-11
infinite loop in testcase


philberty:
I tried this code:

gccrs/gcc/testsuite/rust/compile/builtin_macro_concat.rs

I expected to see this happen: to work

Instead, this happened: 

```
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
test.rs:9:12: error: argument must be a constant literal
```

### Meta

- What version of Rust GCC were you using, git sha if possible. 2669e80c17f4c9d8b534c73ae0c60a51de0b4659

philberty:
@liushuyu i am not sure why bors didn't pick this up.

liushuyu:
![image](https://user-images.githubusercontent.com/6829345/163653145-2f71877d-bc26-4925-910f-6d6783f22e38.png)

I think GitHub bugged out, let's manually close this issue then.

Issue #1084, opened 2022-04-06
macros: Any macro in the return position will result in Rust GCC stuck


liushuyu:
I tried this code:

```rust
macro_rules! add {
    ($a:expr) => { $a };
    ($a:expr, $($b:expr),+) => { $a + add!($($b),*) }
}

fn test() -> i32 {
    add!(add!(1, 2), add!(1, 2))
}

fn main() {
    let _ = test();
}
```

A reduced case for `gccrs`:

```rust
macro_rules! add {
    ($a:expr) => { $a };
    ($a:expr, $($b:expr),+) => { $a + add!($($b),*) }
}

fn main() -> i32 {
    add!(add!(1, 2))
}
```

I expected to see this happen: `gccrs` should compile the code, and the generated program should run and exit with `6` (the reduced case should exit with `3`).

Instead, this happened: `gccrs` stuck in an infinite loop in the macro expansion pass.

### Stack Trace

<details>
<summary>Click to show detailed stack trace</summary>
<p>

```
#0  0x00007ffff79cf4c4 in __memmove_avx_unaligned_erms () from /usr/lib/libc.so.6
#1  0x00000000020f5731 in std::basic_streambuf<char, std::char_traits<char> >::xsputn(char const*, long) ()
#2  0x00000000020e5f54 in std::basic_ostream<char, std::char_traits<char> >& std::__ostream_insert<char, std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*, long) ()
#3  0x0000000000934a03 in std::operator<< <std::char_traits<char> > (__c=<optimized out>, __out=...) at /usr/include/c++/11.2.0/ostream:518
#4  expand_format (fmt=<optimized out>) at ../../gcc/rust/rust-diagnostics.cc:42
#5  expand_message(const char *, typedef __va_list_tag __va_list_tag *) (fmt=<optimized out>, ap=0x7fffffffc788) at ../../gcc/rust/rust-diagnostics.cc:114
#6  0x00000000009353b2 in Rust::Error::Error (this=this@entry=0x7fffffffc960, location=..., fmt=fmt@entry=0x2139080 "unrecognised token %qs for start of %s")
    at ../../gcc/rust/rust-diagnostics.cc:239
#7  0x0000000000a50d65 in Rust::Parser<Rust::MacroInvocLexer>::parse_item (this=0x7fffffffcb80, called_from_statement=called_from_statement@entry=true)
[...]

PR #1107, opened 2022-04-11
tests: Improve timeout handling

liushuyu:
- add a 10-second timeout for individual compile test
- add a xfail test case to ensure the timeout mechanism is working

### DIFF
--- a/gcc/testsuite/lib/rust.exp
+++ b/gcc/testsuite/lib/rust.exp
@@ -166,6 +166,10 @@ proc rust_target_compile { source dest type options } {
     global gluefile wrap_flags
     global ALWAYS_RUSTFLAGS
     global RUST_UNDER_TEST
+    global individual_timeout
+
+    # HACK: guard against infinite loops in the compiler
+    set individual_timeout 10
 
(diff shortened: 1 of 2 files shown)


liushuyu:
This patch is although a hack, could help to prevent issues like:

- #1102 
- #1084 

In fact, the added XFAIL test case is from #1084.

philberty (approved):
This seems good to me but I would rather @dkm or @tschwinge approve/merge this one since they have more experience than me here.

CohenArthur:
Since the test case is passing, wouldn't that cause the same issue of having potentially infinite-looping code merged into master? It seems the only extra output is `WARNING: program timed out` which we should handle properly to report as a CI failure

liushuyu:
> Since the test case is passing, wouldn't that cause the same issue of having potentially infinite-looping code merged into master? It seems the only extra output is `WARNING: program timed out` which we should handle properly to report as a CI failure

As you can see, the test case is an `XFAIL` test case where it's expected to fail. In reality, timing out would produce an excess error with no error location (no line number). In which case, if your test case is well-written (does not have a catch-all case for excess errors), timing out will fail the CI.

PR #1110, opened 2022-04-12
Add Reachability visitors for items with generics

CohenArthur:
This factors generics' predicates visiting in the `ReachabilityVisitor` and calls the function in other items with generic parameters

### DIFF
--- a/gcc/rust/privacy/rust-reachability.cc
+++ b/gcc/rust/privacy/rust-reachability.cc
@@ -31,6 +31,13 @@ maybe_get_vis_item (std::unique_ptr<HIR::Item> &item)
   return static_cast<HIR::VisItem *> (item.get ());
 }
 
+ReachLevel
+ReachabilityVisitor::get_reachability_level (
+  const HIR::Visibility &item_visibility)
+{
+  return item_visibility.is_public () ? current_level : ReachLevel::Unreachable;
+}
+
 void
 ReachabilityVisitor::visit_generic_predicates (
   const std::vector<std::unique_ptr<HIR::GenericParam>> &generics,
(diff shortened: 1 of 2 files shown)


gcc/rust/privacy/rust-reachability.cc
```diff
@@ -31,6 +31,47 @@ maybe_get_vis_item (std::unique_ptr<HIR::Item> &item)
   return static_cast<HIR::VisItem *> (item.get ());
 }
 
+ReachLevel
+ReachabilityVisitor::initial_reach (const HIR::Visibility &item_visibility)
```
philberty, line 35: could this be called get_reachability_level?

gcc/rust/privacy/rust-reachability.cc
```diff
@@ -31,6 +31,47 @@ maybe_get_vis_item (std::unique_ptr<HIR::Item> &item)
...
+{
+  if (item_visibility.is_public ())
```
philberty, line 37: might be a good case to use a ternary statement here, if there is more logic to add then leave it as is.

gcc/rust/privacy/rust-reachability.h
```diff
@@ -46,6 +46,19 @@ class ReachabilityVisitor : public HIR::HIRVisItemVisitor
...
+   * them as reachable or not.
+   */
+  void visit_generic_predicates (
+    const std::vector<std::unique_ptr<HIR::GenericParam>> &generics,
+    ReachLevel item_reach);
+
+  /**
+   * Get the initial reach level for an item based on its visibility.
+   */
+  ReachLevel initial_reach (const HIR::Visibility &item_visibility);
```
philberty, line 60: this could be a static function since you don't need access to self here

gcc/rust/privacy/rust-reachability.cc
```diff
@@ -31,6 +31,47 @@ maybe_get_vis_item (std::unique_ptr<HIR::Item> &item)
...
+
+void
+ReachabilityVisitor::visit_generic_predicates (
+  const std::vector<std::unique_ptr<HIR::GenericParam>> &generics,
+  ReachLevel item_reach)
+{
+  if (item_reach == ReachLevel::Unreachable)
+    return;
+
+  for (auto &generic : generics)
```
philberty, line 51: can this be const auto&

gcc/rust/privacy/rust-reachability.cc
```diff
@@ -31,6 +31,47 @@ maybe_get_vis_item (std::unique_ptr<HIR::Item> &item)
...
+    {
+      if (generic->get_kind () == HIR::GenericParam::TYPE)
+	{
+	  TyTy::BaseType *generic_ty = nullptr;
+	  rust_assert (
```
philberty, line 56: i prefer to do:
```
bool ok = ty_ctx.lookup_type(...)
rust_assert(ok)
```

This makes a gdb session easier to work with.

gcc/rust/privacy/rust-reachability.cc
```diff
@@ -31,6 +31,47 @@ maybe_get_vis_item (std::unique_ptr<HIR::Item> &item)
...
+	    ty_ctx.lookup_type (generic->get_mappings ().get_hirid (),
+				&generic_ty));
+
+	  // FIXME: Can we really get anything else than a TyTy::PARAM here?
+	  // Should we change this to an assertion instead?
+	  if (generic_ty->get_kind () == TyTy::PARAM)
```
philberty, line 62: yeah thats that right, this might as well be an assertion, I should really update that in the rest of the type resolution steps.

philberty (changes requested):
Nice work, this is just very minor style nit's.

gcc/rust/privacy/rust-reachability.h
CohenArthur: Oops - this is a typo actually. We should be returning `current_level` instead of `ReachLevel::Reachable`

philberty (approved):
looks good

PR #1111, opened 2022-04-12
Add AST Private Visibilities

CohenArthur:
When parsing a visibility in `parse_visibility`, it is not an error to
not have a pub token: It simply means we want to create a private
visibility. If we had C++14 or another language, we could instead
represent all visibilities as an optional<AST::Visibility> where the
Visibility class would not need to change. But I think the best course
of action for our case is to instead keep visibilities even when they
are private and have a special case in the `VisKind` enumeration.

This also enables HIR lowering of visibilities to be performed properly for private items

### DIFF
--- a/gcc/rust/ast/rust-ast-full-test.cc
+++ b/gcc/rust/ast/rust-ast-full-test.cc
@@ -306,17 +306,19 @@ SimplePath::as_string () const
 std::string
 Visibility::as_string () const
 {
-  switch (public_vis_type)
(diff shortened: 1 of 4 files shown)


philberty (commented):
Ah this is much better to default to private since we don't really need an error state anyway for the AST::Visibility, since we will simply get an unknown expected keyword parser error.

philberty (approved):
LGTM

Issue #977, opened 2022-02-25
Implement `env!()` builtin macro


CohenArthur:
`env!()` is part of the list of builtin macros enumerated [here](#927)

Its description is as follows:
> Inspects an environment variable at compile time

The goal is to fetch environment variables passed to gccrs, and expand them into the user's compiled code.
```fish
arthur@platypus ~/G/gccrs (add-assert-macro)> cat test.rs
fn main() {
    let s = env!("CUSTOM");

    println!("{}", s);
}
arthur@platypus ~/G/gccrs (add-assert-macro)> rustc test.rs
error: environment variable `CUSTOM` not defined
 --> test.rs:2:13
  |
2 |     let s = env!("CUSTOM");
  |             ^^^^^^^^^^^^^^
  |
  = note: this error originates in the macro `env` (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to previous error

arthur@platypus ~/G/gccrs (add-assert-macro) [1]> CUSTOM='hi!' rustc test.rs
arthur@platypus ~/G/gccrs (add-assert-macro)> ./test
hi!
```
The macro should not expand to a function call that fetches environment variables during the user's program's execution.
```sh
arthur@platypus ~/G/gccrs (add-assert-macro)> ./test
hi!
arthur@platypus ~/G/gccrs (add-assert-macro)> CUSTOM='something else!' ./test
hi!
```

The original macro contains two match arms, including one with a custom error message:
```rust
    macro_rules! env {
        ($name:expr $(,)?) => {{ /* compiler built-in */ }};
        ($name:expr, $error_msg:expr $(,)?) => {{ /* compiler built-in */ }};
    }
```

You should try and support the first form first, by simply fetching an environment variable and creating an associated string using the `make_string` function defined in [`gcc/rust/expand/rust-macro-builtins.cc`](https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/expand/rust-macro-builtins.cc)

The second form could be implemented in a later PR for simplicity

bjorn3:
Rustc has a dep file extension showong which env vars are read and if valid UTF-8 what value saod env vars had. This is used by cargo to rebuild if the env vars change.

CohenArthur:
[...]

PR #1113, opened 2022-04-12
macros: Add env! macro

omachota:
Added the `env!()` macro and relevant test cases

Fixes: #977 

Signed-off-by: Ondřej Machota <ondrejmachota@gmail.com>

### DIFF
--- a/gcc/rust/expand/rust-macro-builtins.cc
+++ b/gcc/rust/expand/rust-macro-builtins.cc
@@ -306,4 +306,74 @@ MacroBuiltin::concat (Location invoc_locus, AST::MacroInvocData &invoc)
   return AST::ASTFragment ({node});
 }
 
+/* Expand builtin macro env!(), which inspects an environment variable at
+   compile time. */
+
+AST::ASTFragment
+MacroBuiltin::env (Location invoc_locus, AST::MacroInvocData &invoc)
+{
+  auto invoc_token_tree = invoc.get_delim_tok_tree ();
+  MacroInvocLexer lex (invoc_token_tree.to_token_stream ());
+  Parser<MacroInvocLexer> parser (std::move (lex));
+
+  auto last_token_id = macro_end_token (invoc_token_tree, parser);
+
+  if (parser.peek_current_token ()->get_id () != STRING_LITERAL)
+    {
+      if (parser.peek_current_token ()->get_id () == last_token_id)
+	rust_error_at (invoc_locus, "env! takes 1 or 2 arguments");
+      else
+	rust_error_at (parser.peek_current_token ()->get_locus (),
+		       "argument must be a string literal");
+      return AST::ASTFragment::create_error ();
+    }
+
+  auto lit_expr = parser.parse_literal_expr ();
+  auto comma_skipped = parser.maybe_skip_token (COMMA);
+
+  std::unique_ptr<AST::LiteralExpr> error_expr = nullptr;
+
+  if (parser.peek_current_token ()->get_id () != last_token_id)
+    {
+      if (!comma_skipped)
+	{
+	  rust_error_at (parser.peek_current_token ()->get_locus (),
+			 "expected token: %<,%>");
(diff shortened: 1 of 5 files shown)


gcc/testsuite/rust/execute/torture/execute.exp
```diff
@@ -26,6 +26,7 @@ dg-init
 set saved-dg-do-what-default ${dg-do-what-default}
 
 set dg-do-what-default "run"
+setenv ENV_MACRO_TEST VALUE
```
CohenArthur, line 29: I'd rather we removed that line...

gcc/testsuite/rust/execute/torture/builtin_macro_env.rs
```diff
@@ -0,0 +1,21 @@
+// { dg-output "VALUE\n" }
```
CohenArthur, line 1:
```suggestion
// { dg-output "VALUE\n" }
// { dg-set-compiler-env-var ENV_MACRO_TEST "VALUE" }
```
... And add this. Taken directly from https://gcc.gnu.org/onlinedocs/gccint/Directives.html for future references

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -306,4 +306,71 @@ MacroBuiltin::concat (Location invoc_locus, AST::MacroInvocData &invoc)
...
+      if (parser.peek_current_token ()->get_id () != STRING_LITERAL)
+	{
+	  rust_error_at (parser.peek_current_token ()->get_locus (),
+			 "argument must be a string literal");
+	  return AST::ASTFragment::create_error ();
+	}
+
+      error_expr = parser.parse_literal_expr ();
+      parser.maybe_skip_token (COMMA);
+    }
```
CohenArthur, line 348: This code allows the following invocation to be parsed properly while it should be rejected:
```rust
env!("ENV" "error message")
```
(Note the missing comma).

I suggest adding it as a test case as it's a bit of an annoying one :D

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -306,4 +306,71 @@ MacroBuiltin::concat (Location invoc_locus, AST::MacroInvocData &invoc)
...
+  if (env_value == nullptr)
+    {
+      if (error_expr == nullptr)
+	{
+	  rust_error_at (invoc_locus, "environment variable '%s' not defined",
+			 lit_expr->as_string ().c_str ());
+	}
+      else
+	rust_error_at (invoc_locus, "%s", error_expr->as_string ().c_str ());
+      return AST::ASTFragment::create_error ();
```
CohenArthur, line 369:
```suggestion
      if (error_expr == nullptr)
	  rust_error_at (invoc_locus, "environment variable '%s' not defined",
			 lit_expr->as_string ().c_str ());
      else
	rust_error_at (invoc_locus, "%s", error_expr->as_string ().c_str ());
      return AST::ASTFragment::create_error ();
```
Very tiny nitpick

gcc/rust/expand/rust-macro-builtins.cc
CohenArthur: It's probably a matter of `if (parser.maybe_skip_token(COMMA) { ... }` to parse the given error message only if a first comma is present

CohenArthur (changes requested):
Thank you for the pull request! This is a great start!

gcc/testsuite/rust/execute/torture/builtin_macro_env.rs
omachota: Thank you for your help. My solution was awkward. I knew there had to be a better one :)

gcc/rust/expand/rust-macro-builtins.cc
omachota: Nice catch :)

CohenArthur (approved):
Great job! Thanks a lot!!!

PR #1112, opened 2022-04-12
rust: Clang/macOS Testing

simonpcook:
This adds a version of the build-and-check job that runs with clang on
macOS.

PR #1114, opened 2022-04-13
rust: Use -Otarget when building and logging warnings

simonpcook:
This will provide some synchronization for output lines, and so will
make comparisons with known warnings as part of CI more reliable.


simonpcook:
Given bors tripped up on #1112 this should make that comparison more robust (if I've read the man page properly)

CohenArthur (approved):
Never used that option but I understand the same thing from the man!

philberty (approved):

PR #1118, opened 2022-04-14
Add reachability visitor to Enum variants

CohenArthur:
This visits all of an enum's variants and their fields if present. To do that properly, this adds a new `EnumItemKind` enum which allows static casting when visiting each variant of the enum (kept as an `EnumItem` class which is derived three times)

### DIFF
--- a/gcc/rust/hir/tree/rust-hir-item.h
+++ b/gcc/rust/hir/tree/rust-hir-item.h
@@ -1588,7 +1588,7 @@ public:
 
   std::string as_string () const;
 
-  Analysis::NodeMapping get_mappings () { return mappings; }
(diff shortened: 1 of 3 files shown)


gcc/rust/hir/tree/rust-hir-item.h
```diff
@@ -1644,12 +1644,19 @@ class TupleStruct : public Struct
 class EnumItem : public Item
 {
   Identifier variant_name;
-
   Location locus;
 
 public:
   virtual ~EnumItem () {}
 
+  enum EnumItemKind
```
philberty, line 1652: This enum will be really nice so we can get rid of the separate enum-visitors in the type-checker

philberty (approved):
LGTM

PR #1124, opened 2022-04-18
CI: do not push Docker image in a fork

liushuyu:
- ci: skip docker image push when in a fork


CohenArthur (approved):
Sounds good

philberty:
This will make @tschwinge happy :)

philberty (approved):
Nice, I was unable to find a way to do this but this seems like the best solution for now

PR #1135, opened 2022-04-19
Don't ICE on traits' `Self` generic parameters

### LINKED ISSUE #1128: rust1: internal compiler error: in visit_generic_predicates, at rust/privacy/rust-reachability.cc:56
I tried this code:

```rust
pub trait Hasher {
    fn write_u8(&mut self, i: u8) {
        self.write(&[i])
    }
}
```

I expected to see this happen: It works.

Instead, this happened:

```
rust1: internal compiler error: in visit_generic_predicates, at rust/privacy/rust-reachability.cc:56
0x204e319 internal_error(char const*, ...)
	???:0
0x8027ef fancy_abort(char const*, int, char const*)
	???:0
0x9ceb45 Rust::Privacy::Resolver::resolve(Rust::HIR::Crate&)
	???:0
0x8c8927 Rust::Session::parse_file(char const*)
	???:0
0x8c9548 Rust::Session::parse_files(int, char const**)
	???:0
Compiler returned: 1
```

### Meta

CohenArthur:
The `Self` type of a Trait is part of its generic parameter list. This
fails lookup since it's not a properly defined type. We need to be able
to identify `Self` parameters and skip them in the reachability pass.

Closes #1128

### DIFF
--- a/gcc/rust/privacy/rust-reachability.cc
+++ b/gcc/rust/privacy/rust-reachability.cc
@@ -53,6 +53,7 @@ ReachabilityVisitor::visit_generic_predicates (
 	  TyTy::BaseType *generic_ty = nullptr;
 	  auto ok = ty_ctx.lookup_type (generic->get_mappings ().get_hirid (),
 					&generic_ty);
+
 	  rust_assert (ok);
 	  rust_assert (generic_ty->get_kind () == TyTy::PARAM);
 
--- a/gcc/rust/typecheck/rust-hir-type-check-type.h
+++ b/gcc/rust/typecheck/rust-hir-type-check-type.h
@@ -189,6 +189,8 @@ public:
 	return nullptr;
       }
 
+    resolver.context->insert_type (param->get_mappings (), resolver.resolved);
+
     return resolver.resolved;
   }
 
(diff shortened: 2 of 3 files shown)


gcc/rust/privacy/rust-reachability.cc
```diff
@@ -53,7 +53,13 @@ ReachabilityVisitor::visit_generic_predicates (
 	  TyTy::BaseType *generic_ty = nullptr;
 	  auto ok = ty_ctx.lookup_type (generic->get_mappings ().get_hirid (),
 					&generic_ty);
-	  rust_assert (ok);
+
+	  // FIXME: This fails lookup on the `Self` generic parameter of the
+	  // trait definition. We can skip it for now, but that's not a real fix
+	  if (!ok)
```
philberty, line 59: This is interesting you should put a debug line in to see what the ID is here and compare with what it is over in rust-hir-trait-resolve.h line 138 and ou should be able to call self->debug() we might be missing an insert call

philberty:
I've found the real issue here the fix is simple but it causes a regression in using qualified paths within traits since there is no associated impl block its working by accident right now the relevant test case is gcc/testsuite/rust/execute/torture/trait3.rs.

Do you mind if I take this over?

CohenArthur:
> I've found the real issue here the fix is simple but it causes a regression in using qualified paths within traits since there is no associated impl block its working by accident right now the relevant test case is gcc/testsuite/rust/execute/torture/trait3.rs.

Great! Nice find. I had completely forgotten about this and was neck-deep in simple path resolving haha

> Do you mind if I take this over?

Please do! The only interesting thing in this PR is the test case, so you can close it whenever you want

philberty:
This will be fixed as part of https://github.com/Rust-GCC/gccrs/pull/1190

PR #1137, opened 2022-04-19
ast: lower: Split rust-lower-base in its own source file

CohenArthur:
The implementation of all functions declared in `rust-ast-lower-base.h` was a big offender of our "definitions in headers" problem, spanning over 900 lines of code. We can do the same for basically all headers in `gcc/rust/hir` which all contain some definitions that have accumulated over time


philberty (approved):
Yeay we need to keep doing this slowly but surely! :D

PR #1142, opened 2022-04-20
backend: process more attributes

liushuyu:
- handle `cold`, `link_section`, `no_mangle` and `deprecated` attributes

### DIFF
--- a/gcc/rust/backend/rust-compile-base.cc
+++ b/gcc/rust/backend/rust-compile-base.cc
@@ -30,6 +30,24 @@
 namespace Rust {
 namespace Compile {
 
+bool
+should_mangle_item (const AST::AttrVec &attrs)
+{
+  for (const auto &attr : attrs)
+    {
+      if (attr.get_path ().as_string ().compare ("no_mangle") == 0)
+	{
+	  if (attr.has_attr_input ())
+	    rust_error_at (
+	      attr.get_locus (),
+	      "attribute %<no_mangle%> does not accept any arguments");
+	  return false;
+	}
+    }
+
+  return false;
+}
+
 void
 HIRCompileBase::setup_attributes_on_fndecl (
   tree fndecl, bool is_main_entry_point, HIR::Visibility &visibility,
@@ -55,6 +73,12 @@ HIRCompileBase::setup_attributes_on_fndecl (
       bool is_inline = attr.get_path ().as_string ().compare ("inline") == 0;
       bool is_must_use
 	= attr.get_path ().as_string ().compare ("must_use") == 0;
+      bool is_cold = attr.get_path ().as_string ().compare ("cold") == 0;
+      bool is_deprecated
+	= attr.get_path ().as_string ().compare ("deprecated") == 0;
+      bool is_link_section
+	= attr.get_path ().as_string ().compare ("link_section") == 0;
+      bool no_mangle = attr.get_path ().as_string ().compare ("no_mangle") == 0;
       if (is_inline)
 	{
 	  handle_inline_attribute_on_fndecl (fndecl, attr);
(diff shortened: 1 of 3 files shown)


gcc/rust/backend/rust-compile-base.cc
```diff
@@ -63,7 +87,106 @@ HIRCompileBase::setup_attributes_on_fndecl (
...
+  else if (input_type == AST::AttrInput::AttrInputType::TOKEN_TREE)
+    {
+      // handle #[deprecated(since = "...", note = "...")]
+      const auto &option = static_cast<const AST::DelimTokenTree &> (input);
+      AST::AttrInputMetaItemContainer *meta_item = option.parse_to_meta_item ();
+      for (const auto &item : meta_item->get_items ())
+	{
+	  // FIXME: does not compile
+	  // std::unique_ptr<AST::MetaNameValueStr> inner =
+	  // item->to_meta_name_value_str ();
```
liushuyu, line 184: This part throws a strange error about `Rust::AST::MetaNameValueStr` is not a complete type and can't be contained in a `std::unique_ptr`.

gcc/rust/backend/rust-compile-base.cc
```diff
@@ -63,7 +87,106 @@ HIRCompileBase::setup_attributes_on_fndecl (
...
+  if (decl_section_name (fndecl))
+    {
+      rust_warning_at (attr.get_locus (), 0, "section name redefined");
+    }
+
+  set_decl_section_name (fndecl, msg_str.c_str ());
+}
+
+void
+HIRCompileBase::handle_deprecated_attribute_on_fndecl (
```
liushuyu, line 154: `deprecated` (`TREE_DEPRECATED`) is an "advisory" marking in GCC, which must be checked by the language frontend itself.

philberty:
Can you do each attribute in separate PR's? I think it makes it easier for review. The TREE_DEPRECIATED part is interesting to think about, since we will likely need to add code to the code-generation pass to emit warnings about it, or it might be enough to detect it by iterating all the fndecls and checking if TREE_USED and TREE_DEPRECIATED and emitting warnings that way but it would be nice to emit the warning from the source location which we could intercept when we emit a CALL_EXPR and we resolve the function address. but there might need to be special handling for dyn trait since its just function pointers here.

liushuyu:
> Can you do each attribute in separate PR's? I think it makes it easier for review. 

Understood. I will split them later.

> The TREE_DEPRECIATED part is interesting to think about, since we will likely need to add code to the code-generation pass to emit warnings about it, or it might be enough to detect it by iterating all the fndecls and checking if TREE_USED and TREE_DEPRECIATED and emitting warnings that way but it would be nice to emit the warning from the source location which we could intercept when we emit a CALL_EXPR and we resolve the function address. but there might need to be special handling for dyn trait since its just function pointers here.

I see, and yes this involves a lot of work there. I will probably only add a stub here to mark `TREE_DEPRECATED` for now, since the analysis pass is more difficult to change.

philberty:
That sounds like a plan. Thanks for splitting them up. Also I think your right lets mark them as TREE_DEPRECATED, I think its completely valid to do this at the GCC tree level. We might be able to get away with adding in a check for TREE_DEPRECATED over here:

https://github.com/Rust-GCC/gccrs/blob/fc22f12c9c707b258f35a1bab0e8154441b972b8/gcc/rust/rust-gcc.cc#L1804

There are things in gcc/tree.h to get the actual function decl from FUNCTION_POINTER_TYPE_P if possible. So in that scenario it might be a decent catch all but I don't know about dyn traits it might be possible to get a similar check when we setup the dyn trait object.

Let me see if I cant how to get the fndecl from FUNCTION_POINTER_TYPE_P for you I've done it before at one point.

philberty:
Actually the code for it is already there lol:

You could add in an if statement:

```
  tree fndecl = fn;
  if (TREE_CODE (fndecl) == ADDR_EXPR)
    fndecl = TREE_OPERAND (fndecl, 0);

if (TREE_CODE (fndecl) == FUNCTION_DECL && TREE_DEPRECATED(fndecl)) {
  RichLocation r(location);
  r.add_range(Location( DECL_SOURCE_LOCATION(fndecl)));
  rust_warning_at(r, "call to deprecated unction");
}
```

liushuyu:
Closing. This pull request has been split into the following:

- #1148 
- #1149 
- #1150 
- #1151

PR #1145, opened 2022-04-21
Fix generic param redefined bug

### LINKED ISSUE #1131: "error: generic param redefined multiple times" when trying to use the same generic param multiple times in the same extern block
I tried this code:

```rust
extern "rust-intrinsic" {
    fn size_of<T>() -> usize;
    fn transmute<T, U>(_: T) -> U;
}
```

I expected to see this happen: It compiles

Instead, this happened:

```
<source>:3:18: error: generic param redefined multiple times
    3 |     fn transmute<T, U>(_: T) -> U;
      |                  ^
<source>:2:16: error: was defined here
    2 |     fn size_of<T>() -> usize;
      |                ^
Compiler returned: 1
```

### Meta

philberty:
When we name-resolve generic parameters their declaration gets inserted in
to the upper-most rib on the stack. Then when this is referenced we lookup
the relevant binding starting from the uppermost rib this lead to the 2nd
extern item adding the same binding into the same rib which caused this
clash. To fix this we need to have a seperate rib for each declaration
so as reusing the same names don't clash with each other.

Fixes #1131

### DIFF
--- a/gcc/rust/resolve/rust-ast-resolve-item.h
+++ b/gcc/rust/resolve/rust-ast-resolve-item.h
@@ -955,6 +955,15 @@ public:
 
   void visit (AST::ExternalFunctionItem &function) override
   {
+    NodeId scope_node_id = function.get_node_id ();
+    resolver->get_name_scope ().push (scope_node_id);
+    resolver->get_type_scope ().push (scope_node_id);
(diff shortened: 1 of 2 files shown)


CohenArthur (approved):
LGTM

Issue #1146, opened 2022-04-21
Code generation coercion sites dont handle complex cases


philberty:
I have a fix for this, we are missing the coercion rule for an array to a slice. There is one other part that needs fixed which is slightly related to this which is how he handles coercion_sites in general the code-generation pass doesn't really handle any complex case but we have all the building blocks to make this robust by reusing the adjustment classes we already use for method resolution so there is a refactor needed to take advantage of that.

_Originally posted by @philberty in https://github.com/Rust-GCC/gccrs/issues/1129#issuecomment-1105653841_

see https://godbolt.org/z/EYdEoETaj

PR #1147, opened 2022-04-21
Add missing coercion rule from array to slice

### LINKED ISSUE #1129: Incorrect "error: expected [[u8]] got [[u8:CAPACITY]]"
I tried this code:

```rust
pub trait Hasher {
    fn finish(&self) -> u64;
    fn write_u8(&mut self, i: u8) {
        self.write(&[i])
    }
}

struct SipHasher;

impl Hasher for SipHasher {
    #[inline]
    fn write(&mut self, msg: &[u8]) {
        loop {}
    }

    #[inline]
    fn finish(&self) -> u64 {
        0
    }
}
```

I expected to see this happen: It compiles.

Instead, this happened:

```
<source>:5:21: error: expected [[u8]] got [[u8:CAPACITY]]
    3 |     fn write(&mut self, bytes: &[u8]);
      |                                 ~
    4 |     fn write_u8(&mut self, i: u8) {
    5 |         self.write(&[i])
      |                     ^
<source>:5:20: error: expected [& [u8]] got [& [u8:CAPACITY]]
    3 |     fn write(&mut self, bytes: &[u8]);
      |                                ~
    4 |     fn write_u8(&mut self, i: u8) {
    5 |         self.write(&[i])
      |                    ^
<source>:5:20: error: Type Resolution failure on parameter
<source>:5:9: error: failed to lookup type to MethodCallExpr
    5 |         self.write(&[i])
      |         ^
<source>:5:9: error: failed to type resolve expression
<source>:4:35: error: expected [()] got [<tyty::error>]
    4 |     fn write_u8(&mut self, i: u8) {
      |                            ~      ^
Compiler returned: 1
```

Note that the `Hasher` trait needs to be used to get this error. Rustc checks default method bodies even when a trait is not used.

### Meta

philberty:
Arrays are coercible into slices, this adds the missing type-resolution
the rule which works for now. The other part of this fix is described in #1146 
for coercion_site to be recursive and reuse the adjustment classes so that
we reuse as much code as possible and handle complex coercion sites.

Fixes #1129

### DIFF
--- a/gcc/testsuite/rust/compile/issue-1129-2.rs
+++ b/gcc/testsuite/rust/compile/issue-1129-2.rs
@@ -1,0 +1,22 @@
+// { dg-additional-options "-w" }
+pub trait Hasher {
+    fn finish(&self) -> u64;
+    fn write(&mut self, bytes: &[u8]);
+    fn write_u8(&mut self, i: u8) {
+        self.write(&[i])
+    }
+}
+
+struct SipHasher;
+
+impl Hasher for SipHasher {
+    #[inline]
+    fn write(&mut self, msg: &[u8]) {
+        loop {}
+    }
+
+    #[inline]
+    fn finish(&self) -> u64 {
+        0
+    }
+}
--- a/gcc/rust/typecheck/rust-tyty-coercion.h
+++ b/gcc/rust/typecheck/rust-tyty-coercion.h
@@ -886,6 +886,22 @@ public:
 			      TyVar (base_resolved->get_ref ()));
   }
 
+  void visit (ArrayType &type) override
+  {
+    // check base type
+    auto base_resolved
+      = base->get_element_type ()->unify (type.get_element_type ());
+    if (base_resolved == nullptr)
+      {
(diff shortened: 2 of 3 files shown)


gcc/testsuite/rust/compile/issue-1129-2.rs
```diff
@@ -0,0 +1,22 @@
+// { dg-additional-options "-w" }
+pub trait Hasher {
+    fn finish(&self) -> u64;
+    fn write(&mut self, bytes: &[u8]);
+    fn write_u8(&mut self, i: u8) {
+        self.write(&[i])
+    }
+}
```
CohenArthur, line 8: Oh interesting, this does not ICE in the reachability pass since there is an implementation of the trait

CohenArthur (approved):
LGTM!

gcc/testsuite/rust/compile/issue-1129-2.rs
philberty: Ah maybe there is something missing for error handling in type-bound-predicates

PR #1148, opened 2022-04-22
backend: handle cold attribute

liushuyu:
- handle the `cold` attribute


philberty (approved):
LGTM, does rust support this attribute on labels by any chance (if it does we can do that in another PR and wire up some handling for that)

liushuyu:
> LGTM, does rust support this attribute on labels by any chance (if it does we can do that in another PR and wire up some handling for that)

I did a test just now:

```rust
fn test1() {
    #[cold]
    'lbl:
    for i in 1..10 {
        break 'lbl;
    }
}

fn main() {}
```

And `rustc` reported:

```
warning: attribute should be applied to a function
 --> test10.rs:2:5
  |
2 |       #[cold]
  |       ^^^^^^^
3 | /     'lbl:
4 | |     for i in 1..10 {
5 | |         break 'lbl;
6 | |     }
  | |_____- not a function
  |
  = note: `#[warn(unused_attributes)]` on by default
```

PR #1149, opened 2022-04-22
backend: handle no_mangle attribute

liushuyu:
- handle the `no_mangle` attribute

### DIFF
--- a/gcc/rust/backend/rust-compile-base.cc
+++ b/gcc/rust/backend/rust-compile-base.cc
@@ -26,32 +26,21 @@
 
 #include "fold-const.h"
 #include "stringpool.h"
+#include "attribs.h"
 
 namespace Rust {
 namespace Compile {
 
-bool
-should_mangle_item (const AST::AttrVec &attrs)
+bool inline should_mangle_item (const tree fndecl)
 {
-  for (const auto &attr : attrs)
-    {
-      if (attr.get_path ().as_string ().compare ("no_mangle") == 0)
-	{
-	  if (attr.has_attr_input ())
-	    rust_error_at (
-	      attr.get_locus (),
-	      "attribute %<no_mangle%> does not accept any arguments");
-	  return false;
-	}
-    }
-
-  return true;
+  return lookup_attribute ("no_mangle", DECL_ATTRIBUTES (fndecl)) == NULL_TREE;
 }
 
 void
-HIRCompileBase::setup_attributes_on_fndecl (
-  tree fndecl, bool is_main_entry_point, HIR::Visibility &visibility,
-  const HIR::FunctionQualifiers &qualifiers, const AST::AttrVec &attrs)
+HIRCompileBase::setup_fndecl (tree fndecl, bool is_main_entry_point,
+			      HIR::Visibility &visibility,
+			      const HIR::FunctionQualifiers &qualifiers,
+			      const AST::AttrVec &attrs)
 {
   // if its the main fn or pub visibility mark its as DECL_PUBLIC
   // please see https://github.com/Rust-GCC/gccrs/pull/137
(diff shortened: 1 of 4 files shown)


gcc/rust/backend/rust-compile-base.cc
```diff
@@ -342,7 +366,10 @@ HIRCompileBase::compile_function (
   // we don't mangle the main fn since we haven't implemented the main shim
   bool is_main_fn = fn_name.compare ("main") == 0;
   std::string asm_name = fn_name;
-  if (!is_main_fn)
+  // TODO(liushuyu): we should probably move this part to
+  // `setup_attributes_on_fndecl` if possible
+  bool should_mangle = should_mangle_item (outer_attrs);
+  if (!is_main_fn && should_mangle)
```
philberty, line 372: I think your right, we could change this by renaming "setup_attributes_on_fndecl" to "setup_fndecl" which implies a more generic setup function.

Then you can change this to remove the asm_name it will default to the ir_symbol_name at this point anyway.
```
  tree fndecl = ctx->get_backend ()->function (compiled_fn_type, ir_symbol_name,
					       asm_name, flags, locus);
```

```
  tree fndecl = ctx->get_backend ()->function (compiled_fn_type, ir_symbol_name,
					       "" /* asm_name */, flags, locus);
```

Then you pull this highlighted code into where you handle no_mangle. You might need to have a flag to check for no_mangle by iterating the attributes then depending on the flag mangle or no mangle at the end.

philberty (changes requested):
You already spotted the mangle code piece so lets fix that as part of this PR. Otherwise great job again.

gcc/rust/backend/rust-compile-base.cc
liushuyu: > I think your right, we could change this by renaming "setup_attributes_on_fndecl" to "setup_fndecl" which implies a more generic setup function.
> 
> Then you can change this to remove the asm_name it will default to the ir_symbol_name at this point anyway.
> 
> ```
>   tree fndecl = ctx->get_backend ()->function (compiled_fn_type, ir_symbol_name,
> 					       asm_name, flags, locus);
> ```
> 
> ```
>   tree fndecl = ctx->get_backend ()->function (compiled_fn_type, ir_symbol_name,
> 					       "" /* asm_name */, flags, locus);
> ```
> 
> Then you pull this highlighted code into where you handle no_mangle. You might need to have a flag to check for no_mangle by iterating the attributes then depending on the flag mangle or no mangle at the end.

I see. Thanks for the suggestion! I will attempt to implement the changes.

liushuyu: Done. I have merged the logic into the renamed `setup_fndecl` function.

gcc/rust/backend/rust-compile-base.cc
```diff
@@ -26,14 +26,21 @@
...
 #include "fold-const.h"
 #include "stringpool.h"
+#include "attribs.h"
 
 namespace Rust {
 namespace Compile {
 
+bool inline should_mangle_item (const tree fndecl)
+{
+  return lookup_attribute ("no_mangle", DECL_ATTRIBUTES (fndecl)) == NULL_TREE;
```
philberty, line 36: I really like how you did this to split up the logic.

I guess the GCC idiom would be that this should be a macro called SHOULD_MANGLE_ITEM_P and put it over in rust-tree.h not sure what design I want in the code-generation part at the moment so lets stick with this for now.

philberty (approved):
LGTM

gcc/rust/backend/rust-compile-base.cc
liushuyu: > I really like how you did this to split up the logic.
> 
> I guess the GCC idiom would be that this should be a macro called SHOULD_MANGLE_ITEM_P and put it over in rust-tree.h not sure what design I want in the code-generation part at the moment so lets stick with this for now.

Noted. I will keep this in mind in the near future.

PR #1151, opened 2022-04-22
backend: handle deprecated attribute

liushuyu:
- handle the `deprecated` attribute

### DIFF
--- a/gcc/rust/backend/rust-compile-base.cc
+++ b/gcc/rust/backend/rust-compile-base.cc
@@ -209,8 +209,13 @@ HIRCompileBase::handle_deprecated_attribute_on_fndecl (
 	}
     }
 
-  DECL_ATTRIBUTES (fndecl) = tree_cons (get_identifier ("deprecated"), value,
-					DECL_ATTRIBUTES (fndecl));
+  if (value)
+    {
+      tree attr_list = build_tree_list (NULL_TREE, value);
+      DECL_ATTRIBUTES (fndecl)
+	= tree_cons (get_identifier ("deprecated"), attr_list,
+		     DECL_ATTRIBUTES (fndecl));
+    }
 }
 
 void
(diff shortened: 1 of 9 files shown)


CohenArthur (commented):
Great stuff. If you look at the [RFC for the deprecated attribute](https://github.com/rust-lang/rfcs/blob/master/text/1270-deprecation.md), you'll see that it is also supposed to be used on types. I think it's fine to leave and merge this PR, but I'd like to see an issue or comment mentionning that this should also handle types :)

gcc/rust/backend/rust-compile-base.cc
```diff
@@ -379,6 +450,8 @@ HIRCompileBase::compile_function_body (Context *ctx, tree fndecl,
       auto compiled_expr = CompileStmt::Compile (s.get (), ctx);
       if (compiled_expr != nullptr)
 	{
+	  if (TREE_CODE (compiled_expr) == CALL_EXPR)
```
philberty, line 453: This won't handle all of the cases to emit the diagnostic such as the tail expression in a block or when there are call expressions as arguments.

There are two methods to make this diagnostic more abstract. We could either implement a new lint along side the rust/lint/rust-lint-deprecated and copy the walk code from the rust-lint-unused-var.cc which reuses the GCC walk_tree code so you can simply have a function like:

```c
static tree
deprecated_call_walk_fn (tree *t, int *walk_subtrees, void *closure)
{
  switch (TREE_CODE (*t))
    {
    case CALL_EXPR:
      maybe_warn_deprecated (t);
      break;

    default:
      break;
    }
  return NULL_TREE;
}
```

The benefit here is we can probably include checks for depreciated types like this too and walk the types used within the tree too. The problem is it will miss the case for constexpr code which will be optimized out but we could handle that directly.

The other option is that for every time we generate a call expression we add this maybe_warn_deprecated into: rust-gcc.cc:call_expression line 1797.

Its probably best to add the check into that function for now

philberty (changes requested):

PR #1155, opened 2022-04-22
Add base for visibility resolving

CohenArthur:
This adds a new visitor whose aim is to resolve the visibility of each VisItem in a crate. The crux of the implementation will be resolving `pub restricted` items, whose module needs to be resolved using name resolution/path resolution. On top of that, we will need to add a new visitor building a "tree" of modules, with a `accessible_from` or `is_child_of` method: All children of a module get to access its items whose visibility is inherited.

### DIFF
--- a/gcc/rust/privacy/rust-visibility-resolver.cc
+++ b/gcc/rust/privacy/rust-visibility-resolver.cc
@@ -17,6 +17,7 @@
 // <http://www.gnu.org/licenses/>.
 
 #include "rust-visibility-resolver.h"
+#include "rust-ast.h"
 #include "rust-hir.h"
 #include "rust-hir-item.h"
 
@@ -55,6 +56,7 @@ VisibilityResolver::resolve_visibility (const HIR::Visibility &visibility,
       return true;
     case HIR::Visibility::PUBLIC:
       // FIXME: We need to handle the restricted path here
+      // FIXME: We also need to handle 2015 vs 2018 edition conflicts
       to_resolve = ModuleVisibility::create_public ();
       return true;
     default:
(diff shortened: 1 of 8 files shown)


gcc/rust/privacy/rust-visibility-resolver.cc
```diff
@@ -55,6 +56,7 @@ VisibilityResolver::resolve_visibility (const HIR::Visibility &visibility,
       return true;
     case HIR::Visibility::PUBLIC:
       // FIXME: We need to handle the restricted path here
+      // FIXME: We also need to handle 2015 vs 2018 edition conflicts
```
philberty, line 59: I wouldn't worry about that untill the end of the milstone and we can think about it when we have something like libcore working and need to make decisions about versions.

gcc/rust/privacy/rust-visibility-resolver.cc
```diff
@@ -0,0 +1,164 @@
...
+      to_resolve = ModuleVisibility::create_restricted (peek_module ());
+      return true;
+    case HIR::Visibility::PUBLIC:
+      // FIXME: We need to handle the restricted path here
+      // FIXME: We also need to handle 2015 vs 2018 edition conflicts
+      to_resolve = ModuleVisibility::create_public ();
+      return true;
+    case HIR::Visibility::ERROR:
+      return false;
+    }
```
philberty, line 64: I think C++ wants a default: or a default return of false here to avoid missing return's since we don't have block expressions in c++.

gcc/rust/privacy/rust-visibility-resolver.cc
```diff
@@ -0,0 +1,164 @@
...
+}
+
+DefId
+VisibilityResolver::peek_module ()
+{
+  // We're always inserting a top module - the crate
+  // But we have to check otherwise `.back()` is UB
+  if (module_stack.empty ())
```
philberty, line 72: I would write this as rust_assert(!module_stack.empty());

gcc/rust/privacy/rust-visibility-resolver.cc
CohenArthur: Oops I fixed this but must have lost it in a rebase

CohenArthur: Ah, I had just forgotten to force push :D Fixed

gcc/rust/privacy/rust-visibility-resolver.cc
CohenArthur: Done in c3a8113

PR #1157, opened 2022-04-22
Resolve remaining simple visibilites

CohenArthur:
Based on #1155, so you can review just the last commit.

This adds the remaining simple visitors for visibility resolving, as well as a helper method for doing so.

### DIFF
--- a/gcc/rust/hir/tree/rust-hir-item.h
+++ b/gcc/rust/hir/tree/rust-hir-item.h
@@ -633,6 +633,7 @@ public:
   virtual void accept_vis (HIRVisItemVisitor &vis) = 0;
 
   Visibility &get_visibility () { return visibility; }
+  const Visibility &get_visibility () const { return visibility; }
 
   std::string as_string () const override;
 };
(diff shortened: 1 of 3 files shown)


gcc/rust/privacy/rust-privacy-common.h
```diff
@@ -0,0 +1,64 @@
...
+  {
+    return ModuleVisibility (Type::Restricted, module_id);
+  }
+
+  static ModuleVisibility create_public ()
+  {
+    return ModuleVisibility (Type::Public, UNKNOWN_DEFID);
+  }
+
+  Type get_kind () { return kind; }
```
philberty, line 51: this can be get_kind() const

gcc/rust/privacy/rust-privacy-common.h
```diff
@@ -0,0 +1,64 @@
...
+
+  DefId &get_module_id () { return module_id; }
```
philberty, line 53: I think if your returning a reference to DefId it should be const DefId& get_module_id() const { ... }

philberty (approved):
LGTM its a really good starting base, just very minor nit about const but its not imporant

Issue #1141, opened 2022-04-20
Macro invocations are not expanded in the `if` expressions


antego:
Related to the #1119 

Looks like the macro expansion works incorrectly in the comparison expressions and in the `if` expressions

I tried this code:

```rust
  let res = "test1" == "test1";
  if res {
    print(0);
  }
  let res = (concat!("test2") == "test3");
  if res {
    print(1);
  }
  if concat!("test2") == "test3" {
    print(2);
  }
```

I expected the code to compile successfully. Instead, the compilation failed with the error:
```
fatal error: Failed to lower expr: [MacroInvocation:
 outer attributes: none
 concat!(("test2"))
 has semicolon: false]
compilation terminated.
```

### Meta

From the zulip chat:

```
@antego:
Some conclusions from the ast above:

macros are expanded outside of the if expressions, but they remove all other parts of the expression. concat!("test2") == "test3" became just test2
macros are not expanded inside the if expressions, leading to the error

Questions:

Are the above conclusions correct?
Any hints about where to look next? I was planning to go through the code and find the place where the macro expansion is called and check how the expanded macro is handled for a simple expression vs if expression.

@CohenArthur:
Macro expansion is done in our attribute visitor, which you can find the implementation of in gcc/rust/expand/[rust-attribute-visitor.cc](http://rust-attribute-visitor.cc/). The code is a little bit hard to follow as it does a lof of in-place replacements of the expanded AST fragments (btw, we could probably think about splitting the attribute visitor and in-place expansion phase. Even though it might be a slowdown?).

The reason it crashes and you get the "Couldn't lower expression" error is that the macro fragment simply does not get used in the proper place. You can look at the visitor for AST::LetExpr values and see how you are supposed to expand an expression in-place.

I've checked the code, and it hasn't been implemented for all AST::If{Expr, LetExpr...} nodes yet. You can probably do the same as a LetExpr and add tests to see your progress gradually.

So to answer your questions precisely:

a/ Yes and no! These are probably side-effects of the code just not existing yet. Sorry about that!
b/ Yes! That's completely right

The rust-attribute-visitor.cc file is a good place to look at, and specifically LetExpr or ArithmeticExpr which do expression expansion :)
```

liushuyu:
Probably related: #1084 and #1048

CohenArthur:
Fixed by #1170

PR #1161, opened 2022-04-23
Implement macro expansion for ComparisonExpr, LazyBooleanExpr, AssignmentExpr

antego:
Following up on https://github.com/Rust-GCC/gccrs/issues/1141

Currently the macro expansion doesn't work for ComparisonExpr, LazyBooleanExpr, AssignmentExpr. To fix this, I just copied the code from the `ArithmeticOrLogicalExpr` and it seemed to work. I don't like the code duplication, happy to try refactoring it into a separate function.

Will work on the macro expansion in the `if` expressions next.

### DIFF
--- a/gcc/rust/expand/rust-attribute-visitor.cc
+++ b/gcc/rust/expand/rust-attribute-visitor.cc
@@ -643,10 +643,19 @@ AttrVisitor::visit (AST::ComparisonExpr &expr)
 
   /* should have no possibility for outer attrs as would be parsed
    * with outer expr */
-  expr.get_left_expr ()->accept_vis (*this);
+  auto &l_expr = expr.get_left_expr ();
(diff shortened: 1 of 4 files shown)


CohenArthur (approved):
Yes sadly this visitor does contain a lot of code repetition (some others in the code too). It's a bit hard to refactor this, I had tried but didn't spend much time on it. Maybe we could do something with a templated function, something like
```c++
template <typename T>
void maybe_expand_expr(T &expr) {
    auto fragment = expander.take_expanded_fragment(*this);
    if (fragment.should_expand)
        expr = fragment.take_expression_fragment();
}
```
You can do that in another pull request if you're interested :)

antego:
> Maybe we could do something with a templated function, something like

Thanks, I'll play with it

PR #1162, opened 2022-04-23
CI: catch malformed test cases

liushuyu:
- catch malformed test cases and use GitHub Actions' annotation feature to mark the failures

### DIFF
--- a/.github/emit_test_errors.pl
+++ b/.github/emit_test_errors.pl
@@ -1,0 +1,25 @@
+#!/bin/perl -n
+
+sub analyze_errors() {
+    /^(FAIL|ERROR|XPASS):\s([^:\s]+):?\s+(.+)/;
+
+    my $type     = $1;
+    my $filename = $2;
+    my $message  = $3;
+    my $line;
+
+    if ( !$type ) { return; }
+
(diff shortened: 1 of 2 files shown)


liushuyu:
Demo:

Screenshot

liushuyu:
> Does this script output some magic output to stdout that GitHub picks this up to annotate the files?

Yes. The script produces GitHub Actions Annotation Commands to the `stdout`. You can see the documentation for this "magic format" here: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions.

.github/emit_test_errors.pl
```diff
@@ -0,0 +1,25 @@
...
+
+    my $command = "::error file=gcc/testsuite/$filename";
+    if ($line) {
+        $command = "$command,line=$line";
+    }
+
+    print "$command,title=Test failure ($type)::$message\n";
+}
+
+analyze_errors();
```
CohenArthur, line 25: I honestly have no idea on how correct this script is and I personally cannot maintain it at all if something goes wrong at some point :DD But I'm fine with it if Philip is

CohenArthur (approved):

Issue #1158, opened 2022-04-22
Support SimplePath resolution in visibilities


philberty:
Visibilities support AST::SimplePaths such as:

```
pub(in crate::mod::child::grandchild)
```

We need to as part of name-resolution resolve these paths so that we can use this information later on in privacy checking.

- [ ] Add helper to AST::Visibility has_path() return ! path.is_error()
- [ ] Add new helper to Resolver::ResolverBase to resolve visibility which takes the reference to the AST::Visibility
- [ ] In the visibility resolver we need to resolve the path if the visibility has a path

### Assumption

Let s assume for now the Simple Path does not support generics which will simplify this a lot.

### Resolving Simple Path

We need to build a CanonicalPath for this we should be able to reuse the existing code in https://github.com/Rust-GCC/gccrs/blob/4337aea3be387921ee60520d2188906d8762367b/gcc/rust/resolve/rust-ast-resolve-expr.h#L28

Lets add a new:

```
static void go (AST::SimplePath *expr, NodeId parent);
```

Before you can build the canonical path you need a new helper to resolve a SimplePathSegment to a CanonicalPath segment

https://github.com/Rust-GCC/gccrs/blob/4337aea3be387921ee60520d2188906d8762367b/gcc/rust/resolve/rust-ast-resolve-type.h#L154

You need a new helper for SimplePathSegment which will be the same as the PathExprSegment except it can ignore generics and it just needs the string of the segment and its relevant node_id.

Now that you have this in place you can build the full path by creating a new resolve_simple_path_segments by reusing the code from:

https://github.com/Rust-GCC/gccrs/blob/4337aea3be387921ee60520d2188906d8762367b/gcc/rust/resolve/rust-ast-resolve.cc#L567

The only difference really is the different types but a SimplePath should be fully resolved at name-resolution time so the piece in the resolve_segments for PathExprSegments should not allow for that empty else case but be an error condition.

PR #1164, opened 2022-04-25
Add `NodeId`s to `SimplePath{Segment}`s

CohenArthur:
This is a necessary first step which addresses #1158

I am working on the visibility resolver and simple-path name resolver in parallel and will need those changes. 

I'm not sure if the `has_path` method is necessary since we will have `HIR::Visibilities` in the visibility resolver, which will be desugared. Maybe I am doing this wrong and we should be working on `AST::Visibilities` instead?

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -325,12 +325,14 @@ class SimplePathSegment : public PathSegment
 {
   std::string segment_name;
   Location locus;
+  NodeId node_id;
 
   // only allow identifiers, "super", "self", "crate", or "$crate"
 public:
   // TODO: put checks in constructor to enforce this rule?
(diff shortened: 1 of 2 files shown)


philberty (approved):
LGTM but something to be really careful of is unlike PathInExpression which is allocated and managed in the std::unique_ptr we can get away with no copy constructor. But in this case you might need to add in copy and copy assignment constructors to ensure the NodeId is preserved or you might end up in the case where the node id is lost. This will only occur if the object containing the SimplePath relies on copy. Otherwise you might be grand but keep this in mind if you run into issues or just add copy constructors for peace of mind.

I only bring this up because I had a nightmare last year trying to figure out what happened my NodeIds in something and it turned out to be a missing copy constructor lol.

CohenArthur:
That sounds absolutely awful haha, thanks for the heads up!

PR #1166, opened 2022-04-25
Support inference of generic parameters on paths behind reference's

### LINKED ISSUE #1165: Identifier expressions are not inferring generic arguments
I tried this code:

```rust
struct Foo<T>(T);

fn main()  {
    let a = &Foo(123);
}
```
see: https://godbolt.org/z/7drh355Mc

I expected to see this happen: That the compiler infers T is i32

Instead, this happened: The generic structure is not infered and returns error

### Meta

- What version of Rust GCC were you using, git sha if possible. 1286acc34e826039ebc8f09ad36dddc30f726283

philberty:
We used a hack in the parser to turn simple cases such as &Foo(..) into:

  BorrowExpr
    CallExpr(
      IdentifierExpr + <Argument-expressions>))

The IdentifierExpr here is parsed as a PathExpression but to simplify
things at the time it seemed logic to see these as identifier expressions
but this is actually a Path and we need to be careful about generic
arguments here. Identifiers are simply identifiers and must not be changed
or coherence of inference variables will become a jumble of inference
variables trying to infer one another inside purely generic code.

This patch leaves the PathInExpressions as Path's instead of trying to be
clever and turn them into identifiers.

Fixes #1165

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -12591,16 +12591,6 @@ Parser<ManagedTokenSource>::null_denotation (const_TokenPtr tok,
 	      // struct/enum expr struct
 	      if (!restrictions.can_be_struct_expr && !not_a_block)
 		{
-		  // assume path is returned if not single segment
-		  if (path.is_single_segment ())
-		    {
-		      // have to return an identifier expression or something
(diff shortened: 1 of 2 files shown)


CohenArthur (approved):
Oh, that feels good, removing a hack :DD Great!

PR #1170, opened 2022-04-26
Implement macro expansion in `IfExpr`, `IfExprConseqElse`, `IfExprConseqIf`, `IfExprConseqIfLet`.

antego:
Addresses #1141. Following up on #1161.

This change adds support for the macros inside the `if` condition expressions.

Things to note:
1. Judging by my research, the `IfExprConseqIfLet` isn't used. The parser treats the syntax `let var = if ...` as a `let` expression followed by `if` expression:
Screen Shot 2022-04-26 at 9 47 29 am
2. I didn't add the macro expansion to the `IfLetExpr...` family of expressions because I wasn't able to write a test case reproducing the missing macro expansion. I've tried the following code
```rust
fn main() -> i32 {
    let mut res = 0;

    enum E {
        X(u8),
        Y(u8),
        Z(u8),
    }
    let v = E::Y(12);
    if let E::Y(n) = v {
        res = 1;
    }

    0
}
```
which caused the compiler error
```
FAIL: rust/compile/macro43.rs (internal compiler error: in append_reference_for_def, at rust/resolve/rust-name-resolver.h:227)
```

### DIFF
--- a/gcc/rust/expand/rust-attribute-visitor.cc
+++ b/gcc/rust/expand/rust-attribute-visitor.cc
@@ -1630,6 +1630,9 @@ AttrVisitor::visit (AST::IfExpr &expr)
   // can't strip condition expr itself, but can strip sub-expressions
   auto &condition_expr = expr.get_condition_expr ();
   condition_expr->accept_vis (*this);
+  auto cond_fragment = expander.take_expanded_fragment (*this);
+  if (cond_fragment.should_expand ())
+    condition_expr = cond_fragment.take_expression_fragment ();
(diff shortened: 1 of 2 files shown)


philberty (approved):
LGTM

CohenArthur (approved):
Great stuff :DD Thank you for working on this!

CohenArthur:
@antego I'd say open issues for your two notes: The first one is something that we can maybe discuss about removing (or it may be caused by a parser bug), and the second one needs to be fixed eventually. It would be really helpful :)

bors r+

PR #1172, opened 2022-04-26
Resolve visibility paths 

### LINKED ISSUE #1158: Support SimplePath resolution in visibilities

CohenArthur:
This tackles the remaining issues in #1158. ~~It also contains the commit from #1168 so you should wait for bors to merge that before reviewing it :D~~ done 

The paths get resolved but their `NodeId`s are not used in the privacy visibility resolver yet.

Closes #1158

### DIFF
--- a/gcc/rust/resolve/rust-ast-resolve-base.cc
+++ b/gcc/rust/resolve/rust-ast-resolve-base.cc
@@ -17,15 +17,27 @@
 // <http://www.gnu.org/licenses/>.
 
 #include "rust-ast-resolve-base.h"
+#include "rust-ast-resolve-expr.h"
+#include "rust-ast-resolve-path.h"
+#include "rust-item.h"
 
 namespace Rust {
 namespace Resolver {
 
 bool
-resolve_visibility (const AST::Visibility &vis)
+ResolverBase::resolve_visibility (const AST::Visibility &vis)
 {
-  gcc_unreachable ();
-  return false;
+  if (vis.has_path ())
+    {
+      auto path = vis.get_path ();
+      ResolvePath::go (&path, parent);
+
+      // Do we need to lookup something here?
+      // Is it just about resolving the names correctly so we can look them up
+      // later?
+    }
+
+  return true;
 }
 
 // Default visitors implementations
--- a/gcc/rust/ast/rust-item.h
+++ b/gcc/rust/ast/rust-item.h
@@ -701,6 +701,7 @@ public:
 
   std::string as_string () const;
   const SimplePath &get_path () const { return in_path; }
+  SimplePath &get_path () { return in_path; }
 
 protected:
   // Clone function implementation - not currently virtual but may be if
--- a/gcc/rust/resolve/rust-ast-resolve-item.cc
+++ b/gcc/rust/resolve/rust-ast-resolve-item.cc
@@ -238,6 +238,8 @@ ResolveItem::visit (AST::Module &module)
   mappings->insert_canonical_path (mappings->get_current_crate (),
 				   module.get_node_id (), cpath);
 
+  resolve_visibility (module.get_visibility ());
+
   NodeId scope_node_id = module.get_node_id ();
   resolver->get_name_scope ().push (scope_node_id);
   resolver->get_type_scope ().push (scope_node_id);
@@ -267,6 +269,8 @@ ResolveItem::visit (AST::TupleStruct &struct_decl)
   mappings->insert_canonical_path (mappings->get_current_crate (),
 				   struct_decl.get_node_id (), cpath);
 
+  resolve_visibility (struct_decl.get_visibility ());
+
   NodeId scope_node_id = struct_decl.get_node_id ();
   resolver->get_type_scope ().push (scope_node_id);
 
@@ -286,6 +290,8 @@ ResolveItem::visit (AST::TupleStruct &struct_decl)
       if (field.get_field_type ()->is_marked_for_strip ())
 	continue;
 
+      resolve_visibility (field.get_visibility ());
+
       ResolveType::go (field.get_field_type ().get (),
 		       struct_decl.get_node_id ());
     }
@@ -302,6 +308,8 @@ ResolveItem::visit (AST::Enum &enum_decl)
   auto cpath = canonical_prefix.append (decl);
   mappings->insert_canonical_path (mappings->get_current_crate (),
 				   enum_decl.get_node_id (), cpath);
+
+  resolve_visibility (enum_decl.get_visibility ());
 
   NodeId scope_node_id = enum_decl.get_node_id ();
   resolver->get_type_scope ().push (scope_node_id);
@@ -328,6 +336,9 @@ ResolveItem::visit (AST::Enum &enum_decl)
 void
 ResolveItem::visit (AST::EnumItem &item)
 {
+  // Since at this point we cannot have visibilities on enum items anymore, we
+  // can skip handling them
+
   auto decl
     = CanonicalPath::new_seg (item.get_node_id (), item.get_identifier ());
   auto path = prefix.append (decl);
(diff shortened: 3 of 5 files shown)


gcc/rust/resolve/rust-ast-resolve-base.cc
```diff
@@ -0,0 +1,653 @@
...
+      // Is it just about resolving the names correctly so we can look them up
+      // later?
+    }
+
+  return true;
+}
+
+// Default visitors implementations
+
+void
```
philberty, line 44: Eventually, we want to do the same refactoring @dkm did for HIR over here. Since it will clean it up alot.

gcc/rust/privacy/rust-visibility-resolver.cc
```diff
@@ -45,6 +45,17 @@ VisibilityResolver::go (HIR::Crate &crate)
...
 
+// FIXME: At this point in the pipeline, we should not be dealing with
+// `AST::SimplePath`s anymore! We need to be dealing with their "resolved
+// counterpart", so probably a NodeId/HirId/DefId.
+
+// static bool
+// resolve_module_path (std::vector<HIR::Module> &module_stack,
+// 		     const AST::SimplePath &restriction, DefId &id)
+// {
+//   return false;
```
philberty, line 56: yep there will be an HIR simple path which we might desugar but this should insert into the name/type namespace what the resolved NodeId is then we just use this in the rest of the pipeline.

philberty: Not sure this change needs to be in this patch though.

gcc/rust/resolve/rust-ast-resolve-expr.h
```diff
@@ -42,16 +42,39 @@ class ResolvePath : public ResolverBase
...
+				NodeId expr_node_id, Location expr_locus);
+};
+
+class ResolveSimplePathSegmentToCanonicalPath
+{
+public:
+  static CanonicalPath resolve (const AST::SimplePathSegment &seg)
+  {
+    // FIXME: Since this is so simple, maybe it can simply be a tiny function?
+    return CanonicalPath::new_seg (seg.get_node_id (), seg.get_segment_name ());
```
philberty, line 76: Yeah that sounds good. There are a bunch like this that could be simple static functions.

gcc/rust/resolve/rust-ast-resolve.cc
```diff
@@ -195,6 +195,92 @@ ResolvePath::resolve_path (AST::PathInExpression *expr)
...
+}
+
+void
+ResolvePath::resolve_simple_path_segments (
+  CanonicalPath prefix, size_t offs,
+  const std::vector<AST::SimplePathSegment> &segs, NodeId expr_node_id,
+  Location expr_locus)
+{
+  /**
+   * TODO: We also need to handle types and functions here
```
philberty, line 257: functions end up in the name scope so this is already handled but you could add in the lookup for the type-scope like we talked about

gcc/rust/resolve/rust-ast-resolve.cc
```diff
@@ -195,6 +195,92 @@ ResolvePath::resolve_path (AST::PathInExpression *expr)
...
+
+      if (resolver->get_name_scope ().lookup (path, &resolved_node))
+	{
+	  resolver->insert_resolved_name (seg.get_node_id (), resolved_node);
+	  resolver->insert_new_definition (seg.get_node_id (),
+					   Definition{expr_node_id, parent});
+	}
+    }
+
+  if (resolved_node != UNKNOWN_NODEID)
```
philberty, line 276: this is an error condition for SimplePaths so if this is UNKNOWN_NODEID then we should emit an error

gcc/rust/resolve/rust-ast-resolve.cc
```diff
@@ -195,6 +195,92 @@ ResolvePath::resolve_path (AST::PathInExpression *expr)
 		    expr->get_node_id (), expr->get_locus ());
 }
 
+void
+ResolvePath::resolve_path (AST::SimplePath *expr)
+{
+  // resolve root segment first then apply segments in turn
+  auto &segs = expr->get_segments ();
```
philberty, line 202: It might be possible for SimplePaths to just do one big for loop instead of splitting this across two functions.

gcc/rust/resolve/rust-ast-resolve-base.cc
```diff
@@ -0,0 +1,653 @@
...
+bool
+ResolverBase::resolve_visibility (const AST::Visibility &vis)
+{
+  if (vis.has_path ())
+    {
+      auto path = vis.get_path ();
+      ResolvePath::go (&path, parent);
+
+      // Do we need to lookup something here?
+      // Is it just about resolving the names correctly so we can look them up
```
philberty, line 35: Yep thats all you need to do here.

philberty (changes requested):
This looks like 99% there, just some minor nits.

You could add in the lookup type scope if you want to. And I think the SimplePath resolution could possibly be simpler with one big for loop:

```c
CanonicalPath simple_path_buffer = empty;
for (seg : path.segments()) {
  canonical_path s = canonical_path_from_simple_Segment;
  simple_path_buffer = simple_path_buffer.append(s);
  
  // do lookup of name scope
  // do lookup of type scope
  
  insert_resolved_node_ids -> segment_node_id
}

If the final segment node_id is fully resolved insert the mapping of the Simple path -> that node id so you don't have to lookup the final segment nodeid and can just use the SimplePath
```

gcc/rust/resolve/rust-ast-resolve-base.cc
CohenArthur: Opened #1160

gcc/rust/privacy/rust-visibility-resolver.cc
CohenArthur: Yeah this part is still very early and I think it should be changed so that the `HIR::Visibility` does not keep a `SimplePath` but simply its associated `DefId` as you said. Do you want me to remove the comment? It was simply for later on to keep track of stuff I need to do in the Privacy visibility resolver

gcc/rust/resolve/rust-ast-resolve-expr.h
CohenArthur: I like the API of having a class do that, but I was unsure of where to put it haha. I'm fine with keeping it as a class especially if it gets more complex when resolving types etc

philberty: Yeah i would be in favor of grouping them together or turning them into simple functions in another PR

CohenArthur:
I'm gonna push the cleaned up version of this, but to have proper tests we need to fix some location info on SimplePaths. I'll tackle that before merging this

philberty (approved):
I think this looks good now!

Issue #1159, opened 2022-04-22
Resolve SimplePaths to DefId's


philberty:
This change depends on #1158 

- [x] Add missing mappings to HIR::SimplePath #1176 
- [x] Add missing mappings to HIR::SimpePathSegment #1176
- [x] Remove the default locus argument of builtin_location on HIR::SimplePathSegment #1174 

Resolving a SimplePath to a DefId its achieved via the name-resolver class:

```
   Parameter: SimplePath* expr;

    NodeId ast_node_id = expr.get_mappings ().get_nodeid ();

    // then lookup the reference_node_id
    NodeId ref_node_id = UNKNOWN_NODEID;
    if (resolver->lookup_resolved_name (ast_node_id, &ref_node_id))
      {
	// these ref_node_ids will resolve to a pattern declaration but we are
	// interested in the definition that this refers to get the parent id
	Definition def;
	if (!resolver->lookup_definition (ref_node_id, &def))
	  {
	    // FIXME
	    // this is an internal error
	    rust_error_at (expr.get_locus (),
			   "unknown reference for resolved name");
	    return;
	  }
	ref_node_id = def.parent;
      }
```

if we are assuming this path can or should only resolve to modules we can ignore the type resolver name-space.

```
    if (ref_node_id == UNKNOWN_NODEID)
      {
        // this is an internal error but error messages are useful for now
        // down the line we will be able to rely on the changes make in #1158 
        // that this will have already errored out for us and we can just return
	rust_error_at (expr.get_locus (), "path: %s",
		       expr.as_string ().c_str ());
	return;
      }
```

Now we have resolved the expression to its definitions NodeId which is all within the AST. 

### Getting the DefId from NodeId

Note there is a bug with our mappings class here but we can fix this either as part of this issue or in a separate issue/pr we need to reverse map the NodeId back to an Hir node so that we can look up the Hir node to the DefID but the mappings class as nested these reverse mappings with the associated CrateNum which is wrong but we can ignore this for now.

```
    // node back to HIR
    HirId ref;
    if (!mappings->lookup_node_to_hir (expr.get_mappings ().get_crate_num (),
				       ref_node_id, &ref))
      {
	// this is an internal error like above but its useful for debugging for now
	rust_error_at (expr.get_locus (), "reverse lookup failure");
	return;
      }
```

Now that we have the associated HirId of the resolved path we can look it up again in the mappings and again this wrongly associates the crate num here but ignore this for now.

```
[...]

PR #1174, opened 2022-04-26
Remove default location for `SimplePath`s

CohenArthur:
This removes the default argument for the locus of `SimplePath`s. Addresses #1159 

There are a lot of places in the code where it is not possible to insert a "correct" location for these `SimplePath`s: namely, when creating the use statement for the standard library's prelude, or when creating the `SimplePath`s of attributes. I've left a few FIXMEs in these cases because I'm unsure about the proper course of action.

### DIFF
--- a/gcc/rust/ast/rust-ast-full-test.cc
+++ b/gcc/rust/ast/rust-ast-full-test.cc
@@ -4695,7 +4695,10 @@ MetaNameValueStr::to_attribute () const
 {
   LiteralExpr lit_expr (str, Literal::LitType::STRING,
 			PrimitiveCoreType::CORETYPE_UNKNOWN, {}, Location ());
-  return Attribute (SimplePath::from_str (ident),
+  // FIXME: What location do we put here? Is the literal above supposed to have
+  // an empty location as well?
+  // Should MetaNameValueStr keep a location?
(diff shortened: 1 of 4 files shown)


gcc/rust/ast/rust-ast-full-test.cc
```diff
@@ -4695,7 +4695,10 @@ MetaNameValueStr::to_attribute () const
 {
   LiteralExpr lit_expr (str, Literal::LitType::STRING,
 			PrimitiveCoreType::CORETYPE_UNKNOWN, {}, Location ());
-  return Attribute (SimplePath::from_str (ident),
+  // FIXME: What location do we put here? Is the literal above supposed to have
+  // an empty location as well?
+  // Should MetaNameValueStr keep a location?
+  return Attribute (SimplePath::from_str (ident, Location ()),
```
philberty, line 4701: Yeah this is because MetaNameValueStr is missing location info, there are quite a few places we have missing location info. Raise issues for each of these but don't fix them as part of this PR.

Just preserve this FIXME comment and over time raise issues for each of them, tagged as good-first-pr.

philberty (approved):
LGTM mark each of those places missing location info as FIXME's and raise issues much of the AST is still missing location info. We should only use empty location when its an error node or empty node.

PR #1176, opened 2022-04-26
Keep `SimplePath{Segment}`s in mappings

CohenArthur:
Addresses #1159


philberty (approved):
LGTM our usage of crate-num etc will get fixed in a few weeks it might affect you when getting the DefId in the 2nd issue.

PR #1184, opened 2022-04-26
Fix bad name resolution on path with generic segments

### LINKED ISSUE #1173: Bad name resolution on path with generic segments
I tried this code:

```rust
mod mem {
    extern "rust-intrinsic" {
        fn transmute<U, V>(_: U) -> V;
    }
}

pub trait Hasher {
    fn write_u16(&mut self, i: u16) {
        self.write(&mem::transmute::<_, [u8; 2]>(i))
    }
}

pub struct SipHasher;

impl Hasher for SipHasher {
    #[inline]
    fn write(&mut self, msg: &[u8]) {}
}

```

I expected to see this happen: compile without failure

Instead, this happened: ```failed to resolve mem:transmute<_, [u8 2]>```

### Meta

- What version of Rust GCC were you using, git sha if possible. ab845217887468b500b624f0837023122002c534

philberty:
This fix breaks down into two commits the first commit fixes the issue directly
then the second one fixes the regression that was introduced in the first.

The TLDR for this patch set is that when we resolve the path here:

  module::transmute

The name-resolver wronly added the specified generic arguments into the canonical
path meaning that the root segment could not be name-resolved such that during
type resolution we iterate and apply the generic arguments appropriately to the root
segment.

In fixing this it introduced a regression exposing a few more complex issues with
inference variables but for a more complete explanation about that please read the
commit message from ea38a59ee8329bb7b309400a7de57d53ad7bde31

Fixes #1173

### DIFF
--- a/gcc/rust/hir/tree/rust-hir-path.h
+++ b/gcc/rust/hir/tree/rust-hir-path.h
@@ -262,6 +262,8 @@ public:
 
   std::vector<PathExprSegment> &get_segments () { return segments; }
 
+  const std::vector<PathExprSegment> &get_segments () const { return segments; }
+
   PathExprSegment &get_root_seg () { return segments.at (0); }
 
   PathExprSegment get_final_segment () const { return segments.back (); }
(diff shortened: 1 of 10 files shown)


CohenArthur (approved):
LGTM

Issue #915, opened 2022-02-09
support repr layout options of aligned and packed on structures


philberty:
https://doc.rust-lang.org/reference/type-layout.html#representations

PR #1188, opened 2022-04-27
Support align and packed repr layout on structs

dafaust:
This is a start at handling the various layout options supported by Rust, beginning with `#[repr(align(N))]` and `#[repr(packed(N))]`, on structs and tuple structs.

There are several other layout options which remain to be supported such as `#[repr(C)]`, `#[repr(transparent)]`, combinations e.g. `#[repr(C, packed(2))]`, as well as layouts on union and enum types.

Fixes: #915

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-base.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-base.cc
@@ -260,5 +260,72 @@ TypeCheckBase::resolve_literal (const Analysis::NodeMapping &expr_mappings,
   return infered;
 }
 
+TyTy::ADTType::ReprOptions
+TypeCheckBase::parse_repr_options (const AST::AttrVec &attrs, Location locus)
+{
+  TyTy::ADTType::ReprOptions repr;
+  repr.pack = 0;
+  repr.align = 0;
+
+  for (const auto &attr : attrs)
+    {
+      bool is_repr = attr.get_path ().as_string ().compare ("repr") == 0;
+      if (is_repr)
+	{
+	  const AST::AttrInput &input = attr.get_attr_input ();
+	  bool is_token_tree = input.get_attr_input_type ()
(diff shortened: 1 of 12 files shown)


gcc/rust/typecheck/rust-hir-type-check-base.cc
```diff
@@ -260,5 +260,72 @@ TypeCheckBase::resolve_literal (const Analysis::NodeMapping &expr_mappings,
...
+	  rust_assert (is_token_tree);
+	  const auto &option = static_cast<const AST::DelimTokenTree &> (input);
+	  AST::AttrInputMetaItemContainer *meta_items
+	    = option.parse_to_meta_item ();
+
+	  const std::string inline_option
+	    = meta_items->get_items ().at (0)->as_string ();
+
+	  // TODO: it would probably be better to make the MetaItems more aware
+	  // of constructs with nesting like #[repr(packed(2))] rather than
```
philberty, line 287: Yeah i think your right here the AST Meta stuff is not very helpful to work with at the moment.

gcc/rust/typecheck/rust-tyty.h
```diff
@@ -1254,6 +1254,20 @@ class ADTType : public BaseType, public SubstitutionRef
...
+    // bool is_c;
+    // bool is_transparent;
+    //...
+
+    // For align and pack: 0 = unspecified. Nonzero = byte alignment.
+    // It is an error for both to be nonzero, this should be caught when
+    // parsing the #[repr] attribute.
+    unsigned char align = 0;
+    unsigned char pack = 0;
+  };
```
philberty, line 1269: I think it might be worth having some kind of error node state here. For example, during the type-checking of this structure, you could represent an error state by a flag or something so that when you are compiling the type we know if it's ok or not so we can error or ignore the options. This will only really matter down the line when we get rid of the saw_errors guards between each compiler pass.

philberty (approved):
Nice work LGTM

Issue #1132, opened 2022-04-18
rust1: internal compiler error: in visit, at rust/typecheck/rust-hir-type-check-path.cc:144


bjorn3:
I tried this code:

```rust
mod mem {
    extern "rust-intrinsic" {
        fn transmute<U, V>(_: U) -> V;
    }
}

pub trait Hasher {
    fn write(&mut self, bytes: &[u8]);
    fn write_u16(&mut self, i: u16) {
        self.write(&mem::transmute::<_, [u8; 2]>(i))
    }
}

pub struct SipHasher;

impl Hasher for SipHasher {
    #[inline]
    fn write(&mut self, msg: &[u8]) {}
}
```

I expected to see this happen: *explanation*

Instead, this happened:

```
rust1: internal compiler error: in visit, at rust/typecheck/rust-hir-type-check-path.cc:144
0x204e319 internal_error(char const*, ...)
	???:0
0x8027ef fancy_abort(char const*, int, char const*)
	???:0
0xa10a9e Rust::Resolver::TypeCheckExpr::visit(Rust::HIR::PathInExpression&)
	???:0
0x9c137c Rust::Resolver::TypeCheckExpr::Resolve(Rust::HIR::Expr*, bool)
	???:0
0x9c2364 Rust::Resolver::TypeCheckExpr::visit(Rust::HIR::CallExpr&)
	???:0
0x9c137c Rust::Resolver::TypeCheckExpr::Resolve(Rust::HIR::Expr*, bool)
	???:0
0x9c4fca Rust::Resolver::TypeCheckExpr::visit(Rust::HIR::BorrowExpr&)
	???:0
0x9c137c Rust::Resolver::TypeCheckExpr::Resolve(Rust::HIR::Expr*, bool)
	???:0
0x9d5a10 Rust::TyTy::TypeCheckMethodCallExpr::visit(Rust::TyTy::FnType&)
	???:0
0x9cd012 Rust::Resolver::TypeCheckExpr::visit(Rust::HIR::MethodCallExpr&)
	???:0
0x9c137c Rust::Resolver::TypeCheckExpr::Resolve(Rust::HIR::Expr*, bool)
	???:0
0x9b7b6a Rust::Resolver::TypeCheckExpr::visit(Rust::HIR::BlockExpr&)
	???:0
0x9c137c Rust::Resolver::TypeCheckExpr::Resolve(Rust::HIR::Expr*, bool)
	???:0
[...]

PR #1190, opened 2022-04-28
Fix ICE in reachability class and refactor associated types code

### LINKED ISSUE #1105: Refactor setup_associated_types
I think the name might be a little unclear. Can we overload the previous `setup_associated_types` or maybe give it a name reflecting the fact that it performs a different operation?

_Originally posted by @CohenArthur in https://github.com/Rust-GCC/gccrs/pull/1086#discussion_r847057432_

philberty:
There are several fixes going on to solve these issues which overlap with one
another so it seemed best to pack them within the same PR.

The main issue for #1128 was that we did not resolve a trait when it was unused
leading to us hitting the ICE in the privacy pass. Since the type check context was
empty for the trait since it was not resolved. To fix this we needed to refactor the
trait resolver to resolve the trait as part of iterating the crate. This resulted in some
regressions in the testsuite so this is why we need the the other commits. Which
required us to finally perform the refactor specified in #1105 to fix these.

Fixes #1105 #1128 #1132


CohenArthur (approved):
LGTM!

Issue #1130, opened 2022-04-18
rust1: internal compiler error: in TyVar, at rust/typecheck/rust-tyty.cc:238


bjorn3:
I tried this code:

```rust
pub trait Hasher {
    fn finish(&self) -> u64;
    fn write(&mut self, bytes: &[u8]);
    fn write_u8(&mut self, i: u8) {
        self.write(&[i])
    }
    fn write_u16(&mut self, i: u16) {
        self.write(&i.to_ne_bytes())
    }
    fn write_u32(&mut self, i: u32) {
        self.write(&i.to_ne_bytes())
    }
    fn write_u64(&mut self, i: u64) {
        self.write(&i.to_ne_bytes())
    }
    fn write_u128(&mut self, i: u128) {
        self.write(&i.to_ne_bytes())
    }
    fn write_usize(&mut self, i: usize) {
        self.write(&i.to_ne_bytes())
    }
    fn write_i8(&mut self, i: i8) {
        self.write_u8(i as u8)
    }
    fn write_i16(&mut self, i: i16) {
        self.write_u16(i as u16)
    }
    fn write_i32(&mut self, i: i32) {
        self.write_u32(i as u32)
    }
    fn write_i64(&mut self, i: i64) {
        self.write_u64(i as u64)
    }
    fn write_i128(&mut self, i: i128) {
        self.write_u128(i as u128)
    }
    fn write_isize(&mut self, i: isize) {
        self.write_usize(i as usize)
    }
}

pub struct SipHasher;

impl Hasher for SipHasher {
    #[inline]
    fn write(&mut self, msg: &[u8]) {
    
    }

    #[inline]
    fn finish(&self) -> u64 {
        0
    }
}

```

I expected to see this happen: Compiles.

Instead, this happened:

```
<source>:5:21: error: expected [[u8]] got [[u8:CAPACITY]]
    3 |     fn write(&mut self, bytes: &[u8]);
      |                                 ~
    4 |     fn write_u8(&mut self, i: u8) {
    5 |         self.write(&[i])
      |                     ^
<source>:5:20: error: expected [& [u8]] got [& [u8:CAPACITY]]
    3 |     fn write(&mut self, bytes: &[u8]);
      |                                ~
    4 |     fn write_u8(&mut self, i: u8) {
    5 |         self.write(&[i])
      |                    ^
<source>:5:20: error: Type Resolution failure on parameter
<source>:5:9: error: failed to lookup type to MethodCallExpr
[...]

PR #1192, opened 2022-04-28
Add an assertion to avoid peeking when the stack is empty

philberty:
This will ensure we get a proper ICE instead of memory corruption/segv.

Addresses #1130


CohenArthur (approved):
Imagine if `std::vector::back` returned an `std::optional` instead of a `{const}reference` with a sprinkle of undefined behavior :)

PR #1193, opened 2022-04-28
Remove unused parameter caller from generating Call expressions

philberty:
Within const context the fncontext maybe empty which in turn results in a
segv for generating const calls which will be evaluated by the const-expr
code anyway.

Addresses #1130


CohenArthur (approved):
LGTM!

PR #1194, opened 2022-04-28
Add missing type inferencing rules for arrays and slices

philberty:
These are missing rulles to allow arrays or slices being coerced with an inference
variable to resolve correctly. This occurs when we use interesting function calls such
as:

```
mod mem {
    extern "rust-intrinsic" {
        fn size_of<T>() -> usize;
        fn transmute<U, V>(_: U) -> V;
    }
}

fn main() {
    let a = 123;
    let b: [u8; mem::size_of::<i32>()] = mem::transmute(a);
}
```

We can add this test case when we implement the transmute intrinsic.


CohenArthur (commented):
Can we add the code in your comment as a test case?

bjorn3:
`transmute` has two restrictions I'm not sure you have added here:

1. Input and output size must be identical.
2. Input and output size must be statically known before monomorphization so eg `transmute::<T, U>` is not allowed, but eg `transmute::<PhantomData<T>, PhantomData<U>>` is allowed. Note that types like `[T; 0]` are not considered to be a fixed size for the purpose of this check.

philberty:
> `transmute` has two restrictions I'm not sure you have added here:
> 
>     1. Input and output size must be identical.
> 
>     2. Input and output size must be statically known before monomorphization so eg `transmute::<T, U>` is not allowed, but eg `transmute::<PhantomData<T>, PhantomData<U>>` is allowed. Note that types like `[T; 0]` are not considered to be of a fixed size for the purpose of this check.

Thanks, @bjorn3 this PR is purely about getting code-generation for the implementation of transmute. Those checks are more related to our missing implementation of the sized auto trait. Those checks will be added in a separate issue and PR's.

CohenArthur (approved):

PR #1197, opened 2022-04-29
Support recursive coercion sites

### LINKED ISSUE #1146: Code generation coercion sites dont handle complex cases

philberty:
There are two changes are part of this PR one where we update the address_expression
helper to take a ptrtype tree. Since the default build_address_expr_loc was automatically
building a pointer type of the TREE_TYPE of the operand which is not correct since we
sometimes want REFERENCE_TYPES instead of POINTER_TYPES. 

The 2nd commit enhances the coercion_site code to recursively walk the tree and their
types to handle coercions to dyn traits or do array bounds checking etc. This gets rid of
the get_root hack used and cleans up the code in general.

Fixes #1146

### DIFF
--- a/gcc/rust/backend/rust-compile-base.cc
+++ b/gcc/rust/backend/rust-compile-base.cc
@@ -317,7 +317,7 @@ HIRCompileBase::mark_addressable (tree exp, Location locus)
 }
 
 tree
-HIRCompileBase::address_expression (tree expr, Location location)
+HIRCompileBase::address_expression (tree expr, tree ptrtype, Location location)
 {
   if (expr == error_mark_node)
     return error_mark_node;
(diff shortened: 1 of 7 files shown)


CohenArthur (approved):
Great! Are there any tests we can add for this? Something like maybe a double or triple reference or something? Otherwise, this looks really good

PR #1199, opened 2022-04-30
Add new as_name interface for Dynamic types

philberty:
The Gimple names of our dyn trait objects were looking like:

  const struct dyn [HIR Trait: FnLike->[C: 0 Nid: 31 Hid: 38 Lid: 13] [(FN call ), ]<Self, &isize, &isize>] & const f

This is a horrible name but useful for debugging this patch fixes this so
we have a seperate naming for generating the type. So now it looks like:

  const struct dyn [FnLike<Self, &isize, &isize>] & const f


CohenArthur (approved):
LGTM

PR #1202, opened 2022-05-02
Resolve module visibility properly

### LINKED ISSUE #1159: Resolve SimplePaths to DefId's

CohenArthur:
Closes #1159

### DIFF
--- a/gcc/rust/privacy/rust-visibility-resolver.cc
+++ b/gcc/rust/privacy/rust-visibility-resolver.cc
@@ -24,8 +24,9 @@
 namespace Rust {
 namespace Privacy {
 
-VisibilityResolver::VisibilityResolver (Analysis::Mappings &mappings)
-  : mappings (mappings)
+VisibilityResolver::VisibilityResolver (Analysis::Mappings &mappings,
+					Resolver::Resolver &resolver)
+  : mappings (mappings), resolver (resolver)
 {}
 
 void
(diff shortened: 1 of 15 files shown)


philberty (approved):
LGTM

CohenArthur:
@philberty Sorry, I didn't realize this would run bad on existing tests. You can review the last commit only to see how that got fixed. If you approve, I'll rebase so that the buildbot is happy with my changes

gcc/rust/privacy/rust-visibility-resolver.cc
```diff
@@ -45,16 +46,56 @@ VisibilityResolver::go (HIR::Crate &crate)
...
+  // FIXME: Add a hint here if we can find the path in another scope, such as
+  // a type or something else
+  // TODO: For the hint, can we point to the original item's definition if
+  // present?
+
+  Resolver::Definition def;
+  rust_assert (resolver.lookup_definition (ref_node_id, &def));
+
+  // FIXME: Is that what we want?
+  ref_node_id = def.parent;
```
philberty, line 77: This is fine, for now just leave it as a FIXME. I was to get rid of this step in the future.

gcc/rust/privacy/rust-visibility-resolver.cc
```diff
@@ -45,16 +46,56 @@ VisibilityResolver::go (HIR::Crate &crate)
...
-// {
-//   return false;
-// }
+  HirId ref;
+  rust_assert (
+    mappings.lookup_node_to_hir (restriction.get_mappings ().get_crate_num (),
+				 ref_node_id, &ref));
+
+  auto module
+    = mappings.lookup_module (restriction.get_mappings ().get_crate_num (),
```
philberty, line 85: also this will get simplified down the line to not require the crate num here but this is fine for now

philberty (approved):
Still looks good :D

CohenArthur:
bors cancel

Issue #854, opened 2022-01-04
Layout of trait objects differs from rustc


philberty:
I tried this code:

```rust
extern "C" {
    fn printf(s: *const i8, ...);
}

struct Foo(i32);
trait Bar {
    fn baz(&self);
}

impl Bar for Foo {
    fn baz(&self) {
        unsafe {
            let a = "%i\n\0";
            let b = a as *const str;
            let c = b as *const i8;

            printf(c, self.0);
        }
    }
}

fn dynamic_dispatch(t: &dyn Bar) {
    t.baz();
}

fn main() {
    let a = &Foo(123);
    let b: &dyn Bar = a;
    dynamic_dispatch(b);
}
```

I expected to see this happen: 

```
Temporary breakpoint 1, test::main () at test.rs:27
27          let a = &Foo(123);
(gdb) n
28          let b: &dyn Bar = a;
(gdb) n
29          dynamic_dispatch(b);
(gdb) p b
$1 = test::&Bar {pointer: 0x555555588004 "{\000", vtable: 0x5555555948b8}
(gdb) p b.vtable
$2 = (*mut [usize; 3]) 0x5555555948b8
(gdb) p *b.vtable
$3 = [93824992252368, 4, 4]
```

Instead, this happened:

```
Temporary breakpoint 1, main () at test.rs:27
27          let a = &Foo(123);
(gdb) n
28          let b: &dyn Bar = a;
(gdb) n
29          dynamic_dispatch(b);
(gdb) p b
$1 = (const dyn [Bar] &) @0x7fffffffdc40: {__receiver_trait_obj_ptr: 0x7fffffffdc5c, __0: 0x401162 <TestCrate::Foo::baz>}
```

see: 

https://github.com/bminor/binutils-gdb/blob/79541a6d9220e800ebc5278594105982d6e1d80c/gdb/rust-lang.c#L231-L262

### Meta

- What version of Rust GCC were you using, git sha if possible. 69f6be3ee483c9895b4b5187a44b3e1c8be2ba63

PR #1205, opened 2022-05-03
Take advantage of OBJ_TYPE_REF'S in dyn calls

### LINKED ISSUE #996: Take advantage of vtable-verify.cc in GCC middle-end
In order to take full advantage of GCC for vtable method resolution we can get security features by using OBJ_TYPE_REF's for all vtable usage in dyn traits. GCC's pass here is simply looking for all gimple statements:

```
  /* Iterate through the immediate uses of the current variable.  If
     it's a virtual function call, we're done.  Otherwise, if there's
     an LHS for the use stmt, add the ssa var to the work list
     (assuming it's not already in the list and is not a variable
     we've already examined.  */
```

This can generate code to call out the libvtv but more importantly using OBJ_TYPE_REF's give much more information for the optimizers which is what we want here.

philberty:
OBJ_TYPE_REF's are the gcc nodes that signify that this is a virtual call
which gives a hint to the optimizers for devirtualization. This patch also enhaces
the dyn object to hold an array of pointers instead of multiple name fields of addresses.

Fixes #996

### DIFF
--- a/gcc/rust/backend/rust-compile-type.cc
+++ b/gcc/rust/backend/rust-compile-type.cc
@@ -532,25 +532,19 @@ TyTyResolveCompile::visit (const TyTy::DynamicObjectType &type)
 
   tree uint = ctx->get_backend ()->integer_type (
     true, ctx->get_backend ()->get_pointer_size ());
-  tree uintptr_ty = ctx->get_backend ()->pointer_type (uint);
+  tree uintptr_ty = build_pointer_type (uint);
 
-  Backend::typed_identifier f ("__receiver_trait_obj_ptr", uintptr_ty,
+  Backend::typed_identifier f ("pointer", uintptr_ty,
 			       ctx->get_mappings ()->lookup_location (
 				 type.get_ty_ref ()));
   fields.push_back (std::move (f));
 
-  for (size_t i = 0; i < items.size (); i++)
-    {
-      // mrustc seems to make a vtable consisting of uintptr's
-      tree uint = ctx->get_backend ()->integer_type (
-	true, ctx->get_backend ()->get_pointer_size ());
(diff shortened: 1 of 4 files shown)


gcc/rust/backend/rust-compile-type.cc
```diff
@@ -532,25 +532,19 @@ TyTyResolveCompile::visit (const TyTy::DynamicObjectType &type)
...
-      tree uintptr_ty = ctx->get_backend ()->pointer_type (uint);
-
-      Backend::typed_identifier f ("__" + std::to_string (i), uintptr_ty,
-				   ctx->get_mappings ()->lookup_location (
-				     type.get_ty_ref ()));
-      fields.push_back (std::move (f));
-    }
+  tree vtable_size = build_int_cst (size_type_node, items.size ());
+  tree vtable_type = ctx->get_backend ()->array_type (uintptr_ty, vtable_size);
+  Backend::typed_identifier vtf ("__vtptr", vtable_type,
```
CohenArthur, line 544: Can we run into any issue where the identifier created here would also be a valid "user" identifier? Would it make more sense to instead use characters that the user could never input in a valid identifier? Something like `@vtptr`, `--vtptr` or `#vtptr` etc

CohenArthur (approved):
Feel free to tell me I'm worrying about nothing :D LGTM otherwise!

gcc/rust/backend/rust-compile-type.cc
philberty: Yeah i know what you mean in C++ they actually use the create_tmp_var code to generate a prefix record with a unique identifier which is a neat trick.

But in rustic they actually don't do that: https://github.com/Rust-GCC/gccrs/issues/854

They actually seem to make a structure like: struct dyn{ pointer , vtable }. So to get the gdb integration i should actually be using those field names instead.

PR #1206, opened 2022-05-03
Use correct format specifiers for unisnged HOST_WIDE_INT

philberty:
The code here was wrongly assuming the unsigned long interface which is not
correctly for all targets.


CohenArthur (approved):
Awww that's too bad haha. I'm assuming this fixes our issues on i386, correct?

philberty:
It should fix it i haven't actually tested it on the build farm yet but it's still the right thing to do

Issue #1207, opened 2022-05-03
Allow `match` on boolean expressions


dafaust:
...
Replacing the `false` with `inside_loop` causes an ICE later on when compiling the match:
```
#0  fancy_abort (file=0x2580950 "../../gccrs/gcc/rust/backend/rust-compile-expr.cc", line=207, 
    function=0x258071e "visit") at ../../gccrs/gcc/diagnostic.cc:2018
#1  0x00000000010d11a8 in Rust::Compile::CompileExpr::visit (this=0x7fffffffcdc0, expr=...)
    at ../../gccrs/gcc/rust/backend/rust-compile-expr.cc:207
#2  0x0000000000f6d172 in Rust::HIR::MatchExpr::accept_vis (this=0x32c0c80, vis=...)
    at ../../gccrs/gcc/rust/hir/tree/rust-hir-full-test.cc:4898
#3  0x0000000000ed6ea5 in Rust::Compile::CompileExpr::Compile (expr=0x32c0c80, ctx=0x7fffffffd720)
    at ../../gccrs/gcc/rust/backend/rust-compile-expr.h:38
#4  0x0000000000ed3340 in Rust::Compile::CompileBlock::visit (this=0x7fffffffcf20, expr=...)
    at ../../gccrs/gcc/rust/backend/rust-compile.cc:93
#5  0x0000000000f6d98e in Rust::HIR::BlockExpr::accept_vis (this=0x32c0d00, vis=...)
    at ../../gccrs/gcc/rust/hir/tree/rust-hir-full-test.cc:5138
#6  0x0000000000ed5f5d in Rust::Compile::CompileBlock::compile (expr=0x32c0d00, ctx=0x7fffffffd720, result=
    0x0) at ../../gccrs/gcc/rust/backend/rust-compile-block.h:34
#7  0x00000000010d9579 in Rust::Compile::CompileExpr::visit (this=0x7fffffffd0d0, expr=...)
    at ../../gccrs/gcc/rust/backend/rust-compile-expr.h:593
#8  0x0000000000f6d242 in Rust::HIR::LoopExpr::accept_vis (this=0x32cb0e0, vis=...)
    at ../../gccrs/gcc/rust/hir/tree/rust-hir-full-test.cc:4922
#9  0x0000000000ed6ea5 in Rust::Compile::CompileExpr::Compile (expr=0x32cb0e0, ctx=0x7fffffffd720)
[...]

PR #1209, opened 2022-05-03
Allow match on boolean expressions

dafaust:
Enables compiling `match` expressions where the scrutinee is a boolean expression. Also enable compiling match arms with Literal patterns, since `true` and `false` literals are commonly used with matches on boolean expressions.

Fixes: #1207

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -211,13 +211,18 @@ CompileExpr::visit (HIR::MatchExpr &expr)
       return;
     }
 
-  rust_assert (scrutinee_expr_tyty->get_kind () == TyTy::TypeKind::ADT);
+  TyTy::TypeKind scrutinee_kind = scrutinee_expr_tyty->get_kind ();
+  rust_assert (scrutinee_kind == TyTy::TypeKind::BOOL
+	       || scrutinee_kind == TyTy::TypeKind::ADT);
 
-  // this will need to change but for now the first pass implementation, lets
-  // assert this is the case
-  TyTy::ADTType *adt = static_cast<TyTy::ADTType *> (scrutinee_expr_tyty);
-  rust_assert (adt->is_enum ());
-  rust_assert (adt->number_of_variants () > 0);
+  if (scrutinee_kind == TyTy::TypeKind::ADT)
+    {
+      // this will need to change but for now the first pass implementation,
+      // lets assert this is the case
+      TyTy::ADTType *adt = static_cast<TyTy::ADTType *> (scrutinee_expr_tyty);
+      rust_assert (adt->is_enum ());
+      rust_assert (adt->number_of_variants () > 0);
+    }
 
   TyTy::BaseType *expr_tyty = nullptr;
   if (!ctx->get_tyctx ()->lookup_type (expr.get_mappings ().get_hirid (),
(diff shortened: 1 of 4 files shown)


dafaust:
oops forgot the clang-format

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -211,13 +211,18 @@ CompileExpr::visit (HIR::MatchExpr &expr)
...
-  TyTy::ADTType *adt = static_cast<TyTy::ADTType *> (scrutinee_expr_tyty);
-  rust_assert (adt->is_enum ());
-  rust_assert (adt->number_of_variants () > 0);
+  if (scrutinee_kind == TyTy::TypeKind::ADT)
+    {
+      // this will need to change but for now the first pass implementation,
+      // lets assert this is the case
+      TyTy::ADTType *adt = static_cast<TyTy::ADTType *> (scrutinee_expr_tyty);
+      rust_assert (adt->is_enum ());
+      rust_assert (adt->number_of_variants () > 0);
```
CohenArthur, line 224: The 3 assertions in there make me wonder if it wouldn't make sense to have a `rust_unimplemented` macro. We want to ICE in that case, but we also want the user to know that we just don't do proper error handling *yet* on match arms, or that we can't *yet* pattern match on structures, or can't *yet* produce a tasteful error for matchin on enums with 0 variants and so on.

I think it might make for a better user experience and encourage the users to report issues or contribute?

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -247,16 +252,30 @@ CompileExpr::visit (HIR::MatchExpr &expr)
...
+	  match_scrutinee_expr, 0, expr.get_scrutinee_expr ()->get_locus ());
+      match_scrutinee_expr_qualifier_expr
+	= ctx->get_backend ()->struct_field_expression (
+	  scrutinee_first_record_expr, 0,
+	  expr.get_scrutinee_expr ()->get_locus ());
+    }
+  else
+    {
+      // FIXME: match on other types of expressions not yet implemented.
+      gcc_assert (0);
```
CohenArthur, line 277:
```suggestion
      gcc_unreachable ();
```
nit but this way it's easier to grep for all those cases :)

CohenArthur (approved):
LGTM :) Leaving @philberty to deal with the nitty gritty backend stuff haha

gcc/rust/backend/rust-compile-pattern.cc
```diff
@@ -80,6 +80,20 @@ CompilePatternCaseLabelExpr::visit (HIR::WildcardPattern &pattern)
...
+{
+  // Compile the literal
+  HIR::LiteralExpr *litexpr
+    = new HIR::LiteralExpr (pattern.get_pattern_mappings (),
+			    pattern.get_literal (), pattern.get_locus (),
+			    std::vector<AST::Attribute> ());
+
+  tree lit = CompileExpr::Compile (litexpr, ctx);
+
+  case_label_expr = build_case_label (lit, NULL_TREE, associated_case_label);
```
philberty, line 94: Oh no way this is really cool, I didn't realise you could do this! But of course you can. I do all the name/type resolution work so that we could support matching complex tuples but doing the match on them really confused me but this opens it up.

In theory we should be able to match integers this way and the case labels support ranges here so that we can support matching integer ranges.

philberty (approved):
This is brilliant i would suggest trying to add in matches for integers and integer ranges. If we have this working then we could refactor the code a bit so that we support this issue

https://github.com/Rust-GCC/gccrs/issues/1081 which I think we just iterate each of the items and match in sucsession to finally hit the match arm block.

gcc/rust/backend/rust-compile-expr.cc
philberty: I think the assertion makes sense here for now because we are saying that we only support bool and ADT enums right now. This patch made me realise I have been over thinking the case labels so actually this patch is really cool

gcc/rust/backend/rust-compile-expr.cc
dafaust: oh whoops I probably was thinking in `rust_assert (0)`.  `gcc_unreachable ()` is definitely better, but for some reason I wasn't thinking of using `gcc_x` for errors in the frontend. But of course it works :smile:

dafaust:
> This is brilliant i would suggest trying to add in matches for integers and integer ranges. If we have this working then we could refactor the code a bit so that we support this issue
> 
> #1081 which I think we just iterate each of the items and match in sucsession to finally hit the match arm block.

Agreed!  I was just thinking about this - it should be just as simple as bools. I wasn't thinking about it for this PR because looking at the enum matching stuff made me think it would be more complicated than it actually should be.

gcc/rust/backend/rust-compile-expr.cc
CohenArthur: Oh definitely, it makes complete sense for now! I was pointing that out as a side issue :) Something we might wanna discuss and consider

Issue #1196, opened 2022-04-28
erroneous "cannot 'break' outside of a loop" from match within LoopExpr


dafaust:
Bumped into this while investigating #253

I tried this code:

```rust
fn bar () -> i32 {

    let mut x = 0;
    let mut y = 0;

    loop {
        match x < 100 {
            true => {
                y += x;
                x += 1;
            },
            _ => { break; },
        }
    }

    return y;
}

fn main () {
    let z = bar ();
}

```

I expected to see this happen: compiles without issue

Instead, this happened: 
```
/home/dfaust/playpen/gccrs/lower-while/bar.rs:13:20: error: cannot ‘break’ outside of a loop
   13 |             _ => { break; },
      |                    ^
/home/dfaust/playpen/gccrs/lower-while/bar.rs:13:20: error: failed to type resolve expression
```

<details><summary> backtrace </summary>
<p>

```
/home/dfaust/playpen/gccrs/lower-while/bar.rs:13:20: error: cannot ‘break’ outside of a loop
   13 |             _ => { break; },
      |                    ^

Breakpoint 5, Rust::Resolver::TypeCheckExpr::visit (this=0x7fffffffcbd0, expr=...)
    at ../../gccrs/gcc/rust/typecheck/rust-hir-type-check-expr.h:904
904		return;
(gdb) bt
#0  Rust::Resolver::TypeCheckExpr::visit (this=0x7fffffffcbd0, expr=...)
    at ../../gccrs/gcc/rust/typecheck/rust-hir-type-check-expr.h:904
#1  0x0000000000f6b5f4 in Rust::HIR::BreakExpr::accept_vis (this=0x32c2da0, vis=...)
    at ../../gccrs/gcc/rust/hir/tree/rust-hir-full-test.cc:4076
#2  0x0000000001044bd5 in Rust::Resolver::TypeCheckExpr::Resolve (expr=0x32c2da0, inside_loop=false)
    at ../../gccrs/gcc/rust/typecheck/rust-hir-type-check-expr.h:49
#3  0x000000000104f614 in Rust::Resolver::TypeCheckStmt::visit (this=0x7fffffffccd0, stmt=...)
    at ../../gccrs/gcc/rust/typecheck/rust-hir-type-check-stmt.h:50
#4  0x0000000000f6c5a2 in Rust::HIR::ExprStmtWithoutBlock::accept_vis (this=0x32cab40, vis=...)
[...]

PR #1211, opened 2022-05-03
Preserve inside_loop context when type checking match

dafaust:
Previously, we would lose the context of being inside a loop when compiling a `match`.
This would lead to incorrect error messages like "cannot 'break' outside of a loop" when
trying to break out of a loop from within a `match` expression.

Fixes: #1196


CohenArthur (approved):
LGTM! Good catch

philberty (approved):
LGTM I really hate that I implemented this check this way. I think we could be putting this as part of the TypeCheck context. Or potentially move this check to be part of the code-generation. But I think this will get fixed up as part of work into the iterator lang items we should be able to desugar all loops into HIR::LoopExpr at that point.

PR #1214, opened 2022-05-04
Replace SSH cloning with HTTPS cloning in README.md

Zopolis4:
HTTPS cloning is the standard across many projects as well as being simpler and more beginner-friendly.

Signed-off-by: Maximilian Downey Twiss <creatorsmithmdt@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ $ apt install build-essential libgmp3-dev libmpfr-dev libmpc-dev flex bison auto
 Clone the repository
 
 ```bash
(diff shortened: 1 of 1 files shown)


CohenArthur (approved):
Thanks

CohenArthur:
I think this is indeed better because it's not just about contributing to the compiler - but simply building it as a "user" (even though we don't have a lot of users right now :D )

philberty (approved):
Sounds good https is better for users

PR #1215, opened 2022-05-04
Report `pub restricted` violations

CohenArthur:
Still in the very early stages and non functional. Needs #1202 to be merged first


gcc/rust/privacy/rust-pub-restricted-visitor.cc
```diff
@@ -0,0 +1,178 @@
...
+  // If there is no visibility in the mappings, then the item is private and
+  // does not contain any restriction
+  if (!mappings.lookup_visibility (item_id, &visibility))
+    return true;
+
+  for (auto mod = module_stack.rbegin (); mod != module_stack.rend (); mod++)
+    if (*mod == visibility.get_module_id ())
+      return true;
+
+  rust_error_at (locus, "restricted path is not an ancestor of the "
```
philberty, line 41: What is the rustc error message like for this case?

gcc/rust/privacy/rust-visibility-resolver.cc
```diff
@@ -45,16 +45,56 @@ VisibilityResolver::go (HIR::Crate &crate)
...
-// }
+  NodeId ref_node_id = UNKNOWN_NODEID;
+  if (!resolver.lookup_resolved_name (ast_node_id, &ref_node_id))
+    {
+      invalid_path.emit_error ();
+      return false;
+    }
+  // FIXME: Add a hint here if we can find the path in another scope, such as
+  // a type or something else
+  // TODO: For the hint, can we point to the original item's definition if
```
philberty, line 69: What do you mean here?

philberty (approved):
LGTM just have a query about one of your comments I might be able to point you in the right direction

gcc/rust/privacy/rust-visibility-resolver.cc
CohenArthur: What I mean is that when we have the following code:
```rust
mod a {
     mod b {
         struct Inner;
    }
}
pub (in a::b::Inner) struct Outer;
````

Let's say we're in this pass, which does not care about whether or not the path used in the restriction is allowed (this is the role of the `PubRestrictedVisitor`. Here, we just want to resolve it to its proper `DefId`).

The visibility used here is a valid *path*: `a::b::Inner` is resolved properly in the SimplePath resolving. However, it does not point to a module, but to a type. The FIXME here says that we should think about hinting regarding the definition of `Inner`. Have a `rust_note` like "cannot use type as visibility restriction" and "item defined here". So we should look through the resolved type scope as well, or check if the visibility used actually refers to a function and not a module, etc etc.

I think I quite know how to do it, it's mostly about doing it piece by piece and focusing on semantics rather than user experience regarding visibilities for now :)

philberty: Ah ok i see what you mean. That should be ok if you fail to lookup in the name scope you can lookup in the type scope, do we care about the label scope? Then the mappings class we could add more enum's so we could switch on the types or add a visitor or something. I did something similar this over here:

https://github.com/Rust-GCC/gccrs/blob/9ea940e4dcabbf99fbb44c125a0af7cf82e48146/gcc/rust/typecheck/rust-hir-inherent-impl-overlap.h#L104

philberty:
As usual great work :)

PR #1217, opened 2022-05-05
Fix size used in unsized adjustments

philberty:
When we coerce from an array to a slice we take the full capacity of the
array as the size to the FatPtr slice object but this was off by one. This
The TYPE_MAX_VALUE is not the correct method of accessing this but instead
it needs to take into account other values to get the correct HOST_WIDE_INT


CohenArthur (approved):
I'm way out of my depth here but really interested in the explanation!

philberty:
> I'm way out of my depth here but really interested in the explanation!

This was a copy paste from the other front-ends i think the idea of TYPE_MAX_VAL is all about the max array index value but to get the capacity value of +1 we need to do that safely using the host wide int magic which needs to take into all the details like precision.

Issue #1177, opened 2022-04-26
Compiler error when compiling the `IfLet` expression.


antego:
related to #1170 

Compilation fails for the `IfLet` expression.

### Code

```rust
fn main() -> i32 {
    let mut res = 0;

    enum E {
        X(u8),
        Y(u8),
        Z(u8),
    }
    let v = E::Y(12);
    if let E::Y(n) = v {
        res = 1;
    }

    0
}
```

### Meta

- gccrs git sha: https://github.com/Rust-GCC/gccrs/commit/b74044fb6278e373da607a8f1f5df2193ce27d65

### Error output

```
FAIL: rust/compile/macro43.rs (internal compiler error: in append_reference_for_def, at rust/resolve/rust-name-resolver.h:227)```
```

antego:
AST dump
```
Crate: 
 inner attributes: none
 items: 
  
i32 main()
BlockExpr:

  outer attributes: none
  inner attributes: none
 statements: 
 
  outer attributes: none
 let mut res = 0
 E
 Generic params: none
 Where clause: none
 Items: 
  X(
  outer attributes: none
 u8)

  outer attributes: none
 let v = CallExpr: 
  outer attributes: none
 Function expr: E::X
 Call params:
  4
  ExprStmtWithBlock: 
IfLetExpr: 
    outer attributes: none
 Condition match arm patterns: 
  TupleStructPattern: 
 Path: E::X
 Tuple struct items: 
  n
 Scrutinee expr: v
 If let block expr:    BlockExpr:
   
     outer attributes: none
     inner attributes: none
    statements: 
    ExprStmtWithoutBlock:
     AssignmentExpr: 
 left: res
 right: n
    final expression: none
   
 final expression: 
0
```

antego:
HIR dump
```
HIR::Crate: 
 inner attributes: none
 items: 
  private 
i32 main()
BlockExpr:
{
 outer attributes: none
 inner attributes: none
 statements: 
 Outer attributes: none
 let mut res = ( 0 ([C: 0 Nid: 6 Hid: 23]))
 private E
 Generic params: none
 Where clause: none
 Items: 
  X [Tuple variant](outer attributes: none
private u8)
 Outer attributes: none
 let v = E::X::[C: 0 Nid: 24 Hid: 33](( 4 ([C: 0 Nid: 25 Hid: 34])),)::[C: 0 Hid: 35]
  ExprStmtWithBlock: 
none (this should not happen and is an error)
[...]

PR #1218, opened 2022-05-05
Lower IfLet expressions

antego:
This PR addresses https://github.com/Rust-GCC/gccrs/issues/1177. This change implements the "hir-lowering" part of the plan outlined here https://github.com/Rust-GCC/gccrs/issues/1177#issuecomment-1118530135.

### DIFF
--- a/gcc/rust/hir/rust-ast-lower-block.h
+++ b/gcc/rust/hir/rust-ast-lower-block.h
@@ -120,6 +120,35 @@ private:
   bool terminated;
 };
 
+class ASTLoweringIfLetBlock : public ASTLoweringBase
+{
+  using Rust::HIR::ASTLoweringBase::visit;
+
+public:
+  static HIR::IfLetExpr *translate (AST::IfLetExpr *expr)
+  {
+    ASTLoweringIfLetBlock resolver;
+    expr->accept_vis (resolver);
+    if (resolver.translated != nullptr)
+      {
+	resolver.mappings->insert_hir_expr (
+	  resolver.translated->get_mappings ().get_crate_num (),
+	  resolver.translated->get_mappings ().get_hirid (),
+	  resolver.translated);
+      }
+    return resolver.translated;
+  }
+
+  ~ASTLoweringIfLetBlock () {}
+
+  void visit (AST::IfLetExpr &expr) override;
+
+private:
+  ASTLoweringIfLetBlock () : ASTLoweringBase (), translated (nullptr) {}
+
+  HIR::IfLetExpr *translated;
+};
+
 class ASTLoweringExprWithBlock : public ASTLoweringBase
 {
   using Rust::HIR::ASTLoweringBase::visit;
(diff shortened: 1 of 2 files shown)


philberty:
Would you like a review @antego or would you still like to do more here?

antego:
Ideally I'd like to make the compilation work. But now I'm not sure since there are possibly too many changes for one PR judging by this https://github.com/Rust-GCC/gccrs/issues/1177#issuecomment-1118530135

CohenArthur:
@antego all the steps mentioned in the issue should be done in multiple PRs imo. It's a good idea to have a PR dedicated to lowering, one to name resolution, one to type checking, one to backend compilation etc.

antego:
> It's a good idea to have a PR dedicated to lowering, one to name resolution, one to type checking, one to backend compilation etc.

Thanks! I'll create separate PRs.

CohenArthur (commented):
I think that overall this looks good :)

gcc/rust/hir/rust-ast-lower-block.h
```diff
@@ -120,6 +120,39 @@ class ASTLoweringIfBlock : public ASTLoweringBase
   bool terminated;
 };
 
+class ASTLoweringIfLetBlock : public ASTLoweringBase
```
CohenArthur, line 123: Any reason why you chose to add a new visitor class instead of simply implementing the function in `ASTLoweringIfBlock`? It seems you're not using new class members or anything, so it might be simpler to reuse this existing class.

antego: The only reason is that `translated` field in the `ASTLoweringIfBlock` is of type `HIR::IfExpr` and can't accomodate the `HIR::IfLetExpr`.

gcc/rust/hir/rust-ast-lower-block.h
```diff
@@ -120,6 +120,39 @@ class ASTLoweringIfBlock : public ASTLoweringBase
...
+
+  void visit (AST::IfLetExpr &expr) override;
+
+private:
+  ASTLoweringIfLetBlock ()
+    : ASTLoweringBase (), translated (nullptr), terminated (false)
+  {}
+
+  HIR::IfLetExpr *translated;
+  bool terminated;
```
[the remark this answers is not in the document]
antego: Honestly, I don't know what `terminated` means. Yeah, the `terminated` is not used in the implementation, I'll remove it.

gcc/rust/hir/rust-ast-lower-block.h
CohenArthur: Then I think it would make more sense to turn that into an `HIR::ExprWithBlock *` but @philberty might think your approach is better. I'm honestly fine with both, I just think it's easier for you to not have to reimplement an entire class/visitor :)

gcc/rust/hir/rust-ast-lower-block.h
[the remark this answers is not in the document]
philberty: Yeah, this terminated piece needs to go away eventually, it was a bad implementation to handle an awkward case in block expressions.

gcc/rust/hir/rust-ast-lower-block.h
philberty: Refactoring to use ExprWithBlock might be a really good idea to cleanup the code a bit here but lets do that as a separate piece since it will affect other code we can raise this as a good-first-pr refactor.

philberty (approved):
LGTM we can do the refactor down the line.

CohenArthur (approved):
bors r+

PR #1219, opened 2022-05-05
Add `Optional<T>` type

CohenArthur:
This adds a tagged union to try and simulate a sum type. This is safer and more ergonomic
 than one of the two alternatives we're currently using in the compiler:
 
 1. Storing a raw pointer, which can be `nullptr` or valid
 
 This is wildly unsafe, and usable in conjunction with local references, stack
 variables, or pointers managed elsewhere, which can cause crashes, hard to
 debug issues or undefined behavior. Likewise, if you do not check for the
 pointer's validity, this will cause a crash.
 
 2. Storing an extra boolean alongside the object
 
 This causes implementors to use a "dummy object": Either an empty version or
 an error version. But what happens if what you really wanted to store was
 the empty or error version? You can also easily incorporate logic bugs if you
 forget to check for the associated boolean.
 
 The `Optional<T>` type has the same "ergonomic" cost: You need to check
 whether your option is valid or not. However, the main advantage is that it
 is more restrictive: You can only acess the member it contains "safely".
 It is similar to storing a value + an associated boolean, but has the
 advantage of making up only one member in your class.
 You also benefit from some helper methods such as `map()`.
 You also get helper functions and operator overloading to "seamlessly"
 replace raw pointer alternatives.
 
  ```c++
 MyType *raw_pointer = something_that_can_fail();
 if (raw_pointer)
     raw_pointer->method();
  // or
  Optional<MyType> opt = something_that_can_fail2();
  if (opt)
      opt->method();
 
  // equivalent to
 
  if (opt.is_some())
      opt.get().method();
  ```
  
  This will be very useful for parent modules when resolving `super` paths :)

### DIFF
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -289,7 +289,8 @@ RUST_INCLUDES = -I $(srcdir)/rust \
 	-I $(srcdir)/rust/util \
 	-I $(srcdir)/rust/typecheck \
 	-I $(srcdir)/rust/privacy \
-	-I $(srcdir)/rust/lint
+	-I $(srcdir)/rust/lint \
+	-I $(srcdir)/rust/util
 
 # add files that require cross-folder includes - currently rust-lang.o, rust-lex.o
 CFLAGS-rust/rust-lang.o += $(RUST_INCLUDES)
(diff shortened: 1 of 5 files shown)


philberty (approved):
Definitely need to start using this :)

PR #1222, opened 2022-05-06
Destructure our generics, placeholers or projections during coercion

philberty:
When we coerce types we need to destructure from the generics in order to
apply these coercion rules correctly or we end up returning a bad
error_mark_node.

This was found while fixing https://github.com/Rust-GCC/gccrs/pull/1220

### DIFF
--- a/gcc/rust/backend/rust-compile.cc
+++ b/gcc/rust/backend/rust-compile.cc
@@ -198,12 +198,15 @@ CompileStructExprField::visit (HIR::StructExprFieldIdentifier &field)
 // Shared methods in compilation
 
 tree
-HIRCompileBase::coercion_site (tree rvalue, TyTy::BaseType *actual,
-			       TyTy::BaseType *expected, Location lvalue_locus,
-			       Location rvalue_locus)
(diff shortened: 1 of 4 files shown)


gcc/rust/backend/rust-compile.cc
```diff
@@ -198,13 +198,50 @@ CompileStructExprField::visit (HIR::StructExprFieldIdentifier &field)
...
+      TyTy::PlaceholderType *p
+	= static_cast<TyTy::PlaceholderType *> (expected);
+      rust_assert (p->can_resolve ());
+      expected = p->resolve ();
+    }
+  else if (expected->get_kind () == TyTy::TypeKind::PROJECTION)
+    {
+      TyTy::ProjectionType *p = static_cast<TyTy::ProjectionType *> (expected);
+      expected = p->get ();
+    }
```
CohenArthur, line 243: Maybe it would be worth to refactor this into a separate function if we want to reuse it later? Since if I understand correctly it's exactly the same code as above

CohenArthur (approved):
Minor nit but looks good :)

PR #1223, opened 2022-05-06
Add new monomorphize_clone interface

### LINKED ISSUE #1133: Bad type resolution for associated types
I tried this code:

```rust
extern "rust-intrinsic" {
}

struct FatPtr<T> {
    data: *const T,
    len: usize,
}

pub union Repr<T> {
    rust: *const [T],
    rust_mut: *mut [T],
    raw: FatPtr<T>,
}

pub enum Option<T> {
    None,
    Some(T),
}

#[lang = "Range"]
pub struct Range<Idx> {
    pub start: Idx,
    pub end: Idx,
}

#[lang = "const_slice_ptr"]
impl<T> *const [T] {
        let a = unsafe { Repr { rust: self }.raw };
        a.len
    }

    pub const fn as_ptr(self) -> *const T {
        self as *const T
    }
}

#[lang = "const_ptr"]
impl<T> *const T {
    pub const unsafe fn offset(self, count: isize) -> *const T {
        unsafe { offset(self, count) }
    }

    }

    pub const fn as_ptr(self) -> *const T {
        self as *const T
    }
}

const fn slice_from_raw_parts<T>(data: *const T, len: usize) -> *const [T] {
    unsafe {
        Repr {
            raw: FatPtr { data, len },
        }
        .rust
    }
}

#[lang = "index"]
trait Index<Idx> {
    type Output;

    fn index(&self, index: Idx) -> &Self::Output;
}

pub unsafe trait SliceIndex<T> {
    type Output;

    fn get(self, slice: &T) -> Option<&Self::Output>;

    unsafe fn get_unchecked(self, slice: *const T) -> *const Self::Output;

    fn index(self, slice: &T) -> &Self::Output;
}

unsafe impl<T> SliceIndex<[T]> for usize {
    type Output = T;

    fn get(self, slice: &[T]) -> Option<&T> {
        unsafe { Option::Some(&*self.get_unchecked(slice)) }
    }

    unsafe fn get_unchecked(self, slice: *const [T]) -> *const T {
        // SAFETY: the caller guarantees that `slice` is not dangling, so it
        // cannot be longer than `isize::MAX`. They also guarantee that
        // `self` is in bounds of `slice` so `self` cannot overflow an `isize`,
        // so the call to `add` is safe.
        unsafe { slice.as_ptr().add(self) }
    }

    fn index(self, slice: &[T]) -> &T {
        // N.B., use intrinsic indexing
        // &(*slice)[self]
        unsafe { &*self.get_unchecked(slice) }
    }
}

unsafe impl<T> SliceIndex<[T]> for Range<usize> {
    type Output = [T];

    fn get(self, slice: &[T]) -> Option<&[T]> {
        if self.start > self.end
        /* || self.end > slice.len() */
        {
            Option::None
        } else {
            unsafe { Option::Some(&*self.get_unchecked(slice)) }
        }
    }

    unsafe fn get_unchecked(self, slice: *const [T]) -> *const [T] {
        unsafe {
            let a: *const T = slice.as_ptr();
            let b: *const T = a.add(self.start);
            slice_from_raw_parts(b, self.end - self.start)
        }
    }

    fn index(self, slice: &[T]) -> &[T] {
        unsafe { &*self.get_unchecked(slice) }
    }
}

impl<T, I> Index<I> for [T]
where
    I: SliceIndex<[T]>,
{
    type Output = I::Output;

    fn index(&self, index: I) -> &I::Output {
        index.index(self)
    }
}

fn main() -> i32 {
    let a/*: [i32; 5]*/ = [1, 2, 3, 4, 5];
    let b/*: &[i32]*/ = &a[1..3];
    let c = b[1];

    0
}

```

Instead, this happened: 

```
rust1: error: expected [[T=i32]] got [i32]
test.rs:78:34: error: expected [& [T=i32]] got [& <placeholder:<Projection=<i32>::T=i32>>]
   78 |     fn index(self, slice: &T) -> &Self::Output;
      |                                  ^
......
  124 |     fn index(self, slice: &[T]) -> &[T] {
      |                                    ~

```

The code begin generated is correctly getting the slice access:

```
i32 test::main ()
{
  struct 
{
  usize start;
  usize end;
} D.186;
  struct 
{
  i32 * data;
  usize len;
} D.187;
  i32 D.188;
  i32 a[5];
  const i32 & const b;
  const i32 c;

  try
    {
      a[0] = 1;
      a[1] = 2;
      a[2] = 3;
      a[3] = 4;
      a[4] = 5;
      D.186.start = 1;
      D.186.end = 3;
      D.187.data = &a;
      D.187.len = 4;
      b = <T as test::Index>::index<i32, Range<usize>> (&D.187, D.186);
      _1 = <T as test::Index>::index<i32, usize> (b, 1);
      c = *_1;
      D.188 = 0;
      return D.188;
    }
  finally
    {
      a = {CLOBBER};
    }
}

```

But you can see the slice from the let binding 'b' has the wrong type here. So I think this is the known issue with one of our lookups for associated impl's.

### Meta

philberty:
This interface is required when we monomorphize since types might be

 placeholder -> projection -> param -> actual-type

This gives us a clean clone interface to destructure this into its actual
type. This is important so for example if we have a generic trait function
and resolve the function for a specific set of generics we update the
associated types. This works, then when we go to substitute the same
generic trait again the placeholders are again updated throwing off the
typing for the first type.

There is more cleanup we can do once this PR is in to make this part of
the general monomorphization path so that we sort out these as part
of the fntype

Fixes #1133

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-expr.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-expr.cc
@@ -469,7 +469,8 @@ TypeCheckExpr::resolve_operator_overload (
   fn->monomorphize ();
 
   // get the return type
-  TyTy::BaseType *function_ret_tyty = type->get_return_type ()->clone ();
+  TyTy::BaseType *function_ret_tyty
+    = type->get_return_type ()->monomorphized_clone ();
 
   // store the expected fntype
   context->insert_operator_overload (expr.get_mappings ().get_hirid (), type);
--- a/gcc/rust/typecheck/rust-tyty.cc
+++ b/gcc/rust/typecheck/rust-tyty.cc
@@ -284,6 +284,32 @@ TyVar::subst_covariant_var (TyTy::BaseType *orig, TyTy::BaseType *subst)
   return TyVar (subst->get_ref ());
 }
 
+TyVar
+TyVar::clone () const
+{
+  TyTy::BaseType *c = get_tyty ()->clone ();
+  return TyVar (c->get_ref ());
+}
+
+TyVar
+TyVar::monomorphized_clone () const
+{
+  auto mappings = Analysis::Mappings::get ();
+  auto context = Resolver::TypeCheckContext::get ();
+
+  // this needs a new hirid
+  TyTy::BaseType *c = get_tyty ()->monomorphized_clone ();
+  c->set_ref (mappings->get_next_hir_id ());
+
+  // insert it
+  context->insert_type (Analysis::NodeMapping (mappings->get_current_crate (),
+					       UNKNOWN_NODEID, c->get_ref (),
+					       UNKNOWN_LOCAL_DEFID),
+			c);
+
+  return TyVar (c->get_ref ());
+}
+
 void
 InferType::accept_vis (TyVisitor &vis)
 {
@@ -372,6 +398,12 @@ InferType::clone () const
   return clone;
 }
 
+BaseType *
+InferType::monomorphized_clone () const
+{
+  return clone ();
+}
+
 bool
 InferType::default_type (BaseType **type) const
 {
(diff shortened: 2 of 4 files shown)


CohenArthur (approved):
LGTM! Nice job!

PR #1238, opened 2022-05-10
Assigning members to structures is a coercion site

### LINKED ISSUE #1235: Bad type resolution on &[T] vs *const T
I tried this code:

```rust
struct FatPtr<T> {
    data: *const T,
    len: usize,
}

pub union Repr<T> {
    rust: *const [T],
    rust_mut: *mut [T],
    raw: FatPtr<T>,
}

impl<T> [T] {
    pub const fn is_empty(&self) -> bool {
        self.len() == 0
    }

    pub const fn len(&self) -> usize {
        unsafe { Repr { rust: self }.raw.len }
    }
}

```

I expected to see this happen: *compile without error

Instead, this happened:

```
<source>:18:31: error: expected [* const [T=T?]] got [& [T=T REF: 46]]
    7 |     rust: *const [T],
      |           ~                    
......
   18 |         unsafe { Repr { rust: self }.raw.len }
      |                               ^
```

see https://godbolt.org/z/KGE8qxzWc

### Meta

- What version of Rust GCC were you using, git sha if possible. 74e836599ce80a11b1fe28065ed7aae6ffa3b7e2

philberty:
Fix missing coercion site call which allows the coecion of a reference to
a pointer type.

Fixes #1235

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-struct.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-struct.cc
@@ -44,7 +44,7 @@ TypeCheckStructExpr::visit (HIR::StructExprStructFields &struct_expr)
 	= TypeCheckExpr::Resolve (struct_expr.struct_base->base_struct.get (),
 				  false);
       struct_def
-	= (TyTy::ADTType *) struct_path_resolved->unify (base_resolved);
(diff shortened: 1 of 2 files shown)


CohenArthur (approved):
LGTM!

PR #1240, opened 2022-05-10
Change Artihmetic Operation to be a cast type operation

### LINKED ISSUE #1234: Bitwise shift operator doesn't cast shift amount
I tried this code:

```rust
fn foo() {
    1u8 << 2u32;
}
```

I expected to see this happen: The shift amount gets casted as necessary to allow this.

Instead, this happened:

```
<source>:2:12: error: expected [u8] got [u32]
    2 |     1u8 << 2u32;
      |     ~      ^
Compiler returned: 1
```

### Meta

  ```
  gccrs (Compiler-Explorer-Build-gcc-74e836599ce80a11b1fe28065ed7aae6ffa3b7e2-binutils-2.38) 12.0.1 20220118 (experimental)
  ```

philberty:
Arithmetic operations like this need a cast to support the range of integer
types which are allow here.

Fixes #1234

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-expr.h
+++ b/gcc/rust/typecheck/rust-hir-type-check-expr.h
@@ -555,7 +555,17 @@ public:
 	return;
       }
 
-    infered = lhs->unify (rhs);
+    switch (expr.get_expr_type ())
+      {
+      case ArithmeticOrLogicalOperator::LEFT_SHIFT:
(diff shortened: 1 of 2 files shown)


bjorn3:
Does this PR also apply to operations like additions? It should only apply to bitshifts.

philberty:
@bjorn3 yes it does I didn't realise it should only be applied to bitshifts I can change that thanks for letting us know.

CohenArthur (approved):
LGTM

PR #1241, opened 2022-05-10
Implement name resolution for the IfLet expression.

antego:
Addresses #1177.

Guidance from the ticket #1177:

> You should be able to copy how we do name resolution for the if blocks from that file. Though if let statements add a new binding and scope level so you need to declare the pattern so you should be able to see how that is done from rust-ast-resolve-stmt.

I don't understand how to modify the block expression resolution so that it can handle the `IfLet` expression. For now, I copied the code from the `MatchExpr` resolution. Not sure how to test it either and what is the desired expected result of the name resolution so I just hope that reviewers will spot the errors.

I created this PR in order to get some guidance about how to proceed with it. Thanks!

### DIFF
--- a/gcc/rust/resolve/rust-ast-resolve-expr.cc
+++ b/gcc/rust/resolve/rust-ast-resolve-expr.cc
@@ -199,6 +199,31 @@ ResolveExpr::visit (AST::IfExprConseqIf &expr)
 }
 
 void
+ResolveExpr::visit (AST::IfLetExpr &expr)
+{
+  resolve_expr (expr.get_value_expr ().get (), expr.get_node_id ());
+
+  NodeId scope_node_id = expr.get_node_id ();
+  resolver->get_name_scope ().push (scope_node_id);
+  resolver->get_type_scope ().push (scope_node_id);
+  resolver->get_label_scope ().push (scope_node_id);
+  resolver->push_new_name_rib (resolver->get_name_scope ().peek ());
+  resolver->push_new_type_rib (resolver->get_type_scope ().peek ());
+  resolver->push_new_label_rib (resolver->get_type_scope ().peek ());
+
+  for (auto &pattern : expr.get_patterns ())
+    {
+      PatternDeclaration::go (pattern.get (), expr.get_node_id ());
+    }
+
+  resolve_expr (expr.get_if_block ().get (), expr.get_node_id ());
+
+  resolver->get_name_scope ().pop ();
+  resolver->get_type_scope ().pop ();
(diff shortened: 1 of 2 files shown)


CohenArthur (approved):
That looks good to me, but waiting for @philberty's expert opinion. One way to test this would be to add a test case using an `if let` construct with a pattern which depends on name resolution, something like:
```rust
mod a {
     enum Foo {
         Bar,
         Baz(i32),
      }
}

fn main() -> i32 {
    let value = a::Foo::Baz(15);
    if let a::Foo::Baz(value) = a {
        return 0;
    }
    
    1
}
```

antego:
> What sort of test-case are you using at the moment?

@philberty still trying to compile the code from #1177 
```
fn main() -> i32 {
    let mut res = 0;

    enum E {
        X(u8),
    }
    let v = E::X(4);
    if let E::X(n) = v {
        res = n;
    }

    0
}
```
It fails with
```
rust1: internal compiler error: in append_reference_for_def, at rust/resolve/rust-name-resolver.cc:200
```
I don't know if this error should be fixed by this PR or not.

CohenArthur:
@antego I think this might be because you're not inserting a new definition? I'm not really sure as I've never worked on pattern resolution before. But the assertion is being triggered because the `defId` was not declared in this scope or an outer one. If we look at the `ResolvePattern` or `PatternDeclaration` class, they often include calls to `resolver->insert_new_definition` which might be what you need?

Sorry if this doesn't help :DD Trying to figure out something so you can progress!

antego:
@CohenArthur appreciate the attempt but I still can't understand why `MatchExpr` works without calling the `insert_new_definition`.

While trying out different things I noticed that the part of test code which causes the error is actually the line
```
    let v = E::X(4);
```
The minimal test case that fails with the error above is
```
```

antego:
This test case
```
enum E {
    X(u8),
}

fn main() -> i32 {
    let mut res = 0;
    let v = E::X(4);
    if let E::X(n) = v {
        res = n;
    }

    0
}
```
fails with the error
```
error: failed to type resolve expression
    8 |     if let E::X(n) = v {
      |     ^

```
So I guess the name resolution part is figured out and the type resolution is next?

antego:
I guess the failing name resolution for enums should be addressed in another issue?

antego:
@CohenArthur @philberty are you happy to merge this PR since the name resolution doesn't throw the error anymore?

philberty:
bors try

philberty (approved):

PR #1244, opened 2022-05-10
Allow match on integer and char types

dafaust:
Enable compiling match expressions on other primitive types like `i32`, `usize` and `char`.

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -212,7 +212,8 @@ CompileExpr::visit (HIR::MatchExpr &expr)
     }
 
   TyTy::TypeKind scrutinee_kind = scrutinee_expr_tyty->get_kind ();
-  rust_assert (scrutinee_kind == TyTy::TypeKind::BOOL
+  rust_assert ((TyTy::is_primitive_type_kind (scrutinee_kind)
+		&& scrutinee_kind != TyTy::TypeKind::NEVER)
 	       || scrutinee_kind == TyTy::TypeKind::ADT);
 
   if (scrutinee_kind == TyTy::TypeKind::ADT)
@@ -222,6 +223,13 @@ CompileExpr::visit (HIR::MatchExpr &expr)
       TyTy::ADTType *adt = static_cast<TyTy::ADTType *> (scrutinee_expr_tyty);
       rust_assert (adt->is_enum ());
       rust_assert (adt->number_of_variants () > 0);
+    }
+  else if (scrutinee_kind == TyTy::TypeKind::FLOAT)
+    {
+      // FIXME: CASE_LABEL_EXPR does not support floating point types.
+      // Find another way to compile these.
+      sorry_at (expr.get_locus ().gcc_location (),
+		"match on floating-point types is not yet supported");
     }
 
   TyTy::BaseType *expr_tyty = nullptr;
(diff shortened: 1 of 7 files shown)


dafaust:
In theory these changes would also allow match on float types, but the  `CASE_LABEL_EXPR` tree nodes do not support floating-point types, so we cannot compile them yet.

CohenArthur:
> In theory these changes would also allow match on float types, but the `CASE_LABEL_EXPR` tree nodes do not support floating-point types, so we cannot compile them yet. We'll have to find another way to handle these.

We might not even have to support them for long! https://github.com/rust-lang/rust/issues/41620

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -223,6 +224,13 @@ CompileExpr::visit (HIR::MatchExpr &expr)
       rust_assert (adt->is_enum ());
       rust_assert (adt->number_of_variants () > 0);
     }
+  else if (scrutinee_kind == TyTy::TypeKind::FLOAT)
+    {
+      // FIXME: CASE_LABEL_EXPR does not support floating point types.
+      // Find another way to compile these.
+      sorry_at (expr.get_locus ().gcc_location (),
```
CohenArthur, line 231: This is great :D I think it would be nice to have a `rust_sorry` wrapper to which we can give a `Location` directly instead of a `location_t`. But in another PR :)

CohenArthur (approved):
LGTM! Great stuff

philberty:
Interesting from testing on rustc we get:

```
warning: floating-point types cannot be used in patterns
 --> <source>:7:9
  |
7 |         15f32 => unsafe {
  |         ^^^^^
  |
  = note: `#[warn(illegal_floating_point_literal_pattern)]` on by default
  = note: for more information, see issue #41620 <https://github.com/rust-lang/rust/issues/41620>

warning: 1 warning emitted
```

When I tried to use a float as a match pattern. Though testing the rust gcc codegen they seem to use conditionals instead of switch expressions which is interesting. https://godbolt.org/z/nx3Y3nEbd

I think for match expressions its looking like we need to have specific handles based on the type so we could have a match code generation-specific for floats using conditionals. It also seemed from my previous look into this that matching a tuple needs special handling as well.

gcc/rust/typecheck/rust-tyty.h
```diff
@@ -67,6 +67,9 @@ enum TypeKind
   ERROR
 };
 
+bool
```
philberty, line 70: should this be extern bool is_primitive_type_kind();

philberty (approved):
LGTM.

I wonder can you have a think about this issue:  https://github.com/Rust-GCC/gccrs/issues/1081

I added the name-resolution and type-resolution steps here: https://github.com/Rust-GCC/gccrs/pull/1144

I got stumped on the code-generation, but I think you have all the primitives required to do this now. I think it would be worth looking at this compiler explorer gimple dump might give some idea how we can do it:

https://godbolt.org/z/xKG68raPr

Not sure whats going on there but I think the code-gen is switching on the discriminate in the enum but not sure if its doing anything with the integer literal but it might be optimized out not 100% sure.

philberty:
@dafaust feel free to merge when your ready :D

dafaust:
> @dafaust feel free to merge when your ready :D

Will do, thanks :) just wanted to fix the extern thing.

PR #1248, opened 2022-05-12
Support RangePatterns in matches

dafaust:
Add name resolution, lowering, type checking and compilation for `RangePattern`s in matches.

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-pattern.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-pattern.cc
@@ -273,5 +273,82 @@ TypeCheckPattern::visit (HIR::LiteralPattern &pattern)
 			     pattern.get_literal (), pattern.get_locus ());
 }
 
+void
+TypeCheckPattern::visit (HIR::RangePattern &pattern)
+{
+  // Resolve the upper and lower bounds, and ensure they are compatible types
+  TyTy::BaseType *upper = nullptr, *lower = nullptr;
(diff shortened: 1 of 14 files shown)


gcc/rust/typecheck/rust-hir-type-check-pattern.cc
```diff
@@ -273,5 +273,82 @@ TypeCheckPattern::visit (HIR::LiteralPattern &pattern)
...
 }
 
+void
+TypeCheckPattern::visit (HIR::RangePattern &pattern)
+{
+  // Resolve the upper and lower bounds, and ensure they are compatible types
+  TyTy::BaseType *upper = nullptr, *lower = nullptr;
+
+  // It would be nice to factor this out into a helper since the logic for both
+  // bounds is exactly the same...
```
CohenArthur, line 283:
```suggestion
  // TODO: It would be nice to factor this out into a helper since the logic for both
  // bounds is exactly the same...
```
Or FIXME, whichever you like more. But I think that makes it easier to find and fix tiny issues like those

CohenArthur (approved):
This is super cool!!!!! Great job!! :D

philberty (approved):
LGTM i was wondering if you needed the Range lang-item structs but they don't really help anything here.

dafaust:
(fixing up that comment from the github UI added a commit which broke the clang-format lint - fixed+pushed locally)

PR #1250, opened 2022-05-17
Support extern-blocks defined within blocks

### LINKED ISSUE #976: Cannot access extern items defined in inner block
I tried this code:

```rust
fn main() {
    {
        extern "C" {
            fn puts(s: *const i8);
        }

        unsafe {
            puts("hi\0" as *const str as *const i8);
        }
    }
}
```

I expected to see this happen: No errors, rustc compiles this snippet fine

Instead, this happened:
```rust
test.rs:14:13: error: Cannot find path ‘puts’ in this scope
   14 |             puts("hi\0");
      |             ^
```

on master

philberty:
This adds support for declaring extern blocks within blocks. So this adds
the missing name resolution visitor for the statement context. Then we
extract out a common extern block hir lowering function for both contexts.

The type resolution step needs to be updated to be like the code-generation
step so that we can solve these paths in a query rather than a top-down
approach but this is a known issue.

The final step was to support query-based compilation to extern functions.

Fixes #976

### DIFF
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -94,6 +94,7 @@ GRS_OBJS = \
     rust/rust-ast-resolve-expr.o \
     rust/rust-ast-resolve-type.o \
     rust/rust-ast-resolve-path.o \
+    rust/rust-ast-resolve-stmt.o \
     rust/rust-hir-type-check.o \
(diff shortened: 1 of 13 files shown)


CohenArthur (approved):
Ah, I see that your clang-format version got updated :D LGTM!

PR #1252, opened 2022-05-17
privacy: Handle calls to functions defined in previous ancestors

CohenArthur:
Previously, we would only check if `current_module` was a direct
descendant of the item's module. However, we also need to visit each of this
item's module's children recursively.


philberty (approved):
LGTM

Issue #1253, opened 2022-05-17
`macro-issue1233.rs` is broken on big-endian systems


CohenArthur:
The issue probably comes to how we perform cfg expansion in blocks, regarding statements and final expressions.
Here is a test case failing as well on LE systems.
```rust
// { dg-additional-options "-w" }

macro_rules! impl_uint {
    ($($ty:ident = $lang:literal),*) => {
        $(
            impl $ty {
                pub fn to_le(self) -> Self {
                    #[cfg(not(target_endian = "big"))]
                    {
                        self
                    }
                    #[cfg(target_endian = "big")]
                    {
                        self
                    }
                }
            }
        )*
    }
}

impl_uint!(u8 = "u8", u16 = "u16", u32 = "u32");
```

The order of the "blocks + cfg" in the test case matter, which means that what is probably happening is we are parsing the first block + cfg as a statement. When the second block + cfg gets removed, due to the predicate not being met, we end up with a statement and no return expression. So the function gets a Self as a statement, ignores that, returns a (), and we get a type error.

And for an example that `rustc` can compile without complaining about redefining primitive items:
```rust
// { dg-additional-options "-w" }

macro_rules! impl_uint {
    ($($ty:ident = $lang:literal),*) => {
        $(
            struct $ty;

            impl $ty {
                pub fn to_le(self) -> Self {
                    #[cfg(not(target_endian = "big"))]
                    {
                        self
                    }
                    #[cfg(target_endian = "big")]
                    {
                        self
                    }
                }
            }
        )*
    }
}

impl_uint!(z1 = "z1", z16 = "z16", z12 = "z12");
```

CohenArthur:
The corresponding struct in `rustc` simply contains one `Vec<Stmt>` and no optional tail expression: https://github.com/rust-lang/rust/blob/77972d2d0134fb597249b3b64dcf9510a790c34e/compiler/rustc_ast/src/ast.rs#L567-L583

CohenArthur:
This should be fixed by #2156 as well

Issue #1233, opened 2022-05-09
double free or corruption (out)


bjorn3:
### Code

```rust
macro_rules! impl_uint {
    ($($ty:ident = $lang:literal),*) => {
        $(
            impl $ty {
                pub fn to_le(self) -> Self {
                    #[cfg(not(target_endian = "little"))]
                    {
                        self.swap_bytes()
                    }
                }
            }
        )*
    }
}

impl_uint!(u8 = "u8", u16 = "u16", u32 = "u32");
```

### Meta

- What version of Rust GCC were you using, git sha if possible.
  ```
  gccrs (Compiler-Explorer-Build-gcc-74e836599ce80a11b1fe28065ed7aae6ffa3b7e2-binutils-2.38) 12.0.1 20220118 (experimental)
  ```

### Error output

```
<output>
```

<details><summary>Backtrace</summary>
  <p>
  
  ```
gccrs-snapshot - 670ms

double free or corruption (out)
rust1: internal compiler error: Aborted
0x205eb59 internal_error(char const*, ...)
	???:0
0x7f360563303b raise
	???:0
0x7f3605612858 abort
	???:0
0x8c9bbe Rust::Session::expansion(Rust::AST::Crate&)
	???:0
0x8ccfa4 Rust::Session::parse_file(char const*)
	???:0
0x8cdc18 Rust::Session::parse_files(int, char const**)
	???:0
Please submit a full bug report,
with preprocessed source if appropriate.
Please include the complete backtrace with any bug report.
See <https://gcc.gnu.org/bugs/> for instructions.
Compiler returned: 1
  ```
  
  </p>
</details>

philberty:
@dkm it seems compiler explorer is missing debug symbols does it make sense for us to compile with -g on compiler explorer?

CohenArthur:
Investigated a little more and the crash seems to originate from deallocating the `std::vector<ContextType>` in the `MacroExpander` class. I'll look into it, thanks a lot!

PR #1254, opened 2022-05-18
issue #1233: Do not rely on the endianness for testing

CohenArthur:
This testcase uncovered a very interesting bug requiring a refactor of
our `AST::Block` class (#1253), but should still be fixed/adapted in the
meantime so that the BE builds on our buildbot do not fail.

I've tested this newtestcase with a compiler from 74e836599ce80a11b1fe28065ed7aae6ffa3b7e2, which was the commit pointed out in #1233. The same ICE would still trigger, so I can safely say that this is a different exemple showing the same underlying issue. I'll work on fixing #1253 but it is a refactor we need to think about a little.

This should make all the architectures on buildbot happy again!


philberty (approved):
LGTM

PR #1255, opened 2022-05-18
Privacy report expressions

CohenArthur:
This adds privacy reporting to all expression types inside our HIR. It also restricts the `PrivacyReporter` visitor to `HIR::Expression`s and `HIR::Stmt`s as a lot of the previous types did not need to be visited.

### DIFF
--- a/gcc/rust/privacy/rust-privacy-reporter.cc
+++ b/gcc/rust/privacy/rust-privacy-reporter.cc
@@ -39,10 +39,13 @@ void
 PrivacyReporter::check_for_privacy_violation (const NodeId &use_id,
 					      const Location &locus)
 {
-  NodeId ref_node_id;
+  NodeId ref_node_id = UNKNOWN_NODEID;
 
   // FIXME: Don't assert here - we might be dealing with a type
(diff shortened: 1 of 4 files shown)


CohenArthur:
Sorry for the really large commit! It didn't make too much sense to split it as the changes are extremely similar

gcc/rust/privacy/rust-privacy-reporter.cc
```diff
@@ -93,161 +96,218 @@ void
...
-
-void
-PrivacyReporter::visit (HIR::TypePathSegmentGeneric &segment)
-{}
-
 void
 PrivacyReporter::visit (HIR::TypePathSegmentFunction &segment)
-{}
+{
+  // FIXME: Do we need to do anything for this?
```
philberty, line 109: We haven't done anything for these yet in general i always have to lookup the rust reference to remember what they are.

CohenArthur: Mmh interesting. I'll look into it when figuring out the last bits of type privacy reporting

Issue #1256, opened 2022-05-18
Add location to `HIR::Type` base class


CohenArthur:
This will be very useful when reporting privacy errors. The "private type errors" currently look like this:
```rust
gcc/testsuite/rust/compile/privacy5.rs:15:16: error: definition is private in this context
   15 |         fn any(a0: green::Foo, a1: green::Bar) {}
      |                ^
```

while it would make for a much better user experience to have the caret point to the type, which is the root cause of the issue here.

PR #1258, opened 2022-05-18
Report privacy errors on explicit types

CohenArthur:
~~Needs #1255, so skip reviewing the first commit or wait for it to be merged~~ Done and rebased :)

### DIFF
--- a/gcc/rust/hir/tree/rust-hir.h
+++ b/gcc/rust/hir/tree/rust-hir.h
@@ -486,6 +486,8 @@ protected:
   virtual Type *clone_type_impl () const = 0;
 
   Analysis::NodeMapping mappings;
+
+  // FIXME: How do we get the location here for each type?
 };
 
 // A type without parentheses? - abstract
--- a/gcc/rust/privacy/rust-privacy-reporter.cc
+++ b/gcc/rust/privacy/rust-privacy-reporter.cc
@@ -6,9 +6,10 @@
 namespace Rust {
 namespace Privacy {
 
-PrivacyReporter::PrivacyReporter (Analysis::Mappings &mappings,
-				  Resolver::Resolver &resolver)
-  : mappings (mappings), resolver (resolver),
+PrivacyReporter::PrivacyReporter (
+  Analysis::Mappings &mappings, Resolver::Resolver &resolver,
+  const Rust::Resolver::TypeCheckContext &ty_ctx)
+  : mappings (mappings), resolver (resolver), ty_ctx (ty_ctx),
     current_module (Optional<NodeId>::none ())
 {}
 
@@ -52,7 +53,11 @@ PrivacyReporter::check_for_privacy_violation (const NodeId &use_id,
   if (!resolver.lookup_resolved_name (use_id, &ref_node_id))
     resolver.lookup_resolved_type (use_id, &ref_node_id);
 
-  rust_assert (ref_node_id != UNKNOWN_NODEID);
+  // FIXME: Assert here. For now, we return since this causes issues when
+  // checking inferred types (#1260)
+  // rust_assert (ref_node_id != UNKNOWN_NODEID);
+  if (ref_node_id == UNKNOWN_NODEID)
+    return;
 
   ModuleVisibility vis;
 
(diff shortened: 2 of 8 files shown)


gcc/rust/privacy/rust-privacy-reporter.cc
```diff
@@ -48,11 +49,19 @@ PrivacyReporter::check_for_privacy_violation (const NodeId &use_id,
...
     resolver.lookup_resolved_type (use_id, &ref_node_id);
 
-  rust_assert (ref_node_id != UNKNOWN_NODEID);
+  // FIXME: Assert here. For now, we return since this causes issues when
+  // checking inferred types (#1260)
+  // rust_assert (ref_node_id != UNKNOWN_NODEID);
+  if (ref_node_id == UNKNOWN_NODEID)
+    return;
+
+  rust_debug ("ARTHUR: resolved_id = %d", ref_node_id);
```
philberty, line 64: Maybe check this debug message to check_for_privacy_violation: hirid %d, ref_node_id

gcc/rust/privacy/rust-privacy-reporter.cc
```diff
@@ -97,6 +106,102 @@ PrivacyReporter::check_for_privacy_violation (const NodeId &use_id,
...
+	// FIXME: Can we use `resolve` here? Is that what we should do?
+	static_cast<const TyTy::PlaceholderType *> (ty)->resolve ());
+    case TyTy::PROJECTION:
+      return recursive_check (
+	static_cast<const TyTy::ProjectionType *> (ty)->get ());
+    case TyTy::NEVER:
+    case TyTy::CLOSURE:
+    case TyTy::ERROR:
+    case TyTy::INFER:
+      rust_unreachable ();
```
philberty, line 174: I think in this case you can just return because they are like builtins

gcc/rust/privacy/rust-privacy-reporter.cc
```diff
@@ -97,6 +106,102 @@ PrivacyReporter::check_for_privacy_violation (const NodeId &use_id,
...
+      for (auto &param : static_cast<const TyTy::FnPtr *> (ty)->get_params ())
+	recursive_check (param.get_tyty ());
+      return recursive_check (
+	static_cast<const TyTy::FnPtr *> (ty)->get_return_type ());
+    case TyTy::TUPLE:
+      for (auto &param :
+	   static_cast<const TyTy::TupleType *> (ty)->get_fields ())
+	recursive_check (param.get_tyty ());
+      return;
+    case TyTy::PLACEHOLDER:
```
philberty, line 163: Placeholders, projections and ParamTypes need to be recursive, so there is a helper for that, i will be raising in a separate PR you will end up calling destructure on them to get the actual type out because there are cases where you can end up with:

Tuple->field_a->placeholder->projection->ParamType->i32

philberty (approved):
LGTM minor tweaks we can make when I raise the recursive destructure PR

gcc/rust/hir/tree/rust-hir.h
```diff
@@ -486,6 +486,8 @@ class Type : public Node
   virtual Type *clone_type_impl () const = 0;
 
   Analysis::NodeMapping mappings;
+
+  // FIXME: How do we get the location here for each type?
```
philberty, line 490: raise an issue for this

philberty (approved):

gcc/rust/hir/tree/rust-hir.h
CohenArthur: #1256

gcc/rust/privacy/rust-privacy-reporter.cc
CohenArthur: I think `CLOSURE` might be the only one that needs special handling here as closures can contain explicit types:
```rust
mod foo {
    struct Priv;
}

fn outer() {
    let closure = |value: foo::Priv| -> foo::Priv { value };
}
```

I'll make them all return except for closures, does that sound good?

CohenArthur: We can also handle closures in this PR maybe but I was unsure as to how well they were handled in the rest of the compiler

PR #1267, opened 2022-05-22
Marklive: support match expr

thomasyonug:
This makes `MarkLive` support `HIR::MatchExpr` and adds a related test case.


gcc/rust/lint/rust-lint-marklive.h
```diff
@@ -204,6 +204,21 @@ class MarkLive : public MarkLiveBase
...
 
+  void visit (HIR::MatchExpr &expr) override
+  {
+    expr.get_scrutinee_expr ()->accept_vis (*this);
+    std::vector<HIR::MatchCase> &cases = expr.get_match_cases ();
+    for (auto &&caz : cases)
+      {
+	auto case_arm = caz.get_arm ();
+	if (case_arm.has_match_arm_guard ())
+	  {
```
dkm, line 215: nit: that's a style preference, but I prefer single statement if block to be {}-less and I think it's closer to the GCC code base...

thomasyonug: Thanks, it's better to keep closer to the GCC code base

philberty (approved):
LGTM nice work! its good to have you back @thomasyonug :D

Issue #192, opened 2021-01-27
If let expressions


philberty:
https://doc.rust-lang.org/reference/expressions/if-expr.html

This work will depend on #189 being done first

PR #1279, opened 2022-05-25
Implement type checking for the `IfLet` expression.

antego:
This change implements the type check for the `IfLet` expression. This is the adapted type checking code from the `MatchExpr`.

Tested on this code
```
enum E {
    X(u8),
}

fn main() -> i32 {
    let mut res = 0;
    let v = E::X(4);
    if let E::X(n) = v {
        res = n;
    }

    0
}
```
Compilation finishes without errors.

Next is the implementation of the code generation but I'll need help to find where it's located.

Addresses #192 #1177

### DIFF
--- a/gcc/rust/hir/tree/rust-hir-expr.h
+++ b/gcc/rust/hir/tree/rust-hir-expr.h
@@ -3477,6 +3477,19 @@ public:
   void accept_vis (HIRFullVisitor &vis) override;
   void accept_vis (HIRExpressionVisitor &vis) override;
 
+  std::unique_ptr<Expr> &get_scrutinee_expr ()
+  {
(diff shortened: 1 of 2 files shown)


philberty (approved):
LGTM eventually we will have helpers like gcc's global trees for TyTy unit-type and error type nodes.

antego:
> Next is the implementation of the code generation but I'll need help to find where it's located.

I guess this should be implemented as part of `rust-compile-expr.h` https://github.com/Rust-GCC/gccrs/blob/39ae84825c5cc15edbb2fa9f5ecc70f93870e3ee/gcc/rust/backend/rust-compile-expr.h

Issue #1271, opened 2022-05-24
&Str has the same layout as &[T]


philberty:
see https://github.com/rust-lang/rust/blob/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/core/src/str/mod.rs#L224-L227

PR #1280, opened 2022-05-25
Str's have the same layout as [T]

### LINKED ISSUE #1023: Slice type error on &str -> &&str
Directly related to #391 

I tried this code:

```rust

fn foo(e: &str) -> &str { // { dg-bogus "expected" "#391" { xfail *-*-* } }
    &"" // { dg-bogus "expected" "#391" { xfail *-*-* } }
}
```

Instead, this happened: bogus typechecking errors

### Meta

- What version of Rust GCC were you using: b0db17b72ee

philberty:
Raw strings have a very specific type layout which maps over to Slices. It
also has very specific type checking rules so for example:

    let a:&str = "TEST 1";
    let b:&str = &"TEST 2";

Are both the same type this is likely to be for all DST's but lets do one
rule at a time.

Fixes #1023 #1271

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -1014,13 +1014,28 @@ tree
 CompileExpr::compile_string_literal (const HIR::LiteralExpr &expr,
 				     const TyTy::BaseType *tyty)
 {
+  tree fat_pointer = TyTyResolveCompile::compile (ctx, tyty);
+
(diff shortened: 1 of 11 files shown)


bjorn3:
```
let b:&str = &"TEST 2";
```

This is true for all reference types it seems. Rustc will dereference as much as needed and then take a single reference. In this case it would do `&**"TEST 2"`.

philberty:
bors try

dafaust (approved):
Cool! LGTM

PR #1286, opened 2022-05-29
gccrs const folding: port over cp_walk_subtrees()

abbasfaisal:
Card: [Link](https://github.com/Rust-GCC/gccrs/projects/16#card-82441327)

Most of TREE_CODEs in the original function belonged to cp/cp-tree.def which have been removed so this ported function is a lot smaller.

### DIFF
--- a/gcc/rust/backend/rust-tree.cc
+++ b/gcc/rust/backend/rust-tree.cc
@@ -674,4 +674,90 @@ pointer_offset_expression (tree base_tree, tree index_tree, location_t location)
 			  base_tree, offset);
 }
 
+// forked from gcc/cp/tree.cc cp_walk_subtrees
+/* Apply FUNC to all language-specific sub-trees of TP in a pre-order
+   traversal.  Called from walk_tree.  */
+
+tree
+rs_walk_subtrees (tree *tp, int *walk_subtrees_p, walk_tree_fn func, void *data,
+		  hash_set<tree> *pset)
+{
+  enum tree_code code = TREE_CODE (*tp);
+  tree result;
+
+#define WALK_SUBTREE(NODE)                                                     \
+  do                                                                           \
+    {                                                                          \
+      result = rs_walk_tree (&(NODE), func, data, pset);                       \
+      if (result)                                                              \
+	goto out;                                                              \
+    }                                                                          \
+  while (0)
+
+  if (TYPE_P (*tp))
+    {
+      /* If *WALK_SUBTREES_P is 1, we're interested in the syntactic form of
+	 the argument, so don't look through typedefs, but do walk into
+	 template arguments for alias templates (and non-typedefed classes).
+
+	 If *WALK_SUBTREES_P > 1, we're interested in type identity or
+	 equivalence, so look through typedefs, ignoring template arguments for
+	 alias templates, and walk into template args of classes.
+
+	 See find_abi_tags_r for an example of setting *WALK_SUBTREES_P to 2
+	 when that's the behavior the walk_tree_fn wants.  */
+      if (*walk_subtrees_p == 1 && typedef_variant_p (*tp))
+	{
+	  *walk_subtrees_p = 0;
+	  return NULL_TREE;
+	}
+    }
+
+  /* Not one of the easy cases.  We must explicitly go through the
+     children.  */
+  result = NULL_TREE;
+  switch (code)
+    {
+    case TREE_LIST:
+      WALK_SUBTREE (TREE_PURPOSE (*tp));
+      break;
+
+    case RECORD_TYPE:
+      if (TYPE_PTRMEMFUNC_P (*tp))
+	WALK_SUBTREE (TYPE_PTRMEMFUNC_FN_TYPE_RAW (*tp));
+      break;
+
+    case CONSTRUCTOR:
+      if (COMPOUND_LITERAL_P (*tp))
+	WALK_SUBTREE (TREE_TYPE (*tp));
+      break;
+
+    case DECL_EXPR:
+      /* User variables should be mentioned in BIND_EXPR_VARS
+	 and their initializers and sizes walked when walking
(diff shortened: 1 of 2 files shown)


CohenArthur:
> Card: [Link](https://github.com/Rust-GCC/gccrs/projects/16#card-82441327)
> 
> This PR is WIP and opened as RFC.
> 
>     * This copies over all the macros in cp-tree.h to rust-tree.h as this makes it easy to port further stuff from cp. This compiles even if the functions and structs inside the unused macros haven't been ported yet. The problem this would leave us with is it will make it difficult to change these to be look relevant to rust rather than c++. I think it would be easy to do that if we ported them one by one. Thoughts?

Agreed. I think this change is really big, and I don't mind it being merged, but I fear it might be hard to work with. I think an approach that would be better for you and people reviewing PRs is adding changes incrementally, bringing over everything you need to make them work in the meantime. Of course, @philberty will have the last word as he has a better understanding of your project :)

>     * The template stuff relavant to cp_walk_subtrees() in cp has also been copied over. I guess we would need to remove it all?

I think you're right
 
>     * Other than above two, there is C++ specific stuff in struct definitions and functions that I guess will need to be reviewed line by line to weed out.
> 
>     * [This ](https://github.com/Rust-GCC/gccrs/compare/master...abbasfaisal:master#diff-c53c9c947998d46ed9d083290a905468978cb0023c9cc4d52708c487b99b75b3R684) and [this ](https://github.com/Rust-GCC/gccrs/compare/master...abbasfaisal:master#diff-c53c9c947998d46ed9d083290a905468978cb0023c9cc4d52708c487b99b75b3R684) bit need a careful look.
> 
>     * Most copied over things are still prefixed with cp. I will change this once above points are okayed.

gcc/rust/backend/rust-tree.cc
```diff
@@ -674,4 +674,289 @@ pointer_offset_expression (tree base_tree, tree index_tree, location_t location)
...
+
+    case LAMBDA_EXPR:
+      /* Don't walk into the body of the lambda, but the capture initializers
+	 are part of the enclosing context.  */
+      for (tree cap = LAMBDA_EXPR_CAPTURE_LIST (*tp); cap;
+	   cap = TREE_CHAIN (cap))
+	WALK_SUBTREE (TREE_VALUE (cap));
+      break;
+
+    case CO_YIELD_EXPR:
```
philberty, line 915: we dont need to port over YIELD

gcc/rust/backend/rust-tree.cc
```diff
@@ -674,4 +674,289 @@ pointer_offset_expression (tree base_tree, tree index_tree, location_t location)
...
+		  && !TREE_STATIC (TREE_OPERAND (*tp, 0)))))
+	{
+	  tree decl = TREE_OPERAND (*tp, 0);
+	  WALK_SUBTREE (DECL_INITIAL (decl));
+	  WALK_SUBTREE (DECL_SIZE (decl));
+	  WALK_SUBTREE (DECL_SIZE_UNIT (decl));
+	}
+      break;
+
+    case LAMBDA_EXPR:
```
philberty, line 907: we don't need to port over lambda expr (yet anyway)

philberty (commented):
This PR has given me alot to think about after a conversation with @dafaust I think the fact your ported over LAMBDA_EXPR is pretty interesting.

I hope you don''t mind letting me look at this for another day. My gut feeling is that we need to remove the stuff for traitsand cp binding levels and namespaces also templates oh and thunk functions.

Leave this PR as is for another day and I will finish the review I just want to think about a few things here since the LAMBDA_EXPR will be really useful for us but I want to know how tied up is that gimplification code with the tsubst stuff for templates.

dafaust:
> * This copies over all the macros in cp-tree.h to rust-tree.h as this makes it easy to port further stuff from cp. This compiles even if the functions and structs inside the unused macros haven't been ported yet. The problem this would leave us with is it will make it difficult to change these to be look relevant to rust rather than c++. I think it would be easy to do that if we ported them one by one. Thoughts?

I agree, I think porting things one-by-one as we have use for them may be better. Skimming through the additions to rust-tree.h there are a lot of macros, struct members, bitfield elements etc. that I'm not sure will be relevant to rust, and I think going back to find and remove them later will be more trouble.

philberty (changes requested):
Hi @abbasfaisal apologies for the delay in response here, I really like how you were able to port a lot of this over. But I think as a rule of thumb we should ignore any code that requires you to port over the extra TREE_CODE's from the CPP front-end it would also include things like the bool is_auto function or CLASS_TEMPLATE etc.

We want to port over the walk tree code because this is part of the code to replace param_decls with the arguments from last I remember walking the constexpr code.

So for this I would expect you can just this branch and just start by removing the is_auto function for example and keep going until you didn't include any extra new TREE_CODEs

abbasfaisal:
@philberty As discussed offline:
- I have removed the TREE_CODEs not in gcc/tree.h which has helped trim this function and remove most of the earlier ported stuff from rust-tree.h so this PR is now more concise.
- It seems use of TREE_CODEs inside cp/cp-tree.def does not cause build error which caused the earlier confusion.

CohenArthur (approved):
I think this is a good start. It will make more sense once you start adding the proper entrypoint and see code evaluated by this function. I think it's good to merge this for now, and it will probably receive some changes soon. Good job :)

dafaust (approved):
Nice I think this looks good :)
Just one thing @abbasfaisal maybe you can update the original comment a bit to reflect the current changes, just because bors stuffs that message into the git log; I think some of the bullets are no longer relevant (for this specific merge at least) and might be a little confusing in the log.

abbasfaisal:
> Nice I think this looks good :) Just one thing @abbasfaisal maybe you can update the original comment a bit to reflect the current changes, just because bors stuffs that message into the git log; I think some of the bullets are no longer relevant (for this specific merge at least) and might be a little confusing in the log.

Done :)

Issue #1261, opened 2022-05-19
Add AST dump visitor


philberty:
Lets create a new visitor to do an AST dump this dump will output any AST item to be as close to real rust formatted code as possible.

So for example we can send in an AST::Function and get back:

```
fn function_name() ... { ... }
```

This visitor should keep a context for indentation, and be used to take over our current AST dump which is abusing as_string which is not the right place to do this as it loses the context information of the indentation levels.

We should give the visitor an output stream  (std::ofstream - yes one of the few times streams in C++ are good) to write the dump to. This lets us give any kind of stream so we are not tied to abusing std::string which will not scale very well.

NOTE this issue will super seed the old: #7

CohenArthur:
Since the goal is to dump valid rust code, we should also think of a way to test that we produce valid rust code. I think this might be something for the [testing project](https://github.com/rust-gcc/testing) as it might be a little hard to do with dejagnu?

CohenArthur:
This is already done within the compiler and simply needs to cover more cases as with go

PR #1287, opened 2022-05-30
Add AST dump visitor

CohenArthur:
This adds a base for pretty printing our AST through a new visitor (Addresses #1261).

This adds enough code to pretty print "functions" without types in the parameters or return type. Looking for feedback on the implementation before we keep going with more code :)

Here's a tiny patch if you want to see the output whenever you're running `-frust-dump-all`:

```diff
diff --git a/gcc/rust/rust-session-manager.cc b/gcc/rust/rust-session-manager.cc
index 0b0e5af915f..3e39c6a6e17 100644
--- a/gcc/rust/rust-session-manager.cc
+++ b/gcc/rust/rust-session-manager.cc
@@ -31,6 +31,7 @@
 #include "rust-cfg-parser.h"
 #include "rust-lint-scan-deadcode.h"
 #include "rust-lint-unused-var.h"
+#include "rust-ast-dump.h"
 
 #include "diagnostic.h"
 #include "input.h"
@@ -1023,6 +1024,8 @@ Session::dump_ast (Parser<Lexer> &parser, AST::Crate &crate) const
 
   parser.debug_dump_ast_output (crate, out);
   out.close ();
+
+  AST::Dump (std::cerr).go (crate);
 }
 
 void
```

output:
```rust
arthur@platypus ~/G/r/gccrs (add-ast-dump-visitor) [1]> cat test.rs
fn foo0(a: i32) {
    let a = 15;
    let b: i32 = 15;
}

fn foo1(a: &i32) {}

struct S;

fn foo2(a: S) {}
fn foo3(a: S) -> i32 {}

arthur@platypus ~/G/r/gccrs (add-ast-dump-visitor)> build/gcc/rust1 test.rs -frust-dump-all
fn foo0(a: ) {
	let a = ;
	let b:  = ;

}

fn foo1(a: ) {

}

fn foo2(a: ) {

}

fn foo3(a: ) -> {

}

test.rs:11:1: error: expected [i32] got [()]
   11 | fn foo3(a: S) -> i32 {}
      | ^                ~

Analyzing compilation unit

 phase setup                        :   0.00 (  0%)   0.00 (  0%)   0.01 (100%)   129k ( 94%)
 TOTAL                              :   0.00          0.00          0.01          137k
```

### DIFF
--- a/gcc/rust/ast/rust-ast-dump.cc
+++ b/gcc/rust/ast/rust-ast-dump.cc
@@ -1,0 +1,648 @@
+// Copyright (C) 2020-2022 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#include "rust-ast-dump.h"
+
+namespace Rust {
+namespace AST {
+
+void
+Dump::visit (Token &tok)
+{}
+
+void
+Dump::visit (DelimTokenTree &delim_tok_tree)
+{}
+
+void
+Dump::visit (AttrInputMetaItemContainer &input)
+{}
+
+void
+Dump::visit (IdentifierExpr &ident_expr)
+{}
+
+void
+Dump::visit (Lifetime &lifetime)
+{}
+
+void
+Dump::visit (LifetimeParam &lifetime_param)
+{}
+
+// rust-path.h
+void
+Dump::visit (PathInExpression &path)
+{}
+
+void
+Dump::visit (TypePathSegment &segment)
+{}
+
+void
+Dump::visit (TypePathSegmentGeneric &segment)
+{}
+
+void
+Dump::visit (TypePathSegmentFunction &segment)
+{}
+
+void
+Dump::visit (TypePath &path)
+{}
+
+void
+Dump::visit (QualifiedPathInExpression &path)
+{}
+
+void
+Dump::visit (QualifiedPathInType &path)
+{}
+
+// rust-expr.h
+void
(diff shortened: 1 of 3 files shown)


gcc/rust/ast/rust-ast-dump.cc
```diff
@@ -0,0 +1,753 @@
...
+Indent::Indent () : tabs (0) {}
+
+std::ostream &
+operator<< (std::ostream &stream, const Indent &indent)
+{
+  for (size_t i = 0; i < indent.tabs; i++)
+    stream << '\t';
+
+  return stream;
+}
```
ndrwnaguib, line 33: I've run a benchmark for appending the tab character using a loop vs a string constructor

```cpp
#include <benchmark/benchmark.h>
#include <ostream>

#define tabs 5e+5

static void BM_AppendUsingStringConstructor(benchmark::State &state) {
  std::ostream stream(NULL);
  for (auto _ : state)
    stream << std::string(tabs, '\t');
}

BENCHMARK(BM_AppendUsingStringConstructor);
//
static void BM_AppendUsingInsertionOperator(benchmark::State &state) {
  std::ostream stream(NULL);
  for (auto _: state)
    for (size_t i = 0; i < tabs; i++)
      stream << '\t';
}
BENCHMARK(BM_AppendUsingInsertionOperator);
//
BENCHMARK_MAIN();
```

Some numbers:

Tabs set to `5e+5`
```stdout
--------------------------------------------------------------------------
Benchmark                                Time             CPU   Iterations
--------------------------------------------------------------------------
BM_AppendUsingStringConstructor      11169 ns        10953 ns        66106
BM_AppendUsingInsertionOperator    5920976 ns      5913260 ns          117
```

Here is a one-liner too

```bash
for tabs in 1e+3, 1e+5, 1e+10, 1e+20, 1e+50; do echo "Tabs set to ${tabs}\n" && sed -i "s/TABS/${tabs}/" benchmark.cc && g++ main.cpp -std=c++11 -isystem benchmark/include -Lbenchmark/build/src -lbenchmark -lpthread -o benchmark && ./benchmark; done
```

--- 

The number of iterations is massively higher for string constructors, but takes much less time however.

dkm: You may want to check how other parts of GCC are doing this same kind of tasks (gimple dump would look really close to this for example, I guess). I know that when looking at dumping the HIR in s-expr, looking at the RTL dumper was useful :)

philberty: Thats a really good point about RTL that will be good for @ndrwnaguib to look at.

ndrwnaguib: thank you for sharing that, I will definitely take a look on the RTL dumper.

PR #1291, opened 2022-06-02
Fix bad impl item overlap check

### LINKED ISSUE #1289: Bad impl overlap error
I tried this code: https://godbolt.org/z/58zdz1ETj

```rust
extern "C" {
    fn printf(s: *const i8, ...);
}

mod intrinsics {
    extern "rust-intrinsic" {
    }
}

#[lang = "mut_ptr"]
impl<T> *mut T {
    pub const unsafe fn offset(self, count: isize) -> *mut T {
        unsafe { intrinsics::offset(self, count) as *mut T }
    }

    }
}

#[lang = "const_ptr"]
impl<T> *const T {
    pub const unsafe fn offset(self, count: isize) -> *mut T {
        unsafe { intrinsics::offset(self, count) as *mut T }
    }

    }
}

fn main() -> i32 {
    let a: *mut _ = &mut 123;
    unsafe {
        let b = a.add(123);
    }

    0
}
```

Instead, this happened: 

```
<source>:13:9: error: duplicate definitions with name offset
   13 |     pub const unsafe fn offset(self, count: isize) -> *mut T {
      |         ^
......
   24 |     pub const unsafe fn offset(self, count: isize) -> *mut T {
      |         ~
<source>:17:9: error: duplicate definitions with name add
   17 |     pub const unsafe fn add(self, count: usize) -> Self {
      |         ^
......
   28 |     pub const unsafe fn add(self, count: usize) -> Self {
      |         ~
```

### Meta

- What version of Rust GCC were you using, git sha if possible. 7a94948e65db2fef6fd93c94e70ead168b71513d

philberty:
This issue here was the equality checks for reference
and pointer types did not check the mutability.
    
Fixes #1289

### DIFF
--- a/gcc/rust/typecheck/rust-tyty.cc
+++ b/gcc/rust/typecheck/rust-tyty.cc
@@ -2441,6 +2441,9 @@ ReferenceType::is_equal (const BaseType &other) const
     return false;
 
   auto other2 = static_cast<const ReferenceType &> (other);
+  if (mutability () != other2.mutability ())
+    return false;
+
   return get_base ()->is_equal (*other2.get_base ());
 }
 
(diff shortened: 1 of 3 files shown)


dafaust (approved):
Good catch! LGTM

Issue #869, opened 2022-01-10
Support for range loops


philberty:
I tried this code:

```rust
fn test()  {
  for i in 0..10 {
      
  }
}
```

I expected to see this happen: compile without error
Instead, this happened: 

```
<source>:2:3: fatal error: Failed to lower expr: [ForLoopExpr: 
 outer attributes: none
 Label: none
 Pattern: i
 Iterator expr: 0..10
 Loop block: BlockExpr:

  outer attributes: none
  inner attributes: none
 statements: none
 final expression: none
]
    2 |   for i in 0..10 {
      |   ^
compilation terminated.
Compiler returned: 1
```

### Meta

- What version of Rust GCC were you using, git sha if possible.

philberty:
Looks as though we need quite alot of libcore here for loops:

```rust
mod intrinsics {
    extern "rust-intrinsic" {
        pub fn wrapping_add<T>(a: T, b: T) -> T;
        pub fn rotate_left<T>(a: T, b: T) -> T;
        pub fn rotate_right<T>(a: T, b: T) -> T;
        pub fn offset<T>(ptr: *const T, count: isize) -> *const T;
    }
}

mod mem {
    extern "rust-intrinsic" {
        fn transmute<T, U>(_: T) -> U;
        fn size_of<T>() -> usize;
    }
}

macro_rules! impl_uint {
    ($($ty:ident = $lang:literal),*) => {
        $(
            impl $ty {
                pub fn wrapping_add(self, rhs: Self) -> Self {
                    // intrinsics::wrapping_add(self, rhs)
                    self + rhs
                }

                pub fn rotate_left(self, n: u32) -> Self {
                    unsafe {
                        intrinsics::rotate_left(self, n as Self)
                    }
                }

                pub fn rotate_right(self, n: u32) -> Self {
                    unsafe {
                        intrinsics::rotate_right(self, n as Self)
                    }
                }

                pub fn to_le(self) -> Self {
                    #[cfg(target_endian = "little")]
                    {
                        self
                    }
                }

                pub const fn from_le_bytes(bytes: [u8; mem::size_of::<Self>()]) -> Self {
                    Self::from_le(Self::from_ne_bytes(bytes))
                }

                pub const fn from_le(x: Self) -> Self {
                    #[cfg(target_endian = "little")]
                    {
                        x
                    }
                }

                pub const fn from_ne_bytes(bytes: [u8; mem::size_of::<Self>()]) -> Self {
[...]

PR #1292, opened 2022-06-02
Add name resolution to for loops

philberty:
This just adds basic name resolution and hir lowering for
the loop. Eventually we will make all Loops into a single
HIR::LoopExpr but for now its really useful having a separate
visitor to implement this to avoid regressions in the short
term.

Addresses #869


CohenArthur (approved):
LGTM

PR #1293, opened 2022-06-02
Fixup name canonicalization for impl blocks

philberty:
When we generate the path for impl items we need to base this of the Self
type but this was ignoring cases like pointers, references or slices. This
meant generic slices had the same path has generic pointers etc. The only
reason we didn't end up with a linker symbol clash is due to the symbol
hash.


CohenArthur (approved):
LGTM

Issue #1227, opened 2022-05-09
Resolve "special" paths properly


CohenArthur:
- [x] `crate` path
- [x] `self` path
- [x] `super` path

PR #1294, opened 2022-06-02
Add new mappings for items within a module

philberty:
This patch adds two new interfaces to the mappings class.

  - Lookup the child items within a module
  - Lookup the parent module id from the child item id

Each of these API's are going to be required so that we can resolve two new
types of path segment:

  super::foo::bar

Where we need to be able to lookup the super module node id from the
current scope node id. To then lookup the children within that scope. The
other path is simpler such as: crate::foo::bar. Where we lookup the items
within the current Crate NodeId and follow on as normal.

Addresses #1227


CohenArthur (approved):
LGTM!

PR #1295, opened 2022-06-02
AST Dump impl traits

CohenArthur:
This adds proper AST formatting for inherent impl blocks and traits.

This does not handle trait impls yet (`impl Trait for Type`) but it should be really easy to refactor and add.

### DIFF
--- a/gcc/rust/ast/rust-ast-dump.cc
+++ b/gcc/rust/ast/rust-ast-dump.cc
@@ -515,24 +515,90 @@ Dump::visit (StaticItem &static_item)
 {}
 
 void
+Dump::format_function_common (std::unique_ptr<Type> &return_type,
+			      std::unique_ptr<BlockExpr> &block)
+{
+  if (return_type)
+    {
+      stream << "-> ";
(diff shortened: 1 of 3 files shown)


gcc/rust/ast/rust-ast-dump.cc
```diff
@@ -728,19 +847,28 @@ Dump::visit (RawPointerType &type)
...
+{
+  type.get_elem_type ()->accept_vis (*this);
+}
 
 void
 Dump::visit (InferredType &type)
-{}
+{
+  // FIXME: Is that always valid?
+  stream << "_";
```
philberty, line 870: What do you mean? Its fairly common to see it being used in interesting ways in the libcore code

philberty (approved):
nice job

gcc/rust/ast/rust-ast-dump.cc
CohenArthur: Just wondering if there are any other ways to format inferred types, but no. I'll remove that comment :)

CohenArthur:
bors cancel

PR #1296, opened 2022-06-02
Ast dump trait impl

CohenArthur:
Needs #1295

This adds the dumping of trait implementations for types. However, it is incomplete as I don't believe we can yet handle associated types and associated constants properly? Am I looking in the wrong place?

Will dig deeper, but opening for any early review.

### DIFF
--- a/gcc/rust/ast/rust-ast-dump.cc
+++ b/gcc/rust/ast/rust-ast-dump.cc
@@ -624,7 +624,21 @@ Dump::visit (InherentImpl &impl)
 
 void
 Dump::visit (TraitImpl &impl)
-{}
+{
+  stream << "impl ";
+  impl.get_trait_path ().accept_vis (*this);
+  stream << " for ";
+  impl.get_type ()->accept_vis (*this);
+
+  stream << " {\n";
+  indentation.increment ();
+
+  for (auto &item : impl.get_impl_items ())
+    item->accept_vis (*this);
+
+  indentation.decrement ();
+  stream << "\n}\n";
+}
 
 void
 Dump::visit (ExternalStaticItem &item)


gcc/rust/ast/rust-ast-dump.cc
```diff
@@ -364,7 +366,10 @@ Dump::visit (AsyncBlockExpr &expr)
 // rust-item.h
 void
 Dump::visit (TypeParam &param)
-{}
+{
+  // Is it possible to have a null type here?
+  param.get_type ()->accept_vis (*this);
```
philberty, line 371: I need to double-check but off the top of my head there is a case it can be null but it might be a different data structure. I am thinking about the case of:



```
impl<T=i32> Foo<T> {}
```

philberty, line 371: I need to double-check but off the top of my head there is a case it can be null but it might be a different data structure. I am thinking about the case of:



```
impl<T=i32> Foo<T> {}
```

CohenArthur: I think that is a different data structure yeah. I'll make sure before merging this

CohenArthur: Yep, this counts as an inherent impl and is covered by #1295

CohenArthur:
Unsure on how I closed this. Reopening

CohenArthur:
bors cancel
Ooops thought reviews were done :) Sorry!

jdupak:
@CohenArthur shouldn't there be `indentation` at the beginning? E.g. (once module dump is implemented)
```rust
mod test_indent {
    trait Test1 {}

    type X = (i32, i32);

    impl Test1 for X {}
}
```

jdupak:
> @CohenArthur shouldn't there be `indentation` at the beginning? E.g. (once module dump is implemented)
> 
> ```rust
> mod test_indent {
>     trait Test1 {}
> 
>     type X = (i32, i32);
> 
>     impl Test1 for X {}
> }
> ```

Never mind. I see you add it inside the module dump logic.

Issue #1195, opened 2022-04-28
`transmute` has two restrictions I'm not sure you have added here:


philberty:
`transmute` has two restrictions I'm not sure you have added here:

1. Input and output size must be identical.
2. Input and output size must be statically known before monomorphization so eg `transmute::<T, U>` is not allowed, but eg `transmute::<PhantomData<T>, PhantomData<U>>` is allowed. Note that types like `[T; 0]` are not considered to be of a fixed size for the purpose of this check.

_Originally posted by @bjorn3 in https://github.com/Rust-GCC/gccrs/issues/1194#issuecomment-1112401967_

PR #1298, opened 2022-06-07
rust/intrinsic: add a basic size check for `transmute`

liushuyu:
- rust/intrinsic: add a basic size check for `transmute`

partially addresses #1195

### DIFF
--- a/gcc/rust/backend/rust-compile-intrinsic.cc
+++ b/gcc/rust/backend/rust-compile-intrinsic.cc
@@ -20,6 +20,8 @@
 #include "rust-compile-context.h"
 #include "rust-compile-type.h"
 #include "rust-compile-fnparam.h"
+#include "rust-diagnostics.h"
+#include "rust-location.h"
 #include "rust-tree.h"
 #include "tree-core.h"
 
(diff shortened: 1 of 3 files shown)


gcc/rust/backend/rust-compile-intrinsic.cc
```diff
@@ -500,6 +504,34 @@ transmute_intrinsic_handler (Context *ctx, TyTy::BaseType *fntype_tyty)
...
+    = target_size_expr ? TREE_INT_CST_LOW (target_size_expr) : 0;
+  unsigned HOST_WIDE_INT source_size
+    = source_size_expr ? TREE_INT_CST_LOW (source_size_expr) : 0;
+
+  // size check for concrete types
+  // TODO(liushuyu): check alignment for pointers; check for dependently-sized
+  // types
+  if (target_size != source_size)
+    {
+      rust_error_at (fntype->get_locus (),
```
liushuyu, line 524: I am not sure how to get the original call-site location from the function declaration here.

philberty: Ah yes good point there is no mechanism for this here. Let me take a look to see what might be the best approach for this and I will raise an issue with a guide to add this for someone to implement.

philberty: Lets raise this as a seperate issue to get the call site info as this PR looks good to go now

philberty (approved):
Nice work, I've raised https://github.com/Rust-GCC/gccrs/issues/1304 for the missing call site location info I will update that issue where we can wire that up.

PR #1299, opened 2022-06-07
rust/lex: skip broken string expression

### LINKED ISSUE #1297: infinite loop is parsing rustc testcase
I tried this code:

https://github.com/rust-lang/rust/blob/6609c6734de4df43e24d7672f8ae8786ecc8047e/src/test/ui/rustdoc/check-doc-alias-attr.rs

I expected to see this happen: error out appropriately like the test case shows

Instead, this happened: infinite loop

### Meta

- What version of Rust GCC were you using, git sha if possible. 957914b4284213b1d72c9c4210892367acaf1419

liushuyu:
- rust/lex: skip broken string expression and unstuck the parser when it could not advance

should fix #1297

### DIFF
--- a/gcc/rust/lex/rust-lex.cc
+++ b/gcc/rust/lex/rust-lex.cc
@@ -1856,6 +1856,35 @@ Lexer::parse_raw_identifier (Location loc)
     }
 }
 
+// skip broken string input (unterminated strings)
+void
+Lexer::skip_broken_string_input (int current_char)
+{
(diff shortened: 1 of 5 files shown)


CohenArthur (approved):
I think this is a good idea! There are other parts of the parser where this kind of stuff must be done better as well, so this is great :D Thank you!

philberty (approved):
Nice work!

PR #1300, opened 2022-06-08
docker: Fix GCCRS_BUILD info

CohenArthur:
Sometimes, the GCCRS_BUILD file would end up empty. I believe this is
due to how Dockerfiles work in that there was a "race condition" between
the two `RUN` commands. Hopefully this should fix it.


philberty (approved):
LGTM

Issue #1230, opened 2022-05-09
Handle `super::super::super...` paths


CohenArthur:
We should be able to recursively fetch a module through "super" paths

Issue #1153, opened 2022-04-22
Handle `crate` sugar for visibilities


CohenArthur:
`pub(crate)` and `crate` are specifiers for the same visibility, with `crate` being feature gated (https://doc.rust-lang.org/beta/unstable-book/language-features/crate-visibility-modifier.html)f

PR #1307, opened 2022-06-10
Handle super and crate in path resolution

### LINKED ISSUE #1251: Handle complex root paths
```rust
mod a {
    pub mod b {
        pub mod a {
            pub fn foo() {}
        }
    }

    pub fn bidule() {
        crate::a::b::a::foo()
    }
}

fn main() {}
```

output:
```rust
<source>:11:9: error: Cannot find path 'crate' in this scope
   11 |         crate::a::b::a::foo()
      |         ^
```

philberty:
This patch completely reimplements our name resolution process for Paths in
general. This patch gets rid of the old Resolver::Definition structures
which were used so that we can map patterns back to LetStmts and establish
an hierarchy of nodes. This was not nessecary and complicated name/type
resolution.

TypePaths and PathInExpression are similar but have a slightl tweak in the
order they lookup the relevant scopes for types. But overall work the same.
There are several cases of paths you must consider in type resolution:

- i32 (simple type path)
- Self::A (associated type reference)
- mod::foo::impl_item() (reference to impl item)
- super::foo (reference to item)
- crate::foo
- T::bound()

In name resolution we cannot always fully resolve a path but have to rely
on the type-resolution to fully resolve a path as it may require careful
thought. For example a path like:

  mod::foo::item()

might be for a generic foo<T>(), which might have two specialized impl
blocks so the best the name resolution can do is resolve mod::foo then
leave it up to the type resolution to figure out which item this is. We
might have i32 which is just a simple lookup.

Apart from that this patch introduces a module parent child hierachy so
that on paths such as super::foo we resolve super to be our parent module
scope then foo can be resolved with the lookup in the items for that
module.

More over this patch gets rid of some but not all of the awkward name
canonicalization to try and patch paths directly. More cleanup is still
needed here to make the name resolution step easier to read. This was
notable in the Qualified path handling where we can now rely on the type
resolver to setup the associated types properly rather than the name
resolver requiring us to resolve this directly.

Fixes #1251 #1230
Addresses #1227 #1153

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -350,8 +350,16 @@ public:
   Location get_locus () const { return locus; }
   NodeId get_node_id () const { return node_id; }
   const std::string &get_segment_name () const { return segment_name; }
-
-  // TODO: visitor pattern?
+  bool is_super_path_seg () const
+  {
+    return as_string ().compare ("super") == 0;
+  }
+  bool is_crate_path_seg () const
+  {
+    return as_string ().compare ("crate") == 0;
+  }
+  bool is_lower_self () const { return as_string ().compare ("self") == 0; }
+  bool is_big_self () const { return as_string ().compare ("Self") == 0; }
 };
 
 // A simple path without generic or type arguments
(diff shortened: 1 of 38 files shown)


CohenArthur (approved):
LGTM :) Great stuff. Everything is clearer and easy to follow. I'm glad you removed the Definitions that we needed to carry everywhere!

Issue #1229, opened 2022-05-09
32bit build warning


philberty:
```
g++  -fno-PIE -c  -I ../../gccrs/gcc/rust -I ../../gccrs/gcc/rust/lex -I ../../gccrs/gcc/rust/parse -I ../../gccrs/gcc/rust/ast -I ../../gccrs/gcc/rust/analysis -I ../../gccrs/gcc/rust/backend -I ../../gccrs/gcc/rust/expand -I ../../gccrs/gcc/rust/hir/tree -I ../../gccrs/gcc/rust/hir -I ../../gccrs/gcc/rust/resolve -I ../../gccrs/gcc/rust/util -I ../../gccrs/gcc/rust/typecheck -I ../../gccrs/gcc/rust/privacy -I ../../gccrs/gcc/rust/lint -I ../../gccrs/gcc/rust/util -DIN_GCC_FRONTEND -g -O0     -DIN_GCC     -fno-exceptions -fno-rtti -fasynchronous-unwind-tables -W -Wall -Wno-narrowing -Wwrite-strings -Wcast-qual -Wmissing-format-attribute -Woverloaded-virtual -pedantic -Wno-long-long -Wno-variadic-macros -Wno-overlength-strings   -DHAVE_CONFIG_H -I. -Irust -I../../gccrs/gcc -I../../gccrs/gcc/rust -I../../gccrs/gcc/../include -I../../gccrs/gcc/../libcpp/include -I../../gccrs/gcc/../libcody  -I../../gccrs/gcc/../libdecnumber -I../../gccrs/gcc/../libdecnumber/bid -I../libdecnumber -I../../gccrs/gcc/../libbacktrace   -o rust/rust-session-manager.o -MT rust/rust-session-manager.o -MMD -MP -MF rust/.deps/rust-session-manager.TPo ../../gccrs/gcc/rust/rust-session-manager.cc
../../gccrs/gcc/rust/rust-session-manager.cc: In function ‘bool Rust::validate_crate_name(const string&, Rust::Error&)’:
../../gccrs/gcc/rust/rust-session-manager.cc:102:22: warning: format ‘%ld’ expects argument of type ‘long int’, but argument 4 has type ‘size_t {aka unsigned int}’ [-Wformat=]
        kMaxNameLength);
                      ^

[...]

PR #1308, opened 2022-06-12
Fix formatting error on 32-bits targets

dkm:
Printing size_t as [unsigned] long (%ld or %lu) raises warnings on 32-bits
targets. As the GCC pretty printer doesn't have the equivalent of libc's %z/%zu,
fix all formats to use unsigned long and cast values.

refs #1229

Co-authored-by: Rainer Orth <ro@gcc.gnu.org>

### DIFF
--- a/gcc/rust/expand/rust-macro-expand.cc
+++ b/gcc/rust/expand/rust-macro-expand.cc
@@ -707,10 +707,12 @@ MacroExpander::match_repetition (Parser<MacroInvocLexer> &parser,
     rust_error_at (rep.get_match_locus (),
 		   "invalid amount of matches for macro invocation. Expected "
 		   "between %s and %s, got %lu",
-		   lo_str.c_str (), hi_str.c_str (), match_amount);
+		   lo_str.c_str (), hi_str.c_str (),
+		   (unsigned long) match_amount);
 
   rust_debug_loc (rep.get_match_locus (), "%s matched %lu times",
-		  res ? "successfully" : "unsuccessfully", match_amount);
+		  res ? "successfully" : "unsuccessfully",
+		  (unsigned long) match_amount);
 
   // We have to handle zero fragments differently: They will not have been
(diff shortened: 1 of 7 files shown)


tschwinge:
This raises a lot of `log_warnings` (#1026) diagnostics -- so that's working, good!  ;-P
Instead of plain `%z`, you want `%zu` for `size_t`, or `%zi`/`%zd` for `ssize_t`.

---

Though, I see almost no `z` length modifier usage in GCC.  Are there compatibility concerns?  (Still with GCC requiring C++11 nowadays?)  (For avoidance of doubt: I'm all in favor of using `z` if permissible.)

dkm:
Woops of course should have used `zu` :)
I'll dig a bit to see if there's more to its absence than simply force of habit :)

CohenArthur:
I believe the `z` modifier was added for C99 so we should be good on the compatibility part?

The man page for printf states the following:

> glibc 2.1 adds length modifiers hh, j, t, and z and conversion characters a and A.

and 2.1 was released waaaaay before GCC 4.8 / C++11, so I assume we should be in the clear?

I think this is definitely something that should go in

CohenArthur (approved):
LGTM :)

dkm:
Reverted back to cast as %z spec not supported by GCC pretty printers

CohenArthur (approved):
Ah well that's a bit sad :) So no one is interested in patching `pp_format` and submitting it? :P

dkm:
> Ah well that's a bit sad :) So no one is interested in patching `pp_format` and submitting it? :P

Thought about that, but that was before looking at the code :) I'm not sure that's really worth the effort...

dafaust (approved):
LGTM. Bit of a shame about %z but I won't blame anyone for not wanting to patch that :P

dkm:
Thanks for merging it, forgot about it sorry !

PR #1309, opened 2022-06-13
Fix lexing of empty comments continuing till next line

### LINKED ISSUE #1306: Empty comment before `match` causes parse error
I tried this code:

```rust
fn foo (x: i8) -> i32 {
    //
    match x {
        1 => { return 1; }
        _ => { return 0; }
    }
}
```

I expected to see this happen: Compiles without error.

Instead, this happened: 
```
bug.rs:3:11: error: expecting ‘}’ but ‘=>’ found
    3 |     match x {
      |           ^
bug.rs:3:11: error: error may be from having an expression (as opposed to statement) in the body of the function but not last
bug.rs:6:1: error: unrecognised token ‘}’ for start of item
    6 |     }
      | ^
bug.rs:6:1: error: failed to parse item in crate
```

Compiles fine without the empty comment line.

### Meta

- What version of Rust GCC were you using, git sha if possible: e6d32d59f1e6524df667262c509910e44151bbf2

nirmal-j-patel:
Empty comments (comments without any characters including spaces after //) had a bug during lexing. The lexer did not recheck the current character after skipping / characters. When there was no character after //, the lexer skipped the next newline character. This caused lexer to count the next line as a part of the comment to. This commit fixes this bug by rechecking current character after skipping two / characters.

Fixes #1306

Signed-off-by: Nirmal Patel <nirmal@nirmal.dev>

### DIFF
--- a/gcc/rust/lex/rust-lex.cc
+++ b/gcc/rust/lex/rust-lex.cc
@@ -489,6 +489,7 @@ Lexer::build_token ()
 	      // (but not an inner or outer doc comment)
 	      skip_input ();
 	      current_column += 2;
+	      current_char = peek_input ();
 
 	      // basically ignore until line finishes
 	      while (current_char != '\n' && current_char != EOF)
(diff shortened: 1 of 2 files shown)


CohenArthur (approved):
Good find :D Nice little bugfix and thank you for adding the test case.

PR #1312, opened 2022-06-14
Refactor generic parameter parsing and report order errors

### LINKED ISSUE #1311: Lifetimes parameters after generic types do not get parsed properly
I tried this code:

```rust
struct Foo<T, U, 'a, 'b, W, 'c, V>;
```

I expected to see this happen: No parsing error. `rustc` still errors out, but with a message about the order of parameters

Instead, this happened: Parsing error about how `'a` could not be parsed as a type parameter

### Meta

- What version of Rust GCC were you using: 1c679d7f42834a72956d6fb39c8d298dd6dc4b4e

CohenArthur:
Closes #1311

This allows us to parse lifetimes and types (and later const generics)
in any order without necessarily erroring out for the wrong reason. It also simplifies the
code greatly and makes it easier to modify.

This also removes an unused duplicate `parse_generic_params` function


dafaust (approved):
Very nice refactor, LGTM!

PR #1315, opened 2022-06-15
Add base for parsing const generic application

CohenArthur:
1. Refactor const generic declaration

The default value for const generics now benefits from using the same
function as parsing a regular const generic expression

2. `Parser::parse_type` should not always add errors

In the case that we are parsing a const generic and not a type, we
should not emit bogus errors from `parse_type` such as "unexpected token
in type: LITERAL". Thus, we add a flag to the function to not always
add errors to the error table

3. Figure out how to deal with ambiguities

In the following cases, parsing is ambiguous:

```rust
let a: Foo<N>;
```

What is N? Is it a type to be used as a generic argument? Is it a const
value to be used for a const generic argument? We need to keep both
possibilities and differentiate later during typechecking. We need to
figure out if it would be better to keep the ambiguity in our future
`ConstGenericArg` type (something like Kind::ConstVarOrType) or modify
our current `AST::Type` to maybe get differentiated later as a const
variable, which seems more annoying.

Finally, since the const evaluation is not implemented yet, we are
getting some bogus errors in the testcase. This commit simply serves as
a necessary base: parsing const generics before we can apply them.

### DIFF
--- a/gcc/rust/ast/rust-path.h
+++ b/gcc/rust/ast/rust-path.h
@@ -134,6 +134,8 @@ struct GenericArgs
   std::vector<Lifetime> lifetime_args;
   std::vector<std::unique_ptr<Type> > type_args;
   std::vector<GenericArgsBinding> binding_args;
+  // TODO: Handle const generics here as well.
+  // We can probably keep a vector of `Expr`s for this.
   Location locus;
 
 public:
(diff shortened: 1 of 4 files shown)


dafaust (approved):
LGTM

PR #1316, opened 2022-06-16
Add const generic declaration to AST

CohenArthur:
This uses a number of hacks and workarounds to avoid implementing the equivalent HIR node in the same PR. I'll raise issues for each of the `FIXME`s in the code. Needs one commit from #1315 to avoid unecessary conflicts.


dafaust (approved):
LGTM

PR #1317, opened 2022-06-16
AST for const generic arguments/const application

CohenArthur:
This commit adds a new type to the AST which contains const generic
arguments. The `GenericArgs` structure now also keeps a vector of them,
and the parser was modified accordingly. We need to handle these extra
arguments in the various resolvers and type-checking phase.

As pointed out before, const generic arguments are ambiguous at the AST
level. Which is why this "sum type" contains a "Clear" variant when
there is no ambiguity at all (`<5>`, `<{ 5 }>`, `<{ 5 + 15 }>` or
`<{ N }>`) and an "Ambiguous" one if the value in question might be
referring to a type argument (`<N>`) instead of a const generic
argument.

### DIFF
--- a/gcc/rust/ast/rust-ast-full-test.cc
+++ b/gcc/rust/ast/rust-ast-full-test.cc
@@ -2609,6 +2609,20 @@ GenericArgs::as_string () const
 	}
     }
 
+  // const args
+  if (!const_args.empty ())
+    {
+      auto i = const_args.begin ();
+      auto e = const_args.end ();
+
+      for (; i != e; i++)
+	{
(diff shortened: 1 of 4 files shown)


dafaust (approved):
Hm... I'm not sure I quite understand why you split up #1315, #1316 and this one into three separate PRs. IMO they could all be one with the commit messages speaking for the different components of changes. Took me a minute to realize they all share 0dbfdb5

But that's not an objection! The changes themselves LGTM (both on the individual PRs and in sum). So please merge the 3 of them separately or however you'd like to do it. Good stuff :+1:

PR #1322, opened 2022-06-17
Fixes #1263; diagnostics/feature/wrapper

### LINKED ISSUE #1263: Add `rust_sorry{_at}`
A gcc macro called `sorry_at` exists in the "diagnostics" header. However, since it takes a `location_t` type as first parameter, we cannot directly call it with our location info, which are instances of the `Location` class, which themselves wrap a `location_t`.

This ends up requiring the following code:
```c++
sorry_at (locus.gcc_location(), "fmt");
```

We should add a wrapper around this macro for our frontend specifically, similarly to the existing `rust_error` or `rust_debug_loc` macros. The idea would simply be to take a `Location` as macro and forward the inner gcc location and the remaining arguments.

ndrwnaguib:
* wrapping GCC's diagnostics `sorry_at` with a new rust macro named `rust_sorry_at` which accepts `Location` instead of the default `location_t`.
* refactoring previous usage of `sorry_at` to `rust_sorry_at`

Fixes #1263

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -239,8 +239,8 @@ CompileExpr::visit (HIR::MatchExpr &expr)
     {
       // FIXME: CASE_LABEL_EXPR does not support floating point types.
       // Find another way to compile these.
-      sorry_at (expr.get_locus ().gcc_location (),
-		"match on floating-point types is not yet supported");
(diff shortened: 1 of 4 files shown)


ndrwnaguib:
> Good stuff! Thank you for that :) Can you change your PR's message a little bit so that the `Fixes #1263` is in and the issue gets closed automatically? Otherwise perfect :)

Thanks Arthur, just did. :)

PR #1327, opened 2022-06-20
Lower const generics structures

### LINKED ISSUE #1318: Lower `AST::ConstGenericParam` properly
As of right now, this lowers to a fake lifetime in order to avoid assertion failures later on in the compiler. We need to implement proper lowering to a valid HIR equivalent

CohenArthur:
This PR adds lowering for both the `ConstGenericParam` type and the `ConstGenericArg` one. Ideally, we should be able to avoid ambiguites (`let a: Foo<N>;` -> what is N? A type? A const?) by going through the name resolver, and resolve either to a different `AST::ConstGenericArg` either to a correct `AST::Type` or equivalent. So I think we can avoid doing the same "Ambiguous/Clear" dance in the `HIR::ConstGenericArg` type.

If that ends up not being possible, we can always add it and disambiguate during typechecking or other.

Closes #1318

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -1381,6 +1381,9 @@ public:
   }
 
   bool has_type () { return type != nullptr; }
+  bool has_default_value () { return default_value != nullptr; }
+
+  const Identifier &get_name () const { return name; }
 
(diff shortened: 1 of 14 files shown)


dafaust (approved):
Nice, LGTM
And good idea adding the FIXME markers in typecheck/ :)

PR #1329, opened 2022-06-21
gccrs const folding port: start porting potential_constant_expression…

abbasfaisal:
Card: [Link](https://github.com/Rust-GCC/gccrs/projects/16#card-82300805).

Since [this](https://github.com/Rust-GCC/gccrs/blob/master/gcc/cp/constexpr.cc#L8350) is a huge function, we will start by porting chunks on which it depends as that would reduce the chance of missing something important and be easier to review.

This changeset ports over the below typedefs, structs and functions in addition to the macros:

Typedefs:
 - [tsubst_flags_t](https://github.com/Rust-GCC/gccrs/blob/7eef766dc5a8abda2ca2cf8d535cdf160f40b50c/gcc/cp/cp-tree.h#L5533)

Structs:
 - [lang_type](https://github.com/Rust-GCC/gccrs/blob/master/gcc/cp/cp-tree.h#L2300)
 - [tree_pair_s](https://github.com/Rust-GCC/gccrs/blob/master/gcc/cp/cp-tree.h#L2281)

Functions:
 - [cp_expr_loc_or_loc](https://github.com/Rust-GCC/gccrs/blob/master/gcc/cp/cp-tree.h#L8073)
 - [cp_expr_loc_or_input_loc](https://github.com/Rust-GCC/gccrs/blob/master/gcc/cp/cp-tree.h#L8082)
 - [cp_expr_location](https://github.com/Rust-GCC/gccrs/blob/master/gcc/cp/tree.cc#L6056)
 - [is_really_empty_class](https://github.com/Rust-GCC/gccrs/blob/master/gcc/cp/class.cc#L8867)
 - [is_empty_class](https://github.com/Rust-GCC/gccrs/blob/master/gcc/cp/class.cc#L8850)
 - [array_type_nelts_top](https://github.com/Rust-GCC/gccrs/blob/master/gcc/cp/tree.cc#L3055)

I feel we do not need [is_really_empty_class](https://github.com/abbasfaisal/gccrs/blob/port-potential-const-expression-1-part-by-part/gcc/rust/backend/rust-tree.cc#L786) and [is_empty_class](https://github.com/abbasfaisal/gccrs/blob/port-potential-const-expression-1-part-by-part/gcc/rust/backend/rust-tree.cc#L827) as they are concerned with classes, record types and unions. The tree codes they surround come from [tree.def](https://github.com/Rust-GCC/gccrs/blob/master/gcc/tree.def). I think the class and record tree codes are not relevant to Rust. Feel free to say so and I will remove these functions or if you would like me to dig more into them.

### DIFF
--- a/gcc/rust/backend/rust-tree.cc
+++ b/gcc/rust/backend/rust-tree.cc
@@ -760,4 +760,92 @@ out:
 #undef WALK_SUBTREE
 }
 
+// forked from gcc/cp/tree.cc cp_expr_location
+
+/* Like EXPR_LOCATION, but also handle some tcc_exceptional that have
+   locations.  */
+
+location_t
+cp_expr_location (const_tree t_)
+{
+  tree t = CONST_CAST_TREE (t_);
+  if (t == NULL_TREE)
+    return UNKNOWN_LOCATION;
+
+  return EXPR_LOCATION (t);
+}
+
+// forked from gcc/cp/class.cc
+
+/* Returns true if TYPE contains no actual data, just various
+   possible combinations of empty classes.  If IGNORE_VPTR is true,
+   a vptr doesn't prevent the class from being considered empty.  Typically
+   we want to ignore the vptr on assignment, and not on initialization.  */
+
+bool
+is_really_empty_class (tree type, bool ignore_vptr)
+{
+  if (CLASS_TYPE_P (type))
+    {
+      tree field;
+      tree binfo;
+      tree base_binfo;
+      int i;
+
+      /* CLASSTYPE_EMPTY_P isn't set properly until the class is actually laid
+	 out, but we'd like to be able to check this before then.  */
+      if (COMPLETE_TYPE_P (type) && is_empty_class (type))
+	return true;
+
+      if (!ignore_vptr && TYPE_CONTAINS_VPTR_P (type))
+	return false;
+
+      for (binfo = TYPE_BINFO (type), i = 0;
+	   BINFO_BASE_ITERATE (binfo, i, base_binfo); ++i)
+	if (!is_really_empty_class (BINFO_TYPE (base_binfo), ignore_vptr))
+	  return false;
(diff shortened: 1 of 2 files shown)


CohenArthur (approved):
Great job getting that in :) I think this PR, like your previous one, will make a lot more sense once you start actually calling into the functions later on.

> I feel we do not need [is_really_empty_class](https://github.com/abbasfaisal/gccrs/blob/port-potential-const-expression-1-part-by-part/gcc/rust/backend/rust-tree.cc#L786) and [is_empty_class](https://github.com/abbasfaisal/gccrs/blob/port-potential-const-expression-1-part-by-part/gcc/rust/backend/rust-tree.cc#L827) as they are concerned with classes, record types and unions. The tree codes they surround come from [tree.def](https://github.com/Rust-GCC/gccrs/blob/master/gcc/tree.def). I think the class and record tree codes are not relevant to Rust. Feel free to say so and I will remove these functions or if you would like me to dig more into them.

Since the goal is ultimately to port over the C++ constant expression folder, I wouldn't worry too much about having *some* unused stuff merged in. We can always weed it out later on. Plus, we might even think about simply linking against the C++ const folder in the future, so we will have extra stuff anyway :)

I think we should get this merged and used as soon as possible :D Great job!

dafaust (approved):
LGTM :+1: 

> Plus, we might even think about simply linking against the C++ const folder in the future, so we will have extra stuff anyway :)

This is an interesting idea, I'd be curious to see if/how it works in practice. I was actually wondering whether some of the ported-but-not-renamed functions in this PR will give link errors if we try to build a gcc with both the rust and c++ frontends enabled. Definitely not a blocker for this to be merged!

It might turn out that if we want to link against the C++ const folder directly we'd need to factor it out of the C++ frontend to somewhere common, like how C and C++ share some functionality in the c-family directory. I'm not very familiar with the build processes for the front ends though so maybe these are issues long since solved :P I need to read up.

PR #1330, opened 2022-06-23
Add location to HIR::Type base class and improve privacy error reporting

### LINKED ISSUE #1256: Add location to `HIR::Type` base class

graphman65:
Fix #1256

Adds location to the HIR::Type base class, removes it from all the subclasses and then use it  in the `check_type_privacy` method to improve privacy error reporting.

The privacy error now looks like this (instead of what you can see in the example #1256).

```rust
gcc/testsuite/rust/compile/privacy5.rs:15:20: error: definition is private in this context
   15 |         fn any(a0: green::Foo, a1: green::Bar) {}
      |                    ^
```

### DIFF
--- a/gcc/rust/hir/tree/rust-hir-path.h
+++ b/gcc/rust/hir/tree/rust-hir-path.h
@@ -623,7 +623,6 @@ class TypePath : public TypeNoBounds
 public:
   bool has_opening_scope_resolution;
   std::vector<std::unique_ptr<TypePathSegment> > segments;
-  Location locus;
 
 protected:
   /* Use covariance to implement clone function as returning this object rather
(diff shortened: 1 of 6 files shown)


graphman65:
I couldn't find a way to make `dg-error` check that the error is at the correct location (not just line but also character position) so couldn't really add a test case for this 😞

tschwinge:
You may use 'dg-begin-multiline-output'/'dg-end-multiline-output' ('gcc/testsuite/lib/multiline.exp') to verify verbatim text, or use syntax like '{ dg-error "31:lvalue required as unary" }' to require a diagnostic in column 31, for example.
From a quick look, it appears as if both these are not documented in 'gcc/doc/sourcebuild.texi', I'm afraid...  :-|  (Patches welcome, of course.)

graphman65:
Right, Thanks @tschwinge. I updated the PR with the proper test case :)

CohenArthur (approved):
Fantastic stuff, thanks a lot for working on this. This is very clean and concise and achieves the goal perfectly. And thank you @tschwinge for more intricate dejagnu knowledge :D

PR #1332, opened 2022-06-24
hir/add: pretty printing visitor skeleton

ndrwnaguib:
This commit adds a visitor skeleton for pretty printing of the HIR. Also,
modifies the session manager to call the new entry point
`Rust::HIR::Dump::crate` which simply prints the crate the way it used to be.

### DIFF
--- a/gcc/rust/hir/rust-hir-dump.cc
+++ b/gcc/rust/hir/rust-hir-dump.cc
@@ -1,0 +1,446 @@
+// Copyright (C) 2020-2022 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#include "rust-hir-dump.h"
+
+namespace Rust {
+namespace HIR {
+
+Dump::Dump (std::ostream &stream) : stream (stream) {}
+
+void
+Dump::go (HIR::Crate &crate)
+{
+  // TODO: print crate inner_attrs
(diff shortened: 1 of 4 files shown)


gcc/rust/hir/rust-hir-dump.h
```diff
@@ -0,0 +1,191 @@
...
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#ifndef RUST_HIR_DUMP_H
+#define RUST_HIR_DUMP_
```
CohenArthur, line 20:
```suggestion
#define RUST_HIR_DUMP_H
```

gcc/rust/hir/rust-hir-dump.cc
```diff
@@ -0,0 +1,446 @@
...
+{
+  // TODO: print crate inner_attrs
+  // TODO: print crate items
+  // TODO: print crate mappings
+  stream << crate.as_string ();
+}
+
+void
+visit (IdentifierExpr &)
+{}
```
CohenArthur, line 37: ... like so :) 

However, these do not refer to the `Dump` class's methods as you forgot the `Dump::` prefix. So they are simply functions in that file, without any reference to the class.

I would suggest removing all default definitions in the header (so removing the `{}` and turning them into function declarations, with a semicolon) and adding the `Dump::` prefix on all of those here.

Doing it that way also has the advantage of saving on compilation times, since we will only be changing the source file and not the header whenever modifications are made to your visitor.

You can do this quite easily with two regexes and `sed`, or your editor's equivalent :D 

`sed -i 's/{}/;/g' gcc/rust/hir/rust-hir-dump.h`
`sed -i 's/visit/Dump::&/g' gcc/rust/hir/rust-hir-dump.cc`

CohenArthur (commented):
This is a very good start Andrew! Thank you for the work. Don't get discouraged by the multiple comments, they're very minor and easy to fix and it will be in a mergeable state in no-time :) As soon as this is fixed we'll merge this, and then I think we should look into adding another dump option, something like `-frust-dump-hir-experimental` so that we could have both the existing HIR dump and yours. I'll create an issue for that

ndrwnaguib:
@CohenArthur thanks for your comments and suggestions, just rebased the same commit. Also, having a separate command for the experimental HIR dumping is a good idea, will work on that.

PR #1335, opened 2022-06-24
cli: add "hir-pretty" as a "dump" option

### LINKED ISSUE #1333: Add new `-frust-dump-hir-experimental` to use the new `HIR::Dump` interface
- [ ] Add new `kHIRExperimentalDumpFile` (`gccrs.hir-experimental.dump` for example)
- [ ] Add new enum value `CompileOptions::HIR_EXPERIMENTAL_DUMP`
- [ ] Handle "hir-experimental" as a dump option in the `Session::enable_dump` function
- [ ] Add a call into the `HIR::Dump` interface if the option is enabled, right after the call to `dump_hir` in the `Session::parse_file` function

ndrwnaguib:
Fixes #1333 

Instead of mixing the ongoing project of pretty dumping the HIR with the current
implementation, this commit adds a new option "-frust-dump-hir-pretty" which
calls the hir pretty dump visitor class. Closes https://github.com/Rust-GCC/gccrs/issues/1333


gcc/rust/rust-session-manager.cc
```diff
@@ -53,6 +53,7 @@ const char *kLexDumpFile = "gccrs.lex.dump";
 const char *kASTDumpFile = "gccrs.ast.dump";
 const char *kASTExpandedDumpFile = "gccrs.ast-expanded.dump";
 const char *kHIRDumpFile = "gccrs.hir.dump";
+const char *kHIRPrettyDumpFile = "gccrs.hir.pretty.dump";
```
CohenArthur, line 56:
```suggestion
const char *kHIRPrettyDumpFile = "gccrs.hir-pretty.dump";
```
nit but this way it stays inline with the other file options :)

CohenArthur (approved):
Perfect thank you!

PR #1342, opened 2022-06-28
Refactor name-resolution pass

philberty:
This is a patch series which cleans up the code in the name resolution pass. Prior to
#1307 we used to build up canonical path segments to do name resolution of types
and paths, this lead to confusing code in how we handle impl blocks Self type and
in general. This fixes that confusion to make the code more maintainable.


CohenArthur (approved):
Great stuff. This looks much cleaner than before. I'm really happy you managed to get rid of all those `parent` `NodeId`s. Congrats

PR #1343, opened 2022-06-29
Ensure we support generic arguments on mangled type-paths

philberty:
This allows us to have the generic arguments within type paths for mangled
symbols such as those in impl-blocks. This also updates our symbol
managling to mangle comma's to $C$.


CohenArthur (approved):
LGTM. That bit of code is used elsewhere in the compiler, so I think it would be a nice cleanup possibility later on :)

PR #1349, opened 2022-06-30
Add generic argument disambiguation

CohenArthur:
ast: Add disambiguation for generic arguments

This commits adds a new resolver visitor, `ResolveGenericArgs`, which
takes care of differentiating const generics from type generics before
the typechecking phase. This is done solely based on name resolution.

The algorithm is as follows:

is that name referring to a type?
  -> disambiguate to a type
is that name referring to a value?
  -> disambiguate to a const value
else
  -> disambiguate to type

Since types are the default expected behavior, this allows us to report
type errors properly during typechecking.

### DIFF
--- a/gcc/rust/resolve/rust-ast-resolve-path.cc
+++ b/gcc/rust/resolve/rust-ast-resolve-path.cc
@@ -88,6 +88,8 @@ ResolvePath::resolve_path (AST::PathInExpression *expr)
       // resolve any generic args
       if (segment.has_generic_args ())
 	{
+	  auto empty = CanonicalPath::create_empty ();
+	  ResolveGenericArgs::go (segment.get_generic_args (), empty, empty);
 	  ResolveType::type_resolve_generic_args (segment.get_generic_args ());
 	}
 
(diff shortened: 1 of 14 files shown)


gcc/rust/resolve/rust-ast-resolve-type.cc
```diff
@@ -30,11 +30,28 @@ ResolveTypeToCanonicalPath::canonicalize_generic_args (AST::GenericArgs &args)
...
 
   size_t i = 0;
-  size_t total = args.get_type_args ().size ();
+  size_t total = args.get_generic_args ().size ();
 
-  for (auto &ty_arg : args.get_type_args ())
+  for (auto &arg : args.get_generic_args ())
     {
-      buf += ty_arg->as_string ();
+      // FIXME: Arthur: Is that function used solely for pretty-printing/dumping
```
philberty: This is changed in the refactor. There is a kind of similar function but its used to generate the canonical path to something like:

impl<T> Foo<T> { fn test(self) {} }

Where the path is expected to be crate::<impl Foo<T>>::test iirc

gcc/rust/resolve/rust-ast-resolve-path.cc
```diff
@@ -88,6 +88,8 @@ ResolvePath::resolve_path (AST::PathInExpression *expr)
       // resolve any generic args
       if (segment.has_generic_args ())
 	{
+	  auto empty = CanonicalPath::create_empty ();
+	  ResolveGenericArgs::go (segment.get_generic_args (), empty, empty);
```
philberty, line 92: I think ResolveGenericArgs::go should be the main entry point for all of this. I think it should disambiguate then call ResolveType on the arguments etc.

philberty (commented):
What do you think?

CohenArthur:
Closed in favor of multiple PRs with more separation

PR #1352, opened 2022-07-01
HIR: fixup MatchArm assignment operator

dafaust:
We overload the assignment operator to clone the vector of patterns for
the match arm. However the existing patterns were not cleared before
inserting the new ones, so the result was to append the new patterns
onto the assignee arm rather than replace them.


dafaust:
Ran into this while working on tuple match compilation.
A quick fix but boy did the effects have me scratching my head for a minute :laughing:

philberty (approved):
nice one this type of stuff has bit me too with bad copy constructors

PR #1355, opened 2022-07-05
Ambiguous generic arg

CohenArthur:
Needs #1354. Only review the last commit :)

ast: Rename ConstGenericArg -> GenericArg

This makes the class clearer and exposes the fact that it may contain a
type OR a const value clearer.

Because we now run in ambiguous cases later in the compiler pipeline,
this commit also adds horrible hacks to simply ignore ambiguous generic
arguments and treat them as type arguments.

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -2878,12 +2878,12 @@ Parser<ManagedTokenSource>::parse_generic_param (EndTokenPred is_end_token)
 	  return nullptr;
 
 	// optional default value
-	auto default_expr = AST::ConstGenericArg::create_error ();
+	auto default_expr = AST::GenericArg::create_error ();
 	if (lexer.peek_token ()->get_id () == EQUAL)
 	  {
 	    lexer.skip_token ();
 	    auto tok = lexer.peek_token ();
-	    default_expr = parse_const_generic_expression ();
+	    default_expr = parse_generic_arg ();
 
 	    if (default_expr.is_error ())
 	      rust_error_at (tok->get_locus (),
@@ -2894,8 +2894,7 @@ Parser<ManagedTokenSource>::parse_generic_param (EndTokenPred is_end_token)
 
 	    // At this point, we *know* that we are parsing a const
 	    // expression
-	    if (default_expr.get_kind ()
-		== AST::ConstGenericArg::Kind::Ambiguous)
+	    if (default_expr.get_kind () == AST::GenericArg::Kind::Either)
 	      default_expr = default_expr.disambiguate_to_const ();
 	  }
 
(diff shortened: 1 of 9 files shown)


gcc/rust/resolve/rust-ast-resolve-type.cc
```diff
@@ -366,12 +361,35 @@ ResolveTypeToCanonicalPath::visit (AST::TypePath &path)
...
 	    if (s->has_generic_args ())
 	      {
-		for (auto &gt : s->get_generic_args ().get_type_args ())
+		for (auto &generic : s->get_generic_args ().get_generic_args ())
 		  {
-		    CanonicalPath arg = CanonicalPath::create_empty ();
-		    bool ok = ResolveTypeToCanonicalPath::go (gt.get (), arg);
-		    if (ok)
-		      args.push_back (std::move (arg));
+		    // FIXME: What do we want to do here in case there is a
```
philberty, line 366: At this point i would expect that the disambiguation to have occurred and name ResolveType or whatever to actually be called on the items here.

gcc/rust/resolve/rust-ast-resolve-type.cc
```diff
@@ -451,5 +469,22 @@ ResolveTypeToCanonicalPath::ResolveTypeToCanonicalPath ()
   : ResolverBase (), result (CanonicalPath::create_empty ())
 {}
 
+void
+ResolveGenericArgs::go (AST::GenericArgs &args)
+{
+  for (auto &arg : args.get_generic_args ())
+    {
+      // FIXME: Arthur: Ugly hack while waiting for disambiguation
+      if (arg.get_kind () == AST::GenericArg::Kind::Either)
```
philberty, line 478: I think this is the right thing to do here because we want a consistent interface for all generic arg resolution to occur here so I don't thing this is ugly.

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -6265,20 +6277,14 @@ Parser<ManagedTokenSource>::parse_path_generic_args ()
...
-	{
-	  auto const_generic_expr = parse_const_generic_expression ();
-	  if (const_generic_expr.is_error ())
-	    break;
-	  else
-	    const_args.emplace_back (std::move (const_generic_expr));
+	  generic_args.emplace_back (std::move (arg));
 	}
 
+      // FIXME: Do we need to break if we encounter an error?
```
philberty, line 6286: I think keep going because we might be able to recover the parser by reading until the delimiter for the arguments.

philberty (approved):
I think this looks good just a minor bit in the canonicalize part. This function gives us back a path for name mangling so if we have:

```
mod a { struct foo;

impl foo { fn test() .. }

 }
 ```
 
 The impl foo we want to get the full canonical path for this type so when we call that we expect the necessary resolution steps to have already occurred or it will error appropriately so I am not sure your changes there are required.

gcc/rust/resolve/rust-ast-resolve-type.cc
CohenArthur: The hack is that when faced with an amibiguous generic, we always resolve it to a type, as what was done before this patch where ambiguous generics were not handled at all. In #1358, this is where disambiguation happens and we say whether this argument is a type argument or a const value argument

CohenArthur:
> The impl foo we want to get the full canonical path for this type so when we call that we expect the necessary resolution steps to have already occurred or it will error appropriately so I am not sure your changes there are required.

Yes, this is what happens in #1358 (unless I misunderstood something :D). This PR was just to add the new "ambiguous" class but always resolve it as a type, as what was done before

PR #1358, opened 2022-07-06
Disambiguate generic args

CohenArthur:
Needs #1355. Only review the last commit.  The next steps are to typecheck const generics and then lower them to TREE

This removes all the hacks previously introduced to resolve ambiguous
generic args as types, and adds proper disambiguation.

The algorithm is as follows:

is that name referring to a type?
-> disambiguate to a type
is that name referring to a value?
-> disambiguate to a const value
else
-> disambiguate to type

Since types are the default expected behavior, this allows us to report
type errors properly during typechecking.


gcc/rust/expand/rust-attribute-visitor.cc
```diff
@@ -132,20 +132,35 @@ AttrVisitor::expand_generic_args (AST::GenericArgs &args)
...
+	      rust_error_at (expr->get_locus (),
+			     "cannot strip expression in this position");
+	    break;
+	  }
+	default:
+	  break;
+	  // FIXME: Figure out what to do here if there is ambiguity. Since the
+	  // resolver comes after the expansion, we need to figure out a way to
+	  // strip ambiguous values here
+	  // TODO: Arthur: Probably add a `mark_as_strip` method to `GenericArg`
```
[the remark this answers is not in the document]
philberty: It could be based on virtual method to get the attributes to avoid needing to cast.

gcc/rust/resolve/rust-ast-resolve-type.cc
```diff
@@ -492,20 +481,79 @@ ResolveTypeToCanonicalPath::ResolveTypeToCanonicalPath ()
...
 void
-ResolveGenericArgs::go (AST::GenericArgs &args)
+ResolveGenericArgs::disambiguate (AST::GenericArg &arg)
 {
-  for (auto &arg : args.get_generic_args ())
+  auto path = canonical_prefix.append (
+    CanonicalPath::new_seg (UNKNOWN_NODEID, arg.get_path ()));
+
+  auto is_type = is_type_name (path);
+  auto is_value = is_const_value_name (path);
```
philberty, line 509: Is this the cast that we have AST::GenericArg as GenericArg::EITHER? I am just thinking it could be a PathInExpression or TypePath which will need us to call the appropriate ResolvePath or ResolveTypePath. 

This will work for all primitive types and values but not for paths. Lets go with this for now because I think the overall logic is right but we need to look at the GenericArg structure but we can do that in a separate PR

gcc/rust/resolve/rust-ast-resolve-type.cc
```diff
@@ -492,20 +481,79 @@ ResolveTypeToCanonicalPath::ResolveTypeToCanonicalPath ()
...
+    arg = arg.disambiguate_to_const ();
+}
+
+void
+ResolveGenericArgs::resolve_disambiguated_generic (AST::GenericArg &arg)
+{
+  switch (arg.get_kind ())
     {
-      // FIXME: Arthur: Ugly hack while waiting for disambiguation
-      if (arg.get_kind () == AST::GenericArg::Kind::Either)
```
philberty, line 501: I am do we need to call disambiguate_to_type in order to get an AST::Type* ? What if we instead disambiguate by casting to an AST::Type* and then calling ResolveType and if it returns non UNKNOWN_NODEID we know its a type else we try ResolveExpr.

The problem here is that if it is not a Type and it is a const value we need a way to suppress the errors similar to how we reuse the parser.

philberty (approved):

gcc/rust/resolve/rust-ast-resolve-type.cc
CohenArthur: I believe that in rustc this happens at the name-resolution level. So this ends up being the simpler solution and one that should work more closely to how rustc does it. This way, we also avoid having to not emit errors in the type resolver for this specific bit

PR #1362, opened 2022-07-06
extern crate loading 

philberty:
WIP still trying to write dejagnu to enable automated tests

Extern crates statements to tell the front-end to look for another library.
The mechanism here is heavily inspired from gccgo, so when we compile a
library for example we invoke:

```
  gccrs -g -O2 -frust-crate=mylib -c src/lib.rs -o src/mylib.o
```

All going well this object file will now contain extra data inside
.rust-export section inside the object file which will be preserved inside
archives and shared objects. When we have another application which uses
this library 'mylib'.

```rust
  extern crate mylib;
  use mylib::foo;

  fn main() {
    foo();
  }
```

We compile using:

```
gcc -g -O2 -frust-crate=test -c src/main.rs -o src/main.o
```

When the extern crate line is hit the front-end will look for mylib.o,
libmylib.a, mylib.rox. If it finds a raw object file it will read the
.rust-export section directly from the object for the public metadata
such as public functions, types constants etc. If it fails to find an
object it might find .rox which is the objdump of the .rust-export to a
raw file, it might even find libmylib.a and read the export directly out
of the archive file reusing code from gccgo to do so.

The full compiler pipeline is reused here, so the metatadata is actually
just real rust code. The benifit here is that Rust supports exporting,
macros and generics so this requires the name-resolution and type info
all to be generated and inserted into the apropriate context classes. Since
the metadata is real rust code it means we can reuse the full pipeline to
generate the code as nessecary. So for the simple case of a public struct
we simply emit the AST dump of this struct directly into the metadata. If
its a non-generic public function we emit and extern rust abi block for
that function. If its a trait we can simply emit the trait with the public
memebers. Generics are more complicated since we need to emit the function
fully for it to be compiled correctly this still needs tests to be added.
The hardest part is non generic impl blocks which is still a WIP.

To finally link the two crates together you run:

```
gcc -g -O2 -o rust-program.exe src/main.o src/mylib.o
```

Fixes: 1169

### DIFF
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -136,6 +136,10 @@ GRS_OBJS = \
     rust/rust-compile-base.o \
     rust/rust-tree.o \
     rust/rust-compile-context.o \
+    rust/rust-export-metadata.o \
+    rust/rust-imports.o \
+    rust/rust-import-archive.o \
+    rust/rust-extern-crate.o \
     $(END)
 # removed object files from here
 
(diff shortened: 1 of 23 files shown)


philberty:
It seems as though mac does not like dumping data into a custom section. This might need us to just always export a seperate file instead.

PR #1363, opened 2022-07-06
Refactor Lexer to support an abstract InputSource class

### LINKED ISSUE #1000: Fix unsafe lex_string() implementation
As per comments committed in #988, the `lex_string`() function has memory safety issues due to its use of a reference to a C++ string, which can lead to problems if the string's lifetime is shorter than the lexer's.

As the lexer was originally designed only to work with files, the implementation of tokenising from a string currently works by using `fmemopen` to open the string as a file. This is then wrapped in a RAIIFile as required by the lexer's constructor. 

The comments suggest one potential fix - allocating a new buffer with fmemopen and copying the string's data into it. 
However, they also seem to suggest that the current implementation of RAIIFile is not capable of doing so. The exact comment is:
```
// But sadly our RAIIFile does not support moving really well... And the
// destructor, which calls fclose(), gets called, triggering a lack of a
// buffer to parse :)
```
I don’t believe that this should occur. The move constructor of RAIIFile copies the fields and then sets the original’s FILE pointer to nullptr in the constructor body, so when the original is set to be destructed (which would be after the move constructor is called), the `fclose(`) is not called on the stream's actual FILE pointer. The move assignment operator does the same - the only potential issue arising if an RAIIFile is assigned to itself (which should not be able to occur).

An alternative solution that I believe would address the problem better conceptually would be to make the lexer into a template class that takes an “InputSource” type argument. 
`InputSource` is currently an internal wrapper inside the lexer that abstracts sourcing characters (`fgetc` for a FILE).
Instead of wrapping a string to look like a file, there could simply be an alternative `StringInputSource` class that sources characters from a string in addition to the previous `FileInputSource` class. 

The advantage of the template solution is that it is conceptually simple, robust and open to further extension, and would be easy to reason about in terms of lifetime, etc. The disadvantage is that it complicates the lexer somewhat and would require all lexer methods to be put in headers, potentially slowing compilation.

### LINKED ISSUE #1203: error: 'fmemopen' was not declared in this scope
I tried compiling the compiler with msys, targeting ucrt64, and my build failed with the following:

In file included from ../../gccrs/gcc/rust/lex/rust-lex.cc:19:
../../gccrs/gcc/rust/lex/rust-lex.h: In static member function 'static Rust::Lexer Rust::Lexer::lex_string(std::string&)':
../../gccrs/gcc/rust/lex/rust-lex.h:171:24: error: 'fmemopen' was not declared in this scope; did you mean 'freopen'?
  171 |     auto string_file = fmemopen (&input[0], input.length (), "r");
      |                        ^~~~~~~~
      |                        freopen

A quick Google suggests fmemopen doesn't exist on Windows?

philberty:
This patch allows us to remove the fmemopen lex_string hack to support
parsing buffers. This will allow us to support mutliple sources such as
metadata imports etc. The patch here updates the parser to hold onto a
reference to the lexer rather than 'owning' the lexer which allows us to
decouple the move semantics here.

Fixes #1203 
Fixes #1000

### DIFF
--- a/gcc/rust/lex/rust-lex.cc
+++ b/gcc/rust/lex/rust-lex.cc
@@ -116,14 +116,17 @@ is_non_decimal_int_literal_separator (char character)
   return character == 'x' || character == 'o' || character == 'b';
 }
 
-// this compiles fine, so any intellisense saying otherwise is fake news
+Lexer::Lexer (const std::string &input)
+  : input (RAIIFile::create_error ()), current_line (1), current_column (1),
+    line_map (nullptr), raw_input_source (new BufferInputSource (input, 0)),
+    input_queue{*raw_input_source}, token_queue (TokenSource (this))
+{}
+
 Lexer::Lexer (const char *filename, RAIIFile file_input, Linemap *linemap)
   : input (std::move (file_input)), current_line (1), current_column (1),
     line_map (linemap),
-    /*input_source (input.get_raw ()), */
-    input_queue{InputSource (input.get_raw ())},
-    /*token_source (this),*/
-    token_queue (TokenSource (this))
+    raw_input_source (new FileInputSource (input.get_raw ())),
+    input_queue{*raw_input_source}, token_queue (TokenSource (this))
 {
   // inform line_table that file is being entered and is in line 1
   if (linemap)
@@ -138,6 +141,7 @@ Lexer::~Lexer ()
    * mentioned in GCC docs as being useful for "just leaving an included header"
    * and stuff like that, so this line mapping functionality may need fixing.
    * FIXME: find out whether this occurs. */
+
   // line_map->stop();
 }
 
--- a/gcc/rust/lex/rust-lex.h
+++ b/gcc/rust/lex/rust-lex.h
@@ -72,10 +72,14 @@ public:
     return *this;
   }
 
+  static RAIIFile create_error () { return RAIIFile (nullptr, nullptr); }
+
   ~RAIIFile () { close (); }
 
   FILE *get_raw () { return file; }
   const char *get_filename () { return filename; }
+
+  bool ok () const { return file; }
 };
 
 class Lexer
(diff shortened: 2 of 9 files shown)


gcc/rust/lex/rust-lex.h
```diff
@@ -223,22 +196,56 @@ class Lexer
...
 
   // Input source wrapper thing.
-  struct InputSource
+  class InputSource
   {
+  public:
+    virtual ~InputSource () {}
+
+    // Overload operator () to return next char from input stream.
+    virtual int operator() () = 0;
```
CohenArthur, line 205: Oof :DD Very cplusplussy! If everyone is fine with it, I won't opppose it, but I personally think this would be a little more clear :)
```suggestion
    virtual int next () = 0;
```

I know I'm going to be confused when I'm gonna see `raw_input_source()` return a character

gcc/rust/lex/rust-lex.cc
```diff
@@ -116,14 +116,17 @@ is_non_decimal_int_literal_separator (char character)
   return character == 'x' || character == 'o' || character == 'b';
 }
 
-// this compiles fine, so any intellisense saying otherwise is fake news
+Lexer::Lexer (const std::string &input)
+  : input (RAIIFile::create_error ()), current_line (1), current_column (1),
+    line_map (nullptr), raw_input_source (new BufferInputSource (input, 0)),
```
CohenArthur, line 121: Any reason why we're using raw pointers and not a `unique_ptr` like usual?

gcc/rust/lex/rust-lex.cc
```diff
@@ -116,14 +116,17 @@ is_non_decimal_int_literal_separator (char character)
...
+{}
+
 Lexer::Lexer (const char *filename, RAIIFile file_input, Linemap *linemap)
   : input (std::move (file_input)), current_line (1), current_column (1),
     line_map (linemap),
-    /*input_source (input.get_raw ()), */
-    input_queue{InputSource (input.get_raw ())},
-    /*token_source (this),*/
-    token_queue (TokenSource (this))
+    raw_input_source (new FileInputSource (input.get_raw ())),
```
CohenArthur, line 128: Same here

gcc/rust/lex/rust-lex.cc
```diff
@@ -138,7 +141,9 @@ Lexer::~Lexer ()
    * mentioned in GCC docs as being useful for "just leaving an included header"
    * and stuff like that, so this line mapping functionality may need fixing.
    * FIXME: find out whether this occurs. */
+
   // line_map->stop();
+  delete raw_input_source;
```
CohenArthur, line 146: This would avoid this manual call to delete

CohenArthur (commented):
Nice changes. I like that

gcc/rust/lex/rust-lex.h
philberty: yeah i agree let me see how painful this is to update otherwise i might leave it as a good first-pr to refactor this

philberty: Looks like it wasn't too bad to update this.

PR #1367, opened 2022-07-06
Initial support for matches on tuples

dafaust:
Initial work on supporting more complex match expressions, starting with tuples. It's proven to be quite a bit more complicated than I originally expected. Still this implementation is very naive, probably very poor performance, and skips all the nuanced stuff like bindings in patterns and pattern guards.

Overview
The general idea is that we lower `MatchExpr`s to `SWITCH_EXPR` trees, but a switch can only operate on one value at a time. When scrutinizing a tuple, we can't build a switch that checks all the elements at once. So instead, just before lowering to trees, we construct a new set of nested match expressions which tests the tuple one element at a time, and is (hopefully) semantically identical to the original match. These new nested matches are simplified, so that each one tests only one thing and therefore *can* lower directly to a `SWITCH_EXPR`.

To see an example, this
```text
match (tupA, tupB, tupC) {
  (a1, b1, c1) => { blk1 }
  (a2, b2, c2) => { blk2 }
  (a1, b3, c3) => { blk3 }
  ...
  _ => { blk4 }
}
```
after one step becomes
```text
match tupA {
  a1 => {
    match (tupB, tupC) {
      (b1, c1) => { blk1 }
      (b3, c3) => { blk3 }
      ...
      _ => { blk4 }
    }
  a2 => {
    match (tupB, tupC) {
      (b2, c2) => { blk2 }
      ...
      _ => { blk4 }
    }
  }
  _ => { blk4 }
}
```
Repeat until all matches are fully simplified, then compile it all into trees.

Now, we might end up with a lot of duplicated basic blocks here as a result, but the optimization in GENERIC and GIMPLE (further on in gcc) is very good and knows how to take care of that for us.

This approach seems somewhat similar to how rustc lowers HIR to MIR, which is hopefully a good sign :)

Addresses: #1081

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -181,6 +181,54 @@ CompileExpr::visit (HIR::DereferenceExpr &expr)
 						known_valid, expr.get_locus ());
 }
 
+
+// Helper for CompileExpr::visit (HIR::MatchExpr).
+// Check that the scrutinee of EXPR is a valid kind of expression to match on.
+// Return the TypeKind of the scrutinee if it is valid, or TyTy::TypeKind::ERROR
+// if not.
+static TyTy::TypeKind
+check_match_scrutinee (HIR::MatchExpr &expr, Context *ctx)
+{
+  TyTy::BaseType *scrutinee_expr_tyty = nullptr;
+  if (!ctx->get_tyctx ()->lookup_type (
+	expr.get_scrutinee_expr ()->get_mappings ().get_hirid (),
+	&scrutinee_expr_tyty))
+    {
+      return TyTy::TypeKind::ERROR;
+    }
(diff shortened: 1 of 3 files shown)


dafaust:
> Not sure I fully understand the is_pattern_mergeable function but I guess its doing an is_equals on patterns?

Right, exactly. It's pretty much just a quick hack. I started out using just a `std::map<Pattern, Vec<MatchCase>>` to match up the sub-cases corresponding to patterns with the same first element. But I quickly encountered the issue that the Patterns of course are nodes and so have their own IDs and HIR IDs, so even when "the patterns are the same" i.e. they match the same things, they will give different keys in the map. A more "proper" solution would probably be to implement comparisons on patterns.

dafaust:
Bit of cleanup, split out another commit and updated the commit messages

PR #1373, opened 2022-07-08
Typecheck const generic params

CohenArthur:
```rust
struct Foo<const N: usize>;
struct Bar<const N: usize = { 15i32 }>; // { dg-error "expected .usize. got .i32." }
```

This PR allows us to typecheck the type of const generics as well as make sure that the optional default expression fits that type


philberty (approved):
LGTM

PR #1375, opened 2022-07-08
backend: properly handles foreign ABIs

liushuyu:
- Properly handles foreign (`extern` functions) ABIs

### DIFF
--- a/gcc/rust/backend/rust-compile-base.cc
+++ b/gcc/rust/backend/rust-compile-base.cc
@@ -17,17 +17,20 @@
 // <http://www.gnu.org/licenses/>.
 
 #include "rust-compile-base.h"
+#include "rust-abi.h"
 #include "rust-compile-item.h"
 #include "rust-compile-stmt.h"
 #include "rust-compile-fnparam.h"
 #include "rust-compile-var-decl.h"
 
+#include "rust-diagnostics.h"
 #include "rust-expr.h"	// for AST::AttrInputLiteral
 #include "rust-macro.h" // for AST::MetaNameValueStr
 
 #include "fold-const.h"
 #include "stringpool.h"
 #include "attribs.h"
+#include "tree.h"
 
 namespace Rust {
 namespace Compile {
@@ -293,6 +296,8 @@ HIRCompileBase::handle_must_use_attribute_on_fndecl (tree fndecl,
 void
 HIRCompileBase::setup_abi_options (tree fndecl, ABI abi)
 {
+  tree abi_tree = NULL_TREE;
+
   switch (abi)
     {
     case Rust::ABI::RUST:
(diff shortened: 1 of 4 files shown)


gcc/rust/backend/rust-compile-base.cc
```diff
@@ -301,22 +306,34 @@ HIRCompileBase::setup_abi_options (tree fndecl, ABI abi)
     case Rust::ABI::CDECL:
       DECL_ATTRIBUTES (fndecl)
 	= tree_cons (get_identifier ("cdecl"), NULL, DECL_ATTRIBUTES (fndecl));
-      break;
+
+      return;
```
philberty, line 310: I feel like we can make this just be handled the same as the other ABI options instead of returning.

gcc/testsuite/rust/debug/win64-abi.rs
```diff
@@ -0,0 +1,11 @@
+pub extern "win64" fn square(num: i32) -> i32 {
+    num * num
+}
+
+fn main() {
```
philberty, line 5: Nice check for the specific target i was worried this test will fail on other arch's.

philberty (approved):

gcc/rust/backend/rust-compile-base.cc
liushuyu: > I feel like we can make this just be handled the same as the other ABI options instead of returning.

Ah, sorry. I should have explained in the comments that `cdecl` is special.

We can't annotate the function with `cdecl` ABI since that's the default ABI GCC will be using, which will cause GCC to emit a warning saying the attribute is useless (which is confusing to the user as the attribute is added by us).

tschwinge: Please, via a new PR, do add such rationale as a source code comment.

liushuyu: > Please, via a new PR, do add such rationale as a source code comment.

Understood. Will do it in a moment.

gcc/testsuite/rust/debug/win64-abi.rs
```diff
@@ -0,0 +1,11 @@
...
+    // { dg-do compile { target { x86_64-*-* } } }
+    // { dg-options "-gdwarf-5 -dA -w -O1" }
+    // MS ABI dictates that the first argument is ecx instead of edi from the sysv world
+    // { dg-final { scan-assembler "%ecx, %ecx" } }
```
tschwinge, line 9: Given a x86_64-pc-linux-gnu build, when running `make check-rust` with `RUNTESTFLAGS='--target_board=unix\{,-m32,-mx32\}'`, for example, we get `PASS` for the (default) `-m64` variant, but `FAIL: rust/debug/win64-abi.rs scan-assembler %ecx, %ecx` for the latter two variants.  The `target { x86_64-*-* }` does apply in this case (and in fact should be extended/generalized to `target i?86-*-* x86_64-*-*`), but will need separate `scan-assembler` directives, conditionalized with `target [...]` or `target { ! [...] }`, with `[...]` being `lp64`/`ia32`/`x32`/etc. as appropriate (see `gcc/doc/sourcebuild.texi`, "Effective-Target Keywords".

tschwinge: Eh, actually for `-mx32`: `error: X32 does not support ‘ms_abi’ attribute`, so that will have to be handled differently, to expect that diagnostic instead.

liushuyu: > Given a x86_64-pc-linux-gnu build, when running `make check-rust` with `RUNTESTFLAGS='--target_board=unix\{,-m32,-mx32\}'`, for example, we get `PASS` for the (default) `-m64` variant, but `FAIL: rust/debug/win64-abi.rs scan-assembler %ecx, %ecx` for the latter two variants. The `target { x86_64-*-* }` does apply in this case (and in fact should be extended/generalized to `target i?86-*-* x86_64-*-*`), but will need separate `scan-assembler` directives, conditionalized with `target [...]` or `target { ! [...] }`, with `[...]` being `lp64`/`ia32`/`x32`/etc. as appropriate (see `gcc/doc/sourcebuild.texi`, "Effective-Target Keywords".

I have tested, `win64` ABI is unsupported on 32-bit x86 targets in `rustc`. `rustc` will report:

```
error[E0570]: `"win64"` is not a supported ABI for the current target
 --> <source>:2:1
  |
2 | pub extern "win64" fn square(num: i32) -> i32 {
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

PR #1378, opened 2022-07-12
hir/dump: simple dumping of crate attrs and adding a few todos

ndrwnaguib:
This is draft PR to track the current implementation of the pretty dumping of HIR. @CohenArthur and I have agreed to keep it as simple as possible for now, and enhancements will be added along the way.

### DIFF
--- a/gcc/rust/hir/rust-hir-dump.cc
+++ b/gcc/rust/hir/rust-hir-dump.cc
@@ -21,15 +21,60 @@
 namespace Rust {
 namespace HIR {
 
-Dump::Dump (std::ostream &stream) : stream (stream) {}
+Dump::Dump (std::ostream &stream) : stream (stream), indent (0) {}
 
 void
 Dump::go (HIR::Crate &crate)
 {
-  // TODO: print crate inner_attrs
-  // TODO: print crate items
-  // TODO: print crate mappings
-  stream << crate.as_string ();
+  stream << "Crate"
+	 << " "
+	 << "{" << std::endl;
+  //
+
+  indent++;
+  stream << std::string (indent, indent_char);
+  stream << "inner_attrs"
+	 << ":"
+	 << " "
+	 << "[";
+  for (auto &attr : crate.inner_attrs)
+    stream << attr.as_string ();
+  stream << "]"
+	 << "," << std::endl;
+  indent--;
+
+  indent++;
+  stream << std::string (indent, indent_char);
+  //
+
+  stream << "items"
+	 << ":"
+	 << " "
+	 << "[";
+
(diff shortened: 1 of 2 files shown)


gcc/rust/hir/rust-hir-dump.cc
```diff
@@ -26,10 +26,54 @@ Dump::Dump (std::ostream &stream) : stream (stream) {}
...
+	 << " "
+	 << "{" << std::endl;
+  //
+
+  indent++;
+  stream << std::string (indent, indent_char);
+  stream << "inner_attrs"
+	 << ":"
+	 << " "
+	 << "[";
```
CohenArthur, line 39: I think this is fine for now, but we should open up an issue about factoring this in a function as you use it often :)

gcc/rust/hir/rust-hir-dump.cc
```diff
@@ -152,8 +202,17 @@ void
...
-Dump::visit (BlockExpr &)
-{}
+Dump::visit (BlockExpr &block_expr)
+{
+  stream << "BlockExpr"
+	 << ":"
+	 << " "
+	 << "[";
+  indent++;
+  // TODO: print statements
```
CohenArthur, line 212:
```suggestion
  // TODO: print statements
  // TODO: print tail expression if present
```

gcc/rust/hir/rust-hir-dump.cc
```diff
@@ -264,8 +323,22 @@ void
...
-Dump::visit (Function &)
-{}
+Dump::visit (Function &function)
+{
+  indent++;
+  stream << std::string (indent, indent_char);
+  stream << "function"
+	 << " "
+	 << ":";
+  stream << "[" << std::endl;
```
CohenArthur, line 333:
```suggestion
  stream << "Function"
	 << " "
  stream << "{" << std::endl;
```  
Since you're printing an object, shouldn't this use curly brackets instead of square brackets?

gcc/rust/hir/rust-hir-dump.cc
```diff
@@ -264,8 +323,22 @@ void
...
+  // TODO: print function params
+  stream << std::string (indent, indent_char);
+  stream << "]" << std::endl;
```
CohenArthur, line 336:
```suggestion
  stream << "}" << std::endl;
```

gcc/rust/hir/rust-hir-dump.cc
ndrwnaguib: Yeah will do.

CohenArthur (approved):
Thanks!

PR #1379, opened 2022-07-13
backend: explain why cdecl is treated differently when setting up the ABIs

liushuyu:
- backend: explain why `cdecl` is treated differently when setting up the ABIs


liushuyu:
Requested by @tschwinge in https://github.com/Rust-GCC/gccrs/pull/1375/files#r919778903

PR #1387, opened 2022-07-15
External Items with Rust ABI need to mangle the full path

philberty:
When compiling external rust abi item requires the fully qualified
canonical path to be mangled in order to link correctly.


bjorn3:
`#[no_mangle]` is what determines if symbol mangling should happen, not the calling convention. You should mangle all functions not marked as `#[no_mangle]` and mangle all imported functions unless they are imported using an `extern "..." { ... }` block.

bjorn3:
@philberty

PR #1394, opened 2022-07-19
Fix parsing grammer of grouped expressions as the block tail expression

### LINKED ISSUE #1393: Parsing block tail expression with bitwise operators
I tried this code:

```rust
fn tst() {
    let bits = 123;
    let x = 0;
    let a = if x == 0 {
        (bits & 0x7fffff) << 1
    } else {
        (bits & 0x7fffff) | 0x800000
    };
}

```

Instead, this happened: 

```
test.rs:5:27: error: expecting ‘}’ but ‘<<’ found
    5 |         (bits & 0x7fffff) << 1
      |                           ^
test.rs:5:27: error: error may be from having an expression (as opposed to statement) in the body of the function but not last
test.rs:6:7: error: failed to parse if body block expression in if expression
    6 |     } else {
      |       ^
test.rs:6:7: error: failed to parse expression in let statement
test.rs:4:5: error: failed to parse statement or expression without block in block expression
    4 |     let a = if x == 0 {
      |     ^
test.rs:9:1: error: unrecognised token ‘}’ for start of item
    9 | }
      | ^
test.rs:9:1: error: failed to parse item in crate
test.rs:9:1: error: attempted to get string for ‘}’, which has no string. returning empty string instead
```

### Meta

- What version of Rust GCC were you using, git sha if possible. 6920d2294b3c6f106478fd3decaa511faf3cac84

philberty:
When we want to compile a grouped expression we had a check for the
LEFT_PAREN during the parse_expr_without_block but this can't be handled
here since in this paticular example the grouped expressions is simply the
lhs of the bit shift expression. This is already handled as part of
parse_expr so we can simply remove the rule here and rely on parse_expr
handling this.

Fixes #1393

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -7276,12 +7276,6 @@ Parser<ManagedTokenSource>::parse_expr_without_block (
     case LEFT_SQUARE:
       // array expr (creation, not index)
       return parse_array_expr (std::move (outer_attrs));
-    case LEFT_PAREN:
-      /* either grouped expr or tuple expr - depends on whether there is a
-       * comma
-       * inside the parentheses - if so, tuple expr, otherwise, grouped expr.
-       */
-      return parse_grouped_or_tuple_expr (std::move (outer_attrs));
       default: {
 	/* HACK: piggyback on pratt parsed expr and abuse polymorphism to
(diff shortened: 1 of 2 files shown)


CohenArthur (approved):
Oh wow lol, that's a *really* small fix :D

PR #1405, opened 2022-07-20
Allow repeating metavars alongside repetitions

### LINKED ISSUE #1400: Allow repeating non-repeating metavars alongside repetitions
When expanding repeating metavars, we have to allow the repetition of non-repeating metavars as well:

```rust
macro_rules! foo {
    ( ( $( $Trait: ident ),+ ) for $Ty: ident ) => {
        $(
            impl $Trait for $Ty {
                fn bar() -> i32 {
                    14
                }
            }
        )+
    }
}

trait Foo {
    fn bar() -> i32;
}

trait Bar {
    fn bar() -> i32;
}

trait Baz {
    fn bar() -> i32;
}

trait Qux {
    fn bar() -> i32;
}

struct S;

foo!{(Foo, Bar, Baz, Qux) for S}
```

This is valid, as the `$Ty` metavar is not a repetition. However, this should fail to compile:
```rust
macro_rules! foo {
//                                 note the repetition here now
    ( ( $( $Trait: ident ),+ ) for $($Ty: ident)+ ) => {
        $(
            impl $Trait for $Ty {
                fn bar() -> i32 {
                    14
                }
            }
        )+
    }
}

trait Foo {
    fn bar() -> i32;
}

trait Bar {
    fn bar() -> i32;
}

trait Baz {
    fn bar() -> i32;
}

trait Qux {
    fn bar() -> i32;
}

struct S;

foo!{(Foo, Bar, Baz, Qux) for S}
```

Obviously, if we were to specify as many `$Ty` as `$Trait` then there would be no issue, but that behavior is already handled.

CohenArthur:
Fixes #1400

### DIFF
--- a/gcc/rust/expand/rust-macro-expand.cc
+++ b/gcc/rust/expand/rust-macro-expand.cc
@@ -83,15 +83,14 @@ MacroExpander::expand_decl_macro (Location invoc_locus,
 
       if (did_match_rule)
 	{
-	  // FIXME: ARTHUR: Comment
-	  // Debugging
-	  for (auto &kv : matched_fragments)
-	    rust_debug ("[fragment]: %s (%ld - %s)", kv.first.c_str (),
-			kv.second.get_fragments ().size (),
-			kv.second.get_kind ()
-			    == MatchedFragmentContainer::Kind::Repetition
-			  ? "repetition"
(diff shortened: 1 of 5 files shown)


philberty (approved):
LGTM

PR #1406, opened 2022-07-20
Add Attribute checking visitor

CohenArthur:
This PR adds a new attribute checker visitor. Its role is to take care of validating builtin attributes and their inputs.

This is currently a draft as there are multiple issues:

1. The visitor is not complete
2. The lexer is broken, but was also broken before in that it didn't allow multiline string literals. There is a FIXME asking for how to check for an `EOF` character when parsing UTF-8 strings
3. The checking of attributes is very limited. Currently, only the `#[doc(alias = "...")]` is being checked. This is incomplete.

I'm looking for feedback on the implementation and advice for my issues. Thanks :)

### DIFF
--- a/gcc/rust/hir/rust-ast-lower-base.cc
+++ b/gcc/rust/hir/rust-ast-lower-base.cc
@@ -872,6 +872,8 @@ ASTLoweringBase::handle_doc_item_attribute (const HIR::Item &item,
   AST::AttrInputMetaItemContainer *meta_item = option.parse_to_meta_item ();
 
   // TODO: add actual and complete checks for the doc attributes
+  //
+  // FIXME: Move this to the AttributeChecker visitor
   rust_assert (meta_item);
 }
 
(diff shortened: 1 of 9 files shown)


CohenArthur:
Ideally, we will need to move all builtin attribute validation to this visitor. It's currently scattered throughout the compiler which makes it hard to navigate

CohenArthur:
@philberty I've fixed the lexer to handle EOF/newlines in strings properly and not crash. This changes a few tests. I've also refactored the attribute checker visitor and I guess it's in a good state, with only missing `visit` implementations? We can open up good-first-pr issues or I can do it

philberty (approved):
LGTM i can't see from the diff why the is_eof check isn't enough in the lexer but if it is an issue its definitely a bug

philberty:
This patch looks like a wip but it doesn't cause any regressions and it's moving forward to a new architecture to allow us to support more stuff so I think it looks good to merge.

CohenArthur:
Nice. I'll open up issues to go alongside it so that we don't forget to fill it up

PR #1409, opened 2022-07-22
checks: Refactor project architecture for checking passes

CohenArthur:
This commit adds a new `checks/` folder in the root of the Rust
frontend, which then contains a folder for `lints` and one for `errors`
passes. Otherwise, some error reporting passes like the privacy checks
or upcoming unsafe visitor would have to live in their own folder in the
root, or be part of another folder like `typecheck` or `hir` which
doesn't necessarily makes sense. This should enable us to find them more
easily going forward


philberty (approved):
LGTM lets get that merged, I like this a lot. We should also really rename the backend folder to be like generic or something instead that backend name bugs me a lot

PR #1410, opened 2022-07-22
unsafe: Add UnsafeCheck visitor

CohenArthur:
This visitor takes care of checking for unsafe expressions in safe
contexts. This first iteration reports the dereferencing of raw
pointers.

### DIFF
--- a/gcc/rust/checks/errors/rust-unsafe-checker.cc
+++ b/gcc/rust/checks/errors/rust-unsafe-checker.cc
@@ -1,0 +1,752 @@
+// Copyright (C) 2020-2022 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#include "rust-unsafe-checker.h"
+#include "rust-hir.h"
+#include "rust-hir-expr.h"
+#include "rust-hir-stmt.h"
+#include "rust-hir-item.h"
+
+namespace Rust {
+namespace HIR {
+
+UnsafeChecker::UnsafeChecker () : context (*Resolver::TypeCheckContext::get ())
+{}
+
+void
+UnsafeChecker::go (HIR::Crate &crate)
+{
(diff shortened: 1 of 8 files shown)


gcc/rust/checks/errors/rust-unsafe-checker.cc
```diff
@@ -0,0 +1,752 @@
...
+UnsafeChecker::visit (ReturnExpr &expr)
+{
+  if (expr.has_return_expr ())
+    expr.get_expr ()->accept_vis (*this);
+}
+
+void
+UnsafeChecker::visit (UnsafeBlockExpr &expr)
+{
+  in_unsafe_context = true;
```
philberty, line 323: I think this should be implemented as a stack. So we do:

```
class UnsafeContext {
  void push_unsafe_context(HirId unsafe_hirid) { unsafe_context.push_back(unsafe_hirid); }
  Optional<HirId> pop_unsafe_context() { 
    // handle case when unsafe_context is empty and return optional::none
    // TODO
  }
  
  bool in_unsafe_context() const { return !unsafe_context.empty(); }

  std::vector<hirid> unsafe_context;
}
```


```
unsafe {   // in_unsafe_context = true
  unsafe { // in_unsafe_context = true
  }            // in_unsafe_context = false
  
  <unsafe_thing> // fails unsafe check
}  // in_unsafe_context = false
```

philberty (changes requested):

gcc/rust/checks/errors/rust-unsafe-checker.h
```diff
@@ -0,0 +1,176 @@
...
+  virtual void visit (ImplTraitTypeOneBound &type) override;
+  virtual void visit (TupleType &type) override;
+  virtual void visit (NeverType &type) override;
+  virtual void visit (RawPointerType &type) override;
+  virtual void visit (ReferenceType &type) override;
+  virtual void visit (ArrayType &type) override;
+  virtual void visit (SliceType &type) override;
+  virtual void visit (InferredType &type) override;
+  virtual void visit (BareFunctionType &type) override;
+};
```
philberty, line 172: I like to make a new line before the closing namespace

gcc/rust/checks/errors/rust-unsafe-checker.cc
CohenArthur: Ah good catch! Didn't even think about that, thanks

gcc/rust/checks/errors/rust-unsafe-checker.cc
```diff
@@ -0,0 +1,779 @@
...
+  rust_assert (!unsafe_contexts.empty ());
+
+  auto last = unsafe_contexts.back ();
+  unsafe_contexts.pop_back ();
+
+  return last;
+}
+
+bool
+UnsafeChecker::is_unsafe_context ()
```
philberty, line 56: this can be const

philberty (approved):
LGTM.

Minor nit that in_unsafe_context could be const but you can fix in the next patches. Another minor nit is I would have made the unsafe context a separate class but there is no need to do that right now.

PR #1415, opened 2022-07-27
Add generic `StackedContexts<T>` type

CohenArthur:


philberty (approved):
LGTM

Another nice thing we could do is like RAIIStackedContextGuard guard(stackedContext) so on the constructor we enter and on the destructor we exit.

Issue #1411, opened 2022-07-22
Add unsafe checks


CohenArthur:
- [ ] Dereferencing raw pointers #1410 
- [ ] Calls to unsafe functions
- [ ] Use of inline assembly
- [ ] Use of mutable static
- [ ] Use of extern static
- [ ] Access to a union's field
- [ ] Call to functions with `#[target(feature = ...)]` attribute
- [ ] Mutation of layout constrained field
- [ ] Borrow of layout constrained field

bjorn3:
> Mutation of layout constrained field

What do you mean with a layout constrained field? A field of a packed struct? Or a field of a type with the `rustc_layout_scalar_valid_range_start` or `rustc_layout_scalar_valid_range_end` attributes? Neither is related to `Freeze`.

CohenArthur:
> > Mutation of layout constrained field
> 
> What do you mean with a layout constrained field? A field of a packed struct? Or a field of a type with the `rustc_layout_scalar_valid_range_start` or `rustc_layout_scalar_valid_range_end` attributes? Neither is related to `Freeze`.

I'm referring to [these two unsafe operations](https://github.com/rust-lang/rust/blob/41419e70366962c9a878bfe673ef4df38db6f7f1/compiler/rustc_mir_build/src/check_unsafety.rs#L534-L535) which seems to be triggered only in some cases like this one: https://github.com/rust-lang/rust/blob/41419e70366962c9a878bfe673ef4df38db6f7f1/compiler/rustc_mir_build/src/check_unsafety.rs#L270-L271

I didn't look much into it, I had just read your comment on #182 and assumed this had to do with `Freeze`. I might be completely off the mark, sorry about that.

bjorn3:
> I'm referring to [these two unsafe operations](https://github.com/rust-lang/rust/blob/41419e70366962c9a878bfe673ef4df38db6f7f1/compiler/rustc_mir_build/src/check_unsafety.rs#L534-L535) which seems to be triggered only in some cases like [this one](https://github.com/rust-lang/rust/blob/41419e70366962c9a878bfe673ef4df38db6f7f1/compiler/rustc_mir_build/src/check_unsafety.rs#L270-L271)

I see. In that case I have another item to add to the list:

* [ ] Borrow of a packed struct's field (currently back compat lint in rustc, will be a hard error in the future)

CohenArthur:
Thank you! Adding it to the list to make sure I don't miss it

PR #1416, opened 2022-07-27
unsafe: Add checks for union field accesses

CohenArthur:
Addresses #1411

### DIFF
--- a/gcc/rust/checks/errors/rust-unsafe-checker.cc
+++ b/gcc/rust/checks/errors/rust-unsafe-checker.cc
@@ -262,10 +262,24 @@ UnsafeChecker::visit (MethodCallExpr &expr)
 void
 UnsafeChecker::visit (FieldAccessExpr &expr)
 {
-  // FIXME: If the receiver is an union, we need to be in an unsafe context to
-  // access it. Make sure to check.
-
   expr.get_receiver_expr ()->accept_vis (*this);
+
+  if (is_unsafe_context ())
+    return;
+
+  TyTy::BaseType *receiver_ty;
+  auto ok = context.lookup_type (
+    expr.get_receiver_expr ()->get_mappings ().get_hirid (), &receiver_ty);
+  rust_assert (ok);
+
+  if (receiver_ty->get_kind () == TyTy::TypeKind::ADT)
+    {
+      auto maybe_union = static_cast<TyTy::ADTType *> (receiver_ty);
+      if (maybe_union->is_union ())
+	rust_error_at (
+	  expr.get_locus (),
+	  "access to union field requires unsafe function or block");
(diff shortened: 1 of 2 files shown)


gcc/rust/checks/errors/rust-unsafe-checker.cc
```diff
@@ -262,10 +262,23 @@ UnsafeChecker::visit (MethodCallExpr &expr)
...
-  // FIXME: If the receiver is an union, we need to be in an unsafe context to
-  // access it. Make sure to check.
-
   expr.get_receiver_expr ()->accept_vis (*this);
+
+  if (is_unsafe_context ())
+    return;
+
+  TyTy::BaseType *receiver_ty;
+  context.lookup_type (expr.get_receiver_expr ()->get_mappings ().get_hirid (),
```
philberty, line 271: This should be:

```
bool ok = context.lookup_type();
rust_assert(ok);
```

I need to make a new interface for code generation, which can default to an error node, but there is some refactoring required for that.

gcc/rust/checks/errors/rust-unsafe-checker.cc
```diff
@@ -262,10 +262,23 @@ UnsafeChecker::visit (MethodCallExpr &expr)
...
+		       &receiver_ty);
+
+  if (receiver_ty->get_kind () == TyTy::TypeKind::ADT)
+    {
+      auto maybe_union = static_cast<TyTy::ADTType *> (receiver_ty);
+      if (maybe_union->get_adt_kind () == TyTy::ADTType::ADTKind::UNION)
```
philberty, line 277: we have a helper on the ADTType is_union which would be best to use here.

```
const auto& adt = static_cast<const TyTy::ADTType&>(*receiver_ty);
if (adt.is_union()) {
   error_message;
}
```

philberty (changes requested):
minor nits

gcc/rust/checks/errors/rust-unsafe-checker.cc
CohenArthur: Fixed in [f532ae5](https://github.com/Rust-GCC/gccrs/pull/1416/commits/f532ae5aa37c8555dd6a99d8502436509db7930f)

gcc/rust/checks/errors/rust-unsafe-checker.cc
CohenArthur: Fixed in [f532ae5](https://github.com/Rust-GCC/gccrs/pull/1416/commits/f532ae5aa37c8555dd6a99d8502436509db7930f)

philberty (approved):
Nice one LGTM

PR #1417, opened 2022-07-27
unsafe: check uses of static variables

CohenArthur:
Addresses #1411 

Accessing an extern static or a mut static is unsafe and requires an unsafe function or block

### DIFF
--- a/gcc/rust/checks/errors/rust-unsafe-checker.cc
+++ b/gcc/rust/checks/errors/rust-unsafe-checker.cc
@@ -25,7 +25,10 @@
 namespace Rust {
 namespace HIR {
 
-UnsafeChecker::UnsafeChecker () : context (*Resolver::TypeCheckContext::get ())
+UnsafeChecker::UnsafeChecker ()
+  : context (*Resolver::TypeCheckContext::get ()),
+    resolver (*Resolver::Resolver::get ()),
+    mappings (*Analysis::Mappings::get ())
 {}
 
 void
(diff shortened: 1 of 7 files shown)


gcc/rust/checks/errors/rust-unsafe-checker.cc
```diff
@@ -60,7 +104,18 @@ UnsafeChecker::is_unsafe_context ()
 
 void
 UnsafeChecker::visit (IdentifierExpr &ident_expr)
-{}
+{
+  NodeId ast_node_id = ident_expr.get_mappings ().get_nodeid ();
+  NodeId ref_node_id;
+  HirId definition_id;
+
+  if (!resolver.lookup_resolved_name (ast_node_id, &ref_node_id))
```
philberty, line 112: This is good, ideally this pass should be permissive of saw_errors eventually but we need to have the safe interface for the type context to return error nodes instead of nullptr to do that properly.

philberty (approved):
Nice work. LGTM

Something we might start to think about doing is for every error diagnostic have some kind of class for them all, so we can enforce good error messages so for example we might have:

class ErrorDiagnostics {
  static void EStaticMutableUnsafe(Location) { <standardized error message> }
}

CohenArthur:
> Nice work. LGTM
> 
> Something we might start to think about doing is for every error diagnostic have some kind of class for them all, so we can enforce good error messages so for example we might have:
> 
> class ErrorDiagnostics { static void EStaticMutableUnsafe(Location) { } }

Yeah, this will definitely come in handy when adding error codes as well! I'll open up an issue regarding this

PR #1419, opened 2022-07-27
Const checker

CohenArthur:
This commit serves as a base to ensure that only the allowed subset of
the language is called from const contexts. For example, this limits
function calls to calls to const functions only when in a const context

You can skip reviewing b4864c7e5a59a19fea6203c4b581889d4d6c20a1 as it's the exact same commit cherry-picked from #1415

### DIFF
--- a/gcc/rust/checks/errors/rust-const-checker.h
+++ b/gcc/rust/checks/errors/rust-const-checker.h
@@ -1,0 +1,183 @@
+// Copyright (C) 2020-2022 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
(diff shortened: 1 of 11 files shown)


gcc/rust/checks/errors/rust-const-checker.h
```diff
@@ -0,0 +1,183 @@
...
+  void go (HIR::Crate &crate);
+
+private:
+  /**
+   * Check that only const functions are called in const contexts
+   */
+  void check_function_call (HirId fn_id, Location locus);
+
+  StackedContexts<HirId> const_context;
+  Resolver::Resolver resolver;
```
philberty, line 43: You should be taking a reference instead of a copy to the resolver and mappings classes.

philberty (changes requested):
minor nit but otherwise LGTM

gcc/rust/checks/errors/rust-const-checker.h
CohenArthur: Oh fuck, this is a typo. This is what I meant to do originally. Sorry!

PR #1421, opened 2022-07-28
expand: further improves the handling of recursive macros

### LINKED ISSUE #1403: Infinite loop/hang when expanding recursive statement macros
@liushuyu I think this is similar to the issue you fix in #1401, if you're interested

```rust
macro_rules! stmt {
    ($s:stmt) => {
        $s
    };
    ($s:stmt, $($ss:stmt),*) => {
        $s;
        stmt!($($ss),*);
    };
}

fn main() {
    stmt!(
        struct S;
    );
    stmt!(
        struct A;,
        struct B;,
        struct C;,
        struct D;,
        struct E;
    );
}

```

I expected to see this happen: Proper expansion and a bunch of unused warnings

Instead, this happened: Infinite hang/loop

liushuyu:
- expand: further improves the handling of recursive macros.
Now all the attribute visitor that expands the macro fragment will recursively expand the macro using the unified expanding function `expand_macro_fragment_recursive` instead of expanding the fragment only once.

Fixes #1403

### DIFF
--- a/gcc/rust/expand/rust-attribute-visitor.h
+++ b/gcc/rust/expand/rust-attribute-visitor.h
@@ -17,6 +17,7 @@
 // <http://www.gnu.org/licenses/>.
 
 #include "rust-ast-visitor.h"
+#include "rust-ast.h"
 #include "rust-macro-expand.h"
 
 namespace Rust {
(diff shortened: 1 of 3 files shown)


liushuyu:
Remark: Although the infinite loop has been fixed, the expansion result for #1403 does not look 100% correct compared to `rustc`. I am not sure what caused this at the moment.

gcc/rust/expand/rust-attribute-visitor.h
```diff
@@ -16,7 +16,10 @@
 // along with GCC; see the file COPYING3.  If not see
 // <http://www.gnu.org/licenses/>.
 
+#include <vector>
```
philberty, line 19: IIRC vector is included as part of rust-system.h

philberty (approved):
LGTM @CohenArthur will know better than Ibut I can't see any issues.

philberty:
Nice work once again @liushuyu

philberty:
If I was being pedantic, it would be nice if you could include a small sentence explaining what the patch does in your commit message.

CohenArthur (approved):
This looks good. Can you go into more details as to how the behavior is different from `rustc` when expanding this? And thanks for the effort you put into this :)

liushuyu:
> If I was being pedantic, it would be nice if you could include a small sentence explaining what the patch does in your commit message.

I see. I will change that.

liushuyu:
> Can you go into more details as to how the behavior is different from `rustc` when expanding this?

Currently, this GCC implementation expands the macro to this (from `gccrs.ast-expanded.dump`):

```
void main()
BlockExpr:

  outer attributes: none
  inner attributes: none
 statements: 
 struct S
 Generic params: none
 Where clause: none
 Struct fields: none (unit)
 struct E
 Generic params: none
 Where clause: none
 Struct fields: none (unit)
 final expression: none
```

For reference, `rustc` will expand the same file to this (from `cargo expand` command):

```rust
#![feature(prelude_import)]
#[prelude_import]
use std::prelude::rust_2021::*;
#[macro_use]
extern crate std;
fn main() {
    struct S;
    struct A;;
    struct B;;
    struct C;;
    struct D;;
    struct E;
}
```

liushuyu:
If the previous example is not obvious, try the following example:

```rust
macro_rules! stmt {
    ($s:stmt) => {
        $s
    };
    ($s:stmt, $($ss:stmt),*) => {
        $s;
        stmt!($($ss),*);
    };
}

fn test() -> i32 {
    stmt!(
        let a = 1
    );
    stmt!(
        let b = 2,
        let c = 3,
        let d = 4,
        let e = 5,
        let f = b + c + d + e
    );
    f
}

fn main() {
    let _ = test();
}
```

CohenArthur:
@liushuyu thanks for the example. I think this PR is good in that it removes a bug (infinite loop) but the output is, as you noted, completely incorrect. I'll open an issue for this but I still want to merge your PR. If you want to have a look at the issue I'm about to open, feel free, but I will get to it at some point as I think this is a big issue with our macro pass.

bors r+

Thanks a lot for the time you put into this :)

PR #1423, opened 2022-07-28
Rename `rust1` to `rs1`

CohenArthur:
@edelsohn if you'd like to have a look


tschwinge:
In my opinion, before renaming individual components, shouldn't we first establish some general plan about how to name all things (project, repository, language in GCC, compiler proper, etc.)?

philberty:
@edelsohn  ^^

edelsohn:
There isn't a lot of historical consistency for the naming of the "compiler proper" other than a "1" in the name: cc1, cc1plus, cc1obj, cc1objplus, gnat1, go1, f771, f951, jc1, d21, lto1.

Whatever executable name you prefer that includes "1" matches the pattern. rs1 is fine.

Zopolis4:
I do think `rs1` matches that pattern better than `rust1`, though. Most languages don't have their full names (apart from the really short ones like Go, C and D, its hard to shorten those.) `rs1` also matches `jc1` quite well.

philberty:
Okay, the consensus is that we close this until we have some guidelines and we _need_ to do it.

PR #1424, opened 2022-07-28
Desugar HIR use and extern crate delcarations

philberty:
These declarations don't have any meaning in the HIR as these statements
are used during name resolution which contain AST::SimplePaths which can
be fully resolved at that stage.


CohenArthur (approved):
+0, -637: The best kind of pull request :DDD LGTM!

bjorn3:
`pub use` still has a meaning, right? It needs to be included in the crate metadata.

philberty:
Darn thats a good point @bjorn3 i will look at this again in a few weeks.

philberty:
I'm going to close this PR and just desugar the ExternBlock out since it doesnt seem to have any effect

Issue #1115, opened 2022-04-13
Only allow compilation of one file at a time


CohenArthur:
I actually would like to add the following changes so that this assumption will always be true:

```diff
diff --git a/gcc/rust/rust-lang.cc b/gcc/rust/rust-lang.cc
index 5ecd79b15..5f06862eb 100644
--- a/gcc/rust/rust-lang.cc
+++ b/gcc/rust/rust-lang.cc
@@ -168,6 +168,12 @@ grs_langhook_parse_file (void)
 {
   rust_debug ("Preparing to parse files. ");
 
+  /* Rust only allows the "root file" to be supplied */
+  if (num_in_fnames > 1) {
+    rust_error_at(Location(), "multiple input filenames provided");
+    return;
+  }
+
   Rust::Session::get_instance ().parse_files (num_in_fnames, in_fnames);
 }
 
```

_Originally posted by @liushuyu in https://github.com/Rust-GCC/gccrs/pull/1088#discussion_r844624558_

First of all, this is more of a question: Is it something we want to do? I think probably yes, but I'm curious about what everyone thinks.

philberty:
Yeah the lang-hook interface expects to be able to handle multiple files, we need to change that. If we make this change can we should update the rust-session-manager::parse_files to be Parse and for it to just take a single path though does this have an interaction with stdin to think about? I remember @dkm added code for that at one point.

I would like to see the Session::parse_file function be changed into something like Session::compilation_pipeline(filename) just so that we signify that this is designed to run a pipeline on a single file.

PR #1425, opened 2022-07-28
Cleanup frontend entrypoint

CohenArthur:
This removes a lot of code that was simply ported from the Go frontend, and forbids the usage of multiple input files.

Addresses #1115

### DIFF
--- a/gcc/rust/rust-session-manager.cc
+++ b/gcc/rust/rust-session-manager.cc
@@ -574,13 +574,16 @@ Session::enable_dump (std::string arg)
 /* Actual main entry point for front-end. Called from langhook to parse files.
  */
 void
-Session::parse_files (int num_files, const char **files)
+Session::handle_input_files (int num_files, const char **files)
 {
+  if (num_files != 1)
+    rust_fatal_error (Location (),
+		      "only one file may be specified on the command line");
+
+  const auto &file = files[0];
+
   if (options.crate_name.empty ())
     {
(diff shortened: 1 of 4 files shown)


gcc/rust/rust-session-manager.cc
```diff
@@ -574,13 +574,16 @@ Session::enable_dump (std::string arg)
 /* Actual main entry point for front-end. Called from langhook to parse files.
  */
 void
-Session::parse_files (int num_files, const char **files)
+Session::handle_input_files (int num_files, const char **files)
 {
+  if (num_files != 1)
+    rust_fatal_error (Location (),
+		      "only one file may be specified on the command line");
```
liushuyu, line 581: I am not sure if we want to match `rustc`'s error message closely.

`rustc` will emit this message when multiple files are specified:

```
error: multiple input filenames provided (first two filenames are `main.rs` and `lib.rs`)
```

CohenArthur: I think rustc's message is great :DD We can improve this one of course

philberty (approved):
LGTM.

I've actually spent alot of time reviewing this PR, the main piece in the compiler driver rustspec.cc we will eventually need to change parts. To handle the no_core and no_std attributes there is got to be a way to send information back here from the compiler proper to say yes link in libcore or libstd.

Maybe during metadata export we export some kind of info into a known file and read info there to figure this out but its not important right now.

PR #1427, opened 2022-07-29
unsafe: Check for unsafe function/method calls

CohenArthur:
Addresses #1411


philberty (approved):
I was about to mention a missing dg-error on unsafe7.rs testcase. But you got there first.

LGTM

PR #1429, opened 2022-08-02
expand: correctly handles non-macro nodes

### LINKED ISSUE #1426: Invalid recursive :stmt expansion in fragment specifiers
If the previous example is not obvious, try the following example:

```rust
macro_rules! stmt {
    ($s:stmt) => {
        $s
    };
    ($s:stmt, $($ss:stmt),*) => {
        $s;
        stmt!($($ss),*);
    };
}

fn test() -> i32 {
    stmt!(
        let a = 1
    );
    stmt!(
        let b = 2,
        let c = 3,
        let d = 4,
        let e = 5,
        let f = b + c + d + e
    );
    f
}

fn main() {
    let _ = test();
}
```

_Originally posted by @liushuyu in https://github.com/Rust-GCC/gccrs/issues/1421#issuecomment-1198902446_

liushuyu:
- expand: correctly handles non-macro nodes when expanding macros recursively to avoid improperly stripping them

Fixes #1426

### DIFF
--- a/gcc/rust/expand/rust-macro-expand.h
+++ b/gcc/rust/expand/rust-macro-expand.h
@@ -321,12 +321,28 @@ struct MacroExpander
   AST::ASTFragment take_expanded_fragment (AST::ASTVisitor &vis)
   {
     AST::ASTFragment old_fragment = std::move (expanded_fragment);
+    auto accumulator = std::vector<AST::SingleASTNode> ();
     expanded_fragment = AST::ASTFragment::create_error ();
 
     for (auto &node : old_fragment.get_nodes ())
       {
 	expansion_depth++;
 	node.accept_vis (vis);
+	// we'll decide the next move according to the outcome of the macro
+	// expansion
+	if (expanded_fragment.is_error ())
+	  accumulator.push_back (node); // if expansion fails, there might be a
+					// non-macro expression we need to keep
+	else
+	  {
+	    // if expansion succeeded, then we need to merge the fragment with
+	    // the contents in the accumulator, so that our final expansion
+	    // result will contain non-macro nodes as it should
+	    auto new_nodes = expanded_fragment.get_nodes ();
+	    std::move (new_nodes.begin (), new_nodes.end (),
+		       std::back_inserter (accumulator));
+	    expanded_fragment = AST::ASTFragment (accumulator);
+	  }
 	expansion_depth--;
       }
 
--- a/gcc/testsuite/rust/compile/macro-issue1403.rs
+++ b/gcc/testsuite/rust/compile/macro-issue1403.rs
@@ -1,0 +1,31 @@
+// { dg-do compile }
+// { dg-options "-O1 -gdwarf-5 -dA -w" }
+macro_rules! stmt {
+    ($s:stmt) => {
+        $s
+    };
+    ($s:stmt, $($ss:stmt),*) => {
+        $s;
+        stmt!($($ss),*);
+    };
+}
+
+pub fn test() -> i32 {
+    stmt!(
+        let a = 1
+    );
+    stmt!(
+        let b = 2,
+        let c = 3,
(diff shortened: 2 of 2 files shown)


gcc/testsuite/rust/compile/macro-issue1403.rs
```diff
@@ -0,0 +1,31 @@
...
+        let a = 1
+    );
+    stmt!(
+        let b = 2,
+        let c = 3,
+        let d = 4,
+        let e = 5,
+        let f = b + c + d + e
+    );
+    // { dg-final { scan-assembler "14" } }
```
philberty, line 24: I think it might be better to avoid needing dg-options with a specified optimization level here. I would prefer if this was an execution test and verify the result of 14 you could print it and use dg-output to match the print output.

philberty (approved):
LGTM I like the comments you added

gcc/testsuite/rust/compile/macro-issue1403.rs
tschwinge: ACK to the first part; see #1438 I've just filed.

Generally I suggest to avoid adding more such executions tests -- instead, reduce them ;-) -- for they have considerably more overhead (assemble, link, execute (possibly on a remote target), verify output).

Execution tests will become more useful once we have language support libraries etc. to link against.

gcc/rust/expand/rust-macro-expand.h
```diff
@@ -321,12 +321,28 @@ struct MacroExpander
...
 
     for (auto &node : old_fragment.get_nodes ())
       {
 	expansion_depth++;
 	node.accept_vis (vis);
+	// we'll decide the next move according to the outcome of the macro
+	// expansion
+	if (expanded_fragment.is_error ())
+	  accumulator.push_back (node); // if expansion fails, there might be a
+					// non-macro expression we need to keep
```
CohenArthur, line 335: Do you have any test cases for this or any examples? I have a hard time figuring out what this achieves (sorry, this is on me, not on your code) and why we'd still want to push to the accumulator when we encounter an error.

CohenArthur (commented):
Late to the party, but this looks good to me. Thanks again for working on the macros :)

gcc/rust/expand/rust-macro-expand.h
liushuyu: > Do you have any test cases for this or any examples? I have a hard time figuring out what this achieves (sorry, this is on me, not on your code) and why we'd still want to push to the accumulator when we encounter an error.

This block is what fixed the issue.

When recursively expanding on a macro, the expansion may produce nodes that can't be expanded further (macro transcribed completely to a regular Rust expression). However, they are still part of the expansion due to how `take_expanded_fragment` is invoked in most places.

The explanations above may seem a bit dry, so let's see an example, shall we?

Okay, let's see the simplified version of the test case included in this pull request:

```rust
stmt!(
        let a = 1,
        let b = 2,
    );
```

One the first pass of GCC-rs expansion, this expands to a single fragment:

```rust
{ let a = 1; stmt!(let b = 2,) }
```

In this expanded fragment, there are 2 (two) nodes, `let a = 1` and `stmt!(let b = 2,)`.

What will happen to the first node when we expand this _fragment_ again?

... ...

Hmm... The first node will return an expansion error after calling `accept_vis (vis)` on it since it does not contain a macro expression. So, in this case, if we discard the error, what will happen then...?

See? It's what you might call a hack, but it's the easiest solution I could think of.

CohenArthur: Thanks!!!! That's super clear :D Thanks for taking the time. Don't be afraid to put lengthy explanations like these in comments within the code, I often do it (e.g. the repetition multimap algorithm thingamagig) and it helps me when looking back on it months later (and no one has complained about it yet :D)

PR #1430, opened 2022-08-02
hir: more conformant builtin macro detection

### LINKED ISSUE #971: Allow redefining builltin macros
rustc allows the redefinition of builtin macros. I am unsure whether or not this is an intended behavior, but we should probably look into it! Please feel free to chime in with additional info

```rust
macro_rules! file {
    () => { 14 }
}

fn main() {
    let a = file!(); // = 14, if we remove the declaration above it expands correctly to the file name
}
```

liushuyu:
- hir: more conformant builtin macro detection

Fixes #971

### DIFF
--- a/gcc/rust/util/rust-hir-map.cc
+++ b/gcc/rust/util/rust-hir-map.cc
@@ -18,6 +18,7 @@
 
 #include "rust-hir-map.h"
 #include "rust-ast-full.h"
+#include "rust-diagnostics.h"
 #include "rust-hir-full.h"
 #include "rust-macro-builtins.h"
 #include "rust-mapping-common.h"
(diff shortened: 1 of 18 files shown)


liushuyu:
Pre-condition: needs #1429 to be merged.

philberty (approved):
LGTM we might need to think about this again. Name resolution of macros in the expand pass I am not sure its really working properly there yet so we might do something there that might affect this too so keep that in mind. You might want to look at that in general too.

gcc/rust/util/rust-hir-map.cc
```diff
@@ -818,9 +819,22 @@ Mappings::insert_macro_def (AST::MacroRulesDefinition *macro)
...
+		   });
+  if (should_be_builtin)
+    {
+      auto builtin = builtin_macros.find (macro->get_rule_name ());
+      if (builtin != builtin_macros.end ())
+	macro->set_builtin_transcriber (builtin->second);
+      else
+	rust_error_at (macro->get_locus (),
+		       "cannot find a built-in macro with name %qs",
+		       macro->get_rule_name ().c_str ());
```
tschwinge, line 836: Should also add a test case that runs into this error diagnostic.  (For example, add a variant of all those where you added `#[rustc_builtin_macro]`?)

CohenArthur (commented):
Sound, and this matches `rustc`'s behavior perfectly as well. I don't think we should run into path issues for this since it only applies to the macro definition and not its uses. LGTM!

PR #1438, opened 2022-08-06
Enhance #1426 "Invalid recursive :stmt expansion in fragment specifiers" test case

tschwinge:
... that was recently added in #1429 commit 9fc6a27b5c6ea2c775646c4474b9084da76b1764 "expand: correctly handles non-macro nodes ...".

  - Rename 'macro-issue1403.rs' -> 'macro-issue1426.rs'.
  - Get rid of '-w'; expect diagnostic instead.
  - Get rid of debugging options not relevant here.
  - Get rid of explicit '-O1'; make it a "torture" test case.
  - Replace always-matching 'scan-assembler "14"' ("14" appears in the crate name) with directed dump scanning.
      - ..., just not for '-O0', where such optimization is not done.

### DIFF
--- a/gcc/testsuite/rust/compile/torture/macro-issue1426.rs
+++ b/gcc/testsuite/rust/compile/torture/macro-issue1426.rs
@@ -1,5 +1,5 @@
-// { dg-do compile }
-// { dg-options "-O1 -gdwarf-5 -dA -w" }
+// { dg-additional-options -fdump-tree-ccp1-raw }
+
 macro_rules! stmt {
     ($s:stmt) => {
         $s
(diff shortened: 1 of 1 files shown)


tschwinge:
@liushuyu, others: happy to discuss any questions you may have, as always!

liushuyu:
> @liushuyu, others: happy to discuss any questions you may have, as always!

It makes more sense to scan the GIMPLE tree as this is mostly CPU architectural-independent.

Also, is `ccp1` pass the "constant propagation" pass?

CohenArthur (approved):
This looks good to me. We've been abusing `-w` in a lot of testcases, and it's true that we might be missing other important warnings and pass test cases that shouldn't. Most of our stuff produces errors rather than warnings though, so I'm not too worried, but still.

philberty (approved):
LGTM yeah we are using -w a lot mostly because our dead code scan needs a slight update to handle crates with no main entry point so it assumes everything is unused rustc uses the --crate-type option to distinguish this case. It also doesn't handle the awkward cases of autoderef and things of that nature but in theory, it should be simple enough to update it down the line.

PR #1444, opened 2022-08-08
Handle `rustc_const_{un}stable` properly for intrinsics

### LINKED ISSUE #1440: Missing const check in uint_impl! macro/array size arguments
```rust
                pub const fn from_ne_bytes(bytes: [u8; mem::size_of::<Self>()]) -> Self {
```

I believe this should only work if the `mem::size_of` function is marked as `const` since array sizes are a const context.

_Originally posted by @CohenArthur in https://github.com/Rust-GCC/gccrs/issues/1432#issuecomment-1207672499_

CohenArthur:
Since extern functions cannot be marked as const (they take no
qualifiers) but some intrinsics are const, while still being `extern
"rust-intrinsic", we need to be able to handle the
`#[rustc_const_stable]` and `#[rustc_const_unstable]` attribute.

Technically, this simply indicates that a certain intrinsic is constant
and can be used in const contexts, such as `size_of` or `offset`.

We also need to mark all const intrinsics with `rustc_const_stable` in
the testsuite.

Closes #1440

### DIFF
--- a/gcc/rust/checks/errors/rust-const-checker.cc
+++ b/gcc/rust/checks/errors/rust-const-checker.cc
@@ -235,11 +235,48 @@ ConstChecker::check_function_call (HirId fn_id, Location locus)
     return;
 
   auto maybe_fn = mappings.lookup_hir_item (fn_id);
-  if (!maybe_fn || maybe_fn->get_item_kind () != Item::ItemKind::Function)
+  if (maybe_fn && maybe_fn->get_item_kind () != Item::ItemKind::Function)
     return;
 
-  auto fn = static_cast<Function *> (maybe_fn);
-  if (!fn->get_qualifiers ().is_const ())
+  // There are const extern functions (intrinsics)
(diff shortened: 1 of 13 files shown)


philberty:
We will need this for @abbasfaisal so that we can mark intrinsic as const properly for the constexpr machinery to work properly nice work

philberty (approved):
LGTM i like that this is a query based way of doing it as well.

One thing that would be useful for the work on the constexpr branch would be that when we iterate an extern block in the backend pass would be if there was a mechanism of:

```
static ConstCtx::is_extern_item_const (HIR::ExternalItem&) { ... }
```

So that when we are generating intrinsic we have a helper to mark them as const for the constexpr machinery instead of duplicating your checks here.

CohenArthur:
@philberty very good idea! I've added that in [4ffd884](https://github.com/Rust-GCC/gccrs/pull/1444/commits/4ffd884a69396d828049d4a14d17e6d3f6c8d61f) and added a comment to move it eventually

PR #1445, opened 2022-08-08
compile: Move and rename `SimpleIntrinsics` -> BuiltinsContext

CohenArthur:
This will allow us to use the class for more than just math builtins

### DIFF
--- a/gcc/rust/backend/rust-compile-intrinsic.h
+++ b/gcc/rust/backend/rust-compile-intrinsic.h
@@ -18,6 +18,7 @@
 #define RUST_COMPILE_INTRINSIC
 
 #include "rust-compile-context.h"
+#include "langhooks.h"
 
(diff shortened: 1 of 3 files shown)


gcc/rust/backend/rust-compile-intrinsic.h
```diff
@@ -18,6 +18,7 @@
 #define RUST_COMPILE_INTRINSIC
 
 #include "rust-compile-context.h"
```
philberty, line 20: does it need the include here?

philberty (approved):
LGTM

gcc/rust/backend/rust-compile-intrinsic.h
CohenArthur: You're right, I cleaned it up to include only the essentials. Good point

PR #1452, opened 2022-08-09
transmute: Fix behavior by always performing the raw copy

CohenArthur:
This desugars calls to transmute the following way:

`transmute::<T1, T2>(value)`

->

`*((T2 *) &value)`

This always ends up being optimized into a simple copy for small types
and a memcpy for larger types.

### DIFF
--- a/gcc/rust/backend/rust-compile-intrinsic.cc
+++ b/gcc/rust/backend/rust-compile-intrinsic.cc
@@ -390,33 +390,15 @@ transmute_intrinsic_handler (Context *ctx, TyTy::BaseType *fntype_tyty)
   ctx->push_block (code_block);
 
   // BUILTIN transmute FN BODY BEGIN
-  tree result_type_tree = TREE_TYPE (DECL_RESULT (fndecl));
-  tree result_expr = error_mark_node;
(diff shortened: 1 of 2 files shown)


philberty (changes requested):
Nice work.

IIRC the test here is likely to fail on the build farm it might be worthwhile chaning to use i32 seems to be more portable when printing stuff like this.

Or maybe instead we do dg-scan-gimple for the transmute stuff I think there are some examples in the testsuite for that.

CohenArthur:
> Nice work.
> 
> IIRC the test here is likely to fail on the build farm it might be worthwhile chaning to use i32 seems to be more portable when printing stuff like this.

I've changed the test to not print anything and be more minimal. It's also using 32 bit values now to be sure :)

> Or maybe instead we do dg-scan-gimple for the transmute stuff I think there are some examples in the testsuite for that.

I think this is harder, and I didn't want to lose too much time fighting dejagnu haha. Ideally, what this testcase should simply test is that raw integer types have the same behavior as ADT-like types when transmuting, as that behavior was already the correct one.

philberty (approved):
LGTM

CohenArthur:
Waiting for #1454 because I think there'll be merge conflicts

Issue #871, opened 2022-01-14
'-m32' FAIL: rust/execute/torture/match3.rs   -O0  output pattern test


tschwinge:
At #866 commit 6d0892636e9642e75a858c40c45acd1df79c63e1 "Add support for wildcard patterns within MatchExprs", the new `rust/execute/torture/match3.rs` FAILs its *output pattern test* for `-m32`:

```
PASS: rust/execute/torture/match3.rs   -O0  (test for excess errors)
Setting LD_LIBRARY_PATH to .:.:
Execution timeout is: 300
spawn [open ...]
Foo::A
wildcard
wildcard
Foo::D 20 0
PASS: rust/execute/torture/match3.rs   -O0  execution test
FAIL: rust/execute/torture/match3.rs   -O0  output pattern test
Output was:
Foo::A
wildcard
wildcard
Foo::D 20 0

Should match:
Foo::A
wildcard
wildcard
Foo::D 20 80
```

Same for other optimization levels.

Similar to #844 "'-m32' FAIL: rust/execute/torture/match1.rs -O0 output pattern test"?

philberty:
This is fixed via #1453

PR #1453, opened 2022-08-09
Add in -m32,-m64 configuration to make check-rust

### LINKED ISSUE #1439: x86/x86_64 '-m32': FAIL: rust/compile/torture/issue-1432.rs   -O0  (internal compiler error: in expand_shift_1, at expmed.cc:2672)
..., and similar for other optimization levels.  All-clean for default '-m64' x86_64.

Test case added in recent #1433 commit 7022b9dd107e534896d8383f6bc4ce70b4726cc9 "Fix bad transmute for aggregate types", with code changes.
(With those code changes temporarily reverted for testing, we're running into 'internal compiler error: 'verify_gimple' failed' for *all* configurations I've tested.)

philberty:
This is to gate PR's which might break builds on 32bit systems.

Fixes #1439 #871

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -67,7 +67,7 @@ jobs:
     - name: Run Tests
       run: |
            cd gccrs-build; \
-           make check-rust
+           make check-rust RUNTESTFLAGS="--target_board=unix\{-m32,-m64}"
     - name: Archive check-rust results
       uses: actions/upload-artifact@v2
       with:
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -1,4 +1,4 @@
-name: C/C++ CI
+name: GCC Rust build and test
 
 on:
   push:
@@ -9,7 +9,7 @@ on:
     branches: [ master ]
 
 jobs:
-  build-and-check:
+  build-and-check-ubuntu-64bit:
 
     env:
       # Force locale, in particular for reproducible results re '.github/bors_log_expected_warnings' (see below).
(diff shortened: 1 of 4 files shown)


philberty:
@liushuyu i have updated your win64 test case here so you might like to review this.

.github/workflows/ccpp.yml
```diff
@@ -67,7 +67,7 @@ jobs:
     - name: Run Tests
       run: |
            cd gccrs-build; \
-           make check-rust
+           make check-rust RUNTESTFLAGS="--target_board=unix\{-m32,-m64}"
```
CohenArthur, line 70: I think it would be worth it to also do it on gcc 4.8 as clearly this is a surprising endeavor haha. I don't know if that causes issues on MacOS so I don't know if it's worth adding. @simonpcook what do you think?

I think it's fine to just add that `--target_board=unix\{-m32, -m64}` to the gcc 4.8 step, but eventually if we keep adding these options it'd be nice to make use of the github action matrices to run these jobs in parallel. Just a note for future modifications.

gcc/testsuite/rust/compile/torture/issue-1432.rs
```diff
@@ -67,6 +67,5 @@ impl_uint!(
     u16 = "u16",
     u32 = "u32",
     u64 = "u64",
-    u128 = "u128",
```
CohenArthur, line 70: Interesting that this causes failures on 32 bits. How do we store 128 bits values? I believe they should simply be stored as 4 words on the stack [godbolt link](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,selection:(endColumn:1,endLineNumber:11,positionColumn:1,positionLineNumber:11,selectionStartColumn:1,selectionStartLineNumber:11,startColumn:1,startLineNumber:11),source:'pub+fn+foo()+-%3E+u64+%7B%0A++++let+a+%3D+15%3B%0A++++let+b+%3D+15%3B%0A%0A++++a+%2B+b%0A%7D%0A%0Apub+fn+bar()+-%3E+u128+%7B%0A++++let+a+%3D+15%3B%0A++++let+b+%3D+15%3B%0A%0A++++a+%2B+b%0A%7D'),l:'5',n:'0',o:'Rust+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:r1620,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:rust,libs:!(),options:'-C+opt-level%3D0+--target+i686-unknown-linux-gnu',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1,tree:'1'),l:'5',n:'0',o:'rustc+1.62.0+(Rust,+Editor+%231,+Compiler+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

CohenArthur (commented):
Non-blocking comment regarding GCC 4.8. LGTM otherwise!

liushuyu:
> @liushuyu i have updated your win64 test case here so you might like to review this.

That test only applies to x86_64 (in the reference `rustc` implementation, running this example will fail and produce an error).

I think the modification is acceptable, though.

CohenArthur (approved):

PR #1454, opened 2022-08-10
intrinsics: Refactor compilation of intrinsics

CohenArthur:
This commit adds helpers around caching, compiling and using intrinsic
functions before extending them with instructions specific to the
intrinsic. We can refactor them further, but this is already a bit
easier to deal with when writing new intrinsics


philberty (approved):
LGTM. Nice work, it was much easier to review this by looking at the file instead of the diff.

PR #1458, opened 2022-08-11
Refactor type resolution pass to use fine grained visitors

philberty:
Thanks for @dkm work on fine-grained HIR visitors this patch takes advantage of this
and moves most implementation into CC files and takes advantage of fine-grained visitors
this will help with the maintainability of the code and it also makes it clear what classes do
and don't do.

It's part of an incremental refactor to get rid of the awkward top-level scan pass in the
type resolution pass. This will allow us to move towards a more query-based model
like the code generation pass so that we can support the forward declared items.

The problem is that we our name resolution pass supports forward declared items but our
type resolution pass does not. We know the HIR nodes when type paths or path expressions
resolve to but when it comes to actually resolve this we just fail when we reuse our mappings
class to look up the item.


CohenArthur (approved):
Looks good, I haven't looked into too much detail as that's a really big refactor haha. But it is cleaner

PR #1459, opened 2022-08-11
intrinsics: Add copy_nonoverlapping<T>

### LINKED ISSUE #1450: Implement copy_nonoverlapping intrinsic

CohenArthur:
intrinsics: Add copy_nonoverlapping<T>

This intrinsic is similar to C's memcpy (or in our case, GCC's
__builtin_memcpy) with the order of arguments swapped and knowledge
about the type of the operands. So we can desugar the following calls:

`copy_nonoverlapping::<T>(src, dst, count)`

can be converted to

`__builtin_memcpy(dst, src, count * size_of::<T>())`

~~Sadly, calling this intrinsic results in calls being optimized out for some reason. I'll open a separate issue to deal with that as I don't have the skills to resolve it.~~ fixed by not marking the function's declaration as `TREE_READONLY`

Fixes #1450

### DIFF
--- a/gcc/rust/backend/rust-builtins.h
+++ b/gcc/rust/backend/rust-builtins.h
@@ -122,6 +122,14 @@ private:
     define_builtin ("breakpoint", BUILT_IN_TRAP, "__builtin_trap", "breakpoint",
 		    build_function_type (void_type_node, void_list_node),
 		    builtin_const | builtin_noreturn);
+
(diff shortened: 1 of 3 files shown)


philberty:
If you rebase this on top of #1462 you should be able to close the issue

philberty (approved):
LGTM

PR #1462, opened 2022-08-11
Ensure we set TREE_SIDE_EFFECTS on call expressions

### LINKED ISSUE #1460: Mutable intrinsic with side effects gets optimized out
I tried this code:

```rust
extern "rust-intrinsic" {
    pub fn copy_nonoverlapping<T>(src: *const T, dst: *mut T, count: usize);
}

fn main() -> i32 {
    let i = 15;
    let mut i_copy = 16;

    let i = &i as *const i32;
    let i_copy = &mut i_copy as *mut i32;

    unsafe {
        copy_nonoverlapping(i, i_copy, 1);

        *i_copy - *i
    }
}

```

I expected to see this happen: A copy of `15` into the `i_copy` value, which would result in an exit code of 0

Instead, this happened: A non zero exit code because `i_copy` kept its value of `16`.

The call to `copy_nonoverlapping` got optimized away at some point as the generated function declaration is correct.

Original tree dump:

```c
void copy_nonoverlapping<i32> (const i32 * const src, i32 * const dst, const usize count)
{
  __builtin_memcpy (dst, src, 4 * count);
}


__attribute__((cdecl))
i32 test::main ()
{
  const i32 i = 15;
  i32 i_copy = 16;
  const i32 * const i = &i;
  i32 * const i_copy = &i_copy;

    i32 RUSTTMP.0;
    const i32 i = 15;
    i32 i_copy = 16;
    const i32 * const i = &i;
    i32 * const i_copy = &i_copy;
    i32 RUSTTMP.1;
  {
    RUSTTMP.1 = *i_copy - *i;
  }
  return RUSTTMP.1;
}
```

Note the missing call to `copy_nonoverlapping` in the last block, despite it not being a pure function and not being allowed to be optimized away.

Equivalent proper Rust test case with proper main signature and printing:

```rust
use std::intrinsics::copy_nonoverlapping;

fn main() {
    let i = 15;
    let mut i_copy = 16;

    let i = &i as *const i32;
    let i_copy = &mut i_copy as *mut i32;

    unsafe {
        copy_nonoverlapping(i, i_copy, 1);

        println!("{}", *i_copy - *i);
    }
}
```

This prints zero, indicating the copy operation is correct 

### Meta

- What version of Rust GCC were you using: [2bbffdb](https://github.com/Rust-GCC/gccrs/pull/1459/commits/2bbffdb6c89b3e37e86fcd3e252eeeaf25ca216f)

philberty:
When we compiling call expressions the GCC helper has quite a simple
implementation to determine if there are SIDE_EFFECTS we have more
information in the front-end to determine this. This adds a check that
the call expression is deemed to have no side effects ensure we check this
against the fndecl used.

Fixes #1460

### DIFF
--- a/gcc/rust/rust-gcc.cc
+++ b/gcc/rust/rust-gcc.cc
@@ -1848,6 +1848,9 @@ Gcc_backend::call_expression (tree fn, const std::vector<tree> &fn_args,
       ret = build1_loc (location.gcc_location (), NOP_EXPR, rettype, ret);
     }
 
+  if (!TREE_SIDE_EFFECTS (ret))
+    TREE_SIDE_EFFECTS (ret) = TREE_SIDE_EFFECTS (fndecl);
+
   delete[] args;
   return ret;
 }


CohenArthur (approved):
LGTM

PR #1464, opened 2022-08-11
Remove unused include in HIR

philberty:
This is a hang over from the AST bootstrap of the HIR tree's


CohenArthur (approved):
Looks good. I wonder if it would be worth it to split the expansion pass into a CfgVisitor and an ExpansionVisitor. But we can think about that later on

PR #1465, opened 2022-08-11
intrinsics: Add wrapping_{add, sub, mul}

### LINKED ISSUE #1449: Implement wrapping_add intrinsic

CohenArthur:
Since wrapping arithmetics are guaranteed in Rust, we turn on the -fwrapv and simply desugar wrapping_{add, sub, mul} to their non-checked inner operations. This is the only difference between a wrapping add and a regular addition: The regular addition will gain some checks for overflows, which are simply not used for the wrapping version.

Fixes #1449 

@bjorn3 if you want to have a look :)

### DIFF
--- a/gcc/rust/rust-lang.cc
+++ b/gcc/rust/rust-lang.cc
@@ -152,8 +152,14 @@ grs_langhook_option_lang_mask (void)
 
 /* Initialize the options structure. */
 static void
-grs_langhook_init_options_struct (struct gcc_options * /* opts */)
+grs_langhook_init_options_struct (struct gcc_options *opts)
 {
+  /* Operations are always wrapping in Rust, even on signed integer. This is
+   * useful for the low level wrapping_{add, sub, mul} intrinsics, not for
+   * regular arithmetic operations which are checked for overflow anyway using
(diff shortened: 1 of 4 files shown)


gcc/rust/rust-lang.cc
```diff
@@ -152,8 +152,14 @@ grs_langhook_option_lang_mask (void)
...
 /* Initialize the options structure. */
 static void
-grs_langhook_init_options_struct (struct gcc_options * /* opts */)
+grs_langhook_init_options_struct (struct gcc_options *opts)
 {
+  /* Operations are always wrapping in Rust, even on signed integer. This is
+   * useful for the low level wrapping_{add, sub, mul} intrinsics, not for
+   * regular arithmetic operations which are checked for overflow anyway using
+   * builtins */
+  opts->x_flag_wrapv = 1;
```
philberty, line 161: We should make a note to document when we enable these by default

philberty (approved):
LGTM we could probably think about starting a .md document inside the front-end folder about language options and defaults something. We will eventually need to document all of this in that weird texi format at some stage next year.

PR #1466, opened 2022-08-11
Dont return error_mark_node on bad CallExpr arguments

philberty:
If we return error_mark_node on a bad expression argument we loss alot of
useful debug info from the error node within the -fdump-tree-original which
saves alot of debug printf.


dafaust (approved):
`useful debug info from the error node within the -fdump-tree-original`
wow I did not realize we were missing out on this.
LGTM!!

PR #1467, opened 2022-08-11
Desugar HIR::IdentifierExpr into HIR::PathInExpression

### LINKED ISSUE #1456: Remove `IdentifierExpr`
`IdentifierExpr`s are our hack, and it would be good to remove eventually in favor of just using Path expressions. 

https://github.com/Rust-GCC/gccrs/blob/eca2ac2c23e0c8b438fd696d4f85e35c9210d8dd/gcc/rust/hir/tree/rust-hir.h#L370-L374

https://github.com/Rust-GCC/gccrs/blob/eca2ac2c23e0c8b438fd696d4f85e35c9210d8dd/gcc/rust/parse/rust-parse-impl.h#L12525-L12533

According to [the reference](https://doc.rust-lang.org/reference/expressions/path-expr.html), references to local variables etc. are just Path Expressions; there is no dedicated Identifier Expression.

The use of both in our codebase is sometimes confusing, for example in the following:
```rust
let x = (Foo::A, 5u8);
match x {
   ...
}
```
the use of `x` in the match scrutinee is not an `IdentifierExpr` but rather a Path expression.

As philbert noted on IRC, it will take some work to remove them and fix regressions which pop up, because `IdentifierExpr`s don't infer generic arguments currently while Paths do.

philberty:
This completly removes the HIR::IdentifierExpr and unifies how we handle
generics in general. There was a hack from last year that did not infer
generic arguments on IdentifierExpr's which leads to a type inferencing
behvaiour mismatch which was becoming difficult to debug. This simplifies
everything.

The changes to the test case reflect making our code more compliant to
real rustc apart from compile/traits3.rs which will be fixed as part of the
refactoring effort going on in the type system.

Fixes #1456

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -692,12 +692,6 @@ CompileExpr::visit (HIR::MatchExpr &expr)
 	  }
 	  break;
 
-	  case HIR::Expr::ExprType::Ident: {
-	    // FIXME
-	    gcc_unreachable ();
-	  }
-	  break;
-
 	  case HIR::Expr::ExprType::Path: {
 	    // FIXME
 	    gcc_unreachable ();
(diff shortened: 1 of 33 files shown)


gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -692,12 +692,6 @@ CompileExpr::visit (HIR::MatchExpr &expr)
 	  }
 	  break;
 
-	  case HIR::Expr::ExprType::Ident: {
```
philberty, line 695: Is this the piece you were asking about @dafaust :D hopefully this helps you

dafaust: Yes exactly! I couldn't find a case where we would hit this and not just the `ExprType::Path` below so I started wondering :smile:

dafaust (approved):
Wow this looks great!
I started poking at the same thing but my first take was from the opposite end, by removing the AST `IdentifierExpr` and changing the parser where they were returned, this is much better.

gcc/testsuite/rust/compile/torture/issue-1075.rs
```diff
@@ -19,8 +19,7 @@ impl<T> *const [T] {
     pub const fn len(self) -> usize {
         // SAFETY: this is safe because `*const [T]` and `FatPtr<T>` have the same layout.
         // Only `std` can make this guarantee.
-        let a = unsafe { Repr { rust: self }.raw };
-        a.len
+        unsafe { Repr { rust: self }.raw.len }
```
CohenArthur, line 22: Yaaaaaay glad to see this work :DD

CohenArthur (approved):
Nice job :)

PR #1468, opened 2022-08-11
Revert "Ensure we set TREE_SIDE_EFFECTS on call expressions"

philberty:
This reverts commit f891a13aa2af912584176d2c595c295b77ff5ff9. This fix
ensures the code is added to the block but still as the artifical function
is marked READONLY it ends up being optimized out anyway.


CohenArthur (approved):
I've confirmed that #1450 works when rebased on top of this. Thank you for all the effort you put into this bug haha.

bors r+

PR #1471, opened 2022-08-12
Remove target hooks changes

philberty:
This reverts the unnecessary changes we have made to GCC
so that merging from upstream should be simplified. This is
also making writing the patches for GCC simpler.


philberty:
hmm i didnt get these failures locally I must try and clean build dir for this locally.

philberty:
The expected warnings needed to be updated post the reset of the genhooks.cc

CohenArthur (approved):
Ah :D Good catch on the warnings

PR #1473, opened 2022-08-16
Improve AST dump

CohenArthur:
Let's abuse C++ a little bit to make dumping and the chaining of dumping operations easier. If anyone has input feel free, this is definitely a roundabout way to go about it.

This also adds pretty printing of the expanded AST, which can be useful to debug macros or `cfg` invocations


philberty (approved):
LGTM

PR #1478, opened 2022-08-17
RFC: arrange tuple matches in trees not HIR

dafaust:
Opening this as draft because I'm not sure whether it is the best path forward,
and I'd like to see if anyone has thoughts one way or another.

This commit changes the way we compile matches on tuple scrutinees to do
as much work as possible on GENERIC trees, rather than first rearranging
the HIR representation of the match.

The pros of this change:
 + Much easier to debug and inspect trees as they are constructed
 + More GCC-y (?)
 + Exactly the same process for tuple and paths-to-tuples as scrutinees
    (see the added test, this is newly-working)
 + Should extend a little more directly to other types of scrutinees,
   such as structs

cons:
 - More boilerplate-ish code repeated, should be fixed with some
   refactoring
 - ??

In any case, there remains a LOT more to do on match expressions, and
probably by the end the whole approach will be entirely different.

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -993,17 +993,14 @@ struct PatternMerge
 // properly comparing patterns, and then use an actual map.
 //
 static struct PatternMerge
-sort_tuple_patterns (HIR::MatchExpr &expr)
+sort_tuple_patterns (std::vector<HIR::MatchCase> cases)
 {
-  rust_assert (expr.get_scrutinee_expr ()->get_expression_type ()
-	       == HIR::Expr::ExprType::Tuple);
-
   struct PatternMerge result;
   result.wildcard = nullptr;
   result.heads = std::vector<std::unique_ptr<HIR::Pattern>> ();
   result.cases = std::vector<std::vector<HIR::MatchCase>> ();
 
-  for (auto &match_case : expr.get_match_cases ())
+  for (auto &match_case : cases)
     {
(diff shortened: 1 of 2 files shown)


dafaust:
> This is probably where we need to add our own custom TREE_CODE for MATCH_EXPR. The only problem I see with this patch currently is that we might miss some abstract stuff like the ref keyword on match arms. 

I haven't looked much into a custom TREE_CODE yet but I think you might be right. I am going to look into that more seriously it could make our lives a lot easier here.

I hadn't thought about the ref keyword specifically yet but there is definitely some stuff I have been not worrying about because I don't have good ideas for how to handle them yet, and been focused on some of the more obvious things first. Definitely agree that if we take this approach we'll need to be careful not to lose track of stuff from the HIR that doesn't necessarily translate directly into trees.

>The main benefit here is that constructing the HIR is really awkward and tree would be easier. Especially for pattern exhaustiveness checking. I think there is a GCC SWITCH_STMT_ALL_CASES_P or something that we might be able to reuse code from other front-ends too.

Yes for sure I find it is easier to construct the trees than full new HIR nodes, and a lot easier to examine them thanks to gcc's existing debug prints for trees. But, related to the above, it is easy to miss some of the more abstract rust-y stuff that can come up in matches. I didn't think of the benefit to pattern exhaustiveness checking.

> 
> Overall i am in favour of this change though match expressions are getting complex so you might want to pull this stuff into its own file and header now.

Yep will do. There is a lot of refactoring and cleanup to be done with this code. And a lot more to be added.

Issue #1476, opened 2022-08-16
Tuple patterns with wildcards fail type resolution


dafaust:
I tried this code:

```rust
fn bar (x: u8, y: u8) -> i32 {
    match (x, y) {
        (1, 1) => { return 1; }
        (1, _) => { return -1; }
        _ => { return 0; }
    }
}
```

I expected to see this happen: compiles

Instead, this happened: 
```
bug.rs:4:13: error: expected [u8] got [(u8, u8, )]
    2 |     match (x, y) {
      |               ~
    3 |         (1, 1) => { return 1; }
    4 |         (1, _) => { return -1; }
      |             ^
bug.rs:2:5: error: failed to type resolve expression
    2 |     match (x, y) {
      |     ^
bug.rs:1:1: error: expected [i32] got [<tyty::error>]
    1 | fn bar (x: u8, y: u8) -> i32 {
      | ^                        ~
```

### Meta

- What version of Rust GCC were you using, git sha if possible.
4665766d3cc9854d1609b71f673e96e0a340805c

PR #1491, opened 2022-08-18
typecheck: resolve tuple pattern elt against parent elt

dafaust:
When doing type resolution for an element of a tuple pattern, check it
against the corresponding element of the parent, if the parent is also a
tuple.

Fixes: #1476


CohenArthur (approved):
LGTM. Is there maybe an error test case that you could add?

philberty (commented):
LGTM nice patch

PR #1492, opened 2022-08-19
Redo coercion site code

philberty:
This gets rid of the old visitor method and brings us much closer to the
Rustc rules which from the algo mentioned in the comment's do the checks
in a very specific order which we need to match.


gcc/rust/typecheck/rust-coercion.cc
```diff
@@ -22,59 +22,275 @@ namespace Rust {
...
+      needs_reborrow = true;
+      expected_mutability = to_mutbl;
+
+      adjustments.push_back (
+	Adjustment (Adjustment::AdjustmentType::INDIRECTION, source_ref, ty_a));
+    }
+
+  // there is a bunch of code to ensure something is coerce able to a dyn trait
+  // we need to support but we need to support a few more lang items for that
+  // see:
```
dafaust, line 249: Might be good to stick a FIXME on this

dafaust (approved):
LGTM I like how you've split up the coercion functions. Very easy to follow.
Love to see all the deletions :)

gcc/rust/typecheck/rust-coercion.cc
philberty: Fixed :D

philberty:
Thanks for the review @dafaust I think with some tweaks we can do something similar for casts next which will be exciting :D.

PR #1493, opened 2022-08-22
Fix ICE in dyn impl block

### LINKED ISSUE #1249: ICE on `impl dyn Trait`
When trying to add an `impl` on a `dyn Trait` gccrs produces an ICE which appears to come from the parser.

### Code

```rust
trait T {}

impl dyn T {}
```

[Godbolt](https://godbolt.org/z/oYc1h8bYe)

### Meta

```
gccrs (Compiler-Explorer-Build-gcc-74e836599ce80a11b1fe28065ed7aae6ffa3b7e2-binutils-2.38) 12.0.1 20220118 (experimental)
```

### Error output

```
rust1: internal compiler error: in append, at rust/util/rust-canonical-path.h:90
```

<details><summary>Backtrace</summary>
  <p>
  
  ```
  0x2061c39 internal_error(char const*, ...)
	  ???:0
  0x8070ef fancy_abort(char const*, int, char const*)
	  ???:0
  0x9a2318 Rust::Resolver::CanonicalPath::append(Rust::Resolver::CanonicalPath const&) const
  	???:0
  0x9a2cee Rust::Resolver::ResolveTopLevel::visit(Rust::AST::InherentImpl&)
	  ???:0
  0x9a01a3 Rust::Resolver::NameResolution::go(Rust::AST::Crate&)
	  ???:0
  0x8cdf38 Rust::Session::parse_file(char const*)
	  ???:0
  0x8ceb28 Rust::Session::parse_files(int, char const**)
	  ???:0
  Compiler returned: 1
  ```
  
  </p>
</details>

philberty:
The issue here was that our name resolution assumed qualified path always
has an as clause. We also had missing visitors to generate the canonical path
for impl blocks using dynamic trait objects.

We should investigate implementing the fine-grained visitors for the AST
similar to the ones we have for HIR. This will solve the issue where we
have missing visitors.

Fixes #1249

### DIFF
--- a/gcc/rust/resolve/rust-ast-resolve-type.cc
+++ b/gcc/rust/resolve/rust-ast-resolve-type.cc
@@ -477,6 +477,27 @@ ResolveTypeToCanonicalPath::visit (AST::SliceType &type)
     }
 }
 
+void
+ResolveTypeToCanonicalPath::visit (AST::TraitObjectTypeOneBound &type)
+{
+  CanonicalPath path = CanonicalPath::create_empty ();
+  bool ok
+    = ResolveTypeToCanonicalPath::go (&type.get_trait_bound ().get_type_path (),
+				      path);
+  if (ok)
+    {
+      std::string slice_path = "<dyn " + path.get () + ">";
+      result = CanonicalPath::new_seg (type.get_node_id (), slice_path);
+    }
+}
+
+void
+ResolveTypeToCanonicalPath::visit (AST::TraitObjectType &type)
+{
+  // FIXME is this actually allowed? dyn A+B
+  gcc_unreachable ();
+}
+
 ResolveTypeToCanonicalPath::ResolveTypeToCanonicalPath ()
   : ResolverBase (), result (CanonicalPath::create_empty ())
 {}
--- a/gcc/rust/resolve/rust-ast-resolve-item.cc
+++ b/gcc/rust/resolve/rust-ast-resolve-item.cc
@@ -534,25 +534,17 @@ ResolveItem::visit (AST::InherentImpl &impl_block)
 
   // FIXME this needs to be protected behind nominal type-checks see:
   // rustc --explain E0118
-
-  NodeId resolved_node = ResolveType::go (impl_block.get_type ().get ());
-  if (resolved_node == UNKNOWN_NODEID)
-    {
-      resolver->get_type_scope ().pop ();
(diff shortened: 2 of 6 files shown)


gcc/rust/resolve/rust-ast-resolve-type.cc
```diff
@@ -477,6 +477,27 @@ ResolveTypeToCanonicalPath::visit (AST::SliceType &type)
...
+      std::string slice_path = "<dyn " + path.get () + ">";
+      result = CanonicalPath::new_seg (type.get_node_id (), slice_path);
+    }
+}
+
+void
+ResolveTypeToCanonicalPath::visit (AST::TraitObjectType &type)
+{
+  // FIXME is this actually allowed? dyn A+B
+  gcc_unreachable ();
```
CohenArthur, line 498: Do you mean this?

```rust
trait Foo {}
trait Bar {}

fn takes_dyn_foo_bar(_: &(dyn Foo + Bar)) {} // invalid
fn takes_dyn_foo_send(_: &(dyn Foo + Send)) {} // okay
```

The error message for the first functions is as follows:

```rust
error[[E0225]](https://doc.rust-lang.org/stable/error-index.html#E0225): only auto traits can be used as additional traits in a trait object
 --> src/lib.rs:4:37
  |
4 | fn takes_dyn_foo_bar(_: &(dyn Foo + Bar)) {} // invalid
  |                               ---   ^^^ additional non-auto trait
  |                               |
  |                               first non-auto trait
  |
  = help: consider creating a new trait with all of these as supertraits and using that trait here instead: `trait NewTrait: Foo + Bar {}`
  = note: auto-traits like `Send` and `Sync` are traits that have special properties; for more information on them, visit <https://doc.rust-lang.org/reference/special-types-and-traits.html#auto-traits>
  ```
  
  So it's allowed in some cases :)

CohenArthur (approved):

gcc/rust/resolve/rust-ast-resolve-type.cc
philberty: ah ok thats interesting thats the same error for impl dyn A+B too.

philberty: That error handling must be during type resolution not 100% sure why this is the case in Rust but there is definitely a reason. We have an open bug about handling super traits in dyn objects which i have been meaning to fix for ages i bet its got to do with that.

PR #1497, opened 2022-08-23
Refactor our casts to follow the Rustc implemention

### LINKED ISSUE #1496: Casts can try a coercion
I tried this code:

```rust
extern "C" {
    fn printf(s: *const i8, ...);
}

struct Foo<T>(T);
impl<T> core::ops::Deref for Foo<T> {
    type Target = T;

    fn deref(&self) -> &Self::Target {
        unsafe {
            let a = "foo_deref\n\0";
            let b = a as *const str;
            let c = b as *const i8;

            printf(c);
        }

        &self.0
    }
}

fn main() {
    let foo = Foo(123);
    let bar = &foo as &i32;

    unsafe {
        let a = "%i\n\0";
        let b = a as *const str;
        let c = b as *const i8;

        printf(c, *bar);
    }
}
```

I expected to see this happen: this casts akin to a normal coercion

Instead, this happened:  https://godbolt.org/z/cWPd9cMs7

```
      foo.0 = 123;
      bar = &foo;
```

### Meta

philberty:
This gets rid of our old visitor system for cast type checking. Casts
depend on type coercions as they are meant to attempt a type coercion
before trying a simple cast. This explicitly defines the rules which should
be allowed for simple casts. In rustc they use match expressions to write
a list of casts which should not be allowed. We have likely missed some
rules of what should be allowed but this is at least the start of how
to implement this.

Fixes #1496


dafaust (approved):
LGTM I am a big fan of these refactors, this is a lot easier to follow IMO as well as making the rules for simple casts obvious :+1:

gcc/rust/typecheck/rust-casts.cc
```diff
@@ -0,0 +1,292 @@
...
+
+    case TyTy::TypeKind::REF:
+    case TyTy::TypeKind::POINTER:
+      switch (to.get_ty ()->get_kind ())
+	{
+	case TyTy::TypeKind::REF:
+	case TyTy::TypeKind::POINTER:
+	  return check_ptr_ptr_cast ();
+
+	  // FIXME can you cast a pointer to a integral type?
```
CohenArthur, line 230: Yes!
```rust
(Ptr(m_expr), Int(_)) => self.check_ptr_addr_cast(fcx, m_expr), // ptr-addr-cast
(FnPtr, Int(_)) => Ok(CastKind::FnPtrAddrCast),

// * -> ptr
(Int(_), Ptr(mt)) => self.check_addr_ptr_cast(fcx, mt), // addr-ptr-cast
```

CohenArthur (commented):
Very good cleanup :) LGTM

PR #1499, opened 2022-08-23
Constant folding in gccrs: port over rest of the code from CP frontend

abbasfaisal:


gcc/rust/backend/rust-compile-base.cc
```diff
@@ -695,6 +701,8 @@ HIRCompileBase::compile_constant_item (
 
       gcc_assert (TREE_CODE (bind_tree) == BIND_EXPR);
       DECL_SAVED_TREE (fndecl) = bind_tree;
+      DECL_DECLARED_CONSTEXPR_P (fndecl);
```
CohenArthur, line 704: Shouldn't this be
```suggestion
      DECL_DECLARED_CONSTEXPR_P (fndecl) = 1;
```
?

gcc/rust/backend/rust-compile-intrinsic.cc
```diff
@@ -213,6 +214,9 @@ finalize_intrinsic_block (Context *ctx, tree fndecl)
   DECL_SAVED_TREE (fndecl) = bind_tree;
 
   ctx->push_function (fndecl);
+
+  DECL_DECLARED_CONSTEXPR_P (fndecl);
```
CohenArthur, line 218: Same here

CohenArthur (approved):
Thank you for all this work. I think overall if this doesn't cause any regression, we should merge it and then start adding test cases and fixing stuff. This is a really good start in my opinion

gcc/rust/backend/rust-compile-intrinsic.cc
abbasfaisal: I think you are correct about the first one as that seems to be dealing with compiling a constant item, not sure about this one as this one gets called from quite a few places. I think @philberty would know better as I have never touched this file :)

philberty: Yeah Arthur is right here it should be =1 this is my mistake.

philberty:
Looks like we have 1 failure for -m32 tests let me know how you get on with that assertion.

philberty:
I found the solution to the regression test case:

```
diff --git a/gcc/rust/backend/rust-constexpr.cc b/gcc/rust/backend/rust-constexpr.cc
index ed0fc2ce345..77e8a059465 100644
--- a/gcc/rust/backend/rust-constexpr.cc
+++ b/gcc/rust/backend/rust-constexpr.cc
@@ -2080,15 +2080,7 @@ eval_constant_expression (const constexpr_ctx *ctx, tree t, bool lval,
                                    non_constant_p, overflow_p);
       break;
 
-    case NOP_EXPR:
-      if (REINTERPRET_CAST_P (t))
-       {
-         if (!ctx->quiet)
-           error_at (loc, "%<reinterpret_cast%> is not a constant expression");
-         *non_constant_p = true;
-         return t;
-       }
-      /* FALLTHROUGH.  */
+
     case MODIFY_EXPR:
       r = eval_store_expression (ctx, t, false, non_constant_p, overflow_p);
       break;
@@ -2347,6 +2339,7 @@ eval_constant_expression (const constexpr_ctx *ctx, tree t, bool lval,
       break;
 
       /* FALLTHROUGH.  */
+    case NOP_EXPR:
     case CONVERT_EXPR:
       case VIEW_CONVERT_EXPR: {
        tree oldop = TREE_OPERAND (t, 0);
```

This was a small mistake in the porting. The reason this only happens on 32bit mode is because of:  ba75fe35b53d7531464cf73a4ed52257fe176a1e

On 32bit mode floats have to be handled slightly differently, resulting in this NOP conversion going on.

philberty:
I will leave it up to @abbasfaisal to apply that diff and @CohenArthur to fix up the bors warnings list. I will get back to you with an example test case tomorrow.

gcc/rust/backend/rust-compile-base.cc
abbasfaisal: Fixed.

philberty:
Lets see if this builds now :D

philberty:
I think i have a new fix for the warnings incoming :D

gcc/testsuite/rust/compile/const5.rs
```diff
@@ -0,0 +1,5 @@
+// { dg-options "-w -O0 -fdump-tree-gimple" }
+const A: usize = 123;
+const B: [i32; 5] = [1, 2, 3, 4, 5];
+const C: i32 = B[A];
+// { dg-error "array subscript value .123. is outside the bounds of array" "" { target *-*-* } .-1 }
```
philberty, line 5: here is another cool feature :D @abbasfaisal

philberty:
woohoo looks like its worked!

PR #1501, opened 2022-08-23
Fix port of NOP_EXPR

philberty:
NOP_EXPR should fall through into convert and view_convert_exprs.


abbasfaisal:
There is another NOP_EXPR in potential_constant_expression_1(). Does that also need fixing? (though I have [fixed](https://github.com/Rust-GCC/gccrs/pull/1499/commits/7f5ee8522f70443b8b164476d05828293f8102c1#diff-4a76572d9d397064c76a83ff737f4054b3bd79d5400029ee84f43caa5d18e172L5798) it already as I didn't know you had raised this PR. But I forgot about this one in that PR so this PR is still valid).

Otherwise LGTM!

PR #1503, opened 2022-08-24
Add overflow traps

CohenArthur:
Opening as a draft since the code is really ugly and some tests still fail. I am looking for feedback on the implementation and my approximative use of functions like `temporary_variable` which I feel I might be using the wrong way.

I'll open up an issue of what still needs to be done, namely:
- [x] Properly handle sub and mul operations
- [ ] Disable the check in `release` mode (`-frust-disable-overflow-checks`)
- [ ] Handle specific rust overflow attribute (needs checkup on the name)

I'll open up some PRs to clean up some parts of the backend as well.

### DIFF
--- a/gcc/rust/rust-gcc.cc
+++ b/gcc/rust/rust-gcc.cc
@@ -50,37 +50,9 @@
 #include "rust-linemap.h"
 #include "rust-backend.h"
 #include "rust-object-export.h"
+#include "rust-gcc.h"
 
 #include "backend/rust-tree.h"
-
-// TODO: this will have to be significantly modified to work with Rust
-
-// Bvariable is a bit more complicated, because of zero-sized types.
-// The GNU linker does not permit dynamic variables with zero size.
-// When we see such a variable, we generate a version of the type with
-// non-zero size.  However, when referring to the global variable, we
-// want an expression of zero size; otherwise, if, say, the global
-// variable is passed to a function, we will be passing a
-// non-zero-sized value to a zero-sized value, which can lead to a
-// miscompilation.
-
-class Bvariable
-{
-public:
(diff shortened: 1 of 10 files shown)


dafaust (approved):
I think this is fine. I guess you are basically building:
```
var tmp = lhs;
if (builtin_op_overflow (lhs, rhs, &tmp) == true /* tmp = lhs op rhs overflowed */)
    abort ();
tmp
```
(though I think *technically* the `COND_EXPR` is more the `?:` formation in C)

I'm sure there are other equivalent expressions that could be built to accomplish the same thing but this way seems fine to me. And, well, if it's compiling and working as expected that's certainly a good sign :)
LGTM

gcc/rust/rust-gcc.cc
```diff
@@ -1437,21 +1415,127 @@ Gcc_backend::arithmetic_or_logical_expression (ArithmeticOrLogicalOperator op,
...
+
+  return {abort, builtin};
+}
+
+// Return an expression for the arithmetic or logical operation LEFT OP RIGHT
+// with overflow checking when possible
+tree
+Gcc_backend::arithmetic_or_logical_expression_checked (
+  ArithmeticOrLogicalOperator op, tree left, tree right, Location location,
+  Bvariable *receiver_var)
```
philberty, line 1494: How about updating the signature of this function to take a flag like 'enable_overflow_checks'. Eventually, it will need to take a tree for abort-mode too. So if it takes a method for enabling overflow checks, we can control how they are enabled, which would be helpful.

philberty (changes requested):
I think it would be idea to add a parameter here even though for this PR we will simply be passing true for each usage.

philberty (approved):
LGTM after talking about this as overflow checks only apply to add,sub and mul this needs some more thought and refactoring

PR #1505, opened 2022-08-25
Create canonical process of compiling constant items

philberty:
In order to compile a block expression constant, the simplest way for us
was to reuse what code we have and to generate an artifical function which
does not get added to the translation unit. The constant then becomes
a CALL_EXPR to this artifical function which we can pass to the constexpr
evaluator to resolve the result of this artifical 'CALL_EXPR'.

Before this patch we seperated the difference between block expressions
and non block expressions in constants. So for non block expressions we
simply compiled them as if it was a simple constant but this is not
guaranteed to be the case in rust, for example coercion sites can generate
temporaries during autoderef which we let the constant evaluator resolve
for us. This makes all constants handled in the same way to simplify the
logic here.

### DIFF
--- a/gcc/rust/backend/rust-compile-base.cc
+++ b/gcc/rust/backend/rust-compile-base.cc
@@ -652,65 +652,72 @@ HIRCompileBase::compile_constant_item (
   bool is_block_expr
     = const_value_expr->get_expression_type () == HIR::Expr::ExprType::Block;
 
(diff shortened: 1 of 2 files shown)


CohenArthur (approved):
LGTM

abbasfaisal:
LGTM

PR #1510, opened 2022-08-26
rustc_attrs: Allow `rustc_inherit_overflow_checks` as a builtin attribute

CohenArthur:
We cannot yet handle this attribute, but we should not reject it either.


philberty (approved):
LGTM

PR #1511, opened 2022-08-26
lint: Do not emit unused warnings for public items

CohenArthur:
This fixes the overzealous warnings on public items from the unused pass


philberty (approved):
LGTM but lets give @thomasyonug a chance to review as well :)

PR #1513, opened 2022-08-29
macros: Handle matchers properly in repetitions

CohenArthur:
This fixes an issue encountered when trying to compile libcore 1.29

### DIFF
--- a/gcc/testsuite/rust/compile/macro43.rs
+++ b/gcc/testsuite/rust/compile/macro43.rs
@@ -1,0 +1,64 @@
+macro_rules! nonzero_integers {
+    ( $( $Ty: ident($Int: ty); )+ ) => {
+        $(
+            /// An integer that is known not to equal zero.
+            ///
+            /// This enables some memory layout optimization.
+            /// For example, `Option<NonZeroU32>` is the same size as `u32`:
+            ///
+            /// ```rust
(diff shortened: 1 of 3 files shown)


gcc/testsuite/rust/compile/macro43.rs
```diff
@@ -0,0 +1,64 @@
...
+                /// Returns the value as a primitive type.
+                #[stable(feature = "nonzero", since = "1.28.0")]
+                #[inline]
+                pub fn get(self) -> $Int {
+                    self.0 .0
+                }
+
+            }
+
+            impl_nonzero_fmt! { // { dg-error "unknown macro" }
```
philberty, line 50: was the issue that it crashed here?

philberty (approved):
LGTM

gcc/testsuite/rust/compile/macro43.rs
CohenArthur: No, the issue was in the match arm:
```rust
( $( $Ty: ident($Int: ty); )+ )
```
`Int` was understood as part of a matcher because of the surrounding parentheses (which is valid), but we didn't take into account the fact that it was within a repetition. So matching multiple values for `Int` would trigger the `gcc_unreachable()` call when calling `insert_metavar`

PR #1514, opened 2022-08-29
Fix SEGV in transcribe_type when compiling libcore 1.29

CohenArthur:
Needs #1513, Skip reviewing the first commit

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -24,6 +24,7 @@
 #include "rust-hir-map.h"
 #include "rust-token.h"
 #include "rust-location.h"
+#include "rust-diagnostics.h"
 
 namespace Rust {
 // TODO: remove typedefs and make actual types for these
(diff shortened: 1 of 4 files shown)


gcc/rust/ast/rust-ast.h
```diff
@@ -1868,6 +1869,10 @@ class ASTFragment
    */
 
   bool is_single_fragment () const { return nodes.size () == 1; }
+  bool is_single_fragment (SingleASTNode::NodeType expected) const
```
philberty, line 1872: Not sure about the name of this method. Its a pet nit of mine that I prefer to use overloading only on methods where they already take arguments. It feels like a new method to check if is_single_fragment and that single fragment is of a particular type.

How about `is_single_fragment_of_kind (...)`

gcc/rust/ast/rust-ast.h
```diff
@@ -1874,9 +1874,39 @@ class ASTFragment
     return is_single_fragment () && nodes[0].get_kind () == expected;
   }
 
-  bool is_single_fragment_kind (SingleASTNode::NodeType kind) const
+  void assert_single_fragment (SingleASTNode::NodeType expected) const
```
philberty, line 1877: ah ok so it gets renamed here anyway :)

philberty (approved):
LGTM

gcc/rust/ast/rust-ast.h
CohenArthur: Ah, I see what you mean. I'll rename it. `assert_single_fragment` kinda serves the same purpose but it also crashes if the assertion fails, so it's a bit different from `is_single_fragment`

CohenArthur:
bors retry

PR #1515, opened 2022-08-29
transcriber: Do not infinite loop if the current parsed node is an error

CohenArthur:


gcc/rust/expand/rust-macro-expand.cc
```diff
@@ -727,6 +727,14 @@ parse_many (Parser<MacroInvocLexer> &parser, TokenId &delimiter,
 	break;
 
       auto node = parse_fn ();
+      if (node.is_error ())
+	{
+	  for (auto err : parser.get_errors ())
+	    err.emit_error ();
+
+	  break;
```
philberty, line 735: what about a return instead of a break?

philberty (approved):
LGTM

gcc/rust/expand/rust-macro-expand.cc
CohenArthur: Fixed in https://github.com/Rust-GCC/gccrs/commit/8dfa9676db0f8660c6c745adb59ef6323dca2341 :)

PR #1516, opened 2022-08-29
parser: Parse RangeFullExpr without erroring out

CohenArthur:
Previously the parser would emit the errors despite later allowing the parsed expression to be null.


philberty (approved):
LGTM

PR #1517, opened 2022-08-31
Refactor error handling on unify

philberty:
This is the first pass on a refactor for error handling on unify sites. Eventually we
will be able to get rid of our can_eq interface but this is the first pass to abstract
unify's to use this interface.

Eventually, we will have unify and unify_and_commit. Unify will become our can_eq
interface but will require more helpers around inference variables.

### DIFF
--- a/gcc/rust/typecheck/rust-coercion.cc
+++ b/gcc/rust/typecheck/rust-coercion.cc
@@ -16,6 +16,7 @@
 // along with GCC; see the file COPYING3.  If not see
 // <http://www.gnu.org/licenses/>.
 
+#include "rust-hir-type-check-base.h"
 #include "rust-coercion.h"
 
 namespace Rust {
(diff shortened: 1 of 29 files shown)


gcc/rust/typecheck/rust-coercion.cc
```diff
@@ -183,7 +231,12 @@ TypeCoercionRules::coerce_borrowed_pointer (TyTy::BaseType *receiver,
       return TypeCoercionRules::CoercionResult::get_error ();
     }
 
+  rust_debug ("coerce_borrowed_pointer -- autoderef cycle");
   AutoderefCycle::cycle (receiver);
+  rust_debug ("coerce_borrowed_pointer -- result: [%s] with adjustments: [%zu]",
+	      try_result.is_error () ? "failed" : "matched",
+	      try_result.adjustments.size ());
```
CohenArthur, line 238: I think this might cause issues on 32 bit builds
```suggestion
  rust_debug ("coerce_borrowed_pointer -- result: [%s] with adjustments: [%lu]",
	      try_result.is_error () ? "failed" : "matched",
	      (unsigned long) try_result.adjustments.size ());
```

I believe this is the fix that we had applied? Not sure

gcc/rust/typecheck/rust-hir-dot-operator.cc
```diff
@@ -171,12 +171,20 @@ MethodResolver::select (const TyTy::BaseType &receiver)
     TyTy::FnType *fntype;
   };
 
+  rust_debug ("inherent_impl_fns found {%zu}, trait_fns found {%zu}, "
```
CohenArthur, line 174: Same here

gcc/rust/typecheck/rust-hir-dot-operator.cc
```diff
@@ -171,12 +171,20 @@ MethodResolver::select (const TyTy::BaseType &receiver)
...
+	      "predicate_items found {%zu}",
```
CohenArthur, line 175: Here as well

CohenArthur (approved):
I think this a good cleanup! I have a few nitpicks regarding the format strings but nothing major. I like the changes

gcc/rust/typecheck/rust-coercion.cc
philberty: fixed

PR #1521, opened 2022-09-03
expand: eager evaluate macros inside builtin macros

liushuyu:
- expand: eagerly evaluate the macros inside some of the builtin macros

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -31,6 +31,7 @@ namespace Rust {
 typedef std::string Identifier;
 typedef int TupleIndex;
 struct Session;
+struct MacroExpander;
 
 namespace AST {
 // foward decl: ast visitor
@@ -951,6 +952,8 @@ public:
 
   virtual Location get_locus () const = 0;
 
+  virtual bool is_literal () const { return false; }
+
   // HACK: strictly not needed, but faster than full downcast clone
   virtual bool is_expr_without_block () const = 0;
 
@@ -1471,6 +1474,7 @@ private:
   // One way of parsing the macro. Probably not applicable for all macros.
   std::vector<std::unique_ptr<MetaItemInner> > parsed_items;
   bool parsed_to_meta_item = false;
+  MacroExpander *expander = nullptr;
 
 public:
   std::string as_string () const;
@@ -1495,6 +1499,7 @@ public:
     path = other.path;
     token_tree = other.token_tree;
     parsed_to_meta_item = other.parsed_to_meta_item;
+    expander = other.expander;
 
     parsed_items.reserve (other.parsed_items.size ());
     for (const auto &e : other.parsed_items)
@@ -1522,6 +1527,13 @@ public:
   // TODO: this mutable getter seems kinda dodgy
   SimplePath &get_path () { return path; }
   const SimplePath &get_path () const { return path; }
+
+  void set_expander (MacroExpander *new_expander) { expander = new_expander; }
+  MacroExpander *get_expander ()
+  {
+    rust_assert (expander);
+    return expander;
+  }
 
   void
   set_meta_item_output (std::vector<std::unique_ptr<MetaItemInner> > new_items)
--- a/gcc/rust/expand/rust-macro-builtins.cc
+++ b/gcc/rust/expand/rust-macro-builtins.cc
@@ -17,12 +17,14 @@
 // <http://www.gnu.org/licenses/>.
 
 #include "rust-macro-builtins.h"
+#include "rust-ast.h"
 #include "rust-diagnostics.h"
 #include "rust-expr.h"
 #include "rust-session-manager.h"
 #include "rust-macro-invoc-lexer.h"
 #include "rust-lex.h"
 #include "rust-parse.h"
+#include "rust-attribute-visitor.h"
 
 namespace Rust {
 namespace {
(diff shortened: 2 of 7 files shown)


liushuyu:
The current code is a total mess, I know.

The issue was I discovered some issues in the parser that sometimes parse the expressions incorrectly mid-implementation. So some duck-tape code was thrown in to avoid crashing the parser and ensure no apparent regression was detected by the test suite.

I still opened a PR anyways because I want to collect feedback regarding how to drag the declarative macro expander into the built-in macro expansion. My current approach is to save the `expander` to `MacroInvocData` so that it can be picked up in the built-in expander along with the invocation data.

gcc/testsuite/rust/compile/builtin_macro_concat.rs
```diff
@@ -6,12 +6,13 @@ macro_rules! concat {
...
+    concat!(not_literal); // { dg-error "expected a literal" }
     concat!("message");
     concat!("message",);
     concat!("message", 1, true, false, 1.0, 10usize, 2000u64);
     concat!("message", 1, true, false, 1.0, 10usize, 2000u64,);
-    concat! ("m", not_literal); // { dg-error "argument must be a constant literal" }
-    concat!(not_literal invalid 'm' !!,); // { dg-error "argument must be a constant literal" }
+    concat! ("m", not_literal); // { dg-error "expected a literal" }
+    concat!(not_literal invalid 'm' !!,); // { dg-error "expected token: .,." }
+                                          // { dg-error "expected a literal" "" { target *-*-* } .-1 }
```
liushuyu, line 17: Note that this test case was changed so that it now expects the diagnostics that follow `rustc`'s diagnostics more closely (I'd say 90% similarity).

liushuyu:
Refactored macro expansion and extraction code

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -61,13 +63,115 @@ macro_end_token (AST::DelimTokenTree &invoc_token_tree,
...
+      auto nodes = try_expand_macro_expression (expr.get (), expander);
+      if (nodes.is_error ())
+	{
+	  // not macro
+	  result.push_back (std::move (expr));
+	}
+      else if (!nodes.is_expression_fragment ())
+	{
+	  rust_error_at (expr->get_locus (), "expected expression");
+	  return empty_expr;
```
liushuyu, line 148: I am unsure how to pass the error status to the caller.
So a bogus error may be emitted from the parent function.

gcc/rust/ast/rust-ast.h
```diff
@@ -1523,6 +1528,9 @@ struct MacroInvocData
   SimplePath &get_path () { return path; }
   const SimplePath &get_path () const { return path; }
 
+  void set_expander (MacroExpander *new_expander) { expander = new_expander; }
+  MacroExpander *get_expander () { return expander; }
```
CohenArthur, line 1532: This can produces segfaults if not handled properly. I would suggest changing the `expander` field from
```c++
MacroExpander *expander = nullptr;
```
to
```c++
Optional<MacroExpander&> expander;
```

This is basically the same ergonomis as what you would have with a raw pointer, but forcing you to make sure its content is valid and non-null. The class is defined in `util/rust-optional.h`

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -61,13 +63,115 @@ macro_end_token (AST::DelimTokenTree &invoc_token_tree,
...
+static std::unique_ptr<AST::LiteralExpr>
+try_extract_string_literal_from_fragment (const Location &parent_locus,
+					  std::unique_ptr<AST::Expr> node)
+{
+  auto maybe_lit = static_cast<AST::LiteralExpr *> (node.get ());
+  if (!node || !node->is_literal ()
+      || maybe_lit->get_lit_type () != AST::Literal::STRING)
+    {
+      rust_error_at (parent_locus, "argument must be a string literal");
+      if (node)
```
CohenArthur, line 89: Is this possible?

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -61,13 +63,115 @@ macro_end_token (AST::DelimTokenTree &invoc_token_tree,
...
+  auto vis = Rust::AttrVisitor (*expander);
+  expr->accept_vis (vis);
+  return expander->take_expanded_fragment (vis);
+}
+
+/* Expand and then extract a string literal from the macro */
+
+static std::unique_ptr<AST::LiteralExpr>
+try_extract_string_literal_from_fragment (const Location &parent_locus,
+					  std::unique_ptr<AST::Expr> node)
```
CohenArthur, line 82:
```suggestion
					  std::unique_ptr<AST::Expr>& node)
```

This would enable us to only clone the expression if the expansion is successful.

Then, this requires the following changes...

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -61,13 +63,115 @@ macro_end_token (AST::DelimTokenTree &invoc_token_tree,
...
+  if (!node || !node->is_literal ()
+      || maybe_lit->get_lit_type () != AST::Literal::STRING)
+    {
+      rust_error_at (parent_locus, "argument must be a string literal");
+      if (node)
+	rust_inform (node->get_locus (), "expanded from here");
+      return nullptr;
+    }
+  return std::unique_ptr<AST::LiteralExpr> (
+    static_cast<AST::LiteralExpr *> (node.release ()));
```
CohenArthur, line 94:
```suggestion
  return std::unique_ptr<AST::LiteralExpr> (
    static_cast<AST::LiteralExpr *> (node.clone_expr().release ()));
```

If this pattern of casting from a `std::unique_ptr<T>` to a `std::unique_ptr<U: public T>` happens often, we might think about adding a helper in the `util` folder which performs that `.release()` danse

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -61,13 +63,115 @@ macro_end_token (AST::DelimTokenTree &invoc_token_tree,
...
+  auto result = std::vector<std::unique_ptr<AST::Expr>> ();
+  auto empty_expr = std::vector<std::unique_ptr<AST::Expr>> ();
+
+  auto first_token = parser.peek_current_token ()->get_id ();
+  if (first_token == COMMA)
+    {
+      rust_error_at (parser.peek_current_token ()->get_locus (),
+		     "expected expression, found %<,%>");
+      return empty_expr;
+    }
```
CohenArthur, line 131: Is this not handled by the parser itself already?

gcc/rust/expand/rust-macro-builtins.cc
CohenArthur: Do you mean fetch the errors from the parser? Or return a possible error to the caller of `try_expand_many_expr`? I might be misunderstanding, sorry

CohenArthur (commented):
I think overall this is a very cool PR! Thank you so much for working on this :) Would it be possible to add test-cases or is that not yet possible?

gcc/rust/expand/rust-macro-builtins.cc
liushuyu: > Do you mean fetch the errors from the parser? Or return a possible error to the caller of `try_expand_many_expr`? I might be misunderstanding, sorry

I meant return an error to the caller of `try_expand_many_expr`, since the error should be already emitted in this function,

gcc/rust/expand/rust-macro-builtins.cc
liushuyu: > Is this not handled by the parser itself already?

No. During my testing, the parser will crash if this happens. I did not dig any further since I saw `rustc` doing the same check like this in their codebase.

gcc/rust/expand/rust-macro-builtins.cc
liushuyu: > Is this possible?

Yes, our parser is currently very buggy. All the checks added in my pull requests are necessary to avoid the parser accidentally screwing itself.

Without this check, macros like `() => { }` (incorrect syntax) will resolve to `nullptr` here.

gcc/rust/expand/rust-macro-builtins.cc
liushuyu: This, unfortunately, does not work.

Changing this to a reference type will now require an `lvalue` for the 2nd argument (`node` here), and then:

https://github.com/Rust-GCC/gccrs/blob/4af2d3228e6933f9c0736c9ac587e77aec8f6186/gcc/rust/expand/rust-macro-builtins.cc#L109-L111

`nodes.take_expression_fragment ()` becomes invalid since it cannot be made into an `lvalue` according to the compiler.

I have to do this to workaround the compiler's complaint, but I am not sure if this is semantically correct:

```cpp
  auto expr = nodes.take_expression_fragment ();
  return try_extract_string_literal_from_fragment (
    input_expr->get_locus (), expr);
```

gcc/rust/ast/rust-ast.h
liushuyu: That's a bit problematic.

You see, `Optional<T>` requires `T` to be a complete type. We can't have that since `rust-macro-expand.h` contains `MacroExpander` but also includes `rust-ast.h` itself!

liushuyu:
> I think overall this is a very cool PR! Thank you so much for working on this :) Would it be possible to add test-cases or is that not yet possible?

It's possible, but I want to get the code to pass the review first, and then I will see how brutal the test cases should be :smile:

gcc/rust/ast/rust-ast.h
CohenArthur: Ah, good catch. Then I would simply add a `rust_assert (expander)` before the return statement to avoid segfaults

gcc/rust/expand/rust-macro-builtins.cc
CohenArthur: Yes, this is completely valid. I think if it's not too annoying it would be better to use that, if not feel free to disregard :)

gcc/rust/expand/rust-macro-builtins.cc
CohenArthur: So if the error has already been emitted and you're setting `has_error`, I think it's good enough. Callers are expected to check the value of `has_error` after calling `try_expand_many_expr`. Returning an empty vector is fine :)

CohenArthur (approved):
LGTM! @liushuyu I think this is in a good state to merge now code-wise - I would just like some tests to make sure everything works well and that I can hack on it without fear of breaking anything :DD As soon as there's some we'll merge it :)

liushuyu:
> LGTM! @liushuyu I think this is in a good state to merge now code-wise - I would just like some tests to make sure everything works well and that I can hack on it without fear of breaking anything :DD As soon as there's some we'll merge it :)

I can add some test cases. But it's very late in my timezone at the moment, so probably tomorrow.

CohenArthur:
> > LGTM! @liushuyu I think this is in a good state to merge now code-wise - I would just like some tests to make sure everything works well and that I can hack on it without fear of breaking anything :DD As soon as there's some we'll merge it :)
> 
> I can add some test cases. But it's very late in my timezone at the moment, so probably tomorrow.

Sounds good! Sorry again for the delay, was a crazy two weeks. Thanks for working on this I *really* appreciate it.

liushuyu:
Rebased against the latest `master` branch and add a testcase.

CohenArthur (approved):
Thank you! LGTM :)

bors r+

PR #1527, opened 2022-09-06
rust: Add -frust-compile-until option

CohenArthur:
This option helps ensure that we do not introduce regressions on various
parts of the compilation pipeline. For example, a testcase (or testsuite
from the `testing` project) might pass attribute checking, expansion and
lowering, but fail during typechecking. Should a change suddenly make
that testcase fail expansion, we would not be able to notice it. By
generating tests that run up until expansion, typechecking, compilation
and so forth we ensure that no regressions are added accidentally to
already failing tests/testsuites.

### DIFF
--- a/gcc/rust/lang.opt
+++ b/gcc/rust/lang.opt
@@ -113,6 +113,48 @@ Rust Joined RejectNegative
 
 o
 Rust Joined Separate
+
+frust-compile-until=
+Rust Joined RejectNegative Enum(frust_compile_until) Var(flag_rust_compile_until)
+-frust-compile-until=[ast|attributecheck|expansion|nameresolution|lowering|typecheck|privacy|unsafety|const|copimlation|end]             When to stop in the pipeline when compiling Rust code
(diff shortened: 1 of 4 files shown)


gcc/rust/lang.opt
```diff
@@ -113,6 +113,48 @@ Rust Joined RejectNegative
 
 o
 Rust Joined Separate
+
+frust-compile-until=
+Rust Joined RejectNegative Enum(frust_compile_until) Var(flag_rust_compile_until)
+-frust-compile-until=[2015|2018|2021]             When to stop in the pipeline when compiling Rust code
```
philberty, line 119: looks like the help string is wrong here

philberty (approved):
LGTM apart from the help string in the lang.opt

gcc/rust/lang.opt
CohenArthur: Ah, thank you :) Fixed

PR #1528, opened 2022-09-06
add testcase with struct to test component_ref and constructor codes in eval_constant_expression()

abbasfaisal:


philberty (approved):
nice test case

PR #1532, opened 2022-09-13
attributes: Add #[macro_use] as builtin

### LINKED ISSUE #1531: Fix error when using `#[macro_use]` (unknown attribute)

CohenArthur:
Fixes #1531 

I think the checking for builtin attributes should definitely be moved to the `AttributeChecker` visitor. It's also a bit messy at the moment considering it isn't in effect on *everything*, simply some nodes such as `Module`s. Thoughts?

### DIFF
--- a/gcc/rust/util/rust-attributes.cc
+++ b/gcc/rust/util/rust-attributes.cc
@@ -40,6 +40,7 @@ static const BuiltinAttrDefinition __definitions[]
      {"no_mangle", CODE_GENERATION},
      {"repr", CODE_GENERATION},
      {"path", EXPANSION},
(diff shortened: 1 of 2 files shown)


philberty (approved):
LGTM

CohenArthur:
> Handling the attributes at a higher level seems like a much better idea. The only problem I forsee is that for lang items we map these to DefIds which only exist once HIR lowering occurs. Though we do have NodeId mappings that we could use to work around this. Else we leave the lang item attribute handling during hir lowering and its still all good.

Yes, this seems like a good idea. I'll take a look at cleaning that up at some point.

bors r+

PR #1534, opened 2022-09-14
module lowering: Do not append null pointers as items

### LINKED ISSUE #1533: ICE when compiling macro in nested module architecture
```rust
mod foo {
    mod bar {
        mod baz {
            macro_rules! baz {
                () => {{}};
            }
        }
    }

    macro_rules! foo {
        () => {{}};
    }

    fn foo_f() {
        foo!();
    }

    fn bar_f() {
        baz!();
    }
}

```

CohenArthur:
Some module items do not need to get lowered to HIR such as `macro_rules!` definitions. Hence, module lowering should act the same as crate lowering: Only emplace back the lowered item if it is a valid pointer

Fixes #1533


philberty (approved):
LGTM nice find

PR #1540, opened 2022-09-17
Implement `-frust-incomplete-and-experimental-compiler-do-not-use` flag

CohenArthur:
Remember to change your `./configure` line :)

This is obviously up for discussion and I would be happy to get some more input.

### DIFF
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -51,9 +51,9 @@ GCCRS_D_OBJS = \
 
 RUST_CXXFLAGS = $(CXXFLAGS)
 
-# If the file exists, we're bootstrapping
+# If the file exists, we're bootstrapping. Set a specific flag for the session manager.
 ifneq ($(wildcard ../stage_current)),)
-RUST_CXXFLAGS += -DRUST_RELEASE
+SESSION_FLAGS = -DRUST_BOOTSTRAP_BUILD
 endif
 
 gccrs$(exeext): $(GCCRS_D_OBJS) $(EXTRA_GCC_OBJS) libcommon-target.a $(LIBDEPS)
@@ -173,7 +173,7 @@ rust_OBJS = $(RUST_ALL_OBJS) rust/rustspec.o
 
 # The compiler itself is called rust1 (formerly grs1)
 rust1$(exeext): $(RUST_ALL_OBJS) attribs.o $(BACKEND) $(LIBDEPS)
-	+$(LLINKER) $(RUST_CXXFLAGS) $(ALL_LINKERFLAGS) $(LDFLAGS) -o $@ \
+	+$(LLINKER) $(ALL_LINKERFLAGS) $(LDFLAGS) -o $@ \
 	      $(RUST_ALL_OBJS) attribs.o $(BACKEND) $(LIBS) $(BACKENDLIBS)
 
 # Build hooks.
@@ -336,7 +336,7 @@ RUST_INCLUDES = -I $(srcdir)/rust \
 CFLAGS-rust/rust-lang.o += $(RUST_INCLUDES)
 CFLAGS-rust/rust-lex.o += $(RUST_INCLUDES)
 CFLAGS-rust/rust-parse.o += $(RUST_INCLUDES)
-CFLAGS-rust/rust-session-manager.o += $(RUST_INCLUDES) $(RUST_CXXFLAGS)
+CFLAGS-rust/rust-session-manager.o += $(RUST_INCLUDES) $(SESSION_FLAGS)
 
 # build all rust/lex files in rust folder, add cross-folder includes
 rust/%.o: rust/lex/%.cc
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -173,7 +173,7 @@ rust_OBJS = $(RUST_ALL_OBJS) rust/rustspec.o
 
 # The compiler itself is called rust1 (formerly grs1)
 rust1$(exeext): $(RUST_ALL_OBJS) attribs.o $(BACKEND) $(LIBDEPS)
-	+$(LLINKER) $(ALL_LINKERFLAGS) $(LDFLAGS) -o $@ \
+	+$(LLINKER) $(RUST_CXXFLAGS) $(ALL_LINKERFLAGS) $(LDFLAGS) -o $@ \
 	      $(RUST_ALL_OBJS) attribs.o $(BACKEND) $(LIBS) $(BACKENDLIBS)
 
 # Build hooks.
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -52,7 +52,7 @@ GCCRS_D_OBJS = \
 RUST_CXXFLAGS = $(CXXFLAGS)
 
 # If the file exists, we're bootstrapping. Set a specific flag for the session manager.
-ifneq ($(wildcard ../stage_current)),)
+ifneq ($(wildcard ../stage_current),)
 SESSION_FLAGS = -DRUST_BOOTSTRAP_BUILD
 endif
 
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -49,6 +49,13 @@ GCCRS_D_OBJS = \
    rust/rustspec.o \
    $(END)
 
+RUST_CXXFLAGS = $(CXXFLAGS)
+
+# If the file exists, we're bootstrapping
+ifneq ($(wildcard ../stage_current)),)
+RUST_CXXFLAGS += -DRUST_RELEASE
+endif
+
 gccrs$(exeext): $(GCCRS_D_OBJS) $(EXTRA_GCC_OBJS) libcommon-target.a $(LIBDEPS)
 	+$(LINKER) $(ALL_LINKERFLAGS) $(LDFLAGS) -o $@ \
 	  $(GCCRS_D_OBJS) $(EXTRA_GCC_OBJS) libcommon-target.a \
@@ -269,7 +276,7 @@ rust.uninstall:
 # Enable selftests for the rust frontend
 selftest-rust: s-selftest-rust
 
-RUST_SELFTEST_FLAGS = -xrust $(SELFTEST_FLAGS)
+RUST_SELFTEST_FLAGS = -xrust -frust-incomplete-and-experimental-compiler-do-not-use $(SELFTEST_FLAGS)
 RUST_SELFTEST_DEPS = rust1$(exeext) $(SELFTEST_DEPS)
 
 # Run the rust selftests
@@ -330,8 +337,6 @@ CFLAGS-rust/rust-lang.o += $(RUST_INCLUDES)
 CFLAGS-rust/rust-lex.o += $(RUST_INCLUDES)
 CFLAGS-rust/rust-parse.o += $(RUST_INCLUDES)
 CFLAGS-rust/rust-session-manager.o += $(RUST_INCLUDES)
-
-RUST_CXXFLAGS = $(CXXFLAGS)
 
 # build all rust/lex files in rust folder, add cross-folder includes
 rust/%.o: rust/lex/%.cc
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -336,7 +336,7 @@ RUST_INCLUDES = -I $(srcdir)/rust \
 CFLAGS-rust/rust-lang.o += $(RUST_INCLUDES)
 CFLAGS-rust/rust-lex.o += $(RUST_INCLUDES)
 CFLAGS-rust/rust-parse.o += $(RUST_INCLUDES)
-CFLAGS-rust/rust-session-manager.o += $(RUST_INCLUDES)
+CFLAGS-rust/rust-session-manager.o += $(RUST_INCLUDES) $(RUST_CXXFLAGS)
 
 # build all rust/lex files in rust folder, add cross-folder includes
 rust/%.o: rust/lex/%.cc
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -49,13 +49,6 @@ GCCRS_D_OBJS = \
    rust/rustspec.o \
    $(END)
 
-RUST_CXXFLAGS = $(CXXFLAGS)
-
-# If the file exists, we're bootstrapping. Set a specific flag for the session manager.
-ifneq ($(wildcard ../stage_current),)
-SESSION_FLAGS = -DRUST_BOOTSTRAP_BUILD
-endif
-
 gccrs$(exeext): $(GCCRS_D_OBJS) $(EXTRA_GCC_OBJS) libcommon-target.a $(LIBDEPS)
 	+$(LINKER) $(ALL_LINKERFLAGS) $(LDFLAGS) -o $@ \
 	  $(GCCRS_D_OBJS) $(EXTRA_GCC_OBJS) libcommon-target.a \
@@ -336,7 +329,9 @@ RUST_INCLUDES = -I $(srcdir)/rust \
 CFLAGS-rust/rust-lang.o += $(RUST_INCLUDES)
 CFLAGS-rust/rust-lex.o += $(RUST_INCLUDES)
 CFLAGS-rust/rust-parse.o += $(RUST_INCLUDES)
-CFLAGS-rust/rust-session-manager.o += $(RUST_INCLUDES) $(SESSION_FLAGS)
+CFLAGS-rust/rust-session-manager.o += $(RUST_INCLUDES)
+
+RUST_CXXFLAGS = $(CXXFLAGS)
 
 # build all rust/lex files in rust folder, add cross-folder includes
 rust/%.o: rust/lex/%.cc
--- a/gcc/rust/lang.opt
+++ b/gcc/rust/lang.opt
@@ -74,6 +74,10 @@ frust-mangling=
 Rust Joined RejectNegative Enum(frust_mangling) Var(flag_rust_mangling)
 -frust-mangling=[legacy|v0]     Choose which version to use for name mangling
 
+frust-incomplete-and-experimental-compiler-do-not-use
+Rust Var(flag_rust_experimental)
+Enable experimental compilation of Rust files at your own risk
+
 Enum
 Name(frust_mangling) Type(int) UnknownError(unknown rust mangling option %qs)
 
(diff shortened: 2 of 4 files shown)


dkm:
I like `do-not-use`, that's really clear. Everything else can sound like "ok it's not fully ready, but must be close" (like `rust nightly` or all these `unstable` things). At least, there's no ambiguity with `do-not-use`.

flip1995:
I also think `do-not-use` is the best option here.

tschwinge:
So I understand the idea of the WIP patch under discussion such that it changes the `--enable-languages` option.  However, isn't that somewhat a moot point, because wouldn't people typically `configure` GCC with `--enable-languages=all`, which would then happily include `rust-work-in-progress` (or similar)?

I had understood the discussions at the GNU Tools Cauldron 2022 such that we'd rename the *driver* (GCC binary that users invoke to compile Rust code) from `gccrs` to `gccrs-work-in-progress` -- but maybe I got that wrong?  In addition to that specific GCC driver, we should also implement some safe-guard in the `gcc` binary itself (which people may also invoke).

For that (or, in fact, instead of renaming `gccrs` driver?), we'd probably disable the automatic handling of `*.rs` files (<https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/lang-specs.h>), such that a user has to explicitly invoke `gcc -x rust-work-in-progress *.rs` to get their Rust source code files recognized as such by GCC?

CohenArthur:
Thanks everyone for the feedback. The commit I just force-pushed does the following:

1. Restore the language name to `rust` for `./configure`. As Thomas pointed out, I think this change does not make too much sense as people won't really care about how their compiler has been configured. If people get a hold of gccrs through their distribution's package manager, this literally achieves nothing.
2. Disable the automatic handling of `.rs` files. I've tested it and can confirm it works, as it tries treating `.rs` files like linker scripts. We can probably do something smarter here if we want.
3. Rename `gccrs` to `gccrs-do-not-use`
4. Rename the language for the `-x` option to `rust-do-not-use`. This way, if people invoke `gcc` instead of `gccrs-do-not-use`, they have to specify `gcc -x rust-do-not-use <file>.rs`

dkm:
Would it be easy to route `.rs` file to a dumb compiler simply displaying an error message with instructions to use `-x rust-do-not-use-or-it-will-explode`  or `gccrs-do-not-use` ?

CohenArthur:
> Would it be easy to route `.rs` file to a dumb compiler simply displaying an error message with instructions to use `-x rust-do-not-use-or-it-will-explode` or `gccrs-do-not-use` ?

I think we can even simply call `echo` or something. I'm not really sure how the `lang-specs.h` file works, but it seems to me that you're just passing it the name of the compiler proper and some arguments

CohenArthur:
@dkm I tried and we can call `echo` or any other binary. I think having a "fake compiler" is a good idea, as we can simply `fatal_error` and display something properly to the user

philberty:
What about .... redirecting to rustc like fortran front-end to llvm did!

flip1995:
> What about .... redirecting to rustc like fortran front-end to llvm did! lol

Unsaved Image 1

tschwinge:
> I think having a "fake compiler" is a good idea, as we can simply `fatal_error` and display something properly to the user

ACK.

And, I just realized how this could be done in the most simple way (unless I'm missing something): actually re-instantiate automatic handling of `*.rs` into `gcc/rust/lang-specs.h`, but via a separate `@rust-fail` spec, which compared to `@rust-do-not-use` also is passing a new command-line option (for example: `-fno-rust-I-really-want-this-experimental-stuff`; `gcc/rust/lang.opt`: `frust-I-really-want-this-experimental-stuff`, with `Undocumented` property), and in the GCC/Rust front end, do `if (!flag_rust_experimental) fatal_error("[...]");`.

tschwinge:
Well, I guess this kind of leads us back to (the inverse of) the other idea discussed occasionally: instead of (a) renaming the `gccrs` driver, etc., (b) require the user to specify `gccrs -frust-I-really-want-this-experimental-stuff *.rs`.  Not sure which variant is "best".

Let's consider the state, eventually, when `gccrs` is deemed OK to use:

For (a):

  - We'll rename the `gccrs-do-not-use` driver back to `gccrs`, and users have to adjust accordingly.
    (I suppose we're not intending to continue (also) providing `gccrs-do-not-use` for backwards compatibility, huh...)
  - We'll either remove the `-x rust-do-not-use` language, and users have to adjust: not specify any `-x [...]` anymore.
    Or, here in fact make `-x rust-do-not-use` an alias to (default) `-x rust`, and users don't have to adjust for that.

For (b):

  - We make the `-frust-I-really-want-this-experimental-stuff` command-line option a no-op, and users don't have to adjust.

(..., and in ten years' time, you'll laugh if you still find that option in your build scripts...) ;-)

tschwinge:
This is a bit of a "color of the bikeshed" discussion, but I think it's necessary to spend some energy on this: this is absolutely directly user-facing (so, first impression of GCC/Rust usage), but we should also make the transition path easy, once GCC/Rust is deemed OK to use.

flip1995:
I don't think that people will complain about their build system breaking because a `do-not-use-this`-flag was removed. If they use this flag or renamed binary, they acknowledge that they know what they're doing, which includes that their build script will break in the future. Thinking about backwards compat for highly experimental features is a step too far IMHO. That's like if rust would have to support every iteration of every nightly feature ever.

CohenArthur:
> And, I just realized how this could be done in the most simple way (unless I'm missing something): actually re-instantiate automatic handling of *.rs into gcc/rust/lang-specs.h, but via a separate @rust-fail spec, which compared to @rust-do-not-use also is passing a new command-line option (for example: -fno-rust-I-really-want-this-experimental-stuff

Isn't that creating the same problem? Unless I'm misunderstanding something, what you're proposing is that `gcc test.rs` "expands" to `rust1 -fi-really-want-this-experimental-stuff test.rs`, correct? So the flag being very long/annoying to type is not something the user has to deal with anymore.

> Well, I guess this kind of leads us back to (the inverse of) the other idea discussed occasionally: instead of (a) renaming the gccrs driver, etc., (b) require the user to specify gccrs -frust-I-really-want-this-experimental-stuff *.rs. Not sure which variant is "best".

I think they're not exclusive! I assume that at least some people will use `gccrs` through `cargo-gccrs`, in which case having the binary names `gccrs` or `gccrs-do-not-use` does not matter a lot. However, having to specify `GCCRS_EXTRA_FLAGS=-frust-i-really-want-to-use-this-experimental-broken-compiler cargo gccrs build` will.

Similarly, having a big non-avoidable warning at the beginning of the compilation pipeline would also do the trick.

Regarding future proofing it or breaking it: I don't really have an opinion on this. I the people that will be using `gccrs` in an experimental state recognize that they will have to put in some elbow grease at some point, and yes that might include broken builds once GCC *17* releases ;) We can also do something where `gccrs-do-not-use` and `-x rust-do-not-use` are provided for a few more releases and being deprecated.

CohenArthur:
I'd also like to point out that as a user of `gccrs`, this PR is completely going to break the https://github.com/Rust-GCC/testing project, and I'm okay with fixing it. So it seems that from recent polls, at least 75% of the userbase of `gccrs` is okay with a little breakage :P

tschwinge:
> > And, I just realized how this could be done in the most simple way (unless I'm missing something): actually re-instantiate automatic handling of *.rs into gcc/rust/lang-specs.h, but via a separate @rust-fail spec, which compared to @rust-do-not-use also is passing a new command-line option (for example: -fno-rust-I-really-want-this-experimental-stuff
> 
> Isn't that creating the same problem? Unless I'm misunderstanding something, what you're proposing is that `gcc test.rs` "expands" to `rust1 -fi-really-want-this-experimental-stuff test.rs`, correct? So the flag being very long/annoying to type is not something the user has to deal with anymore.

No, I'm proposing that `gcc test.rs` "expands" to `rust1 -fno-rust-i-really-want-this-experimental-stuff test.rs` (notice `-fno-[...]`), and this would be caught by `if (!flag_rust_experimental) fatal_error("[...]");`, and so the user has to explicitly invoke `gcc -frust-i-really-want-this-experimental-stuff test.rs` which "expands" to `rust1 -fno-rust-i-really-want-this-experimental-stuff -frust-i-really-want-this-experimental-stuff test.rs` (where `-frust-[...]` then overrides the default `-fno-rust-[...]`).

Well, and actually this doesn't need any separate `@rust-fail` spec, or does it?  We'd just set `-fno-rust-i-really-want-this-experimental-stuff` as default?

> > Well, I guess this kind of leads us back to (the inverse of) the other idea discussed occasionally: instead of (a) renaming the gccrs driver, etc., (b) require the user to specify gccrs -frust-I-really-want-this-experimental-stuff *.rs. Not sure which variant is "best".
> 
> I think they're not exclusive! I assume that at least some people will use `gccrs` through `cargo-gccrs`, in which case having the binary names `gccrs` or `gccrs-do-not-use` does not matter a lot.

Right, that's a good point.  `cargo-gccrs` either has to implicitly call `gccrs-do-not-use` -- which is a pointless exercise, as we're trying to make that explicit.  Or, `cargo-gccrs` itself needs some `-i-really-want-this-experimental-stuff` flag that users have to specify explicitly, but then we'd rather settle on an existing flag (that we need anyway for `gccrs`/`gcc` usage):

> However, having to specify `GCCRS_EXTRA_FLAGS=-frust-i-really-want-to-use-this-experimental-broken-compiler cargo gccrs build` will.

... like this, for example.

So my point is: doesn't requiring the `-frust-i-really-want-this-experimental-stuff` command-line option solve all issues at once?

---

> Similarly, having a big non-avoidable warning at the beginning of the compilation pipeline would also do the trick.

I predict that users will just filter that out or otherwise ignore that, like "everyone" does with other compiler diagnostics.  ;-)

ojeda:
> the people that will be using `gccrs` in an experimental state recognize that they will have to put in some elbow grease at some point, and yes that might include broken builds once GCC _17_ releases ;)

To give you some "user input", from the Rust for Linux side (i.e. as a potential project that may use `gccrs` early on): this is completely fine and we will understand if it breaks!

In fact, if we ended up merging any very early support upstream for `gccrs` at some point, I would expect us to pin the `gccrs` version like we did for `rustc` (even if the `rustc` one is not pinned at all by the time we start using `gccrs`), and then do the required changes as needed when upgrading it.

ojeda:
> the people that will be using `gccrs` in an experimental state recognize that they will have to put in some elbow grease at some point, and yes that might include broken builds once GCC _17_ releases ;) We can also do something where `gccrs-do-not-use` and `-x rust-do-not-use` are provided for a few more releases and being deprecated.

In case you may want some "user input", from the Rust for Linux (i.e. as a project that may want to use `gccrs` early): this is completely fine and we would understand if everything breaks! :)

In fact, if we ended up merging some very early `gccrs` support upstream, I would expect that we need to pin the `gccrs` version like we did for `rustc` (even if `rustc` version is not pinned at all by the time we add `gccrs` support).

CohenArthur:
Silly github... doesn't understand conditionals :)

tschwinge:
Re-iterating my main point why instead of renaming the `gccrs` driver to `gccrs-do-not-use` or similar, a GCC flag like `-frust-i-really-want-to-use-this-experimental-broken-compiler` is best: the latter works with Rust code compiled through the `gccrs` font end as well as `gcc` front end, and through `cargo-gccrs` (via `GCCRS_EXTRA_FLAGS` or similar).

And, once mature, we can then simply turn `-frust-i-really-want-to-use-this-experimental-broken-compiler` into a no-op.

---

> `cargo-gccrs` either has to implicitly call `gccrs-do-not-use` -- which is a pointless exercise, as we're trying to make that explicit. Or, `cargo-gccrs` itself needs some `-i-really-want-this-experimental-stuff` flag that users have to specify explicitly, but then we'd rather settle on an existing flag (that we need anyway for `gccrs`/`gcc` usage):
> 
> > However, having to specify `GCCRS_EXTRA_FLAGS=-frust-i-really-want-to-use-this-experimental-broken-compiler cargo gccrs build` will.
> 
> ... like this, for example.

CohenArthur:
Hi all :) I've finally found the time to do some work on this issue. I've added a `-frust-incomplete-and-experimental-compiler-do-not-use` flag (credits to @flip1995 for the wording), which enables boostrapped gccrs to compile a file. Otherwise, a lengthy fatal error is emitted.

On non-bootstrapped build (so for our development for example), the check isn't performed and the flag becomes a no-op

```shell
arthur@platypus ~/G/r/gccrs (frust-very-long)> build/gcc/rust1 test.rs
arthur@platypus ~/G/r/gccrs (frust-very-long)> build/gcc/rust1 test.rs -frust-incomplete-and-experimental-compiler-do-not-use`
```

On the compiler farm machine where I bootstrapped gccrs:

(The formatting does not wrap, which isn't comfortable. If you simply want to read the error message skip ahead)

```shell
cohenarthur@gcc14:~/gccrs$ build/gcc/rust1 test.rs
rust1: fatal error: gccrs is not yet able to compile Rust code properly. Most of the errors produced will be gccrs' fault and not the crate you are trying to compile. Because of this, please reports issues to us directly instead of opening issues on said crate's repository.

Our github repository: https://github.com/rust-gcc/gccrs
Our bugzilla tracker: https://gcc.gnu.org/bugzilla/buglist.cgi?bug_status=__open__&component=rust&product=gcc

If you understand this, and understand that the binaries produced might not behave accordingly, you may attempt to use gccrs in an experimental manner by passing the following flag:

`-frust-incomplete-and-experimental-compiler-do-not-use`

For cargo-gccrs, this means passing

GCCRS_EXTRA_FLAGS="-frust-incomplete-and-experimental-compiler-do-not-use"

as an environment variable.
compilation terminated.
cohenarthur@gcc14:~/gccrs$ build/gcc/rust1 test.rs -frust-incomplete-and-experimental-compiler-do-not-use
```

Here's the text of the error with wrapping:

> rust1: fatal error: gccrs is not yet able to compile Rust code properly. Most of the errors produced will be gccrs' fault and not the crate you are trying to compile. Because of this, please reports issues to us directly instead of opening issues on said crate's repository.
> 
> Our github repository: https://github.com/rust-gcc/gccrs
> Our bugzilla tracker: https://gcc.gnu.org/bugzilla/buglist.cgi?bug_status=__open__&component=rust&product=gcc
> 
> If you understand this, and understand that the binaries produced might not behave accordingly, you may attempt to use gccrs in an experimental manner by passing the following flag:
> 
> `-frust-incomplete-and-experimental-compiler-do-not-use`
> 
> For cargo-gccrs, this means passing
> 
> GCCRS_EXTRA_FLAGS="-frust-incomplete-and-experimental-compiler-do-not-use"
> 
> as an environment variable.
> compilation terminated.

How does everyone feel about this? Do you feel that it is a sufficient deterrent/warning to unaware users?

I'll update this branch to reflect those changes and keep trace of the discussion instead of opening up another PR.

tschwinge:
Yes, that generally looks good to me, just not the `RUST_BOOTSTRAP_BUILD` special-casing: we generally have to enable this behavior, not just for GCC bootstrap builds.

CohenArthur:
> Yes, that generally looks good to me, just not the `RUST_BOOTSTRAP_BUILD` special-casing: we generally have to enable this behavior, not just for GCC bootstrap builds.

Do you mean that even developers working on the project should have to pass the flag when testing/etc?

tschwinge:
> > Yes, that generally looks good to me, just not the `RUST_BOOTSTRAP_BUILD` special-casing: we generally have to enable this behavior, not just for GCC bootstrap builds.
> 
> Do you mean that even developers working on the project should have to pass the flag when testing/etc?

Quite right.  ;-) Same rules for all.  Whether GCC is built '--enable-bootstrap' vs. '--disable-bootstrap' certainly doesn't tell you whether somebody is a GCC/Rust compiler developer.

What might be useful to have is that setting an environment variable `GCC_RUST_INCOMPLETE_AND_EXPERIMENTAL_COMPILER_DO_NOT_USE=1` achieves the same behavior  as `-frust-incomplete-and-experimental-compiler-do-not-use`?

CohenArthur:
> Quite right.  ;-) Same rules for all.  Whether GCC is built '--enable-bootstrap' vs. '--disable-bootstrap' certainly doesn't tell you whether somebody is a GCC/Rust compiler developer.

Argh, this means making changes to our dejagnu expressions or testsuite :(

> What might be useful to have is that setting an environment variable `GCC_RUST_INCOMPLETE_AND_EXPERIMENTAL_COMPILER_DO_NOT_USE=1` achieves the same behavior  as `-frust-incomplete-and-experimental-compiler-do-not-use`?

I like that idea. It might also help with my above comment

tschwinge:
> Argh, this means making changes to our dejagnu expressions or testsuite :(

I don't understand, do you mean in all test cases individually?  In the few GCC/Rust testsuites just generally inject `-frust-incomplete-and-experimental-compiler-do-not-use`?  Or, do I fail to see a problem there?

> > What might be useful to have is that setting an environment variable `GCC_RUST_INCOMPLETE_AND_EXPERIMENTAL_COMPILER_DO_NOT_USE=1` achieves the same behavior  as `-frust-incomplete-and-experimental-compiler-do-not-use`?
> 
> I like that idea. It might also help with my above comment

That won't help, because DejaGnu doesn't have a proper implementation of transmitting environment variables to the (potentially remote) system where the compiler is run.

That suggestion was for manual use during development.

tschwinge:
:-) Should have added: please let me know if you'd like help with these final pieces here.

CohenArthur:
@tschwinge done :)

tschwinge (approved):

philberty:
The only reservation I have now is that we need to sync up with @dkm as this will break godbolt (compiler explorer) testing and hopefully we can default this for our testsuite

tschwinge:
Right.  GCC/Rust test suite has already been addressed, and Compiler Explorer should be changed to invoke by default `gccrs` with `-frust-incomplete-and-experimental-compiler-do-not-use` indeed.

dkm:
I'll probably be able to do the switch tomorrow night  :)

dkm (approved):

dkm:
Let's see if I can have everything merged tonight :)
bors r+

dkm:
wtf-abort

tschwinge:
My last push (fixes an actual stylistic issue and) is meant to help figure out what's wrong with `.github/bors_log_expected_warnings` vs. actual `log_warnings`; <https://github.com/Rust-GCC/gccrs/pull/1540#issuecomment-1339983725>.

CohenArthur:
Aaaaaah my worries were unfounded. Great job bors thank you

dkm:
FYI everything looks fine now, compiler-explorer has been changed and a fresh gccrs installed

PR #1543, opened 2022-09-24
Reintroduce targetrustm hooks without the maze of macro definitions in target headers

ibuclaw:
- Adds skeleton support and documentation for targetrustm hooks.
- Re-adds previously reverted `TARGET_RUST_CPU_INFO` target hook.
- Re-adds the previously reverted `TARGET_RUST_OS_INFO` target hook.
- Implements `TARGET_RUST_CPU_INFO` hook for x86 (only a subset of the original i386-rust.cc was re-added)
- Implements `TARGET_RUST_OS_INFO` hook for darwin (covers both x86 and powerpc).

There's quite a bit to unpick from what was removed, but I hope this shows how firstly `rust-target.h` and `tm_rust.h` should be kept as empty as possible for the `default-rust.cc` source file that pulls in these headers; secondly all rust macros and declarations be put in rust-specific headers, then pulled into tm_rust.h as-needed. From every other `${target}-rust.cc` file, you can include as much as you like in order to pry out the information you need to get at.

I've verified this does work as expected with the `-frust-dump-target_options` switch, e.g with `-mavx2`
```
target_endian: "little"
target_pointer_width: "64"
target_feature: "fxsr"
target_feature: "popcnt"
target_feature: "avx2"
target_feature: "ssse3"
target_feature: "sse4.1"
target_feature: "sse3"
target_feature: "avx"
target_feature: "sse"
target_feature: "xsave"
target_feature: "sse4.2"
target_feature: "sse2"
target_feature: "mmx"
target_arch: "x86_64"
```

I've also built a few darwin cross compilers with the following results on i686-darwin
```
target_pointer_width: "32"
target_env: ""
target_vendor: "apple"
target_endian: "little"
target_os: "macos"
target_family: "unix"
target_feature: "sse3"
target_feature: "sse"
target_feature: "sse2"
target_feature: "mmx"
target_arch: "x86"
```
and on powerpc64-darwin
```
target_pointer_width: "64"
target_env: ""
target_vendor: "apple"
target_endian: "big"
target_os: "macos"
target_family: "unix"
```


CohenArthur (approved):
LGTM! Thanks a lot

ibuclaw:
Added darwin-specific target hooks (it's picked up by both x86 and PPC).

Mostly now it's just a matter of adding other OS and CPU support code in the exact same way.

CohenArthur:
@ibuclaw do you want to merge this in this state or keep working on it? I see the PR is still marked as draft, but it's in a good spot and I don't mind merging it now

ibuclaw:
@CohenArthur depends on how small you want to keep this.  :-)

 I can keep incrementally adding targets here until a suitable baseline set of implementations are present, or carry on adding one target, one PR at a time.

Of what's being tested by the GH runners, the only rust target info that's missing would be Linux support.

ibuclaw:
Removed unnecessary include from i386-rust.cc, I guess this is fine for now, and will just add support for other kinds of targets in future PRs, based on what was previously reverted from the original target support code.

ibuclaw:
Split out the implementation of the target hook from the actual adding of it, so anyone else who's interested can just look at either commit and see clearly what they have to do in order to in add, say, target cpu info for `riscv-*-*` or os info for `*-*-netbsd`.

philberty (approved):
LGTM!!

Nice work this is really great work! Lets merge this Mark's build farm has a bunch of more esoteric targets that would be really nice to keep green you can see their status in our README near the top.

I think let's merge this. As part of merging gccrs upstream, I think we will likely make this a separate set of patches to make the current review process simpler and then once that's done we can get this reviewed.

PR #1548, opened 2022-09-27
Dump module items

CohenArthur:


gcc/rust/ast/rust-ast-dump.cc
```diff
@@ -554,7 +579,23 @@ Dump::visit (Method &method)
 
 void
 Dump::visit (Module &module)
-{}
+{
+  indentation.increment ();
```
jdupak, line 583: I feel this should be after the curly brace (line 588), to indent the body only.

jdupak: Also a space after mod is missing.

CohenArthur: Fixed in [6bb6550](https://github.com/Rust-GCC/gccrs/pull/1548/commits/6bb6550c44d6644bb22e4a66be467802faeb23d8)

jdupak:
I have just noticed that this implementation is incomplete. It ignores the module kind case - external - which should not have curly braces. I will fix that in a followup PR.

jdupak:
(https://doc.rust-lang.org/reference/items/modules.html#the-path-attribute)

CohenArthur:
> I have just noticed that this implementation is incomplete. It ignores the module kind case - external - which should not have curly braces. I will fix that in a followup PR.

Perfect, thanks a lot. I had completely forgotten!

PR #1549, opened 2022-09-27
Dump macro declarations properly

CohenArthur:
Dump `MacroRulesDef` properly

### DIFF
--- a/gcc/rust/ast/rust-ast-dump.cc
+++ b/gcc/rust/ast/rust-ast-dump.cc
@@ -175,11 +175,27 @@ Dump::format_struct_field (StructField &field)
 
 void
 Dump::visit (Token &tok)
-{}
+{
+  stream << tok.as_string ();
+}
 
 void
 Dump::visit (DelimTokenTree &delim_tok_tree)
-{}
+{
+  auto tokens = delim_tok_tree.to_token_stream ();
+
+  indentation.increment ();
+  stream << '\n' << indentation;
+
+  for (const auto &tok : tokens)
+    {
+      stream << ' ';
+      tok->accept_vis (*this);
+    }
+
+  indentation.decrement ();
+  stream << '\n' << indentation;
+}
 
 void
 Dump::visit (AttrInputMetaItemContainer &input)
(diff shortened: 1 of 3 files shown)


gcc/rust/ast/rust-ast-dump.cc
```diff
@@ -1126,22 +1130,138 @@ Dump::visit (ExternBlock &block)
...
-{}
+{
+  auto delimiters = get_delimiters (matcher.get_delim_type ());
+
+  stream << delimiters.first;
+
+  for (auto &match : matcher.get_matches ())
+    {
+      match->accept_vis (*this);
+      stream << ' ';
```
jdupak, line 1211: This produces extra trailing space.

```rust
macro_rules! foo {
        (l:tt ) =>
//           ^
                 { bar ! ( $ l ) ; }
        ;
}
```

gcc/rust/ast/rust-ast-dump.cc
```diff
@@ -1126,22 +1130,138 @@ Dump::visit (ExternBlock &block)
...
+  emit_token_tree (transcriber.get_token_tree ());
+}
+
+void
+Dump::emit_token_tree (const DelimTokenTree &tree)
+{
+  auto tokens = tree.to_token_stream ();
+
+  indentation.increment ();
+  stream << '\n' << indentation;
```
jdupak, line 1229: This \n should not be here. See
```rust
macro_rules! foo {
        (l:tt ) => // HERE SHOULD NOT BE A NEW NEW LINE
                 { bar ! ( $ l ) ; }
        ;
}
```

gcc/rust/ast/rust-ast-dump.cc
```diff
@@ -1126,22 +1130,138 @@ Dump::visit (ExternBlock &block)
...
+
+  for (const auto &tok : tokens)
+    stream << ' ' << tok->as_string ();
+
+  indentation.decrement ();
+  stream << '\n' << indentation;
```
jdupak, line 1235: Same as above.

jdupak:
I am yet unsure where, but the dollar sign in the matcher gets lost:
```rust
macro_rules! foo {
    ($l:tt) => { bar!($l); }
}
```
```rust
macro_rules! foo {
        (l:tt ) =>  { bar ! ( $ l ) ; };
}

```

gcc/rust/ast/rust-ast-dump.h
```diff
@@ -72,38 +72,38 @@ class Dump : public ASTVisitor
...
+  /**
+   * Format together common items of functions: Parameters, return type, block
+   */
   void format_function_common (std::unique_ptr<Type> &return_type,
 			       std::unique_ptr<BlockExpr> &block);
 
-  /**
-   * Format a function's definition parameter
-   */
+  /* Format a function's definition parameter */
```
jdupak, line 81: For doxygen there should be two stars even for single line comment.

jdupak:
This is a tiny detail, but it would be much easier to read it if it could be read from top to bottom. Now this can be fixed without losing git history (since it is just empty braces).

jdupak (changes requested):

gcc/rust/ast/rust-ast-dump.cc
CohenArthur: This is okay since macro parsing relies on parsing tokens. I agree it does not look good however

gcc/rust/ast/rust-ast-dump.cc
CohenArthur: The issue with this is that the delimiters are part of the token stream. We could change it and avoid printing the delimiters (something like `for (size_t i = 1; i < tokens.length() - 1; i++)` etc) but that could be fixed in a later PR imo

dafaust (approved):
LGTM

philberty (approved):
Nice one i didnt realise you had done this already!!! :D

jdupak (approved):
LGTM (note that I have not run it this time.)

PR #1554, opened 2022-10-03
testing: try loop in const function

abbasfaisal:


CohenArthur (approved):
LGTM :) One thing I would suggest is not creating PRs from your `master` branch. Here's a little article with commands to avoid it: https://blog.jasonmeridth.com/posts/do-not-issue-pull-requests-from-your-master-branch/

There are more reasons to not do so than mentionned in this blog, mind you. And this is a tiny nitpick but the PR's content is great :D

PR #1556, opened 2022-10-04
make: Inherit CXXFLAGS, guard against non g++ compilers for warnings

### LINKED ISSUE #1538: Inherit `$(CXXFLAGS)` in `Make-lang.in`
https://gcc.gnu.org/pipermail/gcc-rust/2022-September/000490.html

CohenArthur:
Fixes #1538 

I don't know if there is a better way to do this or if this is what was expected. I also didn't have the heart to look for the proper warning flag equivalents on the MSVC++ side or for other compilers. I think it's safe to assume that if people are boostrapping `gccrs` with something else than `g++`, they're not hacking on the compiler and thus don't need these warnings?


CohenArthur:
We also need to make sure that this file ends up in the master branch, not just the patch branch

PR #1560, opened 2022-10-05
Fix -Wunused using `rust-lang.cc` hack

CohenArthur:
@tschwinge if your patch lands first I'll happily revert that and use `LangInit`


philberty (approved):

philberty:
I think we should just click merge on this and maybe work at merging this onto our master branch too

PR #1562, opened 2022-10-05
Support looking up super trait items during path resolution

### LINKED ISSUE #1555: Super trait method resolution fails
I tried this code:

```rust
extern "C" {
    fn printf(s: *const i8, ...);
}

struct Foo(i32);
trait Bar {
    fn baz(&self);
}

trait Baz: Bar {
    fn qux(&self);
}

impl Bar for Foo {
    fn baz(&self) {
        unsafe {
            let a = "baz %i\n\0";
            let b = a as *const str;
            let c = b as *const i8;

            printf(c, self.0);
        }
    }
}

impl Baz for Foo {
    fn qux(&self) {
        unsafe {
            let a = "qux %i\n\0";
            let b = a as *const str;
            let c = b as *const i8;

            printf(c, self.0);
        }
    }
}

fn static_dispatch<T: Baz>(t: &T) {
    t.baz();
    t.qux();
}

pub fn main() {
    let a;
    a = &Foo(123);

    static_dispatch(a);
}
```

I expected to see this happen: compile correctly

Instead, this happened:
```
<source>:39:7: error: failed to resolve method for 'baz'
   39 |     t.baz();
      |       ^
```

### Meta

- What version of Rust GCC were you using, git sha if possible. 9b1ba11b0b2f873b85dfc7643fe778e974e874b8

philberty:
When we lookup names in paths such as Foo::bar, foo is a type we resolve
and then we lookup 'bar' based on what type Foo is which includes probing
relevant bounds of this type. We currently return a vector of possible
candidates and this patch changes it so that we return a vector of unique
items based on DefId. Ideally we should use move to a std::set for this.
    
The final patch actually updates the trait reference lookup functions to revert
to looking up relevant super traits when required.

Fixes #1555

### DIFF
--- a/gcc/rust/typecheck/rust-hir-trait-ref.h
+++ b/gcc/rust/typecheck/rust-hir-trait-ref.h
@@ -336,6 +336,15 @@ public:
 	    return true;
 	  }
       }
+
+    // lookup super traits
+    for (const auto &super_trait : super_traits)
+      {
+	bool found = super_trait->lookup_trait_item (ident, ref);
+	if (found)
+	  return true;
+      }
+
     return false;
   }
 
@@ -351,6 +360,16 @@ public:
 	if (ident.compare (item.get_identifier ()) == 0)
 	  return &item;
       }
+
+    // lookup super traits
+    for (const auto &super_trait : super_traits)
+      {
+	const TraitItemReference *res
+	  = super_trait->lookup_trait_item (ident, type);
+	if (!res->is_error ())
+	  return res;
+      }
+
     return &TraitItemReference::error_node ();
   }
 
--- a/gcc/testsuite/rust/compile/torture/issue-1555.rs
+++ b/gcc/testsuite/rust/compile/torture/issue-1555.rs
@@ -1,0 +1,48 @@
+extern "C" {
+    fn printf(s: *const i8, ...);
+}
+
+struct Foo(i32);
+trait Bar {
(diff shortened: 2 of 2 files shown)


gcc/rust/typecheck/rust-hir-path-probe.h
```diff
@@ -426,6 +456,20 @@ class PathProbeType : public TypeCheckBase, public HIR::HIRImplVisitor
...
 
+  // we should probably move to a set but not sure how to enforce the
+  // unique-ness based on the DefId yet on std::set
+  void append_candidate (PathProbeCandidate &&candidate)
+  {
+    for (const auto &c : candidates)
+      {
+	bool found = c.get_defid () == candidate.get_defid ();
+	if (found)
+	  return;
```
CohenArthur, line 467: I think this already assumes that `DefId`s are unique, so we should be able to use a set (unless I'm misunderstanding the above comment?)

CohenArthur (approved):
LGTM

gcc/rust/typecheck/rust-hir-path-probe.h
philberty: Yeah, i just need to figure out how to do that with std::set to say take this type but use this as the key

philberty: I've now converted to std::set just checking the CI passes ok

PR #1564, opened 2022-10-06
builtins: Move implementation into source file

CohenArthur:


philberty (approved):
LGTM i love it when we do this!

PR #1565, opened 2022-10-06
Implement data prefetch intrinsics

CohenArthur:
Addresses #658 
Needs #1564 so only review the last commit

This PR implements the two intrinsics related to data prefetching. I have to say, since these are hints for the backend, I am unsure if the current implementation is right. I believe it should be.


philberty (approved):
LGTM

PR #1570, opened 2022-10-06
Apply #1560 master

CohenArthur:
Same as #1560, but in a single more descriptive commit. I'll hold on doing the other fixes until you're okay with the process @philberty


philberty (approved):
So long as the git sha's match the ones we merged into the other branch then everything is good. LGTM

Issue #1576, opened 2022-10-10
Name resolve macros properly


CohenArthur:
Some name resolution is required during macro expansion, in order to resolve macros properly.

- [ ] Resolve imports
- [ ] Resolve macros

`rustc_resolve::macros` as opposed to `rustc_resolve::late`

CohenArthur:
This will probably tie in with #1078

PR #1580, opened 2022-10-11
Add early name resolver

CohenArthur:
This addresses our issue with name resolution (#1576) not acting on macros early enough but does not fix it yet.

This name resolver performs the same macro name resolution as what was previously done by the AttrVisitor visitor and macro expander.

It also resolves macro expressions in builtin-macros properly, as well as expanded AST nodes when necessary.

### DIFF
--- a/gcc/rust/expand/rust-macro-expand.h
+++ b/gcc/rust/expand/rust-macro-expand.h
@@ -25,6 +25,7 @@
 #include "rust-ast.h"
 #include "rust-macro.h"
 #include "rust-hir-map.h"
+#include "rust-early-name-resolver.h"
 #include "rust-name-resolver.h"
 #include "rust-macro-invoc-lexer.h"
 
@@ -323,10 +324,13 @@ struct MacroExpander
     AST::ASTFragment old_fragment = std::move (expanded_fragment);
     auto accumulator = std::vector<AST::SingleASTNode> ();
     expanded_fragment = AST::ASTFragment::create_error ();
+    auto early_name_resolver = Resolver::EarlyNameResolver ();
 
     for (auto &node : old_fragment.get_nodes ())
       {
 	expansion_depth++;
+
+	node.accept_vis (early_name_resolver);
 	node.accept_vis (vis);
 	// we'll decide the next move according to the outcome of the macro
 	// expansion
--- a/gcc/rust/resolve/rust-early-name-resolver.cc
+++ b/gcc/rust/resolve/rust-early-name-resolver.cc
@@ -1,0 +1,1087 @@
+// Copyright (C) 2020-2022 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#include "rust-early-name-resolver.h"
+#include "rust-ast-full.h"
+#include "rust-name-resolver.h"
+
+namespace Rust {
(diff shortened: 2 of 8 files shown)


gcc/rust/expand/rust-macro-expand.h
```diff
@@ -323,10 +324,13 @@ struct MacroExpander
     AST::ASTFragment old_fragment = std::move (expanded_fragment);
     auto accumulator = std::vector<AST::SingleASTNode> ();
     expanded_fragment = AST::ASTFragment::create_error ();
+    auto early_name_resolver = Resolver::EarlyNameResolver ();
 
     for (auto &node : old_fragment.get_nodes ())
       {
 	expansion_depth++;
+
+	node.accept_vis (early_name_resolver);
```
philberty, line 333: thats pretty neat

gcc/rust/resolve/rust-ast-resolve-item.cc
```diff
@@ -968,6 +968,14 @@ ResolveItem::visit (AST::UseDeclaration &use_item)
...
   auto to_resolve = flatten_use_dec_to_paths (use_item);
 
+  // FIXME: I think this does not actually resolve glob use-decls and is going
+  // the wrong way about it. RFC #1560 specifies the following:
+  //
+  // > When we find a glob import, we have to record a 'back link', so that when
+  // a public name is added for the supplying module, we can add it for the
+  // importing module.
+  //
+  // Which is the opposite of what we're doing if I understand correctly?
```
philberty, line 978: Good comment, i noticed this as well. I think there are two options, try and flatten it like the code path we have which would be ideal.

We can deal with this down the line though and just raise an issue for it.

gcc/rust/resolve/rust-early-name-resolver.cc
```diff
@@ -0,0 +1,1087 @@
...
+EarlyNameResolver::EarlyNameResolver ()
+  : resolver (*Resolver::get ()), mappings (*Analysis::Mappings::get ())
+{}
+
+void
+EarlyNameResolver::go (AST::Crate &crate)
+{
+  // FIXME: Is that valid? Why is the regular name resolution doing
+  // mappings->get_next_node_id()?
+  resolver.get_macro_scope ().push (crate.get_node_id ());
```
philberty, line 35: We do this in the constructor of the name resolver because it's a singleton. The idea is we have a root of artificial. root where it contains the builtins such as i32, i8 etc... Then when we reuse this name resolver we push crate specific rib's for each crate we name resolve.

You are doing the correct thing here. But what you could do in the constructor here is have a root rib where you insert the builtin macro's so that you handle builtin's in the same way but I don't think this works for macros since all builtin macros are like the intrinsic where they are declared as stubs.

gcc/rust/resolve/rust-early-name-resolver.cc
```diff
@@ -0,0 +1,1087 @@
...
+
+  for (auto &item : crate.items)
+    item->accept_vis (*this);
+
+  // FIXME: Should we pop the macro scope?
```
philberty, line 40: If you can pop here, I would. The state is held within the resolver singleton, and this might affect the macro resolution of other crates.

gcc/rust/rust-session-manager.cc
```diff
@@ -783,6 +784,9 @@ Session::injection (AST::Crate &crate)
 void
 Session::expansion (AST::Crate &crate)
 {
+  /* We need to name resolve macros and imports here */
+  Resolver::EarlyNameResolver ().go (crate);
```
philberty, line 788: We might need to add something like this to how we load_extern_crate's but lets not worry about that yet.

We don't do any expansion within load_extern_crate because when we are exporting metadata we should have stripped all nodes as expected and macros all expanded. We do need to export macros so we need some kind of method to allow for referencing these macros at some point but yeah lets not worry about that yet.

philberty (approved):
LGTM we definetly needed this and glad to see it as a separate pass!

CohenArthur:
I think there might actually be an issue with how I create the `EarlyNameResolver` willy-nilly since it automatically tries resolving with the crate module as a base scope. I think this will require fixing and I'll try coming up with a test case exposing that behavior

gcc/rust/resolve/rust-early-name-resolver.cc
CohenArthur: Right now we cannot pop as the EarlyNameResolver is being reused in other parts of the codebase after having finished running through the crate's items. My next PR will fix this issue

PR #1583, opened 2022-10-12
Support type resolution on super traits on dyn objects

philberty:
When checking if specified bounds satisfy other bounds we must lookup the super traits. To finish the support for super traits we need to redo the computation of method addresses to support super traits.

Addresses #914

### DIFF
--- a/gcc/rust/backend/rust-compile.cc
+++ b/gcc/rust/backend/rust-compile.cc
@@ -247,6 +247,8 @@ HIRCompileBase::compute_address_for_trait_item (
   // Algo:
   // check if there is an impl-item for this trait-item-ref first
   // else assert that the trait-item-ref has an implementation
+  //
+  // FIXME this does not support super traits
 
   TyTy::TypeBoundPredicateItem predicate_item
     = predicate->lookup_associated_item (ref->get_identifier ());
--- a/gcc/rust/typecheck/rust-hir-trait-ref.h
+++ b/gcc/rust/typecheck/rust-hir-trait-ref.h
@@ -380,6 +380,16 @@ public:
     return item_refs;
   }
 
+  void get_trait_items_and_supers (
+    std::vector<const TraitItemReference *> &result) const
+  {
+    for (const auto &item : item_refs)
+      result.push_back (&item);
+
+    for (const auto &super_trait : super_traits)
+      super_trait->get_trait_items_and_supers (result);
+  }
+
   void on_resolved ()
   {
     for (auto &item : item_refs)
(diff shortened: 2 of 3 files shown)


philberty:
It seems like a good idea to split this patch out, the first part of the fix for support of dyn objects containing super traits is to ensure we correctly check that bounds are satisfied recursively. So that means if we have a dyn B and B has a super trait of A. The object we coerce to a dyn object must satisfy both A and B and if there are more levels recursively keep going.

The final part of the fix here is to compute the vtable including the super trait items this seems better fixed as a separate PR.

gcc/rust/backend/rust-compile.cc
```diff
@@ -247,6 +247,8 @@ HIRCompileBase::compute_address_for_trait_item (
   // Algo:
   // check if there is an impl-item for this trait-item-ref first
   // else assert that the trait-item-ref has an implementation
+  //
+  // FIXME this does not support super traits
```
CohenArthur, line 251: Should this be turned into an issue?

gcc/rust/typecheck/rust-hir-trait-ref.h
```diff
@@ -380,6 +380,16 @@ class TraitReference
...
 
+  void get_trait_items_and_supers (
+    std::vector<const TraitItemReference *> &result) const
+  {
+    for (const auto &item : item_refs)
+      result.push_back (&item);
+
+    for (const auto &super_trait : super_traits)
+      super_trait->get_trait_items_and_supers (result);
+  }
```
CohenArthur, line 391: I believe this is already a problem on the `master` branch and not necessarily related to this PR, but I'm a bit afraid of the recursive function without a stop condition. The following sort of test cases crash:
```rust
pub trait A: B {}

pub trait B: A {}
```

CohenArthur (approved):
I think this looks good :) The recursion issue is a user-experience issue, so that can be addressed later on

gcc/rust/backend/rust-compile.cc
philberty: Its already an issue #914 so this PR fixes the type-resolution part but doesn't finish the code-generation part so this PR does not close that issue.

gcc/rust/typecheck/rust-hir-trait-ref.h
philberty: oh good point will fix

philberty: Let's raise this as a separate issue I have known about this for a while it needs to be caught during resolving a trait.

CohenArthur: #1591

PR #1584, opened 2022-10-12
Add missing lang item mappings

philberty:
Adds the missing fn_once lang item and the missing rust-call abi option.


CohenArthur (approved):
LGTM

PR #1587, opened 2022-10-13
Method resolution must support multiple candidates

### LINKED ISSUE #1588: Bad method resolution based on mutability
I tried this code: https://godbolt.org/z/f9acosW78

```rust
extern "C" {
    fn printf(s: *const i8, ...);
}

#[lang = "deref"]
pub trait Deref {
    type Target;

    fn deref(&self) -> &Self::Target;
}

impl<T> Deref for &T {
    type Target = T;

    fn deref(&self) -> &T {
        unsafe {
            let a = "imm_deref\n\0";
            let b = a as *const str;
            let c = b as *const i8;

            printf(c);
        }

        *self
    }
}

impl<T> Deref for &mut T {
    type Target = T;

    fn deref(&self) -> &T {
        unsafe {
            let a = "mut_deref\n\0";
            let b = a as *const str;
            let c = b as *const i8;

            printf(c);
        }

        *self
    }
}

fn main() -> i32 {
    let foo = &mut 123;
    let res: i32 = *foo;

    unsafe {
        let a = "%i\n\0";
        let b = a as *const str;
        let c = b as *const i8;

        printf(c, res);
    }

    0
}

```

I expected to see this happen: it called the mut_deref version

Instead, this happened: it defaults to the immutable reference impl block

### Meta

- What version of Rust GCC were you using, git sha if possible. 3e7320510e7bb774295ecbd8089a9f51d6475959

philberty:
This patch fixes bad method resolution in our operator_overload_9 case.
When we have a &mut reference to something and we deref we must resolve to
the mutable reference impl block. The interface we are using to resolve
methods is the can_eq interface which allows for permissive mutability
which means allowing for mutable reference being unified with an immutable
one. This meant we actual match against both the immutable and mutable
version leading to multiple candidate error.

Fixes #1588

### DIFF
--- a/gcc/rust/typecheck/rust-hir-dot-operator.h
+++ b/gcc/rust/typecheck/rust-hir-dot-operator.h
@@ -35,6 +35,13 @@ struct MethodCandidate
   }
 
   bool is_error () const { return candidate.is_error (); }
+
+  DefId get_defid () const { return candidate.get_defid (); }
+
+  bool operator< (const MethodCandidate &c) const
+  {
+    return get_defid () < c.get_defid ();
+  }
 };
 
 class MethodResolver : private TypeCheckBase, protected AutoderefCycle
@@ -46,9 +53,10 @@ public:
     TyTy::FnType *fntype;
   };
 
-  static MethodCandidate Probe (const TyTy::BaseType *receiver,
-				const HIR::PathIdentSegment &segment_name,
-				bool autoderef_flag = false);
+  static std::set<MethodCandidate>
+  Probe (const TyTy::BaseType *receiver,
+	 const HIR::PathIdentSegment &segment_name,
+	 bool autoderef_flag = false);
 
   static std::vector<predicate_candidate> get_predicate_items (
     const HIR::PathIdentSegment &segment_name, const TyTy::BaseType &receiver,
@@ -68,7 +76,7 @@ private:
   std::vector<MethodResolver::predicate_candidate> predicate_items;
 
   // mutable fields
-  MethodCandidate try_result;
+  std::set<MethodCandidate> result;
 };
 
 } // namespace Resolver
--- a/gcc/rust/typecheck/rust-tyty.h
+++ b/gcc/rust/typecheck/rust-tyty.h
@@ -135,6 +135,11 @@ protected:
   std::vector<TypeBoundPredicate> specified_bounds;
 };
 
+extern void
+set_cmp_autoderef_mode ();
+extern void
+reset_cmp_autoderef_mode ();
+
 class TyVisitor;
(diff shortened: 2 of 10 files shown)


gcc/rust/typecheck/rust-hir-dot-operator.h
```diff
@@ -35,6 +35,13 @@ struct MethodCandidate
   }
 
   bool is_error () const { return candidate.is_error (); }
+
+  DefId get_defid () const { return candidate.get_defid (); }
+
+  bool operator<(const MethodCandidate &c) const
```
CohenArthur, line 41: Argh :D Operator overloading! Blasphemy! 

```suggestion
  bool less_than(const MethodCandidate &c) const
```

So that 
```c++
method_candidate1 < method_candidate2;
```

becomes
```c++
method_candidate1.less_than(method_candidate2);
```

Is this for a `std::hashmap` or `std::set` or something and I'm not realizing?

philberty: Yeah it seems if we implement operator< std::set works for us is why I implemented it. I really wish STL was more sane than this...

gcc/rust/typecheck/rust-tyty.h
```diff
@@ -135,6 +135,11 @@ class TypeBoundsMappings
   std::vector<TypeBoundPredicate> specified_bounds;
 };
 
+extern void
+set_cmp_autoderef_mode ();
```
philberty, line 139: This is the hack i mentioned @CohenArthur I would be much easier for us to live with this hack for now. Its going to take a few PR's to get our method resolution closer to that match statement in Rustc

dafaust (commented):
I don't think I'm familiar enough with the code context and how we currently do method resolution to be qualified to give a real approval. But from what I do understand this looks like a good fix for the linked issue and a step in the right direction even with the hack and the code changes make sense so... lgtm

gcc/rust/typecheck/rust-tyty.h
CohenArthur, line 139:
```suggestion
void
set_cmp_autoderef_mode ();
```

nit

gcc/rust/typecheck/rust-tyty.h
```diff
@@ -135,6 +135,11 @@ class TypeBoundsMappings
...
+extern void
+reset_cmp_autoderef_mode ();
```
CohenArthur, line 141:
```suggestion
void
reset_cmp_autoderef_mode ();
```

nit

CohenArthur (approved):
I think the hack is okay for now and that PR looks good :)

PR #1590, opened 2022-10-13
AST: more dump improvements

dafaust:
This PR adds new AST dump visitors for several nodes, and cleans up some minor formatting issues in blocks without tail expressions and if expressions.

### DIFF
--- a/gcc/rust/ast/rust-ast-dump.cc
+++ b/gcc/rust/ast/rust-ast-dump.cc
@@ -449,19 +449,43 @@ Dump::visit (GroupedExpr &expr)
 
 void
 Dump::visit (ArrayElemsValues &elems)
-{}
+{
+  auto &vals = elems.get_values ();
+  if (vals.size () >= 1)
+    {
+      vals[0]->accept_vis (*this);
+      for (size_t i = 1; i < vals.size (); i++)
+	{
+	  stream << ", ";
+	  vals[i]->accept_vis (*this);
+	}
+    }
+}
 
 void
 Dump::visit (ArrayElemsCopied &elems)
-{}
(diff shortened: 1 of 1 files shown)


CohenArthur (approved):
LGTM! Thank you :)

philberty (approved):
LGTM

philberty:
I think we are approaching the point where we should get rid of our old AST dump and replace it with this now.

gcc/rust/ast/rust-ast-dump.cc
```diff
@@ -399,23 +474,51 @@ Dump::visit (CompoundAssignmentExpr &expr)
...
+  expr.get_expr_in_parens ()->accept_vis (*this);
+  stream << ')';
+}
 
 void
 Dump::visit (ArrayElemsValues &elems)
-{}
+{
+  auto &vals = elems.get_values ();
+  if (vals.size () >= 1)
```
jdupak, line 487: This is a recurring pattern which is not very nice to read. I think we should make a `emit_list_without_trailing_comma` method. I can add it in some next PR.

CohenArthur: That's a very good idea! Feel free to do so, that would be a nice refactor. You'll probably need to use templates

CohenArthur:
> I think we are approaching the point where we should get rid of our old AST dump and replace it with this now.

Not yet! I think it's still missing some very important things like macros or extra info helpful not only for exporting but for debugging the AST. Since the old one had all that information, this one should as well (with maybe a `verbose` member so that the exporter doesn't get all that extra info for nothing)

PR #1593, opened 2022-10-14
ast: dump TypeAlias

jdupak:
Implements ast dump for TypeAlias

### DIFF
--- a/gcc/rust/ast/rust-ast-dump.cc
+++ b/gcc/rust/ast/rust-ast-dump.cc
@@ -710,7 +710,6 @@ Dump::visit (Function &function)
 {
   emit_visibility (function.get_visibility ());
   stream << "fn " << function.get_function_name ();
-
   if (function.has_generics ())
     emit_generic_params (function.get_generic_params ());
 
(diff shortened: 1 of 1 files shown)


gcc/rust/ast/rust-ast-dump.cc
```diff
@@ -746,7 +745,24 @@ Dump::visit (Function &function)
 
 void
 Dump::visit (TypeAlias &type_alias)
-{}
+{
+  // TODO does this represent associated type
```
philberty, line 749: Good catch, luckily with our AST class hierarchy for associated types we have: AST::TraitItemType

gcc/rust/ast/rust-ast-dump.cc
```diff
@@ -746,7 +745,24 @@ Dump::visit (Function &function)
...
+  // Visibility? type IDENTIFIER GenericParams? ( : TypeParamBounds )?
+  // WhereClause? ( = Type)? ;
+
+  if (type_alias.has_visibility ())
+    emit_visibility (type_alias.get_visibility ());
+  stream << "type " << type_alias.get_new_type_name ();
+  if (type_alias.has_generics ())
+    emit_generic_params (type_alias.get_generic_params ());
+  if (type_alias.has_where_clause ())
+    {} // FIXME: WhereClause
```
philberty, line 761: This could be your next dump it will be tricky but a good learning experience for you.

philberty (approved):
LGTM

philberty:
This is my fault but we might consider updating our clang-format CI here to a later clang-format I was on an old one I think clang-format 10 and there is a breaking change between 10 and >10 so you might need to fix your formatting here manually for now until we fix this.

gcc/rust/ast/rust-ast-dump.cc
jdupak: I just follow the syntax description in the Rust reference. It seems quite useful to get all the cases.

gcc/rust/ast/rust-ast-dump.cc
```diff
@@ -746,7 +745,24 @@ Dump::visit (Function &function)
...
+  stream << " = ";
+  type_alias.get_type_aliased ()->accept_vis (*this);
+  stream << ";\n\n";
```
CohenArthur, line 764: Very tiny nitpick: Why two newlines? Is there a specific reason?

No need to fix this right now, I think it's completely fine to leave it that way. It produces valid Rust code and that's what matters :)

CohenArthur (approved):
LGTM! Thank you! I'll merge this as soon as the formatting issues are resolved

gcc/rust/ast/rust-ast-dump.cc
jdupak: I had an example for that but it actually does not make sense in the general case. Fixed.

PR #1594, opened 2022-10-14
Refactor TraitResolver to not require a visitor

philberty:
We used a visitor to dispatch the HIR::Item so that we could cast it directly to an HIR::Trait and then check for nullptr if it failed. This patch changes this to simply use our new enum item_type_kind to switch so we can directly static_cast.


CohenArthur (approved):
LGTM

PR #1595, opened 2022-10-14
Support outer attribute handling on trait items just like normal items

philberty:
This patch adds a proxy class ItemWrapper to be a proxy allowing us to use the same code paths for attribute handling as we have with normal items. We need this so we can grab the fn trait associated type lang item's. Which are being missed currently.

Addresses #195


CohenArthur (approved):
LGTM

PR #1597, opened 2022-10-15
ast: Module: unloaded module and inner attributes

jdupak:
Adds support for unloaded modules and dumps also inner attributes.

Resolves issue found in #1548.

### DIFF
--- a/gcc/rust/ast/rust-ast-dump.cc
+++ b/gcc/rust/ast/rust-ast-dump.cc
@@ -810,21 +810,44 @@ Dump::visit (Method &method)
 void
 Dump::visit (Module &module)
 {
+  //  Syntax:
+  //	mod IDENTIFIER ;
+  //     | mod IDENTIFIER {
+  //	  InnerAttribute*
+  //	  Item*
+  //	}
+
   emit_visibility (module.get_visibility ());
-  stream << "mod " << module.get_name () << " {\n";
+  stream << "mod " << module.get_name ();
 
-  indentation.increment ();
-
-  for (auto &item : module.get_items ())
+  if (module.get_kind () == Module::UNLOADED)
     {
-      stream << indentation;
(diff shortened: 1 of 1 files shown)


gcc/rust/ast/rust-ast-dump.cc
```diff
@@ -810,21 +810,44 @@ Dump::visit (Method &method)
...
-  indentation.decrement ();
+      indentation.increment ();
 
-  stream << indentation << "}\n";
+      for (auto &item : module.get_inner_attrs ())
+	{
+	  stream << indentation;
+	  emit_attrib (item);
+	  stream << '\n';
+	}
```
CohenArthur, line 838: I think we will need a helper method for emitting multiple attributes as that should be done in a lot more places and I've been lacking haha. It's great that you thought about it

CohenArthur (approved):
Nice job! This looks great. Adding the bit of syntax reference helps a lot for the dumps, I think that's a very good idea.

bors r+

gcc/rust/ast/rust-ast-dump.cc
philberty: I think we have that typedef to AST::AttrVec or something is pretty useful we could have:
```
void emit_outer_attributes(AST::AttrVec& attribs) {
  <insert your for loop here>
}
```

Then you can refactor the other visitors to do the same

jdupak: I am already working on something related. Printing a list on lines using a emit method (or accept_vis) is extremely common and something like this (see below) seems very readable:
```rust
void
Dump::visit (CallExpr &expr)
{
  expr.get_function_expr ()->accept_vis (*this);
  stream << "(\n";
  emit_lines (expr.get_params (), ",");
  stream << indentation << ')';
}
```
The name is WIP ;)

PR #1606, opened 2022-10-18
Perform expansion pass in a fixed point fashion

CohenArthur:
This also adds proper scoping to the Early Name Resolver. The PR is still a draft as some regressions are still present:

~~1. `compile/macro17.rs`~~

~~The test message about reaching recursion limit does not contain location info, hence the error. This should be easy enough to fix~~

2. `compile/builtin_macro_recurse.rs`

This is going to be much harder. Because the EarlyNameResolver is now wired to work in a fixed point fashion, it's much harder to eagerly expand builtin macros compared to before. We'll need to figure that one out. @liushuyu if you have any ideas I'm all ears :)

### DIFF
--- a/gcc/rust/ast/rust-macro.h
+++ b/gcc/rust/ast/rust-macro.h
@@ -545,8 +545,30 @@ protected:
   }
 };
 
+/**
+ * All builtin macros possible
+ */
+enum class BuiltinMacro
+{
+  Assert,
+  File,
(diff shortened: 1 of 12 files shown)


CohenArthur:
I'm hoping that next week I'll have the time to finish this PR. I think I know *how* to go about it, I just need to find the time

PR #1607, opened 2022-10-19
Improve AST Fragment class

CohenArthur:
This changes the APIs around creating AST fragments and refactors the class into its own header and source file, hopefully making it easier to use. This will also help creating "unexpanded" AST fragments for proper builtin macro expansion with the new fixed-point algorithm introduced by #1606 

@liushuyu pinging you since you've worked extensively with the macro system. Would love your review!


philberty (approved):
LGTM

PR #1613, opened 2022-10-24
builtins: Rename all bang macro handlers

CohenArthur:
This renames all builtin macro functions to `<macro>_handler`. This helps avoiding the `ifdef/undef` dance we performed for MacOS in order to compile the `assert` method


dafaust (approved):
Makes sense to me!

PR #1614, opened 2022-10-25
intrinsics: Add rust_sorry wrapper for unimplemented intrinsics

CohenArthur:
This allows us to define intrinsics without implementing their body. This will be useful for atomic intrinsics for example, as there are a lot of them and we should work on implementing them one by one properly and slowly


dafaust (approved):
LGTM!

PR #1615, opened 2022-10-25
intrinsics: Add early implementation for atomic_store_{seqcst, relaxed, release}

CohenArthur:
Needs #1614 so ignore the first commit

This commit adds support for three `atomic_store_*` intrinsics declared
in the core library. The mapping is as follows:

- atomic_store_seqcst(dst, val) -> __atomic_store_n(dst, val,
  __ATOMIC_SEQ_CST)
- atomic_store_relaxed(dst, val) -> __atomic_store_n(dst, val,
  __ATOMIC_RELAXED)
- atomic_store_release(dst, val) -> __atomic_store_n(dst, val,
  __ATOMIC_RELEASE)

The one remaining question is whether `atomic_store_unordered` can be
abstracted as an atomic store with the __ATOMIC_RELAXED ordering.

This commit also performs the overloading "by hand": Replacing
`atomic_store_release<i32>` with `__atomic_store_4` as I cannot get the
generic version to work. This will be done in future improvements.

### DIFF
--- a/gcc/rust/backend/rust-compile-intrinsic.cc
+++ b/gcc/rust/backend/rust-compile-intrinsic.cc
@@ -93,6 +93,17 @@ prefetch_write_data (Context *ctx, TyTy::FnType *fntype)
   return prefetch_data_handler (ctx, fntype, Prefetch::Write);
 }
 
+static tree
+atomic_store_handler_inner (Context *ctx, TyTy::FnType *fntype, int ordering);
+
+static inline std::function<tree (Context *, TyTy::FnType *)>
+atomic_store_handler (int ordering)
+{
+  return [ordering] (Context *ctx, TyTy::FnType *fntype) {
+    return atomic_store_handler_inner (ctx, fntype, ordering);
+  };
+}
+
 static inline tree
 sorry_handler (Context *ctx, TyTy::FnType *fntype)
 {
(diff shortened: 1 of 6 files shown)


bjorn3:
> The one remaining question is whether atomic_store_unordered can be
abstracted as an atomic store with the __ATOMIC_RELAXED ordering.

It can. LLVM says for the monotonic memory ordering (which is the LLVM name of relaxed) that `In addition to the guarantees of unordered, [...]`. In other words unordered is strictly weaker than monotonic. Note that we don't actually expose unordered to the user in stable rust due to the lack of a formal memory model for it.

gcc/rust/backend/rust-compile-intrinsic.cc
```diff
@@ -93,20 +93,46 @@ prefetch_write_data (Context *ctx, TyTy::FnType *fntype)
...
+    {"wrapping_mul", wrapping_mul_handler},
+    {"copy_nonoverlapping", copy_nonoverlapping_handler},
+    {"prefetch_read_data", prefetch_read_data},
+    {"prefetch_write_data", prefetch_write_data},
+    {"atomic_load", sorry_handler},
+    {"atomic_store_seqcst", atomic_store_handler (__ATOMIC_SEQ_CST)},
+    {"atomic_store_release", atomic_store_handler (__ATOMIC_RELEASE)},
+    {"atomic_store_relaxed", atomic_store_handler (__ATOMIC_RELAXED)},
+    {"atomic_store_unordered", sorry_handler},
+    // FIXME: Are _unordered and _relaxed the same?
```
dafaust, line 135: Went hunting a bit since I haven't heard of this.

https://llvm.org/docs/Atomics.html#unordered
Looks like no, `unordered` is even weaker than `relaxed` and comes from Java rather than C/C++.

https://llvm.org/docs/LangRef.html#atomic-memory-ordering-constraints
`This ordering cannot be specified for read-modify-write operations; it is not strong enough to make them atomic in any interesting way.`

That said, since the guarantees of `unordered` are strictly weaker than `relaxed` it should be fine to map this to `__ATOMIC_RELAXED`.

dafaust: Heh bjorn beat me to it :P

CohenArthur:
@bjorn3 @dafaust thank you both! I will put that in this PR ASAP

PR #1620, opened 2022-10-26
Intrinsics unchecked ops

CohenArthur:
Addresses #658

### DIFF
--- a/gcc/rust/backend/rust-compile-intrinsic.cc
+++ b/gcc/rust/backend/rust-compile-intrinsic.cc
@@ -34,6 +34,22 @@
 namespace Rust {
 namespace Compile {
 
+static bool
+is_basic_integer_type (TyTy::BaseType *type)
+{
+  switch (type->get_kind ())
+    {
+    case TyTy::INT:
+    case TyTy::UINT:
+    case TyTy::USIZE:
+    case TyTy::ISIZE:
+      return true;
+    default:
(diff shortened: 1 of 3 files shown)


gcc/rust/backend/rust-compile-intrinsic.cc
```diff
@@ -34,6 +34,22 @@
 namespace Rust {
 namespace Compile {
 
+static bool
+is_basic_integer_type (TyTy::BaseType *type)
+{
+  switch (type->get_kind ())
+    {
+    case TyTy::INT:
+    case TyTy::UINT:
```
bjorn3, line 43: Do these cover all u* and i* sizes other than usize and isize?

CohenArthur: Yes, they do. This is similar to [`rustc`'s`TyKind` enum](https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/ty/enum.TyKind.html) except we can't have sum types, so we're using inheritance to counter that. All of our `TyTy` nodes inherit from the `BaseType` class, which exposes `get_kind()`. Then specifically the `IntType` and `UintType` classes (which inherit from `BaseType`) expose one new enum each and a `get_{u}int_kind()` method

dafaust (approved):
Cool, LGTM

PR #1622, opened 2022-10-27
intrinsics: Add atomic_load_*

CohenArthur:


gcc/rust/backend/rust-builtins.cc
```diff
@@ -71,36 +71,40 @@ BuiltinsContext::setup_math_fns ()
...
+				     integer_type_node, NULL_TREE);
+  };
+
+  // define_builtin ("atomic_store", BUILT_IN_ATOMIC_STORE, "__atomic_store",
+  // NULL,
+  //   atomic_store_type, 0);
+  // define_builtin ("atomic_store_n", BUILT_IN_ATOMIC_STORE_N,
+  // "__atomic_store_n",
+  //   NULL, atomic_store_type, 0);
+
```
dkm, line 88: Is that some leftover comments or is it intended? If the latter, maybe a comment explaining why it's commented would be good?

CohenArthur: Ah, good catch. It's because I cannot get these versions (which are generic and don't require me to specify the size) to work properly. I'll add a comment explaining it

PR #1624, opened 2022-10-28
Ast dump refactor

jdupak:
Unifies all methods to overloads of the visit method and adds new helpers for common patterns (join by separator, visit as lines...)

### DIFF
--- a/gcc/rust/ast/rust-ast-dump.cc
+++ b/gcc/rust/ast/rust-ast-dump.cc
@@ -101,7 +101,8 @@ Dump::visit_items_as_lines (T &collection, const std::string &trailing)
 
 template <typename T>
 void
-Dump::visit_items_as_block (T &collection, char left_brace, char right_brace)
+Dump::visit_items_as_block (T &collection, const std::string &line_trailing,
+			    char left_brace, char right_brace)
 {
   if (collection.empty ())
     {
@@ -112,7 +113,7 @@ Dump::visit_items_as_block (T &collection, char left_brace, char right_brace)
       stream << left_brace << '\n';
 
       indentation.increment ();
-      visit_items_as_lines (collection, ",");
+      visit_items_as_lines (collection, line_trailing);
       indentation.decrement ();
 
       stream << indentation << right_brace << '\n';
@@ -944,7 +945,7 @@ Dump::visit (StructStruct &struct_item)
 
   // FIXME: where-clause
 
-  visit_items_as_block (struct_item.get_fields ());
+  visit_items_as_block (struct_item.get_fields (), ",");
 }
 
 void
@@ -979,7 +980,7 @@ void
 Dump::visit (EnumItemStruct &item)
 {
   stream << item.get_identifier ();
-  visit_items_as_block (item.get_struct_fields ());
+  visit_items_as_block (item.get_struct_fields (), ",");
 }
 
 void
@@ -998,7 +999,7 @@ Dump::visit (Enum &enum_item)
 
   // FIXME: where-clause
 
-  visit_items_as_block (enum_item.get_variants ());
+  visit_items_as_block (enum_item.get_variants (), ",");
 }
 
 void
@@ -1010,7 +1011,7 @@ Dump::visit (Union &union_item)
 
   // FIXME: where-clause
 
-  visit_items_as_block (union_item.get_variants ());
+  visit_items_as_block (union_item.get_variants (), ",");
 }
 
 void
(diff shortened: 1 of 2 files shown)


gcc/rust/ast/rust-ast-dump.cc
```diff
@@ -62,28 +57,86 @@ Dump::go (AST::Item &item)
...
+  visit (item);
+  stream << trailing << '\n';
+}
+
+template <typename T>
+void
+Dump::visit_items_as_lines (T &collection, const std::string &trailing)
+{
+  for (auto &item : collection)
+    visit_as_line (item);
```
CohenArthur, line 97:
```suggestion
    visit_as_line (item, trailing);
```
Unless I'm misunderstanding something?

gcc/rust/ast/rust-ast-dump.cc
```diff
@@ -62,28 +57,86 @@ Dump::go (AST::Item &item)
...
     {
-      const auto &seg = attrib.get_path ().get_segments ().at (i);
-      bool has_next = (i + 1) < attrib.get_path ().get_segments ().size ();
+      stream << left_brace << '\n';
 
-      stream << seg.get_segment_name ();
-      if (has_next)
-	stream << "::";
+      indentation.increment ();
+      visit_items_as_lines (collection, ",");
```
CohenArthur, line 113: I'm not sure I understand why block items should always have a trailing comma character. In the case of an extern block, that comma should be a semicolon if I'm understanding correctly.

gcc/rust/ast/rust-ast-dump.cc
```diff
@@ -951,49 +945,24 @@ Dump::visit (StructStruct &struct_item)
...
-      stream << '\n' << indentation;
-      format_struct_field (field);
-      stream << ',';
-    }
-  indentation.decrement ();
-
-  if (fields.size () > 0)
-    stream << '\n' << indentation;
-  stream << "}\n";
+  visit_items_as_block (struct_item.get_fields (), '{', '}');
```
CohenArthur, line 945: One thing to be careful of, which I can't remember, is whether or not our AST distinguishes between
```rust
struct Foo;
```
and 
```rust
struct Foo {}
```
as they are quite different and should be dumped differently

gcc/rust/ast/rust-ast-dump.h
```diff
@@ -73,46 +73,60 @@ class Dump : public ASTVisitor
...
+   * types implementing `accept_vis` method and for classes for which the
+   * `visit` method is directly implemented.
    */
-  void format_function_common (std::unique_ptr<Type> &return_type,
-			       std::unique_ptr<BlockExpr> &block);
+  template <typename T> void visit (std::unique_ptr<T> &node);
 
   /**
-   * Format a function's definition parameter
+   * Visit all items in given collection, placing the separator in between but
```
CohenArthur, line 84:
```suggestion
   * Visit all items in a given collection, placing the separator in between but
```

gcc/rust/ast/rust-ast-dump.h
```diff
@@ -73,46 +73,60 @@ class Dump : public ASTVisitor
...
+   *
+   * @see visit_as_line
    */
-  void emit_visibility (const Visibility &vis);
+  template <typename T>
+  void visit_items_as_lines (T &collection, const std::string &trailing = "");
 
   /**
-   * Emit an indented string with an optional extra comment
+   * Visit each item in collection as lines inside a block delimited by braces
```
CohenArthur, line 107:
```suggestion
   * Visit each item in a collection as lines inside a block delimited by braces
```
or
```suggestion
   * Visit each item in @collection as lines inside a block delimited by braces
```
if you meant to refer to the parameter name?

CohenArthur (approved):
This looks great! Thanks a lot for spending the time to clean it up

gcc/rust/ast/rust-ast-dump.cc
jdupak: I looked into the parser and yes, that case should be handled. There is a bool `is_unit` which determined it. But is in not related to this refactor - it was not handled before, so I will open a new PR.

jdupak: Solved by #1625

philberty:
I added a comment about the failed test here: https://github.com/Rust-GCC/gccrs/pull/1625#issuecomment-1295969536

CohenArthur:
As mentioned privately, the issue for the tests failing is the dumping of the extra `Self` parameter for traits. So once that is fixed I'll merge this and you can attend #1625. Great work!

jdupak:
I have fixed the issue by adding offset arguments to the list dumping function to allow "slicing". Does this seem reasonable to you?

CohenArthur (approved):
LGTM :) Thanks for fixing all the reviews!

bors r+

PR #1630, opened 2022-11-04
Cleanup builtin handling

CohenArthur:
This PR adds proper GCC-like handling of builtins, which allows us to not re-define all the builtins we use and fight with declaring the types each time.

This is also safer and works better. Sadly, this has unearthed some problems in our handling of builtins, specifically the ones I've added recently (atomic ones and data prefetching). These are the two issues that remain for this PR to be in a nice state which does not cause regressions:

1. `atomic_{store, load}` (and maybe cmp_xchg) do not work on signed integers. I'm currenty working around this by simply using `u32`s instead of `i32`s for the tests, but that's not a valid solution. These intrinsics need to work on all integer types and I will thus need to add some conversions from `i*` to `u*` before calling the intrinsics. The upside is that with these cleanups we should now be able to handle `*size` types easily and cleanly for those intrinsics
2. `__builtin_prefetch()` requires the `locality` argument (third argument) to be a const value. While LLVM will let you build the function and maybe error out later down the line, GCC does not let you pass non compile time known values as locality arguments. Because we are trying to build the following intrinsic:

```rust
fn prefetch_read_data<T>(src: *const T, locality: i32) {
    __builtin_prefetch(src, 1, locality);
}
```

we cannot know that the `locality` arg is a compile-time constant. There are two ways to go around this problem:

a. Try to constant fold the `locality` argument. If this works, it's a compile time constant and we can keep going
b. Somehow map a generic intrinsic directly to a GCC one and inserting a new argument. So instead of generating something like

```rust
fn prefetch_read_data<i32>(src: *const i32, locality: i32) {
    __builtin_prefetch(src, 0, locality)
}
```

we'd swap the call to `prefetch_read_data::<i32>(src, locality)` with `__builtin_prefetch(src, 0, locality)` and enforce `locality` being a compile time constant.

Edited because dynamically dispatching a *compiler hint* does not make any sense at all.

### DIFF
--- a/gcc/rust/backend/rust-builtins.cc
+++ b/gcc/rust/backend/rust-builtins.cc
@@ -14,7 +14,12 @@
 // along with GCC; see the file COPYING3.  If not see
 // <http://www.gnu.org/licenses/>.
 
+#include "rust-diagnostics.h"
+#include "rust-system.h"
 #include "rust-builtins.h"
+
+#include "target.h"
+#include "stringpool.h"
 
 namespace Rust {
 namespace Compile {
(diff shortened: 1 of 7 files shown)


philberty:
I am not sure about this PR, notice when we denied builtins before:

```
define_builtin ("sinf32", BUILT_IN_SINF, "__builtin_sinf", "sinf",
		  math_function_type_f32, builtin_const);
  define_builtin ("sqrtf32", BUILT_IN_SQRTF, "__builtin_sqrtf", "sqrtf",
		  math_function_type_f32, builtin_const);
```

we specified what the type we expect is. For example this is a math f32 builtin if we need to support f64 we need to specify a f64 builtin etc. If you define a generic builtin your going to need to do a lot of specific type casting which is probably not very safe.

I am all on board with making this stuff more GCC'like but I don't really see what this PR is solving other than raising another regression.

PR #1631, opened 2022-11-04
Ast dump where clause

jdupak:
Adds dump of the where clause and recursively needed nodes.
Adds visitor compatibility layer for general references except for unique pointers - needed for lifetimes. The enable is necessary to coexist with the unique_ptr overload.

Depends on #1624 and will be rebased when it is merged.

### DIFF
--- a/gcc/rust/ast/rust-ast-dump.h
+++ b/gcc/rust/ast/rust-ast-dump.h
@@ -81,6 +81,11 @@ private:
   template <typename T> void visit (std::unique_ptr<T> &node);
 
   /**
+   * @see visit<std::unique_ptr<T>>
+   */
+  template <typename T> void visit (T &node);
+
+  /**
    * Visit all items in given @collection, placing the separator in between but
    * not at the end.
    * Start and end offset allow to visit only a "slice" from the collection.
@@ -122,12 +127,12 @@ private:
 			      std::unique_ptr<BlockExpr> &block);
 
   void visit (FunctionParam &param);
-  void visit (const Attribute &attrib);
-  void visit (const Visibility &vis);
+  void visit (Attribute &attrib);
+  void visit (Visibility &vis);
   void visit (std::vector<std::unique_ptr<GenericParam>> &params);
   void visit (TupleField &field);
   void visit (StructField &field);
-  void visit (const SimplePathSegment &segment);
+  void visit (SimplePathSegment &segment);
   void visit (NamedFunctionParam &param);
   void visit (MacroRule &rule);
 
--- a/gcc/rust/ast/rust-ast-dump.h
+++ b/gcc/rust/ast/rust-ast-dump.h
@@ -135,6 +135,8 @@ private:
   void visit (SimplePathSegment &segment);
   void visit (NamedFunctionParam &param);
   void visit (MacroRule &rule);
+  void visit (WhereClause &rule);
+  void visit (std::vector<LifetimeParam> &for_lifetimes);
 
   // rust-ast.h
   void visit (Token &tok);
--- a/gcc/rust/ast/rust-ast-dump.cc
+++ b/gcc/rust/ast/rust-ast-dump.cc
@@ -227,6 +227,28 @@ Dump::visit (StructField &field)
   visit (field.get_field_type ());
 }
 
+// TODO is this unique by type?
+void
+Dump::visit (std::vector<LifetimeParam> &for_lifetimes)
+{
+  // ForLifetimes :
+  //     for GenericParams
+  //
+  // GenericParams :
+  //      < >
+  //   | < (GenericParam ,)* GenericParam ,? >
+  //
+  // GenericParam :
+  //   OuterAttribute* ( LifetimeParam | TypeParam | ConstParam )
+  //
+  // LifetimeParam :
+  //   LIFETIME_OR_LABEL ( : LifetimeBounds )?
+
+  stream << "for <";
+  visit_items_joined_by_separator (for_lifetimes, " + ");
+  stream << "> ";
+}
+
 void
 Dump::visit (Token &tok)
 {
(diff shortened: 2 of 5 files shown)


jdupak:
Example input, taken from the Rust reference examples:
```rust
#![allow(unused)]

fn main() {
    struct B<T>
        where
            T: Iterator,            // Could use A<T: Iterator> instead
            T::Item: Copy,          // Bound on an associated type
            String: PartialEq<T>,   // Bound on `String`, using the type parameter
            i32: Default,           // Allowed, but not useful
    {
        f: T,
    }

    type Surface = i32;
    trait Shape {
        fn draw(&self, surface: Surface);
        fn name() -> &'static str;
    }

    fn draw_twice<T: Shape>(surface: Surface, sh: T) {
        sh.draw(surface);           // Can call method because T: Shape
        sh.draw(surface);
    }

    fn copy_and_draw_twice<T: Copy>(surface: Surface, sh: T) where T: Shape {
        let shape_copy = sh;        // doesn't move sh because T: Copy
        draw_twice(surface, sh);    // Can use generic function because T: Shape
    }

    struct Figure<S: Shape>(S, S);

    fn name_figure<U: Shape>(
        figure: Figure<U>,          // Type Figure<U> is well-formed because U: Shape
    ) {
        println!(
            "Figure of two {}",
            U::name(),              // Can use associated function
        );
    }

    struct A<'a, T>
        where
            i32: Default,           // Allowed, but not useful
            i32: Iterator,          // Error: `i32` is not an iterator
            &'a mut T: Copy,        // (at use) Error: the trait bound is not satisfied
            [T]: Sized,             // (at use) Error: size cannot be known at compilation
    {
        f: &'a T,
    }
    struct UsesA<'a, T>(A<'a, T>);

    fn f<'a, 'b>(x: &'a i32, mut y: &'b i32) where 'a: 'b {
        y = x;                      // &'a i32 is a subtype of &'b i32 because 'a: 'b
        let r: &'b &'a i32 = &&0;   // &'b &'a i32 is well formed because 'a: 'b
    }

    impl<'a> PartialEq<i32> for &'a T {
        // ...
    }

    fn call_on_ref_zero<F>(f: F) where for<'a> F: Fn(&'a i32) {
        let zero = 0;
        f(&zero);
    }

    fn call_on_ref_zero2<F>(f: F) where F: for<'a> Fn(&'a i32) {
        let zero = 0;
        f(&zero);
    }
}
```

Current output:
```rust
fn main() {
	struct B<T> where
		T: Iterator,
		T::Item: Copy,
		String: PartialEq<Ambiguous: T>,
		i32: Default,
{
		f: T,
	}
; /* stmt */
	type Surface = i32;
; /* stmt */
	trait Shape{
				fn draw(&self, surface: Surface) ;

				fn name() -> str;

	}
; /* stmt */
	fn draw_twice<T>(surface: Surfacesh: T) {
		; /* stmt */
		; /* stmt */
	}

; /* stmt */
	fn copy_and_draw_twice<T>(surface: Surfacesh: T)  where
		T: Shape,
{
		let shape_copy = sh; /* stmt */
		draw_twice(
			surface,
			sh,
		); /* stmt */
	}

; /* stmt */
	struct Figure<S>(S, S);
; /* stmt */
	fn name_figure<U>(figure: Figure<Ambiguous: U>) {
		; /* stmt */
	}

; /* stmt */
	struct A<'a, T> where
		i32: Default,
		i32: Iterator,
		T: Copy,
		T: Sized,
{
		f: T,
	}
; /* stmt */
	struct UsesA<'a, T>(A<'aAmbiguous: T>);
; /* stmt */
	fn f<'a, 'b>(x: i32y: i32)  where
		'a: 'b,
{
		y = x; /* stmt */
		let r: i32 = &&0; /* stmt */
	}

; /* stmt */
	impl PartialEq<Ambiguous: i32> for T {

}
; /* stmt */
	fn call_on_ref_zero<F>(f: F)  where
		for <'a> F: Fn(&'a i32),
{
		let zero = 0; /* stmt */
		f(
			&zero,
		); /* stmt */
	}

; /* stmt */
	fn call_on_ref_zero2<F>(f: F)  where
		F: for <'a> Fn(&'a i32),
{
		let zero = 0; /* stmt */
		f(
			&zero,
		); /* stmt */
	}

; /* stmt */
```

gcc/rust/ast/rust-ast-dump.h
```diff
@@ -80,6 +80,18 @@ class Dump : public ASTVisitor
...
+   * @see visit<std::unique_ptr<T>>
+   *
+   * This ugly template trickery is necessary to make the visit overload work
+   * for both references to AST nodes/properties and unique-pointer ones.
+   * Otherwise this overload would be always preferred.
+   */
+  template <typename T1, typename T2,
+	    typename = std::enable_if<
+	      !std::is_convertible<T1, std::unique_ptr<T2>>::value>>
+  void visit (T1 &node);
```
CohenArthur, line 93: I'm not sure I understand why this is a problem or if we need to do that (but that might just be my C++ skills lacking).
In another part of the dump, we have the following code:
```c++
  /**
   * Use the AST Dump as a debugging tool
   */
  template <typename T> static void debug (T &instance)
  {
    auto dump = Dump (std::cerr);

    std::cerr << '\n';
    instance.accept_vis (dump);
    std::cerr << '\n';
  }
  template <typename T> static void debug (std::unique_ptr<T> &instance)
  {
    debug (*instance);
  }
```

Doesn't this effectively achieve the same behavior without the template condition magic?

CohenArthur (commented):
Looks great thank you! Just wondering about some details

gcc/rust/ast/rust-ast-dump.h
jdupak: I will try that. It would be great to get rid of that magic. I wonder if calling the other overload does the trick, because this is pretty much what I started with except that one was not calling the other, but both called accept_vis.

CohenArthur (approved):
LGTM! Thank you :)

PR #1635, opened 2022-11-10
rust: Remove unused variables and fix dangling references

CohenArthur:
This should make the bootstrap build green again.

This commit contains some aggressive refactoring which either removes unused arguments altogether or removes the argument name in some cases where it might get used later on or cannot change due to implementing a virtual method or something like that.

### DIFF
--- a/gcc/rust/backend/rust-compile-base.h
+++ b/gcc/rust/backend/rust-compile-base.h
@@ -55,7 +55,7 @@ protected:
     const TyTy::TypeBoundPredicate *predicate,
     std::vector<std::pair<Resolver::TraitReference *, HIR::ImplBlock *>>
       &receiver_bounds,
-    const TyTy::BaseType *receiver, const TyTy::BaseType *root, Location locus);
+    const TyTy::BaseType *root, Location locus);
 
   bool verify_array_capacities (tree ltype, tree rtype, Location ltype_locus,
 				Location rtype_locus);
@@ -63,7 +63,7 @@ protected:
   tree query_compile (HirId ref, TyTy::BaseType *lookup,
 		      const HIR::PathIdentSegment &final_segment,
 		      const Analysis::NodeMapping &mappings,
-		      Location expr_locus, bool is_qualified_path);
+		      Location expr_locus);
 
   tree resolve_adjustements (std::vector<Resolver::Adjustment> &adjustments,
 			     tree expression, Location locus);
@@ -113,7 +113,7 @@ protected:
 
   static tree indirect_expression (tree expr, Location locus);
 
-  static bool mark_addressable (tree, Location);
+  static bool mark_addressable (tree);
 
   static std::vector<Bvariable *>
   compile_locals_for_block (Context *ctx, Resolver::Rib &rib, tree fndecl);
--- a/gcc/rust/typecheck/rust-hir-trait-ref.h
+++ b/gcc/rust/typecheck/rust-hir-trait-ref.h
@@ -112,8 +112,7 @@ public:
 	break;
 
       case TYPE:
-	return get_type_from_typealias (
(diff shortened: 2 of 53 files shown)


philberty:
I dont like this change. Its all bundled into a single commit and you've changed quite a bit of the code in the type resolution to be copies and not references so its not clear what you are actually fixing or changing here.

gcc/rust/typecheck/rust-hir-trait-ref.h
```diff
@@ -149,8 +148,7 @@ class TraitItemReference
     return new TyTy::ErrorType (get_mappings ().get_hirid ());
   }
 
-  TyTy::BaseType *get_type_from_typealias (/*const*/
-					   HIR::TraitItemType &type) const;
+  TyTy::BaseType *get_type_from_typealias () const;
```
philberty, line 151: This isn't right we mght need to have the TraitItemType to do resolution of the associated type properly when it comes to handling opaque types and generic argument binding which still needs to be done.

gcc/rust/typecheck/rust-hir-type-check-expr.cc
```diff
@@ -1152,7 +1152,7 @@ TypeCheckExpr::visit (HIR::MethodCallExpr &expr)
       rust_debug_loc (expr.get_method_name ().get_generic_args ().get_locus (),
 		      "applying generic arguments to method_call: {%s}",
 		      lookup->debug_str ().c_str ());
-      HIR::GenericArgs &args = expr.get_method_name ().get_generic_args ();
+      HIR::GenericArgs args = expr.get_method_name ().get_generic_args ();
```
philberty, line 1155: Why take a copy here?

gcc/rust/backend/rust-compile-base.h
```diff
@@ -55,15 +55,15 @@ class HIRCompileBase
...
-    const TyTy::BaseType *receiver, const TyTy::BaseType *root, Location locus);
+    const TyTy::BaseType *root, Location locus);
 
   bool verify_array_capacities (tree ltype, tree rtype, Location ltype_locus,
 				Location rtype_locus);
 
   tree query_compile (HirId ref, TyTy::BaseType *lookup,
 		      const HIR::PathIdentSegment &final_segment,
 		      const Analysis::NodeMapping &mappings,
-		      Location expr_locus, bool is_qualified_path);
```
philberty, line 66: The removal of the is_qualified_path flag should really have been a single commit

gcc/rust/backend/rust-compile-base.h
```diff
@@ -55,15 +55,15 @@ class HIRCompileBase
     const TyTy::TypeBoundPredicate *predicate,
     std::vector<std::pair<Resolver::TraitReference *, HIR::ImplBlock *>>
       &receiver_bounds,
-    const TyTy::BaseType *receiver, const TyTy::BaseType *root, Location locus);
+    const TyTy::BaseType *root, Location locus);
```
philberty, line 58: https://github.com/Rust-GCC/gccrs/pull/1583 fixes resolution for super trait calls on dyn objects but code generation is not finished for this yet so we might still need the receiver here.

gcc/rust/typecheck/rust-hir-type-check-expr.cc
CohenArthur: This was due to an invalid warning upstream which I believe has now been fixed. I'll have a look.

CohenArthur:
> I dont like this change. Its all bundled into a single commit and you've changed quite a bit of the code in the type resolution to be copies and not references so its not clear what you are actually fixing or changing here.

This PR was for us to have a green bootstrap build again. In some places where parameters weren't used at all, I did remove them since I thought they'd stay unused forever

PR #1649, opened 2022-11-15
dwarf2out.c: Don't emit DW_LANG_Rust_old

dkm:
DW_LANG_Rust_old is a non-standard DWARF language code used by old rustc compilers before DWARF5 (released in 2017). Just always emit DW_LANG_Rust unless producing strict DWARF for versions before 5. And in that old strict DWARF case just emit DW_LANG_C instead of a non-standard language code.

### DIFF
--- a/gcc/dwarf2out.cc
+++ b/gcc/dwarf2out.cc
@@ -5600,14 +5600,15 @@ is_fortran (const_tree decl)
   return is_fortran ();
 }
 
-/* Return TRUE if the language is Rust.  */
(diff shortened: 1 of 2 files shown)


dkm:
Original patch from : https://inbox.sourceware.org/gcc-rust/Y16WupOiVe7+eNY1@wildebeest.org/

CohenArthur (approved):
LGTM :) Thanks for pushing it!!!! That's really helpful

PR #1651, opened 2022-11-16
Improve dump lex

### LINKED ISSUE #1650: Improve lexer dump
For now, `-frust-dump-lex` option does not work. It creates `gccrs.lex.dump` with no content because lex dump runs after a crate is parsed.
So I would like to improve lex dump so that lex dump runs at the same time as the parser runs.
I already almost completed modifying the code.

tamaroning:
Fixes #1650

### DIFF
--- a/gcc/rust/lex/rust-lex.cc
+++ b/gcc/rust/lex/rust-lex.cc
@@ -118,13 +118,15 @@ is_non_decimal_int_literal_separator (char character)
 
 Lexer::Lexer (const std::string &input)
   : input (RAIIFile::create_error ()), current_line (1), current_column (1),
-    line_map (nullptr), raw_input_source (new BufferInputSource (input, 0)),
+    line_map (nullptr), dump_lex_out (Optional<std::ofstream &>::none ()),
+    raw_input_source (new BufferInputSource (input, 0)),
     input_queue{*raw_input_source}, token_queue (TokenSource (this))
 {}
 
-Lexer::Lexer (const char *filename, RAIIFile file_input, Linemap *linemap)
+Lexer::Lexer (const char *filename, RAIIFile file_input, Linemap *linemap,
+	      Optional<std::ofstream &> dump_lex_opt)
   : input (std::move (file_input)), current_line (1), current_column (1),
-    line_map (linemap),
+    line_map (linemap), dump_lex_out (dump_lex_opt),
     raw_input_source (new FileInputSource (input.get_raw ())),
     input_queue{*raw_input_source}, token_queue (TokenSource (this))
 {
@@ -184,6 +186,45 @@ void
 Lexer::skip_input ()
 {
   skip_input (0);
+}
+
+void
+Lexer::skip_token (int n)
+{
+  // dump tokens if dump-lex option is enabled
+  if (dump_lex_out.is_some ())
+    dump_and_skip (n);
+  else
+    token_queue.skip (n);
+}
+
+void
+Lexer::dump_and_skip (int n)
+{
+  std::ofstream &out = dump_lex_out.get ();
+  bool found_eof = false;
+  const_TokenPtr tok;
+  for (int i = 0; i < n + 1; i++)
+    {
+      if (!found_eof)
+	{
+	  tok = peek_token ();
+	  found_eof |= tok->get_id () == Rust::END_OF_FILE;
+
+	  Location loc = tok->get_locus ();
+
+	  out << "<id=";
+	  out << tok->token_id_to_str ();
+	  out << (tok->has_str () ? (std::string (", text=") + tok->get_str ()
+				     + std::string (", typehint=")
+				     + std::string (tok->get_type_hint_str ()))
+				  : "")
+	      << " ";
+	  out << get_line_map ()->to_string (loc) << " ";
+	}
+
+      token_queue.skip (0);
+    }
 }
 
 void
--- a/gcc/rust/lex/rust-lex.h
+++ b/gcc/rust/lex/rust-lex.h
@@ -22,6 +22,7 @@
 #include "rust-linemap.h"
 #include "rust-buffered-queue.h"
 #include "rust-token.h"
+#include "rust-optional.h"
 
 namespace Rust {
 // Simple wrapper for FILE* that simplifies destruction.
@@ -139,7 +140,9 @@ private:
 
 public:
   // Construct lexer with input file and filename provided
-  Lexer (const char *filename, RAIIFile input, Linemap *linemap);
+  Lexer (const char *filename, RAIIFile input, Linemap *linemap,
+	 Optional<std::ofstream &> dump_lex_opt
+	 = Optional<std::ofstream &>::none ());
 
   // Lex the contents of a string instead of a file
   Lexer (const std::string &input);
@@ -161,9 +164,12 @@ public:
   const_TokenPtr peek_token () { return peek_token (0); }
 
   // Advances current token to n + 1 tokens ahead of current position.
-  void skip_token (int n) { token_queue.skip (n); }
+  void skip_token (int n);
   // Skips the current token.
   void skip_token () { skip_token (0); }
+
+  // Dumps and advances by n + 1 tokens.
+  void dump_and_skip (int n);
 
   // Replaces the current token with a specified token.
   void replace_current_token (TokenPtr replacement);
@@ -196,6 +202,8 @@ private:
   /* Max column number that can be quickly allocated - higher may require
    * allocating new linemap */
   static const int max_column_hint = 80;
+
+  Optional<std::ofstream &> dump_lex_out;
 
   // Input source wrapper thing.
   class InputSource
--- a/gcc/rust/rust-session-manager.cc
+++ b/gcc/rust/rust-session-manager.cc
@@ -439,7 +439,22 @@ Session::compile_crate (const char *filename)
   // parse file here
   /* create lexer and parser - these are file-specific and so aren't instance
    * variables */
-  Lexer lex (filename, std::move (file_wrap), linemap);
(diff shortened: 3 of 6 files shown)


tamaroning:
Currently, output is like this:
```rust
fn main() {
    let a: i32 = 1 + 2;
    return;
}
```

```
<id=FN_TOK a.rs:1:1 <id=IDENTIFIER, text=main, typehint=unknown a.rs:1:4 <id=LEFT_PAREN a.rs:1:8 <id=RIGHT_PAREN a.rs:1:9 <id=LEFT_CURLY a.rs:1:11 <id=LET a.rs:2:5 <id=IDENTIFIER, text=a, typehint=unknown a.rs:2:9 <id=COLON a.rs:2:10 <id=IDENTIFIER, text=i32, typehint=unknown a.rs:2:12 <id=EQUAL a.rs:2:16 <id=INT_LITERAL, text=1, typehint=pure_decimal a.rs:2:18 <id=PLUS a.rs:2:20 <id=INT_LITERAL, text=2, typehint=pure_decimal a.rs:2:22 <id=SEMICOLON a.rs:2:23 <id=RETURN_TOK a.rs:3:5 <id=SEMICOLON a.rs:3:11 <id=RIGHT_CURLY a.rs:4:1
```

gcc/rust/lex/rust-lex.h
```diff
@@ -139,7 +139,8 @@ class Lexer
 
 public:
   // Construct lexer with input file and filename provided
-  Lexer (const char *filename, RAIIFile input, Linemap *linemap);
+  Lexer (const char *filename, RAIIFile input, Linemap *linemap,
+	 bool dump_lex_enabled);
```
CohenArthur, line 143:
```suggestion
	 bool dump_lex_enabled = false);
```

This allows lexer users who don't care about the dumping to not pass an extra parameter (macros, modules, etc)

gcc/rust/lex/rust-lex.h
```diff
@@ -197,6 +198,9 @@ class Lexer
    * allocating new linemap */
   static const int max_column_hint = 80;
 
+  bool dump_lex_enabled;
+  std::ofstream dump_lex_out;
```
CohenArthur, line 202: This is good. I think the "best" way to go about this would be to keep a reference here, so 
```suggestion
  std::ofstream &dump_lex_out;
```

and to create the `ofstream` in the session manager.

This would allow all "dump operations" to stem from the same file and all "dump filenames" to reside in the same place

gcc/rust/lex/rust-lex.cc
```diff
@@ -24,6 +24,9 @@
 #include "safe-ctype.h"
 
 namespace Rust {
+
+const char *kLexDumpFile = "gccrs.lex.dump";
```
CohenArthur, line 28: And would allow to move this to the session manager

gcc/rust/lex/rust-lex.cc
```diff
@@ -186,6 +202,50 @@ Lexer::skip_input ()
...
+	      dump_lex_out << tok->token_id_to_str ();
+	      dump_lex_out << (has_text
+				 ? (std::string (", text=") + tok->get_str ()
+				    + std::string (", typehint=")
+				    + std::string (tok->get_type_hint_str ()))
+				 : "")
+			   << " ";
+	      dump_lex_out << get_line_map ()->to_string (loc) << " ";
+	    }
+
```
CohenArthur, line 239: This should be in a separate function, so that we could have something like this:

```cpp
if (dump_lex_enabled)
    dump_until_n();
   
token_queue.skip(n);
```

gcc/rust/lex/rust-lex.h
CohenArthur: What we could also do here is keep an `Optional<&std::ofstream>` and remove the `dump_lex_enabled` boolean. This way, we simply maybe keep a stream and dump to it when skipping tokens. This would also move the error handling (if the stream fails to open etc) to the session manager, where similar things already happen

gcc/rust/lex/rust-lex.cc
```diff
@@ -186,6 +202,50 @@ Lexer::skip_input ()
   skip_input (0);
 }
 
+void
+Lexer::skip_token (int n)
+{
+  // dump tokens if dump-lex option is enabled
+  if (dump_lex_enabled && !dump_lex_out.fail ())
```
CohenArthur, line 209: `dump_lex_out.fail()` should not be checked here imo. If it fails to open, it's a compiler error and we should trigger an assertion in the constructor

gcc/rust/lex/rust-lex.cc
```diff
@@ -118,19 +121,30 @@ is_non_decimal_int_literal_separator (char character)
...
   // inform line_table that file is being entered and is in line 1
   if (linemap)
     line_map->start_file (filename, current_line);
+  if (dump_lex_enabled)
+    {
+      dump_lex_out.open (kLexDumpFile);
+      if (dump_lex_out.fail ())
+	{
+	  rust_error_at (Linemap::unknown_location (),
+			 "cannot open %s:%m; ignored", kLexDumpFile);
```
CohenArthur, line 145: Do we really want to ignore it?

CohenArthur (commented):
I'm not sure how I feel about this. I like the idea of having a lexer dump, and you've done some good work :) I like the output format too and I think it's informative.

However, I don't like that the lexer has to take care of opening the dump file, and that the dump file name is defined in `rust-lex.cc` instead of alongside the others in the session-manager. It also feels very "intrusive" at the moment - the dump is being done as parsing is done, and this produces a lot of hard to follow code logic. What other dumps do is that they simply analyze/dump whatever it is they look at *after* that pass has been done in the compiler. Now, I understand that the lexer is different in that aspect: since the queue is buffered, and permanently replaced, we can't simply have the following:

```c++
Lexer lex(...);
Parser<Lexer> parser(lex);

auto crate = parser.parse_crate();

dump_lex(lex);
```

since at this point the lexer will have gone through its entire token queue. So the lexer dump *will* be more intrusive than other dumps, but at the moment it feels *too* intrusive

tamaroning:
@CohenArthur Thank you for your speedy review!
I am a super C++ newbie, so your suggestions help me a lot :)
I will fix them in the next 7 days.

CohenArthur:
@tamaroning thank you :) Feel free to ask any questions, either here or on our Zulip. We'll be happy to help!

gcc/rust/lex/rust-lex.cc
tamaroning: I have no idea about that :(

gcc/rust/lex/rust-lex.cc
tamaroning: fixed

gcc/rust/lex/rust-lex.cc
tamaroning: fixed

gcc/rust/lex/rust-lex.cc
tamaroning: fixed

gcc/rust/lex/rust-lex.h
tamaroning: fixed

gcc/rust/lex/rust-lex.h
tamaroning: fixed

tamaroning:
Fixed :)

gcc/rust/lex/rust-lex.cc
```diff
@@ -186,6 +188,56 @@ Lexer::skip_input ()
...
+      dump_until_n (n);
+    }
+  else
+    {
+      token_queue.skip (n);
+    }
+}
+
+void
+Lexer::dump_until_n (int n)
```
CohenArthur, line 206: Ah, so the name I suggested does not make sense anymore since you "either dump or simply skip" in `skip_token (int n)`. So I think this function should be renamed. Something like `void Lexer::dump_and_skip (int n)`

gcc/rust/rust-session-manager.cc
```diff
@@ -439,20 +439,37 @@ Session::compile_crate (const char *filename)
...
 
   // generate crate from parser
   std::unique_ptr<AST::Crate> ast_crate = parser.parse_crate ();
 
+  if (dump_lex_opt.is_some ())
+    {
+      std::ofstream &out = dump_lex_opt.get ();
+      out.close ();
+    }
+
```
CohenArthur, line 468: Is this necessary? The `Optional` should simply go out of scope and destroy its member, in that case the `ofstream` whose destructor takes care of closing the stream?

gcc/rust/util/rust-optional.h
```diff
@@ -194,6 +194,7 @@ template <typename T> class Optional<T &>
 public:
   Optional (const Optional &other) = default;
   Optional (Optional &&other) = default;
+  Optional &operator= (Optional &&other) = default;
```
CohenArthur, line 197: Ah, nice catch

CohenArthur (commented):
Thanks again for the work :) I think this is in a nice state and we can merge it soon

gcc/rust/rust-session-manager.cc
tamaroning: Oh you are correct. Destructor of ofstream is automatically called and closes the stream.

tamaroning:
Thank you for your quick review @CohenArthur . Fixed.

CohenArthur (approved):
Great stuff, thanks a lot @tamaroning !

Could you please rebase all commits into a single commit? Then I'll merge your PR. I really appreciate the work you put into this :)

tamaroning:
@CohenArthur Rebased :)

CohenArthur (approved):
Thank you! And sorry for the delay :)

bors r+

PR #1654, opened 2022-11-17
bdc3a7317139: Fix invalid ltype_length computation

### LINKED ISSUE #1646: Replace invalid `ltype_length` computation
```quote
> +  unsigned HOST_WIDE_INT ltype_length
> +    = wi::ext (wi::to_offset (TYPE_MAX_VALUE (ltype_domain))
> +                - wi::to_offset (TYPE_MIN_VALUE (ltype_domain)) + 1,

TYPE_MIN_VALUE is not checked to be constant, also the correct
check would be to use TREE_CODE (..) == INTEGER_CST, in
the GCC middle-end an expression '1 + 2' (a PLUS_EXPR) would
be TREE_CONSTANT but wi::to_offset would ICE.

> +              TYPE_PRECISION (TREE_TYPE (ltype_domain)),
> +              TYPE_SIGN (TREE_TYPE (ltype_domain)))
> +       .to_uhwi ();

.to_uhwi will just truncate if the value doesn't fit, the same result as
above is achieved with

 unsigned HOST_WIDE_INT ltype_length
    = TREE_INT_CST_LOW (TYPE_MAX_VALUE (..))
      - TREE_INT_CST_LOW (TYPE_MIN_VALUE (...)) + 1;
```

In `HIRCompileBase::verify_array_capacities()`

CohenArthur:
Fixes #1646


CohenArthur:
This one is causing issues due to the value *needing* to be truncated. I think we'll keep the existing implementation for the time being as I'm not comfortable with the `tree` or `wi` APIs and wouldn't know about fixing that.

PR #1655, opened 2022-11-17
gcc-patch-dev: Add CI to patch branch

CohenArthur:
The `gcc-patch-dev` branch currently has no CI, which is problematic. I'll move this commit to the bottom of all the "patch" commits to make sure it does not get included in the patches.

This should ideally be merged before all the patch fixes PRs


dkm:
>  I'll move this commit to the bottom of all the "patch" commits to make sure it does not get included in the patches.

Make sure the commit message is invalid wrt to GCC's commit format using ` ./contrib/gcc-changelog/git_check_commit.py` and you'll never be able to merge it by mistake :). As the body is empty, it should be the case ;)

PR #1657, opened 2022-11-17
ci: Remove warning checks on gcc-patch-dev

CohenArthur:
This is a temporary solution since the gcc-patch-dev branch will eventually disappear in favor of another upstreaming strategy


CohenArthur:
This should be merged before all other `upstream` PRs and they should then be rebased. This will ensure at least a minimal level of CI. We'll still have to do some manual builds before sending the v4

Issue #1626, opened 2022-10-31
bootstrap build failure


philberty:
```
[...]

PR #1670, opened 2022-12-01
'rust-warn += -Wno-unused-parameter'

tschwinge:
... addressing #1626, <https://inbox.sourceware.org/gcc-patches/CAFiYyc1NQT+bBZXALTs7Qhc-ECbKKBt1tVKip_6C241y3ozT6w@mail.gmail.com>.

### DIFF
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -24,6 +24,9 @@
 
 # Use strict warnings for this front end.
 rust-warn = $(STRICT_WARN)
+# ..., with the exception of '-Wunused-parameter'; waiting for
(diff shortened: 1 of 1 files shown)


CohenArthur (approved):
Thank you Thomas! I'll open an issue to remove it once we merge the cleaned up trunk branch. Or we can keep that warning disabled if it annoys people too much :)

philberty:
I use no-unused-parameters only because when we have prototypes like the abstract visitors etc it seems to complain but ideally we shouldn't rely on turning it off because we miss a lot of real good warnings and we should fix those and then remove this when ready.

PR #1688, opened 2022-12-07
ci: Lock Ubuntu version to 20.04

CohenArthur:
This locks all of our CI images to Ubuntu-20.04.

This should be upgraded eventually to 22.04, in time. I'll open up an issue to accompany it.

All of yesterday's PRs will need to be rebased on master once this is merged

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -168,7 +168,7 @@ jobs:
 
   build-and-check-gcc-48:
 
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-18.04
     steps:
     - uses: actions/checkout@v2
 
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -18,7 +18,7 @@ jobs:
       # Force locale, in particular for reproducible results re '.github/bors_log_expected_warnings' (see below).
       LC_ALL: C.UTF-8
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/ccpp.yml
```diff
@@ -168,7 +168,7 @@ jobs:
 
   build-and-check-gcc-48:
 
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
```
tschwinge: That un-does your earlier commit 19d5f921d0ce536e0e9419b3cfc5e5c2fcef9e33 "ci: gcc-4.8 package is only available in Ubuntu 18.04".  Is that intentional?

tschwinge:
> All of yesterday's PRs will need to be rebased on master once this is merged

Do they?  Per my understanding, *bors* does its own merge with the master branch, and tests the result of that, so the PRs will get the updated `.github/bors_log_expected_warnings` that way?

.github/workflows/ccpp.yml
CohenArthur: The diff looks backward. I did try to lock the gcc4.8 job to 20.04 as well, but the package is not available there. This caused the CI for [f7172bc](https://github.com/Rust-GCC/gccrs/pull/1688/commits/f7172bc2892c552484d1ee67fd5eb9f5414679c5) to fail https://github.com/Rust-GCC/gccrs/actions/runs/3637494717/jobs/6138569872

so I reverted it to 18.04. I think it can be squashed/fixed-up in the earlier commit as it doesn't make a lot of sense on its own :) I'll do that once the CI passes

dkm (approved):

PR #1693, opened 2022-12-09
testsuite: Add xfail clauses for big-endian-related type errors

CohenArthur:
This runs properly on one of the BE Sparc machines of the GCC compile farm. Now, it needs to pass the little-endian testsuite :) these dejagnu directives are a bit annoying to weld.

This also needs issue/bugzilla numbers around the `dg-xfail-if`


philberty:
I think we should do this when we get the dg directives right is a good change

CohenArthur:
Should probably be closed, since #2156 has been merged but #2184 is still there. I'll investigate

CohenArthur:
should be closed now that we've fixed the handling for these scopes

PR #1694, opened 2022-12-10
Enable Rust changelog on gcc-patch-dev

CohenArthur:
@tschwinge my empty Changelog commit should probably be squashed with your patch and I'll rebase all of gcc-patch-dev on top of this commit


CohenArthur:
Welp no need for this anymore :D We'll just update with upstream's master. Thank you @tschwinge !

tschwinge:
Sorry for the duplication of work; I had not read until the end of your private message...

(Also, `gcc/rust/Changelog` shouldn't be completely empty, but should contain the page break marker and copyright notice.)  ;-P

PR #1708, opened 2022-12-19
Parse declarative macro (decl_macro 2.0)

tamaroning:
This PR is currentry wip.
Opinions and advices are welcome!

### DIFF
--- a/gcc/rust/ast/rust-macro.h
+++ b/gcc/rust/ast/rust-macro.h
@@ -23,13 +23,11 @@
 #include "rust-ast.h"
 #include "rust-ast-fragment.h"
 #include "rust-location.h"
+#include "rust-item.h"
+#include "rust-make-unique.h"
 
 namespace Rust {
 namespace AST {
-
-// Decls as definitions moved to rust-ast.h
-class MacroItem;
-
 class MacroFragSpec
 {
 public:
@@ -446,8 +444,18 @@ public:
 };
 
 // A macro rules definition item AST node
-class MacroRulesDefinition : public MacroItem
+class MacroRulesDefinition : public VisItem
 {
+public:
+  enum MacroKind
+  {
+    // Macro by Example (legacy macro rules)
+    MBE,
+    // Declarative macros 2.0
+    DeclMacro,
+  };
+
+private:
   std::vector<Attribute> outer_attrs;
   Identifier rule_name;
   // MacroRulesDef rules_def;
@@ -460,6 +468,7 @@ class MacroRulesDefinition : public MacroItem
   std::function<Fragment (Location, MacroInvocData &)> associated_transcriber;
   // Since we can't compare std::functions, we need to use an extra boolean
   bool is_builtin_rule;
+  MacroKind kind;
 
   /**
    * Default function to use as an associated transcriber. This function should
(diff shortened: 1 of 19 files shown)


gcc/rust/ast/rust-macro.h
```diff
@@ -446,8 +443,18 @@ struct MacroRule
 };
 
 // A macro rules definition item AST node
-class MacroRulesDefinition : public MacroItem
+class MacroRulesDefinition : public VisItem
 {
+public:
+  enum MacroKind
+  {
+    // Macro by Example (leagacy macro rules)
```
CohenArthur, line 451:
```suggestion
    // Macro by Example (legacy macro rules)
```

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -1505,14 +1467,18 @@ Parser<ManagedTokenSource>::parse_macro_rules_def (AST::AttrVec outer_attrs)
 	      return std::unique_ptr<AST::MacroRulesDefinition> (
 		new AST::MacroRulesDefinition (
 		  std::move (rule_name), delim_type, std::move (macro_rules),
-		  std::move (outer_attrs), macro_locus));
+		  std::move (outer_attrs), macro_locus,
+		  AST::MacroRulesDefinition::MacroKind::MBE,
+		  AST::Visibility::create_error ()));
```
CohenArthur, line 1472: So anytime we create an `MBE` macro, we'll have no visibility, while we should always have one when creating a `DeclMacro`. I think this means we should change the API a little bit. Instead of having one single constructor used in both situations, you can create two static functions: `AST::MacroRulesDefinition::mbe()` and `AST::MacroRulesDefinition::decl_macro()`, with the `decl_macro` taking an extra `Visibility` argument.

Both these static functions should call into the existing constructor you've modified, which can now be set to private. It's a very tiny nitpick so no worries :)

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -1532,6 +1498,163 @@ Parser<ManagedTokenSource>::parse_macro_rules_def (AST::AttrVec outer_attrs)
...
+	}
+
+      AST::MacroRule macro_rule
+	= AST::MacroRule (std::move (matcher), std::move (transcriber), locus);
+      std::vector<AST::MacroRule> macro_rules;
+      macro_rules.push_back (macro_rule);
+
+      return std::unique_ptr<AST::MacroRulesDefinition> (
+	new AST::MacroRulesDefinition (
+	  std::move (rule_name), AST::DelimType::CURLY, std::move (macro_rules),
```
CohenArthur, line 1571:
```suggestion
	  std::move (rule_name), AST::DelimType::CURLY, {macro_rule},
```

This should allow you to remove the temporary `macro_rules` move, but it's more of a style thing. Feel free to ignore

CohenArthur (commented):
This looks great @tamaroning, thanks a lot :) The existing tests are really good, but I think it's missing some invalid tests - files where the decl macro is invalid. You can look at other test-cases for the dejagnu directives (`dg-error`) we usually use.

tamaroning:
Thank you for your review @CohenArthur ! I fixed code.

tamaroning:
Oops, I mistakenly added incorrect tests. All tests should pass now.

gcc/rust/ast/rust-macro.h
```diff
@@ -479,27 +487,52 @@ class MacroRulesDefinition : public MacroItem
...
       delim_type (delim_type), rules (std::vector<MacroRule> ()),
       locus (Location ()), associated_transcriber (associated_transcriber),
-      is_builtin_rule (true)
+      is_builtin_rule (true), kind (kind)
   {}
 
+public:
+  std::string as_string () const override;
+
+  static MacroRulesDefinition *mbe (Identifier rule_name, DelimType delim_type,
```
CohenArthur, line 515: We shouldn't return raw pointers like this IMO. Either return `std::unique_ptr<MacroRulesDefinition>`, or have two constructors for `MacroRulesDefinition` and not two static functions which return `MacroRulesDefinition`. We want to enforce memory safety wherever possible and returning raw pointers directly could lead to misuse

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -22,6 +22,7 @@
 /* DO NOT INCLUDE ANYWHERE - this is automatically included with rust-parse.h
  * This is also the reason why there are no include guards. */
 
+#include "rust-macro.h"
```
CohenArthur, line 25: This shouldn't be needed. Do you get errors when removing it?

CohenArthur, line 25: This shouldn't be needed. Do you get errors when removing it?

CohenArthur (commented):
Looks nice :DD This is in a good state! I think once you change the API for the two constructors/static function we'll merge this :) Thank you for your patience!

tamaroning:
I followed your suggestion and use `declarative macro` instead of `decl macro` in error message
(because gcc warns about that) .

CohenArthur:
Oh also @tamaroning, here is the failure on the gcc4.8 job:

```
 In file included from ../../gcc/rust/ast/rust-ast-full.h:29:0,
                 from ../../gcc/rust/resolve/rust-ast-resolve-item.h:24,
                 from ../../gcc/rust/rust-lang.cc:39:
../../gcc/rust/ast/rust-macro.h: In static member function ‘static std::unique_ptr<Rust::AST::MacroRulesDefinition> Rust::AST::MacroRulesDefinition::mbe(Rust::Identifier, Rust::AST::DelimType, std::vector<Rust::AST::MacroRule>, std::vector<Rust::AST::Attribute>, Location)’:
../../gcc/rust/ast/rust-macro.h:519:12: error: ‘make_unique’ is not a member of ‘std’
     return std::make_unique<MacroRulesDefinition> (
            ^
../../gcc/rust/ast/rust-macro.h:519:49: error: expected primary-expression before ‘>’ token
     return std::make_unique<MacroRulesDefinition> (
                                                 ^
../../gcc/rust/ast/rust-macro.h: In static member function ‘static std::unique_ptr<Rust::AST::MacroRulesDefinition> Rust::AST::MacroRulesDefinition::decl_macro(Rust::Identifier, std::vector<Rust::AST::MacroRule>, std::vector<Rust::AST::Attribute>, Location, Rust::AST::Visibility)’:
../../gcc/rust/ast/rust-macro.h:530:12: error: ‘make_unique’ is not a member of ‘std’
     return std::make_unique<MacroRulesDefinition> (MacroRulesDefinition (
            ^
../../gcc/rust/ast/rust-macro.h:530:49: error: expected primary-expression before ‘>’ token
     return std::make_unique<MacroRulesDefinition> (MacroRulesDefinition (
                                     
```

Since we are in C++11, we technically do not have access to `std::make_unique`. You can use the `make_unique` function defined in `gcc/rust/util/rust-make-unique.h`, but not the `std::` one.

I suspect this is the same issue on the MacOS job but maybe not, and it's harder to read. If that doesn't fix it I'll have a closer look.

tamaroning:
> Since we are in C++11, we technically do not have access to std::make_unique. You can use the make_unique function defined in gcc/rust/util/rust-make-unique.h, but not the std:: one.

Thanks for noticing :)
I fixed code and pushed branch.

CohenArthur:
Great :D thank you!

bors r+

PR #1709, opened 2022-12-19
Check const functions for mutable references

### LINKED ISSUE #1552: Mutable references are forbidden in const functions
I tried this code:

```rust
const fn foo(a: &mut i32) { // { dg-error "mutable references are not allowed in const functions" }
    *a = 15;
}
```

```rust
error[E0658]: mutable references are not allowed in constant functions
 --> test2.rs:1:14
  |
1 | const fn foo(a: &mut i32) {
  |              ^
  |
  = note: see issue #57349 <https://github.com/rust-lang/rust/issues/57349> for more information

```

Instead, this happened: compiled without issue, even when using the value.

### Meta

- What version of Rust GCC were you using: 8ff035ddc55

## Task List

- [ ] Adding two types of context in `ConstChecker`: functions/methods and const values (and const blocks later?)
- [ ] Check when a type is being visited: If we're on a mutable reference and the context is a function/method, error out

Testcase available on https://github.com/CohenArthur/gccrs/tree/forbid-mut-refs-in-const-fns

dme2:
Fixes #1552

### DIFF
--- a/gcc/rust/checks/errors/rust-const-checker.cc
+++ b/gcc/rust/checks/errors/rust-const-checker.cc
@@ -898,8 +898,12 @@ ConstChecker::visit (RawPointerType &)
 {}
 
 void
-ConstChecker::visit (ReferenceType &)
-{}
+ConstChecker::visit (ReferenceType &type)
+{
+  if (const_context.is_in_context () && type.is_mut ())
+    rust_error_at (type.get_locus (),
+		   "mutable references are not allowed in constant functions");
+}
 
 void
 ConstChecker::visit (ArrayType &type)
--- a/gcc/testsuite/rust/compile/const10.rs
+++ b/gcc/testsuite/rust/compile/const10.rs
@@ -1,0 +1,3 @@
+const fn foo (a: &mut i32) { // { dg-error "mutable references are not allowed in constant functions" }
+	*a = 1;
+}


dkm:
Hello Dave,
Thanks for the new PR. I was checking if we had already some sign of us having a FSF copyright assignment or if you've correctly used Signed-Of to indicate you are using the DCO, but couldn't really find anything.

Can you take care of this for future contribution ? We can help if you need any help or have any question on that matter :)

dme2:
Hi Marc, can I sign off here in the comment section? For future contributions I'll just use git to sign off.

dkm:
You can add an explicit mention in the PR that was already merged, stating that you've read the GCC's DCO and you are using it. Then simply add the correct Sign-off-by to future commit... Should be enough!

We will need to amend your previous commit before merging it in GCC, so having an explicit ACK is probably best!

dme2:
Just signed off on the other PR (https://github.com/Rust-GCC/gccrs/pull/1636#issue-1444430441)

dkm:
Thanks @dme2!

gcc/rust/checks/errors/rust-const-checker.h
```diff
@@ -71,6 +71,9 @@ class ConstChecker : public HIRFullVisitor
   void check_default_const_generics (
     std::vector<std::unique_ptr<GenericParam>> &param, ConstGenericCtx context);
 
+  /* Are we currently in a const context? */
+  bool in_const_context = false;
```
CohenArthur, line 75: This is not necessary and also bug prone. You can read the whole discussion [here](https://github.com/Rust-GCC/gccrs/pull/1410#discussion_r930960483) which is what led us to come up with the `StackedContext` class, which you can find right under your boolean :) You can check whether or not you are in a constant context (like a constant function) by calling `const_context.is_in_context ()`, which will return a boolean.

The class is defined [here](https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/util/rust-stacked-contexts.h), in `gcc/rust/util/rust-stacked-context.h`

gcc/rust/checks/errors/rust-const-checker.cc
```diff
@@ -898,8 +901,12 @@ ConstChecker::visit (RawPointerType &)
 {}
 
 void
-ConstChecker::visit (ReferenceType &)
-{}
+ConstChecker::visit (ReferenceType &type)
+{
+  if (this->in_const_context && type.is_mut ())
+    rust_error_at (type.get_locus (),
+		   "Mutable references are not allowed in constant functions");
```
CohenArthur, line 908:
```suggestion
		   "mutable references are not allowed in constant functions");
```

We usually go for lowercase error strings as it looks better when formatted by GCC

```shell
> gccrs
gccrs: fatal error: no input files
compilation terminated.
```

gcc/rust/checks/errors/rust-const-checker.cc
```diff
@@ -898,8 +901,12 @@ ConstChecker::visit (RawPointerType &)
 {}
 
 void
-ConstChecker::visit (ReferenceType &)
-{}
+ConstChecker::visit (ReferenceType &type)
+{
+  if (this->in_const_context && type.is_mut ())
```
CohenArthur, line 906:
```suggestion
  if (const_context.is_in_context () && type.is_mut ())
```

gcc/rust/checks/errors/rust-const-checker.cc
```diff
@@ -617,7 +617,10 @@ ConstChecker::visit (Function &function)
 {
   auto const_fn = function.get_qualifiers ().is_const ();
   if (const_fn)
-    const_context.enter (function.get_mappings ().get_hirid ());
+    {
+      const_context.enter (function.get_mappings ().get_hirid ());
+      this->in_const_context = true;
```
CohenArthur, line 622: So yeah, no need for that line :) also, you were never setting it back to false, so anytime we'd see one constant function we'd be in a constant context for the rest of the crate we're checking

CohenArthur (commented):
Thank you for contributing and taking the time to sign off @dme2! I'm sorry to be so annoying on which API to use :) let me know if you think the `StackedContext` API deservers improvements or more documentation overall, as I understand it might be quite unclear at first. Thanks again!

dme2:
Thank you for reviewing that, @CohenArthur .The StackedContext API is pretty much exactly what I was looking for! I went ahead and made those recommended changes.

CohenArthur (approved):
Thank you :) Looks great!

bors r+

CohenArthur:
bors cancel

Oops one more thing. @dme2 could you rebase all those commits in a single commit?

dme2:
I think I did that rebase correctly. If more changes are necessary just let me know

CohenArthur:
Thanks @dme2 that's great :)

bors r+

Issue #1475, opened 2022-08-16
Handle `#[target_feature]` attribute


CohenArthur:
This attribute is similar to GCC's `__attribute__((target(<feature>))` and should be treated as such. This also relates to #1411

PR #1711, opened 2022-12-31
unsafe: check use of `target_feature` attribute

snprajwal:
The `target_feature` attribute is for conditional compilation and may or may not compile on all platforms. Using it requires an unsafe function or block.

Addresses #1411

### DIFF
--- a/gcc/rust/util/rust-attributes.cc
+++ b/gcc/rust/util/rust-attributes.cc
@@ -41,6 +41,9 @@ static const BuiltinAttrDefinition __definitions[]
      {"repr", CODE_GENERATION},
      {"path", EXPANSION},
      {"macro_use", NAME_RESOLUTION},
+     // FIXME: This is not implemented yet, see
+     // https://github.com/Rust-GCC/gccrs/issues/1475
+     {"target_feature", CODE_GENERATION},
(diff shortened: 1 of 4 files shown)


gcc/rust/util/rust-attributes.cc
```diff
@@ -41,6 +41,7 @@ static const BuiltinAttrDefinition __definitions[]
      {"repr", CODE_GENERATION},
      {"path", EXPANSION},
      {"macro_use", NAME_RESOLUTION},
+     {"target_feature", CODE_GENERATION},
```
CohenArthur, line 44: Can you add a `FIXME` comment here pointing to #1475 ? Since this is just a stub and doesn't actually implement the `target_feature` attribute, we need to be careful.

CohenArthur (approved):
This looks great! Thank you :D

Just waiting for you to add the comment and we'll merge this

gcc/rust/util/rust-attributes.cc
snprajwal: Done! I've also reworded  the error message to sound a little better

CohenArthur (approved):
Great stuff :) Thank you! I'll merge this soon

PR #1715, opened 2023-01-02
codegen: fix attribute handling

snprajwal:
Attributes were previously evaluated using an `if-else` ladder. This causes multiple attributes on the same item to not be processed. It has been changed to use individual `if` conditions for each attribute instead.

### DIFF
--- a/gcc/rust/backend/rust-compile-base.cc
+++ b/gcc/rust/backend/rust-compile-base.cc
@@ -79,23 +79,23 @@ HIRCompileBase::setup_fndecl (tree fndecl, bool is_main_entry_point,
 	{
 	  handle_inline_attribute_on_fndecl (fndecl, attr);
 	}
-      else if (is_must_use)
+      if (is_must_use)
 	{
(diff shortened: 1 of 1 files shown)


philberty (commented):
This block iterates each attribute one and at time so I think the else if is ok here because each attribute has its own specific handler

CohenArthur:
Ah, @philberty is right. It would be nice to refactor the AST::Attribute type a little bit to make it clearer, for example to allow this loop instead of the one with multiple `if { } else if { }`:

```c++
for (const auto &attr: attrs) {
    switch (attr.kind) {
        case Attr::Inline: handle_inline_attribute_on_fndecl(fndecl, attr);
        case Attr::MustUse: handle_must_use_attribute_on_fndecl(fndecl, attr);
        /* ... */
    }
} 
```

but that's a different thing. @snprajwal I'll close the PR in the meantime but if you're interested in working on such a refactor/improvement we can chat about it on Zulip and we can mentor you :)

snprajwal:
Sounds good, let's discuss it there 👍

PR #1716, opened 2023-01-04
rust: add bound parsing in parse_generic_arg.

TuringKi:
This patch adds parsing for generic parameters like: <F: Foo<Bar: Foo>>. In current version, this pattern  leads to endless loop.


philberty (approved):
LGTM nice find i think i hit this issue before Christmas but forgot to raise an issue

PR #1718, opened 2023-01-05
parse: fix parse for associated type bounds

TuringKi:
This commit fix the parsing error message  of associated type bounds.

the results:

in rustc :https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=30e3104c1e55baf33f1f14ffbf548847
gccrs:
```
../gcc/testsuite/rust/compile/bounds.rs:8:13: error: associated type bounds are unstable
    8 | fn c<F: Foo<Bar: Foo>>() where F::Bar: Copy { // { dg-error "associated type bounds are unstable" }
      |             ^~~~~~
../gcc/testsuite/rust/compile/bounds.rs:12:13: error: associated type bounds are unstable
   12 | fn d<F: Foo<Bar: Foo // { dg-error "associated type bounds are unstable" }
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   13 | 
      |              
   14 | + T0
      | ~~~~         
   15 | 
      |              
   16 | + T1
      | ~~~~         
   17 | + T2 +
      | ~~~~~~       
   18 | 
      |              
   19 | T3      >>() where F::Bar: Copy {
      | ~            



```


TuringKi:
anyone knowns why ci "GCC Rust build and test / build-and-check-gcc-48 (pull_request) " failed?

CohenArthur (commented):
I'm not sure this specific error should happen in the parser. If the feature is enabled, then associated type bounds are allowed (so adding `#![feature(associated_type_bounds)]` in your example works for example). We don't yet have a mechanic for stable and unstable features, or what is and isn't allowed. I think this error should be emitted when that pass is implemented :)

I would not scratch that PR though. The code is good IMO and it would be useful to get started on a `Feature` AST visitor. If you're interested in working on that, I'll gladly mentor you!

TuringKi:
the new commit adds support for  diagnostic underline, just like this:

```
../gcc/testsuite/rust/compile/bad_type1.rs:2:5: error: expected 'bool' got '<integer>'
    2 |     let logical: bool = 123; // { dg-error "expected .bool. got .<integer>." }
      |     ^~~          ~~~~   ~~~

```

CohenArthur:
> definitely yes. I think it will be a good chance to make me be more familiar with the code base :). Could you please give me some introduction? such as others' pr or commits. A path or filename of source code will also fine.

Great :D

So adding feature checking is actually going to be a big piece of work. I don't expect you to complete it all in one go or all by yourself.

The first step would be to determine where that visitor should happen. Should the visitor visit our AST? Our HIR?

I'm hoping that it can be done on the HIR, as that would be easier for you. I think this is what should be done and I assume this will be the best course of action.

If you're willing to, I would ask you to look at rustc's source code and check where those feature checks happen. If you don't know how to proceed I'll take care of that.

Then, you'll need to start by adding a basic skeleton visitor. For this, you'll find examples in these PRs:

- https://github.com/Rust-GCC/gccrs/pull/1410
- https://github.com/Rust-GCC/gccrs/pull/1419

So the associated files are `gcc/rust/checks/errors/rust-unsafe-checker.cc` and `gcc/rust/checks/errors/rust-const-checker.cc`.

In your case, I'd expect the visitor to be called something like `FeatureGating` or similar.

Then, we can look at adding the first actual check, which could the one you're doing here. You'll need to get familiar with the HIR API (and we'll show you how) so that you can figure out whether or not a generic argument has associated type bounds.
You'll also need to keep track of which features have been added in the crate. This will be done earlier in the compilation pipeline, and we can get to this later.

Don't hesitate to ping me here or on Zulip if something is not clear enough. I'll explain it in more details :)

TuringKi:
> > definitely yes. I think it will be a good chance to make me be more familiar with the code base :). Could you please give me some introduction? such as others' pr or commits. A path or filename of source code will also fine.
> 
> Great :D
> 
> So adding feature checking is actually going to be a big piece of work. I don't expect you to complete it all in one go or all by yourself.
> 
> The first step would be to determine where that visitor should happen. Should the visitor visit our AST? Our HIR?
> 
> I'm hoping that it can be done on the HIR, as that would be easier for you. I think this is what should be done and I assume this will be the best course of action.
> 
> If you're willing to, I would ask you to look at rustc's source code and check where those feature checks happen. If you don't know how to proceed I'll take care of that.
> 
> Then, you'll need to start by adding a basic skeleton visitor. For this, you'll find examples in these PRs:
> 
> * [unsafe: Add UnsafeCheck visitor #1410](https://github.com/Rust-GCC/gccrs/pull/1410)
> * [Const checker #1419](https://github.com/Rust-GCC/gccrs/pull/1419)
> 
> So the associated files are `gcc/rust/checks/errors/rust-unsafe-checker.cc` and `gcc/rust/checks/errors/rust-const-checker.cc`.
> 
> In your case, I'd expect the visitor to be called something like `FeatureGating` or similar.
> 
> Then, we can look at adding the first actual check, which could the one you're doing here. You'll need to get familiar with the HIR API (and we'll show you how) so that you can figure out whether or not a generic argument has associated type bounds. You'll also need to keep track of which features have been added in the crate. This will be done earlier in the compilation pipeline, and we can get to this later.
> 
> Don't hesitate to ping me here or on Zulip if something is not clear enough. I'll explain it in more details :)

Many thanks, that's a very nice tutorial.
I made some code travel in rustc, and found the feature_gate is implemented in expansion stage. The function call stacks like this:

```
rustc_driver lib.rs: 

run_compiler -> configure_and_expand -> configure_and_expand_inner -> feature_gate::crate_create()

```

and the gate check like this:
```rust

    gate_all!(trait_alias, "trait aliases are experimental");
    gate_all!(associated_type_bounds, "associated type bounds are unstable");
    gate_all!(crate_visibility_modifier, "`crate` visibility modifier is experimental");
    gate_all!(const_generics, "const generics are unstable");
    gate_all!(decl_macro, "`macro` is experimental");
    gate_all!(box_patterns, "box pattern syntax is experimental");
    gate_all!(exclusive_range_pattern, "exclusive range pattern syntax is experimental");
    gate_all!(try_blocks, "`try` blocks are unstable");
    gate_all!(label_break_value, "labels on blocks are unstable");
    gate_all!(box_syntax, "box expression syntax is experimental; you can call `Box::new` instead");


```
It seams like, all the gate_check happens after the macro expansions:

```rust
// Done with macro expansion!

    if sess.opts.debugging_opts.input_stats {
        println!("Post-expansion node count: {}", count_nodes(&krate));
    }

    if sess.opts.debugging_opts.hir_stats {
        hir_stats::print_ast_stats(&krate, "POST EXPANSION AST STATS");
    }

    if sess.opts.debugging_opts.ast_json {
        println!("{}", json::as_json(&krate));
    }

    resolver.resolve_crate(&krate);

    // Needs to go *after* expansion to be able to check the results of macro expansion.
    sess.time("complete_gated_feature_checking", || {
        rustc_ast_passes::feature_gate::check_crate(&krate, sess);
    });

```

Maybe I can implement the feature of "associated type bounds" check_gate in [rust-macro-expand.cc](https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/expand/rust-macro-expand.cc) over here:

```c++
...
	it++;
    }

  pop_context ();

  // TODO: should recursive attribute and macro expansion be done in the same
  // transversal? Or in separate ones like currently?

  // expand module tree recursively

  // post-process 

// <----- add feature check_gate here!

  // extract exported macros?
}


```
What do you think about the position where i should add the code?  :)

CohenArthur:
> It seams like, all the gate_check happens after the macro expansions:

Ah, nice find :) In that case, you should create an AST visitor. A recent example is the `EarlyNameResolver`, introduced in https://github.com/Rust-GCC/gccrs/pull/1580

So the files in question are `gcc/rust/resolve/rust-early-name-resolver.{h, cc}`.

Then, you'll need to call into your visitor in the session-manager, in `compile_crate` after the call to the `expansion` function (around line 556)

TuringKi:
> Note that some feature gate checks happen when building the compiler session (eg unstable flags) and some happen at arbitrary points after AST expansion when they gate a feature affecting semantics, rather than syntax.

Yes, but that is a huge coding works... >_<  maybe i need step by step.

bjorn3:
Indeed. It makes sense to do it step by step.

gcc/rust/feature/rust-feature.h
```diff
@@ -0,0 +1,76 @@
...
+    uint8_t none;
+    Edition (CompileOptions::Edition e) : edition (e) {}
+    Edition (uint8_t n) : none (n) {}
+  };
+  Feature (State S, const char *N, const char *RS, const char *GS, uint64_t I,
+	   Edition E, const char *D)
+    : state (S), name (N), rustc_since (RS), gccrs_since (GS), issue (I),
+      edition (E), description (D)
+  {
+    Feature::NUM++;
```
CohenArthur, line 42: Global state is not great for this. I'd like to rethink that API as I think it does too much. In your current class, a `Feature` represents a singular feature, but you can also call `Feature::load_hash_map()` to get access to all features defined within the compiler. I think this is an indication the class should be split in two:

1. A `Feature` class which should represent a singular feature. So what you have at the moment is good
2. A `FeatureCtx` or `FeatureTable` class which keeps track of which features are present and their state, and maybe keeps track of which features are being *used* in the present crate. But we can think about that later on

gcc/rust/feature/rust-feature.h
```diff
@@ -0,0 +1,76 @@
...
+  DECLARE_FEATURE (Feature::State::ACCEPTED, (N), (RS), (GS), (I), (E), (D))
+
+#define DECLARE_ACTIVE_FEATURE(N, RS, GS, I, E, D)                             \
+  DECLARE_FEATURE (Feature::State::ACTIVE, (N), (RS), (GS), (I), (E), (D))
+
+#define DECLARE_REMOVED_FEATURE(N, RS, GS, I, E, D)                            \
+  DECLARE_FEATURE (Feature::State::REMOVED, (N), (RS), (GS), (I), (E), (D))
+
+#define DECLARE_STABILIZED_FEATURE(N, RS, GS, I, E, D)                         \
+  DECLARE_FEATURE (Feature::State::STABILIZED, (N), (RS), (GS), (I), (E), (D))
```
CohenArthur, line 74: I think this should be achieved by having multiple constructors/static functions instead.

So something like:
```c++
class Feature {
public:
    static Feature accepted(...) { return Feature(...); }
    static Feature active(...) { return Feature(...); }
    /* and so on */
};
```

gcc/rust/expand/rust-macro-expand.cc
```diff
@@ -317,10 +324,61 @@ MacroExpander::expand_crate ()
   // expand module tree recursively
 
   // post-process
+  feature_gate ();
```
CohenArthur, line 327: I think this should rather be done in the `SessionManager::compile_crate` function. This is where the rest of our pipeline happens

gcc/rust/expand/rust-macro-expand.cc
```diff
@@ -317,10 +324,61 @@ MacroExpander::expand_crate ()
...
 
   // extract exported macros?
 }
 
+void
+MacroExpander::feature_gate ()
```
CohenArthur, line 333: So I would move this function to the SessionManager as well. It's good, but not really related to macro expansion. It simply happens *after* macro expansion in the pipeline.

gcc/rust/feature/rust-feature.cc
```diff
@@ -0,0 +1,50 @@
...
+#include "rust-feature.h"
+
+namespace Rust {
+
+Feature feature_table[] = {
+
+  DECLARE_ACTIVE_FEATURE ("associated_type_bounds", "1.34.0", "0.0.0", 52662, 0,
+			  ""),
+  DECLARE_ACTIVE_FEATURE ("intrinsics", "1.0.0", "0.0.0", 0, 0, ""),
+  DECLARE_ACTIVE_FEATURE ("rustc_attrs", "1.0.0", "0.0.0", 0, 0, ""),
```
CohenArthur, line 30: What is edition `0`? If I understand correctly, this is the equivalent of `Option<Edition>` used in rustc. In that case, we can use `Optional<Edition>` :) the class is defined in `gcc/rust/util/rust-optional.h` and should help you here

gcc/rust/feature/rust-feature.cc
```diff
@@ -0,0 +1,50 @@
...
+
+};
+
+uint32_t Feature::NUM = 0;
```
CohenArthur, line 34: Do we really need the number? We can simply iterate on the hashmap

gcc/rust/feature/rust-feature.cc
```diff
@@ -0,0 +1,50 @@
...
+Feature::load_hash_map ()
+{
+  if (NUM != feature_hash_map.size ())
+    {
+      for (uint32_t i = 0; i < NUM; i++)
+	{
+	  feature_hash_map.insert ({feature_table[i].name, &feature_table[i]});
+	}
+    }
+  return feature_hash_map;
```
CohenArthur, line 47: In C++ we can also construct a constexpr hashmap at compile time, or as a static in a function, so you don't need to get to all that trouble.

Example [here](https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/util/rust-hir-map.cc#L841-L857)

CohenArthur (changes requested):
This PR is getting really big. There are multiple commits which do not relate to feature gates like 24bf4e725f44a8ca8d759a18e5768c3f65a58d66 and d606dd27a7bd3b3cf2440fddaf811aee332bc86c. They need to be put in separate PRs as they make reviewing this one quite difficult.

Furthermore, even on the feature gating side, this is a little big, but it's great work :) Even though there are lots of comments, the code is good and you've put in a lot of work which I appreciate. I'll let you fix these comments and remove the two first commits and I'll rereview your PR. Feel free to ping me when that is done and I'll get to it ASAP!

gcc/rust/ast/passes/rust-feature-gate.h
```diff
@@ -0,0 +1,223 @@
+// Copyright (C) 2020-2022 Free Software Foundation, Inc.
```
philberty, line 1: I might be missing something but why are we creating a pass directory within the AST? I feel like this could  be within the lints pass's directory

philberty (changes requested):
From a scan of the commits there are 2 merge commits, leaving 3 commits of code here. 

Thanks to @u1f35c I now love lots of small commits and I would like this PR to be split up a little more, You could start by having a single commit for each part such as:

1. AST changes
2. Lexer changes
3. macro-expansion changes
4. HIR change
5. Type checking changes
6:: etc...

I think this really helps us in the long run because we have smaller commits with messages to help us remember whats going on.

TuringKi:
@CohenArthur, It's very appreciate for your review,  i will split this big pr and check your comments carefully.

@philberty, very nice pr splitting advice!

gcc/rust/feature/rust-feature.cc
```diff
@@ -0,0 +1,50 @@
+// Copyright (C) 2020-2022 Free Software Foundation, Inc.
```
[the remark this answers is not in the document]
CohenArthur: Yes, that's a good point. I think at the moment we use the `lints` folder more for warnings and the `errors` folder more for hard errors, so maybe both the `rust-feature.{h, cc}` and `rust-feature-gate.{h, cc}` headers and source files should be moved into `gcc/rust/checks/errors/` @TuringKi

PR #1719, opened 2023-01-05
[WIP] Added better let statement pattern support

powerboat9:

- [x] GroupedPattern removal
- [ ] TuplePattern support


powerboat9:
I'll split it up a bit, but I've been running ad-hoc tests and it seems to be working already (besides a lack of formal tests). I already implemented recursive type checking using TypeCheckPattern and spun off CompilePatternLet to handle code generation similar to how you describe. It would be nice to talk about it a bit more on Sunday though.

philberty:
I wonder why that test-case is failing " FAIL: rust/compile/builtin_macro_recurse.rs scan-tree-dump-times gimple ""test1canary"" 1"?

powerboat9:
It works with the following patch:

--- a/gcc/testsuite/rust/compile/builtin_macro_recurse.rs
+++ b/gcc/testsuite/rust/compile/builtin_macro_recurse.rs
@@ -15,7 +15,8 @@ macro_rules! b {
 
 fn main() {
     // { dg-final { scan-tree-dump-times {"test1canary"} 1 gimple } }
-    let _ = concat!(a!(), 1, b!());
+    let v = concat!(a!(), 1, b!());
+    let _ = v;
     // should not error
     concat!(a!(), true, b!(),);
 }

It seems like it's the change in WildcardPattern compilation.

gcc/rust/resolve/rust-ast-resolve-pattern.h
```diff
@@ -70,11 +70,9 @@ class PatternDeclaration : public ResolverBase
       pattern.get_node_id (), pattern.get_locus (), type);
   }
 
-  void visit (AST::WildcardPattern &pattern) override
+  void visit (AST::GroupedPattern &pattern) override
```
philberty, line 73: Its looks like the wildcard pattern visitor was removed maybe you just need to revert that change and just add the new grouped pattern visitor which is a nice single commit

powerboat9:
Everything appears to work now.

gcc/rust/backend/rust-compile-pattern.cc
```diff
@@ -329,5 +335,134 @@ CompilePatternBindings::visit (HIR::StructPattern &pattern)
...
+    {
+    default:
+      gcc_unreachable ();
+    case HIR::TuplePatternItems::TuplePatternItemType::RANGED:
+      rust_fatal_error (pattern.get_locus (),
+			"ranged tuple pattern in let statement");
+      return;
+    case HIR::TuplePatternItems::TuplePatternItemType::MULTIPLE:
+      break;
+    }
```
CohenArthur, line 436: 1. Won't this switch always evaluate to the `default` branch?
2. Are `RANGED` patterns in let statement a programmer error or something the compiler does not support yet? If it is the former, I would switch this to a call to `rust_error`, and in the case of the latter, `rust_sorry_at`

gcc/rust/backend/rust-compile-pattern.cc
```diff
@@ -329,5 +335,134 @@ CompilePatternBindings::visit (HIR::StructPattern &pattern)
...
+    {
+      TyTy::BaseType *ty_sub = nullptr;
+      HirId pattern_id = pattern.get_pattern_mappings ().get_hirid ();
+      bool ok = ctx->get_tyctx ()->lookup_type (pattern_id, &ty_sub);
+      rust_assert (ok);
+
+      tree sub_init
+	= ctx->get_backend ()->struct_field_expression (access_expr, tuple_idx,
+							sub->get_locus ());
+      CompilePatternLet::Compile (&*sub, sub_init, ty_sub, rval_locus, ctx);
```
CohenArthur, line 462:
```suggestion
      CompilePatternLet::Compile (sub.get (), sub_init, ty_sub, rval_locus, ctx);
```
Nit, but isn't this effectively the same?

gcc/rust/backend/rust-compile-pattern.cc
CohenArthur, line 436: 1. Won't this switch statement always evaluate to the `default` case?
2. Are `RANGED` pattersn in let pattern a programmer error or something we do not support yet? If it is the former, I would switch to a call to `rust_error`, and in the case of the latter, `rust_sorry_at`

gcc/rust/hir/rust-ast-lower-pattern.cc
```diff
@@ -243,5 +243,21 @@ ASTLoweringPattern::visit (AST::RangePattern &pattern)
...
+void
+ASTLoweringPattern::visit (AST::GroupedPattern &pattern)
+{
+  auto crate_num = mappings->get_current_crate ();
+  Analysis::NodeMapping mapping (crate_num, pattern.get_node_id (),
+				 mappings->get_next_hir_id (crate_num),
+				 UNKNOWN_LOCAL_DEFID);
+
+  HIR::Pattern *inner
+    = ASTLoweringPattern::translate (&*pattern.get_pattern_in_parens ());
```
CohenArthur, line 255:
```suggestion
    = ASTLoweringPattern::translate (pattern.get_pattern_in_parens ().get());
```
nit, and this one might be wrong

gcc/rust/backend/rust-compile-var-decl.h
```diff
@@ -30,39 +30,42 @@ class CompileVarDecl : public HIRCompileBase, public HIR::HIRPatternVisitor
...
+    Bvariable *var
       = ctx->get_backend ()->local_variable (fndecl, pattern.get_identifier (),
 					     translated_type, NULL /*decl_var*/,
 					     pattern.get_locus ());
 
     HirId stmt_id = pattern.get_pattern_mappings ().get_hirid ();
-    ctx->insert_var_decl (stmt_id, compiled_variable);
+    ctx->insert_var_decl (stmt_id, var);
+
+    locals->push_back (var);
```
CohenArthur, line 53: Is there any reason other than ergonomics for keeping the `locals` within the visitor? The creation of multiple variables?

gcc/testsuite/rust/compile/let-pattern.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
+    let ((_x, _y, _, _z)) = (1, 2, 3, 4);
+}
```
CohenArthur, line 3: So this is a pattern in parentheses and a tuple pattern at once, right? I think it'd be worth it to separate them into two testcases (and maybe keep this one which contains both) so it'd be easier to pinpoint if eventually something fails.

gcc/testsuite/rust/compile/let-pattern.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
+    let ((_x, _y, _, _z)) = (1, 2, 3, 4);
```
philberty, line 2: Can you make another test case:

```rust
fn main() -> i32 {
  let a,b = (1, 2);
  let c = a + b;
  c - a - b
}
```

I want to see the -fdump-tree-gimple. This is really important to get right so we need to be careful about merging this until we are happy.

gcc/rust/backend/rust-compile-var-decl.h
```diff
@@ -30,39 +30,42 @@ class CompileVarDecl : public HIRCompileBase, public HIR::HIRPatternVisitor
   using HIR::HIRPatternVisitor::visit;
 
 public:
-  static ::Bvariable *compile (tree fndecl, tree translated_type,
```
philberty, line 33: I dont believe there is enough context in your git commit message to explain whats going on here.

gcc/rust/backend/rust-compile-stmt.cc
```diff
@@ -84,31 +76,14 @@ CompileStmt::visit (HIR::LetStmt &stmt)
...
   rust_assert (ok);
-  tree stmt_type = TyTyResolveCompile::compile (ctx, ty);
 
   Location lvalue_locus = stmt.get_pattern ()->get_locus ();
   Location rvalue_locus = stmt.get_init_expr ()->get_locus ();
   TyTy::BaseType *expected = ty;
   init = coercion_site (stmt.get_mappings ().get_hirid (), init, actual,
 			expected, lvalue_locus, rvalue_locus);
 
-  auto fnctx = ctx->peek_fn ();
```
philberty, line 95: Where did this go?

philberty (changes requested):
I believe you should raise 2 seperate PR's.

1. the name resolver change.
2. The refactor to extract out CompileLetPattern

I am a bit nervous about the other changes until we see the gimple and I would like to be sure we are happy with the type resolution changes on its own first.

I am assuming you are creating a parent tuple tyty for the grouped pattern hir id. 

The compilation of the actual pattern I would assume we would have followed some more of how we handle match expression restructuring where we don't add new BVariables but actually add in field access expressions which I didn't see.

gcc/rust/backend/rust-compile-var-decl.h
powerboat9: Yep, the creation of multiple/no variables.

gcc/rust/backend/rust-compile-pattern.cc
powerboat9: Yeah. Do I do anything different with Signed-off-by for applying this?

gcc/rust/backend/rust-compile-pattern.cc
powerboat9: 1. I don't think so
2. Not supported yet, so I'll fix that. What is a ranged tuple pattern?

gcc/rust/backend/rust-compile-stmt.cc
powerboat9: Into the CompilePatternLet visitor

gcc/rust/backend/rust-compile-pattern.cc
CohenArthur: No, it's fine. You can add a `Co-authored-by: name <email>` line if you'd like, but for reviews you don't need to :)

gcc/rust/backend/rust-compile-pattern.cc
CohenArthur: I believe ranged tuple pattern refer to `RestPattern` in the Rust Reference [1]. They are a special tuple pattern which can match a tuple of any size.

[1]: https://doc.rust-lang.org/reference/patterns.html#rest-patterns

CohenArthur:
@powerboat9 so can this PR be closed since it's been split up in multiple smaller PRs? It's fine if you'd like to keep it open, just making sure :)

powerboat9:
On second thought, it looks like the GroupedPattern support in let statements relies on CompilePatternLet.

powerboat9:
@philberty This should be ready, now that all of the GroupedPattern changes have been split off.

powerboat9:
Actually, It looks like I still have some tweaks to do.

PR #1724, opened 2023-01-08
Add support for generics associated type binding

### LINKED ISSUE #1720: Support associated type bindings in generics
I tried this code:

```rust
mod core {
    mod ops {
        #[lang = "add"]
        pub trait Add<Rhs = Self> {
            type Output;

            fn add(self, rhs: Rhs) -> Self::Output;
        }
    }
}

pub fn foo<T: core::ops::Add<Output = i32>>(a: T) {
    a + a;
}

pub fn main() {
    foo(123);
    foo(456f32); // should fail
}

```

I expected to see this happen: compile for the i32 case and fail for the f32 case

Instead, this happened: lots of errors see: https://godbolt.org/z/Y8T481shP

### Meta

- What version of Rust GCC were you using, git sha if possible. 0152926ab36ba52153f3f457f6f3bb02bb274073

philberty:
This patch set adds support to specify the associated type via the argument
binding syntax. We are still missing general bounds checking so in order to
properly add more test cases for bad types we need to implement that first.
I have also changed a test case to -fsyntax-only as at the time I always get
confused between Generic arguments of foo=i32 and bar: baz only specifies
the associated type and the latter adds a bound. Currently we are lacking a
way to represent adding a bound in the AST and HIR within generic arguments
so I have raised: #1725 and #1726

Fixes #1720

### DIFF
--- a/gcc/rust/typecheck/rust-tyty.h
+++ b/gcc/rust/typecheck/rust-tyty.h
@@ -675,16 +675,17 @@ class SubstitutionArgumentMappings
 {
 public:
   SubstitutionArgumentMappings (std::vector<SubstitutionArg> mappings,
+				std::map<std::string, BaseType *> binding_args,
 				Location locus,
 				ParamSubstCb param_subst_cb = nullptr,
 				bool trait_item_flag = false)
-    : mappings (mappings), locus (locus), param_subst_cb (param_subst_cb),
-      trait_item_flag (trait_item_flag)
+    : mappings (mappings), binding_args (binding_args), locus (locus),
+      param_subst_cb (param_subst_cb), trait_item_flag (trait_item_flag)
   {}
 
   SubstitutionArgumentMappings (const SubstitutionArgumentMappings &other)
-    : mappings (other.mappings), locus (other.locus),
-      param_subst_cb (other.param_subst_cb),
+    : mappings (other.mappings), binding_args (other.binding_args),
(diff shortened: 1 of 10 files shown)


gcc/rust/typecheck/rust-tyty.h
```diff
@@ -813,6 +826,19 @@ class SubstitutionRef
...
+  {
+    return get_num_associated_bindings () > 0;
+  }
+
+  virtual size_t get_num_associated_bindings () const { return 0; }
+
+  virtual TypeBoundPredicateItem
+  lookup_associated_type (const std::string &search)
+  {
+    return TypeBoundPredicateItem::error ();
```
CohenArthur, line 839: Does this need a `FIXME`/`TODO`?

CohenArthur (approved):
Minor question but LGTM!

CohenArthur:
> Minor question but LGTM!

Train wifi sent my review twice. Sorry!

gcc/rust/typecheck/rust-tyty.h
CohenArthur, line 839: Does this need a `FIXME`/`TODO` comment?

CohenArthur (approved):

gcc/rust/typecheck/rust-tyty.h
CohenArthur: That sounds good to me then :)

PR #1727, opened 2023-01-09
Implemented utf-8 checking

### LINKED ISSUE #1065: Enforce file contents are valid UTF-8 in include_str! macro
Strings and strs are always valid UTF-8 in Rust. We should enforce that in the `include_str!` builtin macro.

https://github.com/Rust-GCC/gccrs/blob/0fa882160df40cee56b5cdd0a2953b4abb4b9d18/gcc/rust/expand/rust-macro-builtins.cc#L232-L233

powerboat9:

Fixes #1065


philberty:
I think it would be good for @dafaust to look at this for you

dkm:
Beware that SoB lines must use your real name (1 commit is OK, the other one and the text in the PR are not) :)

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -241,6 +241,7 @@ load_file_bytes (const char *filename)
   RAIIFile file_wrap (filename);
   if (file_wrap.get_raw () == nullptr)
     {
+      // TODO: is Location () ok?
```
CohenArthur, line 244: In that case, yes. We can't have locations, which point inside a file, if we can't open the file.
```suggestion
```

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -254,6 +255,7 @@ load_file_bytes (const char *filename)
 
   if (fread (&buf[0], fsize, 1, f) != 1)
     {
+      // TODO: is Location () ok?
```
CohenArthur, line 258: Similarly
```suggestion
```

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -358,8 +360,47 @@ MacroBuiltin::include_str_handler (Location invoc_locus,
...
+  int expect_single = 0;
+  for (uint8_t b : bytes)
+    {
+      if (expect_single)
+	{
+	  if ((b & 0xC0) != 0x80)
+	    {
+	      /* character was truncated, exit with expect_single != 0 */
+	      break;
+	    }
```
CohenArthur, line 373: GCC styling convention wants us to remove these brackets
```suggestion
	      /* character was truncated, exit with expect_single != 0 */
	      break;
```

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -358,8 +360,47 @@ MacroBuiltin::include_str_handler (Location invoc_locus,
...
+  std::string str;
+  if (expect_single)
+    {
+      rust_error_at (invoc_locus, "%s wasn't a utf-8 file",
+		     target_filename.c_str ());
+    }
+  else
+    {
+      str = std::string ((const char *) &bytes[0], bytes.size ());
+    }
```
CohenArthur, line 403:
```suggestion
  if (expect_single)
      rust_error_at (invoc_locus, "%s wasn't a utf-8 file",
		     target_filename.c_str ());
  else
      str = std::string ((const char *) &bytes[0], bytes.size ());
```
Same tiny styling nit

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -358,8 +360,47 @@ MacroBuiltin::include_str_handler (Location invoc_locus,
...
+	      expect_single = 1;
+	      break;
+	    }
+	}
+    }
+
+  std::string str;
+  if (expect_single)
+    {
+      rust_error_at (invoc_locus, "%s wasn't a utf-8 file",
```
CohenArthur, line 397:
```suggestion
      rust_error_at (invoc_locus, "%s was not a valid utf-8 file",
```

CohenArthur (commented):
I don't know about any of the actual checks so I'll leave someone with more experience in that regard to review the PR. Other than the tiny nitpicks I have this looks okay to me.

CohenArthur:
@powerboat9 are there any testcases you might add to this PR btw?

philberty:
@powerboat9 is this still WIP

powerboat9:
With tests, this would no longer be WIP

powerboat9:
Let me know if I need to add more tests

gcc/testsuite/rust/compile/invalid_utf8
```diff
@@ -0,0 +1 @@
+�
```
CohenArthur, line 1: I'm really not well versed on UTF-8. Can you explain why this is invalid?

CohenArthur (commented):
I think that looks good, but I'm not able to fully check the algorithm. I'd like to see a test with a regular string (not necessarily from `include_str!`) and an identifier, like a variable name. Then I'm okay with merging it :)

philberty (changes requested):
I will aprove this as it does not regress but you should add a FIXME here that we should be reusing the lexer here instead of hand-coding something here when I compare it to the implementation in gccgo.

   gcc/go/gofrontend/lex.cc:Lex::fetch_char

This is the implementation we will likely need to port over would make this simpler.

gcc/testsuite/rust/compile/invalid_utf8
bjorn3: It is an 0xff byte followed by an 0x0a byte. 0xff is not a valid part of any UTF-8 encoded character. The highest allowed byte is 0xf7 which would theb have to be followed by 4 continuation bytes. Note that your browser may render this sequence as ÿ. For example for me it does when showing the files, but not when showing the discussion. This is caused by the browser autodetecting a different character encoding than UTF-8.

CohenArthur: Ah, thanks @bjorn3 :) yeah I was confused because it does render as `ÿ` in the Files, but I hadn't checked back here. Thanks!

CohenArthur:
@powerboat9 this is still missing the comment about reusing our lexer before we can merge it

powerboat9:
Added ```FIXME``` and removed usage of ```__builtin_clz```.

philberty (approved):
LGTM

PR #1733, opened 2023-01-11
diagnostics: Add underline for tokens in diagnostics.

TuringKi:
Currently, the diagnostics only point to the corresponding token's start position by carets, and lack of underlines for full token.  This commit add support for such underlines in diagnostics by encoding range information in location_t.

the results like this:

```
../gcc/testsuite/rust/compile/bad_as_bool_char.rs:5:14: error: invalid cast 'bool' to 'f32' [E0054]
    5 |   let fone = t as f32;   // { dg-error "invalid cast" }
      |              ^    ~~~
../gcc/testsuite/rust/compile/bad_as_bool_char.rs:6:15: error: invalid cast 'bool' to 'f64' [E0054]
    6 |   let fzero = f as f64;  // { dg-error "invalid cast" }
      |               ^    ~~~
../gcc/testsuite/rust/compile/bad_as_bool_char.rs:8:12: error: invalid cast 'u8' to 'bool' [E0054]
    8 |   let nb = 0u8 as bool;  // { dg-error "invalid cast .u8. to .bool. \\\[E0054\\\]" }
      |            ^~~    ~~~~
../gcc/testsuite/rust/compile/bad_as_bool_char.rs:9:12: error: invalid cast 'bool' to 'char' [E0054]
    9 |   let nc = true as char; // { dg-error "invalid cast" }
      |            ^~~~    ~~~~
../gcc/testsuite/rust/compile/bad_as_bool_char.rs:13:12: error: invalid cast 'char' to 'f32' [E0054]
   13 |   let fa = a as f32;     // { dg-error "invalid cast" }
      |            ^    ~~~
../gcc/testsuite/rust/compile/bad_as_bool_char.rs:14:12: error: invalid cast 'char' to 'bool' [E0054]
   14 |   let bb = b as bool;    // { dg-error "invalid cast .char. to .bool. \\\[E0054\\\]" }
      |            ^    ~~~~
../gcc/testsuite/rust/compile/bad_as_bool_char.rs:17:12: error: invalid cast 'u32' to 'char' [E0054]
   17 |   let ab = t32 as char;  // { dg-error "invalid cast" }
      |            ^~~    ~~~~

```

This commit is split from pr #1718.

### DIFF
--- a/gcc/rust/lex/rust-lex.cc
+++ b/gcc/rust/lex/rust-lex.cc
@@ -447,6 +447,7 @@ Lexer::build_token ()
 	      // match arm arrow
 	      skip_input ();
 	      current_column += 2;
+	      loc += 1;
 
 	      return Token::make (MATCH_ARROW, loc);
 	    }
@@ -455,6 +456,7 @@ Lexer::build_token ()
 	      // equality operator
 	      skip_input ();
 	      current_column += 2;
+	      loc += 1;
 
 	      return Token::make (EQUAL_EQUAL, loc);
 	    }
@@ -473,6 +475,7 @@ Lexer::build_token ()
 	      // return type specifier
 	      skip_input ();
 	      current_column += 2;
+	      loc += 1;
 
 	      return Token::make (RETURN_TYPE, loc);
 	    }
@@ -481,6 +484,7 @@ Lexer::build_token ()
 	      // minus-assign
 	      skip_input ();
 	      current_column += 2;
+	      loc += 1;
 
 	      return Token::make (MINUS_EQ, loc);
 	    }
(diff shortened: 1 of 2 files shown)


gcc/rust/rust-location.h
```diff
@@ -34,6 +34,8 @@ class Location
 
   location_t gcc_location () const { return gcc_loc_; }
 
+  void set_offset (unsigned int offset) { gcc_loc_ += offset; }
```
CohenArthur, line 37: This doesn't set the offset, but it does increment it.
```suggestion
  void increment_offset (unsigned int offset) { gcc_loc_ += offset; }
```
The function in itself is also not "needed" since there are already operator overloads for the `Location` type. So you can do `loc += 1` instead of `loc.set_offset(1)` for example.

I'm not a fan of operator overloading, but this is what we do in other parts of the compiler so I'm fine with it. I'd use these APIs instead :)

CohenArthur (commented):
I think this is a good idea. I wonder if it's possible to do assertions in the testsuite to make sure that the carrets/tildes are positioned nicely? @dkm @tschwinge any idea?

PR #1734, opened 2023-01-11
Add get_locus function for abstract class MetaItemInner.

TuringKi:
This commit adds virtual function get_locus to base class MetaItemInner, which is helpful when we need to print  diagnostics on some sub-classes of MetaItemInner.

This commit is split from pr https://github.com/Rust-GCC/gccrs/pull/1718.

### DIFF
--- a/gcc/rust/ast/rust-expr.h
+++ b/gcc/rust/ast/rust-expr.h
@@ -150,6 +150,8 @@ public:
 
   std::string as_string () const override { return lit_expr.as_string (); }
 
+  Location get_locus () const override { return lit_expr.get_locus (); }
+
   void accept_vis (ASTVisitor &vis) override;
 
   bool check_cfg_predicate (const Session &session) const override;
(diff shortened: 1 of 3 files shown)


gcc/rust/ast/rust-expr.h
```diff
@@ -178,6 +180,12 @@ class MetaItemPathLit : public MetaItem
     return path.as_string () + " = " + lit.as_string ();
   }
 
+  // There are two Locations in MetaItemPathLit (path and lit_expr),
+  //  we have no idea use which of them, just simply return UNKNOWN_LOCATION
+  //  now.
+  // Maybe we will figure out when we really need the location in the future.
+  Location get_locus () const override { return Location (UNKNOWN_LOCATION); }
```
CohenArthur, line 187: I think this should return the "earlier" of the two locations. In the case of `MetaItemPathLit`, which is of the form `path = lit`, I think this should return the path's location.
```suggestion
  Location get_locus () const override { return path.get_locus (); }
```

CohenArthur (approved):
LGTM!

philberty (approved):

PR #1736, opened 2023-01-11
Change how CompileVarDecl outputs Bvariable's

powerboat9:
This allows patterns to declare multiple/no variables

---

### DIFF
--- a/gcc/rust/backend/rust-compile-var-decl.h
+++ b/gcc/rust/backend/rust-compile-var-decl.h
@@ -30,12 +30,11 @@ class CompileVarDecl : public HIRCompileBase, public HIR::HIRPatternVisitor
   using HIR::HIRPatternVisitor::visit;
 
 public:
-  static ::Bvariable *compile (tree fndecl, tree translated_type,
-			       HIR::Pattern *pattern, Context *ctx)
+  static void compile (tree fndecl, tree translated_type, HIR::Pattern *pattern,
+		       std::vector<Bvariable *> &locals, Context *ctx)
   {
-    CompileVarDecl compiler (ctx, fndecl, translated_type);
+    CompileVarDecl compiler (ctx, fndecl, translated_type, locals);
     pattern->accept_vis (compiler);
-    return compiler.compiled_variable;
   }
 
   void visit (HIR::IdentifierPattern &pattern) override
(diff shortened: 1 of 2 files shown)


gcc/rust/backend/rust-compile-var-decl.h
```diff
@@ -30,39 +30,42 @@ class CompileVarDecl : public HIRCompileBase, public HIR::HIRPatternVisitor
   using HIR::HIRPatternVisitor::visit;
 
 public:
-  static ::Bvariable *compile (tree fndecl, tree translated_type,
-			       HIR::Pattern *pattern, Context *ctx)
+  static void compile (tree fndecl, tree translated_type, HIR::Pattern *pattern,
+		       std::vector<Bvariable *> *locals, Context *ctx)
```
CohenArthur, line 34: How about passing a reference here?
```suggestion
		       std::vector<Bvariable *> &locals, Context *ctx)
```

gcc/rust/backend/rust-compile-var-decl.h
```diff
@@ -78,15 +81,16 @@ class CompileVarDecl : public HIRCompileBase, public HIR::HIRPatternVisitor
   void visit (HIR::TupleStructPattern &) override {}
 
 private:
-  CompileVarDecl (Context *ctx, tree fndecl, tree translated_type)
+  CompileVarDecl (Context *ctx, tree fndecl, tree translated_type,
+		  std::vector<Bvariable *> *locals)
```
CohenArthur, line 85: Same here

gcc/rust/backend/rust-compile-var-decl.h
```diff
@@ -78,15 +81,16 @@ class CompileVarDecl : public HIRCompileBase, public HIR::HIRPatternVisitor
...
     : HIRCompileBase (ctx), fndecl (fndecl), translated_type (translated_type),
-      compiled_variable (ctx->get_backend ()->error_variable ())
+      locals (locals)
   {}
 
   tree fndecl;
   tree translated_type;
 
-  Bvariable *compiled_variable;
+  std::vector<Bvariable *> *locals;
```
CohenArthur, line 93: And then keep a reference here instead :)
```suggestion
  std::vector<Bvariable *> &locals;
```

CohenArthur (commented):
Code wise this looks good to me apart from the raw pointer. Waiting for Philip's review as he knows more about this part of the project

philberty (commented):
I've been spending a good bit of time writing and rewriting my thoughts on this grouped pattern support. I think it depends on understanding the rustc behaviour from my side better.

I think the question is when we restructure:

```
a, b = (123, 456)
```

Is a and b a copy of the fields or do the reference the fields? My initial thoughts here are that a,b should be field access expressions and not stack variables, so that they reference the structure storage itself and avoid copying.

Maybe @flip1995 can help answer that for me. If we can figure out what the behaviour is it determines the approach here.

If we go with this approach, we are going to be taking copy's and offloading the destructure logic to the assignment, or we can treat a,b as a special grouped variable so the assignment does not create two new variables but we reuse the pattern binding code from our match expression which then any reference to a or b we actually get back field access expressions. 

I will try and write some rustc examples to figure this out later as well.

# TLDR

So from C perspective is this:

```rust
let a,b = (123, 456);
let c = a + b;
```

solution A:

```c
struct Foo { 
  i32 x;
  i32 y;
}

struct Foo tmp = Foo(123, 456);
i32 a = tmp.0;
i32 b = tmp.1;
i32 c = a + b;
```

This means we get real variables and this patch is the right approach but means we are taking copys.

solution B

```c
struct Foo { 
  i32 x;
  i32 y;
}

struct Foo tmp = Foo(123, 456);
i32 c = tmp.0 + tmp.1;
```

This looses debug info but means a,b just reuse the same storage.

philberty:
Ok it looks like it is infact copy's https://godbolt.org/z/eh93YoYT4

philberty (changes requested):
Ok I think I will be happy with this once you fix up @CohenArthur 's comments. We use references where possible.

powerboat9:
Checks are failing because of dependency downloading errors.

flip1995:
Unlike whatever C++ does with structured bindings ([really weird behavior](https://stackoverflow.com/questions/46114214/lambda-implicit-capture-fails-with-variable-declared-from-structured-binding)), in Rust 

```rust
let t = (123, 456);
let (a, b) = t;
```

Is the same as 

```rust
let a = t.0;
let b = t.1;
```

I would assume, that if you compile in release mode or with `-O`/`-Copt-level=3` the copies are optimized away.

CohenArthur (approved):
LGTM :) I asked the CI to re-run on failed jobs but this should not be a problem

philberty:
Just checking CI is good.

philberty (approved):
Good job thanks for working through this with us!

PR #1737, opened 2023-01-12
Add  support for feature check.

TuringKi:
This commit implements a very basic feature checking module.

The results like this:

```

../gcc/testsuite/rust/compile/feature.rs:1:12: error: unknown feature 'AA'
    1 | #![feature(AA)] //{ dg-error "unknown feature 'AA'" }
      |            ^

```

This commit is split from pr https://github.com/Rust-GCC/gccrs/pull/1718.

### DIFF
--- a/gcc/rust/checks/errors/rust-feature-gate.cc
+++ b/gcc/rust/checks/errors/rust-feature-gate.cc
@@ -1,0 +1,63 @@
+// Copyright (C) 2020-2022 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#include "rust-feature-gate.h"
+#include "rust-feature.h"
+
+namespace Rust {
+
+void
+FeatureGate::check (AST::Crate &crate)
+{
+  std::vector<Feature> valid_features;
+  for (const auto &attr : crate.inner_attrs)
+    {
+      if (attr.get_path ().as_string () == "feature")
+	{
+	  const auto &attr_input = attr.get_attr_input ();
+	  auto type = attr_input.get_attr_input_type ();
+	  if (type == AST::AttrInput::AttrInputType::TOKEN_TREE)
+	    {
+	      const auto &option = static_cast<const AST::DelimTokenTree &> (
+		attr.get_attr_input ());
+	      std::unique_ptr<AST::AttrInputMetaItemContainer> meta_item (
+		option.parse_to_meta_item ());
+	      for (const auto &item : meta_item->get_items ())
+		{
+		  const auto &name = item->as_string ();
+		  auto tname = Feature::as_name (name);
+		  if (!tname.is_none ())
+		    valid_features.push_back (Feature::create (tname.get ()));
+		  else
+		    rust_error_at (item->get_locus (), "unknown feature '%s'",
+				   name.c_str ());
+		}
+	    }
+	}
+    }
+  valid_features.shrink_to_fit ();
+
+  // TODO (mxlol233): add the real feature gate stuff.
+  auto &items = crate.items;
+  for (auto it = items.begin (); it != items.end (); it++)
+    {
+      auto &item = *it;
+      item->accept_vis (*this);
+    }
+}
+} // namespace Rust
--- a/gcc/rust/checks/errors/rust-feature.cc
+++ b/gcc/rust/checks/errors/rust-feature.cc
@@ -1,0 +1,66 @@
+// Copyright (C) 2020-2022 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#include "rust-feature.h"
+#include "rust-session-manager.h"
+
+namespace Rust {
+
+Feature
+Feature::create (Feature::Name name)
+{
+  switch (name)
+    {
+    case Feature::Name::ASSOCIATED_TYPE_BOUNDS:
+      return Feature (Feature::Name::ASSOCIATED_TYPE_BOUNDS,
+		      Feature::State::ACCEPTED, "associated_type_bounds",
+		      "1.34.0", 52662,
+		      Optional<CompileOptions::Edition>::none (), "");
+    case Feature::Name::INTRINSICS:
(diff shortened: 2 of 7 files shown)


gcc/rust/checks/errors/rust-feature.h
```diff
@@ -0,0 +1,54 @@
...
+    ACCEPTED,
+    ACTIVE,
+    REMOVED,
+    STABILIZED,
+  };
+
+  Feature (State S, const char *N, const char *RS, const char *GS, uint64_t I,
+	   const Optional<CompileOptions::Edition> &E, const char *D)
+    : state (S), name (N), rustc_since (RS), gccrs_since (GS), issue (I),
+      edition (E), description (D)
```
CohenArthur, line 40: As mentioned on the other PR

1. In my opinion the parameters are not named appropriately and would benefit from a more descriptive name
2. I think we should remove the `gccrs_since` argument and field in the class

gcc/rust/checks/errors/rust-feature.cc
```diff
@@ -0,0 +1,67 @@
...
+				 Optional<CompileOptions::Edition>::none (),
+				 ""),
+  DECLARE_ACTIVE_FEATURE_IN_MAP ("rustc_attrs", "1.0.0", "0.0.0", 0,
+				 Optional<CompileOptions::Edition>::none (),
+				 ""),
+  DECLARE_ACTIVE_FEATURE_IN_MAP ("decl_macro", "1.0.0", "0.0.0", 0,
+				 Optional<CompileOptions::Edition>::none (),
+				 ""),
+
+};
```
CohenArthur, line 65: Is it not possible to avoid the use of macros here? I thought this would be okay with C++'s `constexpr` and so on

CohenArthur (commented):
Thanks! This is in a very good state. I have one question regarding the `Feature` class: where is the edition coming from? I'm not aware of Rust features being gated behind an edition, but it might be wrong. Is that directly taken from the rustc implementation?

gcc/rust/checks/errors/rust-feature.cc
```diff
@@ -0,0 +1,67 @@
...
+  {                                                                            \
+    (N),                                                                       \
+    {                                                                          \
+      (Feature (DECLARE_ACCEPTED_FEATURE ((N), (RS), (GS), (I), (E), (D))))    \
+    }                                                                          \
+  }
+
+const std::map<std::string, const Feature &> Feature::feature_hash_map = {
+
+  DECLARE_ACTIVE_FEATURE_IN_MAP ("associated_type_bounds", "1.34.0", "0.0.0",
```
philberty, line 51: we need these strings to also have a public enum associated with them then you can use a Feature::ToString for this

philberty (changes requested):
I agree with @CohenArthur we should name our parameters on the Feature constructor here as it helps with readability.

Apart from that the idea here LGTM. There are follow up changes we need to make here.

There is the problem where you use a feature that is not specified in the code and there are features that are allowed by default etc. This scan will only determine the specified features which is the right starting point.

The next PR should be to make the FeatureGate have a function:

```bool featureAvailable (enum Feature);```

So when you do the feature gate check we do it like this:

```
FeatureGate feature_gate;
feature_gate.check (hir);
```

We also need a method:

```bool ok () const;```

this returns true if any errors based on the feature's gates have been emitted.

This along with the public enum will mean we can pass this object as a dependency to the code-generation context so during code-generation we can check if we are compiling something that uses a feature of some kind we can check if its available in the feature gate object. Though we might actually make that a separate error check pass over the crate again but either way it doesn't matter.

**Comment by philberty** [comment]
TLDR i would be happy to merge this PR if you fix:

1. naming of the parameters in the Feature constructor
2. add a new enum for each of the features you have defined so far.

**Comment by philberty** [comment]
Actually after rereading the PR maybe your intention is to also walk the whole crate for features that will require a feature check? I still think the features should be an enum though

**Comment by TuringKi** [comment]
> TLDR i would be happy to merge this PR if you fix:
> 
> 1. naming of the parameters in the Feature constructor
> 2. add a new enum for each of the features you have defined so far.

Thanks for review this pr. I left for Chinese New Year (春节 in chinese). I will check it out about your advice in 1-2 days.

File: gcc/rust/checks/errors/rust-feature.h [hunk 4]
```diff
@@ -0,0 +1,73 @@
...
+
+  static Optional<Name> as_name (const std::string &name);
+  static Feature create (Name name);
+
+private:
+  Feature (Name name, State S, const char *N, const char *RS, uint64_t I,
+	   const Optional<CompileOptions::Edition> &E, const char *D)
+    : m_state (S), m_name (name), m_name_str (N), m_rustc_since (RS), issue (I),
+      edition (E), m_description (D)
+  {}
```
CohenArthur, line 59: @TuringKi the argument names to the function are still not precise enough IMO
```suggestion
  Feature (Name name, State state, const char *name_str, const char *rustc_since, uint64_t issue_number,
	   const Optional<CompileOptions::Edition> &edition, const char *description)
    : m_state (state), m_name (name), m_name_str (name_str), m_rustc_since (rustc_since), issue (issue_number),
      edition (edition), m_description (description)
  {}
```

CohenArthur (commented):
> Thanks for review this pr. I left for Chinese New Year (春节 in chinese). I will check it out about your advice in 1-2 days.

Hope you had a good time :D I've added a suggestion for the arguments name and if you want we can work together on removing the macros and adding @philberty's enum idea

TuringKi:
> > Thanks for review this pr. I left for Chinese New Year (春节 in chinese). I will check it out about your advice in 1-2 days.
> 
> Hope you had a good time :D I've added a suggestion for the arguments name and if you want we can work together on removing the macros and adding @philberty's enum idea

The macro is removed, and the constructor parameter names are changed.

```
enum class Name
  {
    ASSOCIATED_TYPE_BOUNDS,
    INTRINSICS,
    RUSTC_ATTRS,
    DECL_MACRO,
  };
```

I'm not sure this enum definition is ok?

philberty (approved):
LGTM nice work! I am excited to start walking the crate. And you are right to choose the AST here because the HIR is desugared too much.

philberty:
I am happy with this, its a good starting point! Leaving it over to @CohenArthur to merge when he is happy.

gcc/rust/checks/errors/rust-feature-gate.cc
```diff
@@ -0,0 +1,67 @@
...
+		  auto tname = Feature::as_name (name);
+		  if (!tname.is_none ())
+		    {
+		      valid_features.push_back (Feature::create (tname.get ()));
+		    }
+		  else
+		    {
+		      rust_error_at (item->get_locus (), "unknown feature '%s'",
+				     name.c_str ());
+		    }
```
CohenArthur, line 52: Minor GCC coding-style nitpick
```suggestion
		  if (!tname.is_none ())
		      valid_features.push_back (Feature::create (tname.get ()));
		  else
		      rust_error_at (item->get_locus (), "unknown feature '%s'",
				     name.c_str ());
```

gcc/rust/checks/errors/rust-feature.cc
```diff
@@ -0,0 +1,68 @@
...
+};
+
+Optional<Feature::Name>
+Feature::as_name (const std::string &name)
+{
+  if (Feature::name_hash_map.count (name))
+    {
+      return Optional<Feature::Name>::some (Feature::name_hash_map.at (name));
+    }
+  return Optional<Feature::Name>::none ();
```
CohenArthur, line 65:
```suggestion
  if (Feature::name_hash_map.count (name))
      return Optional<Feature::Name>::some (Feature::name_hash_map.at (name));
      
  return Optional<Feature::Name>::none ();
```
this is a great use of `Optional` btw :) thanks for using it

CohenArthur (approved):
Great stuff :) I agree that this is a great starting point. Feel free to ignore the style nitpick and I'll fix them in a subsequent PR, or you can fix them now and I'll merge this PR later today. Thanks for working on this!

TuringKi:
The  coding-style  problem is fixed. :)

TuringKi:
> Great stuff :) I agree that this is a great starting point. Feel free to ignore the style nitpick and I'll fix them in a subsequent PR, or you can fix them now and I'll merge this PR later today. Thanks for working on this!

Thanks. I will add "intrinsic" feature gate after this pr is merged.

CohenArthur:
Thanks a lot @TuringKi ! Great job :D

bors r+

PR #1750, opened 2023-01-17
Implement fixed point macro expansion

### LINKED ISSUE #1795: Implement eager expansion for `env!()`

CohenArthur:
This is a cleanup of #1606's branch, which also contains the code necessary for performing eager macro expansion in builtin macros. 

This commit changes our macro expansion system from an eager and recursive
macro expansion to a fixed-point like system. Instead of, when seeing
a macro invocation, expanding it and all of the macros within it, we
now perform multiple passes of expansion on the entire crate.

This, however, leads to a problem. Rust macros are expanded lazily, but
Rust builtin macros should be expanded eagerly. Due to this, we must
work around the lazy expansion in builtin macros and perform eager
expansion for each pass of the fixed-point, before finally expanding
the builtin when there are no longer any inner macro invocations.

To perform proper macro scoping, the ENR now keeps track of the current
scope (`current_scope` member) and resolves macros accordingly.

This is done through the use of the `scoped` method, which creates a new
scope, runs a specified lambda and then exits the scope. This prevents
pushing/popping errors that we've seen happen already in similar
contexts.

We might think about generalizing it to other classes, providing a
 `Scoped<EntryFn, ExitFn>` class or similar
 
Fixes #1795


CohenArthur:
I have put together a list of builtin macros which need to be expanded eagerly as of rust 1.49. I'll open up a tracking issue with a list of them as they are good first issue for someone interested in the macro system to contribute to, but I'd like to get this merged asap and keep going.

CohenArthur:
The last remaining code issue is debugging why `env` outputs errors twice. Probably something I messed up in the expansion pipeline.

CohenArthur:
> The last remaining code issue is debugging why `env` outputs errors twice. Probably something I messed up in the expansion pipeline.

Fixed by actually stopping the fixed-point when seeing errors :) I'll spend a little more time adding some documentation as the code is not easy and then I'll merge this.

CohenArthur:
bors cancel
bors r+

Issue #1747, opened 2023-01-17
Revert invalid testsuite timeout


CohenArthur:
Introduced in 997894f6372d057e22cd54cc7b9fefdfdc77520a

DrMahad:
- @CohenArthur could you assign me this issue & mentor me. 😄

Parthib314:
Hey @CohenArthur,  can I work on this issue?

CohenArthur:
Yup go ahead @Parthib314 ! Thanks :)

philberty:
Added context: https://github.com/Rust-GCC/gccrs/pull/1759#issuecomment-1406357309

CohenArthur:
Closed by #1759

PR #1759, opened 2023-01-21
Revert 10-second timeout for each test

Parthib314:
* Fixes issue #1747 

**Remove the 10-second timeout for each compile case**

### DIFF
--- a/gcc/testsuite/lib/rust.exp
+++ b/gcc/testsuite/lib/rust.exp
@@ -168,10 +168,7 @@ proc rust_target_compile { source dest type options } {
     global gluefile wrap_flags
     global ALWAYS_RUSTFLAGS
     global RUST_UNDER_TEST
-    global individual_timeout
-
-    # HACK: guard against infinite loops in the compiler
-    set individual_timeout 10
+    
 
     if { [target_info needs_status_wrapper] != "" && [info exists gluefile] } {
 	lappend options "libs=${gluefile}"


Parthib314:
Hey @CohenArthur if you have time then please review this PR.

philberty:
Hi @Parthib314 you need to edit your top message to remove the checklist as all of this info will be contained within your merge commit which is not what we want.

Also I am not sure why we are removing the timeout, I guess it must be something from the gcc-upstream review. @dkm or @tschwinge any thoughts?

Parthib314:
@philberty Is this correct now???

CohenArthur:
> Also I am not sure why we are removing the timeout, I guess it must be something from the gcc-upstream review. @dkm or @tschwinge any thoughts?

@philberty Upon fixing the testsuite for Windows, we realized that no other frontends were setting timeouts this way. There should already be (if I'm not mistaken) a 300 seconds timeout (so 5 minutes), which makes much more sense considering some tests might take longer than 10 seconds to run on slower machines. IIRC, the original commit was also added as a "hack" to circumvent some tests looping infinitely, and maybe we just didn't wait long enough for the 5 minutes timeout to kick in? But the current testsuite passes properly.

CohenArthur (approved):

philberty:
@CohenArthur ah ok yes thats right i forgot about the default timeout of 300 seconds. Lets add your comment to the issue.

PR #1770, opened 2023-01-24
fixes #1766

### LINKED ISSUE #1766: Refactor rust-buffered-queue.h
This is a util class and should be moved into the util folder.

SainiAditya1:

in the git commit message.

- \[x] Run `clang-format`

order to get feedback.
---


gcc/rust/util/rust-buffered-queue.h
```diff
@@ -72,7 +72,6 @@ template <typename T, typename Source> class buffered_queue
 		  end = num_queued_items;
 		  // TODO: would move be better here? optimisation for move with
 		  // shared pointer?
-
```
philberty, line 75: can you remove this empty line deletion as it separates the TODO comment with the comment on the next line

philberty (changes requested):
LGTM though i would like that line deletion to be removed.

Also you need to update your opening PR comment here to remove the check list otherwise this will be added into the bors merge commit.

philberty:
Also I see that you have done this in two commits this should be squashed into 1 commit. You can do this via:

```git rebase -i HEAD~2```

gcc/rust/util/rust-buffered-queue.h
SainiAditya1: @philberty , ok I will fix it.

philberty:
@SainiAditya1 I think this pr should be a:

```git mv rust-buffered-queue.h util/```

There shouldn't be any edits to the file it self

PR #1771, opened 2023-01-25
Add type resolution for grouped patterns

powerboat9:

Adds type resolution for GroupedPattern.


philberty (approved):
LGTM 

I am starting to think once we get your stuff merged in we should look to see if we can desugar the HIR here, so instead of an HIR::GroupedPattern we literly lower the inner pattern and return that.. as this sort of feels very AST'y. But lets not do that yet.

PR #1739, opened 2023-01-15
Initial type bounds checking for all type checks

philberty:
This patchset is my branch for general improvements to the type-system and
for improved type bounds checking in preparation for opaque types.

Fixes #1773 #1786

PR #1774, opened 2023-01-27
Remove monomorphization hack to setup possible associated types

### LINKED ISSUE #1773: Bad type resolution
I tried this code:

```rust
trait Foo {
    type A;

    fn test(a: Self::A) -> Self::A {
        a
    }
}

struct Bar(i32);
impl Foo for Bar {
    type A = i32;
}

struct Baz(f32);
impl Foo for Baz {
    type A = f32;
}

fn main() {
    let a;
    a = Bar(123);

    let b;
    b = Bar::test(a.0);

    let c;
    c = Baz(123f32);

    let d;
    d = Baz::test(c.0);
}

```

I expected to see this happen: resolve the associated types properly.

```
test.rs:24:9: note: resolved_call_expr to: {fn<Bar> (a <placeholder:<Projection=::i32>>,) -> <placeholder:<Projection=::i32>>}
   24 |     b = Bar::test(a.0);
      |         ^~~
```

Instead, this happened: 

This test does work in our current master, but it only works by using a hack to apply monomorphization early as our call expression gets resolved to the trait definition.

```
test.rs:24:9: note: resolved_call_expr to: {fn<Bar> (a <placeholder:>,) -> <placeholder:>}
   24 |     b = Bar::test(a.0);
      |         ^~~
```

Then monomorphizaton gets applied and we get:

```
rust1: note: unify_site id={42} expected={Bar} expr={Bar}
rust1: note: coercion_site id={74} expected={<placeholder:<Projection=::i32>>} expr={i32}
rust1: note: coerce_unsized(source={i32}, target={<placeholder:<Projection=::i32>>})
rust1: note: coerce_default_unify(a={i32}, b={<placeholder:<Projection=::i32>>})
rust1: note: unify_site id={74} expected={<placeholder:<Projection=::i32>>} expr={i32}
```

This results in extra unification going on to resolve inference variables which is not required in this case and is throwing off inference variables down the line in more complex cases.

### Meta

- What version of Rust GCC were you using, git sha if possible. fe828b4f931ca216a4b74322a4fdb527faada53e

philberty:
Remove this hack to improve type-checking for bounds.

Fixes #1773

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-path.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-path.cc
@@ -374,6 +374,34 @@ TypeCheckExpr::resolve_segments (NodeId root_resolved_node_id,
 	    {
 	      // get the associated impl block
 	      associated_impl_block = impl;
+	      HirId impl_block_id = impl->get_mappings ().get_hirid ();
+
+	      AssociatedImplTrait *associated = nullptr;
+	      bool found_impl_trait
+		= context->lookup_associated_trait_impl (impl_block_id,
+							 &associated);
+	      if (!found_impl_trait)
+		{
+		  // FIXME
+		  // Can we error? This seems like we should error out as we
+		  // cannot resolve any possible associated types? This needs
+		  // some thought as it may not be a useable facing error, we
+		  // might be resolving a crate with type error's occuring early
+		  // on so it may not be helpful in the end.
+		}
+	      else
+		{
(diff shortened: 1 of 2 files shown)


philberty:
To fix the generic FIXME i need to do it on my big refactor branch

philberty:
I have the proper fixed within my refactoring branch instead over on https://github.com/Rust-GCC/gccrs/pull/1739

Issue #867, opened 2022-01-06
Bad error handling on wildcard pattern within expression


philberty:
I tried this code:

```rust
fn test() {
    let _ = 42;
    let a = _ + 123;
}

```

I expected to see this happen: 

```
error: expected expression, found reserved identifier `_`
 --> test.rs:3:13
  |
3 |     let a = _ + 123;
  |             ^ expected expression
```

Instead, this happened:

```
test.rs:3:13: error: found unexpected token ‘_’ in null denotation
    3 |     let a = _ + 123;
      |             ^
test.rs:3:15: error: failed to parse expression in let statement
    3 |     let a = _ + 123;
      |               ^
test.rs:3:5: error: failed to parse statement or expression without block in block expression
    3 |     let a = _ + 123;
      |     ^
test.rs:4:1: error: unrecognised token ‘}’ for start of item
    4 | }
      | ^
test.rs:4:1: error: failed to parse item in crate
test.rs:4:1: error: attempted to get string for ‘}’, which has no string. returning empty string instead

```

### Meta

- What version of Rust GCC were you using, git sha if possible. 69f6be3ee483c9895b4b5187a44b3e1c8be2ba63

00AR:
I add a new switch case in **null_denotation**(rust-parse-impl.h) to handle '_'

    case UNDERSCORE:
        add_error (Error (tok->get_locus (),
            "expected expression, found reserved identifier %qs",
            tok->get_token_description ()));
            return nullptr;

It seems to have worked, now the output looks like this

    main.rs:3:13: error: expected expression, found reserved identifier ‘_’
        3 |     let a = _ + 123;
          |             ^
    main.rs:3:15: error: failed to parse expression in let statement
        3 |     let a = _ + 123;
          |               ^
    main.rs:3:5: error: failed to parse statement or expression without block in block expression
        3 |     let a = _ + 123;
          |     ^
    main.rs:4:1: error: unrecognized token ‘}’ for start of item
        4 | }
          | ^
    main.rs:4:1: error: failed to parse item in crate
I wanted to know if this is the right fix?

CohenArthur:
@00AR the fix looks fine to me but the error message is not the clearest IMO. 

rustc example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021

We don't have to copy that error message exactly, just to be clear. Feel free to submit your patch as a pull-request and we can work on it :) thanks for taking the time!

PR #1776, opened 2023-01-28
fixed compiler error message on wildcard pattern within expression

00AR:
Signed-off-by: Abdul Rafey <abdulrafeyq@gmail.com>

fixed compiler error message on wildcard pattern '_'  within expression.
Added a new testcase for the same.
#867

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -13021,6 +13021,12 @@ Parser<ManagedTokenSource>::null_denotation (const_TokenPtr tok,
     case UNSAFE:
       return parse_unsafe_block_expr (std::move (outer_attrs),
 				      tok->get_locus ());
+    case UNDERSCORE:
+      add_error (
+	Error (tok->get_locus (),
+	       "use of %qs is not allowed on the right-side of an assignment",
+	       tok->get_token_description ()));
+      return nullptr;
     default:
       if (!restrictions.expr_can_be_null)
 	add_error (Error (tok->get_locus (),
--- a/gcc/testsuite/rust/compile/issue-867.rs
+++ b/gcc/testsuite/rust/compile/issue-867.rs
@@ -1,0 +1,8 @@
+fn main() {
+    let _ = 42;
+    let a = _ + 123; // { dg-error "use of '_' is not allowed on the right-side of an assignment" }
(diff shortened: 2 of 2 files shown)


gcc/rust/parse/rust-parse-impl.h
```diff
@@ -12900,6 +12900,12 @@ Parser<ManagedTokenSource>::null_denotation (const_TokenPtr tok,
     case UNSAFE:
       return parse_unsafe_block_expr (std::move (outer_attrs),
 				      tok->get_locus ());
+    case UNDERSCORE:
+      add_error (
+	Error (tok->get_locus (),
+	       "Use of %qs is not allowed on the right-side of an assignment",
```
CohenArthur, line 12906:
```suggestion
	       "use of %qs is not allowed on the right-side of an assignment",
```

We usually use lowercase error messages, as it looks better with GCC's internal formatter

CohenArthur (approved):
Looks good to me apart from the very tiny nitpick :) Once you fix it ping me and I'll merge this PR. Thanks for working on this!

gcc/rust/parse/rust-parse-impl.h
00AR: I have made the change. @CohenArthur Thanks for reviewing!

gcc/testsuite/rust/compile/issue-867.rs
```diff
@@ -0,0 +1,8 @@
+fn main() {
+    let _ = 42;
+    let a = _ + 123; // { dg-error "Use of '_' is not allowed on the right-side of an assignment" }
```
CohenArthur, line 3: This test is going to fail now :DD but no worries it's not a big deal. Once that's fixed I'll merge it :)

CohenArthur (approved):
Last stretch haha. Sorry! I should have thought about it in my previous review

gcc/testsuite/rust/compile/issue-867.rs
00AR: > This test is going to fail now

became a bit lazy :D 
Fixed the testcase and this time made sure it passes. Also I rebased this PR too.

CohenArthur (approved):
Great stuff :) Thanks for taking the time!

bors r+

PR #1779, opened 2023-01-30
Add SoB CI

CohenArthur:


.github/workflows/commit-format.yml
```diff
@@ -41,4 +52,32 @@ jobs:
...
+        run: |
+          if echo ${{ secrets.FSF_CONTRIBUTORS }} | grep ${{ github.event.pull_request.user.login }}; then
+            echo "author has an FSF copyright assignment"
+          fi
+
+          retval=0
+          for commit in $(git rev-list origin/${{ github.event.pull_request.base.ref }}..${{ github.event.pull_request.head.sha }});
+          do
+            echo -n "Checking commit sign-off for $commit: "
+            if [[ $(git log $commit -1 --format="%B" | tail -n 2) = Signed-off-by:* ]]; then
```
tschwinge, line 76: Also need to consider `Co-authored-by` tags (any case)?

tschwinge, line 76: Need not just any `Signed-off-by:*`, but rather that it does match the Git commit Author or `Co-authored-by` tag(s)?

CohenArthur: Should they really be considered? I don't mind writing the shell to do so, but it does seem maybe a bit overkill considering we have very little Co-authored-by commits. If you think it's necessary I'll do it :)

CohenArthur: Right. Let me know if the script sort of looks good to you in its current form and I'll make that change :)

CohenArthur:
Rebasing and merging :) this commit does not need a Changelog because it won't be upstreamed to GCC

PR #1783, opened 2023-01-31
Remove HIR::GroupedPattern

powerboat9:

Removes HIR::GroupedPattern and lowers AST::GroupedPattern to its inner pattern directly.


philberty:
Ok let me get back to you about this tomorrow I just want to think about it a little longer. The code LGTM and this makes our Patterns the same as the rustc hir patterns: https://doc.rust-lang.org/beta/nightly-rustc/rustc_hir/hir/enum.PatKind.html

I am wondering if there are any implications we are not considering. It would really help if we fix the horrible coding pattern in the backend where we visit a type lookup the type for that mapping id and fatal_error out. I think I am overthinking everything as usual but yeah I will 99% chance merge this in the morning after fresh coffee.... :D

philberty (approved):
LGTM thanks for bearing with me, I am happy to see this going in :D

PR #1787, opened 2023-01-31
Update `gcc-patch-dev` after first half update

CohenArthur:
Not sure if this is the right way to go about updating `gcc-patch-dev` with GCC's master branch. But we should do this :)

My guess is we'll simply force-push this to the `gcc-patch-dev` branch, but we might be missing the CI commits. Or should this maybe be turned into a cherry-pick of all commits present on GCC's master but not `gcc-patch-dev`?

When trying to simply merge GCC's master into `gcc-patch-dev` a *lot* of conflicts pop up, mostly about Copyright lines but some other more annoying ones.

Locally, this builds and passes the tests so I'm not too worried.


CohenArthur:
I definitely cocked something up as this is missing CI commits and has a lot of conflicts :) I'll take care of it later. If someone wants to take ownership of this PR go ahead

PR #1788, opened 2023-01-31
ci: Add commit format checker

CohenArthur:
Bring over the commit checker from `gcc-patch-dev` with the `gccrs` prefix checker enabled only for `gcc-patch-dev` PRs. I'll open up an issue to make sure that in the future it's easier to keep that folder in sync between the two branches


CohenArthur:
Okay, let's go! Let's make sure this commit is the last ever without a Changelog entry :)

bors r+

Issue #1813, opened 2023-02-01
ICE in handling for `ReferencePattern`s


CohenArthur:
I tried this code:

```rust
fn foo(&a: &char) {}
```

Another testcase:
```rust
fn main() {
    let a = 15u8;
    let a = &a;
    match a {
        &15 => {}
        &14 => {}
        _ => {}
    }
}
```

What I expected to happen: no error

What happened:

```rust
rust1: internal compiler error: in translate, at rust/hir/rust-ast-lower-pattern.cc:33
0x11edb28 Rust::HIR::ASTLoweringPattern::translate(Rust::AST::Pattern*)
	/home/arthur/Git/rust-gcc/gccrs/build/gcc/../../gcc/rust/hir/rust-ast-lower-pattern.cc:33
0x11d8fc3 Rust::HIR::ASTLoweringStmt::visit(Rust::AST::Function&)
	../../gcc/rust/hir/rust-ast-lower-stmt.h:355
0xfb7840 Rust::AST::Function::accept_vis(Rust::AST::ASTVisitor&)
	/home/arthur/Git/rust-gcc/gccrs/build/gcc/../../gcc/rust/ast/rust-ast.cc:5373
0x11d6b96 Rust::HIR::ASTLoweringStmt::translate(Rust::AST::Stmt*, bool*)
	../../gcc/rust/hir/rust-ast-lower-stmt.h:42
0x11d44c1 Rust::HIR::ASTLoweringBlock::visit(Rust::AST::BlockExpr&)
	/home/arthur/Git/rust-gcc/gccrs/build/gcc/../../gcc/rust/hir/rust-ast-lower.cc:117
0xfb7130 Rust::AST::BlockExpr::accept_vis(Rust::AST::ASTVisitor&)
	/home/arthur/Git/rust-gcc/gccrs/build/gcc/../../gcc/rust/ast/rust-ast.cc:5151
0x11d6fd3 Rust::HIR::ASTLoweringBlock::translate(Rust::AST::BlockExpr*, bool*)
	../../gcc/rust/hir/rust-ast-lower-block.h:36
0x11f2c84 Rust::HIR::ASTLoweringItem::visit(Rust::AST::Function&)
	/home/arthur/Git/rust-gcc/gccrs/build/gcc/../../gcc/rust/hir/rust-ast-lower-item.cc:430
0xfb7840 Rust::AST::Function::accept_vis(Rust::AST::ASTVisitor&)
[...]

PR #1818, opened 2023-02-02
Implement lowering ReferencePattern from AST to HIR

powerboat9:
gcc/rust/ChangeLog:

	* ast/rust-pattern.h (ReferencePattern::get_has_two_amps): Add method. (ReferencePattern::get_is_mut): Add method.
	* hir/rust-ast-lower-pattern.cc (ASTLoweringPattern::visit): Add ReferencePattern visitor.
	* hir/rust-ast-lower-pattern.h (ASTLoweringPattern::visit): Add ReferencePattern visitor.

Partially addresses #1813

### DIFF
--- a/gcc/rust/hir/rust-ast-lower-pattern.cc
+++ b/gcc/rust/hir/rust-ast-lower-pattern.cc
@@ -259,5 +259,34 @@ ASTLoweringPattern::visit (AST::GroupedPattern &pattern)
 			       pattern.get_locus ());
 }
 
+void
+ASTLoweringPattern::visit (AST::ReferencePattern &pattern)
+{
+  auto crate_num = mappings->get_current_crate ();
+  Analysis::NodeMapping mapping (crate_num, pattern.get_node_id (),
+				 mappings->get_next_hir_id (crate_num),
+				 UNKNOWN_LOCAL_DEFID);
+
+  HIR::Pattern *inner
+    = ASTLoweringPattern::translate (pattern.get_referenced_pattern ().get ());
+
+  translated
+    = new HIR::ReferencePattern (mapping, std::unique_ptr<HIR::Pattern> (inner),
+				 pattern.get_is_mut () ? Mutability::Mut
+						       : Mutability::Imm,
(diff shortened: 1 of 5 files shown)


gcc/rust/hir/rust-ast-lower-pattern.cc
```diff
@@ -259,5 +259,24 @@ ASTLoweringPattern::visit (AST::GroupedPattern &pattern)
...
+
+  HIR::Pattern *inner
+    = ASTLoweringPattern::translate (pattern.get_referenced_pattern ().get ());
+
+  translated
+    = new HIR::ReferencePattern (mapping, std::unique_ptr<HIR::Pattern> (inner),
+				 pattern.get_is_mut () ? Mutability::Mut
+						       : Mutability::Imm,
+				 pattern.get_has_two_amps (),
+				 pattern.get_locus ());
```
CohenArthur, line 278: Not for this PR, but we could probably think about desugaring double references here into an `HIR::ReferencePattern(mappings, HIR::ReferencePattern(...), mut, ...)` and remove the `has_two_amps` field of the `HIR::ReferencePattern` class. In the AST, having the flag is useful and helpful because the lexer does not know what a double reference is, so you end up going through either an `AMP` ('&') token or a... `LOGICAL_AND` ('&&') token :) but I think this is a good point to strip away that obligation since the HIR does not need to know about this.

CohenArthur: Let me know if you're interested in working on this @powerboat9

CohenArthur (approved):
LGTM :) Thank you! This ties in nicely with some of the issues I opened :DD

philberty (commented):
LGTM I think this should work even though they share the same ast-node-id so it should be grand so long as they have unique hirid's because reference types are covariant ones.

Good work again.

CohenArthur (approved):
@powerboat9 I'm checking both of your commits to ensure they both bootstrap and pass the tests and then we'll merge this :) thanks a lot!

PR #1820, opened 2023-02-02
Support for TuplePattern in let statements

powerboat9:
Added support for TuplePattern in let statements, along with a test.


gcc/rust/backend/rust-compile-pattern.cc
```diff
@@ -387,5 +387,51 @@ CompilePatternLet::visit (HIR::WildcardPattern &pattern)
...
+{
+  rust_assert (!ty->is_unit ());
+  rust_assert (pattern.has_tuple_pattern_items ());
+  switch (pattern.get_items ()->get_pattern_type ())
+    {
+    default:
+      gcc_unreachable ();
+    case HIR::TuplePatternItems::TuplePatternItemType::RANGED:
+      rust_fatal_error (pattern.get_locus (),
+			"ranged tuple pattern in let statement");
```
CohenArthur, line 401: Is this something we cannot compile yet but that should be allowed? In which case, 
```suggestion
      rust_sorry_at (pattern.get_locus (),
			"ranged tuple pattern in let statement");
```
if not, why is it a fatal_error and not a regular error? Is it a compiler error to have a ranged pattern in a tuple pattern, as if "did something go wrong earlier in gccrs to get to that point"? In that case, we should have an assertion failure IMO

gcc/rust/backend/rust-compile-pattern.cc
```diff
@@ -387,5 +387,51 @@ CompilePatternLet::visit (HIR::WildcardPattern &pattern)
     }
 }
 
+void
+CompilePatternLet::visit (HIR::TuplePattern &pattern)
+{
+  rust_assert (!ty->is_unit ());
```
CohenArthur, line 393: Note that
```rust
fn main() {
    let () = ();
}
```
is valid Rust but we cannot parse it yet. It'd be nice to have a TODO or FIXME here and open up an issue (which I can do if you want)

CohenArthur (commented):
Code looks good to me :) thanks!

gcc/rust/backend/rust-compile-pattern.cc
```diff
@@ -387,5 +387,91 @@ CompilePatternLet::visit (HIR::WildcardPattern &pattern)
...
+					       false, pattern.get_locus (),
+					       &init_stmt);
+  tree access_expr
+    = ctx->get_backend ()->var_expression (tmp_var, pattern.get_locus ());
+  ctx->add_statement (init_stmt);
+
+  switch (pattern.get_items ()->get_pattern_type ())
+    {
+    default:
+      gcc_unreachable ();
```
philberty, line 409: I dont't like default being at the start of a switch

philberty (changes requested):
Minor nit on the switch statement, I would like to see two things:

1. Test case involving using the bindings of the pattern.

```
fn main () -> i32 {
  let (a,b) = (1,2)
  a + b - 3
}
```

So it ensures this should return 0.

- This is probably better as another commit/pr how do we handle when the number of elements in the pattern do not match the initilizer.

````let (a,b) = (1,2,3,4,5,6).```

I would be happy to merge if you fix the style nit in the switch statement and add usage of a binding in the test case.

powerboat9:
Requested changes are implemented, though element number mismatches and usages of RestPattern throw ICEs.

philberty (changes requested):
LGTM but you need to rebase this into 1 commit. Great work thanks for updating the test case please raise issues about the ICE when you get a chance

CohenArthur:
Looks good thank you :) bors r+

CohenArthur:
Oops forgot bors needs a newline

bors r+

PR #1821, opened 2023-02-02
Improve WildcardPattern let statement handling

powerboat9:
Improves handling of WildcardPattern in let statements.

### DIFF
--- a/gcc/rust/backend/rust-compile-pattern.cc
+++ b/gcc/rust/backend/rust-compile-pattern.cc
@@ -361,30 +361,14 @@ CompilePatternLet::visit (HIR::IdentifierPattern &pattern)
 void
 CompilePatternLet::visit (HIR::WildcardPattern &pattern)
 {
-  Bvariable *var = nullptr;
-  rust_assert (
-    ctx->lookup_var_decl (pattern.get_pattern_mappings ().get_hirid (), &var));
+  tree init_stmt = NULL;
+  tree stmt_type = TyTyResolveCompile::compile (ctx, ty);
 
-  auto fnctx = ctx->peek_fn ();
-  if (ty->is_unit ())
-    {
-      ctx->add_statement (init_expr);
+  ctx->get_backend ()->temporary_variable (ctx->peek_fn ().fndecl, NULL_TREE,
+					   stmt_type, init_expr, false,
+					   pattern.get_locus (), &init_stmt);
 
-      tree stmt_type = TyTyResolveCompile::compile (ctx, ty);
-
-      auto unit_type_init_expr
-	= ctx->get_backend ()->constructor_expression (stmt_type, false, {}, -1,
-						       rval_locus);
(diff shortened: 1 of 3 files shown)


gcc/testsuite/rust/compile/builtin_macro_recurse.rs
```diff
@@ -15,7 +15,7 @@ macro_rules! b {
 
 fn main() {
     // { dg-final { scan-tree-dump-times {"test1canary"} 1 gimple } }
-    let _ = concat!(a!(), 1, b!());
+    let _v = concat!(a!(), 1, b!());
```
CohenArthur, line 18: Why was this causing an issue?

CohenArthur (commented):
Thanks for the PR :) Once this is approved, please rebase your commits in a single one as to not have a test failure on the first one.

gcc/testsuite/rust/compile/builtin_macro_recurse.rs
powerboat9: Something about not storing the string reference in a variable.

CohenArthur: I'm not getting any issue when running the file or another code sample like
```rust
fn main() {
    let _ = "a";
}
```

can you reproduce the behavior? I'd rather not change the test if everything is good, but at the same time I want to make sure this PR does not accidentally introduce a regression. The code above is valid Rust and so is the testcase

CohenArthur: If this
```
FAIL: rust/compile/builtin_macro_recurse.rs scan-tree-dump-times gimple ""test1canary"" 1
FAIL: rust/compile/closure_multi_args.rs (test for excess errors)
```
is the error you mean then there is an issue in your PR and we're now miscompiling wildcard patterns. Sorry :(

philberty (changes requested):
So here your chaning the implementation of the wildcard to become a temporary variable which means debug info won't exist for it. I think if your going to make changes like this you need to have an explanation about it in your commit message.

I think this probably is the right thing to do because its unused and we don't care about the debug info.

powerboat9:
I reworded my commit a bit (s/Improve/Simplify/) to make its purpose a bit more clear. It isn't strictly necessary for new features, and is more of a clean-up-the-codebase kind of thing.

philberty:
Thanks @powerboat9 I am a sticker for commit messages please see an example of mine even for small commits: https://github.com/Rust-GCC/gccrs/commit/e612713029296dac8429e4b00ec7659bcad5cffe

I really believe that subject + description is important over time. I think your message is ok for now but just keep it in mind that we think commit messages are important for behaviour changes like this since we no longer emit debug info for unused locals after this commit.

CohenArthur:
I'll add this to the merge queue as soon as #1858 is merged :)

Issue #413, opened 2021-05-07
Bug in stmt vs expr parsing


flip1995:
### Code:

```rust
pub fn test() { return }
```

[Godbolt](https://godbolt.org/z/n44aWMKxW)

**I got this:**

```
<source>:1:25: error: expecting '}' but 'end of file' found
    1 | pub fn test() { return }
      |                         ^
<source>:1:17: error: error may be from having an expression (as opposed to statement) in the body of the function but not last
    1 | pub fn test() { return }
      | 
```

**I expected to see this:**

No error; successful compilation.

---

Adding a `;` after the `return` compiles successfully. 

If I had to guess what's going on here, I'd say that `return` is parsed as a keyword and is not seen as a whole expression on its own. So the parser inserts an empty expression after the `return`, so that the compiler thinks that the `return` is not the last expression and therefore requires a `;`

philberty:
this lookes fixed https://godbolt.org/z/osEqsrfc4

PR #1827, opened 2023-02-05
Move expr into base class ExprStmt

powerboat9:
Moves some of the functionality of HIR::ExprStmtWith{,out}Block into HIR::ExprStmt.

### DIFF
--- a/gcc/rust/hir/tree/rust-hir-stmt.h
+++ b/gcc/rust/hir/tree/rust-hir-stmt.h
@@ -157,7 +157,10 @@ protected:
 class ExprStmt : public Stmt
 {
   // TODO: add any useful virtual functions
+protected:
+  std::unique_ptr<Expr> expr;
 
+private:
   Location locus;
 
 public:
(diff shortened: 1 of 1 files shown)


CohenArthur (approved):
The code changes look good to me. Any reason why you moved the expr inside `ExprStmt` rather than `ExprStmtWith{,out}Block`? Just refactoring?

powerboat9:
Refactoring, with an end goal of replacing HIR::ExprStmtWith{,out}Block with HIR::ExprStmt.

philberty:
I am anxious about making this change to HIR before we fix the block expression issue within the AST do we have an open issue for that on @CohenArthur ?

philberty:
Here it is: https://github.com/Rust-GCC/gccrs/issues/1253 we should fix this first, if we change the HIR structure before the AST is fixed here we might end up in a situation where we need to revert this change and do it again to fit the model that was used to first the issue in the AST

philberty (changes requested):
I am not in favour of this change until https://github.com/Rust-GCC/gccrs/issues/1253 is fixed

powerboat9:
I'm not sure how unifying HIR::ExprStmt.* would interfere with solving #1253. If anything, it seems like it would make the eventual solution more simple. Regardless, this doesn't actually unify HIR::ExprStmt.*, its just a refactor.

philberty:
@powerboat9  I understand, but from my experince here there is a risk that we end up fixing that issue with block expression and tail expression handling in the AST which might cause quite a few changes to the HIR expression handling which might just negate your change here. I think we should avoid changing Stmt and Expression's until we are happy with block expressions, which have been notoriously tricky to get 100% correct.

CohenArthur:
@powerboat9 I agree with @philberty here. I think we might end up in a situation where that part of the HIR changes so much from fixing #1253 that theses changes will be void or hinder us. Sorry

philberty:
Sorry @powerboat9 I am going to close this PR and we can revisit this down the line when we fix the block expression issues like #1253 and https://github.com/Rust-GCC/gccrs/issues/413

PR #1833, opened 2023-02-08
Update `gcc-patch-dev` after first half update

CohenArthur:


CohenArthur:
Changelog failures are fine as it's for a merge commit. There are no diffs between GCC's master branch and this, apart from missing github workflows which is what we want. I'll merge this ASAP.

PR #1838, opened 2023-02-09
lex: Prevent directories in RAIIFile

### LINKED ISSUE #1830: `include_str` & `include_bytes` macro crash on empty string
I tried this code:

```rust
#[rustc_builtin_macro]
macro_rules! include_str {
    () => {{}};
}

fn main() {
    let x = include_str!("");
}
```

```
error: couldn't read src/: Is a directory (os error 21)
```

Instead, this happened:
```
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
rust1: internal compiler error: Aborted
0x1109ecf crash_signal
	../../gcc/toplev.cc:314
0x7f07ec29ed5f ???
	./signal/../sysdeps/unix/sysv/linux/x86_64/sigaction.c:0
0x7f07ec29ece1 __GI_raise
	../sysdeps/unix/sysv/linux/raise.c:51
0x7f07ec288536 __GI_abort
	./stdlib/abort.c:79
0xa8fe80 __gnu_cxx::new_allocator<unsigned char>::allocate(unsigned long, void const*)
	/usr/include/c++/10/ext/new_allocator.h:115
0xa8fe80 std::allocator_traits<std::allocator<unsigned char> >::allocate(std::allocator<unsigned char>&, unsigned long)
	/usr/include/c++/10/bits/alloc_traits.h:460
0xa8fe80 std::_Vector_base<unsigned char, std::allocator<unsigned char> >::_M_allocate(unsigned long)
	/usr/include/c++/10/bits/stl_vector.h:346
0xa8fe80 std::_Vector_base<unsigned char, std::allocator<unsigned char> >::_M_create_storage(unsigned long)
	/usr/include/c++/10/bits/stl_vector.h:361
0xa8fe80 std::_Vector_base<unsigned char, std::allocator<unsigned char> >::_Vector_base(unsigned long, std::allocator<unsigned char> const&)
	/usr/include/c++/10/bits/stl_vector.h:305
0xa8fe80 std::vector<unsigned char, std::allocator<unsigned char> >::vector(unsigned long, std::allocator<unsigned char> const&)
	/usr/include/c++/10/bits/stl_vector.h:511
0xa8fe80 load_file_bytes
	../../gcc/rust/expand/rust-macro-builtins.cc:284
0xaa0796 Rust::MacroBuiltin::include_str_handler(Location, Rust::AST::MacroInvocData&)
	../../gcc/rust/expand/rust-macro-builtins.cc:421
0x99ed93 Rust::AST::Fragment std::__invoke_impl<Rust::AST::Fragment, Rust::AST::Fragment (*&)(Location, Rust::AST::MacroInvocData&), Location, Rust::AST::MacroInvocData&>(std::__invoke_other, Rust::AST::Fragment (*&)(Location, Rust::AST::MacroInvocData&), Location&&, Rust::AST::MacroInvocData&)
	/usr/include/c++/10/bits/invoke.h:60
0x99ed93 std::enable_if<std::__and_<std::__not_<std::is_void<Rust::AST::Fragment> >, std::is_convertible<std::__invoke_result<Rust::AST::Fragment (*&)(Location, Rust::AST::MacroInvocData&), Location, Rust::AST::MacroInvocData&>::type, Rust::AST::Fragment> >::value, Rust::AST::Fragment>::type std::__invoke_r<Rust::AST::Fragment, Rust::AST::Fragment (*&)(Location, Rust::AST::MacroInvocData&), Location, Rust::AST::MacroInvocData&>(Rust::AST::Fragment (*&)(Location, Rust::AST::MacroInvocData&), Location&&, Rust::AST::MacroInvocData&)
	/usr/include/c++/10/bits/invoke.h:142
0x99ed93 std::_Function_handler<Rust::AST::Fragment (Location, Rust::AST::MacroInvocData&), Rust::AST::Fragment (*)(Location, Rust::AST::MacroInvocData&)>::_M_invoke(std::_Any_data const&, Location&&, Rust::AST::MacroInvocData&)
	/usr/include/c++/10/bits/std_function.h:292
0xa3b2fe std::function<Rust::AST::Fragment (Location, Rust::AST::MacroInvocData&)>::operator()(Location, Rust::AST::MacroInvocData&) const
	/usr/include/c++/10/bits/std_function.h:622
0xa3b2fe Rust::MacroExpander::expand_invoc(Rust::AST::MacroInvocation&, bool)
	../../gcc/rust/expand/rust-macro-expand.cc:272
0xa87533 Rust::AttrVisitor::visit(Rust::AST::LetStmt&)
	../../gcc/rust/expand/rust-attribute-visitor.cc:3202
0xa87533 Rust::AttrVisitor::visit(Rust::AST::LetStmt&)
	../../gcc/rust/expand/rust-attribute-visitor.cc:3162
0xa8b89a void Rust::AttrVisitor::expand_macro_children<std::vector<std::unique_ptr<Rust::AST::Stmt, std::default_delete<Rust::AST::Stmt> >, std::allocator<std::unique_ptr<Rust::AST::Stmt, std::default_delete<Rust::AST::Stmt> > > >, std::unique_ptr<Rust::AST::Stmt, std::default_delete<Rust::AST::Stmt> > >(Rust::MacroExpander::ContextType, std::vector<std::unique_ptr<Rust::AST::Stmt, std::default_delete<Rust::AST::Stmt> >, std::allocator<std::unique_ptr<Rust::AST::Stmt, std::default_delete<Rust::AST::Stmt> > > >&, std::function<std::unique_ptr<Rust::AST::Stmt, std::default_delete<Rust::AST::Stmt> > (Rust::AST::SingleASTNode)>)
	../../gcc/rust/expand/rust-attribute-visitor.h:69
```

### Meta

- GCC commits

```
7f66312db45e3e157123108934f6cf0c71a1c235
```
(master)

```
a158a665b35dbacc14f256cad54961d4b25d3c90
```
(macro rework with eager expansion)

### LINKED ISSUE #1842: Add a regression test for empty string in `include_str` and `include_bytes`
We recently fixed the issue with empty string in these two macros. It would be good for us to introduce some tests to avoid this situation to arise again.

P-E-P:
RAIIFile constructor was accepting directory filename. This lead to unattended directory opening in some part of the code (load_file_bytes) which resulted in ice. Since RAIIFile are used for the lexer, removing the ability to open directories with RAIIFile fixes those issues and prevent future mistakes.

gcc/rust/ChangeLog:

	* lex/rust-lex.h: Add requirement for file write permission.

Signed-off-by: Pierre-Emmanuel Patry <pierre-emmanuel.patry@embecosm.com>

Fixes #1830 
Fixes #1842

### DIFF
--- a/gcc/rust/lex/rust-lex.h
+++ b/gcc/rust/lex/rust-lex.h
@@ -38,13 +38,37 @@ private:
       fclose (file);
   }
 
+  static bool allowed_filetype (const struct stat &statbuf)
+  {
+    // The file could be either
+    // - a regular file
+    // - a char device (/dev/null...)
+    return S_ISREG (statbuf.st_mode) || S_ISCHR (statbuf.st_mode);
+  }
+
 public:
   RAIIFile (const char *filename) : filename (filename)
   {
     if (strcmp (filename, "-") == 0)
-      file = stdin;
+      {
+	file = stdin;
+      }
     else
-      file = fopen (filename, "r");
+      {
+	struct stat statbuf;
+	if (!(file = fopen (filename, "r")))
+	  {
+	    return;
+	  }
+
+	if (-1 == fstat (fileno (file), &statbuf)
+	    || !allowed_filetype (statbuf))
+	  {
+	    fclose (file);
+	    file = nullptr;
+	    errno = EISDIR;
+	  }
+      }
   }
 
   /**
--- a/gcc/testsuite/rust/compile/issue-1830_bytes.rs
+++ b/gcc/testsuite/rust/compile/issue-1830_bytes.rs
@@ -1,0 +1,8 @@
+#[rustc_builtin_macro]
+macro_rules! include_bytes {
+    () => {{}};
+}
+
+fn main() {
+    include_bytes!(""); // { dg-excess-errors "Is a directory" }
+}
--- a/gcc/testsuite/rust/compile/issue-1830_str.rs
+++ b/gcc/testsuite/rust/compile/issue-1830_str.rs
@@ -1,0 +1,8 @@
+#[rustc_builtin_macro]
+macro_rules! include_str {
+    () => {{}};
+}
+
+fn main() {
+    include_str!(""); // { dg-excess-errors "Is a directory" }
+}


CohenArthur (approved):
LGTM! Thank you :)

powerboat9:
What happens if you try to include a file you don't have permission to write to?

P-E-P:
> What happens if you try to include a file you don't have permission to write to?

It should simply fail with the usual permission denied OS error message, this PR should not affect this behavior.

powerboat9:
But what if you have permission to read it but not to write to it? (ex: 744 owned by root on unix-like systems)

P-E-P:
> But what if you have permission to read it but not to write to it? (ex: 744 owned by root on unix-like systems)

I recognize it should fail. We may try to `opendir` the file instead, I believe `dirent.h` is available on all targeted platforms ? That said, I'm not entirely convinced by this approach, trying to open every file as a directory and checking if it fail feels kinda wrong.

dkm:
I'm not sure to really understand why you don't simply test if `filename` is a directory or not?

P-E-P:
> I'm not sure to really understand why you don't simply test if `filename` is a directory or not?

I was assuming `stat` was not available, hence why I tried to find other solutions.

CohenArthur:
@powerboat9 @dkm from my limited understanding it seems that:

1. `stat` did not seem to be available on OSX, since its man page indicates it is a syscall and is hence usually Linux specific. This is why @P-E-P and I decided not to go that route. It seems however, that the syscall also exists on BSD systems and would thus be available. However,
2. `stat` is not available on Windows, I'm pretty sure. We have something called `_stat` which may or may not do the same thing, and other functions with funny looking names to do the same thing. But I think this is something we need to take into account, similarly to how we make sure to handle file separators properly on both *nix and Windows systems.
2. `opendir` is not available on Windows directly, again. 

So I don't think there's an easy solution here. We did in fact not think about opening files where the user does not have write permission. This might need a little more thinking

philberty:
There is bound to be some kind of helpers for this in libiberty

philberty:
Maybe not...

CohenArthur:
That's fine then. We'll have nightstrike ping us on IRC if we do end up breaking the windows build anyways :) so let's go with `stat`

powerboat9:
It gets converted because its passed as an argument to a vector constructor - implicitly being converted to an unsigned type.

CohenArthur:
@P-E-P I added some debug prints to show the situation:

```
rust1: note: [ARTHUR]: fseek (f, 0L, SEEK_END) = 0
rust1: note: [ARTHUR]: fsize = 9223372036854775807
rust1: note: [ARTHUR]: fseek (f, 0L, SEEK_SET) = 0
```

this is just by printing the result of the first fseek, the call to ftell, and the second fseek.

```diff
diff --git a/gcc/rust/expand/rust-macro-builtins.cc b/gcc/rust/expand/rust-macro-builtins.cc
index f6ddb1a803e..936649d4d87 100644
--- a/gcc/rust/expand/rust-macro-builtins.cc
+++ b/gcc/rust/expand/rust-macro-builtins.cc
@@ -246,9 +246,13 @@ load_file_bytes (const char *filename)
     }
 
   FILE *f = file_wrap.get_raw ();
-  fseek (f, 0L, SEEK_END);
+  auto one = fseek (f, 0L, SEEK_END);
   long fsize = ftell (f);
-  fseek (f, 0L, SEEK_SET);
+  auto two = fseek (f, 0L, SEEK_SET);
+
+  rust_debug ("[ARTHUR]: fseek (f, 0L, SEEK_END) = %d", one);
+  rust_debug ("[ARTHUR]: fsize = %ld", fsize);
+  rust_debug ("[ARTHUR]: fseek (f, 0L, SEEK_SET) = %d", two);
 
   std::vector<uint8_t> buf (fsize);
 ```
 
 Note that I am using `%d` and `%ld` to indicate signed values, and that the call to `ftell` is assigned to a `long` value. This is probably due to some funky reexports within GCC or equivalent.
 
Unless I'm missing something obvious it seems to be this is not working as intended.

**Comment by powerboat9** [comment]
Huh. I suppose we have to use ```stat()``` then?

File: gcc/rust/lex/rust-lex.h [hunk 1]
```diff
@@ -38,13 +38,37 @@ struct RAIIFile
...
+	  {
+	    return;
+	  }
+
+	if (-1 == fstat (fileno (file), &statbuf)
+	    || !allowed_filetype (statbuf))
+	  {
+	    fclose (file);
+	    file = nullptr;
+	    errno = EISDIR;
```
CohenArthur, line 69: Setting errno here is IMO fine but we should maybe have a think about it. Should the RAIIFile class be responsible for emitting its own errors? As it is currently, we always check if `file_wrap.ok()` or `file_wrap.get_raw() != nullptr` before calling `rust_error_at` or `rust_fatal_error`. But I think it could be worth it to make this class a little more complex and self-contained to emit its own errors. With an optional flag to disable that behavior. But this can definitely be done later on as part of a cleanup and is fine for now :)

CohenArthur (approved):
LGTM I'm happy to merge this. Thank you!

PR #1856, opened 2023-02-14
`libproc_macro` rust interface

### LINKED ISSUE #1843: `libproc_macro` rust interface
We require a rust interface for our future `libproc_macro` library we will bind using ffi to our c++ version of the `libproc_macro`.

Here are the components required:

- [x] [Delimiter](https://doc.rust-lang.org/proc_macro/enum.Delimiter.html)
- [x] [Group](https://doc.rust-lang.org/proc_macro/struct.Group.html)
- [x] [Ident](https://doc.rust-lang.org/proc_macro/struct.Ident.html)
- [x] [LexError](https://doc.rust-lang.org/proc_macro/struct.LexError.html)
- [x] [Literal](https://doc.rust-lang.org/proc_macro/struct.Literal.html)
- [x] [Punct](https://doc.rust-lang.org/proc_macro/struct.Punct.html)
- [x] [Spacing](https://doc.rust-lang.org/proc_macro/enum.Spacing.html)
- [x] [Span](https://doc.rust-lang.org/proc_macro/struct.Span.html)
- [x] [TokenStream](https://doc.rust-lang.org/proc_macro/struct.TokenStream.html)
- [x] [TokenTree](https://doc.rust-lang.org/proc_macro/enum.TokenTree.html)
- [x] [is_available](https://doc.rust-lang.org/proc_macro/fn.is_available.html)
- [x] [token_stream](https://doc.rust-lang.org/proc_macro/token_stream/index.html)

Additional things to achieve:

- [x] Rework `Clone` trait on string owning types. We may want to do deep clone instead of copying the fields with a new alloc and `memcpy` of the internal string.
- [x] Add byte length as argument in FFI C interface ? Rust string are not null terminated.

Implement the following traits for most types that require it. We may create a new issue for this as Sync and Send traits are not working yet with gccrs. We can use the marker trait for most structures, however enumerations will require a nightly feature.

- [ ] `!Sync` for most types
- [ ] `!Send` for most types

P-E-P:
Add the rust interface types for `libproc_macro`.

Fixes #1843


P-E-P:
@CohenArthur could you please add this PR to the libproc_macro project ?

librust/proc_macro/rust/bridge/span.rs
```diff
@@ -0,0 +1,32 @@
...
+//! All methods accessing source location in rust are unstable, hence this
+//! implementation with an empty structure.
+
+#[derive(Copy, Clone, Debug)]
+#[repr(C)]
+pub struct Span {}
+
+impl Span {
+    pub fn call_site() -> Self {
+        Span {}
```
bjorn3, line 14: It is necessary to distinguish between call_site, expansion_site and mixed_site for macro hygiene purposes. In addition for spans originating from the macro input, there is macro hygiene that needs to be recorded too. I'm not sure if gccrs already implements macro hygiene, but if it doesn't, omitting it here for now shouldn't matter.

librust/proc_macro/rust/bridge/punct.rs
```diff
@@ -1,16 +1,23 @@
 use bridge::span::Span;
+use std::convert::TryInto;
+use std::ffi::c_uchar;
 use Spacing;
 
 #[repr(C)]
 #[derive(Clone, Debug)]
 pub struct Punct {
-    pub(crate) ch: char,
+    pub(crate) ch: c_uchar,
```
bjorn3, line 9: A rust char is an u32 with a range of all unicode scalar values. A C char is a u8 which is unable to represent arbitrary unicode scalar values.

bjorn3:
By the way do you want me to keep reviewing as you are implementing it or do you prefer if I wait until this is no longer a draft?

librust/proc_macro/rust/bridge/punct.rs
P-E-P: > By the way do you want me to keep reviewing as you are implementing it or do you prefer if I wait until this is no longer a draft?

To be fair, the way to achieve this still contains some dark spots. I really appreciate those early comments as they help me  to either confirm I'm going in the right direction or that the code/design is flawed.

P-E-P:
> By the way do you want me to keep reviewing as you are implementing it or do you prefer if I wait until this is no longer a draft?

To be fair, the way to achieve this still contains some dark spots. I really appreciate those early comments as they help me  to either confirm I'm going in the right direction or that the code/design is flawed.

librust/proc_macro/rust/bridge/literal.rs
```diff
@@ -0,0 +1,212 @@
...
+#[derive(Clone)]
+pub enum Unsigned {
+    Unsigned8(u8),
+    Unsigned16(u16),
+    Unsigned32(u32),
+    Unsigned64(u64),
+    // FIXME: 128 bits ffi is not safe for now
+    // https://github.com/rust-lang/rust/issues/54341
+    //
+    // Unsigned128(u128),
```
bjorn3, line 21: `Unsigned128(u64, u64)` would work, right?

P-E-P: Yes, this is what I had in my mind.

P-E-P:
I'm thinking about that extra layer of indirection. At the beginning I thought that would allow us to better separate ffi interface from the "standard" rust interface. However, the more I think about it, the less I like it (It begins to turn into a somewhat ugly and unnecessary mess with the `TokenTree` enum). What prevent us from having directly `#[repr(C)]` struct exposed ? I believe someone had another argument for this pattern but I can't remember who said that. @CohenArthur any thoughts about that ?

bjorn3:
For enums you have to keep using a private type as you can't make enum variants private. For structs directly exposing an `#[repr(C)]` struct would work for as long as all fields are private.

librust/proc_macro/rust/bridge/ident.rs
```diff
@@ -0,0 +1,59 @@
...
+#[derive(Clone, Debug)]
+pub struct Ident {
+    pub(crate) is_raw: bool,
+    pub(crate) val: *const c_char,
+}
+
+impl Ident {
+    pub fn new(string: &str, _span: Span) -> Self {
+        let string = CString::new(string).expect("Cannot convert to CString");
+        unsafe { Ident__new(string.as_ptr()) }
```
CohenArthur, line 21: I believe this causes a use-after-free:

https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.as_ptr

`as_ptr` states that the pointer is only alive for as long as the `CString` is. However, `string` goes ouf of scope and gets dropped at the end of the function, so `Ident.val` will point to freed memory.

librust/proc_macro/rust/bridge/ident.rs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    pub fn new_raw(string: &str, _span: Span) -> Self {
+        let string = CString::new(string).expect("Cannot convert to CString");
+        unsafe { Ident__new_raw(string.as_ptr()) }
```
CohenArthur, line 26: Same here

librust/proc_macro/rust/bridge/ident.rs
```diff
@@ -0,0 +1,59 @@
+use bridge::span::Span;
+use std::ffi::{c_char, CStr, CString};
```
CohenArthur, line 2: We can think about this later and this is fine for now, but eventually it'd be nice to try and stay away from `std` as much as possible for this crate. I think it'll still be some time before `gccrs` can compile `std`.

librust/proc_macro/rust/bridge/ident.rs
P-E-P: This should be fine, the C++ part should memcpy the string and set the pointer in the structure with it's own allocator. Every allocated memory will be handled by the C++ part sadly. I believe that technically it should be possible to use rust allocator if 100% of structures are allocated by rust, otherwise the allocator will get corrupted.

librust/proc_macro/rust/bridge/ident.rs
P-E-P: We will need `std` anyway, some types/functions/trait come from the standard library. Mainly `fmt`.

librust/proc_macro/rust/bridge/ident.rs
```diff
@@ -0,0 +1,68 @@
+use bridge::span::Span;
+use std::convert::TryInto;
+use std::ffi::c_uchar;
+use std::fmt;
+
+extern "C" {
+    fn Ident__new(string: *const c_uchar, len: u64) -> Ident;
```
CohenArthur, line 7: Are you not using `c_size_t` here because it's unstable?

librust/proc_macro/rust/bridge/token_stream.rs
```diff
@@ -0,0 +1,156 @@
...
+        Ok(())
+    }
+}
+
+impl FromStr for TokenStream {
+    type Err = LexError;
+    fn from_str(string: &str) -> Result<Self, LexError> {
+        let mut ts = TokenStream::new();
+        if unsafe {
+            TokenStream__from_string(
```
CohenArthur, line 139: So this will return `true` on errors?

CohenArthur (commented):
I get a few errors when trying to compile the library with `rustc` 1.49 but other than that nothing to report :) and we can fix these later down the line.

```rust
arthur@platypus ~/G/r/g/l/p/rust ((b1752289)) [1]> rustc lib.rs --crate-type=lib
error: there is no argument named `byte`
   --> bridge/literal.rs:258:39
    |
258 |                         write!(f, "\\x{byte:02x}")?;
    |                                       ^^^^^^^^^^

error: there is no argument named `ch`
   --> bridge/literal.rs:269:46
    |
269 |                     ' '..='z' => write!(f, "'{ch}'")?,
    |                                              ^^^^

error: there is no argument named `val`
   --> bridge/literal.rs:270:41
    |
270 |                     _ => write!(f, "'\\u{val:x}'")?,
    |                                         ^^^^^^^

error[E0432]: unresolved import `std::ffi::c_uchar`
 --> bridge/ident.rs:3:5
  |
3 | use std::ffi::c_uchar;
  |     ^^^^^^^^^^^^^^^^^ no `c_uchar` in `ffi`

error[E0432]: unresolved import `std::ffi::c_uchar`
 --> bridge/literal.rs:3:5
  |
3 | use std::ffi::c_uchar;
  |     ^^^^^^^^^^^^^^^^^ no `c_uchar` in `ffi`

error[E0432]: unresolved import `std::ffi::c_uchar`
 --> bridge/token_stream.rs:3:5
  |
3 | use std::ffi::c_uchar;
  |     ^^^^^^^^^^^^^^^^^ no `c_uchar` in `ffi`

error: aborting due to 6 previous errors
```

I think we should merge this soon and we'll certainly see how the implementation fares when we start using it

librust/proc_macro/rust/bridge/ident.rs
P-E-P: Yes, unstable type, hence not in 1.49. Furthermore I thought using `u64` would lead to less errors/unexpected behaviors on some funky architectures.

librust/proc_macro/rust/bridge/token_stream.rs
P-E-P: This should probably be documented. At first I expected to pass a pointer to a LexError type through ffi, on error the `LexError` would be modified and the function would return `true`. As for now `LexError` are empty, hence the simplification of the interface, and the pointer removal from the arguments.

P-E-P:
> I get a few errors when trying to compile the library with `rustc` 1.49 but other than that nothing to report :) and we can fix these later down the line.

@CohenArthur ffi types were officially stabilised in [rust 1.64](https://doc.rust-lang.org/src/std/ffi/mod.rs.html#157), how should we proceed for those error ?

CohenArthur:
> > I get a few errors when trying to compile the library with `rustc` 1.49 but other than that nothing to report :) and we can fix these later down the line.
> 
> @CohenArthur ffi types were officially stabilised in [rust 1.64](https://doc.rust-lang.org/src/std/ffi/mod.rs.html#157), how should we proceed for those error ?

Let's not worry about them for now - it's still a ways before we can compile this crate anyways. I'll merge this now so we can start experimenting with it further :)

Issue #1808, opened 2023-02-01
Invalid pattern parsing in nested "or" pattern


CohenArthur:
I tried this code:

```rust
struct C(char);

fn main() {
    let c = C('a');
    match c {
        C('a' | 'b') => {}
        _ => {}
    }
}

```

I expected to see this happen: no error

Instead, this happened: 
```rust
test.rs:6:15: error: unexpected token ‘|’ in tuple struct items
    6 |         C('a' | 'b') => {}
      |        
```
  
### Meta

- What version of Rust GCC were you using, 7f222689dcb

ArberSephirotheca:
Hi, I am interested in this issue and currently looking at the related code, could you please assign me if possible?

P-E-P:
Fixed by #1878

PR #1862, opened 2023-02-15
Add AST::AltPattern class

powerboat9:
Adds AST::AltPattern to represent alternate patterns (ex: ```12 | 13```).

Addresses #1808

### DIFF
--- a/gcc/rust/ast/rust-ast-dump.cc
+++ b/gcc/rust/ast/rust-ast-dump.cc
@@ -1678,6 +1678,10 @@ void
 Dump::visit (SlicePattern &)
 {}
 
+void
+Dump::visit (AltPattern &)
+{}
+
 // rust-stmt.h
 void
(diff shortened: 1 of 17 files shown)


powerboat9:
The ```clang-format``` github action is failing because of lines I didn't change.

CohenArthur (approved):
This is really, really good work @powerboat9 thanks a lot :) I think the class is in a good state and we'll need to be careful when changing the parser to accomodate for it.

The grammar for patterns is actually quite funny as it allows a leading `|`: https://doc.rust-lang.org/reference/patterns.html

All in all great work!

CohenArthur:
Regarding the clang-format check failing... yeah. That has happened to me in the past. Please simply format the offending file in your commit. We need to look at upgrading our version since this one is a bit dodgy at times. Sorry about that :/

powerboat9:
Fixed formatting

PR #1863, opened 2023-02-15
Add trailing newline

powerboat9:
gcc/rust/ChangeLog:

	* checks/errors/rust-feature-gate.h: Add trailing newline before EOF.

I noticed this while preparing #1862, but figured it would be better to address it separately.


CohenArthur (approved):
Thanks :)

PR #1871, opened 2023-02-16
builtins: Implement eager expansion for include_str!()

### LINKED ISSUE #1804: Implement eager expansion for `include_str!()`

omkar-mohanty:
Fixes #1804

### DIFF
--- a/gcc/rust/expand/rust-macro-builtins.cc
+++ b/gcc/rust/expand/rust-macro-builtins.cc
@@ -16,42 +16,24 @@
 // along with GCC; see the file COPYING3.  If not see
 // <http://www.gnu.org/licenses/>.
 
+#include "rust-system.h"
 #include "rust-macro-builtins.h"
+#include "rust-ast-fragment.h"
 #include "rust-ast.h"
-#include "rust-diagnostics.h"
-#include "rust-expr.h"
-#include "rust-session-manager.h"
-#include "rust-macro-invoc-lexer.h"
-#include "rust-lex.h"
-#include "rust-parse.h"
-#include "rust-early-name-resolver.h"
 #include "rust-attribute-visitor.h"
+#include "rust-diagnostics.h"
+#include "rust-early-name-resolver.h"
+#include "rust-expr.h"
+#include "rust-lex.h"
+#include "rust-macro-invoc-lexer.h"
+#include "rust-macro.h"
+#include "rust-parse.h"
+#include "rust-session-manager.h"
 
 namespace Rust {
 namespace {
-
-/**
- * Shorthand function for creating unique_ptr tokens
- */
-static std::unique_ptr<AST::Token>
-make_token (const TokenPtr tok)
-{
-  return std::unique_ptr<AST::Token> (new AST::Token (tok));
-}
-
-std::unique_ptr<AST::Expr>
-make_string (Location locus, std::string value)
-{
-  return std::unique_ptr<AST::Expr> (
-    new AST::LiteralExpr (value, AST::Literal::STRING,
-			  PrimitiveCoreType::CORETYPE_STR, {}, locus));
-}
-
-// TODO: Is this correct?
-static AST::Fragment
-make_eager_builtin_invocation (
-  AST::BuiltinMacro kind, Location locus, AST::DelimTokenTree arguments,
-  std::vector<std::unique_ptr<AST::MacroInvocation>> &&pending_invocations)
+std::string
+make_macro_path_str (AST::BuiltinMacro kind)
 {
   std::string path_str;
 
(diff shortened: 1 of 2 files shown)


gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -376,6 +390,29 @@ AST::Fragment
...
+
+  auto last_token_id = macro_end_token (invoc_token_tree, parser);
+
+  auto expanded_expr = try_expand_many_expr (parser, last_token_id,
+					     invoc.get_expander (), has_error);
+
+  if (has_error)
+    {
+      AST::Fragment::create_error ();
+    }
```
CohenArthur, line 407:
```suggestion
  if (has_error)
      return AST::Fragment::create_error ();
```

gcc/testsuite/rust/compile/builtin_macro_eager3.rs
```diff
@@ -0,0 +1,21 @@
...
+
+macro_rules! foo {
+    () => {
+        "bar.txt"
+    };
+}
+
+fn main () {
+  include_str!(foo!()); // ok
+  include_str!(foo!(), bar!()); // ok
```
CohenArthur, line 20: This is not a proper invocation of the macro sadly :) error from `rustc`:

```rust
error: include_str! takes 1 argument
  --> test.rs:22:5
   |
22 |     include_str!(foo!(), bar!()); // ok
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the macro `include_str` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Also, I can't find any file named `bar.txt` or `foo.txt` in the repo, so I think you might have forgotten to push them?

gcc/rust/expand/rust-macro-builtins.cc
CohenArthur, line 407: `include_str_handler` calls into `parse_single_string_literal`, which already takes care of lexing and parsing the tokens. So by adding another parser here you're effectively doing the same work twice. I think what we should do instead is look at how to modify `parse_single_string_literal` to allow parsing macro invocations. If you look at the function:

```c++
std::unique_ptr<AST::LiteralExpr>
parse_single_string_literal (AST::DelimTokenTree &invoc_token_tree,
                             Location invoc_locus, MacroExpander *expander)
{
  MacroInvocLexer lex (invoc_token_tree.to_token_stream ());
  Parser<MacroInvocLexer> parser (lex);

  auto last_token_id = macro_end_token (invoc_token_tree, parser);

  std::unique_ptr<AST::LiteralExpr> lit_expr = nullptr;

  if (parser.peek_current_token ()->get_id () == STRING_LITERAL)
    {
      lit_expr = parser.parse_literal_expr ();
      parser.maybe_skip_token (COMMA);
      if (parser.peek_current_token ()->get_id () != last_token_id)
        {
          lit_expr = nullptr;
          rust_error_at (invoc_locus, "macro takes 1 argument");
        }
    }
  else if (parser.peek_current_token ()->get_id () == last_token_id)
    rust_error_at (invoc_locus, "macro takes 1 argument");
  else
    rust_error_at (invoc_locus, "argument must be a string literal");

  parser.skip_token (last_token_id);

  return lit_expr;
}
```

you can see that it calls into `parse_literal_expr()`. We can discuss in more details on Zulip on how to go about making these modifications if you'd like, but I think that the first step would be to change its return type from `std::unique_ptr<AST::LiteralExpr>` to `std::unique_ptr<AST::Expr>` and trying to parse a literal expression or a macro invocation.

So I think the issue is a bit more complex than it looks but it's also a great way to learn about our parser and more of the codebase :)

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -17,41 +17,24 @@
...
-#include "rust-macro-invoc-lexer.h"
 #include "rust-lex.h"
+#include "rust-macro-invoc-lexer.h"
+#include "rust-macro.h"
 #include "rust-parse.h"
-#include "rust-early-name-resolver.h"
-#include "rust-attribute-visitor.h"
+#include "rust-session-manager.h"
+#include <memory>
+#include <utility>
```
CohenArthur, line 32: GCC does not like when you include these system headers and you should not have to. You can include "rust-system.h" if you get include failures

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -219,11 +245,35 @@ parse_single_string_literal (AST::DelimTokenTree &invoc_token_tree,
   else if (parser.peek_current_token ()->get_id () == last_token_id)
     rust_error_at (invoc_locus, "macro takes 1 argument");
   else
-    rust_error_at (invoc_locus, "argument must be a string literal");
+    {
+      bool has_error = false;
+
+      auto expanded_expr
+	= try_expand_many_expr (parser, last_token_id, expander, has_error);
```
CohenArthur, line 252: This function tries and parse multiple expressions from an invocation. Think something like `concat!(a, b, 15, "hey")`. This function would parse and return a vector containing `a`, `b`, `15` and `"hey"`. Here, we need to make sure we're only parsing one expression, which could be a literal expression or a macro invocation. So you need to add some more checks after calling this:
```suggestion
      auto expanded_exprs
	= try_expand_many_expr (parser, last_token_id, expander, has_error);
      if (expanded_exprs.size() > 1)
          rust_error_at(invoc_locus, "expected only one argument, got %lu", expanded_exprs.size());
          
      // then check for literal expr or macro invocation etc
```

but I think it would be even simpler to not call into `try_expand_many_expr`. If you look above in the function, we call `parser.parse_literal_expr()`. You could add a call to `parser.parse_macro_invocation({})` and see which one returns a null pointer.

CohenArthur:
@omkar-mohanty I think this PR is also getting too big. You don't need to return a new macro invocation in `assert_handler` or other handlers. Those are separate issues. You can simply change that to something like `rust_assert(lit_expr->is_literal())` and then static cast

tschwinge:
I suppose the discussion around <https://gcc-rust.zulipchat.com/#narrow/stream/266897-general/topic/GSoC.202023/near/329455766> relates to this here, and in particular why we get `FAIL: rust/compile/builtin_macro_eager3.rs (test for excess errors)` from CI testing here, when local testing does `PASS`.

In <https://github.com/Rust-GCC/gccrs/actions/runs/4244169649/jobs/7377900097#step:2:620> you see that CI testing is not actually testing your commit 501876ec5785ac63f95e560316d9e0fb4a3ab698 "Implement and test include_str eager expansion", but instead `HEAD is now at 7a0b07f46 Merge 501876ec5785ac63f95e560316d9e0fb4a3ab698 into 1145917c6b82ed475b84ae0dd6de241bd018721d`, that is, yours merged with current GCC/Rust master branch commit 1145917c6b82ed475b84ae0dd6de241bd018721d.  (Yes, that it quite non-obvious -- I didn't know that until now, either!)

As it happens, in GCC/Rust master branch a new check has recently been added, which your test case runs into (expectedly, I think), and thus the `excess errors`.  So: merge your branch with current GCC/Rust master branch, re-run your testing locally, and fix things up.  :-)

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -17,41 +17,23 @@
...
-#include "rust-session-manager.h"
-#include "rust-macro-invoc-lexer.h"
 #include "rust-lex.h"
+#include "rust-macro-invoc-lexer.h"
+#include "rust-macro.h"
 #include "rust-parse.h"
-#include "rust-early-name-resolver.h"
-#include "rust-attribute-visitor.h"
+#include "rust-session-manager.h"
+#include "rust-system.h"
```
philberty, line 31: rust-system.h should be the first include

CohenArthur (approved):
Thank you :) this looks good to me! I'll spend some time this week refactoring that whole file as it's getting quite messy, so thank you for cleaning it up a little already!

omkar-mohanty:
> Thank you :) this looks good to me! I'll spend some time this week refactoring that whole file as it's getting quite messy, so thank you for cleaning it up a little already!

Thanks for the review!

PR #1878, opened 2023-02-16
Parse AltPattern

powerboat9:
Addresses #1808


CohenArthur (commented):
This is great :) Can you please add testcases such as the one in #1808 with `{ dg-additional-options "-fsyntax-only" }`? This way we can ensure there are no regressions later on.

It'd be nice also to have simpler test-cases and add them here

powerboat9:
Added test case.

CohenArthur (approved):
Thanks!

PR #1879, opened 2023-02-16
Unsafe block parsing fix

### LINKED ISSUE #1422: Parsing error when parsing unsafe blocks in macro transcribers
I tried this code:

```rust
// { dg-output "1\n2\nNaN\n3\n" }

macro_rules! print_num {
    ($l:literal) => {
        unsafe { printf("%d\n\0" as *const str as *const i8, $l); }
    };
}

extern "C" {
    fn printf(s: *const i8, ...);
}

// Check to make sure that expanding macros does not break the flow of calls
fn main() -> i32 {
    print_num!(1);
    print_num!(2);

    unsafe {
        printf("NaN\n\0" as *const str as *const i8);
    }

    print_num!(3);

    0
}

```

Instead, this happened:
```rust
gcc/testsuite/rust/execute/torture/macros22.rs:5:16: error: unexpected token ‘{’ in some sort of unsafe production
    5 |         unsafe { printf("%d\n\0" as *const str as *const i8, $l); }
      |                ^
```

powerboat9:
Fixes #1422 and removes a ```FIXME```.

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -6131,7 +6131,15 @@ Parser<ManagedTokenSource>::parse_stmt (ParseRestrictions restrictions)
       /* if any of these (should be all possible VisItem prefixes), parse a
        * VisItem can't parse item because would require reparsing outer
        * attributes */
-      return parse_vis_item (std::move (outer_attrs));
+      // may also be unsafe block
+      if (lexer.peek_token (1)->get_id () == LEFT_CURLY)
+	{
+	  return parse_expr_stmt (std::move (outer_attrs), restrictions);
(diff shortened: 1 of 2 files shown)


CohenArthur:
Oh @powerboat9 before I merge this: I just checked and your first commit does not pass the testsuite. Can you please rebase your two commits into a single one?

powerboat9:
I just checked the first commit, and it works on my computer.tm. What test does it fail?

CohenArthur:
> I just checked the first commit, and it works on my machinetm. What test does it fail?

Uh. Sorry about the trouble then. I tested it again on my laptop and the tests pass for both commits. I must have had some uncommitted test cases. Sorry for this!

Issue #1864, opened 2023-02-15
Upstream GCC broken for --target=pdp11-aout


jbglaw:
Hi!

Testing revealed that a rust-enabled upstream GCC build (at g:06ca0c9abb260266b688e2c2154c72214bb47076) for --target=pdp11-aout breaks:
```
../gcc/configure '--with-pkgversion=basepoints/gcc-13-5954-g06ca0c9abb2, built at 1676413490' --prefix=/var/lib/laminar/run/gcc-pdp11-aout/57/toolchain-install --enable-werror-always --enable-languages=all --disable-gcov --disable-shared --disable-threads --target=pdp11-aout --without-headers
[...]
make V=1 all-gcc
[...]
[all 2023-02-15 01:32:13] /usr/lib/gcc-snapshot/bin/g++  -fno-PIE -c  -DIN_GCC_FRONTEND -g -O2   -DIN_GCC  -DCROSS_DIRECTORY_STRUCTURE   -fno-exceptions -fno-rtti -fasynchronous-unwind-tables -W -Wall -Wno-narrowing -Wwrite-strings -Wcast-qual -Wmissing-format-attribute -Wconditionally-supported -Woverloaded-virtual -pedantic -Wno-long-long -Wno-variadic-macros -Wno-overlength-strings -Werror -Wno-unused-parameter -fno-common  -DHAVE_CONFIG_H -I. -Irust -I../../gcc/gcc -I../../gcc/gcc/rust -I../../gcc/gcc/../include  -I../../gcc/gcc/../libcpp/include -I../../gcc/gcc/../libcody  -I../../gcc/gcc/../libdecnumber -I../../gcc/gcc/../libdecnumber/dpd -I../libdecnumber -I../../gcc/gcc/../libbacktrace   -o rust/rust-tree.o -MT rust/rust-tree.o -MMD -MP -MF rust/
[...]

Issue #1819, opened 2023-02-02
New build failure on upstream GCC for --target=powerpc-ibm-aix7.{1,2}


jbglaw:
On upstream GCC master, I see new errors for the `powerpc-ibm-aix7.1` and `powerpc-ibm-aix7.2` targets starting with the e66fec8e6ba35edf01f86c2bf6514109aba4ded2 commit (gccrs: const folding port):

```shell
.../configure --enable-werror-always --enable-languages=all --disable-gcov --disable-shared --disable-threads --target=powerpc-ibm-aix7.1 --without-headers
[...]
make V=1 all-gcc
[...]
[...]

PR #1885, opened 2023-02-18
Merge upstream

tschwinge:


tschwinge:
By the way: this must of course not be rebased by GitHub merge queue, but has to become a proper Git merge.

powerboat9:
@tschwinge Nice work, how did you figure out we needed ```tm_p.h```?

tschwinge:
> how did you figure out we needed `tm_p.h`?

Basically like this: the #1819 `error: 'rs6000_xcoff_strip_dollar' was not declared` needs `[source]/gcc/config/rs6000/rs6000-protos.h` and the #1864 `error: 'pdp11_gen_int_label' was not declared` needs `[source]/gcc/config/pdp11/pdp11-protos.h`, so there should exist some *indirection header file* that does `#include "[target]-protos.h"`.  That *indirection header file* probably is generated at build time (because then we can easily expand `[target]`).  Then I found `[build]/gcc/tm_p.h` to be doing exactly that, and confirmed via other existing `#include "tm_p.h"` usage that this is a correct thing to do.

CohenArthur:
Looks good to me :) thank you @tschwinge. @jbglaw we'll merge this and if your github remote build passes we'll close both issues :)

jbglaw:
Just did a pull, wasn't yet there. Once I realize it's there (or you drop me a comment?) I'll schedule a new round for my gccrs builds. It' all in place, so it's easy. ^^

jbglaw:
Hi @CohenArthur , just updated the other two tickets: I got confirmation that the three builds (pdp11 and the two powerpc AIX versions) had a successful run.

PR #1888, opened 2023-02-20
Add AltPattern HIR node

powerboat9:
Adds a HIR node for alternate patterns

### DIFF
--- a/gcc/rust/hir/tree/rust-hir-pattern.h
+++ b/gcc/rust/hir/tree/rust-hir-pattern.h
@@ -1300,6 +1300,73 @@ protected:
   }
 };
 
+// HIR node for alternative patterns
+class AltPattern : public Pattern
+{
+  std::vector<std::unique_ptr<Pattern>> alts;
+  Location locus;
+  Analysis::NodeMapping mappings;
+
+public:
+public:
+  std::string as_string () const override;
+
+  AltPattern (Analysis::NodeMapping mappings,
+	      std::vector<std::unique_ptr<Pattern>> alts, Location locus)
+    : alts (std::move (alts)), locus (locus), mappings (mappings)
+  {}
+
+  // Copy constructor with vector clone
+  AltPattern (AltPattern const &other)
+    : locus (other.locus), mappings (other.mappings)
+  {
+    alts.reserve (other.alts.size ());
+    for (const auto &e : other.alts)
+      alts.push_back (e->clone_pattern ());
+  }
+
+  // Overloaded assignment operator to vector clone
+  AltPattern &operator= (AltPattern const &other)
+  {
+    locus = other.locus;
+    mappings = other.mappings;
+
+    alts.clear ();
+    alts.reserve (other.alts.size ());
+    for (const auto &e : other.alts)
+      alts.push_back (e->clone_pattern ());
+
+    return *this;
+  }
+
+  // move constructors
+  AltPattern (AltPattern &&other) = default;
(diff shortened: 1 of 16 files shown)


gcc/rust/hir/tree/rust-hir-pattern.h
```diff
@@ -1295,6 +1293,72 @@ class SlicePattern : public Pattern
...
+  AltPattern (Analysis::NodeMapping mappings,
+	      std::vector<std::unique_ptr<Pattern>> alts, Location locus)
+    : alts (std::move (alts)), locus (locus), mappings (mappings)
+  {}
+
+  // Copy constructor with vector clone
+  AltPattern (AltPattern const &other)
+    : locus (other.locus), mappings (other.mappings)
+  {
+    alts.reserve (other.alts.size ());
```
CohenArthur, line 1316: Missing a call to `alt.clear()` otherwise you're going to run into some issues if you copy-assign to an `AltPattern` multiple times. This is a problem in some of our other classes as well so no worries
```suggestion
    alts.clear();
    alts.reserve (other.alts.size ());
```

gcc/rust/hir/tree/rust-hir-pattern.h
```diff
@@ -1295,6 +1293,72 @@ class SlicePattern : public Pattern
...
+      alts.push_back (e->clone_pattern ());
+  }
+
+  // Overloaded assignment operator to vector clone
+  AltPattern &operator= (AltPattern const &other)
+  {
+    locus = other.locus;
+    mappings = other.mappings;
+
+    alts.reserve (other.alts.size ());
```
CohenArthur, line 1327: Same here

gcc/rust/typecheck/rust-hir-type-check-pattern.cc
```diff
@@ -407,5 +407,12 @@ TypeCheckPattern::visit (HIR::SlicePattern &)
   gcc_unreachable ();
 }
 
+void
+TypeCheckPattern::visit (HIR::AltPattern &)
+{
+  // TODO
+  gcc_unreachable ();
```
CohenArthur, line 414: I think `rust_sorry_at` would be nice here as well :)

CohenArthur (commented):
Leaving to @philberty to merge but this looks good to me :) only tiny nitpicks. Thank you for working on this!

philberty (changes requested):
LGTM

 - I think for this PR it should be squashed into 1 commit.
 - I would also like the commit message to contain a message mentioning the AltPattern of:
 
 ```
 match x {
    0 | 1 => println!("it's zero or one!"),
}
 ```
 
 It would really help me when adding new HIR nodes what its going to represent.

gcc/rust/hir/tree/rust-hir-pattern.h
powerboat9: That's a copy constructor, so ```alts``` should have just been initialized as an empty vector.

gcc/rust/hir/tree/rust-hir-pattern.h
powerboat9: This is a bug, I'll fix it.

powerboat9:
This should be fixed up now, although I'm not sure if I should squash it all into one comment at this point.

powerboat9:
@CohenArthur All unrelated changes have now been split off.

philberty (approved):
LGTM

Issue #1705, opened 2022-12-15
Tracking issue for updating upstream's gccrs with this gccrs


CohenArthur:
Now that the first set of patches has been pushed, we need to update GCC's master branch with the current state of the compiler.

I want this issue to serve as an infodump and overall discussion on how to achieve this:
1. In case we need to do it again eventually
2. To discuss better ways of doing so
3. For transparency :)

The cutoff for what was pushed to GCC's master seems to have happened around #1497, commit 36a9255, on the 23rd of August 2022.

At present (2022-12-15), there are more than 5000 commits between #1497 and master. This is due to the multiple merges from GCC upstream that we've done in the last few months. I approximate the amount of commits we need to take care of (commits to gccrs specificially) to ~300. I got the list of commits thanks to these commands:

```sh
arthur@platypus ~/G/r/gccrs (master)> git log 36a9255..HEAD --format=%H -- gcc/rust/ > all_commits.old
arthur@platypus ~/G/r/gccrs (master)> git log 36a9255..HEAD --format=%H -- gcc/testsuite/rust >> all_commits.old
arthur@platypus ~/G/r/gccrs (master)> cat all_commits.old | uniq > all_commits
```

better way (thanks @tschwinge)
> Just use `git log 36a9255..HEAD --format=%H -- gcc/rust/ gcc/testsuite/rust/ > all_commits`.
> And, given git remote upstream git+ssh://gcc.gnu.org/git/gcc.git, further restrict to `36a9255..HEAD ^upstream/master` to filter out all GCC upstream master branch commits that we're not interested in tracking here.

I'm sure there are better ways to go about this, but it's fine for now :)

The main issues for updating are the following:

- [ ] Commits since 2022-08-23 need Changelog entries

If you'd like to write Changelogs for some commits, please pick some unticked ones in the list below. If a commit is ticked, then it means the Changelog has already been written or is being written.

Commits up to 72ee612ef41a45d56e03982ba04562949bd5778d have been upstreamed
New commits have been pushed on `first-half-update`. The name still sucks but I don't want to break hashes or anything by changing the branch's name.

These commits are as follows.

- [x] 360a69c4383968deb77bc0c40780dfb3772f7b12 #1721
- [x] 11705326a5f81e9aa9066a9edf83609b65aa4f96 #1721
- [x] ef114f8524a48fe2061225df8606eb3b640e8678 #1745
[...]

PR #1889, opened 2023-02-21
[CL] missing commits from after #1705

CohenArthur:
There were 22 commit Changelogs missing which should now be fixed.


CohenArthur:
How should we deal with 38adbd453ea76814ae8e809b92bdcd6c960dc865 @tschwinge? This isn't really an `Update copyright year` commit since it only concerns `gcc/rust/` files. So should I add a Changelog to it and a `gccrs: ` prefix? What do you think?

CohenArthur:
Also, I don't think [2dd73cf](https://github.com/Rust-GCC/gccrs/pull/1889/commits/2dd73cf151ec8ec1a3ee16cf2cf695bdf321288b) should be in this series... :thinking:

CohenArthur:
> Also, I don't think [2dd73cf](https://github.com/Rust-GCC/gccrs/pull/1889/commits/2dd73cf151ec8ec1a3ee16cf2cf695bdf321288b) should be in this series... thinking

Ah, right. This is because we haven't updated `gcc-patch-dev` yet

tschwinge:
> How should we deal with [38adbd4](https://github.com/Rust-GCC/gccrs/commit/38adbd453ea76814ae8e809b92bdcd6c960dc865) @tschwinge? This isn't really an `Update copyright year` commit since it only concerns `gcc/rust/` files. So should I add a Changelog to it and a `gccrs: ` prefix? What do you think?

Why shouldn't it be just `Update copyright years.` anyway?

But, there's some other problem with that commit: it contains unrelated changes -- how did that happen?  :-O

Ideally, as discussed on IRC yesterday, the individual `Update copyright years.` should (have) be(en) moved into the individual upstreaming commits.  (..., but not sure if any remain, that have not yet been upstreamed.)

philberty (approved):
LGTM

PR #1913, opened 2023-02-22
librust: Add libproc_macro and build system

### LINKED ISSUE #1891: Add C++ internal library to build system
We need to add the `libproc_macro`'s C++ internal library to GCC's build system and figure out how to tie it all together. Let's not focus on the Rust interface for now, but on the internal C++ library which the interface will interact with. If any GCC gurus have tips on how to go about adding a new library to the build system please shout :)

P-E-P:
Add some dummy files in libproc_macro along with it's build system.

ChangeLog:

	* librust/Makefile.am: New file.
	* librust/autogen.sh: New file.
	* librust/configure.ac: New file.
	* librust/libproc_macro/Makefile.am: New file.
	* librust/libproc_macro/proc_macro.cc: New file.
	* librust/libproc_macro/proc_macro.h: New file.

Fixes #1891


tschwinge:
I suppose this is mostly ["cargo-culted"](https://en.wikipedia.org/wiki/Cargo_cult_programming) from other GCC libraries?  (But wow, so much boilerplate code -- but of course not blaming you for that; maybe certain things could be cleaned up generally in GCC as a separate task...)

You'll then also need to modify the top-level build system (top-level `configure.ac`, `Makefile.def`) and supposedly add `target_libs=[...]` in `gcc/rust/config-lang.in` to have this be built.

Let me know if you'd like to get reviewed anything specifically.

Let me know if you need help with (re)generating any files or GNU Autotools stuff generally.

P-E-P:
> I suppose this is mostly ["cargo-culted"](https://en.wikipedia.org/wiki/Cargo_cult_programming) from other GCC libraries? (But wow, so much boilerplate code -- but of course not blaming you for that; maybe certain things could be cleaned up generally in GCC as a separate task...)
> 
> You'll then also need to modify the top-level build system (top-level `configure.ac`, `Makefile.def`) and supposedly add `target_libs=[...]` in `gcc/rust/config-lang.in` to have this be built.
> 
> Let me know if you'd like to get reviewed anything specifically.
> 
> Let me know if you need help with (re)generating any files or GNU Autotools stuff generally.

For now I'm mostly trying to figure out how to compile a simple file with GCC build system. I think I should be able to compile it as a standalone as a first step, I'll then add the library as a module by modifying the files at the top level. I've already fiddled around with automake in the past, but I'm indeed taking a look at the other libraries in an attempt to get it working. There seems to be so much "fixes" and workaround in the other libraries.

This is some kind of "first draft" I've pushed because I had a configure step somehow working. I'm struggling a bit to separate what to keep and what I should throw away. I initially thought it would be better to copy what was existing, throwing every non essential things until it doesn't work. Alas there are so many things I don't know why they've been put there and whether they are still required. Maybe building this part from scratch would be a better approach ? Do you think following a standard autotool build methodology would be suitable ?

tschwinge:
> I'm indeed taking a look at the other libraries in an attempt to get it working. There seems to be so much "fixes" and workaround in the other libraries.

Right...  :-\

> This is some kind of "first draft" I've pushed because I had a configure step somehow working. I'm struggling a bit to separate what to keep and what I should throw away. I initially thought it would be better to copy what was existing, throwing every non essential things until it doesn't work anymore. Alas there are so many things I don't know why they've been put there and whether they are still required. Maybe building this part from scratch would be a better approach ? Do you think following a standard autotool build methodology would be more suitable ?

Hmm, I'm not sure -- I kind of expect some GCC-specifics are necessary to fit libraries into GCC's build process (bootstrap vs. non-bootstrap maybe), build vs. host vs. target configuration, several multilibs builds, etc.

Starting with an existing library might be easier.  Maybe have a look at `libbacktrace` or `zlib`, which are built for both host and target?

Later, once we get to that point, `libsanitizer` (but I'm sure there are others) might serve as an example of a large body of code imported from elsewhere, with GCC-local GNU Autotools build system.

> > Let me know if you need help with (re)generating any files or GNU Autotools stuff generally.
> 
> I don't expect any problem on this part. Did you have a specific problem in mind?

Just using the expected (fixed) versions of GNU Autotools; see <https://gcc.gnu.org/wiki/Regenerating_GCC_Configuration#Regenerating_All_GNU_Autotools_Files>, for example.

P-E-P:
@tschwinge I've something that compiles as a standalone, I've done some modifications at the top level directory (mainly add a target module as well as an host module) and now come the time I regenerate using autotools. I've followed the link you kindly provided (great resource, thank you) but there is a version mismatch. The document talk everywhere about autoconf 2.59 and even provides an download link through ftp, but the code block seems to mention autoconf 2.69. Should I continue with version 2.59 or is that outdated ?

tschwinge:
@P-E-P, that page (as so many others) seems largely "historic" ;-) -- I really only keep the one section: "Regenerating All GNU Autotools Files", <https://gcc.gnu.org/wiki/Regenerating_GCC_Configuration#Regenerating_All_GNU_Autotools_Files>, up to date, for my own use.

That means, use GNU Autoconf 2.69, GNU Automake 1.15.1.

P-E-P:
> That means, use GNU Autoconf 2.69, GNU Automake 1.15.1.

After several attempts I've been able to generate file that actually compile gcc correctly (first tries made the compilation hang indefinitely).

I'm now trying to fix the module description, I thought that without any disable option it would have been compiled by default but this is not the case. I believe I now have to describe a dependency between the rust frontend and the module but that's a little strange because the module was declared as an `host` module as well as a `target` module, so it should be compiled anyway.

P-E-P:
Finally got it! All this time just for a small mistake! I can't believe I lost so many time on this. Well at least now that I now that I've gone through the build system in detail I'm confident with it. The regeneration was at fault, not the configuration. I did not realise the `Makefile.def` did not regenerate the `Makefile.in`, hence the missing rules.

P-E-P:
@tschwinge Should I squash all those commits together ?

CohenArthur (approved):
Looks good to me but I have absolutely zero knowledge of the GCC build system :) I think as long as you have a commit regenerating all generated files that's separate from the commit(s) adding the new library and build-system, it's okay. Waiting for @tschwinge's input but I'm in favor of merging this so we can keep going on libproc_macro :D

PR #1778, opened 2023-01-28
Moved operator.h to util/rust-operators.h.

00AR:
Signed-off-by: Abdul Rafey <abdulrafeyq@gmail.com>

Refactored operator.h. #1768 
Changes shown in rust-operators.h are done by clang-formater.

PR #1914, opened 2023-02-23
[WIP] Create ClonePtr, a smart pointer

powerboat9:
Backed by a std::unique_ptr<T>, a ClonePtr<T> is intended to handle pointers to data which can be cloned with a 'clone_*'-like method (a common pattern in the gccrs codebase). This should reduce the need for custom copy constructors/assignment operators.

I'm not quite sure how to integrate ```rust-smart-ptr.h``` with the build system, so this is marked WIP.


powerboat9:
Looking at #1778, it seems like I don't need to modify any build system specific files to ensure Makefiles track dependencies correctly.

CohenArthur (commented):
So this wraps a `std::unique_ptr<T>`, meaning to construct a `ClonePtr` we have to do something like
```c++
ClonePtr<AST::Expr>(std::unique_ptr<AST::Expr>(AST::LiteralExpr(...)))
```
?

Also, this is going to require significant refactoring, and I don't know how much it will improve our situation. Our AST and HIR are pretty much set, there aren't a lot of classes we need to add to them still, and certainly not a lot of base abstract classes where these `clone_*` methods are declared

powerboat9:
> So this wraps a `std::unique_ptr<T>`, meaning to construct a `ClonePtr` we have to do something like

I'll have to add another constructor, I thought ```unique_ptr``` had an implicit constructor, but it [looks](https://en.cppreference.com/w/cpp/memory/unique_ptr/unique_ptr) like it doesn't.

>  Also, this is going to require significant refactoring, and I don't know how much it will improve our situation. Our AST and HIR are pretty much set, there aren't a lot of classes we need to add to them still, and certainly not a lot of base abstract classes where these clone_* methods are declared

This is intended more for internal use by AST and HIR nodes, not really for use outside of them. I figured that any further changes to utilize ```ClonePtr``` would modify the public interfaces of AST or HIR nodes as little as possible. If we have to fix https://github.com/Rust-GCC/gccrs/pull/1888#discussion_r1112855546, adding a smart pointer and removing a lot of boilerplate seems like a neat solution.

powerboat9:
Found https://github.com/jbcoe/polymorphic_value which is similar -- although its proposed for C20

powerboat9:
Also found https://github.com/Rust-GCC/gccrs/pull/318

CohenArthur:
I'm still not convinced this would make reading the code and reasoning about the code easier :/ I kept this PR in mind since you posted it and I don't think I have seen another PR where having a `ClonePtr` type would have made things easier or better. As I had said, the AST and HIR are unlikely to change a lot now, and adding some more complexity on top of it doesn't sit right with me. The only annoying thing about defining `clone_*` functions is when definining new nodes in these two represenations, which seldom happens

philberty:
Sorry @powerboat9 I am going to close this PR I think when we get libcore working and we have a solid base to the front-end we will be more open to this type of refactoring and high level changes.

PR #1915, opened 2023-02-23
cfg!() eager expansion

### LINKED ISSUE #1792: Implement eager expansion for `cfg!()`

ArberSephirotheca:
Fixed #1792 

Signed-off-by: Zheyuan Chen sephirotheca17@gmail.com
```
gcc/ChangeLog/
* rust/expand/rust-macro-builtins.cc: modify cfg_handler() to expand cfg! eagerly`
* testsuite/rust/compile/builtin_macro_eager4.rs: add test for cfg! macro`
```

**Question**: If we have something like `cfg!(foo!(), bar!())`, which has multiple arguments, do we need to return error, or just simply return the eager expansion. 
I tested `env` macro with `env!(a!(), a!(), a!())`, and gccrs does not emit any error(it will definitely emit error in rustc as `env!` only accepts 1 or 2 arugments ). So I assume we don't check the correctness(for now at least) for eager expansion when we have macro as argument?

### DIFF
--- a/gcc/testsuite/rust/compile/builtin_macro_eager4.rs
+++ b/gcc/testsuite/rust/compile/builtin_macro_eager4.rs
@@ -1,0 +1,22 @@
+#![feature(rustc_attrs)]
+
+#[rustc_builtin_macro]
+macro_rules! cfg {
+  () => {{}};
+}
+
+macro_rules! a{
+  () => {"test"};
+}
+
+
+fn main () {
+  let os = "linux";
+  cfg!(); // { dg-error "cfg-pattern required" }
+  cfg!(1); // { dg-error "expected identifier" }
+  cfg!(windows); // shoud not error
+  cfg!(target_os = "linux"); //should not error
+  cfg!(a!()); // {dg-error "expected 1 cfg-pattern"}
+  cfg!(a!(), windows); // {dg-error "expected 1 cfg-pattern"}
+
+}
--- a/gcc/rust/expand/rust-macro-builtins.cc
+++ b/gcc/rust/expand/rust-macro-builtins.cc
@@ -18,14 +18,14 @@
 
 #include "rust-macro-builtins.h"
 #include "rust-ast.h"
-#include "rust-diagnostics.h"
-#include "rust-expr.h"
-#include "rust-session-manager.h"
-#include "rust-macro-invoc-lexer.h"
-#include "rust-lex.h"
-#include "rust-parse.h"
-#include "rust-early-name-resolver.h"
 #include "rust-attribute-visitor.h"
+#include "rust-diagnostics.h"
+#include "rust-early-name-resolver.h"
+#include "rust-expr.h"
+#include "rust-lex.h"
+#include "rust-macro-invoc-lexer.h"
+#include "rust-parse.h"
+#include "rust-session-manager.h"
 
 namespace Rust {
 namespace {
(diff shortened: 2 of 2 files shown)


gcc/testsuite/rust/compile/builtin_macro_eager4.rs
```diff
@@ -0,0 +1,22 @@
...
+}
+
+
+fn main () {
+  let os = "linux";
+  cfg!(); // { dg-error "cfg-pattern required" }
+  cfg!(1); // { dg-error "expected identifier" }
+  cfg!(windows); // shoud not error
+  cfg!(target_os = "linux"); //should not error
+  cfg!(a!()); // {dg-error "expected 1 cfg-pattern"}
```
CohenArthur, line 19: Maybe add one more test to make sure we get proper cfg-patterns out of macro expansion?
```suggestion
  cfg!(a!()); // {dg-error "expected 1 cfg-pattern"}
  cfg!(b!()); // ok - expands to `linux`
```

CohenArthur (approved):
The code looks good to me. These functions are getting very repetitive so we'll need to work on getting them refactored eventually

gcc/testsuite/rust/compile/builtin_macro_eager4.rs
ArberSephirotheca: I have trouble figuring out what is the propery way to return a cfg-pattern by simple Macro. I tried a few things but none of them work. Could you please enlight me?
Here is my attempts;
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e7bc6709ee61ba4d8e6beb03cd87fa71

CohenArthur: I tried multiple things including something like `cfg!(target_os = linux!())`, and that is not allowed either. So I'm not sure why `cfg!()` is a macro which should be expanded eagerly. @flip1995 or @bjorn3 you might have an idea?

flip1995: Why would eager expansion be necessary for `cfg!`? Grepping through the `rustc` source code I think the only things that are eagerly expanded are

- `concat[_*]!`
- `option_env!` and `env!`
- `compiler_error!`
- `include[_*]!`
- Something in `proc_macro_server.rs`. I haven't looked into what this does.

CohenArthur: I grepped through `rustc` for `MacEager`, which gave a match in the `expand_cfg` function. What pattern did you search for?

flip1995: I grepped for `fully_expand_fragment` that does the actual expansion of an item and is used (according to the [rustc def guide](https://rustc-dev-guide.rust-lang.org/macro-expansion.html?highlight=eager#eager-expansion)) to do eager expansion. And then I traced back the functions that calls that to where they are used.

`expand_cfg` doesn't call that function AFAICT. It calls `parse_meta_item` which is IIUC the same thing you can give it when using it as an attribute. So if you have the code that parses the `cfg` attribute, you should be able to just use that to parse the `cfg!` macro.

CohenArthur: Aaaah, nice, thank you @flip1995! In that case sorry @ArberSephirotheca for misguiding you and giving you a non-issue :/ if you'd like to work on something else, we can discuss it together on Zulip? Really sorry and I hope you didn't spend too much time on this

PR #1928, opened 2023-02-24
Implemented Option_env in rust-macro-buildins

Parthib314:
Also implemented `eagar_expansion` on it.


.github/workflows/clang-format.yml
```diff
@@ -1,6 +1,7 @@
 name: Clang Format Lint
 
 on:
+  workflow_dispatch:
```
CohenArthur, line 4: This should not be in this PR. You also have multiple commits which are from your older PR, right?

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -91,6 +91,9 @@ make_eager_builtin_invocation (
     case AST::BuiltinMacro::Include:
       path_str = "include";
       break;
+    case AST::BuiltinMacro::Option_env:
```
CohenArthur, line 94: You also need to add the variant to the `BuiltinMacro` enum in `rust-macro.h`. Also, this should be named `OptionEnv` instead of `Option_env` :)

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -783,4 +786,70 @@ MacroBuiltin::line_handler (Location invoc_locus, AST::MacroInvocData &)
...
+
+  auto pending = check_for_eager_invocations (expanded_expr);
+  if (!pending.empty ())
+    return make_eager_builtin_invocation (AST::BuiltinMacro::Env, invoc_locus,
+					  invoc_token_tree,
+					  std::move (pending));
+
+  if (expanded_expr.size () < 1 || expanded_expr.size () > 2)
+    {
+      rust_error_at (invoc_locus, "env! takes 1 or 2 arguments");
```
CohenArthur, line 820:
```suggestion
      rust_error_at (invoc_locus, "option_env! takes 1 argument");
```

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -783,4 +786,70 @@ MacroBuiltin::line_handler (Location invoc_locus, AST::MacroInvocData &)
...
+    }
+  if (expanded_expr.size () > 1)
+    {
+      if (!(error_expr
+	    = try_extract_string_literal_from_fragment (invoc_locus,
+							expanded_expr[1])))
+	{
+	  return AST::Fragment::create_error ();
+	}
+    }
```
CohenArthur, line 840:
```suggestion
```

CohenArthur: `option_env` only accepts one argument

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -783,4 +786,70 @@ MacroBuiltin::line_handler (Location invoc_locus, AST::MacroInvocData &)
...
+  parser.skip_token (last_token_id);
+
+  auto env_value = getenv (lit_expr->as_string ().c_str ());
+
+  auto node = AST::SingleASTNode (make_string (invoc_locus, env_value));
+  auto tok
+    = make_token (Token::make_string (invoc_locus, std::move (env_value)));
+
+  // FIXME: Do not return an empty token vector here
+  return AST::Fragment ({node}, std::move (tok));
```
CohenArthur, line 851: `option_env` needs to return an `Option<&'static str>`, not just the string. Which I agree is quite tricky and might cause issues

CohenArthur (changes requested):

.github/workflows/clang-format.yml
Parthib314: Yes sir. I don't know why it's happening I even created a separate branch for this.

gcc/rust/expand/rust-macro-builtins.cc
CohenArthur: In `rustc`, `expand_option_env` return either `option::Option::None` or `option::Option::Some(<value>)`

Issue #1901, opened 2023-02-21
Pointers are coerceable


philberty:
I tried this code: https://godbolt.org/z/aE71GeqjK

```rust
 mod intrinsics {
    extern "rust-intrinsic" {
        pub fn offset<T>(ptr: *const T, count: isize) -> *const T;
    }
}
 
 mod ptr{
    #[lang = "const_ptr"]
    impl<T> *const T {
        pub unsafe fn offset(self, count: isize) -> *const T {
            intrinsics::offset(self, count)
        }
    }
}

fn test(x: *mut u8) {
    x.offset(1);
}
```

I expected to see this happen: compile without error

Instead, this happened: 

```
<source>:17:7: error: failed to resolve method for 'offset'
   17 |     x.offset(1);
      |       ^~~~~~
<source>:17:5: error: failed to type resolve expression
   17 |     x.offset(1);
      |     ^
```

### Meta

- What version of Rust GCC were you using, git sha if possible. d6e33adc7174bfd52afc47cd5792320ca42e13fd

Issue #1895, opened 2023-02-21
Support Iterators


philberty:
This is a parent issue to track progress on getting for-loops working, which actually uses iterators under the hood.

### Task List

- [x] #1894
- [x] #1893
- [x] #869 
- [x] #850 
- [x] #855 
- [x] #1902
- [x] https://github.com/Rust-GCC/gccrs/issues/1981
- [x] https://github.com/Rust-GCC/gccrs/issues/2000
- [x] #1984
- [x] #1899
- [x] #1904
- [x] #1898
- [x] #1897 
- [x] #1896 
- [x] #1901
- [x] #1903
- [x] #1524 
- [x] https://github.com/Rust-GCC/gccrs/issues/2015
- [x] https://github.com/Rust-GCC/gccrs/issues/2019
- [x] https://github.com/Rust-GCC/gccrs/pull/1905
- [x] https://github.com/Rust-GCC/gccrs/pull/1931
- [x] https://github.com/Rust-GCC/gccrs/pull/1945
- [x] https://github.com/Rust-GCC/gccrs/pull/1948
- [x] https://github.com/Rust-GCC/gccrs/pull/1977
- [x] https://github.com/Rust-GCC/gccrs/pull/1999
- [x] https://github.com/Rust-GCC/gccrs/pull/2011
- [x] https://github.com/Rust-GCC/gccrs/pull/2018
- [x] https://github.com/Rust-GCC/gccrs/issues/2330
- [x] https://github.com/Rust-GCC/gccrs/issues/2304
- [x] https://github.com/Rust-GCC/gccrs/issues/2179
- [x] https://github.com/Rust-GCC/gccrs/issues/2166
- [x] https://github.com/Rust-GCC/gccrs/issues/2165
- [x] https://github.com/Rust-GCC/gccrs/issues/2139
- [x] https://github.com/Rust-GCC/gccrs/issues/2136
- [x] https://github.com/Rust-GCC/gccrs/issues/2238
- [x] https://github.com/Rust-GCC/gccrs/issues/2236
- [x] https://github.com/Rust-GCC/gccrs/issues/2195
- [x] https://github.com/Rust-GCC/gccrs/issues/2190
- [x] https://github.com/Rust-GCC/gccrs/issues/2188
- [x] https://github.com/Rust-GCC/gccrs/issues/2180
- [x] https://github.com/Rust-GCC/gccrs/pull/2549

### Goal-test case:

```rust
#![feature(intrinsics)]

[...]

Issue #878, opened 2022-01-20
autoderef should stop upon pointer deref


philberty:
the rustc probe.rs shows that it stops autoderef when it tries pointers as this is unsafe.

philberty:
this was fixed by https://github.com/Rust-GCC/gccrs/pull/2322

PR #1931, opened 2023-02-26
Generic pointers coerceable

### LINKED ISSUE #1903: Generic Pointers are castable
I tried this code: https://godbolt.org/z/Kjao733Pj

```rust
fn test<T>(x: *mut T) {
    let x = x as *mut u8;
}
```

Instead, this happened:

```
<source>:2:13: error: expected '* mut u8' got '* mut T'
    2 |     let x = x as *mut u8;
      |             ^    ~
```

### Meta

philberty:
In order to fix the following issues we need to begin the journey to get rid of
the can_eq interface as well as the autoderef_flag hack inside it. In non-generic
contexts in order to call const pointer lang item methods the can_eq interface
was not possible to infer that these types are valid. This patch set adds a new
unify_and interface so that we can optionally inject inference variables and only
commit them when everything is ok and on failure cleanup. So during method
resolution and picking we can now pick using our TryCoerce interface instead
of the bad can_eq one.

Fixes #1903 #1901 #878
Addresses #1895

### DIFF
--- a/gcc/rust/typecheck/rust-coercion.cc
+++ b/gcc/rust/typecheck/rust-coercion.cc
@@ -373,7 +373,7 @@ TypeCoercionRules::coerce_unsized (TyTy::BaseType *source,
 }
 
 bool
-TypeCoercionRules::select (const TyTy::BaseType &autoderefed)
+TypeCoercionRules::select (TyTy::BaseType &autoderefed)
 {
   rust_debug (
     "autoderef type-coercion select autoderefed={%s} can_eq expected={%s}",
--- a/gcc/rust/typecheck/rust-coercion.cc
+++ b/gcc/rust/typecheck/rust-coercion.cc
@@ -18,7 +18,7 @@
 
 #include "rust-hir-type-check-base.h"
 #include "rust-coercion.h"
-#include "rust-unify.h"
+#include "rust-type-util.h"
 
 namespace Rust {
 namespace Resolver {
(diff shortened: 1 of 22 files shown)


gcc/rust/typecheck/rust-coercion.cc
```diff
@@ -169,13 +197,21 @@ TypeCoercionRules::coerce_unsafe_ptr (TyTy::BaseType *receiver,
...
 
 	return CoercionResult::get_error ();
       }
     }
 
+  bool receiver_is_non_ptr = receiver->get_kind () != TyTy::TypeKind::POINTER;
+  if (autoderef_flag && receiver_is_non_ptr)
+    {
+      // it is unsafe to autoderef to raw pointers
+      return CoercionResult::get_error ();
```
CohenArthur, line 212: Unsafe as in "this is allowed in unsafe contexts"? Or is it always forbidden?

gcc/rust/typecheck/rust-hir-dot-operator.cc
```diff
@@ -86,6 +92,38 @@ MethodResolver::select (const TyTy::BaseType &receiver)
...
+	  const TyTy::ReferenceType &sptr
+	    = *static_cast<const TyTy::ReferenceType *> (impl_self);
+	  const TyTy::ReferenceType &ptr
+	    = *static_cast<const TyTy::ReferenceType *> (raw);
+
+	  // we could do this via lang-item assemblies if we refactor this
+	  bool mut_match = sptr.mutability () == ptr.mutability ();
+	  if (!mut_match)
+	    return true;
+	}
```
CohenArthur, line 126: So having ambiguous implementations for `*mut T` and `*const T` is allowed as you pointed out in zulip, but is it also allowed for `&mut T` and `&T`? Or am I misunderstanding this bit of code?

PR #1933, opened 2023-02-27
an attempt to implement pretty HIR dump 

00AR:
In this draft, I have attempted to implement some part of pretty HIR Dump to show the proposed format for HIR Dump. My approach in this is to avoid printing any empty fields of HIR nodes. In the pretty HIR dump of the code example given below, no attributes are present in the code for the crate, so you will not see  **inner attribute: []** at all. This makes the dump look cleaner as only the available information is printed. 

Try this example for the dump
```rust
fn main() {
    let var1=5;
}

fn test_1(arg1:u32, arg2:f32)->u8 {
    let new1:u8 = 4;
    let new2:u8 = 6 + 5;
    new2
}

fn test(mul:u32)->u32 {
    let first = 5;
    let second = first + 5;
    second
}
```

Seeking feedback for making it better :)
I will be making a GSOC proposal for this project.

### DIFF
--- a/gcc/rust/hir/rust-hir-dump.cc
+++ b/gcc/rust/hir/rust-hir-dump.cc
@@ -26,31 +26,24 @@ Dump::Dump (std::ostream &stream) : stream (stream), indent (0) {}
 void
 Dump::go (HIR::Crate &crate)
 {
-  stream << "Crate"
-	 << " "
-	 << "{" << std::endl;
-  //
-
-  indent++;
-  stream << std::string (indent, indent_char);
-  stream << "inner_attrs"
-	 << ":"
-	 << " "
-	 << "[";
-  for (auto &attr : crate.inner_attrs)
-    stream << attr.as_string ();
-  stream << "]"
-	 << "," << std::endl;
-  indent--;
+  stream << "Crate {" << std::endl;
+  // inner attributes
+  if (!crate.inner_attrs.empty ())
+    {
+      indent++;
+      stream << std::string (indent, indent_char);
+      stream << "inner_attrs: [";
+      for (auto &attr : crate.inner_attrs)
+	stream << attr.as_string ();
+      stream << "]," << std::endl;
+      indent--;
+    }
 
   indent++;
   stream << std::string (indent, indent_char);
   //
 
-  stream << "items"
-	 << ":"
-	 << " "
-	 << "[";
+  stream << "items: [";
 
   stream << std::string (indent, indent_char);
   for (const auto &item : crate.items)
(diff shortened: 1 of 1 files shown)


gcc/rust/hir/rust-hir-dump.cc
```diff
@@ -29,19 +29,21 @@ Dump::go (HIR::Crate &crate)
...
-  indent--;
+  // inner attributes
+  if (!crate.inner_attrs.empty ())
+    {
+      indent++;
+      stream << std::string (indent, indent_char);
+      stream << "inner_attrs"
+	     << ":"
+	     << " "
+	     << "[";
```
CohenArthur, line 40:
```suggestion
      stream << "inner_attrs: ["
```

gcc/rust/hir/rust-hir-dump.cc
```diff
@@ -29,19 +29,21 @@ Dump::go (HIR::Crate &crate)
...
+      for (auto &attr : crate.inner_attrs)
+	stream << attr.as_string ();
+      stream << "]"
+	     << "," << std::endl;
```
CohenArthur, line 44:
```suggestion
      stream << "]," << std::endl;
```

gcc/rust/hir/rust-hir-dump.cc
```diff
@@ -467,11 +603,44 @@ void
...
 {}
 void
-Dump::visit (LetStmt &)
-{}
+Dump::visit (LetStmt &let_stmt)
+{
+  indent++;
+  // TODO: outer attributes
+  stream << std::string (indent, indent_char);
+  stream << "let ";
```
CohenArthur, line 611: Probably clearer to have something like `LetStmt: ` like we have for `Stmt: `, `Expr: `, `Function: ` etc

CohenArthur (commented):
This is good work, thank you for working on this :) I like the idea of not having empty vectors everywhere. I think for your proposal it might be nice to consider a new side of the HIR dump which would be a more configurable dump - so for example, it might be nice to have an option to disable the printing of `HirId`s or `NodId`s. They are useful but might clutter the output if one is not looking for them. I'm looking forward to your proposal!

00AR:
Thanks for reviewing!

>  I think for your proposal it might be nice to consider a new side of the HIR dump which would be a more configurable dump - so for example, it might be nice to have an option to disable the printing of `HirId`s or `NodId`s. 

If I understand correctly, you mean introduce some more options other than ```-frust-dump-hir-pretty```, such as ```-frust-dump-hir-pretty-no-hid``` and ```-frust-dump-hir-pretty-no-nid```.

CohenArthur:
> Thanks for reviewing!
> 
> > I think for your proposal it might be nice to consider a new side of the HIR dump which would be a more configurable dump - so for example, it might be nice to have an option to disable the printing of `HirId`s or `NodId`s.
> 
> If I understand correctly, you mean introduce some more options other than `-frust-dump-hir-pretty`, such as `-frust-dump-hir-pretty-no-hid` and `-frust-dump-hir-pretty-no-nid`.

Yes :) but we can see about this later down the line :) I think this PR is in a good enough state already if you fix the few suggestions I left, and we can merge it :)

CohenArthur (approved):
Great stuff :) is this still a draft or are you happy enough for us to merge it?

00AR:
>are you happy enough for us to merge it?

Yes, I was just waiting for the checks to pass.
Thank you for the review!

Issue #1898, opened 2023-02-21
Support overflowing_add in libcore 


philberty:
I tried this code:

```rust
mod intrinsics {
    extern "rust-intrinsic" {
        pub fn add_with_overflow<T>(x: T, y: T) -> (T, bool);
    }
}

pub enum Option<T> {
    None,
    Some(T),
}

#[lang = "i32"]
impl i32 {
    pub fn checked_add(self, rhs: Self) -> Option<Self> {
        let (a, b) = self.overflowing_add(rhs);
        if b {
            Option::None
        } else {
            Option::Some(a)
        }
    }

    pub fn overflowing_add(self, rhs: Self) -> (Self, bool) {
        let (a, b) = unsafe { intrinsics::add_with_overflow(self as i32, rhs as i32) };
        (a as Self, b)
    }
}
```

I expected to see this happen: compile without error

- [ ] #1896 
- [x] #1897

```
test.rs:24:31: error: unknown builtin intrinsic: add_with_overflow
   24 |         let (a, b) = unsafe { intrinsics::add_with_overflow(self as i32, rhs as i32) };
      |                               ^~~~~~~~~~
```

### Meta

- What version of Rust GCC were you using, git sha if possible.

PR #1938, opened 2023-02-28
Fix casting error behind generics [WIP]

### LINKED ISSUE #1897: Bad cast error when behind generics
I tried this code: https://godbolt.org/z/MWbn8587K

```rust
fn test<T>(a:T) -> T {
    a
}

fn main() {
    let a:i32 = test(123 as i32) as i32;
}
```

Instead, this happened:

```
<source>:6:17: error: invalid cast 'i32' to 'i32' [E0054]
    6 |     let a:i32 = test(123 as i32) as i32;
      |                 ^~~~                ~~~
```

### Meta

goar5670:
Fixes #1897 

gcc/rust/ChangeLog:
```
* typecheck/rust-casts.cc (TypeCastRules::cast_rules): Perform destructure on `from` type
```

gcc/testsuite/ChangeLog:
```
* rust/compile/cast_generics.rs: New test.
```

### DIFF
--- a/gcc/rust/typecheck/rust-casts.cc
+++ b/gcc/rust/typecheck/rust-casts.cc
@@ -60,15 +60,16 @@ TypeCastRules::cast_rules ()
   // https://github.com/rust-lang/rust/blob/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/compiler/rustc_typeck/src/check/cast.rs#L596
   // https://github.com/rust-lang/rust/blob/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/compiler/rustc_typeck/src/check/cast.rs#L654
 
-  rust_debug ("cast_rules from={%s} to={%s}",
-	      from.get_ty ()->debug_str ().c_str (),
+  TyTy::BaseType *from_type = from.get_ty ()->destructure ();
+
+  rust_debug ("cast_rules from={%s} to={%s}", from_type->debug_str ().c_str (),
 	      to.get_ty ()->debug_str ().c_str ());
 
-  switch (from.get_ty ()->get_kind ())
+  switch (from_type->get_kind ())
     {
       case TyTy::TypeKind::INFER: {
 	TyTy::InferType *from_infer
-	  = static_cast<TyTy::InferType *> (from.get_ty ());
+	  = static_cast<TyTy::InferType *> (from_type);
 	switch (from_infer->get_infer_kind ())
 	  {
 	  case TyTy::InferType::InferTypeKind::GENERAL:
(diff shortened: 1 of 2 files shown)


gcc/rust/typecheck/rust-casts.cc
```diff
@@ -63,11 +63,11 @@ TypeCastRules::cast_rules ()
 	      from.get_ty ()->debug_str ().c_str (),
 	      to.get_ty ()->debug_str ().c_str ());
 
-  switch (from.get_ty ()->get_kind ())
+  switch (from.get_ty ()->destructure ()->get_kind ())
```
philberty, line 66: I think it would be better to do something like the unify code over in rust-unify.cc UnifyRules::go

```
TyTy::BaseType *
UnifyRules::go ()
{
  TyTy::BaseType *ltype = lhs.get_ty ();
  TyTy::BaseType *rtype = rhs.get_ty ();

  ltype = lhs.get_ty ()->destructure ();
  rtype = rhs.get_ty ()->destructure ();
```

So that you only need to call it once and it doesn't get lost.

philberty (changes requested):
You found the right fix I think a small tweak to call restructure once would be ideal here. I also think this should be squashed into 1 commit.

Otherwise great job.

philberty:
Just so you know i would really like to merge this soon because I want to work on https://github.com/Rust-GCC/gccrs/issues/1898 tomorrow

philberty:
If your having problems with the git commit format checker you can run the checker locally via: 

```
python3 contrib/gcc-changelog/git_check_commit.py origin/master..HEAD
Checking ed45b943a1fee3310fea361c6e127f5a60ae6d22: OK
Checking 27c4590301a3da7d62d5b7e5a88cdbdc8de23070: OK
Checking 5c9a0e4a49f23f277652af6d1cc8173bc255ad44: OK
Checking d5397d90190d4a4d13c688815924b0a9439f3a98: OK
Checking 92d31693da5401291db6f667d9997a889bf0ea8a: OK
Checking 48c65d249e3fdabdac0947226f375cf911529c6b: OK
Checking 4ae86b6ce74474e65da436ab92bcd9b87ed1833b: OK
Checking fa9c51aa8d435b953c58f0ba2db3a339cd9c6f59: OK
Checking a44cf951951d6a0c747a18b3d7c71d1bb96134e6: OK
Checking b6a8e19529614610519ab97d1afa7e5cd214a2a9: OK
```

gcc/rust/typecheck/rust-casts.cc
goar5670: Edited it, please take another look.

philberty (approved):
LGTM good work

PR #1941, opened 2023-03-01
Unify TupleItems

powerboat9:
Unifies ```TupleStructItems``` and ```TuplePatternItems``` into ```TupleItems```, since they're essentially duplicates of each other.


philberty:
Added @flip1995 how does rustc AST and HIR represent tuples? This patch seems like a good idea to unify them under one structure.

CohenArthur:
@powerboat9 did you have a look at the reference for the AST data structures used by rustc for this? https://doc.rust-lang.org/reference/patterns.html#tuple-struct-patterns

I'm reviewing your PR right now

flip1995:
So a tuple as an expression is just an `ExprKind::Tup` in AST and in HIR. It just holds a list of expressions.

A struct item in the AST is conceptually just an enum with [one variant](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/ast/enum.VariantData.html) . And a tuple struct is not really different from a "normal" struct: it's just a list of fields.

In a struct expression, there is no difference between a tuple struct and a normal struct.

For patterns, there are 2 `PatKind` in the AST: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/ast/enum.PatKind.html#variant.Struct

All of this also applies to the HIR.

powerboat9:
Now unifies TupleItemsMultiple and TupleItemsRanged into TupleItems, generally following https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/ast/enum.PatKind.html and https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/hir/enum.PatKind.html

philberty (approved):
I think this is starting to look good to merge, but please understand in future for such a large refactor open an issue up about this change so it can be discussed or a topic in zulip. Its hard for @CohenArthur and I to review PR's like this without the extra context of what you are trying to do or have more detailed commit messages.

philberty (changes requested):
Hmm I keep going back and forth on accepting this PR on the face of it, its simplifying alot of HIR and AST but none of this stuff was implemented so we cant see what regression we might have.

@goar5670 is working through a lot of these patterns now so i fear this might cause some conflicts at this point.

We are at a critical stage with the project, so i think i would prefer holding on this PR at the moment until i would see the AST dump and HIR dump's implemented properly would help with the confidence here.

powerboat9:
Should I split the first three comments off into their own PR?

philberty:
same here @powerboat9 feel free to reference both of these PR's in an issue

PR #1946, opened 2023-03-01
hir: Refactor ASTLoweringStmt to source file.

CohenArthur:
gcc/rust/ChangeLog:

	* Make-lang.in: Add rust-ast-lower-stmt.o
	* hir/rust-ast-lower-stmt.h: Move definitions to...
	* hir/rust-ast-lower-stmt.cc: ...here.


philberty (approved):
LGTM

philberty:
I have noticed that compiling gccrs is getting better and better as we are cleaning up these headers.

PR #1951, opened 2023-03-01
parser: Fix while let expr parsing

### LINKED ISSUE #1949: Failed to parse while let expression
I tried this code: **with -fsyntax-only**

```rust
pub enum Option<T> {
    None,
    Some(T),
}

fn main() {
    let mut x = Option::Some(3);
    let a = while let Option::Some(1) = x {
        x = Option::None;
    };
}
```

Instead, this happened:
```
while_let_expr.rs:8:19: error: appears to be while let loop but is being parsed by while loop - this may be a compiler issue
    8 |     let a = while let Option::Some(1) = x {
      |                   ^~~
while_let_expr.rs:8:19: error: failed to parse expression in let statement
while_let_expr.rs:8:5: error: failed to parse statement or expression without block in block expression
    8 |     let a = while let Option::Some(1) = x {
      |     ^~~
while_let_expr.rs:10:5: error: unrecognised token ‘}’ for start of item
   10 |     };
      |     ^
while_let_expr.rs:10:5: error: failed to parse item in crate
```

### Meta

- What version of Rust GCC were you using:
9dce6f995c2b3bc021a97872b413807de7d294b1 

Maybe related to #1904

P-E-P:
While let expr return unit but are valid construct in rust, they should therefore be included in the parsing code.

Fixes #1949

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -8373,7 +8373,7 @@ Parser<ManagedTokenSource>::parse_while_let_loop_expr (AST::AttrVec outer_attrs,
     locus = lexer.peek_token ()->get_locus ();
   else
     locus = label.get_locus ();
-  skip_token (WHILE);
+  maybe_skip_token (WHILE);
 
   /* check for possible accidental recognition of a while loop as a while let
    * loop */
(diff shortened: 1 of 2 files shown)


P-E-P:
@CohenArthur @philberty  I switched the `skip_token` to `maybe_skip_token` because the `while` has already been eaten in some cases, however I'm not entirely sure if it is appropriate. Is there some case where we do need an hard error ?

philberty:
My gut feeling is that until we find otherwise, maybe_skip seems like a good option here.

Issue #1872, opened 2023-02-16
Add location info when opening a file fails in `include_{,str,bytes}!()`


CohenArthur:
We are missing proper location info when emitting errors for these three "include" macros. We should error out using the invocation's location, so that the user knows which macro invocation is invalid.

nikos-alexandris:
Hi, I would like to try resolving this issue as my first issue on GCC :)

nikos-alexandris:
Shouldn't this issue be marked as closed? The `include_str` and `include_bytes` macros were fixed with the commits referenced above, and the `include` macro already reports the invocation's location in the error message.

CohenArthur:
Correct! We forgot to add a `Fixes #1872` to https://github.com/Rust-GCC/gccrs/pull/1955 :) thank you @nikos-alexandris

PR #1955, opened 2023-03-02
Add location info when opening a file fails in include_str and include_bytes #1872

nikos-alexandris:


gcc/testsuite/rust/compile/builtin_macro_include_bytes_location_info.rs
```diff
@@ -0,0 +1,10 @@
+#![feature(rustc_attrs)]
+
+#[rustc_builtin_macro]
+macro_rules! include_bytes {
+    () => {{}};
+}
+
+fn main() {
+    include_bytes!("nonexistent.txt"); // { dg-error "cannot open filename (.*?)nonexistent.txt: No such file or directory" }
```
CohenArthur, line 9: Any reason you needed to add the `(.*?)` to the regex?

CohenArthur (approved):
Great! Thank you :) LGTM

gcc/testsuite/rust/compile/builtin_macro_include_bytes_location_info.rs
nikos-alexandris: When running the tests, the name of the file is reported as an absolute path, so this will match any path with the file name at the end

Issue #1077, opened 2022-03-31
Handle `#[macro_use]` attribute


CohenArthur:
- [ ] Implement attribute behavior (reference on the MBE page)
- [ ] Handle `#[macro_use(macro_name)]` variant properly

PR #1957, opened 2023-03-02
enr: Fetch module items during early name resolution

CohenArthur:
This is important as public macros can be present in other modules,
which would otherwise not be loaded until the expansion phase
happening right after the early name resolution.

gcc/rust/ChangeLog:

	* resolve/rust-early-name-resolver.cc (EarlyNameResolver::visit): Move
	unloaded module item loading to...
	* expand/rust-attribute-visitor.cc (AttrVisitor::visit): ...here.

Addresse #1077


philberty (approved):
LGTM

PR #1947, opened 2023-03-01
hir: Add ExportedMacro node and handling.

CohenArthur:
This HIR node represents macros which should be exported into the final Rust metadata files. Because our metadata exporter operates on the HIR, while macros are inherently tied to the AST, we need a way to propagate exported macros up until the metadata export pass on the HIR. Hence the existence of this class, whose sole purpose is to keep enough information for the metadata exporter to retrieve the original AST::MacroRulesDefinition. Handling for actually exporting these macros will come later.

gcc/rust/ChangeLog:

	* hir/tree/rust-hir-item.h (class ExportedMacro): Add new ExportedMacro class.
	* hir/tree/rust-hir.cc (ExportedMacro::accept_vis): New function.
	(ExportedMacro::get_locus): Likewise.
	(ExportedMacro::get_item_kind): Likewise.
	(ExportedMacro::clone_item_impl): Likewise.
	* hir/tree/rust-hir-full-decls.h (class ExportedMacro): Forward declare class.
	* backend/rust-compile-item.h: Add visitor for ExportedMacro.
	* backend/rust-compile-stmt.h: Likewise.
	* checks/errors/privacy/rust-privacy-reporter.cc (PrivacyReporter::visit): Likewise.
	* checks/errors/privacy/rust-privacy-reporter.h: Likewise.
	* checks/errors/privacy/rust-pub-restricted-visitor.cc (PubRestrictedVisitor::visit): Likewise.
	* checks/errors/privacy/rust-pub-restricted-visitor.h: Likewise.
	* checks/errors/privacy/rust-reachability.cc (ReachabilityVisitor::visit): Likewise.
	* checks/errors/privacy/rust-reachability.h: Likewise.
	* checks/errors/privacy/rust-visibility-resolver.cc (VisibilityResolver::visit): Likewise.
	* checks/errors/privacy/rust-visibility-resolver.h: Likewise.
	* checks/errors/rust-const-checker.cc (ConstChecker::visit): Likewise.
	* checks/errors/rust-const-checker.h: Likewise.
	* checks/errors/rust-unsafe-checker.cc (UnsafeChecker::visit): Likewise.
	* checks/errors/rust-unsafe-checker.h: Likewise.
	* hir/rust-ast-lower-item.cc (ASTLoweringItem::visit): Likewise.
	* hir/rust-ast-lower-item.h: Likewise.
	* hir/rust-hir-dump.cc (Dump::visit): Likewise.
	* hir/rust-hir-dump.h: Likewise.
[...]

PR #1958, opened 2023-03-02
Add ExportedMacroCollector

CohenArthur:
Exported macros need to be placed at the root of the crate for metadata
export purposes. This visitor goes through various items susceptibles of
containing exported macros and inserts them at the end of the top-level
items vector within the crate.
    
gcc/rust/ChangeLog:
    
        * Make-lang.in: Add rust-exported-macro-collector.o.
        * metadata/rust-exported-macro-collector.cc: New file.
        * metadata/rust-exported-macro-collector.h: New file.
        * metadata/rust-export-metadata.cc (PublicInterface::gather_export_data):
        Gather public macros before exporting metadata.

Addresses #1077
Needs #1947 so only review the last commit

### DIFF
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -162,6 +162,7 @@ GRS_OBJS = \
     rust/rust-compile-base.o \
     rust/rust-tree.o \
     rust/rust-compile-context.o \
+	rust/rust-exported-macro-collector.o \
     rust/rust-export-metadata.o \
     rust/rust-imports.o \
     rust/rust-import-archive.o \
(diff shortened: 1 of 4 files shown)


philberty (approved):
I will approve this but it just dawned on me. Instead of walking the crate again to find the exported macros. What about as we are doing HIR lowering you keep mappings of the NodeId's for the exported macros since we are walking the fully expanded AST crate that point anyways and just ignoring the macro-decls.

We should be using local-def-id's but they dont exist on our AST but they really should some day lol. But NodeId should be enough might be simpler.

CohenArthur:
> I will approve this but it just dawned on me. Instead of walking the crate again to find the exported macros. What about as we are doing HIR lowering you keep mappings of the NodeId's for the exported macros since we are walking the fully expanded AST crate that point anyways and just ignoring the macro-decls.
> 
> We should be using local-def-id's but they dont exist on our AST but they really should some day lol. But NodeId should be enough might be simpler.

Mh that is a good idea and probably simpler than adding an entirely new visitor. I'll have a go at it before merging anything, to see if it's cleaner/easier. Good catch :)

CohenArthur:
Closing in favor of #1991

PR #1960, opened 2023-03-02
parser: Fix parsing closure parameter

### LINKED ISSUE #1910: Invalid parsing of wildcard closure parameter
I tried this code:

```rust
// { dg-additional-options "-fsyntax-only" }

pub fn encode_normal<T: RawFloat>(x: Unpacked) -> T {
    debug_assert!(
        T::MIN_SIG <= x.sig && x.sig <= T::MAX_SIG,
        "encode_normal: significand not normalized"
    );
    // Remove the hidden bit
    let sig_enc = x.sig & !(1 << T::EXPLICIT_SIG_BITS);
    // Adjust the exponent for exponent bias and mantissa shift
    let k_enc = x.k + T::MAX_EXP + T::EXPLICIT_SIG_BITS as i16;
    debug_assert!(k_enc != 0 && k_enc < T::MAX_ENCODED_EXP, "encode_normal: exponent out of range");
    // Leave sign bit at 0 ("+"), our numbers are all positive
    let bits = (k_enc as u64) << T::EXPLICIT_SIG_BITS | sig_enc;
    T::from_bits(bits.try_into().unwrap_or_else(|_| unreachable!()))
}
```

similar testcase:
```rust
fn foo() {
    let a = |_| 15;
}
```

Instead, this happened: 
```rust
../../rust/library/core/src/num/dec2flt/rawfp.rs:295:67: error: expecting ‘|’ but ‘)’ found
  295 |     T::from_bits(bits.try_into().unwrap_or_else(|_| unreachable!()))
      |                                                                   ^
../../rust/library/core/src/num/dec2flt/rawfp.rs:295:49: error: failed to parse method param in method call
  295 |     T::from_bits(bits.try_into().unwrap_or_else(|_| unreachable!()))
      |                                                 ^
../../rust/library/core/src/num/dec2flt/rawfp.rs:295:18: error: failed to parse expression in struct (or enum) expression tuple
  295 |     T::from_bits(bits.try_into().unwrap_or_else(|_| unreachable!()))
 ```

### Meta

- What version of Rust GCC were you using: 0d5fe7b9aee

goar5670:
Fixes #1910

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -9055,7 +9055,7 @@ Parser<ManagedTokenSource>::parse_closure_param ()
   AST::AttrVec outer_attrs = parse_outer_attributes ();
 
   // parse pattern (which is required)
-  std::unique_ptr<AST::Pattern> pattern = parse_pattern ();
+  std::unique_ptr<AST::Pattern> pattern = parse_pattern_no_alt ();
   if (pattern == nullptr)
     {
       // not necessarily an error
--- a/gcc/testsuite/rust/compile/closure_no_type_anno.rs
+++ b/gcc/testsuite/rust/compile/closure_no_type_anno.rs
@@ -1,0 +1,5 @@
+// { dg-additional-options "-fsyntax-only" }
+
+pub fn foo() {
+  let a = |_| 15;
+}


goar5670:
@CohenArthur after this fix, this case results in an ICE without `-fsyntax-only`, is this expected or should I open an issue for it? For rustc, it asks for a type annotation ([E0282])

goar5670:
Just noticed that Cohen went on vacation so cc @philberty

philberty:
@goar5670 It make sense that Rustc fails this test case as you have specified an inference variable for the closure parameter and it is unused so therefore you end up with a floating inference variable thats not unified against anything. This will ICE for gccrs at the moment as we havent added any support for that yet its on my radar.

when we merge this @goar5670 can you raise an issue with the ICE you get when running without -fsyntax only

PR #1963, opened 2023-03-03
Add length checking for tuple patterns. Resolves issue Rust-GCC#1849

nikos-alexandris:
gcc/rust/ChangeLog:

	* typecheck/rust-hir-type-check-pattern.cc (TypeCheckPattern::visit): Add length checking for tuple patters.

gcc/testsuite/ChangeLog:

	* rust/compile/tuple_mismatch.rs: New test.

Signed-off-by: Nikos Alexandris <nikos-alexandris@protonmail.com>

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-pattern.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-pattern.cc
@@ -262,16 +262,30 @@ TypeCheckPattern::visit (HIR::TuplePattern &pattern)
 	  = *static_cast<HIR::TuplePatternItemsMultiple *> (
 	    pattern.get_items ().get ());
 
-	std::vector<TyTy::TyVar> pattern_elems;
-	for (size_t i = 0; i < ref.get_patterns ().size (); i++)
+	if (parent->get_kind () != TyTy::TUPLE)
 	  {
-	    auto &p = ref.get_patterns ()[i];
-	    TyTy::BaseType *par_type = parent;
-	    if (parent->get_kind () == TyTy::TUPLE)
-	      {
-		TyTy::TupleType &par = *static_cast<TyTy::TupleType *> (parent);
-		par_type = par.get_field (i);
-	      }
+	    rust_error_at (pattern.get_locus (), "expected %s, found tuple",
+			   parent->as_string ().c_str ());
+	    break;
+	  }
+
+	const auto &patterns = ref.get_patterns ();
+	size_t nitems_to_resolve = patterns.size ();
+
+	TyTy::TupleType &par = *static_cast<TyTy::TupleType *> (parent);
+	if (patterns.size () != par.get_fields ().size ())
+	  {
+	    emit_pattern_size_error (pattern, par.get_fields ().size (),
+				     patterns.size ());
+	    nitems_to_resolve
+	      = std::min (nitems_to_resolve, par.get_fields ().size ());
+	  }
+
+	std::vector<TyTy::TyVar> pattern_elems;
+	for (size_t i = 0; i < nitems_to_resolve; i++)
+	  {
+	    auto &p = patterns[i];
+	    TyTy::BaseType *par_type = par.get_field (i);
 
 	    TyTy::BaseType *elem
 	      = TypeCheckPattern::Resolve (p.get (), par_type);
@@ -405,6 +419,22 @@ TypeCheckPattern::visit (HIR::SlicePattern &pattern)
 {
   rust_sorry_at (pattern.get_locus (),
 		 "type checking qualified path patterns not supported");
+}
+
+void
+TypeCheckPattern::emit_pattern_size_error (const HIR::Pattern &pattern,
+					   size_t expected_field_count,
+					   size_t got_field_count)
+{
+  RichLocation r (pattern.get_locus ());
+  r.add_range (mappings->lookup_location (parent->get_ref ()));
+  rust_error_at (r,
+		 "expected a tuple with %lu %s, found one "
+		 "with %lu %s",
+		 (unsigned long) expected_field_count,
+		 expected_field_count == 1 ? "element" : "elements",
+		 (unsigned long) got_field_count,
+		 got_field_count == 1 ? "element" : "elements");
 }
 
 } // namespace Resolver
--- a/gcc/rust/typecheck/rust-hir-type-check-pattern.h
+++ b/gcc/rust/typecheck/rust-hir-type-check-pattern.h
@@ -51,6 +51,10 @@ private:
 				 Analysis::NodeMapping mappings,
 				 Location locus);
 
+  void emit_pattern_size_error (const HIR::Pattern &pattern,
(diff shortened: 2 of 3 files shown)


gcc/rust/typecheck/rust-hir-type-check-pattern.cc
```diff
@@ -262,16 +262,35 @@ TypeCheckPattern::visit (HIR::TuplePattern &pattern)
 	  = *static_cast<HIR::TuplePatternItemsMultiple *> (
 	    pattern.get_items ().get ());
 
+	size_t items = ref.get_patterns ().size ();
```
philberty, line 265: you should change this to be 'number_of_items' or 'nitems'. Or you could do:

```
const auto& patterns = ref.get_patterns();

// then when you want the number of times do:
patterns.size()
```

gcc/rust/typecheck/rust-hir-type-check-pattern.cc
```diff
@@ -262,16 +262,35 @@ TypeCheckPattern::visit (HIR::TuplePattern &pattern)
...
+	    size_t fields_cnt = par.get_fields ().size ();
+	    size_t patterns_cnt = ref.get_patterns ().size ();
+	    rust_error_at (pattern.get_locus (),
+			   "expected a tuple with %lu %s, found one "
+			   "with %lu %s",
+			   (unsigned long) fields_cnt,
+			   fields_cnt == 1 ? "element" : "elements",
+			   (unsigned long) patterns_cnt,
+			   patterns_cnt == 1 ? "element" : "elements");
+	    items = std::min (items, fields_cnt);
```
philberty, line 286: should this just return error_mark_node now that an error has occured

gcc/rust/typecheck/rust-hir-type-check-pattern.cc
```diff
@@ -262,16 +262,35 @@ TypeCheckPattern::visit (HIR::TuplePattern &pattern)
...
+	  }
+
 	std::vector<TyTy::TyVar> pattern_elems;
-	for (size_t i = 0; i < ref.get_patterns ().size (); i++)
+	for (size_t i = 0; i < items; i++)
```
philberty, line 290: refactor this to use a better name

philberty (changes requested):
Looks like a good start i've left some comments and you have created an excellent test case.

What are the rules here when using a tuple struct is it the same sort of logic? If so you might want to tweak the code here to be more permissive

gcc/rust/typecheck/rust-hir-type-check-pattern.cc
nikos-alexandris: This is the reason why i added the `foo` function in my test case (maybe i should include a comment to clarify). If we have something of the form

```rust
fn foo() -> i32 {
  (let b, _) = (true, 2, 3); // obv error
  use b
}
```

if we return early the error says "failed to resolve root segment", whereas if we do it as i did, we resolve the min number of matching elements (here 2) so we get that b is of type bool, and we get an error of type "expected i32, got bool".

nikos-alexandris:
> What are the rules here when using a tuple struct is it the same sort of logic? If so you might want to tweak the code here to be more

Things like these

```rust
struct Tup(i32, i32);

let (_, _) = Tup(1, 2);
let t: Tup = (1, 2);
```

are errors according to rustc. Please correct me if i didn't understand your comment as you meant and thanks for the patience :smiley:

gcc/rust/typecheck/rust-hir-type-check-pattern.cc
philberty: Perfect

philberty:
> > What are the rules here when using a tuple struct is it the same sort of logic? If so you might want to tweak the code here to be more
> 
> Things like these
> 
> ```rust
> struct Tup(i32, i32);
> 
> let (_, _) = Tup(1, 2);
> let t: Tup = (1, 2);
> ```
> 
> are errors according to rustc. Please correct me if i didn't understand your comment as you meant and thanks for the patience smiley

I just noticed in your patch you do a check for

```if (parent->get_kind () != TyTy::TUPLE)```

If this is meant to work with Tuple Structs too you need to check for TyTy::ADT since there is a difference between TyTy::Tuple for normal tiples and Tuple structs which are actually an ADT. 

```
class ADTType : public BaseType, public SubstitutionRef
{
public:
  enum ADTKind
  {
    STRUCT_STRUCT,
    TUPLE_STRUCT,
    UNION,
    ENUM
  };

```

Maybe try this test case you mentioned here.

File: gcc/rust/typecheck/rust-hir-type-check-pattern.cc [hunk 4]
```diff
@@ -262,16 +262,35 @@ TypeCheckPattern::visit (HIR::TuplePattern &pattern)
...
+	  {
+	    size_t fields_cnt = par.get_fields ().size ();
+	    size_t patterns_cnt = ref.get_patterns ().size ();
+	    rust_error_at (pattern.get_locus (),
+			   "expected a tuple with %lu %s, found one "
+			   "with %lu %s",
+			   (unsigned long) fields_cnt,
+			   fields_cnt == 1 ? "element" : "elements",
+			   (unsigned long) patterns_cnt,
+			   patterns_cnt == 1 ? "element" : "elements");
```
philberty, line 285: For an error as well done as this one is, can you extract it our as a separate function so that it becomes:

```emit_pattern_size_error (expected_field_count, got_field_count)```

We have been starting to extract out error diagnostics like this as separate functions so that they can be improved over time. One thing you might like to try is:

```
RichLocation r(pattern.get_locus())
r.add_Range(mappings->lookup_location(parent->get_Ref()));
rust_error_at(r, "error-message")
```

**Comment by nikos-alexandris** [comment]
> If this is meant to work with Tuple Structs too you need to check for TyTy::ADT since there is a difference between TyTy::Tuple for normal tiples and Tuple structs which are actually an ADT.

Yes, this is not meant to work with Tuple Structs; there already is a function

```c++
void
TypeCheckPattern::visit (HIR::TupleStructPattern &pattern)
```

that should handle Tuple Struct patterns, and since we can't assign tuple structs to tuples we shouldn't relax the condition.

nikos-alexandris:
The changes have been applied an squashed into a single commit

philberty:
> > If this is meant to work with Tuple Structs too you need to check for TyTy::ADT since there is a difference between TyTy::Tuple for normal tiples and Tuple structs which are actually an ADT.
> 
> Yes, this is not meant to work with Tuple Structs; there already is a function
> 
> ```c++
> void
> TypeCheckPattern::visit (HIR::TupleStructPattern &pattern)
> ```
> 
> that should handle Tuple Struct patterns, and since we can't assign tuple structs to tuples we shouldn't relax the condition.

Ah ok thanks for the info that makes sense! :D Nice work.

philberty (approved):
LGTM nice work

PR #1965, opened 2023-03-04
Allow all keywords (except `priv`) to follow `:vis` in macro rules

### LINKED ISSUE #1060: Allow identifiers other than "priv" to follow a `:vis`
When checking for follow-set restrictions, we must allow all identifiers other than "priv" to follow a `:vis` fragment.

The code to change is in `gcc/rust/parse/rust-parse.cc`. Follow-up tokens are checked in `peculiar_fragment_match_compatible`. We currently allow all identifiers after a `:vis`, but that needs to be restricted to all identifiers except "priv".

> FOLLOW(vis) = {`,` any keyword or identifier except a non-raw priv; any token that can begin a type; ident, ty, and path nonterminals}.

From: https://doc.rust-lang.org/reference/macro-ambiguity.html

tage64:
Previously, the following macro rules were rejected by gccrs:
```Rust
macro_rules! {
    ($v:vis <KEY_WORD>) => { ... };
}
```

This PR fixes so the above code is accepted by the compiler for all key words like `async` or `unsafe`.
The only exception is the keyword `priv` which is not allowed.
See [this page](https://doc.rust-lang.org/reference/macro-ambiguity.html) for reference. Especially the following excerpt:
> FOLLOW(vis) = {,l any keyword or identifier except a non-raw priv; any token that can begin a type; ident, ty, and path nonterminals}.

Fixes #1060


gcc/rust/parse/rust-parse.cc
```diff
@@ -150,50 +151,51 @@ peculiar_fragment_match_compatible_fragment (
...
+    EXCLAM,	   ASTERISK,
+    AMP,	   LOGICAL_AND,
+    QUESTION_MARK, LIFETIME,
+    LEFT_ANGLE,	   LEFT_SHIFT,
+    UNDERSCORE,
+    // FIXME: Add $crate as valid
+  };
+  // Add all keyword identifiers.
+  // RS_TOKEN_LIST is defined in rust-token.h.
+#define RS_TOKEN(name, _)
```
philberty, line 170: What is the purpose of defining these macros from the lexer?

gcc/rust/parse/rust-parse.cc
```diff
@@ -150,50 +151,51 @@ peculiar_fragment_match_compatible_fragment (
...
+      {AST::MacroFragSpec::EXPR, {MATCH_ARROW, COMMA, SEMICOLON}},
+      {AST::MacroFragSpec::STMT, {MATCH_ARROW, COMMA, SEMICOLON}},
+      {AST::MacroFragSpec::PAT, {MATCH_ARROW, COMMA, EQUAL, PIPE, IF, IN}},
+      {AST::MacroFragSpec::PATH,
+       {MATCH_ARROW, COMMA, EQUAL, PIPE, SEMICOLON, COLON, RIGHT_ANGLE,
+	RIGHT_SHIFT, LEFT_SQUARE, LEFT_CURLY, AS, WHERE}},
+      {AST::MacroFragSpec::TY,
+       {MATCH_ARROW, COMMA, EQUAL, PIPE, SEMICOLON, COLON, RIGHT_ANGLE,
+	RIGHT_SHIFT, LEFT_SQUARE, LEFT_CURLY, AS, WHERE}},
+      {AST::MacroFragSpec::VIS, vis_fragment_follow_set ()},
```
philberty, line 197: I can see you are trying to avoid having duplication I am starting to wonder for this patch would it not just be simpler to just add the keywords here?

gcc/testsuite/rust/compile/xfail/macro_vis_priv_not_in_follow_set.rs
```diff
@@ -0,0 +1,16 @@
+// { dg-excess-errors "warnings" }
+
+// Check that "priv" is not in the follow set of :vis.
+
+// { dg-error "priv not follow :vis" "#359" { xfail *-*-* } .+1 }
```
P-E-P, line 5: I believe this should be a `dg-error` instead. `xfail` test category is used for tests that are expected to fail but that should be fixed someday, they highlight a current bug that should be fixed; not tests that are expected to fail in the long term.

P-E-P (changes requested):
Slight test classification error. Otherwise LGTM, nice work!

gcc/testsuite/rust/compile/xfail/macro_vis_priv_not_in_follow_set.rs
tage64: I moved the test file from `rust/xfail/macro_vis_priv_not_in_follow_set.rs` to `rust/compile/macro48.rs`.
And I also changed the "dg-error" line to:
```Rust
// { dg-error "token `priv` is not allowed after 'vis' fragment" "#359" { target *-*-* } .+2 }
```
Although I am not really sure what every part of that line mean.

gcc/testsuite/rust/compile/macro48.rs
```diff
@@ -0,0 +1,10 @@
+// { dg-excess-errors "allow" }
```
philberty, line 1: hmm we have to be cautious about using dg-excess-errors as it swallows up everything. What are the excess errors?

philberty (changes requested):
I think you are getting there with the test case. I still think that just listing the extra keywords would be a better patch for now so that it follows the pattern that was there in the original code.

tage64:
Now I have hard coded all keywords and removed the dg-excess-errors line. Please comment on the test file though if I'm doing things in an inelegant way.

tage64:
I apologize for the formatting, but clang-format insists on having every keyword on a single line.

gcc/rust/parse/rust-parse.cc
```diff
@@ -166,34 +167,69 @@ peculiar_fragment_match_compatible (const AST::MacroMatchFragment &last_match,
...
-	  FOR,
-	  IMPL,
-	  FN_TOK,
-	  UNSAFE,
-	  TYPEOF,
-	  DYN
-	  // FIXME: Add Non kw identifiers
-	  // FIXME: Add $crate as valid
-	}}};
+	{COMMA,
```
CohenArthur, line 170: Yeah, sadly there's not really a nice way to do this other than listing the keywords. We could maybe clean this up a little bit by having a separate `std::vector` elsewhere in the file. But this is very nitpicky and I think this is good :)

CohenArthur (approved):
LGTM! Thank you :)

CohenArthur:
@tage64 before I merge this, could you refactor all 4 of your commits into a single one describing the change and adding the test? I think it makes sense for a tiny change like that to only be one commit :)

tage64:
I have squashed my commits and added a short description in the commit message.
Please comment though should you have any complaints about the description, changelog entries, or anything else. This is the first time I'm contributing to gccrs, so it is good if I adhear to the best practices right from the beginning.

CohenArthur:
The commit looks good to me @tage64 :) your commit description is precise and good. Thanks!

PR #1966, opened 2023-03-04
parser: Fix handling of multiple left angles in null denotation

### LINKED ISSUE #1880: Invalid parsing of complex qualified path in expression
I tried this code:

```rust
fn main() {
    let _ = <<F as Foo>::A as Bar>::f();
}
```

I expected to see this happen: no parsing error. `rustc` with `-Z parse-only` returns 0

Instead, this happened: 

```rust
test.rs:2:13: error: found unexpected token ‘<<’ in null denotation
    2 |     let _ = <<F as Foo>::A as Bar>::f();
      |             ^~
test.rs:2:15: error: failed to parse expression in let statement
    2 |     let _ = <<F as Foo>::A as Bar>::f();
      |               ^
test.rs:2:5: error: failed to parse statement or expression without block in block expression
    2 |     let _ = <<F as Foo>::A as Bar>::f();
      |     ^~~
test.rs:3:1: error: unrecognised token ‘}’ for start of item
    3 | }
      | ^
test.rs:3:1: error: failed to parse item in crate
```

or

```rust
test.rs:2:35: error: found null denotation scope resolution operator, and have not written handling for it
    2 |     let _ = <<F as Foo>::A as Bar>::f();
      |                                   ^~
test.rs:2:37: error: failed to parse expression in let statement
    2 |     let _ = <<F as Foo>::A as Bar>::f();
      |                                     ^
test.rs:2:5: error: failed to parse statement or expression without block in block expression
    2 |     let _ = <<F as Foo>::A as Bar>::f();
      |     ^~~
test.rs:3:1: error: unrecognised token ‘}’ for start of item
    3 | }
      | ^
test.rs:3:1: error: failed to parse item in crate
```

with basic handling of `<<` in `null_denotation`

### Meta

- What version of Rust GCC were you using: 1bd1e466453

goar5670:
Fixes #1880

### DIFF
--- a/gcc/testsuite/rust/compile/complex_qualified_path_in_expr.rs
+++ b/gcc/testsuite/rust/compile/complex_qualified_path_in_expr.rs
@@ -1,0 +1,34 @@
+struct S; // { dg-warning "struct is never constructed" "" { target *-*-* } .-1 }
+impl S {
+  fn f() -> i32 { return 0; }
+  // { dg-warning "associated function is never used" "" { target *-*-* } .-1 }
+}
+
+trait T1 {
+  type A;
+  fn f() -> i32 { return 0; }
+}
+
+impl T1 for S {
+  type A = S;
+}
+
+trait T2 {
+  type B;
+  fn f() -> i32 { return 0; }
+}
+
+impl T2 for S {
+  type B = S;
+}
+
+trait T3 {
+  fn f() -> i32 { return 0; }
+}
+
+impl T3 for S {}
+
+fn main() {
+  let _ = <<S as T1>::A as T2>::f();
+  let _ = <<<S as T1>::A as T2>::B as T3>::f();
+}
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -12650,6 +12650,13 @@ Parser<ManagedTokenSource>::parse_expr (int right_binding_power,
 	  || id == RIGHT_SQUARE)
 	return nullptr;
     }
+
+  if (current_token->get_id () == LEFT_SHIFT)
+    {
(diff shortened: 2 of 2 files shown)


gcc/testsuite/rust/compile/complex_qualified_path_in_expr.rs
```diff
@@ -0,0 +1,6 @@
+// { dg-additional-options "-fsyntax-only" }
```
philberty, line 1: have you tried this without -fsyntax-only? I merged in https://github.com/Rust-GCC/gccrs/pull/1954 which fixes issues there

philberty (changes requested):
I would like to see the test-case updated to not use -fsyntax-only or if there is an issue for an issue to be raised

gcc/testsuite/rust/compile/complex_qualified_path_in_expr.rs
goar5670: Good point. I tried it as per your suggestion and it exposed a couple of other issues.
1. `let _ = <<S as T1>::A as T2>::f()` failing: This was caused by `rust-hir-path.h`'s `gcc_unreachable ()` at line 948 in combination with the missing initialization for `associated_segment` in `HIR::QualifiedPathInType`'s constructor (`rust-hir-path.h` line 941) I'm guessing it wasn't intentional, right?
2. After fixing 1, `let _ = <<<S as T1>::A as T2>::B as T3>::f();` failed: This I couldn't solve completely but I'll share what I understood. `ResolveRelativeQualTypePath::resolve_qual_seg` expects a resolved node id to be returned which is not applicable if the the `Type` (as in reference to < Type as TypePath >) in the `QualifiedPathType` is a `QualifiedPathInType`. I'm not sure what the correct way of handling this is but would like to hear your input.

philberty (approved):
LGTM

PR #1970, opened 2023-03-06
Fix handling of coherence related lang_items

### LINKED ISSUE #1896: unknown lang item i32
I tried this code:

```rust
#[lang = "i32"]
impl i32 {}
```

Instead, this happened: 

```
test.rs:12:3: error: unknown lang item
   12 | #[lang = "i32"]
      |   ^~~~

```

### Meta

- What version of Rust GCC were you using, git sha if possible. 87e7c9dedacd4d12636e6c626aab14bcecf686f2

powerboat9:
Fixes #1896. I don't think we have coherence checks yet, but the test should detect regressions.


philberty (changes requested):
I am not sure i like this approach yet for two reasons:

1. We are lossing the ability to lookup the const_ptr lang item and we still cant lookup the i32 langitem for example
2. In rustc they have lang items defined on the primitive impl blocks how do we implement a check that lang item i32 has already been defined if we remove all of this?

We don't need to copy Rustc and i dont like jumping ahead when we dont have any test cases for this area either.

I would prefer that we just add the mappings for each of the lang items instead of grouping them like this because we don't know how much we dont know in order to implement the rest of libcore.

powerboat9:
@philberty Fixed

philberty (approved):
LGTM thanks for updating this

dkm:
little nit: your commit is missing the SIgned-Off-By. Not a big deal, but would probably make the task of pushing all these in gcc easier if it was systematically there :)

powerboat9:
@dkm Ah, sorry. I was trying out a new rebase-heavy workflow, and forgot to sign off on the final comment.

PR #1971, opened 2023-03-06
typecheck: Add basic typechecking for ReferenceType

goar5670:
provide basic resolve and typechecking for `ReferencePattern`.

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-pattern.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-pattern.cc
@@ -410,8 +410,13 @@ TypeCheckPattern::visit (HIR::QualifiedPathInExpression &pattern)
 void
 TypeCheckPattern::visit (HIR::ReferencePattern &pattern)
 {
-  rust_sorry_at (pattern.get_locus (),
-		 "type checking qualified path patterns not supported");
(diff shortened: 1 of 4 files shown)


powerboat9:
@goar5670 the clang format check failed

gcc/rust/typecheck/rust-hir-type-check-pattern.cc
```diff
@@ -410,8 +410,13 @@ TypeCheckPattern::visit (HIR::QualifiedPathInExpression &pattern)
 void
 TypeCheckPattern::visit (HIR::ReferencePattern &pattern)
 {
-  rust_sorry_at (pattern.get_locus (),
-		 "type checking qualified path patterns not supported");
+  if (parent->get_kind () != TyTy::TypeKind::REF)
+    rust_error_at (pattern.get_locus (), "expected %s, found reference",
```
philberty, line 414: You could add a dg-error test case for this

philberty (changes requested):
Seems good to me but lets add a test case for this

gcc/rust/typecheck/rust-hir-type-check-pattern.cc
goar5670: Done 👍 please recheck

philberty (approved):
LGTM

PR #1974, opened 2023-03-08
hir: Provide basic handling for `ReferencePattern`

goar5670:
Added an implementation for
`CompilePatternBindings::visit (HIR::ReferencePattern)` where we
dereference the initial expression and recurse.
Added an implementation for
`CompilePatternBindings::visit (HIR::IdentifierPattern)` as well since it's
the simplest base case.

In addition to this, a small refactor for the shared code in
`StructPattern` and `TupleStructPattern` visits was added as a helper
function called `create_tmp_param_var`.

Addresses #1813


gcc/testsuite/rust/compile/ref_pattern_fn_param2.rs
```diff
@@ -0,0 +1,12 @@
+#[lang = "fn_once"]
+pub trait FnOnce<Args> {
```
philberty, line 2: you don't need the fn_once lang item for this.

gcc/testsuite/rust/compile/ref_pattern_fn_param2.rs
```diff
@@ -0,0 +1,12 @@
...
+    #[lang = "fn_once_output"]
+    type Output;
+
+    extern "rust-call" fn call_once(self, args: Args) -> Self::Output;
+}
+
+fn foo(&b: &i32) -> bool {
```
philberty, line 9: Can you write another function to call foo?

gcc/rust/backend/rust-compile-fnparam.cc
```diff
@@ -117,5 +120,23 @@ CompileSelfParam::compile (Context *ctx, tree fndecl, HIR::SelfParam &self,
...
+tree
+CompileFnParam::create_tmp_match_scrutinee (tree decl_type)
+{
+  // generate the anon param
+  tree tmp_ident = create_tmp_var_name ("RSTPRM");
+  std::string cpp_str_identifier = std::string (IDENTIFIER_POINTER (tmp_ident));
+
+  decl_type = ctx->get_backend ()->immutable_type (decl_type);
+  compiled_param
+    = ctx->get_backend ()->parameter_variable (fndecl, cpp_str_identifier,
```
philberty, line 135: That likely wont work for match expressions because we dont need parameters

philberty (changes requested):
Hmm this is hard to review i cant see what you are trying to do. This is why commit messages are _really_ important.

It would also be important for you to show -fdump-tree-gimple output for this in the PR comments to show what the result of this is. As far as I can tell the reference pattern is syntatic sugar to enforce the parameter or pattern binding to be a reference. It might even coerce to a reference as match binding not sure about that though.

gcc/rust/backend/rust-compile-fnparam.cc
goar5670: If I understand correctly, this part will never be reached by a `match` expression. It's only called from a `CompileFnParam::visit` which in turn is only called in function compilation and intrinsic functions. However, the naming for the function itself might be misleading since it has nothing to do with `match`, it was a misunderstanding on my part, I apologize.

Also, I think `StructPattern` and `TupleStructPattern` are not supported in match yet. We have no tests for them (only enums) and when I tried a test, it resulted in an ICE in the type-checking stage.

goar5670:
I'm sorry for not clarifying, I'll write a detailed commit message, and here's the gimple output, followed by the updated test case (after adding a function call).
```
__attribute__((cdecl))
void test::main ()
{
  const bool a;

  a = test::foo (0);
}


__attribute__((cdecl))
bool test::foo (const i32 b)
{
  bool D.88;

  D.88 = b == 0;
  return D.88;
}
```

```rust
fn foo(&a: &i32) -> bool {
  b == 0
}

fn main() {
  let a = foo(&0);
}
```

philberty:
> 

Your gimple output looks wrong to me. For example take a look at this: https://godbolt.org/z/j7aTbG6M1

This rust function

```
fn test(x: &i32) -> bool {
    *x == 0
}
```

Is gimplified into:

```
__attribute__((cdecl))
bool example::test (const i32 & const x)
{
  bool D.86;

  _1 = *x;
  D.86 = _1 == 0;
  return D.86;
}
```

So you can see the parameter in gimple is a reference type.

In your example:

```
fn foo(&b: &i32) -> bool {
  b == 0
}
```
With your gimple output as:

```
__attribute__((cdecl))
bool ref_pattern_fn_param2::foo (const i32 b)
{
  bool D.88;

  D.88 = b == 0;
  return D.88;
}
```

Your parameter of 'b' should still be a reference, but is not an i32 here other wise the function ABI is broken. As far as I can tell the reference pattern binding is about creating a sneaky way to destructure types like we have some support for in match expressions. So this example should really be creating a parameter of &i32 but creating one of those implicit_pattern_bindings to the *b which is an indirection expression.

goar5670:
@philberty what about this gimple output, does it seem correct?
```
__attribute__((cdecl))
void test::main ()
{
  i32 D.88;

  _1 = 0;
  D.88 = _1;
  RUSTTMP.2 = test::foo (&D.88);
}


__attribute__((cdecl))
bool test::foo (const i32 & const RSTPRM.0)
{
  bool D.89;

  _1 = *RSTPRM.0;
  D.89 = _1 == 0;
  return D.89;
}
```

philberty (changes requested):
This looks really really good i think it would be important to add another test case for example in gcc/testsuite/rust/execute/torture/ref-pattern1.rs

```
fn foo (&a: &i32, b: i32) -> i32 {
  a + b
}

fn main() -> i32 {
  let a = 4;
  foo(&a, 2) - 6
}
```

This should ensure that this returns zero and it derefs properly

goar5670:
> This looks really really good i think it would be important to add another test case for example in gcc/testsuite/rust/execute/torture/ref-pattern1.rs
> 
> ```
> fn foo (&a: &i32, b: i32) -> i32 {
>   a + b
> }
> 
> fn main() -> i32 {
>   let a = 4;
>   foo(&a, 2) - 6
> }
> ```
> 
> This should ensure that this returns zero and it derefs properly

Good point! added it.

philberty (approved):
LGTM

PR #1976, opened 2023-03-11
Impl `Rust::HIR::Dump::visit`

chosen-ox:
In my opinion, `Session::dump_hir` is to dump items in one crate using `items->as_string()`  while `Session::dump_hir_pretty` is to dump items with identation using `items->accept_vis()`. So the diff between two dump methods are mainly identation (e.g. hierarchy)? (If it's incorrect, please tell me the right one). 
My plan is implementing some `HIR::Dump::visit()` based on `as_string()`.  In first commit, I dump inner attrs for `Dump::visit(BlockExpr &)` according to [BlockExpr::as_string()](https://github.com/Rust-GCC/gccrs/blob/2c1e7b55dab5564f92d890c4953f9fd387bb30d7/gcc/rust/hir/tree/rust-hir.cc#L929-#L943). 

I wonder if I am on the right track. Please give some suggestions. Thank you!

### DIFF
--- a/gcc/rust/hir/rust-hir-dump.cc
+++ b/gcc/rust/hir/rust-hir-dump.cc
@@ -238,7 +238,22 @@ Dump::visit (BlockExpr &block_expr)
   stream << "BlockExpr: [";
   indentation.increment ();
   stream << std::endl;
+
   // TODO: inner attributes
+  stream << std::string (indent, indent_char);
+  stream << "inner attributes: ";
+  if (!block_expr.inner_attrs.empty ())
+    {
+      for (const auto &attr : block_expr.inner_attrs)
+	{
+	  stream << std::endl;
+	  stream << std::string (indent, indent_char);
+	  stream << attr.as_string ();
+	  // stream << attr.accept_vis(*self);
+	}
+    }
+
+  stream << std::endl;
 
   // statements
   if (block_expr.has_statements ())
--- a/gcc/rust/hir/rust-hir-dump.cc
+++ b/gcc/rust/hir/rust-hir-dump.cc
@@ -51,7 +51,9 @@ Dump::go (HIR::Crate &crate)
       stream << std::endl;
       item->accept_vis (*this);
     }
+  stream << std::endl;
   stream << indentation;
+
   stream << "]," << std::endl;
   indentation.decrement ();
   //
@@ -66,8 +68,39 @@ Dump::go (HIR::Crate &crate)
 }
 
 void
-Dump::visit (Lifetime &)
-{}
+Dump::visit (AST::Attribute &attribute)
+{
+  std::string path_str = attribute.get_path ().as_string ();
+  stream << path_str;
+  if (attribute.has_attr_input ())
+    stream << attribute.get_attr_input ().as_string ();
+}
+
+void
+Dump::visit (Lifetime &lifetime)
+{
+  if (lifetime.is_error ())
+    {
+      stream << "error lifetime";
+      return;
+    }
+
+  switch (lifetime.get_lifetime_type ())
+    {
+    case AST::Lifetime::LifetimeType::NAMED:
+      stream << "'" << lifetime.get_name ();
+      break;
+    case AST::Lifetime::LifetimeType::STATIC:
+      stream << "'static";
+      break;
+    case AST::Lifetime::LifetimeType::WILDCARD:
+      stream << "'_";
+      break;
+    default:
+      stream << "ERROR-MARK-STRING: lifetime type failure";
+      break;
+    }
+}
 void
 Dump::visit (LifetimeParam &)
 {}
@@ -235,44 +268,48 @@ Dump::visit (ClosureExpr &)
 void
 Dump::visit (BlockExpr &block_expr)
 {
-  stream << "BlockExpr: [";
-  indentation.increment ();
-  stream << std::endl;
+  stream << "BlockExpr: [\n";
 
+  indentation.increment ();
   // TODO: inner attributes
-  stream << std::string (indent, indent_char);
-  stream << "inner attributes: ";
   if (!block_expr.inner_attrs.empty ())
     {
-      for (const auto &attr : block_expr.inner_attrs)
+      stream << indentation << "inner_attrs: [";
+      indentation.increment ();
+      for (auto &attr : block_expr.inner_attrs)
 	{
-	  stream << std::endl;
-	  stream << std::string (indent, indent_char);
-	  stream << attr.as_string ();
-	  // stream << attr.accept_vis(*self);
+	  stream << "\n";
+	  stream << indentation;
+	  visit (attr);
 	}
+      indentation.decrement ();
+      stream << "\n" << indentation << "]\n";
     }
 
-  stream << std::endl;
-
   // statements
+  // impl null pointer check
+
   if (block_expr.has_statements ())
     {
       auto &stmts = block_expr.get_statements ();
       for (auto &stmt : stmts)
 	{
 	  stream << indentation << "Stmt: {\n";
-	  // stream << indentation;
 	  stmt->accept_vis (*this);
 	  stream << "\n";
 	  stream << indentation << "}\n";
 	}
     }
 
-  // // TODO: print tail expression if exists
+  // final expression
+  if (block_expr.has_expr ())
+    {
+      stream << indentation << "final expression:";
+      stream << "\n" << indentation << block_expr.expr->as_string ();
+    }
 
   indentation.decrement ();
-  stream << indentation << "]";
+  stream << "\n" << indentation << "]";
 }
 
 void
--- a/gcc/rust/hir/rust-hir-dump.h
+++ b/gcc/rust/hir/rust-hir-dump.h
@@ -37,6 +37,7 @@ private:
   Indent indentation;
   std::ostream &stream;
 
+  void visit (AST::Attribute &attribute);
   virtual void visit (Lifetime &) override;
   virtual void visit (LifetimeParam &) override;
   virtual void visit (PathInExpression &) override;


philberty (changes requested):
I think you are on the right track but avoid using as_strings, its better to keep using the visitor pattern here so it becomes more reusable. Please take a look at how @jdupak did it

philberty:
@jdupak ^^ can you review this

jdupak:
Sure, I will look at it on Monday.

jdupak:
**Review of the code changes:**
Quick search shows, that there is already one instance of inner attribute dumping on line 35 so we should compare it with that:

- The "inner attributes: " string should only be dumped when not empty - move in inside the if.
- Same goes for the final end of line.
- The list should be delimited by square brackets.
- There is no end of line printed between the items in the existing instance.

For the individual attributes, you should add a new overload of Dump::visit and use it to dump it. You might notice that the attribute type is shared between AST and HIR, so using the as_string method would be confusing as the representation might differ between the two dumps.

Little style thing - most of the code uses '\n' instead of std::endl. It is more concise and easier to read and c++ should substitute it correctly.

Since there are two instances of inner attr dump already, it might be best to extract it into a separate method (I would call it `visit_inner_attrs` because outer attrs have the same type).

jdupak:
> In my opinion, `Session::dump_hir` is to dump items in one crate using `items->as_string()` while `Session::dump_hir_pretty` is to dump items with identation using `items->accept_vis()`. So the diff between two dump methods are mainly identation (e.g. hierarchy)? (If it's incorrect, please tell me the right one). My plan is implementing some `HIR::Dump::visit()` based on `as_string()`. In first commit, I dump inner attrs for `Dump::visit(BlockExpr &)` according to [BlockExpr::as_string()](https://github.com/Rust-GCC/gccrs/blob/2c1e7b55dab5564f92d890c4953f9fd387bb30d7/gcc/rust/hir/tree/rust-hir.cc#L929-#L943).

@philberty could you clarify, how the multiple hir dumps are supposed to work (coexist/replace each other...)?

Anyway, items->accept_vis() is the preferred one with wrapper template method Dump::visit():
```c++
template <typename T>
void
Dump::visit (std::unique_ptr<T> &node)
{
  node->accept_vis (*this);
}

template <typename T>
void
Dump::visit (T &node)
{
  node.accept_vis (*this);
}
```
This way, you can add overloads of the Dump::visit method for objects not implementing accept_vis. This way the syntax is unified and you can reuse the helpers like `visit_items_joined_by_separator` to print lists.

Please tell me if you need to further clarify this.

jdupak:
@philberty We should also unify the indentation approach.

chosen-ox:
> **Review of the code changes:** Quick search shows, that there is already one instance of inner attribute dumping on line 35 so we should compare it with that:
> 
For sure, that one is to dump Crate. It's easy to follow.
> * The "inner attributes: " string should only be dumped when not empty - move in inside the if.
> * Same goes for the final end of line.
> * The list should be delimited by square brackets.
> * There is no end of line printed between the items in the existing instance.
> 


> For the individual attributes, you should add a new overload of Dump::visit and use it to dump it. You might notice that the attribute type is shared between AST and HIR, so using the as_string method would be confusing as the representation might differ between the two dumps.
> 
In my observation, AST and HIR has different `as_string()` methods distinguished by namespace. They have different implementations. BTW, in `rust-hir-dump.cc`, many `visit` functions use `as_string()`. For example, 
```rust
Dump::visit (LiteralExpr &literal_expr)
{
  stream << literal_expr.get_literal ().as_string () << " "
	 << literal_expr.get_mappings ().as_string ();
}
```
It use `literal ().as_string ()`. We should clarify which items are dumped by `visit` and which items are dumped by `as_string()`. 
> Little style thing - most of the code uses '\n' instead of std::endl. It is more concise and easier to read and c++ should substitute it correctly.
> 
Yes, I found that problem too. `Dump::visit (BlockExpr &block_expr)` mixes std::endl with "\n". So I follow the convention. I will modify it.
> Since there are two instances of inner attr dump already, it might be best to extract it into a separate method (I would call it `visit_inner_attrs` because outer attrs have the same type).
>
Sure.

jdupak:
> In my observation, AST and HIR has different as_string() methods distinguished by namespace. They have different implementations. BTW, in rust-hir-dump.cc, many visit functions use as_string().

Usually yes, but if you look at the rust-hir.h:214 - the `AttrVec` is shared. The idea is to keep all the logic related to dumping in the specialized file, not with the object itself. You can find a lot of examples doing it otherwise as all the is WIP and it was done differently in the past.

In special cases, where it is just printing literal with no special logic, as_string is probably OK.

chosen-ox:
> Usually yes, but if you look at the rust-hir.h:214 - the `AttrVec` is shared. The idea is to keep all the logic related to dumping in the specialized file, not with the object itself. You can find a lot of examples doing it otherwise as all the is WIP and it was done differently in the past.
> 
> In special cases, where it is just printing literal with no special logic, as_string is probably OK.
>
woops! I probably get what you mean. I traced `AttrVec` and find it's a vector whose member type is `Attribute`. `Attribute` belongs to `AST`. So its `as_string()` function is for dumping as AST. You're right. I should add a new overload of `Dump::visit(Attribute &)`. But I can keep the `AttrVec` shared as it's just a vector.
Do I totally understand? Feel free to correct me! :)

jdupak:
> > Usually yes, but if you look at the rust-hir.h:214 - the `AttrVec` is shared. The idea is to keep all the logic related to dumping in the specialized file, not with the object itself. You can find a lot of examples doing it otherwise as all the is WIP and it was done differently in the past.
> > In special cases, where it is just printing literal with no special logic, as_string is probably OK.
> 
> woops! I probably get what you mean. I traced `AttrVec` and find it's a vector whose member type is `Attribute`. `Attribute` belongs to `AST`. So its `as_string()` function is for dumping as AST. You're right. I should add a new overload of `Dump::visit(Attribute &)`. But I can keep the `AttrVec` shared as it's just a vector.
> 
> This is indented for possible diff of dumping `attribute` between AST and HIR and for a more beautiful code and structure.
> 
> Do I totally understand? Feel free to correct me! :)

Yes, sounds right to me. :+1:

jdupak:
The unified indent approach is now in master. Please rebase :)

chosen-ox:
@jdupak Hi, there's a lot of things to do to fix compile error. Now I add `Dump::visit (AST::Attribute &attribute)`. Could you review it and give some suggestions? If everything goes well, I would rebase and then request Philip to review the code.

chosen-ox:
> The unified indent approach is now in master. Please rebase :)

Rebased :)

gcc/rust/ast/rust-ast.h
```diff
@@ -32,6 +32,10 @@ typedef std::string Identifier;
 typedef int TupleIndex;
 struct Session;
 struct MacroExpander;
+// forward declaration Dump to pass the compilation
+namespace HIR {
+class Dump;
```
jdupak, line 37: This "contamination" is not great, but it is the best we can do right now.

gcc/rust/hir/rust-hir-dump.cc
```diff
@@ -237,10 +270,25 @@ Dump::visit (BlockExpr &block_expr)
 {
   stream << "BlockExpr: [";
   indentation.increment ();
-  stream << std::endl;
+  stream << "\n";
+
   // TODO: inner attributes
+  stream << indentation;
```
jdupak, line 276: At least for me, it seems more readable to put those on one line to better correspond to the output format.

gcc/rust/hir/rust-hir-dump.cc
```diff
@@ -237,10 +270,25 @@ Dump::visit (BlockExpr &block_expr)
...
+  stream << "inner attributes: ";
+  if (!block_expr.inner_attrs.empty ())
```
jdupak, line 278: Is this correct, when inner_attrs is empty?

gcc/rust/hir/rust-hir-dump.cc
```diff
@@ -237,10 +270,25 @@ Dump::visit (BlockExpr &block_expr)
...
+    {
+      for (auto &attr : block_expr.inner_attrs)
+	{
+	  stream << "\n";
+	  stream << indentation;
+	  // stream << attr.as_string ();
```
jdupak, line 284: Once you are doing the final cleanup, do not leave dead-code behind. ;)

jdupak (changes requested):
This mostly looks good.
- The mixing of AST and HIR is not nice but seems to be the best we can do, so OK.
- I have doubts that the new lines are printed correctly for empty lists (some or all) (**this is the main problem I see**).
- Other than that there are just some style details.

gcc/rust/hir/rust-hir-dump.cc
chosen-ox: Finished

gcc/rust/hir/rust-hir-dump.cc
chosen-ox: Sure, I fixed it :)

gcc/rust/ast/rust-ast.h
chosen-ox: Refactor is neccessary in the future.

chosen-ox:
> This mostly looks good.
> 
> * The mixing of AST and HIR is not nice but seems to be the best we can do, so OK.
> * I have doubts that the new lines are printed correctly for empty lists (some or all) (**this is the main problem I see**).
> * Other than that there are just some style details.

Thank you for your patience. I have fixed all. Have a good day!

CohenArthur (commented):
Mh, I am really not a fan of us having to include HIR bits within the AST. I think this will have an effect on compilation speed overall. Is there really no other way to go about it? Especially since the `Dump::visit(AST::Attribute&)` implementation is exactly the same as `AST::Attribute::as_string()`. I don't know which would be cleaner

jdupak:
> Mh, I am really not a fan of us having to include HIR bits within the AST. I think this will have an effect on compilation speed overall. Is there really no other way to go about it? Especially since the `Dump::visit(AST::Attribute&)` implementation is exactly the same as `AST::Attribute::as_string()`. I don't know which would be cleaner

@CohenArthur There are multiple ways. The problem here is that there is an overlap between AST and HIR nodes and therefore for visitor it need to have a vtable pointer for both AST and HIR dump. Any of these will remove the problem:
- Eliminate the overlap - i.e. duplicate the node for AST and HIR.
- Do not use visitor for it - use an overload of the visit method - which will work as it is not used via inheritance.

Now looking at the code, it does not use visitor even in the AST dump, so the second option seems like a natural solution.

gcc/rust/hir/rust-hir-dump.cc
```diff
@@ -236,25 +269,43 @@ void
...
   // TODO: inner attributes
+  if (!block_expr.inner_attrs.empty ())
+    {
+      stream << "\n";
+      stream << indentation << "inner_attrs: [";
+      for (auto &attr : block_expr.inner_attrs)
+	{
+	  stream << "\n";
+	  stream << indentation;
+	  attr.accept_vis (*this);
```
jdupak, line 283:
```suggestion
	  visit(attr);
```

gcc/rust/ast/rust-ast.cc
```diff
@@ -26,6 +26,7 @@ along with GCC; see the file COPYING3.  If not see
 #include "rust-lex.h"
 #include "rust-parse.h"
 #include "rust-operators.h"
+#include "rust-hir-dump.h"
```
jdupak, line 29:
```suggestion
```

gcc/rust/ast/rust-ast.cc
```diff
@@ -176,6 +177,12 @@ Attribute::as_string () const
     return path_str + attr_input->as_string ();
 }
 
+void
+Attribute::accept_vis (HIR::Dump &vis)
+{
+  vis.visit (*this);
+}
+
```
jdupak, line 185:
```suggestion
```

gcc/rust/ast/rust-ast.h
```diff
@@ -32,6 +32,10 @@ typedef std::string Identifier;
...
+}
```
jdupak, line 38:
```suggestion
```

gcc/rust/ast/rust-ast.h
```diff
@@ -545,6 +549,8 @@ struct Attribute
   std::string as_string () const;
 
   // no visitor pattern as not currently polymorphic
+  // virtual for future subclass
+  virtual void accept_vis (HIR::Dump &);
```
jdupak, line 553:
```suggestion
```

gcc/rust/hir/rust-hir-dump.h
```diff
@@ -32,6 +32,7 @@ class Dump : public HIRFullVisitor
 public:
   Dump (std::ostream &stream);
   void go (HIR::Crate &crate);
+  virtual void visit (AST::Attribute &attribute);
```
jdupak, line 35:
```suggestion
  void visit (AST::Attribute &attribute);
```

Also should be private.

jdupak (commented):
This is what needs to be done to resolve the HIR/AST "contamination".

gcc/rust/hir/rust-hir-dump.cc
```diff
@@ -236,25 +269,43 @@ void
...
 {
   stream << "BlockExpr: [";
+
   indentation.increment ();
-  stream << std::endl;
   // TODO: inner attributes
+  if (!block_expr.inner_attrs.empty ())
+    {
+      stream << "\n";
+      stream << indentation << "inner_attrs: [";
```
jdupak, line 278: The '[' seems not to be terminated.

gcc/rust/hir/rust-hir-dump.cc
```diff
@@ -236,25 +269,43 @@ void
...
+	{
+	  stream << "\n";
+	  stream << indentation;
+	  visit (attr);
+	}
+    }
 
   // statements
+  // impl null pointer check
+  stream << "\n";
```
jdupak, line 289: I think this should be before the brace on line 285 in case there are no attributes.

jdupak (changes requested):
Almost there :) Style and structure is good. It would be best to compile some examples and check the output (and post it here).

gcc/rust/hir/rust-hir-dump.cc
chosen-ox: In my opinion. If there's no attrs. Stmts will dump in a new line instead of the line with "BlockExpr: [".

gcc/rust/hir/rust-hir-dump.cc
chosen-ox: > Almost there :) Style and structure is good. It would be best to compile some examples and check the output (and post it here).

I will try. But I will finish my homework first 😭.

gcc/rust/hir/rust-hir-dump.cc
jdupak: You are right, it was just a little hard to see. Moving line 277 before the if (after 247) and line 289 after 284 would do the same (right?) and would be easier to follow.

gcc/rust/hir/rust-hir-dump.cc
jdupak: No worries. I am just procrastinating my homework by giving this review :lol:

gcc/rust/hir/rust-hir-dump.cc
chosen-ox: I add "\n" behind {} and []. Now seems more clear.

jdupak: Thanks! Much better.

chosen-ox:
I compile one example. Get the right dump. LGTM. But though I add `#![allow(unused_variables)]`, compiler still told me that `test.rs:3:13: warning: unused name ‘i’ `. This might be a bug?


```rust
// test.rs
fn is_unix_platform() -> bool {
        #![allow(unused_variables)]
        let i = 5;
        false
}
fn main() {
    // Statements here are executed when the compiled binary is called.

    // Print text to the console.
}
```

```
Crate {
        items: [
                Function {
                        func_name: is_unix_platform,
                        return_type: bool,
                        BlockExpr: [
                                inner_attrs: [
                                allow((unused_variables))
                                ]
                                Stmt: {
                                        LetStmt: {
                                                i = Expr: {
                                                        5 [C: 0 Nid: 11 Hid: 23]
                                                        }
                                        }

                                }
                                final expression:
                                ( false ([C: 0 Nid: 13 Hid: 25]))
                        ]
                        node_mappings: [C: 0 Nid: 15 Hid: 27 Lid: 2]
                }

                Function {
                        func_name: main,
                        return_type: void,
                        BlockExpr: [
                        ]
                        node_mappings: [C: 0 Nid: 19 Hid: 29 Lid: 3]
                }

        ],
        node_mappings: [C: 0 Nid: 20 Hid: 30]
}
```

gcc/rust/hir/rust-hir-dump.cc
```diff
@@ -235,29 +268,46 @@ Dump::visit (ClosureExpr &)
...
-  stream << "BlockExpr: [";
+  stream << "BlockExpr: [\n";
+
   indentation.increment ();
-  stream << std::endl;
   // TODO: inner attributes
+  if (!block_expr.inner_attrs.empty ())
+    {
+      stream << indentation << "inner_attrs: [";
+      for (auto &attr : block_expr.inner_attrs)
```
jdupak, line 278: Just add indentation increment and decrement around the for loop (see the example).

jdupak (approved):
LGTM with one exception

gcc/rust/hir/rust-hir-dump.cc
chosen-ox: Fixed.

jdupak:
Good job!

gcc/rust/hir/rust-hir-dump.cc
chosen-ox: > Just add indentation increment and decrement around the for loop (see the example).

I have updated the [new dump](https://github.com/Rust-GCC/gccrs/pull/1976#issuecomment-1468716299).

chosen-ox:
> Good job!

Thank you for your guide and teach~

CohenArthur:
> But though I add #![allow(unused_variables)], compiler still told me that test.rs:3:13: warning: unused name ‘i’ 

Yeah, we don't handle that attribute yet :) no worries, it's got nothing to do with your PR

CohenArthur (approved):
Looks good to me @chosen-ox :) thanks for spending so much time on this PR!

chosen-ox:
> 

My pleasure. Thank you all!

philberty:
> 

Sorry for late reply @jdupak we should be removing the old AST and HIR dumps so that your work and this one become the only AST and HIR dump. The typecheck one needs to be deleted as well I will raise issues for this

PR #1978, opened 2023-03-11
Unify AST and HIR dump indentation

jdupak:
Relates to #1976 (#1976 should be rebased before merging)

make check-rust passed on every commit

Please check the changelog - I have done it for the first time.

### DIFF
--- a/gcc/rust/util/rust-dump.h
+++ b/gcc/rust/util/rust-dump.h
@@ -1,0 +1,51 @@
+// Copyright (C) 2021-2023 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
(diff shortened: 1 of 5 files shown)


gcc/rust/util/rust-dump.h
```diff
@@ -0,0 +1,49 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+// Common definitions useful for textual dump of IRs (AST and HIR).
+
+#ifndef RUST_DUMP_H
+#define RUST_DUMP_H
+
```
philberty, line 23: This header seems to depend on rust-system you should probably add it here.

philberty (approved):
LGTM

gcc/rust/util/rust-dump.h
jdupak: Fixed

PR #1979, opened 2023-03-11
ci: check Signed-off-by lines (and others) are at the end

dkm:
Check Signed-off-by/Reviewed-by/Co-authored-by lines are all grouped and last in commit log.

### DIFF
--- a/.github/workflows/commit-format.yml
+++ b/.github/workflows/commit-format.yml
@@ -91,6 +91,22 @@ jobs:
             fi
           done;
 
-          exit $retval;
+          ## check no Signed-off-by in the middle.
+          while read -r f; do
+            echo -n "$f : "
+               if awk 'BEGIN { in_footer = 0; error = 0;}
+                       { DEFAULT = 1; }
+                       /^Signed-off-by/ { in_footer = 1; DEFAULT = 0;}
+                       /^Co-authored-by/ { in_footer = 1; DEFAULT = 0;}
+                       /^Reviewed-by/  { in_footer = 1; DEFAULT = 0;}
+                       /^\s*$/  { DEFAULT = 0; }
+                       DEFAULT { if (in_footer == 1) { printf $0 " appears after some Signed-off-by/Co-authored-by line "; error = 1; exit (1);} }';
+               then
+                   echo OK
+               else
+                   echo KO
+                   retval=1
+               fi < <(git log "$f" -1 --format="%B")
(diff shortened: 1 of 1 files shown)


dkm:
This is a test/proposal. I've seen that several commits have SoB lines before changelog entries, which is unusual and should probably be avoided.

We also have commits with duplicated SoB lines for the same author but with different names (pseudo + real name), or still SoB with pseudo, but this is hardly checked automatically...

mvvsmk:
Hey @dkm, I don't know how helpful this observation is, but I noticed that whenever we use `git gcc-commit-mklog -s`, it adds my Sob before the changelog, which could also explain its recurrence in multiple commits.

And the current CI check it by testing the last commit with `git gcc-verify` , which passes if the Sob is before the changelog. I didn't look into this further, but it seems that `git gcc-verify` only checks for the presence of a Sob and a changelog, not in what orientation they are placed. :)

dkm:
Yes, that's why. The SoB use has been introduced rather recently in gcc (see DCO introduction) and most scripts are not really handling this corner case (that's still very low prio). The changelog entries are simply added at the end of the existing log...

But I think having the SoB lines at the end of the commits will help when it comes to review commits before merging them in gcc... So I'm experimenting :)

.github/workflows/commit-format.yml
```diff
@@ -91,6 +91,23 @@ jobs:
             fi
           done;
 
+          ## check no Signed-of-by in the middle.
```
CohenArthur, line 94:
```suggestion
          ## check no Signed-off-by in the middle.
```

CohenArthur (approved):
I'll trust you on the shell script side of things :P I don't personally care about SoB lines being in the middle of the commit, especially since this is what `git gcc-commit-mklog -s` does by default, but I agree that it looks better at the end and I'm okay with enforcing it. We could probably make some changes to the `gcc-commit-mklog` command to take that into account too :D

.github/workflows/commit-format.yml
dkm: I think the gcc-commit-mklog is not handling this because the SoB is still rarely used in GCC, so the intersection with dev using this script is even smaller (if not empty... until gccrs?). But good point, I'll hack it to have it put the line at the end.

CohenArthur:
I'm in favor of merging this and trying it out :)

Issue #1136, opened 2022-04-19
unrecognised token ‘|’ in grouped or tuple pattern after first pattern


philberty:
Other example which should work as soon as we fix the above example:
```rust
enum NewTypeOrStruct {
    One(i32),
    Two{other: String, value: i32},
} 

fn foo((NewTypeOrStruct::One(foo) | NewTypeOrStruct::Two{other: _, value: foo}): NewTypeOrStruct) -> i32 {
    foo + 15i32
}

fn main() {
    foo(NewTypeOrStruct::One(15));
    foo(NewTypeOrStruct::Two {
        other: String::from("other"),
        value: 14
    });
}
```

_Originally posted by @CohenArthur in https://github.com/Rust-GCC/gccrs/issues/995#issuecomment-1061139942_

philberty:
@flip1995  @CohenArthur @SimplyTheOther or @bjorn3 is this a grouped pattern? https://doc.rust-lang.org/reference/patterns.html our parser doesn't handle the case of the '|' to fix this i think it is a grouped pattern but not sure how the AST looks like here in Rustc

flip1995:
It's a grouped pattern. Function parameters are not normal ``Pattern``s, but ``PatternNoTopAlt``. It can't be a ``TuplePattern`` in this case, because a ``TuplePattern`` either has to be ``..`` or must include a ``,``.

P-E-P:
[...]

PR #1983, opened 2023-03-13
resolve: Handle multiple bindings to the same identifier

goar5670:
This is an attempt to provide handling for multiple bindings to the same identifier. It works for function and closure parameters only for now (i.e. E0415). Additionally, it lays down the base for resolving `AltPattern` bindings.

I followed rustc's implementation for the most part.
https://github.com/rust-lang/rust/blob/master/compiler/rustc_resolve/src/late.rs#L3168.

If initially approved, I'll extend it for pattern bindings as well (i.e. E0416) and add tests.

Partially addresses #1136

### DIFF
--- a/gcc/rust/resolve/rust-ast-resolve-item.cc
+++ b/gcc/rust/resolve/rust-ast-resolve-item.cc
@@ -77,13 +77,16 @@ ResolveTraitItems::visit (AST::TraitItemFunc &func)
   if (function.has_return_type ())
     ResolveType::go (function.get_return_type ().get ());
 
+  std::vector<PatternBinding> bindings
+    = {PatternBinding (PatternBoundCtx::Product, std::set<Identifier> ())};
+
   // we make a new scope so the names of parameters are resolved and shadowed
   // correctly
   for (auto &param : function.get_function_params ())
     {
       ResolveType::go (param.get_type ().get ());
-      PatternDeclaration::go (param.get_pattern ().get (),
-			      Rib::ItemType::Param);
+      PatternDeclaration::go (param.get_pattern ().get (), Rib::ItemType::Param,
+			      bindings);
     }
 
   if (function.has_where_clause ())
@@ -141,13 +144,16 @@ ResolveTraitItems::visit (AST::TraitItemMethod &func)
   ResolveType::go (&self_type_path);
   PatternDeclaration::go (&self_pattern, Rib::ItemType::Param);
 
+  std::vector<PatternBinding> bindings
+    = {PatternBinding (PatternBoundCtx::Product, std::set<Identifier> ())};
+
   // we make a new scope so the names of parameters are resolved and shadowed
   // correctly
   for (auto &param : function.get_function_params ())
     {
       ResolveType::go (param.get_type ().get ());
-      PatternDeclaration::go (param.get_pattern ().get (),
-			      Rib::ItemType::Param);
+      PatternDeclaration::go (param.get_pattern ().get (), Rib::ItemType::Param,
+			      bindings);
     }
 
   if (function.has_where_clause ())
@@ -496,13 +502,16 @@ ResolveItem::visit (AST::Function &function)
   if (function.has_return_type ())
     ResolveType::go (function.get_return_type ().get ());
 
+  std::vector<PatternBinding> bindings
+    = {PatternBinding (PatternBoundCtx::Product, std::set<Identifier> ())};
+
   // we make a new scope so the names of parameters are resolved and shadowed
   // correctly
   for (auto &param : function.get_function_params ())
     {
       ResolveType::go (param.get_type ().get ());
-      PatternDeclaration::go (param.get_pattern ().get (),
-			      Rib::ItemType::Param);
+      PatternDeclaration::go (param.get_pattern ().get (), Rib::ItemType::Param,
+			      bindings);
     }
 
   // resolve the function body
@@ -633,13 +642,16 @@ ResolveItem::visit (AST::Method &method)
   ResolveType::go (&self_type_path);
   PatternDeclaration::go (&self_pattern, Rib::ItemType::Param);
 
+  std::vector<PatternBinding> bindings
+    = {PatternBinding (PatternBoundCtx::Product, std::set<Identifier> ())};
+
   // we make a new scope so the names of parameters are resolved and shadowed
   // correctly
   for (auto &param : method.get_function_params ())
     {
       ResolveType::go (param.get_type ().get ());
-      PatternDeclaration::go (param.get_pattern ().get (),
-			      Rib::ItemType::Param);
+      PatternDeclaration::go (param.get_pattern ().get (), Rib::ItemType::Param,
+			      bindings);
     }
 
   // resolve any where clause items
--- a/gcc/rust/resolve/rust-ast-resolve-pattern.cc
+++ b/gcc/rust/resolve/rust-ast-resolve-pattern.cc
@@ -24,6 +24,78 @@ namespace Rust {
 namespace Resolver {
 
 void
+PatternDeclaration::go (AST::Pattern *pattern, Rib::ItemType type)
+{
+  std::vector<PatternBinding> bindings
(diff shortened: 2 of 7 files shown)


gcc/rust/resolve/rust-ast-resolve-pattern.h
```diff
@@ -25,6 +25,14 @@
 namespace Rust {
 namespace Resolver {
 
+enum PatternBoundCtx
+{
+  Product,
+  Or,
+};
+
+using PatternBinding = std::pair<PatternBoundCtx, std::set<Identifier>>;
```
philberty, line 34: This seems more like a type-def than a using. Please avoid using statements. Or make this your own struct

gcc/rust/resolve/rust-ast-resolve-pattern.h
```diff
@@ -57,17 +65,51 @@ class PatternDeclaration : public ResolverBase
 public:
   static void go (AST::Pattern *pattern, Rib::ItemType type)
   {
-    PatternDeclaration resolver (type);
+    auto bindings = std::vector (1, std::make_pair (PatternBoundCtx::Product,
```
philberty, line 68: what type of initilizer for std::vector is this? I cant read it

gcc/rust/resolve/rust-ast-resolve-item.cc
```diff
@@ -77,12 +77,15 @@ ResolveTraitItems::visit (AST::TraitItemFunc &func)
   if (function.has_return_type ())
     ResolveType::go (function.get_return_type ().get ());
 
+  auto bindings = std::vector (1, std::make_pair (PatternBoundCtx::Product,
```
philberty, line 80: I dont understand this vector at all

philberty (changes requested):
I think this is a great start, I think the initilizer you are using for std::vector is super unclear. I cant tell what your element type is.

Please try to use std::vector<mytype> bla = {x ,y ,z} where possible its easier to read

gcc/rust/resolve/rust-ast-resolve-pattern.h
```diff
@@ -92,8 +134,11 @@ class PatternDeclaration : public ResolverBase
   void visit (AST::RangePattern &pattern) override;
 
 private:
-  PatternDeclaration (Rib::ItemType type) : ResolverBase (), type (type) {}
+  PatternDeclaration (std::vector<PatternBinding> *bindings, Rib::ItemType type)
+    : ResolverBase (), bindings (bindings), type (type)
+  {}
 
+  std::vector<PatternBinding> *bindings = nullptr;
```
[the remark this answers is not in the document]
goar5670: I guess we can make it work with just references. Does it make a difference?

gcc/rust/resolve/rust-ast-resolve-pattern.h
```diff
@@ -57,17 +65,51 @@ class PatternDeclaration : public ResolverBase
...
   static void go (AST::Pattern *pattern, Rib::ItemType type)
   {
-    PatternDeclaration resolver (type);
+    auto bindings = std::vector (1, std::make_pair (PatternBoundCtx::Product,
+						    std::set<Identifier> ()));
+    PatternDeclaration resolver (&bindings, type);
     pattern->accept_vis (resolver);
   };
 
+  static void go (AST::Pattern *pattern, std::vector<PatternBinding> *bindings,
```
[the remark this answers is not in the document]
goar5670: I'll move the pattern visits into rust-ast-resolve.pattern.cc as well.

goar5670:
Thanks for the review @philberty, I added the test. There should be another test for tuples and structs but it fails right now because of how match arm patterns are handled. I'll add it as a part of #1985.

Please take another look whenever you're free.

gcc/rust/resolve/rust-ast-resolve-pattern.h
```diff
@@ -55,45 +63,25 @@ class PatternDeclaration : public ResolverBase
...
-  void visit (AST::IdentifierPattern &pattern) override
-  {
-    // if we have a duplicate id this then allows for shadowing correctly
-    // as new refs to this decl will match back here so it is ok to overwrite
-    resolver->get_name_scope ().insert (
-      CanonicalPath::new_seg (pattern.get_node_id (), pattern.get_ident ()),
-      pattern.get_node_id (), pattern.get_locus (), type);
-  }
+  static void go (AST::Pattern *pattern, Rib::ItemType type);
+  static void go (AST::Pattern *pattern, std::vector<PatternBinding> *bindings,
```
philberty, line 67: If your adding another override you should keep the pattern the same so you have:

```c++
static void go (AST::Pattern *pattern, Rib::ItemType type);
static void go (AST::Pattern *pattern, Rib::ItemType type, std::vector<PatternBinding>* bindings);
```

gcc/rust/resolve/rust-ast-resolve-pattern.cc
```diff
@@ -23,6 +23,71 @@
...
+  PatternDeclaration resolver (bindings, type);
+  pattern->accept_vis (resolver);
+}
+
+void
+PatternDeclaration::visit (AST::IdentifierPattern &pattern)
+{
+  rust_assert (!bindings->empty ());
+  if (bindings->back ().first == PatternBoundCtx::Product
+      && bindings->back ().second.find (pattern.get_ident ())
```
philberty, line 49: This if statement is not readable. If your doing something like this, you need to check for:

```
bool have_bindings = bindings.size() > 0;
bool have_product_bindings = have_bindings && bindings.back().first == PatternBoundCtx::Product;
...

if (have_product_bindings && .. )
```

I am not sure about this pattern why is this a vector? because you are only checking binds.back, I don't see any example where you add other bindings

gcc/rust/resolve/rust-ast-resolve-pattern.h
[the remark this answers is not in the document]
philberty: From reading how you use this vector, It doesnt seem like your saving anything by making this a poiinter. Pointers mean the ownership is unclear, because we dont know if this is meant to be free'd or its been allocated at all. We can avoid some of those issues by using const.

For your situation i would turn this into:

```std::vector<PatternBinding> bindings```. And just rely on copy constructors.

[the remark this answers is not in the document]
Comment (philberty) [reply gcc/rust/resolve/rust-ast-resolve-pattern.h:84]: If you really need it to be persisted and updated across multiple visitors you should use a ```std::vector<PAtternBindings>&``` using a reference is safer because you cant free it but still acts like a pointer so this implies the ownership is up to the initial caller.

File: gcc/rust/resolve/rust-ast-resolve-pattern.h [hunk 8]
```diff
@@ -25,6 +25,14 @@
 namespace Rust {
 namespace Resolver {
 
+enum PatternBoundCtx
+{
+  Product,
```
philberty, line 30: can you comment what each of these mean? Its not clear to me what they represent

philberty (changes requested):
I've left some comments i think i am starting to see what you are trying to do but i think it needs some tidy up.

gcc/rust/resolve/rust-ast-resolve-pattern.cc
goar5670: We need it to be a vector for when we add name resolution support for `AltPattern`. The only place we should push into the vector is inside `AltPattern` see [here](https://github.com/rust-lang/rust/blob/master/compiler/rustc_resolve/src/late.rs#L3222). I'm planning to work on `AltPattern` name resolution right after this PR gets merged so I think it's better to make it a vector from now.

gcc/rust/resolve/rust-ast-resolve-pattern.h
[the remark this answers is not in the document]
goar5670: Ok, I'll go with the reference then.

goar5670:
@philberty, responded to your comments. Hopefully, it's ready to be merged now.

CohenArthur (approved):
Looks good :D Great job!

PR #1985, opened 2023-03-13
gccrs: Refactor match arm pattern vectors into a single pattern

### LINKED ISSUE #1982: parse_match_arm_patterns refactor
In `rust-parse-impl.h`'s `parse_match_arm_patterns`:
```rust
...
  // parse required pattern - if doesn't exist, return empty
  std::unique_ptr<AST::Pattern> initial_pattern = parse_pattern ();
  if (initial_pattern == nullptr)
    {
      // FIXME: should this be an error?
      return patterns;
    }
  patterns.push_back (std::move (initial_pattern));

  // DEBUG
  rust_debug ("successfully parsed initial match arm pattern");

  // parse new patterns as long as next char is '|'
  const_TokenPtr t = lexer.peek_token ();
  while (t->get_id () == PIPE)
    {
...
```

I think the while loop is never reached since `parse_pattern` is supposed to parse all alternative patterns. Also, the return type of `parse_match_arm_patterns` should be refactored into a simple `Pattern`, or maybe it can be deleted, not sure.

goar5670:
Fixes #1982.

### DIFF
--- a/gcc/rust/ast/rust-ast.cc
+++ b/gcc/rust/ast/rust-ast.cc
@@ -1814,14 +1814,13 @@ IfLetExpr::as_string () const
   str += append_attributes (outer_attrs, OUTER);
 
   str += "\n Condition match arm patterns: ";
-  if (match_arm_patterns.empty ())
+  if (match_arm_pattern == nullptr)
     {
       str += "none";
     }
   else
     {
-      for (const auto &pattern : match_arm_patterns)
-	str += "\n  " + pattern->as_string ();
+      str += "\n  " + match_arm_pattern->as_string ();
     }
 
   str += "\n Scrutinee expr: " + value->as_string ();
@@ -2058,14 +2057,13 @@ WhileLetLoopExpr::as_string () const
     str += loop_label.as_string ();
 
   str += "\n Match arm patterns: ";
-  if (match_arm_patterns.empty ())
+  if (match_arm_pattern == nullptr)
     {
       str += "none";
     }
   else
     {
-      for (const auto &pattern : match_arm_patterns)
-	str += "\n  " + pattern->as_string ();
+      str += "\n  " + match_arm_pattern->as_string ();
     }
 
   str += "\n Scrutinee expr: " + scrutinee->as_string ();
@@ -2144,14 +2142,13 @@ MatchArm::as_string () const
   std::string str = append_attributes (outer_attrs, OUTER);
 
   str += "\nPatterns: ";
-  if (match_arm_patterns.empty ())
+  if (match_arm_pattern == nullptr)
     {
       str += "none";
     }
   else
     {
-      for (const auto &pattern : match_arm_patterns)
-	str += "\n " + pattern->as_string ();
+      str += "\n " + match_arm_pattern->as_string ();
     }
 
   str += "\nGuard expr: ";
--- a/gcc/rust/ast/rust-expr.h
+++ b/gcc/rust/ast/rust-expr.h
@@ -3470,14 +3470,14 @@ protected:
 class WhileLetLoopExpr : public BaseLoopExpr
 {
   // MatchArmPatterns patterns;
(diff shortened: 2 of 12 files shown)


gcc/rust/ast/rust-ast.cc
```diff
@@ -1814,14 +1814,13 @@ IfLetExpr::as_string () const
   str += append_attributes (outer_attrs, OUTER);
 
   str += "\n Condition match arm patterns: ";
-  if (match_arm_patterns.empty ())
+  if (match_arm_pattern == nullptr)
```
philberty, line 1817: This looks wrong, checking for null should be avoided checking for empty is much safer.

philberty (commented):
Initially i think this makes the AST and HIR structures more unsafe, which means we now have to check for nullptr during resolution, ir-dumps, typechecking, lints and codegen. By having an empt vector it was just empty and we can safely try to iterate it regardless.

From the issue you linked you reference that a while loop is not reached. Can you find a test case which fails to parse? You can verity the AST by looking at the AST dump, thought they may not be complete for patterns. We should be cautious about chaning the AST and HIR at this stage not that we can't do this but we need good reason for it.

P-E-P (approved):
LGTM, I've tested with a few edge case of mine but could'nt break it.

P-E-P:
> Can you find a test case which fails to parse ?

Parsing works fine right now with the old version. The comment made about the for loop was about reachability. Since pattern can contain pipes, we do not need to parse those pipes in all external functions.

At the beginning we were parsing like this:
```
| A | B | C | D ...
  ^   ^   ^   ^
  |   |   |   |
  +---+---+----+-- parse_pattern()
```
Now the whole pipe thing is a single pattern expression
```
| A | B | C | D ...
^^^^^^^^^^^^^^^
          |
          +-- parse_pattern()
```

goar5670:
> Initially i think this makes the AST and HIR structures more unsafe, which means we now have to check for nullptr during resolution, ir-dumps, typechecking, lints and codegen. By having an empt vector it was just empty and we can safely try to iterate it regardless.

Is there a reason we allow nullptr for patterns in the first place? I think the correct way to handle this would be to exit at the parsing stage whenever parse_pattern returns a nullptr.  Or would we still need to check for nullptr at all stages even with this handling?

> From the issue you linked you reference that a while loop is not reached. Can you find a test case which fails to parse? You can verity the AST by looking at the AST dump, thought they may not be complete for patterns. We should be cautious about chaning the AST and HIR at this stage not that we can't do this but we need good reason for it.

It doesn't necessarily fail to parse, the while loop was just useless/outdated. It was effectively just checking for alternative patterns so it became unreachable after introducing alt pattern parsing in #1878.

@philberty

goar5670:
> This change effectively removes the possibility for `if-let`s and `while-let`s to have multiple patterns. This is fine as of today, however `if/while-let` chains will probably eventually get stabilized, and we will need to support the unstable feature even if that isn't the case https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=2707039d4e0f0285701c5a02787941ad
> 
> So I think we would ultimately end up reversing that change. Unless I misunderstood the intent?

This is true, but I don't think the support for `if let` chains will necessarily depend on this vector of patterns. The chains can be linked by a combination of ands and ors so I think a more sophisticated structure will have to be implemented for that.

and just FYI, I'm not hard pushing for it or anything. If anybody doesn't agree to the changes, let's close the PR no problem.

CohenArthur:
Yeah I'm not really sure how if-let chains will eventually be implemented and how we'll desugar them and compile them. I think it might be easier to keep a vector of patterns, as there isn't really an `and` pattern apart from product type patterns, which might be difficult to reuse here. I just wouldn't want to merge your work to maybe remove it later down the line :)

I'm personally against the code change (but nothing against you! and thank you for exploring this! I appreciate it :) ) simply because we might end up removing it and/or it might cause us some pain when trying to support if-let chains in the future.

Issue #1807, opened 2023-02-01
Parsing issues on closure parameters and return type


CohenArthur:
This following line in libcore 1.49 throws our parser quite the curveball:
```rust
 let is_zero = &|&&d: &&u8| -> bool { d == b'0' };
 ```
 
 output:
 
 ```rust
 ../../rust/library/core/src/num/dec2flt/mod.rs:297:26: error: unrecognised token ‘&&’ in type
  297 |     let is_zero = &|&&d: &&u8| -> bool { d == b'0' };
      |                          ^~
../../rust/library/core/src/num/dec2flt/mod.rs:297:26: error: failed to parse type in closure parameter
../../rust/library/core/src/num/dec2flt/mod.rs:297:21: error: could not parse closure param
  297 |     let is_zero = &|&&d: &&u8| -> bool { d == b'0' };
      |                     ^~
../../rust/library/core/src/num/dec2flt/mod.rs:297:32: error: found unexpected token ‘->’ in null denotation
  297 |     let is_zero = &|&&d: &&u8| -> bool { d == b'0' };
      |                                ^~
../../rust/library/core/src/num/dec2flt/mod.rs:297:35: error: failed to parse expression in let statement
  297 |     let is_zero = &|&&d: &&u8| -> bool { d == b'0' };
      |                                   ^~~~
../../rust/library/core/src/num/dec2flt/mod.rs:297:5: error: failed to parse statement or expression without block in block expression
  297 |     let is_zero = &|&&d: &&u8| -> bool { d == b'0' };
      |     ^~~
  ```

CohenArthur:
This issue is actually made of two parts:
1. The parsing of types which are actually multiple references (`&&u8` here) #1812 
2. The proper handling of reference patterns (`&&d` here)

goar5670:
Now results in an ICE

<details><summary>Backtrace</summary>
  <p>
  
  ```
rust1: internal compiler error: Segmentation fault
0x1173f23 crash_signal
	../../gccrs/gcc/toplev.cc:314
0x7f5a480f851f ???
	./signal/../sysdeps/unix/sysv/linux/x86_64/libc_sigaction.c:0
0xafcee4 Rust::Analysis::NodeMapping::get_hirid() const
	../../gccrs/gcc/rust/util/rust-hir-map.cc:51
[...]

PR #1989, opened 2023-03-14
hir: Improve pattern bindings handling in closure parameters

goar5670:
Addresses #1807

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -2918,9 +2918,8 @@ CompileExpr::generate_closure_function (HIR::ClosureExpr &expr,
       tree compiled_param_var = ctx->get_backend ()->struct_field_expression (
 	args_param_expr, i, closure_param.get_locus ());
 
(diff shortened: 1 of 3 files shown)


philberty (approved):
LGTM but i am not 10% this fixes https://github.com/Rust-GCC/gccrs/issues/1807 as it seemed to be a parser issue this is applying the pattern logic properly to closures which is good and adds new test cases.

philberty:
This is good to merge if you can confirm what happens when you test this PR with #1807 and we can go from there.

goar5670:
It doesn't completely fix it yet, but the issue is not with reference patterns anymore (or parsing for that matter). It results in an ICE. I think it has something to do with the `&` before the closure pipe or the b'0'.

goar5670:
The way I see it, we have 2 choices. Close 1807 and open a new issue for the new behaviour. Or just leave it open and change the description to 'Addresses' instead of  'Fixes'

goar5670:
> In this case, you should change your description to Addresses and add a comment showing the ICE on the issue. Which can be fixed later on.

Ok done 👍

PR #1990, opened 2023-03-14
Cleanup

powerboat9:
I noticed I missed a few things for #1888, so I took the opportunity to do some general cleanup.


philberty (approved):
LGTM

PR #1991, opened 2023-03-15
Lower public macros to mappings

CohenArthur:
@philberty as you pointed out in #1958, using the mappings to store exported macros ends up being shorter and cleaner. This PR should probably be split in two so I'll keep it as a draft for now.

This correctly exports public macros to .rox files. I'll spend some time adding tests to ensure this behavior is respected.

- Revert "hir: Add ExportedMacro node and handling."
- mappings: Keep exported macro IDs
- lowering: Add lowering of exported macros


P-E-P (approved):
LGTM

PR #1993, opened 2023-03-16
resolve: Add name resolution for AltPattern

goar5670:
~Doesn't even fix~ Addresses #1136

The main changes in this commit can be logically split into two parts:

1) Pushing new pattern binding contexts in alt patterns. At the start   
of each AltPattern we push an 'Or' context which represents the relation
between the bindings of different alts.

   Before we resolve each alt arm we need to push a 'Product' context to represent
   the relation between the bindings of this specific alt arm and each other.
   This 'Product' context is removed after the alt arm visit and the its bindings
   are pushed into the 'Or' context.

   Eventually, the 'Or' context is removed as well after it fulfills its duty.
   Its bindings are then pushed into the previous context similarly and so on.

2) Checking for consistent bindings between the alt arms which is handled
by check_bindings_consistency. The info necessary for this check is held
by binding_info_map inside PatternDeclaration class. The binding_info_map
only holds bindings for one alt arm at a time. After every alt arm visit, 
these bindings info are pushed into a vec (that contains all alt arms info
and will eventually be passed to check_bindings_consistency) and emptied out
for the next alt arm visit. At the end, all the info from all the alt arms
are pushed again into binding_info in addition to the initial bindings that
were there before the alt arms visits, this way the binding_info will contain
all the binding_info present in the AltPattern (as it should).

In addition to that, some refactors were made (e.g. add_new_binding function)    
and some errors emitted, no biggie.

### DIFF
--- a/gcc/rust/resolve/rust-ast-resolve-pattern.cc
+++ b/gcc/rust/resolve/rust-ast-resolve-pattern.cc
@@ -28,9 +28,8 @@ PatternDeclaration::go (AST::Pattern *pattern, Rib::ItemType type)
 {
   std::vector<PatternBinding> bindings
     = {PatternBinding (PatternBoundCtx::Product, std::set<Identifier> ())};
-  PatternDeclaration resolver (bindings, type);
-  pattern->accept_vis (resolver);
-};
+  PatternDeclaration::go (pattern, type, bindings);
+}
 
 void
 PatternDeclaration::go (AST::Pattern *pattern, Rib::ItemType type,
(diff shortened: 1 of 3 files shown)


gcc/rust/resolve/rust-ast-resolve-pattern.cc
```diff
@@ -38,49 +37,25 @@ PatternDeclaration::go (AST::Pattern *pattern, Rib::ItemType type,
 {
   PatternDeclaration resolver (bindings, type);
   pattern->accept_vis (resolver);
+
+  for (auto &[ident, info] : resolver.missing_bindings)
```
philberty, line 41: this is C++17 style we only allow up to c++11 this is likely why the gcc4.8 build failed too

gcc/rust/resolve/rust-ast-resolve-pattern.cc
```diff
@@ -197,6 +171,160 @@ PatternDeclaration::visit (AST::TuplePattern &pattern)
...
+    PatternBinding (PatternBoundCtx::Or, std::set<Identifier> ()));
+
+  // This is a hack to avoid creating a separate visitor class for the
+  // consistency checks. We empty out the binding_info_map before each iteration
+  // to separate between the alts' binding_maps. And right after the alt
+  // visit...
+  auto tmp_binding_map = binding_info_map;
+  binding_info_map.clear ();
+
+  std::vector<std::map<Identifier, BindingInfo>> alts_binding_maps;
```
philberty, line 190: I think the C++ STL is really poor at dealing with stuff like this because the template arguments get messy, I think you should make the std::map<Identifier, BindingInfo> a typedef or a separate proxy class it means it decouples it a bit if it needs changed.

philberty: This would have made it easier for me to review the check_consistency function because it was unclear why iterating over the same vector like this made any sense

gcc/rust/resolve/rust-ast-resolve-pattern.h
```diff
@@ -45,6 +45,30 @@ struct PatternBinding
...
+struct BindingInfo
+{
+  bool is_mut;
+  bool is_ref;
+  Location locus;
+
+  BindingInfo (bool is_mut, bool is_ref, Location locus)
+    : is_mut (is_mut), is_ref (is_ref), locus (locus)
+  {}
+  BindingInfo (){};
```
philberty, line 59: brilliant this i think its just missing the copy constructor and assignment operator which will avoid some niggly issues

gcc/rust/resolve/rust-ast-resolve-pattern.h
```diff
@@ -45,6 +45,30 @@ struct PatternBinding
   {}
 };
 
+// Info that gets stored in the map. Helps us detect if two bindings to the same
+// identifier have different mutability or ref states.
+struct BindingInfo
+{
+  bool is_mut;
```
philberty, line 52: I think it would be ideal to reuse the Mutability enum from the rust/util/rust-common.h

gcc/rust/resolve/rust-ast-resolve-pattern.h
```diff
@@ -45,6 +45,30 @@ struct PatternBinding
   {}
 };
 
+// Info that gets stored in the map. Helps us detect if two bindings to the same
+// identifier have different mutability or ref states.
+struct BindingInfo
```
philberty, line 50: I think BindingInfo is too general and it sounds like it should contain the pattern or identifier. I would call this: BindingTypeInfo

philberty, line 50: I could see this struct becoming more complex I think you should make this a class and make these fields private and give them getters.

philberty (changes requested):
Excellent work, I left some style comments but this is really good work!

goar5670:
Thanks for the review @philberty, I applied your comments. Can you take another look?

gcc/rust/resolve/rust-ast-resolve-pattern.cc
```diff
@@ -197,6 +185,162 @@ PatternDeclaration::visit (AST::TuplePattern &pattern)
...
+  // This is a hack to avoid creating a separate visitor class for the
+  // consistency checks. We empty out the binding_info_map before each iteration
+  // to separate between the alts' binding_maps. And right after the alt
+  // visit...
+  auto tmp_binding_map = binding_info_map;
+  binding_info_map.clear ();
+
+  std::vector<BindingMap> alts_binding_maps;
+
+  for (auto &alt : pattern.get_alts ())
```
philberty, line 206: I probably would have called this pattern but its very minor nit

philberty (approved):
LGTM

PR #1994, opened 2023-03-16
Unify HIR::IfExprConseqIf{,Let} into HIR::IfExprConseqElse

powerboat9:
This should simplify 'if' expression handling to match simplifications to 'if let' expression handling in order to simplify 'if let' lowering.


powerboat9:
https://doc.rust-lang.org/stable/nightly-rustc/rustc_hir/enum.ExprKind.html#variant.If

rustc does something similar, so these changes shouldn't impair future feature development.

powerboat9:
yak shaving

CohenArthur (approved):
Thanks :) I think having our HIR being simpler than our AST where it makes sense is nice.

philberty (approved):
LGTM... YAK shaved...

PR #1995, opened 2023-03-16
Unify HIR::IfLetExprConseqIf{,Let} into HIR::IfLetExprConseqElse

powerboat9:
This should allow for 'if let' expressions to be lowered more easily.


philberty (approved):
LGTM

PR #2003, opened 2023-03-18
Clean tuple minor

powerboat9:
Some of the changes from #1941, split off into a smaller pull request.


philberty:
Sorry @powerboat9 I think we are not ready to accept such a big refactor to the HIR yet maybe when we get more of libcore working then lets revisit this. Raise an issue that this type of refactor should be possible though so we don't forget about this desugaring

PR #2006, opened 2023-03-19
Remove unused ResolvePattern visitor

goar5670:
Remove `ResolvePattern` visitor class.


philberty (approved):
LGTM just for note we are trying to split up our headers and move more stuff into .cc impl files so keep that in mind while your going through the code base.

PR #2007, opened 2023-03-20
Add name resolution and ast lowering for SlicePattern

goar5670:
Implement ast lowering and name resolution for `SlicePattern`


philberty (approved):
LGTM

Issue #850, opened 2022-01-04
Option Type requires lang items


philberty:
I tried this code: https://godbolt.org/z/YdPTMGqEd

```rust
extern "C" {
    fn printf(s: *const i8, ...);
}

enum Option<T> {
    #[lang = "None"]
    None,
    #[lang = "Some"]
    Some(T),
}

fn divide(numerator: f64, denominator: f64) -> Option<f64> {
    if denominator == 0.0 {
        None
    } else {
        Some(numerator / denominator)
    }
}

fn main() {
    let result = divide(2.0, 3.0);

    match result {
        Some(x) => unsafe {
            let a = "Result: %i\n\0";
            let b = a as *const str;
            let c = b as *const i8;

            printf(c, x);
        },
        None => unsafe {
            let a = "Cannot divide by 0\n\0";
            let b = a as *const str;
            let c = b as *const i8;

            printf(c);
        },
    }
}

```

I expected to see this happen: compile without error

Instead, this happened:

```
<source>:10:9: error: failed to find name: None
   10 |         None
      |         ^
<source>:12:9: error: Cannot find path 'Some' in this scope
   12 |         Some(numerator / denominator)
      |         ^
<source>:20:9: error: Cannot find path 'Some' in this scope
   20 |         Some(x) => println!("Result: {}", x),
      |         ^
```

### Meta

- What version of Rust GCC were you using, git sha if possible. 69f6be3ee483c9895b4b5187a44b3e1c8be2ba63

goar5670:
Hello, I want to work on this issue if it's not too difficult. I played with it for a bit and found out that it fails at the resolution step because of an incomplete canonical path but the problem is that we handle lang items later in ast lowering, so I'm currently stuck :) what do you suggest here?

philberty:
This might be a tricky one the test case is missing the prelude use statement which injects the name of Some and None so you don't need a full canonical path so the patch should be relating to use statements.

Issue #855, opened 2022-01-04
Rust result type requires lang items


philberty:
I tried this code:

```rust
pub enum Result<T, E> {
    #[lang = "Ok"]
    Ok(T),

    #[lang = "Err"]
    Err(E),
}

fn main() {
    let result = Ok(123);
    let err = Err(false);
}

```

I expected to see this happen: compile without error
Instead, this happened: 

```
<source>:10:18: error: Cannot find path 'Ok' in this scope
   10 |     let result = Ok(123);
      |                  ^
<source>:11:15: error: Cannot find path 'Err' in this scope
   11 |     let err = Err(false);
      |               ^
```

### Meta

- What version of Rust GCC were you using, git sha if possible. 69f6be3ee483c9895b4b5187a44b3e1c8be2ba63

bjorn3:
This is not because of the lang items. This is because Ok and Err are re-exported separately in the prelude: https://github.com/rust-lang/rust/blob/2b681ac06b1a6b7ea39525e59363ffee0d1a68e5/library/core/src/prelude/v1.rs#L44

philberty:
Ah, brilliant news. Now I need to know why are they defined as lang items?

bjorn3:
They are used by `expr_ok` and `expr_try` which seem to only be used by the deprecated `RustcEncodable` and `RustcDecodable` derive macros: https://github.com/rust-lang/rust/blob/2b681ac06b1a6b7ea39525e59363ffee0d1a68e5/compiler/rustc_expand/src/build.rs#L348-L357

PR #2011, opened 2023-03-20
Support use declaration to export the type properly

philberty:
Fixes: #850 #855


philberty:
I have the fix for these two issues ready to go as well

CohenArthur (commented):
LGTM :)

PR #2012, opened 2023-03-21
ast: Add conversion to token stream

P-E-P:
Add several functions to convert an ast back to a token stream. This may be used later either to unify the ast dumping and modifications on ast nodes in procedural macros.

### DIFF
--- a/gcc/rust/ast/rust-ast-tokenstream.cc
+++ b/gcc/rust/ast/rust-ast-tokenstream.cc
@@ -1586,20 +1586,81 @@ TokenStream::visit (ExternCrate &crate)
 }
 
 void
-TokenStream::visit (UseTreeGlob &)
-{}
+TokenStream::visit (UseTreeGlob &use_tree)
+{
+  switch (use_tree.get_glob_type ())
+    {
+      case UseTreeGlob::PathType::PATH_PREFIXED: {
+	auto path = use_tree.get_path ();
+	visit (path);
+	tokens.push_back (Rust::Token::make (SCOPE_RESOLUTION, Location ()));
+      }
+      break;
+    case UseTreeGlob::PathType::NO_PATH:
+      tokens.push_back (Rust::Token::make (SCOPE_RESOLUTION, Location ()));
+      break;
+    case UseTreeGlob::PathType::GLOBAL:
+      break;
+    }
+  tokens.push_back (Rust::Token::make (ASTERISK, Location ()));
+}
 
 void
-TokenStream::visit (UseTreeList &)
-{}
+TokenStream::visit (UseTreeList &use_tree)
+{
+  switch (use_tree.get_path_type ())
+    {
+      case UseTreeList::PathType::PATH_PREFIXED: {
+	auto path = use_tree.get_path ();
+	visit (path);
+	tokens.push_back (Rust::Token::make (SCOPE_RESOLUTION, Location ()));
+      }
+      break;
+    case UseTreeList::PathType::NO_PATH:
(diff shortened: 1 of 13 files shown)


gcc/rust/ast/rust-ast-tokenstream.h
CohenArthur, line 1: Great stuff here, the header looks very good and the documentation is nice!

gcc/rust/ast/rust-ast-tokenstream.cc
```diff
@@ -0,0 +1,2718 @@
...
+template <typename T>
+void
+TokenStream::visit_as_line (T &item, std::vector<TokenPtr> trailing)
+{
+  visit (item);
+  for (auto &token : trailing)
+    {
+      tokens.push_back (token);
+    }
+  // tokens.push_back (Rust::Token::make (NEWLINE, Location ()));
```
CohenArthur, line 75: Deadcode?

gcc/rust/ast/rust-ast-tokenstream.cc
```diff
@@ -0,0 +1,2718 @@
...
+
+template <typename T>
+void
+TokenStream::visit_as_line (T &item, std::vector<TokenPtr> trailing)
+{
+  visit (item);
+  for (auto &token : trailing)
+    {
+      tokens.push_back (token);
+    }
```
CohenArthur, line 74: I believe the GNU style prefers
```suggestion
      tokens.push_back (token);
```

but this is a *very tiny nitpick*

CohenArthur (approved):
Looks great :) I think yeah we should start integrating it and see where it needs adjustments. This doesn't cause regressions or anything and is necessary for proc macros.

gcc/rust/ast/rust-ast-tokenstream.h
```diff
@@ -0,0 +1,310 @@
...
+#ifndef RUST_AST_TOKENSTREAM_H
+#define RUST_AST_TOKENSTREAM_H
+
+namespace Rust {
+namespace AST {
+
+class TokenStream : public ASTVisitor
+{
+public:
+  TokenStream (std::vector<TokenPtr> &container);
```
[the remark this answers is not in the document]
P-E-P: I'm not entirely sure about this. You're assuming we always want to get a tokenstream at a crate or item level. But attributes and proc macros may be placed in very odd situations, I would prefer have a getter and simply visit the ast from wherever we need. This would be more flexible.

gcc/rust/ast/rust-ast-tokenstream.cc
P-E-P: Yes, the `NEWLINE` token do not exist anymore (I'm not even sure it existed at all, there are some comment remaining in `rust-token.h`). This was later replaced with a call to an empty newline function in [5541d52](https://github.com/Rust-GCC/gccrs/pull/2012/commits/5541d522ff36ef8d46845b102548acd2afc3b588). Since TokenStream are a kind of superset of the dump code I thought we would refactor all of this in order to avoid maintaining the same behavior. If you think this is not right I can reverse the "formatting" commit.

CohenArthur (approved):
This does not introduce regressions, and will allow shortening the AST dump greatly, so I'm in favor of merging it now as it's getting big :)

P-E-P:
> This does not introduce regressions, and will allow shortening the AST dump greatly, so I'm in favor of merging it now as it's getting big :)

Sure, let's do this once the CI pass.

CohenArthur:
weird, this should have been done automatically by github when I rebased and pushed :thinking:

PR #2014, opened 2023-03-21
ast: Dump each MacroRule properly without the extra semicolon

CohenArthur:
Having the extra semicolon causes parsing errors when importing macros
exported using the AST dump, as no rule expects multiple tokens after a
single macro rule definition.

gcc/rust/ChangeLog:

	* ast/rust-ast-dump.cc (Dump::visit): Remove extraneous semicolon when
	dumping macro rules.


philberty (approved):
LGTM

PR #2016, opened 2023-03-22
Split Cfg stripping and macro expansion phase

CohenArthur:
- expand: Move cfg-attrs related functions out of MacroExpander
- expand: Add new ExpandVisitor class
- expand: Rename AttrVisitor -> CfgStrip

In order to import macros properly, we need to be able to name resolve extern crates, import their metdata, and select the macros out of it (among other things). Due to how the expansion pipeline was structured, extern crate importing was performed at the same time as macro expansion, which obviously causes issues when name resolving imported macros. However, directly importing extern crates in the Name Resolver has its own issues: we might be in a situation where the EarlyNameResolver imports a crate and its items, but the AttrVisitor then realizes this extern crate should be stripped because of a cfg-directive evaluating to false.

We effectively have two options to handle this case:

1/ Keep mappings of items imported for each extern crate, and if we do end up stripping an imported crate, strip all of the items associated with it

This is problematic for two reasons:

1. We effectively do a lot of work for... nothing, as we will search the external crate, load its metadata and parse it, only to discard it
2. Most importantly, this is also wrong. cfg-directives are often used for conditional compilation, for example for targeting multiple systems or architectures. Consider the following:

```rust
#[cfg(target_os = "linux")]
#[macro_use]
extern crate my_linux_crate;

#[cfg(target_os = "windows")]
#[macro_use]
extern crate my_msdos_crate;
```

it does not make sense and is also wrong to try and load crates that might not be present on the system: there's a very good chance that in such a setup, our user on Linux will not have compiled the Windows-related crate, and will end up with a very silly "cannot find `my_msdos_crate` crate".

Hence the second option:

2/ Split the cfg-stripping and macro-expansion pass

By first running cfg-stripping *before* running EarlyNameResolution, we are able to strip unwanted extern crates, load the remaining ones within the EarlyNameResolver, import macros, and perform proper name resolution and expansion.

Our `AttrVisitor` class was effectively doing the following actions within each of its visitors:

```cpp
void AttrVisitor::visit(AST::CallExpr &stmt) {
  // check if the call needs to be stripped
  expand_cfg_attrs(stmt.get_outer_attrs())
  if (fails_cfg_expansion()) {
    stmt.mark_for_strip();
    return;
  }

  // the call expression might be a macro invocation - expand it
  maybe_expand_expr(stmt.get_called_expression());

  // recursively substrip arguments given to the call
  for (auto &arg: stmt.get_arguments())
    arg->accept_vis();

  // each argument might be a macro invocation - expand it
  for (auto &arg: stmt.get_arguments())
    maybe_expand_expr(arg);

  for (auto &arg: stmt.get_arguments())
    if (arg.is_marked_for_strip())
      arg.skip() // more complex than this but basically this
}
```

You can see that we are alternating between checking cfg-predicates, stripping nodes from our AST, and expanding macros. The new architecture is much simpler, with one visitor taking care of cfg-predicates and stripping, and one of macro expansion:

```cpp
void CfgStrip::visit(AST::CallExpr &stmt) {
  // check if the call needs to be stripped
  expand_cfg_attrs(stmt.get_outer_attrs())
  if (fails_cfg_expansion()) {
    stmt.mark_for_strip();
    return;
  }

  // recursively substrip arguments given to the call
  for (auto &arg: stmt.get_arguments())
    arg->accept_vis();

  for (auto &arg: stmt.get_arguments())
    if (arg.is_marked_for_strip())
      arg.skip() // more complex than this but basically this
}

void ExpandVisitor::visit(AST::CallExpr &stmt) {
  // the call expression might be a macro invocation - expand it
  maybe_expand_expr(stmt.get_called_expression());
 
  // each argument might be a macro invocation - expand it
  for (auto &arg: stmt.get_arguments())
    maybe_expand_expr(arg);
}
```

There is a bit more that goes into this, but this is overall a big simplification of our expansion pipeline and should make it easier to extend

### DIFF
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -81,7 +81,7 @@ GRS_OBJS = \
     rust/rust-mangle.o \
     rust/rust-compile-resolve-path.o \
     rust/rust-macro-expand.o \
-    rust/rust-attribute-visitor.o \
+    rust/rust-cfg-strip.o \
 	rust/rust-expand-visitor.o \
     rust/rust-macro-invoc-lexer.o \
     rust/rust-macro-substitute-ctx.o \
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -82,6 +82,7 @@ GRS_OBJS = \
     rust/rust-compile-resolve-path.o \
     rust/rust-macro-expand.o \
     rust/rust-attribute-visitor.o \
+	rust/rust-expand-visitor.o \
     rust/rust-macro-invoc-lexer.o \
     rust/rust-macro-substitute-ctx.o \
     rust/rust-macro-builtins.o \
--- a/gcc/rust/expand/rust-attribute-visitor.cc
+++ b/gcc/rust/expand/rust-attribute-visitor.cc
@@ -141,6 +141,20 @@ AttrVisitor::go (AST::Crate &crate)
       crate.strip_crate ();
       // TODO: maybe create warning here? probably not desired behaviour
     }
+
+  auto &items = crate.items;
+
+  for (auto it = items.begin (); it != items.end ();)
+    {
+      auto &item = *it;
+      item->accept_vis (*this);
+
+      if (item->is_marked_for_strip ())
+	it = items.erase (it);
+      else
+	it++;
+    }
+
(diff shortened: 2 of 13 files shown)


philberty (approved):
I like this patch set LGTM

PR #2051, opened 2023-03-27
Fix parsing of array expressions.

### LINKED ISSUE #2048: Wrong grammar for array literals/index operator
I tried this code:

```rust
pub fn f() {
    [0][0];
}
```

I expected to see this happen: This should compile.

Instead, this happened: Syntax error:
```
<source>:2:8: error: expecting '}' but '[' found
    2 |     [0][0];
      |        ^
<source>:2:8: error: error may be from having an expression (as opposed to statement) in the body of the function but not last
Compiler returned: 1
```

### Meta

```
gccrs (Compiler-Explorer-Build-gcc-be27571c110d99fe55324a9c2b6622787b08fb52-binutils-2.38) 13.0.1 20230221 (experimental)
```
So probably be27571c110d99fe55324a9c2b6622787b08fb52 (even though Godbolt says 2023-02-21)?

Godbolt link: https://godbolt.org/z/nnfY7eoe7

powerboat9:
Fixes #2048.


CohenArthur (changes requested):
Looks good to me :) Could you add an execution test to make sure we don't eventually add regressions to this?

```rust
fn main() -> i32 {
    [0, 2, 3][1] - 2
}
```

The test outputs properly, so your patch is good imo

powerboat9:
Added the execution test

CohenArthur (approved):
Thanks :)

PR #2055, opened 2023-03-28
docker: Do not run update-alternatives on invalid path

### LINKED ISSUE #2047: Remove docker build automation
Docker image build is failing lets remove this but maybe keep the dockerfile around

CohenArthur:
We do not need to set the system's compiler to `gccrs`, as we are mostly
interested in *having* gccrs - not necessarily using it as a C and C++
compiler as well.

ChangeLog:

	* Dockerfile: Remove update-alternatives invocation.

Fixes #2047

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,11 +64,6 @@ RUN /bin/sh -c set -ex; \
     echo '/usr/local/lib64' > /etc/ld.so.conf.d/local-lib64.conf; \
     ldconfig -v
 
-RUN /bin/sh -c set -ex; \
-    dpkg-divert --divert /usr/bin/gcc.orig --rename /usr/bin/gcc; \
-    dpkg-divert --divert /usr/bin/g++.orig --rename /usr/bin/g++; \
-    update-alternatives --install /usr/bin/cc cc /usr/local/bin/gcc 999
-
 FROM rust:latest
 COPY --from=gcc-builder /usr/ /usr/
 COPY --from=gcc-builder /GCCRS_BUILD /GCCRS_BUILD


Dockerfile
```diff
@@ -66,8 +66,7 @@ RUN /bin/sh -c set -ex; \
 
 RUN /bin/sh -c set -ex; \
     dpkg-divert --divert /usr/bin/gcc.orig --rename /usr/bin/gcc; \
-    dpkg-divert --divert /usr/bin/g++.orig --rename /usr/bin/g++; \
-    update-alternatives --install /usr/bin/cc cc /usr/local/bin/gcc 999
+    dpkg-divert --divert /usr/bin/g++.orig --rename /usr/bin/g++;
```
dkm, line 69: Why do you need this block ? ISTM that you could remove it completely if you don't want to mess with system compiler.

CohenArthur: I don't mind removing the entire block - My philosophy was leaving as many lines of the Dockerfile untouched as possible, but I'm fine with changing it more :)

CohenArthur: Fixed in a732373

philberty:
I do sort of wonder should we just stop pushing the docker image at somepoint from github?

philberty (approved):
LGTM

CohenArthur:
> I do sort of wonder should we just stop pushing the docker image at somepoint from github?

I'm not in favor of this since https://github.com/rust-gcc/testing relies on it, but if it becomes too much of a hassle we can think about alternatives for sure

philberty:
> > I do sort of wonder should we just stop pushing the docker image at somepoint from github?
> 
> I'm not in favor of this since https://github.com/rust-gcc/testing relies on it, but if it becomes too much of a hassle we can think about alternatives for sure

Ah I forgot about that, though maybe we need to think about the registry location to be an organization name rather than my account.

PR #2056, opened 2023-03-28
Add proper support for `pub(crate)`

### LINKED ISSUE #1448: pub(crate) not supported and gives error without source location
I tried this code:

```rust
pub(crate) struct Foo;
```

I expected to see this happen: Compiles

Instead, this happened:

```
rust1: error: cannot use non-module path as privacy restrictor
Compiler returned: 1
```

### Meta

```
gccrs (Compiler-Explorer-Build-gcc-8809ee8c6a5621e830f3cfe66c381f986e63c7f2-binutils-2.38) 12.0.1 20220118 (experimental)
```

### LINKED ISSUE #2000: Bad privacy error
I tried this code: https://godbolt.org/z/zW3zWEfoT

```rust
struct Foo;

impl Foo {
    pub(crate) fn bla() {

    }
}
```

Instead, this happened: 

```
crab1: error: cannot use non-module path as privacy restrictor
```

### Meta

- What version of Rust GCC were you using, git sha if possible. a1f940d193c6cdb13483690a4f4a7d501ad7040e

CohenArthur:
- ast: Fix location for pub(crate)
- resolver: Allow SimplePath to resolve to their root segment
- privacy: Check for pub(crate) when resolving visibility path.

Fixes #2000
Fixes #1448

This causes the function to return true and the checks to pass, but it
requires more thinking - how should we deal with pub(crate) in the current system?
Should we simply treat it as a pub item in the current crate, but export it as
a private item in the metadata?

I'll look into this more in the coming days.

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -413,7 +413,7 @@ public:
   {
     std::vector<AST::SimplePathSegment> single_segments
       = {AST::SimplePathSegment (std::move (str), locus)};
-    return SimplePath (std::move (single_segments));
+    return SimplePath (std::move (single_segments), false, locus);
   }
 
   const std::vector<SimplePathSegment> &get_segments () const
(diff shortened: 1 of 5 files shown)


philberty (approved):
LGTM nice one

PR #2059, opened 2023-03-28
Unify AST::Tuple{Struct,Pattern}Items

powerboat9:
Removes the redundant distinction between ```AST::TupleStructItems``` and ```AST::TuplePatternItems```. Split off from #2003


CohenArthur:
[`TupleStructItems`](https://doc.rust-lang.org/reference/patterns.html#tuple-struct-patterns) and [`TuplePatternItems`](https://doc.rust-lang.org/reference/patterns.html#tuple-patterns) are different, so I'm not sure this is a valid cleanup - unless I'm misunderstanding?

powerboat9:
https://doc.rust-lang.org/reference/patterns.html#tuple-struct-patterns
https://doc.rust-lang.org/reference/patterns.html#tuple-patterns

The Tuple*Items are both just lists of patterns

powerboat9:
@CohenArthur

powerboat9:
@philberty

CohenArthur:
> `TupleStructItems` and `TuplePatternItems` are both just lists of patterns. `TuplePatternItems` is described differently because it has to be distinct from `GroupedPattern`.

Yes, but we'll need to handle them differently potentially, and it makes sense to keep the information of where they originate from. I'm not in favor of AST changes which are not reflected in the Rust reference personally

PR #2061, opened 2023-03-29
Fix bad cast error to bool

### LINKED ISSUE #2026: Bad cast as bool
I tried this code: https://github.com/rust-lang/rust/blob/aa3ca1994904f2e056679fce1f185db8c7ed2703/src/test/ui/cast-as-bool.rs

```rust
fn main() {
    let u = 5 as bool;
  
}
```

```
  //~^ ERROR cannot cast as `bool`
```

Instead, this happened: compiled without error

### Meta

emanuele-em:
In rust is not allowed to cast from int to bool. This patch handles the case when we cast a integer to bool with 'as bool'

close #2026 

- [x] Run `clang-format`


emanuele-em:
image

Issue #2052, opened 2023-03-27
ICE when compiling closures that have a single expression as body (not a block)


narpfel:
I tried this code:

```rust
#[lang = "fn_once"]
pub trait FnOnce<Args> {
    #[lang = "fn_once_output"]
    type Output;

    extern "rust-call" fn call_once(self, args: Args) -> Self::Output;
}

pub fn f() {
    || 42;
}
```

I expected to see this happen: This should compile.

Instead, this happened: ICE:
<details>
<summary>Backtrace</summary>

```
crab1: internal compiler error: in generate_closure_function, at rust/backend/rust-compile-expr.cc:2887
0x23e927e internal_error(char const*, ...)
	???:0
0xadbcea fancy_abort(char const*, int, char const*)
	???:0
0xd79de7 Rust::Compile::CompileExpr::visit(Rust::HIR::ClosureExpr&)
	???:0
0xd72114 Rust::Compile::CompileExpr::Compile(Rust::HIR::Expr*, Rust::Compile::Context*)
	???:0
0xd70967 Rust::Compile::CompileStmt::visit(Rust::HIR::ExprStmtWithoutBlock&)
	???:0
0xd70bb1 Rust::Compile::CompileStmt::Compile(Rust::HIR::Stmt*, Rust::Compile::Context*)
	???:0
0xd99e0a Rust::Compile::HIRCompileBase::compile_function_body(Rust::Compile::Context*, tree_node*, Rust::HIR::BlockExpr&, bool)
	???:0
0xd9b980 Rust::Compile::HIRCompileBase::compile_function(Rust::Compile::Context*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, Rust::HIR::SelfParam&, std::vector<Rust::HIR::FunctionParam, std::allocator<Rust::HIR::FunctionParam> >&, Rust::HIR::FunctionQualifiers const&, Rust::HIR::Visibility&, std::vector<Rust::AST::Attribute, std::allocator<Rust::AST::Attribute> >&, Location, Rust::HIR::BlockExpr*, Rust::Resolver::CanonicalPath const*, Rust::TyTy::FnType*, bool)
	???:0
0xd6fd2b Rust::Compile::CompileItem::visit(Rust::HIR::Function&)
	???:0
0xbd6709 Rust::Compile::CompileCrate::go()
	???:0
[...]

PR #2066, opened 2023-03-29
gccrs: fix ICE when closure body is not a block

philberty:
Fixes: #2052

gcc/rust/ChangeLog:

	* backend/rust-compile-expr.cc (CompileExpr::generate_closure_function):
	  when its not a block we dont have any ribs to generate locals from

gcc/testsuite/ChangeLog:

	* rust/execute/torture/issue-2052.rs: New test.


CohenArthur (approved):
LGTM :)

PR #2072, opened 2023-03-29
gccrs: Add testcase to show matching of enum variants

### LINKED ISSUE #852: Dataless enum variants can be matched without full path
I tried this code: https://godbolt.org/z/7eqP9vrTn

```rust
extern "C" {
    fn printf(s: *const i8, ...);
}

enum Foo {
   A,
   B(i32),
}

fn main() {
    let result = Foo::B(123);

    match result {
        A => unsafe {
            let a = "A\n\0";
            let b = a as *const str;
            let c = b as *const i8;

            printf(c);
        },
        Foo::B(x) => unsafe {
            let a = "Result: %i\n\0";
            let b = a as *const str;
            let c = b as *const i8;

            printf(c, x);
        },
    }
}

```

Instead, this happened:

```
rust1: error: expected [Foo{A = 1, B (0:Int:i32:(Ref: 28 TyRef: 8[8,]):bounds:[])}] got [<tyty::error>]
<source>:13:5: error: failed to type resolve expression
   13 |     match result {
      |     ^
<source>:10:1: error: expected [()] got [<tyty::error>]
   10 | fn main() {
      | ^
```

### Meta

philberty:
Fixes #852

gcc/testsuite/ChangeLog:

	* rust/compile/issue-852.rs: New test.

### DIFF
--- a/gcc/testsuite/rust/compile/issue-852.rs
+++ b/gcc/testsuite/rust/compile/issue-852.rs
@@ -1,0 +1,30 @@
+// { dg-options "-w" }
+extern "C" {
+    fn printf(s: *const i8, ...);
+}
+
+enum Foo {
+    A,
+    B(i32),
+}
+
+fn main() {
+    let result = Foo::B(123);
+
+    match result {
+        A => unsafe {
+            let a = "A\n\0";
+            let b = a as *const str;
+            let c = b as *const i8;
+
+            printf(c);
+        },
+        Foo::B(x) => unsafe {
+            let a = "Result: %i\n\0";
+            let b = a as *const str;
+            let c = b as *const i8;
+
+            printf(c, x);
+        },
+    }
+}


gcc/testsuite/rust/compile/issue-852.rs
```diff
@@ -0,0 +1,30 @@
...
+enum Foo {
+    A,
+    B(i32),
+}
+
+fn main() {
+    let result = Foo::B(123);
+
+    match result {
+        A => unsafe {
```
CohenArthur, line 15:
```suggestion
        Foo::A => unsafe {
```

CohenArthur: Otherwise the second match arm should never be entered - I didn't catch up on the Zulip discussion which addresses this tho :D

PR #2076, opened 2023-03-30
Add derive proc_macro expansion stubs

### LINKED ISSUE #1986: Stubs for proc_macro expansion
Implement some stubs in gccrs in order to facilitate the future integration of procedural macros expansion.

P-E-P:
Add proc macros expansion stub functions.

Fixes #1986

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -980,6 +980,8 @@ public:
 
   virtual void set_node_id (NodeId id) { node_id = id; }
 
+  virtual std::vector<Attribute> &get_outer_attrs () = 0;
+
 protected:
   // Constructor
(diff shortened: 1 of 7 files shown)


P-E-P:
I'd like some insight on [0e1d6c8](https://github.com/Rust-GCC/gccrs/pull/2076/commits/0e1d6c8845e47315621c280e9b017757354a09b3) since it modifies the behavior of a widespread component.

CohenArthur (approved):
Sounds good to me :) we talked with @P-E-P offline and this will require some more bits and pieces to do proper context-dependant expansion, so that for example this code
```rust
#[derive(Copy)]
struct Foo { value: i32 }
```

gets expanded to
```rust
struct Foo { value: i32 }

impl Copy for Foo {
    fn copy(&self) -> Self { ... }
 }
 ```

PR #2078, opened 2023-03-31
gccrs: Fix bad cast as a char

### LINKED ISSUE #2027: Bad cast as a char
I tried this code: https://github.com/rust-lang/rust/blob/aa3ca1994904f2e056679fce1f185db8c7ed2703/src/test/ui/cast_char.rs

```rust
fn main() {
    const XYZ: char = 0x1F888 as char;
    //~^ ERROR only u8 can be cast into char
    const XY: char = 129160 as char;
    //~^ ERROR only u8 can be cast into char
    const ZYX: char = '\u{01F888}';

}
```

I expected to see this happen: Bad cast into char error

Instead, this happened: compile without error

### Meta

emanuele-em:
In rust cast to char is allowed only from u8 type. This patch handles fix the case when the type is infered from an integer value, allowing only the u8 case'

Fixes #2027

gcc/rust/ChangeLog:

	* typecheck/rust-casts.cc (TypeCastRules::cast_rules): case INTEGRAL handles TypeKind::CHAR

gcc/testsuite/ChangeLog:

	* rust/compile/cast5.rs: New test.

- [x] Run `clang-format`

### DIFF
--- a/gcc/rust/typecheck/rust-casts.cc
+++ b/gcc/rust/typecheck/rust-casts.cc
@@ -64,7 +64,6 @@ TypeCastRules::cast_rules ()
 
   rust_debug ("cast_rules from={%s} to={%s}", from_type->debug_str ().c_str (),
 	      to.get_ty ()->debug_str ().c_str ());
-
   switch (from_type->get_kind ())
     {
       case TyTy::TypeKind::INFER: {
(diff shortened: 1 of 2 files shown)


philberty (approved):
LGTM

PR #2081, opened 2023-04-01
Add floppyMIT dual license

jdupak:
Any user of 3.5" floppy disks is clearly superior to other users and shall not be limited by the GPL license. Hence the floppyMIT license must be added to this project as a dual license.


tschwinge:
Oh no.  I don't have 3.5" floppy anymore, only have 5.25".  :'-(

Or Datasette, of course.

dkm:
signal-2023-04-01-102303
I find it a bit harsh, considering my minitel doesn't have any floppy/cd/parallel/serial interface... I feel a bit excluded :(

powerboat9:
The license as written seems to allow the user to "deal in the Software without restriction", which would suggest that copying the software to non-volatile storage other than a floppy disk is allowed.

Issue #1996, opened 2023-03-16
break rust


bugaevc:
I tried this code:

```rust
fn main() {
    break rust;
}
```

I expected to see this happen:

```
error: internal compiler error: It looks like you're trying to break rust; would you like some ICE?
note: the compiler expectedly panicked. this is a feature.
note: we would appreciate a joke overview: https://github.com/rust-lang/rust/issues/43162#issuecomment-320764675
```

Instead, this happened:

```
<source>:2:11: error: failed to find name: rust
    2 |     break rust;
      |           ^~~~
```

IMO this is an essential feature that GCC Rust absolutely has to support! 😄 In addition to `break rust;`, I propose also accepting `break gcc;` as a GNU extension.

### Meta

- What version of Rust GCC were you using, git sha if possible.

I was using Godbolt:

gccrs (Compiler-Explorer-Build-gcc-a1f940d193c6cdb13483690a4f4a7d501ad7040e-binutils-2.38) 13.0.1 20230221 (experimental)

bugaevc:
I've taken a look into implementing this — and didn't make much progress. So dumping my findings & understanding here, maybe you could provide me with some guidance. Please keep in mind that this is my first time really looking into the gccrs internals, and not just casually browsing the source / history.

The "failed to find name" errors has the following backtrace:

```c
#0  rust_error_at (location=..., fmt=fmt@entry=0x22d9800 "failed to find name: %s") at ../../gcc/rust/rust-diagnostics.cc:188
#1  0x0000000000b82afb in Rust::Resolver::ResolveExpr::visit (this=0x7fffffffd580, expr=...) at ../../gcc/rust/resolve/rust-ast-resolve-expr.cc:125
#2  0x0000000000b83667 in Rust::Resolver::ResolveExpr::go (canonical_prefix=..., prefix=..., expr=0x32043f0) at ../../gcc/rust/resolve/rust-ast-resolve-expr.cc:35
#3  Rust::Resolver::ResolveExpr::visit (this=<optimized out>, expr=...) at ../../gcc/rust/resolve/rust-ast-resolve-expr.cc:387
#4  0x0000000000b83a94 in Rust::Resolver::ResolveExpr::go (expr=0x3204370, prefix=..., canonical_prefix=...) at ../../gcc/rust/resolve/rust-ast-resolve-expr.cc:35
[...]

PR #2086, opened 2023-04-03
break rust 💥

bugaevc:
Since https://github.com/Rust-GCC/gccrs/issues/1996 has received positive feedback, here's a tentative implementation 😄

TODO:
- [x] Better / funnier error message
  * Suggestions welcome!
  * ...or is the current one fine?
  * It would be nice to make it a little bit more explicit that this is not _really_ an ICE, there is no bug to report, etc.
    * rustc says: "the compiler expectedly panicked. this is a feature."
- [x] Fix/verify Changelog entries
  * I've no idea how to do GNU-style changelogs! I've read [this doc](https://www.gnu.org/prep/standards/html_node/Change-Logs.html) and skimmed through some other gccrs commits, and tried to write some entries based on that
  * glibc people have accepted my patches without Changelog entries; are they really required?
- [x] Try to make the error point at the whole statement and not just the name
  * Or... is it worth it?
- [x] Tests!
  * `make check-rust` seems to pass locally, that's a start
  * Need to test that `break rust` and `break gcc` still compile just fine if `rust`/`gcc` is actually in scope
  * Need to test that `break (rust)` still causes the old error
  * How do I even test for something that is supposed to ICE?
- [x] clang-format is angry with me, need to appease it

### DIFF
--- a/gcc/rust/resolve/rust-ast-resolve-expr.cc
+++ b/gcc/rust/resolve/rust-ast-resolve-expr.cc
@@ -29,9 +29,9 @@ namespace Resolver {
 
 void
 ResolveExpr::go (AST::Expr *expr, const CanonicalPath &prefix,
-		 const CanonicalPath &canonical_prefix)
+		 const CanonicalPath &canonical_prefix, bool funny_error)
 {
-  ResolveExpr resolver (prefix, canonical_prefix);
+  ResolveExpr resolver (prefix, canonical_prefix, funny_error);
   expr->accept_vis (resolver);
 }
 
@@ -643,8 +643,10 @@ ResolveExpr::resolve_closure_param (AST::ClosureParam &param,
 }
 
 ResolveExpr::ResolveExpr (const CanonicalPath &prefix,
-			  const CanonicalPath &canonical_prefix)
-  : ResolverBase (), prefix (prefix), canonical_prefix (canonical_prefix)
+			  const CanonicalPath &canonical_prefix,
+			  bool funny_error)
+  : ResolverBase (), prefix (prefix), canonical_prefix (canonical_prefix),
+    funny_error (funny_error)
 {}
 
 } // namespace Resolver
--- a/gcc/rust/resolve/rust-ast-resolve-expr.cc
+++ b/gcc/rust/resolve/rust-ast-resolve-expr.cc
@@ -23,6 +23,7 @@
 #include "rust-ast-resolve-type.h"
 #include "rust-ast-resolve-pattern.h"
 #include "rust-ast-resolve-path.h"
+#include "diagnostic.h"
 
 namespace Rust {
 namespace Resolver {
@@ -105,6 +106,45 @@ ResolveExpr::visit (AST::AssignmentExpr &expr)
   VerifyAsignee::go (expr.get_left_expr ().get ());
 }
 
+/* The "break rust" Easter egg.
+
+   Backstory: once upon a time, there used to be a bug in rustc: it would ICE
+   during typechecking on a 'break' with an expression outside of a loop.  The
+   issue has been reported [0] and fixed [1], but in recognition of this, as a
+   special Easter egg, "break rust" was made to intentionally cause an ICE.
+
+   [0]: https://github.com/rust-lang/rust/issues/43162
+   [1]: https://github.com/rust-lang/rust/pull/43745
+
+   This was made in a way that does not break valid programs: namely, it only
+   happens when the 'break' is outside of a loop (so invalid anyway).
+
+   GCC Rust supports this essential feature as well, but in a slightly
+   different way.  Instead of delaying the error until type checking, we emit
+   it here in the resolution phase.  We, too, only do this to programs that
+   are already invalid: we only emit our funny ICE if the name "rust" (which
+   must be immediately inside a break-with-a-value expression) fails to
+   resolve.  Note that "break (rust)" does not trigger our ICE, only using
+   "break rust" directly does, and only if there's no "rust" in scope.  We do
+   this in the same way regardless of whether the "break" is outside of a loop
+   or inside one.
+
+   As a GNU extension, we also support "break gcc", much to the same effect,
+   subject to the same rules.  */
+
+/* The finalizer for our funny ICE.  This prints a custom message instead of
+   the default bug reporting instructions, as there is no bug to report.  */
+
+static void ATTRIBUTE_NORETURN
+funny_ice_finalizer (diagnostic_context *context, diagnostic_info *diagnostic,
+		     diagnostic_t diag_kind)
+{
+  gcc_assert (diag_kind == DK_ICE_NOBT);
+  default_diagnostic_finalizer (context, diagnostic, diag_kind);
+  fnotice (stderr, "You have broken GCC Rust. This is a feature.\n");
+  exit (ICE_EXIT_CODE);
+}
+
 void
 ResolveExpr::visit (AST::IdentifierExpr &expr)
 {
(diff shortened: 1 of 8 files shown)


bugaevc:
There are some tests now 🙂

> How do I even test for something that is supposed to ICE?

There's apparently `dg-ice` made exactly for this case!

> clang-format is angry with me, need to appease it

clang-format should be happy now; but I may have to revert some of the changes in case we decide on a longer error message text

gcc/rust/resolve/rust-ast-resolve-expr.cc
```diff
@@ -384,7 +389,18 @@ ResolveExpr::visit (AST::BreakExpr &expr)
...
-    ResolveExpr::go (expr.get_break_expr ().get (), prefix, canonical_prefix);
+    {
+      bool funny_error = false;
+      AST::Expr &break_expr = *expr.get_break_expr ().get ();
+      if (break_expr.get_ast_kind () == AST::Kind::IDENTIFIER)
+	{
+	  std::string ident
+	    = static_cast<AST::IdentifierExpr &> (break_expr).as_string ();
+	  if (ident == "rust" || ident == "gcc")
+	    funny_error = true;
```
philberty, line 400: Why not just emit the error here without the need for the funny_error flag?

philberty (commented):
I think its looking ok just not sure we need the the funny_error flag

gcc/rust/resolve/rust-ast-resolve-expr.cc
bugaevc: This was [suggested by @CohenArthur](https://github.com/Rust-GCC/gccrs/issues/1996#issuecomment-1491915749). Basically, at this point we don't yet know that there will be an error; we only want to emit an error if the identifier later fails to resolve. For example (this is `break-rust1.rs`, one of the test case added by this PR):

```rust
let rust = "crab";
let res = loop {
    break rust;
};
```

This should still compile just fine. [Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8edc59d0dcbcd30231eb182aae97a6ca). (Actually, as you can see by commenting out the definition of `rust` in the playground, rustc still emits the "boring" error if it encounters `break rust` inside a loop, because they do it at a different stage (when recovering from `break` outside of a loop). I didn't replicate that; gccrs will still ICE on that example.)

gcc/rust/resolve/rust-ast-resolve-expr.cc
```diff
@@ -120,6 +120,11 @@ ResolveExpr::visit (AST::IdentifierExpr &expr)
     {
       resolver->insert_resolved_type (expr.get_node_id (), resolved_node);
     }
+  else if (funny_error)
+    {
+      rust_error_at (expr.get_locus (), "oopsie woopsie! you broke GCC Rust");
+      internal_error_no_backtrace ("you broke me");
```
CohenArthur, line 126: I think we could use a `rust_fatal_error` here instead and not ICE - the ICE function prompts the user to submit a full bug report, which we don't want here
```suggestion
      rust_fatal_error (expr.get_locus (),
			"are you trying to break %s? how dare you?",
			expr.as_string ().c_str ());
```
outputs with the following message:

```
arthur@platypus ~/G/r/gccrs ((8657790c))> build/gcc/crab1 main.rs
main.rs:2:11: fatal error: are you trying to break rust? how dare you?
    2 |     break rust;
      |           ^~~~
compilation terminated.
```
which I think is less misleading than the current output

```
arthur@platypus ~/G/r/gccrs ((8657790c)) [1]> build/gcc/crab1 main.rs
main.rs:2:11: error: oopsie woopsie! you broke GCC Rust
    2 |     break rust;
      |           ^~~~
crab1: internal compiler error: you broke me
```

CohenArthur (approved):
The tests look good, as well as the implementation :) nice work! Once everything is approved we'll probably ask you to rebase your commits into one, but that can be done later. Thank you for working on this!

CohenArthur:
> Since #1996 has received positive feedback, here's a tentative implementation smile
> 
> TODO:
> 
>     * [ ]  Better / funnier error message
>       
>       * Suggestions welcome!
>       * ...or is the current one fine?
>       * It would be nice to make it a little bit more explicit that this is not _really_ an ICE, there is no bug to report, etc.

Submitted as a review suggestion :)

>         
>         * rustc says: "the compiler expectedly panicked. this is a feature."
> 
>     * [ ]  Fix/verify Changelog entries
>       
>       * I've no idea how to do GNU-style changelogs! I've read [this doc](https://www.gnu.org/prep/standards/html_node/Change-Logs.html) and skimmed through some other gccrs commits, and tried to write some entries based on that
>       * glibc people have accepted my patches without Changelog entries; are they really required?

Yes, we do have to write Changelog entries for GCC. You can run `contrib/gcc-git-customization.sh` to get new git subcommands, such as `git gcc-commit-mklog` which will produce a Changelog skeleton in your commit's body. You can use it like a regular `git commit`, so you can do things like `git gcc-commit-mklog --amend` to fix your existing commits.

> >     * [ ]  Try to make the error point at the whole statement and not just the name
>       
>       * Or... is it worth it?

Complicated and not worth it imo :) this is already nice and fun haha.

>     * [x]  Tests!
>       
>       * `make check-rust` seems to pass locally, that's a start
>       * Need to test that `break rust` and `break gcc` still compile just fine if `rust`/`gcc` is actually in scope
>       * Need to test that `break (rust)` still causes the old error
>       * How do I even test for something that is supposed to ICE?
> 
>     * [x]  clang-format is angry with me, need to appease it

gcc/rust/resolve/rust-ast-resolve-expr.cc
bugaevc: > I think we could use a `rust_fatal_error` here instead and not ICE

No, we really want this to be an ICE. The backstory here is that rustc [used to ICE](https://github.com/rust-lang/rust/issues/43162) (not just emit a regular error) when type-checking `break some_expr()` outside of a loop. This was fixed (to just emit an error), but `break rust` was intentionally kept an ICE, as a fun Easter egg. So we need to ICE too, a regular error won't cut it.

> the ICE function prompts the user to submit a full bug report, which we don't want here

Yes. But maybe we can get the best of both worlds:

```
$ gcc -frust-incomplete-and-experimental-compiler-do-not-use /tmp/break-rust3.rs
/tmp/break-rust3.rs:2:11: internal compiler error: are you trying to break rust? how dare you?
    2 |     break rust;
      |           ^~~~
You broke GCC Rust. This is a feature.
```
Yes, this is an ICE with rich location. How cool is that!

(Also, "broke" or "have broken"?)

bugaevc:
> Submitted as a review suggestion :)

And implemented 🙂

> Yes, we do have to write Changelog entries for GCC. You can run `contrib/gcc-git-customization.sh` to get new git subcommands, such as `git gcc-commit-mklog` which will produce a Changelog skeleton in your commit's body. You can use it like a regular git commit, so you can do things like `git gcc-commit-mklog --amend` to fix your existing commits.

Does that mean that my existing changelongs are not fine? Any specific things for me to fix?

CohenArthur:
> Does that mean that my existing changelongs are not fine? Any specific things for me to fix?

They're great! I hadn't checked them yet haha, I thought you hadn't written any and was looking for help on how to proceed. The ones you wrote are really good

philberty (approved):
LGTM i think i can live with a flag for this :D

bugaevc:
I think this still needs more comments in the source code (and not just on GH):
* _What_ we are doing (reporting a funny error, but only if this unresolved identifier is `rust` or `gcc` and is immediately inside a `break` expression) -- this would answer why we need a flag
* _Why_ we are doing it (why this particular Easter egg at all? why it must be an ICE and not just an error?)

philberty:
@bugaevc are you going to add another commit with more comments?

bugaevc:
Wrote a comment (several of them actually). Please tell me if these are comprehensive enough (and please point out any grammar / spelling / other mistakes) 🙂. I opted to fold this into the last commit though, and slightly expanded the commit message.

Assuming this passes CI, it should be good to go from my side 🚀

gcc/rust/resolve/rust-ast-resolve-expr.cc
bugaevc: > (Also, "broke" or "have broken"?)

Went for "have broken", but this can be reverted if "broke" is better / more correct.

gcc/rust/resolve/rust-ast-resolve-expr.cc
bugaevc: Hopefully the comment answers this now :)

bugaevc:
Thank you for guiding me! And please don't forget to mention this important feature in the next progress report :smiley:

eli-schwartz:
Fascinatingly, it appears the german translators have gone on strike: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114629

PR #2097, opened 2023-04-04
Support qualified trait item call expression

### LINKED ISSUE #2070: Failed to resolve CallExpr to trait item
I tried this code: https://godbolt.org/z/GxeszbfTb

```rust
trait Foo {
    fn get(self) -> i32;
}

struct Bar(i32);
impl Foo for Bar {
    fn get(self) -> i32 {
        self.0
    }
}

fn type_bound_test<T: Foo>(a: T) -> i32 {
    Foo::get (a) // <-- this call it works if you do a.get ()
}

fn main() {
    let a;
    a = Bar(456);

    let b;
    b = type_bound_test(a);
}

```

Instead, this happened: 

```
<source>:13:10: error: failed to resolve path segment using an impl Probe
   13 |     Foo::get (a)
      |          ^~~
<source>:13:5: error: failed to type resolve expression
   13 |     Foo::get (a)
      |     ^~~
<source>:13:5: error: Failed to resolve expression of function call
<source>:13:5: error: failed to type resolve expression
```

### Meta

- What version of Rust GCC were you using, git sha if possible. 3eb5dcb67c40a2d26fe05442d1d64dcb3b9d1d09

philberty:
Fixes #2070

### DIFF
--- a/gcc/rust/backend/rust-compile-base.cc
+++ b/gcc/rust/backend/rust-compile-base.cc
@@ -27,6 +27,9 @@
 #include "rust-diagnostics.h"
 #include "rust-expr.h"	// for AST::AttrInputLiteral
 #include "rust-macro.h" // for AST::MetaNameValueStr
+#include "rust-hir-path-probe.h"
+#include "rust-type-util.h"
+#include "rust-compile-implitem.h"
 
 #include "fold-const.h"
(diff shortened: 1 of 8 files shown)


CohenArthur (approved):
LGTM :)

PR #2099, opened 2023-04-05
ast: Add difference between attributes

### LINKED ISSUE #2096: Differentiate inner and outer attributes.
For now we have no way to differentiate an inner from an outer attribute in the AST. This distinction is required for proc macros.

P-E-P:
Add a boolean to tell inner and outer attributes ast nodes appart.

Fixes #2096

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -484,7 +484,7 @@ Parser<ManagedTokenSource>::parse_inner_attributes ()
 
 // Parse a inner or outer doc comment into an doc attribute
 template <typename ManagedTokenSource>
-AST::Attribute
+std::tuple<AST::SimplePath, std::unique_ptr<AST::AttrInput>, Location>
 Parser<ManagedTokenSource>::parse_doc_comment ()
 {
   const_TokenPtr token = lexer.peek_token ();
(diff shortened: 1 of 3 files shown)


gcc/rust/parse/rust-parse-impl.h
```diff
@@ -507,7 +507,12 @@ AST::Attribute
 Parser<ManagedTokenSource>::parse_inner_attribute ()
 {
   if (lexer.peek_token ()->get_id () == INNER_DOC_COMMENT)
-    return parse_doc_comment ();
+    {
+      auto values = parse_doc_comment ();
+      return AST::Attribute (std::move (std::get<0> (values)),
+			     std::move (std::get<1> (values)),
+			     std::get<2> (values), true);
```
CohenArthur, line 514: I think this would be a bit clearer if you destructured the tuple
```suggestion
      // probably missing some `&&` or std::move somewhere
      auto path = std::move(std::get<0>(values));
      auto input = std::move(std::get<1>(values));
      auto loc = std::get<2>(values);
      return AST::Attribute (path, input, loc, true);
```

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -533,7 +538,10 @@ Parser<ManagedTokenSource>::parse_inner_attribute ()
   if (!skip_token (LEFT_SQUARE))
     return AST::Attribute::create_empty ();
 
-  AST::Attribute actual_attribute = parse_attribute_body ();
+  auto values = parse_attribute_body ();
+  auto actual_attribute = AST::Attribute (std::move (std::get<0> (values)),
+					  std::move (std::get<1> (values)),
+					  std::get<2> (values), true);
```
CohenArthur, line 544: same here. And when we ~~rewrite it in Rust~~ switch to C++14/17, we can use `auto [path, input, loc] = parse_attribute_body ();`

gcc/rust/parse/rust-parse-impl.h
P-E-P: Fixed.

gcc/rust/parse/rust-parse-impl.h
P-E-P: Fixed.

CohenArthur (approved):
Thank you!

PR #2104, opened 2023-04-06
libproc_macro: Add literal type

P-E-P:
Add `Literal` type matching the rust's one.

### DIFF
--- a/libgrust/libproc_macro/literal.cc
+++ b/libgrust/libproc_macro/literal.cc
@@ -25,7 +25,6 @@
 #include <cstdlib>
 
 namespace Literal {
-
 extern "C" {
 
 void
@@ -55,23 +54,13 @@ Literal__drop (Literal *lit)
 Literal
 Literal__string (const unsigned char *str, std::uint64_t len)
 {
-  unsigned char *data = new unsigned char[len];
-  StringPayload str_payload = {data, len};
-  std::memcpy (data, str, len);
-  LiteralPayload payload;
-  payload.string_payload = str_payload;
-  return {STRING, payload};
+  return Literal::make_string (str, len);
 }
 
 Literal
 Literal__byte_string (const std::uint8_t *bytes, std::uint64_t len)
 {
-  std::uint8_t *data = new std::uint8_t[len];
-  ByteStringPayload bstr_payload = {data, len};
-  std::memcpy (data, bytes, len);
-  LiteralPayload payload;
-  payload.byte_string_payload = bstr_payload;
-  return {BYTE_STRING, payload};
+  return Literal::make_byte_string (bytes, len);
 }
 
 bool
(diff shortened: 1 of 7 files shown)


CohenArthur (approved):
I like the tagged unions approach, but I'm wondering if a more C++ like representation wouldn't be better. I very much like this one however. But we will need to add safe APIs to ensure people aren't accessing fields willy-nilly (so getters with assertions on the `tag` for example). This can come in a later PR

P-E-P:
> I like the tagged unions approach, but I'm wondering if a more C++ like representation wouldn't be better. I very much like this one however. But we will need to add safe APIs to ensure people aren't accessing fields willy-nilly (so getters with assertions on the `tag` for example). This can come in a later PR

Those types shall correspond directly to their rust counterpart, we're limited by the C abi here, we cannot rearrange types. I strictly followed this [RFC](https://github.com/rust-lang/rfcs/blob/master/text/2195-really-tagged-unions.md).

A commit with different abstractions is coming, but I still need to figure out some details (typically allocation failure report to the rust interface).

P-E-P:
Failing job is interesting, looks like the bootstrap is failing because of some ICE in the cpp compiler ?
```
../../../libgrust/libproc_macro/literal.cc: In function 'Literal Literal__byte_string(uint8_t*, uint64_t)':
../../../libgrust/libproc_macro/literal.cc:64:79: internal compiler error: in reshape_init_class, at cp/decl.c:5208
   return {.tag = BYTE_STRING, .payload = {.byte_string_payload = bstr_payload}};
```

libgrust/libproc_macro/literal.cc
```diff
@@ -0,0 +1,77 @@
...
+  LiteralPayload payload;
+  payload.byte_string_payload = bstr_payload;
+  return {BYTE_STRING, payload};
+}
+
+extern "C" bool
+Literal__from_string (const unsigned char *str, std::uint64_t len, Literal *lit)
+{
+  // FIXME: implement this function with parser
+  return false;
```
CohenArthur, line 76: I wonder if it wouldn't be better for there to be a panic there saying this isn't implemented yet

CohenArthur (approved):

libgrust/libproc_macro/literal.cc
P-E-P: This is indeed probably better.

P-E-P:
> The API looks great to me. One question is - how do we deal with running the actual destructors on the C++ side? As it stands right now, the structures have no destructors, so should we call `Literal__drop` from the C++ side as well? Do we _want_ a C++ destructor here or would that make ownershipt too difficult to manage?

We cannot have a direct dectructor in the struct since that would make it incompatible with C linkage, thus unusable on the rust side. Moreover, we cannot enforce move on this struct due to those abi specifications, hence why we also cannot enforce any mechanism of automated deletion. We could create a function that takes a `struct Literal &&` though, it would free the internal buffer before dropping the other fields @CohenArthur .

CohenArthur:
I'm happy with that :) Again, I think it's the sort of thing where we should merge it and then tweak it as we start using it to really handle proc macros :)

libgrust/libproc_macro/literal.cc
```diff
@@ -0,0 +1,281 @@
...
+// see the files COPYING3 and COPYING.RUNTIME respectively.  If not, see
+// <http://www.gnu.org/licenses/>.
+
+#include "literal.h"
+#include <cstring>
+#include <cstdlib>
+
+namespace Literal {
+
+extern "C" void
```
philberty, line 29: Would it be better to use an extern "C" block instead of specificly for each function?

philberty (approved):
Just minor style thing but looks good. I just realised though we probably need to change the clang-format automation to check the libgrust directory too not just the rust directory

CohenArthur (approved):
LGTM! Thank you!

libgrust/libproc_macro/literal.cc
```diff
@@ -0,0 +1,284 @@
...
+Literal
+Literal::make_u8 (std::uint8_t value, bool suffixed)
+{
+  UnsignedPayload unsigned_payload;
+  unsigned_payload.unsigned8 = value;
+  Unsigned val{UNSIGNED_8, unsigned_payload};
+  UnsignedSuffixPayload payload{val, suffixed};
+
+  return make_unsigned (payload);
+}
```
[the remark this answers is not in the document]
P-E-P: Fixed.

PR #2107, opened 2023-04-06
gccrs: Introduce AST::Visitable class for AST

dkm:
AST::Visitable is an abstract class with a unique accept_vis() method.
Make all abstract AST node class inherit from this class.

Allows for easy definition of operations on nodes that must accept a
visitor.

The static Dump::dump() is an example of such use: the static method
accepts any AST node, creates a Dump visitor and have it visit the AST
starting at the node.

This change also inserts a debug(Visitable&) function in the global
namespace to make it easy to call from the debugger (similar to
debug_tree or debug(rtx*) functions).

```
gcc/rust/ChangeLog:
            * ast/rust-ast-dump.cc (Dump::debug): New.
            * ast/rust-ast-dump.h (Dump::debug): Untemplate it.
            (debug): New.
            * ast/rust-ast.h (class Visitable): New.
            (class TokenTree): Inherit from Visitable.
            (class MacroMatch): Likewise.
            (class AttrInput): Likewise.
            (class MetaItemInner): Likewise.
            (class Pattern): Likewise.
            (classTypeParamBound): Likewise.
            (class GenericParam): Likewise.
            (class TraitItem): Likewise.
            (classInherentImplItem): Likewise.
            (class TraitImplItem): Likewise.
            (class ExternalItem): Likewise.
            (class SingleASTNode): Likewise.

```

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -46,8 +46,14 @@ enum Kind
   MACRO_INVOCATION,
 };
 
+class Visitable
+{
+public:
+  virtual void accept_vis (ASTVisitor &vis) = 0;
+};
+
 // Abstract base class for all AST elements
-class Node
+class Node : public Visitable
 {
 public:
   /**
@@ -72,7 +78,7 @@ enum DelimType
 class Token;
 
 // A tree of tokens (or a single token) - abstract base class
-class TokenTree
+class TokenTree : public Visitable
 {
 public:
   virtual ~TokenTree () {}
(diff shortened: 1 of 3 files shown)


gcc/rust/ast/rust-ast.h
```diff
@@ -876,7 +874,7 @@ class MetaListNameValueStr;
 
 /* Base statement abstract class. Note that most "statements" are not allowed in
  * top-level module scope - only a subclass of statements called "items" are. */
-class Stmt : public Node
+class Stmt : public Node, public Visitable
```
CohenArthur, line 877: Wouldn't it be easier for `Node` to inherit from `Visitable`? This way you don't have to tweak Expr, Stmt, Type, and maybe more classes. It also makes sense semantically IMO

CohenArthur (approved):
Great idea and great execution :) thank you!

gcc/rust/ast/rust-ast.h
dkm: Hmm maybe this could save a few lines and be more correct yes. I'll give it a try :)

dkm:
Implemented your suggestion @CohenArthur

PR #2113, opened 2023-04-09
Remove dependency on C++ for building libgrust

cepawiel:
Prior to this change building libgrust during a cross compile would fail to configure due to libstdc++ not having been built and would fail to find a valid C++ preprocessor. At this time nothing in libgrust requires C++ so this patch solves the issue by removing any C++ usage from the library.

libgrust/ChangeLog:
	* Makefile.in: Regenerated
	* configure: Regenerated from configure.ac
	* configure.ac: Removed AC_PROG_CXX
	* libproc_macro/Makefile.am: Changed .cc extension to .c
	* libproc_macro/Makefile.in: Regenerated from Makefile.am
	* libproc_macro/proc_macro.c: Renamed from .cc to .c

PR Checklist:

- \[X] GCC development requires copyright assignment or the Developer's Certificate of Origin sign-off, see https://gcc.gnu.org/contribute.html or https://gcc.gnu.org/dco.html
- \[X] Read contributing guidlines
- \[X] `make check-rust` passes locally
- \[ ] Run `clang-format`


tschwinge:
Is this really the correct fix -- I suppose we actually do want to use C++ in libgrust, @P-E-P?

P-E-P:
> Is this really the correct fix -- I suppose we actually do want to use C++ in libgrust, @P-E-P?

Yes C++ is required in libgrust, part of it is in rust but that's simply an interface for users to interact with it. #2104 add some cpp to libgrust and many more PR are waiting to be created.

cepawiel:
Just to be clear, having C++ support around to build libgrust will then require Rust support to be built as an additional GCC pass when cross-compiling (as far as I can tell), and will also require adding a dependency between libgrust and libstdc++ by adding `dependencies = { module=configure-target-libgrust; on=all-target-libstdc++-v3; };` to Makefile.def. None of which is an issue, just wanting to make sure that's the intended solution.

Using the Dreamcast homebrew scene's toolchain as a brief example (since that's why I'm working on this) of build steps:

1. Build GCC with --enable-languages=c
2. Build Newlib
3. Build GCC with --enable-languages=c,c++,objc,obj-c++
4. Build GCC with --enable-languages=c,c++,objc,obj-c++,rust

P-E-P:
Opened #2582 to collect every discussions on this issue. #2143 is still up (but stale) to fix this problem.

Issue #2021, opened 2023-03-24
Delete as_string HIR dump


philberty:
see https://github.com/Rust-GCC/gccrs/blob/7d38d01886da81d2c38673887dc8802efd0d5432/gcc/rust/rust-session-manager.cc#L924

PR #2114, opened 2023-04-10
Deleted the as_string AST dump

### LINKED ISSUE #2022: Delete as_string AST dump
see https://github.com/Rust-GCC/gccrs/blob/7d38d01886da81d2c38673887dc8802efd0d5432/gcc/rust/rust-session-manager.cc#L887

mvvsmk:
Fixes #2022 
Removed the Session::dump_ast function, and all the elements associated with it.

gcc/rust/ChangeLog:

	* rust-session-manager.cc (Session::enable_dump): Removed else if (arg == "parse").
	(Session::compile_crate): Removed the statement that called the dump_ast.
	(Session::dump_ast): Removed dump_ast function.
	* rust-session-manager.h (struct CompileOptions): Removed the PARSER_AST_DUMP option.

Signed-off-by: M V V S Manoj Kumar <mvvsmanojkumar@gmail.com>

Do let me know if I need to change something or missed anything .

### DIFF
--- a/gcc/rust/rust-session-manager.cc
+++ b/gcc/rust/rust-session-manager.cc
@@ -283,7 +283,7 @@ Session::enable_dump (std::string arg)
     {
       rust_error_at (
 	Location (),
-	"dump option was not given a name. choose %<lex%>, %<parse%>, "
+	"dump option was not given a name. choose %<lex%>, "
 	"%<register_plugins%>, %<injection%>, %<expansion%>, %<resolution%>,"
 	" %<target_options%>, %<hir%>, or %<all%>");
       return false;
@@ -296,10 +296,6 @@ Session::enable_dump (std::string arg)
   else if (arg == "lex")
     {
       options.enable_dump_option (CompileOptions::LEXER_DUMP);
-    }
-  else if (arg == "parse")
-    {
-      options.enable_dump_option (CompileOptions::PARSER_AST_DUMP);
     }
   else if (arg == "ast-pretty")
     {
@@ -495,10 +491,6 @@ Session::compile_crate (const char *filename)
   handle_crate_name (*ast_crate.get ());
 
   // dump options except lexer dump
-  if (options.dump_option_enabled (CompileOptions::PARSER_AST_DUMP))
-    {
-      dump_ast (parser, *ast_crate.get ());
-    }
   if (options.dump_option_enabled (CompileOptions::AST_DUMP_TOKENSTREAM))
     {
       dump_tokenstream (*ast_crate.get ());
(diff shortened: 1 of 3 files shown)


philberty (changes requested):
You should also remove the function:

```c++
// Parses crate and dumps AST to stderr, recursively.
void
Parser<ManagedTokenSource>::debug_dump_ast_output (AST::Crate &crate,
						   std::ostream &out)
{
  out << crate.as_string ();
}
```

From rust-parse-impl.h and it looks like clang-format is complaining

philberty:
Ah clang-format is complaining but its unreleated to your changes do a rebase ontop of master and see if that sorts it out

CohenArthur:
@mvvsmk the issue is that your original code has a whitespace after the curly bracket. The clang-format removes that extra whitespace.

I would really suggest running clang-format locally for things like this, which I agree are extremely annoying to fix haha. I have my clang-format set up to run whenever I save a file in my editor and it helps a lot. So the change you have to make is this:

```diff
-<space><space><space><space>}<space>
+<space><space><space><space>}
```

mvvsmk:
Thank you @CohenArthur for helping me out.  I did have clang-format to run when I saved the file, but idk why it has a problem with that perticular statement, as in my IDE auto formats it again for some reason even though I didn't touch while making my changes, I'll look into it and have it properly configured.

mvvsmk:
> You should also remove the function:
> 
> ```c++
> // Parses crate and dumps AST to stderr, recursively.
> template <typename ManagedTokenSource>
> void
> Parser<ManagedTokenSource>::debug_dump_ast_output (AST::Crate &crate,
> 						   std::ostream &out)
> {
>   out << crate.as_string ();
> }
> ```
> 
> From rust-parse-impl.h and it looks like clang-format is complaining

@philberty I had completed the changes you requested, could you please have a look at it again, I have also rebased the PR so that github dosen't show that the PR is outdated. :) this PR should also fix #2021

gcc/rust/rust-session-manager.cc
```diff
@@ -283,7 +283,7 @@ Session::enable_dump (std::string arg)
     {
       rust_error_at (
 	Location (),
-	"dump option was not given a name. choose %<lex%>, %<parse%>, "
+	"dump option was not given a name. choose %<lex%>, "
```
CohenArthur, line 286: oh, I had never realized that we were parsing the string directly here instead of doing magic in `gcc/rust/lang.opt` like we do for the other enums options we've added. I think this is fine but we should eventually think about moving it here maybe? but this is not relevant to your PR so no worries

CohenArthur (approved):
LGTM!

gcc/rust/rust-session-manager.cc
mvvsmk: We had another file for parsing command line args?, I will check that out. :D

mvvsmk:
update : just rebased the PR, so that it doesn't create any problems, my branch had became like 105 commits behind.

PR #2119, opened 2023-04-12
libproc_macro: Add Punct type

P-E-P:
Add Punct structure matching the rust definition.

Needs #2117

PR #2120, opened 2023-04-12
libproc_macro : Add Tokenstream type

P-E-P:
Add remaining structures. Most of those structures are interdependent hence the big patch.

Needs #2119


P-E-P:
I'm still unsure about the drop interface. A pointer match the rust interface but it may falsely carry the information that the pointer is freed afterward.

libgrust/libproc_macro/tokenstream.cc
```diff
@@ -0,0 +1,119 @@
...
+  data[size] = tree;
+  size++;
+}
+
+void
+TokenStream::drop (TokenStream *stream)
+{
+  for (std::uint64_t i = 0; i < stream->size; i++)
+    {
+      TokenTree::TokenTree::drop (stream->data + i);
```
CohenArthur, line 72:
```suggestion
      TokenTree::TokenTree::drop (&stream->data[i]);
```
nit: any reason you're not doing this? I personally find it a little clearer but if you don't agree with it your version is fine as well :)

CohenArthur (commented):
Looks good to me :DD Really nice work!

libgrust/libproc_macro/tokenstream.cc
P-E-P: Ok, rebased and changed.

PR #2123, opened 2023-04-14
Rust: Make 'tree'-level 'MAIN_NAME_P' work

tschwinge:
A thing I discovered while testing GCC/Rust for GCN target (AMD GPUs).  (Not a typical thing anybody would do -- but it should, and now does, work.)  See the Git commit log for details.

The first variant, commit 41c7e1d0a707f97327f097022e9e6fa7f6f13533 "Rust: Make 'tree'-level 'MAIN_NAME_P' work" would be the classic way of addressing this: set `main_identifier_node` to `main` as part of the front end setup, and then we need the `gcc/rust/backend/rust-compile-base.cc:HIRCompileBase::compile_function` bit, because the `main` function otherwise doesn't actually have the name `main`, but has the crate name prefixed, and only the `DECL_ASSEMBLER_NAME` forced to `main` (to satisfy the actual C library-level linkage, I suppose).

But then I came up with the second variant, commit 705de6d54ca95b8277a15fe9a2255af6168ae086 "Rust: Make 'tree'-level 'MAIN_NAME_P' work", which simply does set `main_identifier_node` to the actual `main` function name, with the crate name prefixed.

I didn't drop the first variant (bit instead `git revert`ed it), so that we have it available in Git history, in case it's actually preferable over the second variant.

### DIFF
--- a/gcc/rust/backend/rust-compile-base.cc
+++ b/gcc/rust/backend/rust-compile-base.cc
@@ -523,6 +523,12 @@ HIRCompileBase::compile_function (
 
   // we don't mangle the main fn since we haven't implemented the main shim
   bool is_main_fn = fn_name.compare ("main") == 0;
+  if (is_main_fn)
+    {
+      rust_assert (!main_identifier_node);
+      /* So that 'MAIN_NAME_P' works.  */
+      main_identifier_node = get_identifier (ir_symbol_name.c_str ());
+    }
   std::string asm_name = fn_name;
 
   unsigned int flags = 0;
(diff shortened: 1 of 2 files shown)


philberty (approved):
LGTM, I think your first commit looks like the way other front-ends do it, so how about going with that commit?

CohenArthur:
@tschwinge for some reason the merge bot sees merge conflicts, can you rebase and we can try again?

tschwinge:
As `git rebase` would "detach" Git commit IDs from their references in the commit logs, I decided to `git merge` this one manually, after checking that it's still good.

Issue #166, opened 2021-01-19
ErrorPropagationExpression


philberty:
https://doc.rust-lang.org/reference/expressions/operator-expr.html#the-question-mark-operator

PR #2124, opened 2023-04-14
Lower ErrorPropagationExpr from AST to HIR

bugaevc:
[I have no idea what I'm doing](https://github.com/vkurilin/ihniwid), but using a `?` was causing an ICE and now it does not (I get regular errors instead), so, good? A step towards https://github.com/Rust-GCC/gccrs/issues/166 perhaps?

### DIFF
--- a/gcc/rust/hir/rust-ast-lower-expr.cc
+++ b/gcc/rust/hir/rust-ast-lower-expr.cc
@@ -674,6 +674,21 @@ ASTLoweringExpr::visit (AST::DereferenceExpr &expr)
 }
 
 void
+ASTLoweringExpr::visit (AST::ErrorPropagationExpr &expr)
+{
+  HIR::Expr *propagating_expr
+    = ASTLoweringExpr::translate (expr.get_propagating_expr ().get ());
+
+  auto crate_num = mappings->get_current_crate ();
+  Analysis::NodeMapping mapping (crate_num, expr.get_node_id (),
+				 mappings->get_next_hir_id (crate_num),
+				 UNKNOWN_LOCAL_DEFID);
+  translated = new HIR::ErrorPropagationExpr (
+    mapping, std::unique_ptr<HIR::Expr> (propagating_expr),
+    expr.get_outer_attrs (), expr.get_locus ());
+}
+
+void
 ASTLoweringExpr::visit (AST::MatchExpr &expr)
 {
   translated = ASTLoweringExprWithBlock::translate (&expr, &terminated);
--- a/gcc/rust/hir/rust-ast-lower-expr.h
+++ b/gcc/rust/hir/rust-ast-lower-expr.h
@@ -111,6 +111,7 @@ public:
   void visit (AST::ContinueExpr &expr) override;
   void visit (AST::BorrowExpr &expr) override;
   void visit (AST::DereferenceExpr &expr) override;
+  void visit (AST::ErrorPropagationExpr &expr) override;
   void visit (AST::MatchExpr &expr) override;
   void visit (AST::RangeFromToExpr &expr) override;
   void visit (AST::RangeFromExpr &expr) override;


philberty (approved):
LGTM

This adds the HIR lowering with no desugaring, but we are probably missing the name-resolution for this and type-checking then we need code-generation etc...

philberty:
IF you fix the commit message format, then this is good to merge.

bugaevc:
Thanks!

> This adds the HIR lowering with no desugaring

Yes. I saw that there exists `HIR::ErrorPropagationExpr`, and understood it to mean that desugaring is expected to happen at some later stage, after the AST-to-HIR lowering. Was that the right call? At which stage should it be desugared?

I was going to look into how lang items are implemented next, to see if I could find/add `try_trait` / `try_trait_v2` and do something about it.

> IF you fix the commit message format, then this is good to merge.

I will, but that will have to wait until tomorrow.

CohenArthur (approved):
LGTM :) If you're interested in taking care of the later parts of the pipeline we'd be more than happy to help, if you want to chat on our Zulip to get quicker feedback on what to do etc feel free! Thank you :D

bugaevc:
> If you're interested in taking care of the later parts of the pipeline

I'd love to!

> if you want to chat on our Zulip to get quicker feedback on what to do etc feel free!

Maybe, but I must admit the prospect of joining yet another IM platform does not sound very appealing 🙂 ~~Can't we all just go back to IRC and plain text mailing lists~~. So while I'm undecided, would you be so kind as to give me some guidance here on GH?

* What is the compilation pipeline?
    * The source code is parsed into AST which is then lowered to HIR, that much I understand.
    * What happens next? GENERIC? GIMPLE? (Maybe you could point me to where in the source this happens, too.)
    * Is there a MIR?
* Does/should the desugaring happen during a HIR-to-HIR transform pass? Or maybe do different desugarings happen in different phases?
* What are the major differences between AST and HIR? From what I've seen so far (and done in this PR), it's mostly a 1-to-1 translation.
* What does this `Analysis::NodeMapping` thing even do?
    * I'm not dumb, I'm just not very familiar with GCC / gccrs internals, that's all 🙂
* For desugaring, I need to look up a trait marked with `lang = "try"`; how would I do that? Is there some sort of lang item -marked item registry where I could do something like `registry ()->look_up (RustLangItem::TRY)`?
    * I need to look into how traits for arithmetic ops are looked up and just do the same. But maybe you can just tell me the answer straight up.

bugaevc:
> * For desugaring, I need to look up a trait marked with lang = "try"; how would I do that? Is there some sort of lang item -marked item registry where I could do something like registry ()->look_up (RustLangItem::TRY)?
>     * I need to look into how traits for arithmetic ops are looked up and just do the same.

I found this:

https://github.com/Rust-GCC/gccrs/blob/74b4d6ddf6758addd3b93985aa56382fa00c43e1/gcc/rust/typecheck/rust-hir-type-check-expr.cc#L1581-L1595

(by the way, instead of just a `return false` this should probably emit a warning saying it could not find the required lang item?)

But also, the following looks just wrong:

https://github.com/Rust-GCC/gccrs/blob/74b4d6ddf6758addd3b93985aa56382fa00c43e1/gcc/rust/typecheck/rust-hir-type-check-expr.cc#L1597-L1616

We should not attempt to resolve a method named, say, `add`, but instead solely look at whether or not the receiver implements the one trait marked with the lang item. And indeed, testing the following:

```rust
#[lang = "add"]
pub trait Add<Rhs = Self> {
    type Output;
    fn add(self, rhs: Rhs) -> Self::Output;
}

trait AlsoAdd {
    fn add(self);
}

struct MyStruct(i32);

impl Add for MyStruct {
    type Output = i32;
    fn add(self, rhs: MyStruct) -> i32 {
        self.0 + rhs.0
    }
}

impl AlsoAdd for MyStruct {
    fn add(self) { }
}

fn foo() ->i32 {
    MyStruct(35) + MyStruct(42)
}
```

I get the error:

```
bad-add.rs:25:5: error: multiple candidates found for possible operator overload
   15 |     fn add(self, rhs: MyStruct) -> i32 {
      |     ~~
......
   21 |     fn add(self) { }
      |     ~~
......
   25 |     MyStruct(35) + MyStruct(42)
      |     ^~~~~~~~
bad-add.rs:25:5: error: cannot apply this operator to types MyStruct{MyStruct (0:i32)} and MyStruct{MyStruct (0:i32)}
bad-add.rs:25:5: error: failed to type resolve expression
```

Who cares whether or not there happens to be another method named `add` on my type! I'm not invoking `.add()`, I'm using `+`, the `Add` trait. [rustc playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=05b01a811c693299b37017e9c481f136)

CohenArthur:
> > If you're interested in taking care of the later parts of the pipeline
> 
> I'd love to!
> 
> > if you want to chat on our Zulip to get quicker feedback on what to do etc feel free!
> 
> Maybe, but I must admit the prospect of joining yet another IM platform does not sound very appealing slightly_smiling_face ~Can't we all just go back to IRC and plain text mailing lists~. 

I'll address all of the other points you raised in your message shortly, but I just wanted to say we also have an IRC channel and mailing list if you do prefer that :D #gccrust on irc.oftc.net and https://gcc.gnu.org/mailman/listinfo/gcc-rust. So if you'd rather chat on there, feel free!

bugaevc:
> but I just wanted to say we also have an IRC channel and mailing list if you do prefer that :D #gccrust on irc.oftc.net and https://gcc.gnu.org/mailman/listinfo/gcc-rust. So if you'd rather chat on there, feel free!

I was not being very serious, but thank you! 😄

The next round of questions:
* What is `TraitReference`, and how's it different from `HIR::Trait`? When would I want to use one vs the other?
* How do I programmatically do `<Type as Trait>::method(receiver)`? I see that for programmatically doing `receiver.method()`, there is `MethodResolver::Probe ()`, but that contains a non-trivial amount of logic.
* How are generic vs concrete items represented? How do I properly work with a "somewhat specialized" item, by which I mean something like the following:
    ```rust
    struct SomeItem<T>;

    trait Foo {
        type Bar;
    }

    fn my_func<U: Foo>() {
        SomeItem<U::Bar>;
    }
    ```

    Here, `SomeItem<U::Bar>` is not the generic `SomeItem<>`, but a seemingly concrete, monomorphized type, except not really, because the type parameter we instantiate it with is not a specific type like `i32`, but rather itself dependent on whatever `U` is. Also do you happen to know the proper terminology to talk about this?

CohenArthur:
> * What is the compilation pipeline?
>     * The source code is parsed into AST which is then lowered to HIR, that much I understand.

That's correct :)

>     * What happens next? GENERIC? GIMPLE? (Maybe you could point me to where in the source this happens, too.)

Our HIR is then translated to GCC's highest level intermediate representation, GENERIC (or TREE which is the same). That happens in all the visitors within `gcc/rust/backend/`, like `rust-compile-expr.cc` and so on.

>     * Is there a MIR?

There isn't :) we don't have the need for one at the moment.

> * Does/should the desugaring happen during a HIR-to-HIR transform pass? Or maybe do different desugarings happen in different phases?

Most of the desugaring will happen during the AST-lowering pass. I believe it could be the case for the error propagation expressions as well but they are very complex, so maybe not.

> * What are the major differences between AST and HIR? From what I've seen so far (and done in this PR), it's mostly a 1-to-1 translation.

It is a 1-to-1 translation except in the cases where it can be simplified, as well as for AST nodes which cannot exist this late in the pipeline (macro invocations for example). There are some exceptions to this rule like the `asm!` macro whose content needs to be available up until the codegen stage.

> * What does this `Analysis::NodeMapping` thing even do?
>     * I'm not dumb, I'm just not very familiar with GCC / gccrs internals, that's all 🙂

The `NodeMapping` class helps in keeping track of HIR nodes and the AST node they correspond to as well as the crate they originate from. The class is defined in `rust-hir-map.h`. It's complicated but very useful :D And this isn't a dumb question, it's not an easy thing to grasp nor do we have a lot of documentation about it.

> > * For desugaring, I need to look up a trait marked with `lang = "try"`; how would I do that? Is there some sort of lang item -marked item registry where I could do something like `registry ()->look_up (RustLangItem::TRY)`?
> >     * I need to look into how traits for arithmetic ops are looked up and just do the same. But maybe you can just tell me the answer straight up.

> I found this:

> https://github.com/Rust-GCC/gccrs/blob/74b4d6ddf6758addd3b93985aa56382fa00c43e1/gcc/rust/typecheck/rust-hir-type-check-expr.cc#L1581-L1595

yes, that's exactly what you'll have to do. I would suggest also looking at https://github.com/Rust-GCC/gccrs/blob/74b4d6ddf6758addd3b93985aa56382fa00c43e1/gcc/rust/typecheck/rust-hir-type-check-expr.cc#L1457-L1579

which does the same lookup as well as some error handling.

> (by the way, instead of just a return false this should probably emit a warning saying it could not find the required lang item?)

> But also, the following looks just wrong:

> https://github.com/Rust-GCC/gccrs/blob/74b4d6ddf6758addd3b93985aa56382fa00c43e1/gcc/rust/typecheck/rust-hir-type-check-expr.cc#L1597-L1616

> We should not attempt to resolve a method named, say, `add`, but instead solely look at whether or not the receiver implements the one trait marked with the lang item. And indeed, testing the following:

```rust
#[lang = "add"]
pub trait Add<Rhs = Self> {
    type Output;
    fn add(self, rhs: Rhs) -> Self::Output;
}

trait AlsoAdd {
    fn add(self);
}

struct MyStruct(i32);

impl Add for MyStruct {
    type Output = i32;
    fn add(self, rhs: MyStruct) -> i32 {
        self.0 + rhs.0
    }
}

impl AlsoAdd for MyStruct {
    fn add(self) { }
}

fn foo() ->i32 {
    MyStruct(35) + MyStruct(42)
}
```

> I get the error:

```
bad-add.rs:25:5: error: multiple candidates found for possible operator overload
   15 |     fn add(self, rhs: MyStruct) -> i32 {
      |     ~~
......
   21 |     fn add(self) { }
      |     ~~
......
   25 |     MyStruct(35) + MyStruct(42)
      |     ^~~~~~~~
bad-add.rs:25:5: error: cannot apply this operator to types MyStruct{MyStruct (0:i32)} and MyStruct{MyStruct (0:i32)}
bad-add.rs:25:5: error: failed to type resolve expression
```

> Who cares whether or not there happens to be another method named `add` on my type! I'm not invoking `.add()`, I'm using `+`, the `Add` trait. [rustc playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=05b01a811c693299b37017e9c481f136)

and I agree with this. Could you open an issue with the above code? it would be nice to have error handling here as well if the `add` lang item is not defined.

bugaevc:
Thank you!

> Could you open an issue with the above code?

I'd rather open a pull request :D but for that, please reply to the second set of questions.

philberty:
I think some of this will be a bit simpler when i merge my work over here: https://github.com/Rust-GCC/gccrs/pull/2140

Traits a complicated, so you can have a plain basic trait, so we need a way to work with them which is what the TraitReference object is, which is mostly just a helper object around it. The other part is the TypeBoundPredicate Rustc has something similar cant remember the name but this is all about managing generic traits.

Generic Traits are a pain and the current master code has one main issue here, where Trait has an implicit Self generic type parameter. This is left as a generic in the case of the Self type bound predicate but in every other case such as:

impl Foo<i32> for u8 {...}

This means we really create a type-bound predicate to wrap up the generics here properly which means we end up with

Foo<u8, i32>

The current master code does not track the Self parameter properly here which is confusing but i have fixes on the way. This is all required to handle complicated cases of associated types and trait disambiguation.

The other part to be aware of the is another bug in master for qualified path syntax such is always resolving the to the trait item and we are relying on code generation to always monomorphize it which is wrong again i have fixes for this laready.

You test case:

```
struct SomeItem<T>;

trait Foo {
    type Bar;
}

fn my_func<U: Foo>() {
    SomeItem<U::Bar>;
}
```

I havented chaged if this is working or broken but here this is a generic unit struct the generic U has a type bound predicate of Foo and then we its argument is the trait item of Bar but i have a feeling if this is not working its likely missing the piece i mentioned earlier thats not tracking the Self type properly here.

Associated types like this are really awkward because assoicated types can be the one implemented on the impl block or the placeholder on the trait which will need setup if possible when we get more information. So in the case the best case you can hope for us that SomeItem will turn into SomeItem<Placeholder-type> because there is not enough type info to get any further.

PR #2125, opened 2023-04-14
[WIP] Fix generic parameter parsing loop

### LINKED ISSUE #2102: Infinite loop on parse error in generics
Apparently we're not skipping the offending token. In addition, we're asking for the wrong string from the token, so we see an empty string.

I tried this code:

```rust
struct A<$
```

```
error: expected one of `#`, `>`, `const`, identifier, or lifetime, found `$`
 --> inf.rs:1:10
  |
1 | struct A<$
  |          ^ expected one of `#`, `>`, `const`, identifier, or lifetime

```

Instead, this happened:

```
inf.rs:1:10: error: attempted to get string for ‘$’, which has no string. returning empty string instead
    1 | struct A<$
      |          ^
inf.rs:1:10: error: unexpected token when parsing generic parameters: ‘’
inf.rs:1:10: error: attempted to get string for ‘$’, which has no string. returning empty string instead
inf.rs:1:10: error: unexpected token when parsing generic parameters: ‘’
inf.rs:1:10: error: attempted to get string for ‘$’, which has no string. returning empty string instead
inf.rs:1:10: error: unexpected token when parsing generic parameters: ‘’
inf.rs:1:10: error: attempted to get string for ‘$’, which has no string. returning empty string instead
inf.rs:1:10: error: unexpected token when parsing generic parameters: ‘’
[...]
```

### Meta

92b137048583c2c22c585097119dba6a5fe06228

powerboat9:
Should fix #2102, still trying to figure out how to write the test


philberty (changes requested):
LGTM from reading the issue i think a test case of:

```
struct A<$
// dg-error ...
```

What is the output you get now?

powerboat9:
I have a test case like that, but ```dg-error``` doesn't seem to detect duplicate errors.

CohenArthur:
@powerboat9 I think the test case is more about the infinite loop, so you should add it. If the testcase does not infinite loop, then #2102 is fixed in my opinion. Your fix looks good

philberty:
Yeah you might have success with trying the dg-excess-errors option if there are a cascade of errors

powerboat9:
Ok, I got it working.

CohenArthur (approved):
LGTM. I think there was an issue with using `dg-timeout` in our testsuite but I can't reproduce it, and not having it causes an infinite loop without the fix, so it's clearly needed.

P-E-P (approved):
LGTM

CohenArthur (approved):

PR #2122, opened 2023-04-13
Fix Bad formatting in AST expanded dump with associated types

ArberSephirotheca:
gcc/rust/ChangeLog:

        * ast/rust-ast-dump.cc: implement visit() function for QualifiedPathInType.
                                Fixed bad formatting of type alias in ast dump.

Fixes #2023 

Output of AST expanded dump when running the code the linked issue:
```
macro_rules! forward_ref_binop{
	(impl $imp:ident , $method:ident for $t:ty , $u:ty) => 
		{ forward_ref_binop ! ( impl $ imp , $ method for $ t , $ u , # [ stable ( feature = "rust1" , since = "1.0.0" ) ] ) ; }
	;
	(impl $imp:ident , $method:ident for $t:ty , $u:ty , # [$attr:meta]) => 
		{ # [ $ attr ] impl < a > $ imp < $ u > for & a $ t { type Output = < $ t as $ imp < $ u >> :: Output ; # [ inline ] fn $ method ( self , other : $ u ) -> < $ t as $ imp < $ u >> :: Output { $ imp :: $ method ( * self , other ) } } # [ $ attr ] impl < a > $ imp < & a $ u > for $ t { type Output = < $ t as $ imp < $ u >> :: Output ; # [ inline ] fn $ method ( self , other : & a $ u ) -> < $ t as $ imp < $ u >> :: Output { $ imp :: $ method ( self , * other ) } } # [ $ attr ] impl < a , b > $ imp < & a $ u > for & b $ t { type Output = < $ t as $ imp < $ u >> :: Output ; # [ inline ] fn $ method ( self , other : & a $ u ) -> < $ t as $ imp < $ u >> :: Output { $ imp :: $ method ( * self , * other ) } } }
	;
}

#[lang = "add"]
pub trait Add{
		type Output;

	fn add(self, rhs: RHS) -> Self::Output;

}

macro_rules! add_impl{
	($($t:ty)*) => 
		( $ ( # [ stable ( feature = "rust1" , since = "1.0.0" ) ] impl Add for $ t { type Output = $ t ; fn add ( self , other : $ t ) -> $ t { self + other } } forward_ref_binop ! { impl Add , add for $ t , $ t } ) * )
	;
}

impl Add for usize {
	type Output = usize;
	fn add(self, other: usize) -> usize {
		self + other /* tail expr */

	}


}

impl Add<usize> for &'a usize {
	type Output = <usize as Add<usize>>::Output;
[...]

PR #2128, opened 2023-04-17
Various tokenstream fixes

P-E-P:
Upon looking at #2122 I discovered a bug in `QualifiedPathInTypes`, the scope resolution token was wrong, a colon was used instead. Whilst investigating the bug I found a few more.

- Fix a case were outer attributes were incorrectly set as inner attributes
- Fix the lifetime token string representation to use a single quote alone
- Fix the output of a function's body when there is no return type.
- Fix missing path in function call tokenstream output
- Fix the original bug around scope resolution token output


CohenArthur (approved):
LGTM!

PR #2134, opened 2023-04-18
Add frust-extern option

P-E-P:
Add a new option to gather extern crates location.

gcc/rust/ChangeLog:

	* lang.opt: Add frust-extern option.
	* rust-session-manager.cc (Session::handle_extern_option): Add frust-extern option handler.
	* rust-session-manager.h (struct Session): Add an unordered map to store library name and locations.

### DIFF
--- a/gcc/rust/rust-session-manager.cc
+++ b/gcc/rust/rust-session-manager.cc
@@ -190,6 +190,11 @@ Session::handle_option (
       }
       break;
 
+      case OPT_frust_extern_: {
+	std::string input (arg);
+	ret = handle_extern_option (input);
+      }
+      break;
     case OPT_frust_crate_:
       // set the crate name
       if (arg != nullptr)
@@ -247,6 +252,20 @@ Session::handle_option (
     }
 
   return ret;
+}
+
+bool
+Session::handle_extern_option (std::string &input)
+{
+  auto pos = input.find ('=');
+  if (std::string::npos == pos)
+    return false;
+
+  std::string libname = input.substr (0, pos);
+  std::string path = input.substr (pos + 1);
+
+  extern_crates.insert ({libname, path});
+  return true;
 }
 
 bool
--- a/gcc/rust/rust-session-manager.h
+++ b/gcc/rust/rust-session-manager.h
@@ -290,6 +290,7 @@ struct Session
   /* This should really be in a per-crate storage area but it is wiped with
    * every file so eh. */
   std::string injected_crate_name;
+  std::map<std::string, std::string> extern_crates;
 
   /* extra files get included during late stages of compilation (e.g. macro
    * expansion) */
@@ -370,6 +371,8 @@ private:
 
   // handle cfg_option
   bool handle_cfg_option (std::string &data);
+
+  bool handle_extern_option (std::string &data);
 };
 
 } // namespace Rust
--- a/gcc/rust/lang.opt
+++ b/gcc/rust/lang.opt
@@ -58,6 +58,10 @@ frust-crate=
 Rust Joined RejectNegative
 -frust-crate=<name>             Set the crate name for the compilation
 
+frust-extern=
+Rust RejectNegative Joined
+-frust-extern=              Specify where an external library is located
+
 frust-debug
 Rust Var(flag_rust_debug)
 Dump various Rust front end internals.


gcc/rust/rust-session-manager.cc
```diff
@@ -249,6 +254,21 @@ Session::handle_option (
...
 }
 
+bool
+Session::handle_extern_option (std::string &input)
+{
+  auto pos = input.find ('=');
+  if (std::string::npos == pos)
+    {
+      return false;
+    }
```
CohenArthur, line 264:
```suggestion
  if (std::string::npos == pos)
      return false;
```
nit, also shouldn't we error out here? not sure why rustc accepts `rustc --extern <something_without_an_equal_sign>` and I haven't looked into it, so maybe we should not error out here

gcc/rust/rust-session-manager.cc
```diff
@@ -249,6 +254,21 @@ Session::handle_option (
...
+  std::string libname = input.substr (0, pos);
+  std::string path = input.substr (pos + 1);
+
+  extern_crates.insert ({libname, path});
```
CohenArthur, line 268: mh, I'm not sure if we want to keep this as a local hashmap or actually reach into the extern crate importer and add the candidate there. @philberty will have opinions :D

CohenArthur (commented):
I think this is a good start, but it does not yet cover the entirety of what we want from the option (which is fine). we'll also need to add tests later

gcc/rust/rust-session-manager.cc
P-E-P: No we shouldn't. At least rustc does not, it simply ignores the dependency.

gcc/rust/rust-session-manager.cc
P-E-P: I put a hashmap here because I thought this question was kinda out of scope for this PR. Some crates can be renamed with this mechanism, others are proc macros and should be loaded (immediately ? probably not).

P-E-P: In fact, rustc even use a [BTreeMap for this kind of thing](https://github.com/rust-lang/rust/blob/master/compiler/rustc_session/src/config.rs#L2181), and advise against any [unsorted containers](https://github.com/rust-lang/rust/blob/master/compiler/rustc_session/src/config.rs#L436).

gcc/rust/rust-session-manager.cc
CohenArthur: mh, I wonder why that is the case and why there isn't at least a warning or error. @bjorn3 @flip1995 any idea?

gcc/rust/rust-session-manager.cc
CohenArthur: yeah, that's fair. I'm happy with us keeping these crates in the `SessionManager` and then accessing them when loading crates later down the line. IMO it makes sense. and yeah I guess you should switch from `std::unordered_map` to a `std::map` which should be a sorted tree

gcc/rust/rust-session-manager.cc
P-E-P: I think this is due to the unstable feature with options [here](https://github.com/rust-lang/rust/blob/master/compiler/rustc_session/src/config.rs#L2238). But still, no option and no path seems to raise no error.

flip1995: I have no idea how the argument parsing and logic in rustc works, so I can't really help here. I just know that last time I looked, `rustc` uses https://docs.rs/getargs/latest/getargs/ to parse arguments.

gcc/rust/rust-session-manager.h
```diff
@@ -290,6 +290,7 @@ struct Session
   /* This should really be in a per-crate storage area but it is wiped with
    * every file so eh. */
   std::string injected_crate_name;
+  std::unordered_map<std::string, std::string> extern_crates;
```
CohenArthur, line 293: Do you want to switch this to a sorted tree and then we can merge it?
```suggestion
  std::map<std::string, std::string> extern_crates;
```

CohenArthur (approved):

gcc/rust/rust-session-manager.h
P-E-P: Rebased and changes :+1:

Issue #2144, opened 2023-04-21
Invalid branching in wildcard pattern


CohenArthur:
I tried this code:

```rust
enum Foo {
    A,
    B(i32),
}

fn main() -> i32 {
    let result = Foo::B(123);

    // { dg-warning "statement will never be executed" "" { target *-*-* } .+1 }
    let value = match result {
        A => 15,
        Foo::B(x) => x,
    };

    value - 15
}

```

I expected to see this happen: running this program exits with a `0` exit code.

The "equivalent" Rust code being:
```rust
enum Foo {
    A,
    B(i32),
}

fn main() {
    let result = Foo::B(123);

    // { dg-error "statement will never be executed" "" { target *-*-* } .+1 }
    let value = match result {
        A => 15,
        Foo::B(x) => x,
    };

    println!("{}", value - 15)
}
```

which gives out a lot of warnings but prints zero.

Instead, this happened: the code exited with `108` as a return value, which is `123 - 15`

### Meta

- What version of Rust GCC were you using: 428a31804b8

PR #2145, opened 2023-04-21
patterns: Add execution testcases to verify proper pattern matching

CohenArthur:
These testcases are similar to the ones added in #2072 but regarding
execution. One more issue was opened from them - #2144

gcc/testsuite/ChangeLog:

	* rust/execute/torture/issue-1852-1.rs: New test.
	* rust/execute/torture/issue-1852.rs: New test.


philberty (approved):
LGTM

PR #2147, opened 2023-04-21
libproc_macro: fix literal drop

### LINKED ISSUE #2146: Literal drop function may corrupt heap
The internal array of the Literal struct in libgrust uses a `new[]` operator, a `delete[]` operator should be used to free the corresponding memory safely. Currently a simple `delete` operator is used, which is an undefined behavior.

82e26f836210a2defed7e49d7d2655b5525f0561

P-E-P:
Needs #2120 

Fixes #2146


CohenArthur (approved):
LGTM :)

PR #2150, opened 2023-04-24
Proc macro type

P-E-P:
Add a proc_macro type to identify the location of a proc macro function in the proc macro library. This model is slightly based upon the rustc format, the main difference resides within the fact we do not use a bridge/server communication but instead a function pointer directly to the macro function.

### DIFF
--- a/libgrust/libproc_macro/proc_macro.cc
+++ b/libgrust/libproc_macro/proc_macro.cc
@@ -21,3 +21,32 @@
 // <http://www.gnu.org/licenses/>.
 
 #include "proc_macro.h"
+
+namespace ProcMacro {
+
+Procmacro
+Procmacro::make_derive (const char *trait_name, const char **attributes,
+			std::uint64_t size, CustomDeriveMacro macro)
+{
+  ProcmacroPayload payload;
+  payload.custom_derive = {trait_name, attributes, size, macro};
+  return {CUSTOM_DERIVE, payload};
+}
+
+Procmacro
+Procmacro::make_attribute (const char *name, AttributeMacro macro)
+{
+  ProcmacroPayload payload;
+  payload.attribute = {name, macro};
+  return {ATTR, payload};
+}
+
+Procmacro
+Procmacro::make_bang (const char *name, BangMacro macro)
+{
+  ProcmacroPayload payload;
+  payload.bang = {name, macro};
+  return {BANG, payload};
+}
+
+} // namespace ProcMacro
--- a/libgrust/libproc_macro/proc_macro.h
+++ b/libgrust/libproc_macro/proc_macro.h
@@ -42,9 +42,9 @@ using BangMacro = TokenStream (*) (TokenStream);
 struct CustomDerivePayload
 {
   // TODO: UTF-8 function name
-  char *trait_name;
+  const char *trait_name;
   // TODO: UTF-8 attributes
-  char **attributes;
+  const char **attributes;
   std::uint64_t attr_size;
   CustomDeriveMacro macro;
 };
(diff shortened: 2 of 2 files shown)


libgrust/libproc_macro/proc_macro.h
```diff
@@ -23,6 +23,69 @@
...
+  BangPayload bang;
+};
+
+struct Procmacro
+{
+  ProcmacroTag tag;
+  ProcmacroPayload payload;
+
+public:
+  Procmacro make_derive (char *trait_name, char **attribute, std::uint64_t size,
```
philberty, line 83: can this be const char* trait_name? where possible

libgrust/libproc_macro/proc_macro.cc
```diff
@@ -21,3 +21,32 @@
...
 #include "proc_macro.h"
+
+namespace Procmacro {
+
+Procmacro
+Procmacro::make_derive (char *trait_name, char **attributes, std::uint64_t size,
+			CustomDeriveMacro macro)
+{
+  ProcmacroPayload payload;
+  payload.custom_derive = {trait_name, attributes, size, macro};
```
philberty, line 32: I'm wondering should you map this over to a std::string in the struct so we dont have to worry about doing things like strdup

philberty (changes requested):
I think if the structs were managing within the extern C API we need to be careful with the string pointers here. Iff they are all internal strings then maybe its ok but if we are getting input from else where you might want to think about the memory management here and use strdup to take copies of the strings and have cleanup functions for the structs.

I would however be super wrong about whats going on lol

libgrust/libproc_macro/proc_macro.cc
P-E-P: To be fair I'm not even sure we need the constructors, these structures should be embedded in the final proc_macro library and thus will be directly mapped to the file's content. The strings won't ever be created from us but rather by the backend from the trait name etc... in the ast.

libgrust/libproc_macro/proc_macro.cc
```diff
@@ -21,3 +21,32 @@
 // <http://www.gnu.org/licenses/>.
 
 #include "proc_macro.h"
+
+namespace Procmacro {
```
CohenArthur, line 25:
```suggestion
namespace ProcMacro {
```
nit, feel free to ignore

CohenArthur (commented):
This looks good to me. I think we'll need to start using the abstractions to figure out how good it is to interact with them. I had the same concerns as @philberty regarding strings and memory allocations but I see what you mean by "the strings won't be created from us, but simply borrowed from the AST". In that case however, I'd like to see more `const char *` pointers if possible - since we're just borrowing, and we're not going to change these strings, it seems to me that it should be possible.

P-E-P:
> This looks good to me. I think we'll need to start using the abstractions to figure out how good it is to interact with them. I had the same concerns as @philberty regarding strings and memory allocations but I see what you mean by "the strings won't be created from us, but simply borrowed from the AST". In that case however, I'd like to see more `const char *` pointers if possible - since we're just borrowing, and we're not going to change these strings, it seems to me that it should be possible.

Is it even worth keeping those constructors ?

CohenArthur:
> > This looks good to me. I think we'll need to start using the abstractions to figure out how good it is to interact with them. I had the same concerns as @philberty regarding strings and memory allocations but I see what you mean by "the strings won't be created from us, but simply borrowed from the AST". In that case however, I'd like to see more `const char *` pointers if possible - since we're just borrowing, and we're not going to change these strings, it seems to me that it should be possible.
> 
> Is it even worth keeping those constructors ?

I think so, yeah :) we can always remove them later on

CohenArthur (approved):

PR #2152, opened 2023-04-25
builtin: Cleanup handling of builtin macros

CohenArthur:
This commit regroups information related to builtin macros in one place instead
of spreading it over multiple files. It also adds a simple bi-directional
hashmap in order to perform lookups from a key as well as a value.

gcc/rust/ChangeLog:

	* ast/rust-macro.cc (builtin_macro_from_string): Move function.
	* ast/rust-macro.h (enum class): Move enum.
	(builtin_macro_from_string): Move function.
	* expand/rust-macro-builtins.cc (builtin_macro_from_string): New function.
	(make_macro_path_str): Use new bi-map.
	(parse_single_string_literal): Use new `BuiltinMacro` enum.
	(MacroBuiltin::include_bytes_handler): Likewise.
	(MacroBuiltin::include_str_handler): Likewise.
	(MacroBuiltin::compile_error_handler): Likewise.
	(MacroBuiltin::concat_handler): Likewise.
	(MacroBuiltin::env_handler): Likewise.
	(MacroBuiltin::include_handler): Likewise.
	(MacroBuiltin::sorry): New function.
	* expand/rust-macro-builtins.h (enum class): Move enum here.
	(builtin_macro_from_string): New function declaration.
	* resolve/rust-early-name-resolver.cc (EarlyNameResolver::visit): Use
	new function.
	* util/rust-hir-map.cc (Mappings::insert_macro_def): Remove old
	builtin macro map.

### DIFF
--- a/gcc/rust/expand/rust-macro-builtins.cc
+++ b/gcc/rust/expand/rust-macro-builtins.cc
@@ -32,54 +32,91 @@
 #include "bi-map.h"
 
 namespace Rust {
+static const BiMap<std::string, BuiltinMacro> builtins = {{
+  {"assert", BuiltinMacro::Assert},
+  {"file", BuiltinMacro::File},
+  {"line", BuiltinMacro::Line},
+  {"column", BuiltinMacro::Column},
+  {"include_bytes", BuiltinMacro::IncludeBytes},
+  {"include_str", BuiltinMacro::IncludeStr},
+  {"stringify", BuiltinMacro::Stringify},
+  {"compile_error", BuiltinMacro::CompileError},
+  {"concat", BuiltinMacro::Concat},
+  {"env", BuiltinMacro::Env},
+  {"option_env", BuiltinMacro::OptionEnv},
+  {"cfg", BuiltinMacro::Cfg},
+  {"include", BuiltinMacro::Include},
+  {"format_args", BuiltinMacro::FormatArgs},
+  {"format_args_nl", BuiltinMacro::FormatArgsNl},
+  {"concat_idents", BuiltinMacro::ConcatIdents},
+  {"module_path", BuiltinMacro::ModulePath},
+  {"asm", BuiltinMacro::Asm},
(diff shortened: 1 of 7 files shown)


gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -31,54 +31,116 @@
 #include "rust-session-manager.h"
 
 namespace Rust {
+
+// very simple bi-directional hashmap
+template <typename K, typename V> class BiMap
```
P-E-P, line 36: Wouldn't this class fit more into the utils folder ?

P-E-P (approved):
Small nit, otherwise lgtm. Nice work :)

gcc/rust/expand/rust-macro-builtins.cc
CohenArthur: Yes agreed, but I don't see it being used very often so I've put it here for now. I don't mind moving it if you think it's useful :)

P-E-P: Tbh splitting it now would help other people finding it and reusing it. We probably do not want to deter other people to use such a nice tool simply because it requires a quick refactor before using it.

powerboat9:
It looks like you may need to use ```std::underlying_type<BuiltinMacro>::type``` as the bidirectional map key, or provide a custom implementation of ```std::hash<BuiltinMacro>```.

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=60970

CohenArthur:
> It looks like you may need to use `std::underlying_type<BuiltinMacro>::type` as the bidirectional map key, or provide a custom implementation of `std::hash<BuiltinMacro>`.
> 
> https://gcc.gnu.org/bugzilla/show_bug.cgi?id=60970

ah yeah, this is an issue we've had in the past. what we usually do is implement hash for our enum types. I'll take care of it

PR #2156, opened 2023-04-26
Fix tail expression stripping

powerboat9:
Should address #1446

### DIFF
--- a/gcc/rust/ast/rust-ast.cc
+++ b/gcc/rust/ast/rust-ast.cc
@@ -4213,6 +4213,31 @@ Attribute::is_parsed_to_meta_item () const
   return has_attr_input () && attr_input->is_meta_item ();
 }
 
+void
+BlockExpr::strip_tail_expr ()
+{
+  if (expr)
+    {
+      expr = nullptr;
+
+      // HACK: try to turn the last statement into a tail expression
+      if (statements.size () && statements.back ()->is_expr ())
+	{
+	  auto &stmt = static_cast<ExprStmt &> (*statements.back ());
+
+	  if (stmt.get_type () == ExprStmt::ExprStmtType::WITH_BLOCK)
+	    {
+	      auto &stmt_block = static_cast<ExprStmtWithBlock &> (stmt);
+	      if (!stmt_block.is_semicolon_followed ())
+		{
+		  expr = std::move (stmt_block.get_expr ());
+		  statements.pop_back ();
+		}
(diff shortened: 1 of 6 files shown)


philberty (commented):
This is not a bad solution actually... i wonder will it solve our big-endian test case issue

philberty (approved):
I think this seems like a reasonable approach for this case but i would like to see what @CohenArthur and @P-E-P think

CohenArthur (approved):
This is a great idea! Good job @powerboat9 :D can you please add the testcases from the issue you reference? I'm in favor of merging this and like the solution

P-E-P (approved):

powerboat9:
Do I have to worry about copyright assignment if I add the test from #1446 verbatim?

CohenArthur:
> Do I have to worry about copyright assignment if I add the test from #1446 verbatim?

I'm pretty sure it's fine unless specified otherwise. That test comes from the standard library anyway, and is just an expanded/adapted version (sorry for the lines numbers - quick ripgrep and didn't look up the proper option)

```rust
421-            #[stable(feature = "rust1", since = "1.0.0")]
422-            #[rustc_const_stable(feature = "const_int_conversions", since = "1.32.0")]
423-            #[inline]
424:            pub const fn to_le(self) -> Self {
425-                #[cfg(target_endian = "little")]
426-                {
427-                    self
428-                }
429-                #[cfg(not(target_endian = "little"))]
430-                {
431-                    self.swap_bytes()
432-                }
433-            }
```

powerboat9:
I'm not sure whether I should stick a ```Signed-off-by``` on the last commit because it comes from the standard library, someone will probably have to verify that everything is fine while upstreaming

CohenArthur:
> I'm not sure whether I should stick a `Signed-off-by` on the last commit because it comes from the standard library, someone will probably have to verify that everything is fine while upstreaming

It doesn't come from the standard library, it's been adapted. I've asked GCC people on IRC and it's never been considered an issue to commit testcases submitted by someone else when you are fixing the behavior, unless specifed otherwise. If you're happy with the commit please sign-off on it as it will be needed for proper DCO when upstreaming. You can also craft a similar testcase if you'd like instead :)

CohenArthur:
> Ok, it's fixed now. I just wanted to be cautious -- copyright assignment seems pretty important.

yes haha, and you did good :D better safe than sorry!

Issue #2041, opened 2023-03-24
ICE in match expression pattern


philberty:
### Code

Tried https://github.com/rust-lang/rust/blob/aa3ca1994904f2e056679fce1f185db8c7ed2703/src/test/ui/issue-15896.rs

https://godbolt.org/z/aYYPjYaPq

```rust
fn main() {
    enum R { REB(()) }
    struct Tau { t: usize }
    enum E { B(R, Tau) }

    let e = E::B(R::REB(()), Tau { t: 3 });
    let u = match e {
        E::B(
          Tau{t: x},
          //~^ ERROR mismatched types
          //~| expected type `main::R`
          //~| found type `main::Tau`
          //~| expected enum `main::R`, found struct `main::Tau`
          _) => x,
    };
}
```

### Meta

- What version of Rust GCC were you using, git sha if possible. be27571c110d99fe55324a9c2b6622787b08fb52

### Error output

```
crab1: internal compiler error: in visit, at rust/resolve/rust-ast-resolve-pattern.cc:135
```

<details><summary>Backtrace</summary>
  <p>
  
  ```
0x23e463e internal_error(char const*, ...)
	???:0
0xad9596 fancy_abort(char const*, int, char const*)
	???:0
0xce55cd Rust::Resolver::PatternDeclaration::visit(Rust::AST::StructPattern&)
	???:0
0xce46f7 Rust::Resolver::PatternDeclaration::visit(Rust::AST::TupleStructPattern&)
	???:0
0xce4844 Rust::Resolver::PatternDeclaration::go(Rust::AST::Pattern*, Rust::Resolver::Rib::ItemType, std::vector<Rust::Resolver::PatternBinding, std::allocator<Rust::Resolver::PatternBinding> >&)
	???:0
0xcea423 Rust::Resolver::ResolveExpr::visit(Rust::AST::MatchExpr&)
	???:0
0xce78a7 Rust::Resolver::ResolveExpr::go(Rust::AST::Expr*, Rust::Resolver::CanonicalPath const&, Rust::Resolver::CanonicalPath const&)
	???:0
0xcf1335 Rust::Resolver::ResolveStmt::visit(Rust::AST::LetStmt&)
	???:0
0xce8a7a Rust::Resolver::ResolveExpr::visit(Rust::AST::BlockExpr&)
	???:0
[...]

PR #2164, opened 2023-04-27
Add accessors to HIR::StructPatternFieldIdentPat

powerboat9:
A small patch to address #2041, split off since it's a shared dependency for two other patches.


philberty:
what on earth is going on with github....

PR #2170, opened 2023-04-27
Tokenstream conversion

### LINKED ISSUE #1398: Allow keywords to be regular identifiers for proc macro server
In rustc itself keywords are lexed as regular identifiers AFAIK. Only the parser distinguishes between them. Even if you make the lexer distinguish between them in the lexer of gccrs, for proc macro support you will need to emulate the rustc behavior as there is no separate keyword token type in the proc macro api.

_Originally posted by @bjorn3 in https://github.com/Rust-GCC/gccrs/pull/1397#issuecomment-1188934444_

P-E-P:
Add a way to convert a vector of token to a rust "Tokenstream" type back and forth.

Needs #2153 
Fixes #1398


gcc/rust/util/rust-token-converter.cc
```diff
@@ -0,0 +1,586 @@
...
+			 const const_TokenPtr &token)
+{
+  std::string::size_type sz;
+  auto str = token->as_string ();
+  switch (token->get_type_hint ())
+    {
+      case CORETYPE_F32: {
+	auto value = std::stof (str, &sz);
+	ts.push (ProcMacro::TokenTree::make_tokentree (
+	  ProcMacro::Literal::make_f32 (value, sz == str.length ())));
```
CohenArthur, line 45: so floats are suffixed when the size is equal to the token's string's size? So if we have the token `15f32`, which is suffixed (unless I'm misunderstanding?): we'll `stdof` it, get `15` and a size of 2 in `sz` (?). So the expression `sz == str.length()` will be false and our f32 literal won't be suffixed?

gcc/rust/util/rust-token-converter.cc
```diff
@@ -0,0 +1,586 @@
...
+      break;
+    case CORETYPE_UINT:
+      uvalue = std::stoull (str, &sz);
+      ts.push (ProcMacro::TokenTree::make_tokentree (
+	ProcMacro::Literal::make_usize (uvalue, sz == str.length ())));
+      break;
+    case CORETYPE_UNKNOWN:
+      svalue = std::stoll (str, &sz);
+      ts.push (ProcMacro::TokenTree::make_tokentree (
+	ProcMacro::Literal::make_i32 (svalue, false)));
```
CohenArthur, line 123: does `rustc` transform all integer tokens without a type hint into i32s?

gcc/rust/util/rust-token-converter.cc
```diff
@@ -0,0 +1,586 @@
...
+	  case EQUAL_EQUAL: {
+	    auto str = token->as_string ();
+	    auto it = str.cbegin ();
+	    for (; it != str.cend () - 1; it++)
+	      {
+		trees.back ().push (ProcMacro::TokenTree::make_tokentree (
+		  ProcMacro::Punct::make_punct (*it, ProcMacro::JOINT)));
+	      }
+	    trees.back ().push (ProcMacro::TokenTree::make_tokentree (
+	      ProcMacro::Punct::make_punct (*it, ProcMacro::ALONE)));
```
CohenArthur, line 256: so we push back a `JOINT` punc token and an `ALONE` one?

gcc/rust/util/rust-token-converter.cc
```diff
@@ -0,0 +1,586 @@
...
+ * @param result Reference to the output vector.
+ */
+static void
+from_ident (const ProcMacro::Ident &ident, std::vector<const_TokenPtr> &result)
+{
+  std::string value (reinterpret_cast<const char *> (ident.val), ident.len);
+  if (ident.is_raw)
+    {
+      value = "r#" + value;
+    }
```
CohenArthur, line 324:
```suggestion
  if (ident.is_raw)
      value = "r#" + value;
```

gcc/rust/util/rust-token-converter.cc
```diff
@@ -0,0 +1,586 @@
...
+}
+
+static void
+string_literal (const ProcMacro::StringPayload &payload,
+		std::vector<const_TokenPtr> &result)
+{
+  // TODO: UTF-8 string
+  result.push_back (Token::make_string (
+    Location (),
+    std::string (reinterpret_cast<const char *> (payload.data), payload.len)));
```
CohenArthur, line 339: why is the `reinterpret_cast` needed if `payload` is already `const`?

gcc/rust/util/rust-token-converter.cc
```diff
@@ -0,0 +1,586 @@
...
+    case ProcMacro::NONE:
+      from_tokenstream (g.stream, result);
+      break;
+    default:
+      gcc_unreachable ();
+    }
+}
+
+/**
+ * Dispatch TokenTree's conversion to it's inner type depending on it's tag.
```
CohenArthur, line 531:
```suggestion
 * Dispatch TokenTree's conversion to its inner type depending on its tag.
```

gcc/rust/util/rust-token-converter.cc
```diff
@@ -0,0 +1,586 @@
...
+ */
+static void
+from_tokenstream (const ProcMacro::TokenStream &ts,
+		  std::vector<const_TokenPtr> &result)
+{
+  std::vector<std::uint32_t> punct_accumulator;
+  for (std::uint64_t i = 0; i < ts.size; i++)
+    {
+      from_tokentree (ts.data[i], punct_accumulator, result);
+    }
```
CohenArthur, line 575:
```suggestion
  for (const auto &token: ts.data)
      from_tokentree (token, punct_accumulator, result);
```

gcc/rust/util/rust-token-converter.cc
```diff
@@ -0,0 +1,586 @@
...
+      switch (token->get_id ())
+	{
+	// Literals
+	case FLOAT_LITERAL:
+	  dispatch_float_literals (trees.back (), token);
+	  break;
+	case INT_LITERAL:
+	  dispatch_integer_literals (trees.back (), token);
+	  break;
+	// FIXME: Why does BYTE_CHAR_LITERAL is not handled by rustc ?
```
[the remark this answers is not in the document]
P-E-P: No, I really mean rustc, not gccrs. I don't know why but at the time I've written this there was no byte char literal. I'll double check to be sure though.

gcc/rust/util/rust-token-converter.cc
P-E-P: Yes, there is an inversion of the condition here (and everywhere this mechanism is applied), good catch.

gcc/rust/util/rust-token-converter.cc
P-E-P: No, we push all characters but the last one as joint if any, and then the last one `alone`. This way both single and complex punctuation are handled.

gcc/rust/util/rust-token-converter.cc
P-E-P: This is due to a rust interface incompatibility, cpp string requires a `const char *` while rust works with `const unsigned char *`. I'm aware this is clunky at best but I have not better solution for now. This will be solved with utf-8 introduction in gccrs though.

gcc/rust/util/rust-token-converter.cc
```diff
@@ -0,0 +1,586 @@
...
+      break;
+      case CORETYPE_F64: {
+	auto value = std::stod (str, &sz);
+	ts.push (ProcMacro::TokenTree::make_tokentree (
+	  ProcMacro::Literal::make_f64 (value, sz == str.length ())));
+      }
+      break;
+    default:
+      gcc_unreachable ();
+    }
```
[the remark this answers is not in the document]
P-E-P: @CohenArthur This comment highlight a situation that bother me a little. We're going back from an AST all the way down to a token vector. `Literals` in rust proc_macro [require me to choose a type](https://doc.rust-lang.org/proc_macro/struct.Literal.html) as there is no plain "integer"/"float" literal. Not relying on a the `CORETYPE` hint means I would have to set a "default" (i32 and f64) type which could then be promoted to a different "true" type. While this could potentially work I'm not entirely convinced by this approach, something feels a bit off.

gcc/rust/util/rust-token-converter.cc
P-E-P: Fixed in c94c6886b1ff39b39c8bebffa0c701baa1f7d3af

gcc/rust/util/rust-token-converter.cc
P-E-P: Fixed in [0182ef1](https://github.com/Rust-GCC/gccrs/pull/2170/commits/0182ef1edc9d0e5431cd2b5e5b552754886f997d)

gcc/rust/util/rust-token-converter.cc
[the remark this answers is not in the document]
P-E-P: The documentation [talks about the byte character](https://doc.rust-lang.org/proc_macro/struct.Literal.html) but there is no function for it. They may be using the unsufixed u8 ?

philberty:
@P-E-P do you want me to merge this or would you rather wait for @CohenArthur from vacation?

gcc/rust/util/rust-token-converter.cc
P-E-P: Fixed with rebase.

gcc/rust/util/rust-token-converter.cc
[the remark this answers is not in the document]
P-E-P: Looks likeI was wrong on this one, the proc_macro interface does not expose it but there is an `Integer` type.

For the following input:
```rust
pub fn default_type() { let a = 3 ; }
```
We get the following dump:
```
(type: TokenStream [Ident { ident: "pub", span: #0 bytes(387..390) }, Ident { ident: "fn", span: #0 bytes(391..393) }, Ident { ident: "default_type", span: #0 bytes(394..406) }, Group { delimiter: Parenthesis, stream: TokenStream [], span: #0 bytes(406..408) }, Group { delimiter: Brace, stream: TokenStream [Ident { ident: "let", span: #0 bytes(415..418) }, Ident { ident: "a", span: #0 bytes(419..420) }, Punct { ch: '=', spacing: Alone, span: #0 bytes(421..422) }, Literal { kind: Integer, symbol: "3", suffix: None, span: #0 bytes(423..424) }, Punct { ch: ';', spacing: Alone, span: #0 bytes(424..425) }], span: #0 bytes(409..427) }])
```

where we can find the following:

```
Literal { kind: Integer, symbol: "3", suffix: None, span: #0 bytes(423..424) }
```

gcc/rust/util/rust-token-converter.cc
```diff
@@ -0,0 +1,586 @@
...
+static void
+dispatch_integer_literals (ProcMacro::TokenStream &ts,
+			   const const_TokenPtr &token)
+{
+  std::string::size_type sz;
+  auto str = token->as_string ();
+  unsigned long long uvalue;
+  long long svalue;
+
+  switch (token->get_type_hint ())
```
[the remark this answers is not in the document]
P-E-P: For reference https://github.com/Rust-GCC/gccrs/pull/2170#discussion_r1188340315

gcc/rust/util/rust-token-converter.cc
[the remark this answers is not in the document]
P-E-P: After a deeper look into it, there is no constructor but there is an internal representation for byte char. As there is multiple flaws with the current literal implementation, I won't fix this in this PR, but rather rework the literal representation entirely.

gcc/rust/util/rust-token-converter.cc
CohenArthur: that's fine with me - just thought it would be possible at first glance, but I forgot we were doing the allocation manually for `ts.data`. all good :)

CohenArthur:
Rebased manually since the merge queue was stuck on it for three days

PR #2174, opened 2023-04-27
Type check StructPatternFieldIdentPat

powerboat9:


philberty (approved):
LGTM nice work.

Side thing to check (not related to your PR) we use a break to get out of the switch statement here but i am just wondering what would happen if we just returned early after an error if that would be safer/simpler but maybe we miss a chance to keep looking for more errors. Do you have an opinion here about the error handling.

PR #2175, opened 2023-04-28
Make key and key/value config options seperate

### LINKED ISSUE #2101: `cfg(key = "value")` should not imply `cfg(key)`
If there is any cfg item with a value for some key, then we consider cfg(key) to be true. This is wrong.

I tried this code with `-frust-cfg=key="value"`:

```rust
#[cfg(key)]
pub fn f() {}
```

I expected to see this happen: `f` is not compiled

Instead, this happened: `f` is compiled

[Godbolt with comparison to rustc](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DEArgoKkl9ZATwDKjdAGFUtEywZ7HAGTwNMAHLuAEaYxHoADqgKhHYMLm4ekdGxAr7%2BQSyh4VyWmNa2AkIETMQECe6euVaYNnHFpQTpgSFheuaNFUnVJWXNmdkSAJSWqCbEyOwcAKQATADM0wCsAELIVMAQANaYAJ5DywAi0xoAghEmwQDUVAw3EENX0wDsKy%2BHHCO0nEu8nhxaUioTgAJTMBCuCjGE0wTwWPFIBE0nxGWxASw0%2Bk4kj%2ByKBnF4ChAmKRAM%2BpDgsBgiBQqBYEToYXIlDQ9MZ4WIXAAbAAOTFYABueEmADU8JgAO4AeQijE4CJotAIYSJEGCeOCflKu3lvE1zGIuylwW0tVJCNZbEEUoYtB1ZNIWBYhmA4gd%2BGIZrwAswRIdmFUtRMyt15EE%2BTxtDwwWI2pcWDxBGIeBYupGVAMwAUYslMrl3F4/EEIjE7CkMkEihU6gdulm%2BhdIFM5n00aJkBGqAihQYfoAtH31sBpvNDjtdiPjksnHNZgKxCZMMsZ7NZoT8l77BBHF0qqQfH4Wlk2rkojEe7vkue4v1Wjk8gV6r1yq5Ku0N3Uis/b8f7x0ypePSND%2BgxcCMULjJMwxYhwvykP8gLAhwVzNgQyBXFwAB0vKYRoVwQLghAkHC8xgbwpJaEMqLopi3wcDipCphi8F4khhLEoiyIjJSNKsgy9BkBQEB8eyIDAMgyDEAofYKMwEQKAgqAWIKwqYDm0qyv8Cp0MqUmUOqDr6tqoZGYaxqmjYoaWowBA2naeJOi6bqAh6Xo%2Bn6gIBkGIYFmGyp0YCUYxnGGBTICSYpmmfCZtm4oafmCqyCW4jlkW8hKGoeK6LkBhGE24KtsE7YPEC3ZxP2VCoYOGyTuOk7LrO85uEu06zuuj5bjur7dPuDDoCBJ6kGeqTxN1e7DT2A1/h%2BPYNABY3vh1DBzU0h4DIN/4voke6bVN0EQTC0F0XBCG8EhVyqDyXJ9lykhXAA4k4TggkI%2BGPc9r3OuYYSPAR%2BBEMQJFkZxZJUaQCCYEwWDhCVdEMUxmKnfiHDsSSXHUcxdHzLiDpsSDlFfJwa4sbjBL4yipA%2BlJcQgJIQA%3D%3D)

**N.B.** rustc apparently had our current behaviour before 1.6.

### Meta

92b137048583c2c22c585097119dba6a5fe06228

powerboat9:
Fixes #2101


philberty (commented):
Hmm i am not sure whats going on here, why do we need a set of the keys? Can we not just use the map.find ? not 100% sure whats going on here.

powerboat9:
```features_kv``` stores ```cfg(key = "value")``` entries and ```features_k``` stores ```cfg(key)``` entries.

flip1995:
No idea how rustc implements this. I think there is only one map for known config values: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_session/config/struct.CheckCfg.html#structfield.expecteds But I'm not sure if that was exactly the question.

powerboat9:
```ExpectedValues<T>``` stores a hashset with key type ```Option<T>```. Since ```std::optional``` is C++ 17, I used a seperate hashset to store ```(x, None)``` entries.

CohenArthur:
@powerboat9 how about using the same hashmap with an empty hashset as the `None` equivalent? I think your change is fine, but having the two maps could get a bit confusing - keeping a single `unordered_map<string, unordered_set<string>>` would work fine. We also have `Optional` in `gcc/rust/util/` if you'd like to try that

powerboat9:
Waiting on #2211 and #2213 now for a ```unordered_set<Optional<std::string>>```

powerboat9:
Now waiting on #2224, but that should be it

CohenArthur (commented):
I think this is good. I'm wondering if we could simplify part of the code, as I want to make sure we can maintain it in the future, but overall LGTM :) just one comment but I'm in favor of merging this

powerboat9:
Implemented suggestion

CohenArthur (approved):
Thank you! Looks great :)

PR #2183, opened 2023-05-03
fixup: Fix bootstrap build

### LINKED ISSUE #2181: Bootstrap build failure
Bootstrap build is failing on master since afcde5f8175 with the following error:
```
make[3]: *** No rule to make target '../libgrust/libproc_macro/libproc_macro.a', needed by 'crab1'.  Stop.
```

Failing [action](https://github.com/Rust-GCC/gccrs/actions/runs/4869971300/jobs/8685132648)

P-E-P:
The bootstrap was failing due to a missing target which was not available during the bootstrap.

ChangeLog:

	* Makefile.def: Add libgrust target to bootstrap.
	* Makefile.in: Regenerate.

Fixes #2181

### DIFF
--- a/Makefile.def
+++ b/Makefile.def
@@ -147,7 +147,7 @@ host_modules= { module= libcc1; extra_configure_flags=--enable-shared; };
 host_modules= { module= gotools; };
 host_modules= { module= libctf; bootstrap=true; };
 host_modules= { module= libsframe; bootstrap=true; };
-host_modules= { module= libgrust; };
+host_modules= { module= libgrust; bootstrap=true; };
 
 target_modules = { module= libstdc++-v3;
 		   bootstrap=true;
--- a/Makefile.in
+++ b/Makefile.in
@@ -1299,7 +1299,9 @@ all-host: maybe-all-libctf
 @if libsframe-no-bootstrap
 all-host: maybe-all-libsframe
 @endif libsframe-no-bootstrap
+@if libgrust-no-bootstrap
 all-host: maybe-all-libgrust
+@endif libgrust-no-bootstrap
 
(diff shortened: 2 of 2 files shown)


P-E-P:
@CohenArthur This should be merged and then fixed up with  while upstreaming on https://github.com/Rust-GCC/gccrs/commit/afcde5f8175f294baf89141bcc24b8a6ba6499fd.

@tschwinge Is this the correct fix ? I found no documentation around `bootstrap`, I had to solely rely on the `makefile.def` in an attempt to understand what should be fixed and I may have missed something.

tschwinge:
Looking into this.

I wonder, don't we also need `dependencies = { module=all-gcc; on=all-libgrust; hard=true; };`?

P-E-P:
> I wonder, don't we also need `dependencies = { module=all-gcc; on=all-libgrust; hard=true; };`?

Won't this will compile libgrust even when the rust language is disabled ?

tschwinge:
This feels a bit weird what we're doing here...  ;-|

One step back.

So this relates to #2153 "Modularize libproc macro" commit afcde5f8175f294baf89141bcc24b8a6ba6499fd "libproc_macro: Build statically".  Does this `libgrust/libproc_macro` code need to be _(a)_ (statically) linked into the compiler only vs. also _(b)_ needs to be available in another way (target library)?  For linking into the compiler, should it simply be built only (case _(a)_) vs. additionally (case _(b)_) as part of the Rust front end build process?  (Even if the sources live in `libgrust/libproc_macro/` to cater for case _(b)_?

CohenArthur:
Merging this and then we'll have a look at #2242

PR #2185, opened 2023-05-05
Parse AST::AttrInputMacro attributes

powerboat9:
Allows parsing of ```#[doc = include_str!(...)]``` for ```libcore```. Depends on #2177

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -793,6 +793,21 @@ Parser<ManagedTokenSource>::parse_attr_input ()
 
 	t = lexer.peek_token ();
 
+	// attempt to parse macro
+	// TODO: macros may/may not be allowed in attributes
+	// this is needed for "#[doc = include_str!(...)]"
+	if (is_simple_path_segment (t->get_id ()))
+	  {
+	    std::unique_ptr<AST::MacroInvocation> invoke
+	      = parse_macro_invocation ({});
+
+	    if (!invoke)
+	      return nullptr;
+
+	    return std::unique_ptr<AST::AttrInput> (
+	      new AST::AttrInputMacro (std::move (invoke)));
+	  }
+
 	/* Ensure token is a "literal expression" (literally only a literal
 	 * token of any type) */
 	if (!t->is_literal ())
--- a/gcc/testsuite/rust/compile/doc_macro.rs
+++ b/gcc/testsuite/rust/compile/doc_macro.rs
@@ -1,0 +1,1 @@
+#![doc = concat!("AB")]


gcc/rust/parse/rust-parse-impl.h
```diff
@@ -793,6 +793,20 @@ Parser<ManagedTokenSource>::parse_attr_input ()
 
 	t = lexer.peek_token ();
 
+	// attempt to parse macro
+	// TODO: macros may/may not be allowed in attributes
+	// this is needed for "#[doc = !include_str(...)]"
```
CohenArthur, line 798:
```suggestion
	// this is needed for "#[doc = include_str!(...)]"
```

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -793,6 +793,20 @@ Parser<ManagedTokenSource>::parse_attr_input ()
...
+	if (is_simple_path_segment (t->get_id ()))
```
CohenArthur, line 799: we can have more complex macro invocations than single paths here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d81ca952b1610239de6a3955dd175728

```rust
#[macro_export]
macro_rules! foo {
    () => { "hello" }
}

#[doc = crate::foo!()]
struct S;
```

powerboat9: I can change it, but ```parse_macro_invocation``` currently only accepts simple paths

https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/parse/rust-parse-impl.h#L1845

powerboat9: Oh, wait, *single* paths

CohenArthur: Oh, actually this might be my mistake. The snippet I put in my comment is already parsed properly by your fix, so that's all good. I'm the one that got confused by "simple/single"!

CohenArthur (commented):
One thing I just realized is that this mechanic is present in modern Rust, but not in Rust 1.49 which we are targeting. So I'm not sure if we should be merging it right now or how to handle it exactly.

CohenArthur:
Oh actually this works with *eager* macro invocations like `concat!()`, but not general purpose macro invocations. So that should be all good

powerboat9:
Rebased to rely on the now-merged #2177

CohenArthur (approved):
So I'm happy with merging this. I'll open an issue to discuss how to handle the restrictions around eager macro invocations in this attribute, but what I think we should do is keep a list of allowed macros when expanding the invocation of a doc attribute.

PR #2200, opened 2023-05-09
Handle keywords in macro fragments

### LINKED ISSUE #2192: Parse ```self``` as ```$_:ident```
I tried this code:

```rust
macro_rules! foo { ($_:ident) => {} }

pub fn foo() {
    foo!(self);
}
```

Instead, this happened:

```
test.rs:4:5: error: Failed to match any rule within macro
    1 | macro_rules! foo { ($_:ident) => {} }
      | ~~~~~~~~~~~
......
    4 |     foo!(self);
      |     ^~~
```

### Meta

powerboat9:
Fixes #2192

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -999,6 +999,24 @@ Parser<ManagedTokenSource>::parse_delim_token_tree ()
     }
 }
 
+// Parses an identifier/keyword as a Token
+template <typename ManagedTokenSource>
+std::unique_ptr<AST::Token>
+Parser<ManagedTokenSource>::parse_identifier_or_keyword_token ()
+{
+  const_TokenPtr t = lexer.peek_token ();
+
+  if (t->get_id () == IDENTIFIER || token_id_is_keyword (t->get_id ()))
+    {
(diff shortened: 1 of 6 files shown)


gcc/rust/parse/rust-parse-impl.h
```diff
@@ -999,6 +999,89 @@ Parser<ManagedTokenSource>::parse_delim_token_tree ()
     }
 }
 
+// checks if a token is an identifier/keyword
+template <typename ManagedTokenSource>
+bool
+Parser<ManagedTokenSource>::is_ident_like_tk (TokenId id)
```
philberty, line 1005: I think is_keyword_token would be a better name

philberty (changes requested):
This looks good to me just a small nit with the naming of that helper. Other wise great job!

powerboat9:
Renamed to ```is_keyword_or_ident_token```since it also matches identifiers

CohenArthur (commented):
Looks good to me, just wondering about the public parsing function you introduced - we can name it `parse_ident_token` instead and it would be clearer IMO.

CohenArthur (approved):

PR #2204, opened 2023-05-10
libproc_macro: Change rust literal internals

### LINKED ISSUE #2182: Handle 128 bits literals in cpp side
Currently 128 bits literals are represented as a tuple `(u64, u64)` on the rust side since 128 bit ffi type has not been stabilized but those types were left unimplemented in the C++ side of the proc macro library.

- [ ] Implement 128 bits literals in libproc_macro c++
- [ ] Add 128 bits literals to token converter

### LINKED ISSUE #2201: libproc_macro: Rework literal type
Current type does not match the internal rust representation and this lead to multiple problems with the token conversion.

P-E-P:
Rust interface literal internals were taking a type and storing that type. This lead to multiple problems such as various conversion from string to int/float/other type as well as dead end on undetermined types (type checker runs at a later stage).

Fixes #2182 
Fixes #2201

### DIFF
--- a/libgrust/libproc_macro/ffistring.cc
+++ b/libgrust/libproc_macro/ffistring.cc
@@ -1,0 +1,62 @@
+// Copyright (C) 2023 Free Software Foundation, Inc.
+//
+// This file is part of the GNU Proc Macro Library.  This library is free
+// software; you can redistribute it and/or modify it under the
+// terms of the GNU General Public License as published by the
+// Free Software Foundation; either version 3, or (at your option)
+// any later version.
+
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// Under Section 7 of GPL version 3, you are granted additional
+// permissions described in the GCC Runtime Library Exception, version
+// 3.1, as published by the Free Software Foundation.
+
+// You should have received a copy of the GNU General Public License and
+// a copy of the GCC Runtime Library Exception along with this program;
+// see the files COPYING3 and COPYING.RUNTIME respectively.  If not, see
+// <http://www.gnu.org/licenses/>.
+
+#include <cstring>
+#include "ffistring.h"
+
+namespace ProcMacro {
+void
+FFIString::drop (FFIString *str)
+{
+  delete[] str->data;
+  str->len = 0;
+}
+
+FFIString
+FFIString::make_ffistring (const std::string &str)
+{
+  return make_ffistring (reinterpret_cast<const unsigned char *> (str.c_str ()),
+			 str.length ());
+}
+
+FFIString
+FFIString::make_ffistring (const unsigned char *data, std::uint64_t len)
+{
+  const unsigned char *inner = new unsigned char[len];
+  return {inner, len};
+}
+
+FFIString
+FFIString::clone () const
+{
+  unsigned char *inner = new unsigned char[this->len];
+  std::memcpy (inner, this->data, this->len);
+  return {inner, this->len};
+}
+
+std::string
+FFIString::to_string () const
+{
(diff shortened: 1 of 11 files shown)


libgrust/libproc_macro/ffistring.h
```diff
@@ -0,0 +1,55 @@
...
+{
+  const unsigned char *data;
+  std::uint64_t len;
+
+public:
+  FFIString clone () const;
+  std::string to_string () const;
+  static FFIString make_ffistring (const std::string &str);
+  static FFIString make_ffistring (const unsigned char *data,
+				   std::uint64_t len);
```
CohenArthur, line 41: Any reason for not having constructors here instead?
```suggestion
  FFIString (const std::string &str);
  FFIString (const unsigned char *data,
				   std::uint64_t len);
```

libgrust/libproc_macro/ffistring.cc
```diff
@@ -0,0 +1,62 @@
...
+FFIString::make_ffistring (const std::string &str)
+{
+  return make_ffistring (reinterpret_cast<const unsigned char *> (str.c_str ()),
+			 str.length ());
+}
+
+FFIString
+FFIString::make_ffistring (const unsigned char *data, std::uint64_t len)
+{
+  const unsigned char *inner = new unsigned char[len];
```
CohenArthur, line 44: Missing `std::memcpy` call here?

libgrust/libproc_macro/rust/bridge/ffistring.rs
```diff
@@ -0,0 +1,48 @@
...
+
+#[repr(C)]
+#[derive(Debug)]
+pub struct FFIString {
+    data: *const c_uchar,
+    len: u64,
+}
+
+impl FFIString {
+    pub fn new(string: &str) -> FFIString {
```
CohenArthur, line 20: I think `clippy` will warn here, won't it? `Self::new` should indicate the creation of an empty `FFIString`, so equivalent to what you have right now with `FFIString::new("")`, and we should instead implement `From<&str> for FFIString`? What do you think?

CohenArthur (commented):
Looks good to me :) a few comments but I'm happy with merging this

libgrust/libproc_macro/rust/bridge/ffistring.rs
P-E-P: If there would be any trait here to use that would probably be [`FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html) instead of `From<T>`. I'm not a fan of importing traits or any rust mechanism that is not strictly required in this particular crate as we want to compile it early on. Using more rust things means it may not compile because the compiler misses a bit somewhere that we do not have implemented yet. I mean,  this could be changed later once the compiler is working "better". What are your thoughts about this ?

libgrust/libproc_macro/ffistring.cc
P-E-P: Good catch.

libgrust/libproc_macro/ffistring.h
CohenArthur: Ah, makes sense! Thank you!

libgrust/libproc_macro/ffistring.cc
P-E-P: Fixed in [6792422](https://github.com/Rust-GCC/gccrs/pull/2204/commits/6792422d6fb893eeaf16698b5fee1ff520e26b76)

libgrust/libproc_macro/rust/bridge/ffistring.rs
P-E-P: We do use FromStr in this crate, thus I was enclined to change this code, but I then remembered it returned a Result. I therefore implemented the `From<S: AsRef<str>>` instead in [f5f749b](https://github.com/Rust-GCC/gccrs/pull/2204/commits/f5f749b1fb415ea23fdf7be559970fa04909b30f).

PR #2206, opened 2023-05-10
Block dot statements

### LINKED ISSUE #2189: Unsafe block as null denotation
I tried this code:

```rust
#[allow(unused_unsafe)]

pub fn foo() {
    unsafe { (0, 1) }.0;
}
```

Instead, this happened:

```
test.rs:4:22: error: found unexpected token ‘.’ in null denotation
    4 |     unsafe { (0, 1) }.0;
```

### Meta

matthewjasper:
`{}.f;`, `{}.a();`, `{}.await;` and `{}?;` are all syntactically valid expression statements in Rust. The current approach of have separate "with block" and "without block" statement expression kinds does not work well with this.

* Remove `AST::ExprStatementWithBlock` and `AST::ExprStatementWithoutBlock` - this distinction is currently unused and inspecting the inner expression can work as an alternative in the future.
* Move statement expression parsing to `parse_expr`, which now checks if it should return early for block expressions if `expr_can_be_stmt` is set.

Closes #2189


philberty (approved):
Nice job, this LGTM but I would like @CohenArthur to review before merging

P-E-P (approved):
This is really nice, furthermore this will greatly simplify the code on some aspect.

gcc/rust/ast/rust-expr.h
```diff
@@ -2720,17 +2721,12 @@ class RangeExpr : public ExprWithoutBlock
...
-    rust_assert (false);
+    outer_attrs = std::move (new_attrs);
   }
 
-  std::vector<Attribute> &get_outer_attrs () override
-  {
-    // RangeExpr cannot have any outer attributes
-    rust_assert (false);
-  }
+  std::vector<Attribute> &get_outer_attrs () override { return outer_attrs; }
```
CohenArthur, line 2729: I'm not sure about these changes - I don't think there's any way for `RangeExpr`s to have outer attributes at the moment

gcc/rust/ast/rust-ast.cc
```diff
@@ -1273,6 +1273,10 @@ ExprStmtWithBlock::as_string () const
     {
       indent_spaces (enter);
       str += expr->as_string ();
+      if (semicolon_followed)
+	{
+	  str += ";";
+	}
```
CohenArthur, line 1279: very small GNU style nitpick
```suggestion
	  str += ";";
```

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -7210,315 +7210,46 @@ Parser<ManagedTokenSource>::parse_method ()
...
-	  }
-      }
+      if (skip_token (SEMICOLON))
+	{
+	  has_semi = true;
+	}
+      else if (expr->is_expr_without_block ())
+	{
+	  return nullptr;
+	}
```
CohenArthur, line 7248: same here but feel free to ignore these nitpicks
```suggestion
      if (skip_token (SEMICOLON))
	  has_semi = true;
      else if (expr->is_expr_without_block ())
	  return nullptr;
```

CohenArthur (commented):
Those changes look great to me and they simplify parsing quite a lot. Thanks a lot and very good work!

gcc/rust/ast/rust-expr.h
P-E-P: Correct, see [the reference](https://doc.rust-lang.org/reference/expressions.html#expression-attributes).

matthewjasper: This only appears to apply to range expressions with a lower bound, rustc accepts the following
```rust
fn main() {
    #[cfg(all())]
    ..234;
}
``` 

That said, attribute parsing for expressions seems pretty complicated, so I've reverted the parsing to be as close as possible to the original handling before this patch.

CohenArthur: I wonder if that testcase passing is valid and not a bug - this is an expression, so we should definitely be hitting https://doc.rust-lang.org/stable/error_codes/E0658.html. But this is outside the scope of this PR :D

gcc/testsuite/rust/compile/macro_call_statement.rs
```diff
@@ -0,0 +1,11 @@
...
+
+macro_rules! panic {
+    () => (
+        crate::f()
+    );
+}
+
+fn main() {
+    panic!();
+}
```
CohenArthur, line 11: Is this test being part of this PR intentional?

CohenArthur (approved):
One last tiny question about the tests but that looks very good to me and I'm looking forward to merge it!

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -6200,13 +6207,6 @@ Parser<ManagedTokenSource>::parse_stmt (ParseRestrictions restrictions)
 	  return parse_vis_item (std::move (outer_attrs));
 	}
       break;
-    case SUPER:
-    case SELF:
-    case CRATE:
-    case DOLLAR_SIGN:
-      // almost certainly macro invocation semi
-      return parse_macro_invocation_semi (std::move (outer_attrs));
```
matthewjasper, line 6208: ^

gcc/testsuite/rust/compile/macro_call_statement.rs
matthewjasper: yes, it's fixed by the code change I just posted a comment pointing to.

gcc/rust/ast/rust-expr.h
CohenArthur: to clarify, we should be hitting `error[[E0658]](https://doc.rust-lang.org/stable/error_codes/E0658.html): attributes on expressions are experimental`

PR #2215, opened 2023-05-12
ast: Add `outer_attrs` to all `Item`s

CohenArthur:
gcc/rust/ChangeLog:

	* ast/rust-ast.h: Add `outer_attrs` to Node.
	* ast/rust-expr.h: Make use of new inheritance methods.
	* ast/rust-item.h: Likewise.
	* ast/rust-macro.h: Likewise.
	* ast/rust-path.h: Likewise.
	* ast/rust-stmt.h: Likewise.

This PR adds a vector of outer attributes to almost all nodes, which makes procedural macros easier to handle - there is no longer the need to `static_cast` to a node known to have outer attributes.

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -932,6 +932,11 @@ public:
   // behavior that we have items that can also be expressions?
   bool is_item () const override { return true; }
 
+  virtual std::vector<Attribute> &get_outer_attrs () = 0;
+  virtual const std::vector<Attribute> &get_outer_attrs () const = 0;
+
+  virtual bool has_outer_attrs () const { return !get_outer_attrs ().empty (); }
+
 protected:
   // Clone function implementation as pure virtual method
   virtual Item *clone_item_impl () const = 0;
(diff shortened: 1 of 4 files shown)


P-E-P (commented):
I'm not convinced this is the way we should go. Could you elaborate on your motivations ? What exactly was bothering you ? Was that for macro "identification" or macro "expansion" ?

CohenArthur:
> I'm not convinced this is the way we should go. Could you elaborate on your motivations ? What exactly was bothering you ? Was that for macro "identification" or macro "expansion" ?

that was for macro expansion: the issue is that `Item` does not implement `get_outer_attrs()` and `has_outer_attrs()`, which seems invalid. when looking at it in more detail, it seemed to me that basically all of our AST nodes should probably have a vector of outer attributes - with some restrictions around `RangeExpr` for example. But I'm not against implementing this "only" for `Item`s.

CohenArthur:
@P-E-P I reworked the PR entirely to only add these methods to the `Item` base class which makes more sense. Thanks for reporting this :)

PR #2219, opened 2023-05-16
Change return type of ```token_id_keyword_string```

powerboat9:
This should return a ```std::string``` instead of a ```const char *```


CohenArthur (approved):
LGTM

PR #2221, opened 2023-05-16
Handle keyword metavariables during macro expansion

### LINKED ISSUE #2194: Handle ```$type``` macro parameter
I tried this code:

```rust
macro_rules! foo {($type:ident) => {
    let $type = 12;
}} 

pub fn foo() {
    foo!(_a);
}
```

Instead, this happened:

```
test.rs:2:10: error: unexpected token in macro transcribe: expected ‘(’ or identifier after ‘$’, got ‘type’
    2 |     let $type = 12;
      |          ^~~~
```

### Meta

powerboat9:
Fixes #2194

### DIFF
--- a/gcc/rust/expand/rust-macro-substitute-ctx.cc
+++ b/gcc/rust/expand/rust-macro-substitute-ctx.cc
@@ -58,7 +58,8 @@ SubstituteCtx::check_repetition_amount (size_t pattern_start,
       if (macro.at (i)->get_id () == DOLLAR_SIGN)
 	{
 	  auto &frag_token = macro.at (i + 1);
-	  if (frag_token->get_id () == IDENTIFIER)
+	  if (token_id_is_keyword (frag_token->get_id ())
+	      || frag_token->get_id () == IDENTIFIER)
 	    {
 	      auto it = fragments.find (frag_token->get_str ());
 	      if (it == fragments.end ())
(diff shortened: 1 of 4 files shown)


philberty (changes requested):
Hmmm I think this LGTM but just wondering if get_str_handle_keyword should just replace the get_str function? I think the only reason get_str dodesnt work is jsut because any keyword has a null token but has a token id set and this fixes it. I wonder can you check that?

Otherwise i think the get_str and get_str_handle_keyword just seems confusing.

powerboat9:
Should be fixed now

gcc/rust/expand/rust-macro-substitute-ctx.cc
```diff
@@ -199,8 +200,13 @@ std::pair<std::vector<std::unique_ptr<AST::Token>>, size_t>
 SubstituteCtx::substitute_token (size_t token_idx)
 {
   auto &token = macro.at (token_idx);
+
+  if (token_id_is_keyword (token->get_id ()))
+    goto case_keyword;
```
philberty, line 205: do you need this goto? 

I wonder can you turn the case IDENTIFIER into the default case and leave the case LEFT_PAREN as is?

philberty:
Also thanks for fixed the get_str is looking brilliant

powerboat9:
```goto``` removed

CohenArthur (approved):
LGTM

PR #2222, opened 2023-05-16
Improve keyword check while parsing a macro match

powerboat9:
Use ```token_id_is_keyword``` to shorten a large list of cases

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -2063,57 +2063,6 @@ Parser<ManagedTokenSource>::parse_macro_match ()
 	const_TokenPtr t2 = lexer.peek_token (1);
 	switch (t2->get_id ())
 	  {
-	  case ABSTRACT:
-	  case AS:
-	  case ASYNC:
-	  case BECOME:
-	  case BOX:
-	  case BREAK:
-	  case CONST:
-	  case CONTINUE:
-	  case CRATE:
-	  case DO:
-	  case DYN:
-	  case ELSE:
-	  case ENUM_TOK:
-	  case EXTERN_TOK:
-	  case FALSE_LITERAL:
-	  case FINAL_TOK:
-	  case FN_TOK:
-	  case FOR:
-	  case IF:
-	  case IMPL:
-	  case IN:
-	  case LET:
-	  case LOOP:
-	  case MACRO:
-	  case MATCH_TOK:
-	  case MOD:
-	  case MOVE:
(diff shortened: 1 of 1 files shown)


powerboat9:
Also prevents parsing of ```crate``` as a valid macro match

powerboat9:
@philberty Looking through some of my old branches, parsing ```crate``` as a valid macro match is still an issue. I've refactored to remove the ```goto```, is it ok now?

powerboat9:
https://github.com/Rust-GCC/gccrs/compare/master...powerboat9:gccrs:fix-key

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -2045,59 +2045,13 @@ Parser<ManagedTokenSource>::parse_macro_match ()
...
-	  case TYPE:
-	  case TYPEOF:
-	  case UNSAFE:
-	  case UNSIZED:
-	  case USE:
-	  case VIRTUAL:
-	  case WHERE:
-	  case WHILE:
-	  case YIELD:
+	  case_keyword:
```
CohenArthur, line 2054: Jumping from outside the switch to inside the switch is not great IMO. I'm not in favor of this change. Do you have a testcase for the crate issue you mention? What about removing it from the list of cases in the switch statement?

powerboat9: Since this pull request was closed and hasn't been reopened, the ```Files changed``` menu hasn't updated.

https://github.com/Rust-GCC/gccrs/compare/master...powerboat9:gccrs:fix-key

CohenArthur: Ah, my bad - can you reopen a pull-request? This will be easier. The new diff looks good to me, I would just like to see testcases

powerboat9: I can't, I don't have the right permissions

CohenArthur: oh, that's weird. It tells me I can't either because the branch was force-pushed or recreated. Can you open another one or not at all?

powerboat9: I could, but does it work now? I force pushed the original commit

CohenArthur: the button just unlocked for me lol. I reopened this one. I assume it's not in a valid state then?

CohenArthur (approved):
Looks good to me, thank you for fixing it!

PR #2224, opened 2023-05-19
Add operator== for Optional.

powerboat9:


CohenArthur (changes requested):
I'm not a fan of operator overloading and we have very little of it in the compiler - what's your use case for it? Would you settle for an `equals` method or something?

powerboat9:
```std::unordered_set``` needs to be able to hash values and compare them for equality

CohenArthur (approved):

PR #2226, opened 2023-05-19
Fix macro resolutions in middle of multiple nested macro_use modules

### LINKED ISSUE #2209: Unknown macro with nested macro use modules
I tried this code (minimized from core, macro there is called `int_impl`):

```rust
#[macro_use]
mod num {
    #[macro_use]
    mod macros {
        macro_rules! a {
            () => ()
        }

    }

    a!();
}

fn main() -> i32 { 0 }
```

I expected to see this happen: Compiles

Instead, this happened: Error

```text
tmp/num/mod.rs:4:1: error: unknown macro: [a]
    4 | a!();
      | ^
```

### Meta

875cd5ee9c6f037c7f5ace00b75631a1c45970a5

matthewjasper:
Closes #2209

### DIFF
--- a/gcc/rust/resolve/rust-early-name-resolver.cc
+++ b/gcc/rust/resolve/rust-early-name-resolver.cc
@@ -63,8 +63,6 @@ EarlyNameResolver::accumulate_escaped_macros (AST::Module &module)
 	    continue;
 	  }
 
-	item->accept_vis (*this);
-
 	if (item->get_ast_kind () == AST::Kind::MACRO_RULES_DEFINITION)
 	  escaped_macros.emplace_back (item->clone_item ());
       }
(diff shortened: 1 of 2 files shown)


CohenArthur (commented):
This looks good to me - the one thing is, could the test call into `a!()` as well? This would ensure we don't introduce regressions down the line by accident. Great work otherwise :)

matthewjasper:
There's already an invocation of `a` in the test, do you want another one?

CohenArthur:
Ah yes sorry @matthewjasper I meant in the `main` function. Since you `#[macro_use]` both modules, it is available and in scopes. I've tested it locally on your branch and it already works - it's just to make sure

CohenArthur (approved):

Issue #2186, opened 2023-05-05
Fix ```libcore``` parsing


powerboat9:
I merged #2177 and #2131 into a local test branch and then attempted to compile ```libcore```. The parsing phase experienced the following errors:

- [ ] ```cannot find a built-in macro with name ‘panic’```
- [ ] ```cannot find a built-in macro with name ‘unreachable’```
- [ ] ```unknown macro: [int_impl]```
- [ ] ```unknown macro: [uint_impl]```
- [ ] ```unknown macro: [matches]```
- [ ] ```unknown macro: [assert]```
- [ ] ```cannot find a built-in macro with name ‘format_args’```
- [ ] ```cannot find a built-in macro with name ‘const_format_args’```
- [ ] ```cannot find a built-in macro with name ‘format_args_nl’```
- [ ] ```cannot find a built-in macro with name ‘option_env’```
- [ ] ```cannot find a built-in macro with name ‘concat_idents’```
- [ ] ```cannot find a built-in macro with name ‘concat_bytes’```
- [ ] ```cannot find a built-in macro with name ‘module_path’```
- [ ] ```cannot find a built-in macro with name ‘log_syntax’```
- [ ] ```cannot find a built-in macro with name ‘trace_macros’```
- [ ] ```cannot find a built-in macro with name ‘derive’```
- [ ] ```cannot find a built-in macro with name ‘derive_const’```
- [ ] ```cannot find a built-in macro with name ‘test’```
- [ ] ```cannot find a built-in macro with name ‘bench’```
- [ ] ```cannot find a built-in macro with name ‘test_case’```
- [ ] ```cannot find a built-in macro with name ‘global_allocator’```
- [ ] ```cannot find a built-in macro with name ‘alloc_error_handler’```
- [ ] ```cannot find a built-in macro with name ‘cfg_accessible’```
- [ ] ```cannot find a built-in macro with name ‘cfg_eval’```
- [ ] ```cannot find a built-in macro with name ‘type_ascribe’```
- [ ] ```cannot find a built-in macro with name ‘RustcDecodable’```
- [ ] ```cannot find a built-in macro with name ‘RustcEncodable’```

While a lot of these issues are duplicates of #927, I figured this issue would still be useful for tracking ```libcore``` progress.

powerboat9:
It looks like I accidentally shallow cloned the latest version of rust, as opposed to ```1.49.0```. The new list of errors is as follows:

[...]

PR #2227, opened 2023-05-20
Handle file loading from "non-mod-rs" files properly

powerboat9:
This seems to address many of the remaining ```unknown macro``` errors from #2186

https://doc.rust-lang.org/reference/items/modules.html

https://dev-doc.rust-lang.org/beta/edition-guide/rust-2018/module-system/path-clarity.html#no-more-modrs


CohenArthur (commented):
The changes seem okay to me, but they introduce a lot of test failures

philberty (commented):
This LGTM but not 100% sure i can see whats going on with the test cases, why did they need to be moved around like this?

P-E-P (approved):

gcc/rust/ast/rust-ast.cc
```diff
@@ -3310,21 +3310,40 @@ Module::process_file_path ()
...
+    {
+      including_fname = including_fpath.substr (dir_slash_pos + 1);
+    }
+
+  auto path_string = filename_from_path_attribute (get_outer_attrs ());
+
+  if (!path_string.empty () || !module_scope.empty ()
+      || including_fname == "mod.rs" || including_fname == "lib.rs"
+      || including_fname == "main.rs" || including_fname.size () < 3
+      || including_fname.compare (including_fname.size () - 3, 3, ".rs"))
```
CohenArthur, line 3340: this condition is very big - why is the size of `including_fname` a determining factor here? also, can this condition be moved in an outer static function?

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -2430,8 +2430,27 @@ Parser<ManagedTokenSource>::parse_module (AST::Visibility vis,
...
+	    auto slash_idx = filename.rfind (file_separator);
+	    if (slash_idx == std::string::npos)
+	      slash_idx = 0;
+	    else
+	      slash_idx++;
+	    filename = filename.substr (slash_idx);
+
+	    if (filename != "mod.rs" && filename != "lib.rs"
+		&& filename != "main.rs" && filename.size () >= 3
+		&& !filename.compare (filename.size () - 3, 3, ".rs"))
```
CohenArthur, line 2447: same question here

CohenArthur (approved):
That looks good to me - just the two comments to address and I'm okay with merging this. Thank you for working on this!

PR #2235, opened 2023-05-23
Bug fixes autoderef cycles using derefs

### LINKED ISSUE #2190: autoderef should support where clause `Deref` bounds
This assertion is wrong:

https://github.com/Rust-GCC/gccrs/blob/bb1e40da864addfc2d8dfc310928b7a4e7b13185/gcc/rust/typecheck/rust-autoderef.cc#L206-L209

This code causes the assertion to trigger: https://godbolt.org/z/b5s1Tj6vc

```rust
fn foo<T: Deref<Target = i32>>(t: &T) -> i32 {
    t.max(2)
}
```

Autoderef should be able to handle `Deref` coming from where clauses (and associated type bounds for that matter).

### LINKED ISSUE #2236: Type bounds fails to find Deref is implemented for i32
After fixing #2190 i was able to find our Type Bounds checking code is unable to realise the generic reference impl blocks do apply to i32 here. I think Rustc uses lang items here to say yes this impl block is applied to this type but i need to double check how they do this. We could reuse an autoderef cycle to match here as well for i32 vs &T from our coercion code.

I tried this code:

```rust
#[lang = "deref"]
trait Deref {
    type Target;
    fn deref(&self) -> &Self::Target;
}

impl<T> Deref for &T {
    type Target = T;

    fn deref(&self) -> &T {
        *self
    }
}

impl i32 {
    fn max(self, other: i32) -> i32 {
        if self > other {
            self
        } else {
            other
        }
    }
}

fn foo<T: Deref<Target = i32>>(t: &T) -> i32 {
    t.max(2)
}

fn main() -> i32 {
    let a: i32 = 1;
    foo::<i32>(&a) - 2
}
```

I expected to see this happen: Compile

Instead, this happened: 

```
test.rs:32:11: error: bounds not satisfied for i32 ‘Deref’ is not satisfied
   26 | fn foo<T: Deref<Target = i32>>(t: &T) -> i32 {
      |           ~~~~~
......
   32 |     foo::<i32>(&a) - 2
      |           ^~~
```

### Meta

- What version of Rust GCC were you using, git sha if possible. 586d85e4947d36852faac4c27229e1ba3205e270

philberty:
Fixes #2190 
Fixes #2236

### DIFF
--- a/gcc/rust/backend/rust-compile-base.cc
+++ b/gcc/rust/backend/rust-compile-base.cc
@@ -766,27 +766,28 @@ HIRCompileBase::named_constant_expression (tree type_tree,
 }
 
 tree
-HIRCompileBase::resolve_method_address (
-  TyTy::FnType *fntype, HirId ref, TyTy::BaseType *receiver,
-  const HIR::PathIdentSegment &segment,
-  const Analysis::NodeMapping &expr_mappings, Location expr_locus)
+HIRCompileBase::resolve_method_address (TyTy::FnType *fntype,
+					TyTy::BaseType *receiver,
+					Location expr_locus)
 {
+  DefId id = fntype->get_id ();
+  rust_assert (id != UNKNOWN_DEFID);
+
   // Now we can try and resolve the address since this might be a forward
   // declared function, generic function which has not be compiled yet or
   // its an not yet trait bound function
-  HIR::ImplItem *resolved_item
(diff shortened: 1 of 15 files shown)


CohenArthur (approved):

philberty:
I've added the fix for #2236 here to keep this issue together

Issue #2172, opened 2023-04-27
Tracking issue for builtin derive macros


CohenArthur:
- [ ] #2154 
- [ ] Add new `DeriveMacroAccumulator` visitor to visit all required fields/variants of a `struct`/`enum`/`union` getting "derived"

This is different from custom derive macros, which will only receive a token stream - we have more information within the compiler and can make use of them

- [ ] Handle all builtin derives (#927): `Clone`, `Copy`, `Debug`, `Default`, `Hash`, `{Partial}Eq`, `{Partial}Ord`)
- [ ] Add good typechecking errors for when derive constraints are not met:

```rust
struct S;
#[derive(Clone)]
struct SuperS(S);
```
should fail because `S` does not implement `Clone`. The code produced by the macro should be similar to this:

```rust
struct S;

struct SuperS(S);

impl Clone for SuperS {
    fn clone(&self) -> Self {
        Self(self.0.clone())
    }
}     
```

Without any derive-specific behavior, we'll get an error about how `self.0` does not implement clone, which is true but not helpful for users as the code responsible will have been generated by the compiler.

During typechecking, we should check that the members of a type getting derived also implement the trait being derived, and produce errors more related to deriving ("consider annotating `S` with `#[derive(Clone)]`" etc)

PR #2253, opened 2023-06-01
Add #[derive(Copy)] support

CohenArthur:
Addresses #2172 

From looking at `rustc`'s dump of a simple struct which is `#[derive(Copy)]`, it does not seem like we need to do anything particular like instantiating `AssertParamIsCopy<T>`s for each field's type. So I'm not sure how the "this field does not implement Copy" get generated and when. We'll have to investigate this further

### DIFF
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -90,6 +90,7 @@ GRS_OBJS = \
 	rust/rust-ast-builder.o \
 	rust/rust-derive.o \
 	rust/rust-derive-clone.o \
+	rust/rust-derive-copy.o \
     rust/rust-macro-invoc-lexer.o \
     rust/rust-macro-substitute-ctx.o \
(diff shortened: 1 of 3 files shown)


matthewjasper:
"this field does not implement Copy" is checked for all Copy implementations (when type checking the impl

philberty:
I think there is a piece about any assignment coercion site for non reference/ptr types technically turns into a copy call. So I think there is a piece about typechecking this then code generating it too.

Its kind of weird though because Rustc does take shortcuts here like for example the slices it does follow all the code for typechecking then ignores it all for codegen. So bet this will be another one of those cases where technically it should be code generating stuff to call copy but because simple primitive types don't require this they "optimize" it away super early before hitting LLVM

philberty (approved):
LGTM i llove how they are split into separate files readability win for us!

PR #2260, opened 2023-06-01
Make MatchedFragmentContainer a tree-like structure

powerboat9:
This should be useful for improving/introducing proper nested-repetition handling.

### DIFF
--- a/gcc/rust/expand/rust-macro-expand.cc
+++ b/gcc/rust/expand/rust-macro-expand.cc
@@ -78,7 +78,8 @@ MacroExpander::expand_decl_macro (Location invoc_locus,
 
   // find matching arm
   AST::MacroRule *matched_rule = nullptr;
-  std::map<std::string, MatchedFragmentContainer> matched_fragments;
+  std::map<std::string, std::unique_ptr<MatchedFragmentContainer>>
+    matched_fragments;
   for (auto &rule : rules_def.get_rules ())
     {
       sub_stack.push ();
@@ -109,8 +110,13 @@ MacroExpander::expand_decl_macro (Location invoc_locus,
       return AST::Fragment::create_error ();
     }
 
-  return transcribe_rule (*matched_rule, invoc_token_tree, matched_fragments,
-			  semicolon, peek_context ());
+  std::map<std::string, MatchedFragmentContainer *> matched_fragments_ptr;
+
+  for (auto &ent : matched_fragments)
+    matched_fragments_ptr.emplace (ent.first, ent.second.get ());
+
+  return transcribe_rule (*matched_rule, invoc_token_tree,
+			  matched_fragments_ptr, semicolon, peek_context ());
 }
 
 void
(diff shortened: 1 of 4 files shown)


gcc/rust/expand/rust-macro-expand.h
```diff
@@ -90,66 +90,88 @@ class MatchedFragmentContainer
...
    * Add a matched fragment to the container
    */
-  void add_fragment (MatchedFragment fragment)
-  {
-    rust_assert (!is_single_fragment ());
+  void add_fragment (MatchedFragment fragment);
 
-    fragments.emplace_back (fragment);
-  }
+  // const std::string &get_fragment_name () const { return fragment_name; }
```
CohenArthur, line 114: dead code?

gcc/rust/expand/rust-macro-expand.cc
```diff
@@ -109,8 +110,13 @@ MacroExpander::expand_decl_macro (Location invoc_locus,
       return AST::Fragment::create_error ();
     }
 
-  return transcribe_rule (*matched_rule, invoc_token_tree, matched_fragments,
-			  semicolon, peek_context ());
+  std::map<std::string, MatchedFragmentContainer *> matched_fragments_ptr;
+
+  for (auto &ent : matched_fragments)
+    matched_fragments_ptr.emplace (ent.first, ent.second.get ());
```
CohenArthur, line 116: Is there no way to avoid this and instead make changes to `transcribe_rule` and `SubstituteCtx` to accept an `std::map<std::string, std::unique_ptr<MatchedFragmentContainer>>`?

CohenArthur (commented):
I think this PR is a good idea, thank you for the work!

gcc/rust/expand/rust-macro-expand.h
powerboat9: The diff algorithm produces a strange result here, but that line was moved and not created -- there's a matching ```- // const std::string &get_fragment_name () const { return fragment_name; }``` line in the diff.

gcc/rust/expand/rust-macro-expand.cc
powerboat9: Not without some finagling. Each repetition substitution would have to either clone ```MatchedFragmentContainer```s or temporarily move them out of shared state.

gcc/rust/expand/rust-macro-expand.h
CohenArthur: ah, I see, thanks :) that's fine then, or you can remove it in this PR if you'd like since you're cleaning up the class

gcc/rust/expand/rust-macro-expand.h
```diff
@@ -180,8 +202,9 @@ class SubstitutionScope
     auto it = current_map.find (fragment.fragment_ident);
 
     if (it == current_map.end ())
-      current_map.insert ({fragment.fragment_ident,
-			   MatchedFragmentContainer::metavar (fragment)});
+      current_map.emplace (fragment.fragment_ident,
+			   std::move (
+			     MatchedFragmentContainer::metavar (fragment)));
```
CohenArthur, line 207:
```suggestion
			     MatchedFragmentContainer::metavar (fragment));
```
oh, that `move` is unnecessary and causes clang to emit a warning - I'll merge the PR once this is removed

CohenArthur (changes requested):

gcc/rust/expand/rust-macro-expand.h
powerboat9: This should be fixed, just waiting for the tests to run

CohenArthur (approved):
Thank you!

PR #2261, opened 2023-06-01
Rework the HIR dump pass

### LINKED ISSUE #693: HIR Dump
Our HIR dump looks like this currently:

```
BlockExpr:
{
 outer attributes: none
 inner attributes: none
 statements: 
 ExprStmtWithoutBlock:
  MethodCallExpr: 
 Object (receiver) expr: ( t ([C: 0 Nid: 97 Hid: 77]))
 Method path segment: 
baz
 Call params:none
 final expression: none
}::[C: 0 Nid: 102 Hid: 81 Lid: 26]
```

Which is very messy and is a clone of our AST as_string methods which is fine for gdb sessions but bad for textural output.

- [ ] Converge on an S-expression format, which is more extensible and easier to read
- [x] Add command line switch https://github.com/Rust-GCC/gccrs/pull/222
- [x] Add dump entry point https://github.com/Rust-GCC/gccrs/blob/28f527c9598339cf834a30b5ee1f14258b8ecbb2/gcc/rust/rust-session-manager.cc#L874-L888
- [x] Dump to file 
- [ ] Print block statements
- [ ] Print block tail expression if present
- [ ] Print crate mapping attributes
- [ ] Print function parameters
- [ ] Visit function's definition block

dkm:
Nearly complete rewrite of the HIR dump pass.
    
fixes #693

### DIFF
--- a/gcc/rust/hir/rust-hir-dump.cc
+++ b/gcc/rust/hir/rust-hir-dump.cc
@@ -17,678 +17,2411 @@
 // <http://www.gnu.org/licenses/>.
 
 #include "rust-hir-dump.h"
+#include "rust-abi.h"
+#include "rust-hir-item.h"
+#include "rust-hir-path.h"
+#include "rust-hir-type.h"
+#include "rust-hir.h"
+#include <string>
 
 namespace Rust {
 namespace HIR {
 
-Dump::Dump (std::ostream &stream) : stream (stream) {}
+// Dump Format for HIR
+//
+// SomeHIRNode [
+//    field: ...
+//    field: ...
+//    field: ...
+// ]
+//
+// When a field is a collection of other HIR object:
+//   field {
+//     SomeHIRNode [ ... ]
+//     SomeOtherHIRNode [ ... ]
+//   }
+//
+// If a field is an empty collection:
+//  field: empty
+//
+// If a field is optional and is currently not holding anything:
+//   field: none
+
+std::string Dump::delims[2][2] = {
+  {std::string ("{"), std::string ("}")},
+  {std::string ("["), std::string ("]")},
+};
 
 void
-Dump::go (HIR::Crate &crate)
+Dump::go (HIR::Crate &e)
 {
-  stream << "Crate {" << std::endl;
-  // inner attributes
-  if (!crate.get_inner_attrs ().empty ())
+  begin ("Crate");
+  do_inner_attrs (e);
+  do_mappings (e.get_mappings ());
+
+  visit_collection ("items", e.get_items ());
+  end ("Crate");
+}
+
+Dump::Dump (std::ostream &stream) : stream (stream) {}
+
+/**
+ * Writes TEXT with a final newline if ENDLINE is true.
+ * If TEXT is starting the current line, its first line is indented.
+ * If TEXT is multiline, all followning lines are also indented.
+ *
+ * @param text Text to emit
+ * @param endline If true, newline is emitted after text
+ */
+void
+Dump::put (std::string text, bool endline)
+{
+  if (beg_of_line)
     {
-      indentation.increment ();
       stream << indentation;
-      stream << "inner_attrs: [";
-      for (auto &attr : crate.get_inner_attrs ())
-	stream << attr.as_string ();
-      stream << "]," << std::endl;
-      indentation.decrement ();
+      beg_of_line = false;
     }
 
-  indentation.increment ();
-  stream << indentation;
-  //
+  // keep multiline string indented
+  std::string::size_type pos = 0;
+  std::string::size_type prev = 0;
+  auto first = true;
+  while ((pos = text.find ('\n', prev)) != std::string::npos)
+    {
+      if (!first)
+	stream << std::endl << indentation;
+      first = false;
 
-  stream << "items: [";
+      stream << text.substr (prev, pos - prev);
+      prev = pos + 1;
+    }
 
-  stream << indentation;
-  for (const auto &item : crate.get_items ())
+  if (first)
+    stream << text;
+
+  if (endline)
     {
       stream << std::endl;
-      item->accept_vis (*this);
+      beg_of_line = endline;
     }
-  stream << std::endl;
-  stream << indentation;
-
-  stream << "]," << std::endl;
-  indentation.decrement ();
-  //
-
-  indentation.increment ();
-  stream << indentation;
-  stream << "node_mappings: ";
-  stream << crate.get_mappings ().as_string ();
-  indentation.decrement ();
-
-  stream << "\n}" << std::endl;
 }
+
+/**
+ * Called when starting to emit info for an HIR node.
+ * Emits NAME and an opening delimiter denoted by D.
+ *
+ * @param name Name of HIR node
+ * @param d Delimiter
+ */
+void
+Dump::begin (std::string name, enum delim d)
+{
+  if (!beg_of_line)
+    put ("");
+  put (name + " " + delims[d][0], true);
+  indentation.increment ();
+}
+
+/**
+ * Called when ending the dump info for an HIR node. Emits a C++-style
+ * comment with NAME and a closing delimiter denoted by D.
+ *
+ * @param name Name of HIR node
+ * @param d Delimiter
+ */
+void
+Dump::end (std::string name, enum delim d)
+{
+  indentation.decrement ();
+  if (!beg_of_line)
+    stream << std::endl;
+  put (delims[d][1] + " // " + name);
+}
+
+/**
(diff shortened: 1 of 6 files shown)


dkm:
Initial work for reworking the HIR dump (easier indentation handling, less formatting work needed,...). Originally motivated by some debugging. Currently simply trying to have something easier to use while keeping the actual output format.
Sample bellow:
```

Crate [
    
    items [
        
        Function {
            func_name: size,
            return_type: usize,
            
            BlockExpr [
                final expression: ( 15 ([C: 0 Nid: 5 Hid: 22]))

            ] // BlockExpr
            node_mappings: [C: 0 Nid: 7 Hid: 24 Lid: 2]

        } // Function
        
        Function {
            func_name: main,
            return_type: void,
            
            BlockExpr [
                
                Stmt [
                    
                    LetStmt {
                        a
                        Expr {
                            
                            ArrayExpr (INCOMPLETE -- CONTENT NOT DISPLAYED) [
                                ArrayExpr:
                                 inner attributes: none

                            ] // ArrayExpr

                        } // Expr

                    } // LetStmt

                ] // Stmt

            ] // BlockExpr
            node_mappings: [C: 0 Nid: 23 Hid: 34 Lid: 3]
...
```

One big improvement is that the dump now clearly shows when it dumps something not yet handled :clown_face:

CohenArthur:
That output looks great :D I like it! I also like the "unimplemented" bit - that's good info to have

dkm:
> That output looks great :D I like it! I also like the "unimplemented" bit - that's good info to have

I still think s-expr would be nicer and easier to parse/browse... But as a first step, I'll try to simply have the content.

gcc/rust/hir/tree/rust-hir-expr.h
```diff
@@ -24,6 +24,7 @@
 #include "rust-hir.h"
 #include "rust-hir-path.h"
 #include "rust-operators.h"
+#include <memory>
```
philberty, line 27: I have a funny feeling this should be included by rust-system.h I think @tschwinge found I was including it and it was marked as poisoned or something in a bootstrap build.

gcc/rust/hir/tree/rust-hir-expr.h
```diff
@@ -1105,8 +1108,8 @@ class ArrayIndexExpr : public ExprWithoutBlock
   void accept_vis (HIRFullVisitor &vis) override;
   void accept_vis (HIRExpressionVisitor &vis) override;
 
-  Expr *get_array_expr () { return array_expr.get (); }
-  Expr *get_index_expr () { return index_expr.get (); }
+  std::unique_ptr<Expr> &get_array_expr () { return array_expr; }
```
philberty, line 1111: You are amazing this has been bugging me that the HIR is super inconsistent like that and returning raw pointers. Thanks for doing this

gcc/rust/hir/tree/rust-hir-expr.h
```diff
@@ -533,8 +539,10 @@ class LazyBooleanExpr : public OperatorExpr
   void accept_vis (HIRFullVisitor &vis) override;
   void accept_vis (HIRExpressionVisitor &vis) override;
 
-  Expr *get_lhs () { return main_or_left_expr.get (); }
+  void visit_lhs (HIRFullVisitor &vis) { main_or_left_expr->accept_vis (vis); }
+  void visit_rhs (HIRFullVisitor &vis) { right_expr->accept_vis (vis); }
 
+  Expr *get_lhs () { return main_or_left_expr.get (); }
```
philberty, line 545: I think this is another case where we should be returning std::unique_ptr

gcc/rust/hir/tree/rust-hir-expr.h
```diff
@@ -459,6 +462,9 @@ class ComparisonExpr : public OperatorExpr
   void accept_vis (HIRFullVisitor &vis) override;
   void accept_vis (HIRExpressionVisitor &vis) override;
 
+  void visit_lhs (HIRFullVisitor &vis) { main_or_left_expr->accept_vis (vis); }
+  void visit_rhs (HIRFullVisitor &vis) { right_expr->accept_vis (vis); }
+
   Expr *get_lhs () { return main_or_left_expr.get (); }
   Expr *get_rhs () { return right_expr.get (); }
```
philberty, line 469: same here

philberty (commented):
Great work cant wait to get it merged. Are you still going for s-expressions I have seen dumps in rustc which looked like the AST and another which looked like s-expressions. I don't really care but s-expressions might be useful so we can dump mapping info which would be interesting and eventually extend this pass to dump the type check info too

gcc/rust/hir/tree/rust-hir-expr.h
tschwinge: Yes, instead please (early) `#define INCLUDE_MEMORY`, see <https://github.com/Rust-GCC/gccrs/blob/2a274c1b7bfaa860f41b4bdcfb7008df0bc4cc26/gcc/system.h#L742-L748>.

dkm: Hmmm, clangd did some nasty stuff behind my back. It automatically adds include files that break the build. I did not add any header file intentionally, so this one is probably the work of clangd.

gcc/rust/hir/tree/rust-hir-expr.h
dkm: I was not really sure what to do, but I've noticed the inconsistencies. I can fix them along the way, that's easy and helps me understand the HIR anyway :)

gcc/rust/hir/tree/rust-hir-expr.h
dkm: ack

dkm:
> Great work cant wait to get it merged. Are you still going for s-expressions I have seen dumps in rustc which looked like the AST and another which looked like s-expressions. I don't really care but s-expressions might be useful so we can dump mapping info which would be interesting and eventually extend this pass to dump the type check info too

As often for this kind of "this will be quick" work, my initial plan was to keep current format (a rustc-like HIR dump) and have it quickly working. When the first pass is ok-ish, it /should/ not be too hard to move to s-expr. I've tried to use util function with start/stop call, so it should not be that hard to use s-expr. But as a first step, this will be an unspecified rustc-ish HIR dump.

dkm:
So, I think it should be ready to be merged.
Eyeballing the results of some test cases seems ok.
I've executed the full testsuite with the extra `-frust-dump-hir-pretty` and there's no more crash due to assertions or anything else.

I still need to fixup the commit log and add some inline doc/comment, but it should be ready for an initial review (only look at `-dump.{h,cc}` files, the others changes come from #2405 that should go first).

dkm:
I've added few comments, amended the changelog, fixed few minor leftovers... So hopefully this is fine now.

dkm:
Oooops, there are regression after latest rebase, I'll need to investigate before this can be merged.

dkm:
Ok, regressions are fixed, good to go, again.

gcc/rust/hir/rust-hir-dump.cc
```diff
@@ -17,678 +17,2457 @@
...
+    case UseTreeRebind::NewBindType::IDENTIFIER:
+      return "IDENTIFIER";
+    case UseTreeRebind::NewBindType::WILDCARD:
+      return "WILDCARD";
+    }
+  gcc_unreachable ();
+}
+
+std::string
+convert_mut_to_str (Mutability mut)
```
philberty, line 92: potentially you could move this over to util/rust-common.h i think and leave it as an inline helper

gcc/rust/hir/rust-hir-dump.cc
```diff
@@ -17,678 +17,2457 @@
...
     {
       stream << std::endl;
-      item->accept_vis (*this);
+      beg_of_line = endline;
     }
-  stream << std::endl;
-  stream << indentation;
+}
+
+static std::string delims[2][2] = {
```
philberty, line 150: i think you should keep this along side the Delim enum as they are associated

philberty (approved):
LGTM lets get this merged only two minor style things that come to mind.

I think we migt have some changes when we get it merged and start using it but i want to see it first

philberty:
Feel free to merge when your ready

gcc/rust/hir/rust-hir-dump.cc
dkm: ack

gcc/rust/hir/rust-hir-dump.cc
dkm: I've implemented this change, and renamed the function at the same time to `enum_to_str`, a bit more explicit.

PR #2265, opened 2023-06-03
ast: dump literals correctly

### LINKED ISSUE #2220: AST Dump does not print literals properly
I tried this code:

```rust
fn main() {
    let c = 'c';
    let b = b'y';
    let s = "jinko";
}
```

I expected to see this happen: The AST dump prints the literal properly, adding the extra `'` and `"` where necessary

Instead, this happened: This is the AST dump

```rust
fn main() {
	let c = c;
	let b = y;
	let s = jinko;
}
```

### Meta

- What version of Rust GCC were you using e02b4427f87be6d5f433e1965d36b3f197fca90f

### Guide

To fix this issue, have a look at the AST dump implementation in `rust-ast-dump.cc`. Specifically, the function you'll need to fix is the following:

https://github.com/Rust-GCC/gccrs/blob/e02b4427f87be6d5f433e1965d36b3f197fca90f/gcc/rust/ast/rust-ast-dump.cc#L508-L511

You can `switch` on the expr's `get_lit_type()` in order to handle these special literals properly

charmitro:
This commit fixes printing of literals based on their type.

Previous implementation printed literals the same, regardless their type. Now we are printing:
  * int, float, bool don't require special printing
  * char        -> `'<char>'`
  * string      -> `"<string>"`
  * byte        -> `b'<byte>'`
  * byte_string -> `b"<byte_string>"`

Fixes #2220

### DIFF
--- a/gcc/rust/ast/rust-ast-dump.cc
+++ b/gcc/rust/ast/rust-ast-dump.cc
@@ -507,7 +507,34 @@ Dump::visit (QualifiedPathInType &path)
 void
 Dump::visit (LiteralExpr &expr)
 {
-  stream << expr.as_string ();
+  switch (expr.get_lit_type ())
+    {
+    case Literal::CHAR:
+      stream << "'" << expr.as_string () << "'";
+      return;
+
+    case Literal::STRING:
+      stream << "\"" << expr.as_string () << "\"";
+      return;
+
+    case Literal::BYTE:
+      stream << "b'" << expr.as_string () << "'";
+      return;
+
+    case Literal::BYTE_STRING:
+      stream << "b\"" << expr.as_string () << "\"";
+      return;
+
+    case Literal::INT:
+    case Literal::FLOAT:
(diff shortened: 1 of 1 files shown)


gcc/rust/ast/rust-ast-dump.cc
```diff
@@ -507,7 +507,40 @@ Dump::visit (QualifiedPathInType &path)
...
+      stream << expr.as_string ();
+      return;
+
+    case Literal::FLOAT:
+      stream << expr.as_string ();
+      return;
+
+    case Literal::BOOL:
+      stream << expr.as_string ();
+      return;
```
CohenArthur, line 538:
```suggestion
    case Literal::INT:
    case Literal::FLOAT:
    case Literal::BOOL:
      stream << expr.as_string ();
      return;
```

nit :)

CohenArthur (approved):
LGTM! Thank you :D

Issue #1078, opened 2022-03-31
Handle `$crate` properly in macros


CohenArthur:
A special "metavar" exists in macro invocations, which helps reference items from the module in macro invocations. We must implement its handling properly for the Imports and Visibility milestone as it is a core part of exporting macros to the rest of a crate

CohenArthur:
When a macro is exported, the #[macro_export] attribute can have the local_inner_macros keyword added to automatically prefix all contained macro invocations with $crate::. This is intended primarily as a tool to migrate code written before $crate was added to the language to work with Rust 2018's path-based imports of macros. Its use is discouraged in new code.

```rust
#[macro_export(local_inner_macros)]
macro_rules! helped {
    () => { helper!() } // Automatically converted to $crate::helper!().
}

#[macro_export]
macro_rules! helper {
    () => { () }
}
```

PR #2267, opened 2023-06-04
Handle crate metavariable

powerboat9:
Addresses #1078 for the time being

### DIFF
--- a/gcc/rust/expand/rust-macro-substitute-ctx.cc
+++ b/gcc/rust/expand/rust-macro-substitute-ctx.cc
@@ -31,7 +31,13 @@ SubstituteCtx::substitute_metavar (
   if (it == fragments.end ())
     {
       // fail to substitute
-      return false;
+
+      // HACK: substitute ($ crate) => (crate)
+      if (metavar->get_id () != CRATE)
+	return false;
+
+      expanded.push_back (metavar->clone_token ());
+      return true;
     }
   else
     {


gcc/rust/expand/rust-macro-substitute-ctx.cc
```diff
@@ -31,7 +31,13 @@ SubstituteCtx::substitute_metavar (
...
     {
       // fail to substitute
-      return false;
+
+      // HACK: substitute ($ crate) => (crate)
+      if (metavar->get_id () != CRATE)
+	return false;
+
+      expanded.push_back (metavar->clone_token ());
+      return true;
```
CohenArthur, line 40: I'm not sure this is enough to get the proper behavior of `$crate`, and we will run into issues when importing the `core` crate if I'm not mistaken. For example, when expanding [`panic!()`](https://doc.rust-lang.org/src/std/macros.rs.html#14-20), we need to make sure we're calling into `core`'s `panic_2015!` or `panic_2018!` macro and not a similarly named macro defined by the user in the crate where `panic!` is being *called*. Unless I am misunderstanding?

powerboat9: Yeah, ```$crate``` would need to be handled better for macro exporting purposes. As long as a macro isn't exported, ```$crate``` => ```crate``` should still work. I figured a partial fix to get ```libcore``` parsing would still be useful, and of course this shouldn't resolve #1078.

PR #2274, opened 2023-06-07
Proc macro statements

P-E-P:
Requires #2218

### DIFF
--- a/gcc/rust/ast/rust-ast.cc
+++ b/gcc/rust/ast/rust-ast.cc
@@ -4228,8 +4228,10 @@ BlockExpr::normalize_tail_expr ()
   if (!expr)
     {
       // HACK: try to turn the last statement into a tail expression
-      if (statements.size () && statements.back ()->is_expr ())
+      if (!statements.empty () && statements.back ()->is_expr ())
 	{
+	  // Watch out: This reference become invalid when the vector is
+	  // modified.
 	  auto &stmt = static_cast<ExprStmt &> (*statements.back ());
 
 	  if (!stmt.is_semicolon_followed ())
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -884,6 +884,15 @@ class MetaListNameValueStr;
 class Stmt : public Node
 {
 public:
+  enum class Kind
+  {
+    Empty,
+    Item,
+    Let,
+    Expr,
+    MacroInvocation,
+  };
+
   // Unique pointer custom clone function
   std::unique_ptr<Stmt> clone_stmt () const
   {
@@ -900,8 +909,10 @@ public:
   virtual bool is_marked_for_strip () const = 0;
   NodeId get_node_id () const { return node_id; }
 
-  virtual bool is_item () const = 0;
+  virtual Kind get_stmt_kind () = 0;
 
+  // TODO: Can we remove these two?
+  virtual bool is_item () const = 0;
   virtual bool is_expr () const { return false; }
 
   virtual void add_semicolon () {}
@@ -930,6 +941,8 @@ public:
   virtual void
   add_crate_name (std::vector<std::string> &names ATTRIBUTE_UNUSED) const
   {}
+
+  Stmt::Kind get_stmt_kind () final { return Stmt::Kind::Item; }
 
   // FIXME: ARTHUR: Is it okay to have removed that final? Is it *required*
   // behavior that we have items that can also be expressions?
--- a/gcc/rust/ast/rust-expr.h
+++ b/gcc/rust/ast/rust-expr.h
@@ -2503,6 +2503,8 @@ public:
   // Normalizes a trailing statement without a semicolon to a tail expression.
   void normalize_tail_expr ();
 
+  void try_convert_last_stmt ();
+
   const std::vector<Attribute> &get_outer_attrs () const { return outer_attrs; }
   std::vector<Attribute> &get_outer_attrs () override { return outer_attrs; }
 
--- a/gcc/rust/ast/rust-expr.h
+++ b/gcc/rust/ast/rust-expr.h
@@ -2487,6 +2487,17 @@ public:
     return expr;
   }
 
+  std::unique_ptr<Expr> take_tail_expr ()
+  {
+    rust_assert (has_tail_expr ());
+    return std::move (expr);
+  }
+
+  void set_tail_expr (std::unique_ptr<Expr> expr)
+  {
+    this->expr = std::move (expr);
+  }
+
   // Removes the tail expression from the block.
   void strip_tail_expr () { expr = nullptr; }
   // Normalizes a trailing statement without a semicolon to a tail expression.
--- a/gcc/rust/expand/rust-expand-visitor.cc
+++ b/gcc/rust/expand/rust-expand-visitor.cc
@@ -149,9 +149,9 @@ expand_item_attribute (AST::Item &item, AST::SimplePath &name,
   return result;
 }
 
+template <typename T>
 static std::vector<std::unique_ptr<AST::Stmt>>
-expand_stmt_attribute (AST::Item &item, AST::SimplePath &name,
-		       MacroExpander &expander)
+expand_stmt_attribute (T &item, AST::SimplePath &name, MacroExpander &expander)
 {
   std::vector<std::unique_ptr<AST::Stmt>> result;
   auto frag = expander.expand_attribute_proc_macro (item, name);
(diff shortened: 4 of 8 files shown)


gcc/rust/ast/rust-expr.h
```diff
@@ -2487,9 +2487,22 @@ class BlockExpr : public ExprWithBlock
     return expr;
   }
 
+  std::unique_ptr<Expr> take_tail_expr ()
+  {
+    rust_assert (has_tail_expr ());
+    return std::move (expr);
+  }
+
+  void set_tail_expr (std::unique_ptr<Expr> expr)
```
CohenArthur, line 2496:
```suggestion
  void set_tail_expr (std::unique_ptr<Expr> &&expr)
```

gcc/rust/expand/rust-expand-visitor.cc
```diff
@@ -230,37 +287,85 @@ ExpandVisitor::expand_inner_items (
 
 void
 ExpandVisitor::expand_inner_stmts (
-  std::vector<std::unique_ptr<AST::Stmt>> &stmts)
+  AST::BlockExpr &expr, std::vector<std::unique_ptr<AST::Stmt>> &stmts)
```
CohenArthur, line 290:
```suggestion
  AST::BlockExpr &expr)
```
we can probably get away by just passing the `expr`, right? And the accessing `expr.get_stmts()`

gcc/rust/expand/rust-expand-visitor.cc
```diff
@@ -230,37 +287,85 @@ ExpandVisitor::expand_inner_items (
...
+				 std::inserter (stmts, it));
+		      // TODO: Improve this ?
+		      // item is invalid since it refers to now deleted,
+		      // cancel the loop increment and break.
+		      it--;
+		      break;
+		    }
+		}
+	    }
+	}
```
CohenArthur, line 363: so this is more of an observation than me requesting a change - is that code very similar to item proc macro expansion, and if yes, do you think we could factor them into one templated function? or are there some big differences I'm not seeing?

gcc/rust/expand/rust-expand-visitor.cc
```diff
@@ -730,12 +835,15 @@ ExpandVisitor::visit (AST::ClosureExprInner &expr)
 void
 ExpandVisitor::visit (AST::BlockExpr &expr)
 {
-  expand_inner_stmts (expr.get_statements ());
+  expand_inner_stmts (expr, expr.get_statements ());
```
CohenArthur, line 838:
```suggestion
  expand_inner_stmts (expr);
```

gcc/rust/expand/rust-expand-visitor.cc
```diff
@@ -730,12 +835,15 @@ ExpandVisitor::visit (AST::ClosureExprInner &expr)
...
 
   expander.push_context (MacroExpander::ContextType::BLOCK);
 
+  expand_tail_expr (expr, expander);
   if (expr.has_tail_expr ())
-    maybe_expand_expr (expr.get_tail_expr ());
+    {
+      maybe_expand_expr (expr.get_tail_expr ());
+    }
```
CohenArthur, line 846:
```suggestion
      maybe_expand_expr (expr.get_tail_expr ());
```

gcc/rust/expand/rust-expand-visitor.h
```diff
@@ -158,7 +158,8 @@ class ExpandVisitor : public AST::ASTVisitor
    * @param values Vector of values to mutate in-place and append into
    */
   void expand_inner_items (std::vector<std::unique_ptr<AST::Item>> &values);
-  void expand_inner_stmts (std::vector<std::unique_ptr<AST::Stmt>> &values);
+  void expand_inner_stmts (AST::BlockExpr &expr,
+			   std::vector<std::unique_ptr<AST::Stmt>> &values);
```
CohenArthur, line 162:
```suggestion
  void expand_inner_stmts (AST::BlockExpr &expr);
```

CohenArthur (commented):
That's super nice work, and I really, really like the "hack" for `BlockExpr`s haha. It's very clever! Good job!

gcc/rust/expand/rust-expand-visitor.cc
P-E-P: Yes it is indeed similar, and I did consider to factor those two functions together multiple times. But there are some small differences in multiple places that make it not so easy to refactor. I'm not yet convinced the result of such refactor would make the code easier and cleaner, quite the opposite in fact. I could experiment and compare both versions in a later PR, I do not want to make this PR bigger/harder than required.

gcc/rust/expand/rust-expand-visitor.cc
P-E-P: Fixed in e145c1eb94d76af6dee7f1ec0bc5f7c92f1d2779

gcc/rust/expand/rust-expand-visitor.cc
P-E-P: Fixed in e145c1eb94d76af6dee7f1ec0bc5f7c92f1d2779

gcc/rust/expand/rust-expand-visitor.cc
P-E-P: Fixed in e145c1eb94d76af6dee7f1ec0bc5f7c92f1d2779

gcc/rust/expand/rust-expand-visitor.h
P-E-P: Fixed in e145c1eb94d76af6dee7f1ec0bc5f7c92f1d2779

philberty (commented):
There is a PR https://github.com/Rust-GCC/gccrs/pull/2282 that might affect this PR lets see what happens when it gets merged

gcc/rust/ast/rust-ast.cc
```diff
@@ -4223,26 +4223,32 @@ Attribute::is_parsed_to_meta_item () const
 }
 
 void
-BlockExpr::strip_tail_expr ()
+BlockExpr::try_convert_last_stmt ()
 {
-  if (expr)
+  rust_assert (expr == nullptr);
+  // HACK: try to turn the last statement into a tail expression
+  if (statements.size () && statements.back ()->is_expr ())
```
philberty, line 4230: I think doing
```
if (!statements.empty() && statements.back ()->is_expr ()) {
}
```

I prefer using empty instead of a size check is more readable

gcc/rust/ast/rust-ast.cc
```diff
@@ -4223,26 +4223,32 @@ Attribute::is_parsed_to_meta_item () const
...
     {
-      expr = nullptr;
+      auto &stmt = static_cast<ExprStmt &> (*statements.back ());
```
philberty, line 4232: Just note it is undefined behaviour to take a reference to a vector element then mutate the vector after then use the same reference again.

You are ok here because you don't use stmt after calling statements.pop_back but if you did it would turn into garbage. I had a horrible issue in the type system where I had a reference to the current context then called query_type which might then change the current context by resolving another function which turned my reference into garbage this was really confusing to track down and it was fixed simply by removing the reference and taking a copy.

Nothing to change here but something to be aware of.

gcc/rust/ast/rust-ast.h
```diff
@@ -884,6 +884,15 @@ class MetaListNameValueStr;
 class Stmt : public Node
 {
 public:
+  enum class Kind
+  {
```
philberty, line 888: Brilliant more of this please :D. Everyone hates visitors and more of this will allow use to use switch statements and write a function something like the GCC walk_tree eventually.

gcc/rust/expand/rust-expand-visitor.cc
```diff
@@ -149,6 +149,63 @@ expand_attribute (std::unique_ptr<AST::Item> &item, AST::SimplePath &name,
   return result;
 }
 
+template <typename T>
+static std::vector<std::unique_ptr<AST::Stmt>>
+expand_stmt_attribute (T &item, AST::SimplePath &name, MacroExpander &expander)
```
philberty, line 154: I feel like this function needs a comment with a small example its not clear what types of T will be valid here because C++ is not as good as Rust lol

philberty (commented):
just some small minor comments but lets see if the other PR's being merged cause issues for you

gcc/rust/expand/rust-expand-visitor.cc
P-E-P: Fixed in [26fe2f9](https://github.com/Rust-GCC/gccrs/pull/2274/commits/26fe2f9120795cb8b53a53c39f82904570cde8d9)

gcc/rust/ast/rust-ast.cc
P-E-P: Fixed in [26fe2f9](https://github.com/Rust-GCC/gccrs/pull/2274/commits/26fe2f9120795cb8b53a53c39f82904570cde8d9).

CohenArthur (approved):
Thank you! Great work :)

PR #2283, opened 2023-06-09
Fix ICE for reference patterns in match statements

dme2:
Fixes https://github.com/Rust-GCC/gccrs/issues/1813.

Not sure that I handled that scrutinee_kind == TyTy::TypeKind::REF 100% correctly in the visit function, so I'll make this a draft.


philberty (approved):
LGTM i think this looks perfectly reasonable for this case. you just need to fix your commit message

Issue #418, opened 2021-05-10
Unicode identifiers


bjorn3:
Unicode identifiers were stabilized in https://github.com/rust-lang/rust/pull/83799. There is a PR (https://github.com/rust-lang/reference/pull/1022) on the reference to clarify the rules regarding unicode identifiers. There are also some lints to detect confusing cases like mixing different scripts.

tamaroning:
Duplicate #2287

PR #2284, opened 2023-06-09
Tokenize Unicode identifiers

tamaroning:
Addresses #2287 #418

- [x] add function which checks XID_Start and XID_Continue to libcpp
- [x] identifiers
- [x] lifetime token

gcc/rust/ChangeLog:
```
	* lex/rust-lex.cc (is_whitespace): add all lacked codepoints valid as whitespaces
	(is_identifier_start): new function to check XID_Start and underscore
	(is_identifier_continue): new function to check XID_Continue
	(Lexer::build_token):tokenize Unicode identifiers
	(Lexer::parse_partial_string_continue):add comments
	(Lexer::parse_partial_unicode_escape):add comments
	(Lexer::parse_raw_identifier):change to use `is_identifier_scontinue`
	(Lexer::parse_identifier_or_keyword):change to use `is_identifier_continue`
	(Lexer::parse_char_or_lifetime):change to use `is_identifier_start` and `is_identifier_continue`
	(Lexer::skip_codepoint_input):do not attempt to skip input when bumping EOF
	* lex/rust-lex.h:add `current_char32` field
```
libcpp/ChangeLog:
```
	* charset.cc (check_xid_property):new function to check XID_Start and XID_Continue
	* include/cpplib.h (check_xid_property):add forward declaration of `check_xid_property`
```

### DIFF
--- a/gcc/rust/lex/rust-lex.cc
+++ b/gcc/rust/lex/rust-lex.cc
@@ -22,6 +22,7 @@
 #include "rust-linemap.h"
 #include "rust-session-manager.h"
 #include "safe-ctype.h"
+#include "cpplib.h"
 
 namespace Rust {
 // TODO: move to separate compilation unit?
@@ -103,17 +104,35 @@ check_valid_float_dot_end (char character)
   return character != '.' && character != '_' && !ISALPHA (character);
 }
 
-// ISSPACE from safe-ctype but may change in future
 bool
-is_whitespace (char character)
+is_whitespace (int character)
 {
-  return ISSPACE (character);
+  // https://doc.rust-lang.org/reference/whitespace.html
+  return character == '\t' || character == '\n' || character == '\v'
+	 || character == '\f' || character == '\r' || character == ' '
+	 || character == 0x0085	 // next line
+	 || character == 0x200e	 // left-to-right mark
+	 || character == 0x200f	 // right-to-left mark
+	 || character == 0x2028	 // line separator
+	 || character == 0x2029; // pragraph separator
 }
 
 bool
 is_non_decimal_int_literal_separator (char character)
 {
   return character == 'x' || character == 'o' || character == 'b';
+}
+
+bool
+is_identifier_start (int codepoint)
+{
+  return (check_xid_property (codepoint) & XID_START) || codepoint == '_';
+}
+
+bool
+is_identifier_continue (int codepoint)
+{
+  return check_xid_property (codepoint) & XID_CONTINUE;
 }
 
 Lexer::Lexer (const std::string &input)
@@ -283,21 +302,21 @@ Lexer::build_token ()
   while (true)
     {
       Location loc = get_current_location ();
-      current_char = peek_input ();
-      skip_input ();
 
       // detect UTF8 bom
       //
       // Must be the first thing on the first line.
       // There might be an optional BOM (Byte Order Mark), which for UTF-8 is
       // the three bytes 0xEF, 0xBB and 0xBF. These can simply be skipped.
-      if (current_line == 1 && current_column == 1 && current_char == 0xef
-	  && peek_input () == 0xbb && peek_input (1) == 0xbf)
+      if (current_line == 1 && current_column == 1 && peek_input () == 0xef
+	  && peek_input (1) == 0xbb && peek_input (2) == 0xbf)
 	{
-	  skip_input (1);
-	  current_char = peek_input ();
-	  skip_input ();
+	  skip_input (2);
 	}
+
+      current_char = peek_input ();
+      current_char32 = peek_codepoint_input ();
+      skip_codepoint_input ();
 
       // detect shebang
       // Must be the first thing on the first line, starting with #!
@@ -311,6 +330,7 @@ Lexer::build_token ()
 	  int n = 1;
 	  while (true)
 	    {
+	      // TODO use utf-8 codepoint to skip whitespaces
 	      int next_char = peek_input (n);
 	      if (is_whitespace (next_char))
 		n++;
@@ -1051,7 +1071,8 @@ Lexer::build_token ()
 	  int peek = peek_input ();
 	  int peek1 = peek_input (1);
 
-	  if (peek == '#' && (ISALPHA (peek1) || peek1 == '_'))
+	  // TODO (tamaron) parse Unicode ident
+	  if (peek == '#' && is_identifier_start (peek1))
 	    {
 	      TokenPtr raw_ident_ptr = parse_raw_identifier (loc);
 	      if (raw_ident_ptr != nullptr)
@@ -1068,8 +1089,8 @@ Lexer::build_token ()
 	    }
 	}
 
-      // find identifiers and keywords
-      if (ISALPHA (current_char) || current_char == '_')
+      // find identifiers and keywords.
+      if (is_identifier_start (current_char32.value))
 	return parse_identifier_or_keyword (loc);
 
       // int and float literals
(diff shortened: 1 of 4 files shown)


tamaroning:
I found this code and am not sure how we handle width of characters 
https://github.com/tamaroning/gccrs/blob/3c0f3ca4e4d8ddeecead05e109ace1917e021321/gcc/rust/lex/rust-lex.cc#L2462-L2463

libcpp/charset.cc
```diff
@@ -1256,6 +1256,38 @@ _cpp_uname2c_uax44_lm2 (const char *name, size_t len, char *canon_name)
...
 
+/* Returns 1 if C has the XID_Start property, 2 if C has the XID_Continue properties, 3 if C has the both properties, 0 otherwise. */
+int check_xid_property (cppchar_t c) {
+  // fast path for ASCII
+  if (c < 0x80) {
+    if (('A' <= c && c <= 'Z') || ('a' <= c && c <= 'z'))
+      return 3;
+    if (('0' <= c && c <= '9') || c == '_')
+      return 2;
+  }
```
tamaroning, line 1267: I think this function should return bit flags but how can I do it?
(i.e. XID_Start: 1, XID_Continue: 1 << 1)
Is just using `int` (or `unsigned int`) and export constants such as `XID_START` and `XID_CONTINUE` good?

tamaroning: Fix to use `unsigned int` and export consants via enum

gcc/rust/lex/rust-lex.cc
```diff
@@ -103,11 +104,17 @@ check_valid_float_dot_end (char character)
...
 {
-  return ISSPACE (character);
+  // https://doc.rust-lang.org/reference/whitespace.html
+  return character == '\t' || character == '\n' || character == '\v'
+	 || character == '\f' || character == '\r' || character == ' '
+	 || character == 0x0085	 // next line
+	 || character == 0x200e	 // left-to-right mark
+	 || character == 0x200f	 // right-to-left mark
+	 || character == 0x2028	 // line separator
+	 || character == 0x2029; // pragraph separator
```
CohenArthur, line 117: Are all of those characters accepted by `rustc` as whitespace?

gcc/rust/lex/rust-lex.cc
```diff
@@ -283,22 +302,22 @@ Lexer::build_token ()
...
 	{
-	  skip_input (1);
-	  current_char = peek_input ();
-	  skip_input ();
+	  skip_input (2);
 	}
 
+      current_char = peek_input ();
+      current_char32 = peek_codepoint_input ();
+      skip_codepoint_input ();
```
CohenArthur, line 319: why are we skipping the codepoint input here but not the char?

gcc/rust/lex/rust-lex.cc
```diff
@@ -2443,28 +2467,29 @@ Lexer::parse_char_or_lifetime (Location loc)
...
+	      current_char32 = peek_codepoint_input ();
 	      length++;
 	    }
 
 	  current_column += length;
 
 	  loc += length - 1;
 
+	  // TODO some keywords cannot be used for a lifetime label
+	  // https://doc.rust-lang.org/reference/tokens.html
```
CohenArthur, line 2491: That's good - please open an issue for this and put the issue number in the comment :)

CohenArthur (commented):
That's great work, thank you!

gcc/rust/lex/rust-lex.cc
tamaroning: If we skip one byte here, only the first byte of current utf-8 character can be skipped by the lexer, which we do not expect. 
For example, if the lexer tokenizes identifier あああ , it should skip the first utf8 character あ, not its first byte.

gcc/rust/lex/rust-lex.cc
tamaroning: Yes. All of these values are defined in the Rust ref.
You can find URL to this just before the selected lines.

gcc/rust/lex/rust-lex.cc
tamaroning: Opend https://github.com/Rust-GCC/gccrs/issues/2306

tamaroning:
Oops, I mistakenly made the last commit. I will undo it later.

tamaroning:
Done

gcc/rust/lex/rust-lex.cc
CohenArthur: Oh, I missed this! Sorry! Thanks for pointing it out haha

gcc/rust/lex/rust-lex.cc
CohenArthur: I see, thank you!

gcc/rust/lex/rust-lex.cc
```diff
@@ -2443,28 +2467,29 @@ Lexer::parse_char_or_lifetime (Location loc)
...
+	      skip_codepoint_input ();
+	      current_char32 = peek_codepoint_input ();
 	      length++;
 	    }
 
 	  current_column += length;
 
 	  loc += length - 1;
 
+	  // TODO some keywords cannot be used for a lifetime label
```
CohenArthur, line 2490:
```suggestion
	  // TODO some keywords cannot be used for a lifetime label #2306 
```

CohenArthur (approved):
Very minor nit :) This looks great to me, thank you @tamaroning!

CohenArthur:
@tamaroning I noticed that your commit does not include a commit title - please add a short description of the commit as title. It should look something like this:

```
    gccrs: we can't check the bounds involving empty placeholder types

    We use placeholders for assoicated types on traits but if we are unifying
    types against a placeholder its not possible to check the bounds as the
    placeholder does not have enough information yet at this point to determine
    if bounds will or won't be satisfied. That check will occur when associated
    types and generics are setup.

    Fixes #2036

    gcc/rust/ChangeLog:

            * typecheck/rust-unify.cc (UnifyRules::go): dont check bounds on placeholders

    gcc/testsuite/ChangeLog:

            * rust/compile/issue-2036.rs: New test.

    Signed-off-by: Philip Herron <herron.philip@googlemail.com>
```

You can see the commit has a title, which can be anything as long as it is a short description of the commit. Then, you have the commit description - but that's not always necessary. Then the Changelog, and finally the `Signed-off-by` line

tamaroning:
@CohenArthur  Thanks for your review! I fixed comments and commit messages.

philberty (changes requested):
This patch looks great though i think what will really help is to split this into two commits. So you put all the libcpp changes into one commit and the gccrs changes in another. The only reason for this is that the libcpp changes will need to be reviewed by other people in the GCC community and it will be easier to do this by separating them out.

Great work!

tamaroning:
> This patch looks great though i think what will really help is to split this into two commits. So you put all the libcpp changes into one commit and the gccrs changes in another. The only reason for this is that the libcpp changes will need to be reviewed by other people in the GCC community and it will be easier to do this by separating them out.

Sounds good :) Done.

P-E-P (approved):
I've not much to say, that's good work.

gcc/rust/lex/rust-lex.cc
```diff
@@ -116,6 +123,18 @@ is_non_decimal_int_literal_separator (char character)
   return character == 'x' || character == 'o' || character == 'b';
 }
 
+bool
+is_identifier_start (int codepoint)
```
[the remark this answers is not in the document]
tamaroning: Yes, I think `uint32_t` is better. If we unify types for paramters of such functions, other several functions should also use the same type.
e.g. is_x_digit, is_octal_digit, etc.
https://github.com/Rust-GCC/gccrs/blob/d535c8259301687e28105105a52e013f138b619e/gcc/rust/lex/rust-lex.cc#L83-L87

PR #2288, opened 2023-06-12
fixup: Fix build dependency

P-E-P:
Fix the missing dependency between gcc and libgrust.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -698,6 +698,17 @@ if test -d ${srcdir}/libphobos; then
     fi
 fi
 
+# Disable libgrust when not using the rust language
+if test -d ${srcdir}/libgrust; then
+  case ,${enable_languages}, in
+    *,rust,*)
+      ;;
+    *)
+      noconfigdirs="$noconfigdirs libgrust"
+      ;;
+  esac
+fi
+
 # Disable Fortran for some systems.
 case "${target}" in
   mmix-*-*)
--- a/Makefile.def
+++ b/Makefile.def
@@ -357,6 +357,7 @@ dependencies = { module=configure-gcc; on=all-ld; };
 dependencies = { module=configure-gcc; on=all-gold; };
 dependencies = { module=configure-gcc; on=all-libiconv; };
 dependencies = { module=all-gcc; on=all-libiberty; hard=true; };
+dependencies = { module=all-gcc; on=all-libgrust; };
 dependencies = { module=all-gcc; on=all-intl; };
 dependencies = { module=all-gcc; on=all-mpfr; };
 dependencies = { module=all-gcc; on=all-mpc; };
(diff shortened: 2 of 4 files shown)


P-E-P:
Add a build dependency as suggested [here](https://github.com/Rust-GCC/gccrs/pull/2183#issuecomment-1534540420) because the build on a single process was not working anymore.

Note that I did not set up a "hard" build dependency because GCC documentation advise against it in most cases. If someone thinks this is a mistake feel free to speak up.

CohenArthur (approved):
LGTM!

philberty (approved):
LGTM i dont see anyother way to do this. I think we are the only front-end to need this so far. The only problem i see when liibgrust starts to depend on Rust do we then have a dependency on Rust at the start? It would be ideal if we can rely on the GCC build and make Rust depend on C++ to achieve this for bootstrapping.

I say we could go for this patch but maybe start a thread on GCC mailing list asking about this.

P-E-P:
> The only problem i see when liibgrust starts to depend on Rust do we then have a dependency on Rust at the start? It would be ideal if we can rely on the GCC build and make Rust depend on C++ to achieve this for bootstrapping.

I may have misunderstood your message but this will be the case, the user app will depend on a rust proc_macro library which depends on cpp proc_macro library. GCC will continue to depend on the cpp proc_macro library.

The cpp proc macro library will probably be renamed in the future to avoid a name collision with the rust one.

configure.ac
```diff
@@ -698,6 +698,17 @@ if test -d ${srcdir}/libphobos; then
     fi
 fi
 
+# Disable libgrust when not using the rust language
+if test -d ${srcdir}/libgrust; then
```
tschwinge, line 702: This `if` isn't actually necessary here, as you're not attempting to use files from `${srcdir}/libgrust/`.  (Background is that the top-level build system is shared between GCC and Binutils/GDB, and `${srcdir}/libgrust/` doesn't exist for the latter, for example.)  However, as such usage is going to appear at some point in time (like "Disable libphobos on unsupported systems" just above, for example), it's OK to leave in that `if`.

tschwinge (approved):
Looks good to me to unblock things.

configure.ac
P-E-P: TBH i wasn't sure why it was put everywhere then I read that some binutils build share that script and thought I better put a seamlessly useless condition rather than breaking an obscure bintutils build somewhere. I'm not yet sure if we want to disable libgrust at some point though.

tschwinge: > I'm not yet sure if we want to disable libgrust at some point though. Or is this only disabled for binutils/GDB ?

It'll have to be disabled for all configurations where it doesn't build (or, rather: only be enabled for configurations where it is supported/known to build).  See <https://github.com/Rust-GCC/gccrs/blob/master/libsanitizer/configure.tgt> as a random example.  Similar to the libsanitizer case, I suppose that for libgrust we'll also have that decided that for each "sub-library" individually: the generic Procedural Macros support library is always built (for it doesn't contain any configuration specifics, supposedly), but, for example, the Rust Standard Library (imported from upstream Rust) only for configurations where it's known to build.

PR #2290, opened 2023-06-12
Remove direct error on macro rule count-related match failure

### LINKED ISSUE #2229: Macros error on match count mismatch too eagerly
I tried this code (minimized from core, https://github.com/rust-lang/rust/blob/e1884a8e3c3e813aada8254edfa120e85bf5ffca/library/core/src/ptr/mod.rs#L1468):

```rust
macro_rules! fnptr_impls_args {
    ($($Arg: ident),+) => {};
    () => {};
}

fnptr_impls_args! {}
fnptr_impls_args! { A }
```

Instead, this happened: 

```text
<source>:86:21: error: invalid amount of matches for macro invocation. Expected between 1 and +inf, got 0
   86 |     ($($Arg: ident),+) => {};
      |
```

### Meta

gccrs (Compiler-Explorer-Build-gcc-c7b7e297ea4b688ef9fb51a8aee2a8b2af39ac1c-binutils-2.40) 13.0.1 20230417 (experimental)

powerboat9:
Fixes #2229


gcc/testsuite/rust/compile/macro6.rs
CohenArthur, line 1: This change is invalid: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ce3350e1796ae54aa9343bea6c67dc25

The macro invocation should still fail - similarly for the other test file, `macro7.rs`

gcc/rust/expand/rust-macro-expand.cc
```diff
@@ -694,13 +694,6 @@ MacroExpander::match_repetition (Parser<MacroInvocLexer> &parser,
       gcc_unreachable ();
     }
 
-  if (!res)
-    rust_error_at (rep.get_match_locus (),
-		   "invalid amount of matches for macro invocation. Expected "
-		   "between %s and %s, got %lu",
-		   lo_str.c_str (), hi_str.c_str (),
-		   (unsigned long) match_amount);
-
```
CohenArthur, line 703: The problem is that we are emitting the error *too* eagerly when another match arm could match the invocation, not that the error is emitted. We cannot just remove the error emission - we need to delay it until all match arms have been tried

CohenArthur (changes requested):

gcc/rust/expand/rust-macro-expand.cc
powerboat9: It does still fail though -- note the ```// { dg-error "Failed to match any rule within macro" }```. While the error is less descriptive, it is similar to the error in the linked rust playground code (```no rules expected the token '12'```) anyway.

PR #2282, opened 2023-06-08
Fix parsing of statement macros 

matthewjasper:
Fixes various bugs in statement macro handling
- Parses braced macros as statements where appropriate
- Checks for semicolons (sometimes) when expanding statement macros
- Handles final statement vs. tail expression changing for macros
- Distinguishes between expression and statement macros in AST

This PR takes an approach similar to rustc (and gccrs before #2206) where statement macros are parsed as statements and the parser "recovers" into parsing an expression when required.  I didn't use the existing block expr vs. non-block expr handling because it doesn't handle match arms correctly (or interpolated expression macros once those are implemented).

Closes #2239

PR #2292, opened 2023-06-13
Ignore semicolon following a macro expansion in expression context

### LINKED ISSUE #2273: Macro Expansion Producing Expression with Semicolon
I tried this code:

```rust
macro_rules! mac {
    () => {return 1;}
}

pub fn foo() -> i32 {
    mac!()
}
```

```
warning: trailing semicolon in macro used in expression position
 --> src/lib.rs:2:20
  |
2 |     () => {return 1;}
  |                    ^
...
6 |     mac!()
  |     ------ in this macro invocation
  |
  = note: for more information, [see issue #79813 <https://github.com/rust-lang/rust/issues/79813>](https://github.com/rust-lang/rust/issues/79813)
  = note: macro invocations at the end of a block are treated as expressions
  = note: to ignore the value produced by the macro, add a semicolon after the invocation of `mac`
  = note: `#[warn(semicolon_in_expressions_from_macros)]` on by default
  = note: this warning originates in the macro `mac` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Instead, this happened:
```
test.rs:2:20: error: tokens here and after are unparsed
    2 |     () => {return 1;}
      |                    ^
```

### Meta

- git sha: ```1b322540fa0fce561cf86e371dd7f304f1343511```

powerboat9:
Fixes #2273

### DIFF
--- a/gcc/rust/expand/rust-macro-expand.cc
+++ b/gcc/rust/expand/rust-macro-expand.cc
@@ -911,6 +911,15 @@ transcribe_expression (Parser<MacroInvocLexer> &parser)
   if (expr == nullptr)
     return AST::Fragment::create_error ();
 
+  // FIXME: make this an error for some edititons
+  if (parser.peek_current_token ()->get_id () == SEMICOLON)
+    {
+      rust_warning_at (
+	parser.peek_current_token ()->get_locus (), 0,
+	"trailing semicolon in macro used in expression context");
+      parser.skip_token ();
+    }
+
   auto end = lexer.get_offs ();
 
   return AST::Fragment ({std::move (expr)}, lexer.get_token_slice (start, end));
--- a/gcc/testsuite/rust/compile/macro-issue2273.rs
+++ b/gcc/testsuite/rust/compile/macro-issue2273.rs
@@ -1,0 +1,7 @@
+macro_rules! mac {
+    () => {();} // { dg-warning "trailing semicolon" }
+}
+
+pub fn foo() {
+    mac!()
+}


powerboat9:
Looks like I accidentally duplicated some of #2282 -- converting this into a draft for reexamination after #2282 is merged.

powerboat9:
It looks like I actually didn't duplicate #2282, as issue #2273 is still present on master.

gcc/rust/expand/rust-macro-expand.cc
```diff
@@ -875,6 +875,15 @@ transcribe_expression (Parser<MacroInvocLexer> &parser)
...
 
+  // FIXME: make this an error for some edititons
+  if (parser.peek_current_token ()->get_id () == SEMICOLON)
+    {
+      rust_warning_at (
+	parser.peek_current_token ()->get_locus (), 0,
+	"trailing semicolon in macro used in expression context");
+      parser.skip_token ();
+    }
+
```
matthewjasper, line 886: The FIXME at line 1050 in this file can be removed.

powerboat9: But this will have to be changed to an error in future editions

CohenArthur (approved):
Looks good :) nice work

gcc/rust/expand/rust-macro-expand.cc
CohenArthur: We'll keep the existing code, but emit an error or a warning depending on which `-frust-edition` option has been passed. Both behaviors need to stay. So I agree that the comment can be removed, or it can be kept for now since we'll need to touch that part of the code again

philberty (approved):
LGTM

PR #2300, opened 2023-06-14
Add proc macro resolve mappings

### LINKED ISSUE #2410: Remove NodeId from `Identifier` and replace Identifiers in attributes with `SimplePath` instead
> > I don't think we need a node_id for an Identifier as we do resolution to the NodeId of the parent item the identifier belongs to.
> > I think remove the node_id, or add a copy constructor and copy assignment constructor. We also wish to have location info but feel free to do this as a separate PR see: #2333
> 
> But I do need a node Id for procedural macro resolution. When I have such derive attribute:
> 
> ```
> #[derive(Toto, serde::Deserialize)]
> ```
> 
> `serde::Deserialize` is a `SimplePath` but `Toto` is an identifier. They both share the same parent attribute with a common node id even though that's two different macros.

I feel like I've seen a lot of discussion around having `NodeId`s for `Identifier`s and the more I think about it, the more it seems like this makes sense *only* for proc macros. In this case, wouldn't it be better to have a different class specifically for these proc macro identifiers? E.g instead of having
```c++
class Identifier {
    std::string identifier;
    NodeId id;
    Location loc;
}
```
and ignore the `NodeId` in all cases *but* proc macro name resolution, why not have
```c++
class Identifier {
    std::string identifier;
    Location loc;
}

class SpecialMagicProcMacroIdentifier {
    Identifier ident;
    NodeId node;
}
```
?

@P-E-P @philberty not sure what you think of this?

_Originally posted by @CohenArthur in https://github.com/Rust-GCC/gccrs/issues/2332#issuecomment-1627724200_
            
- [x] Convert `Identifier` to `SimplePath`
- [x] Remove common path interface (`ProcMacroInvocable`)
- [x] #2458

### LINKED ISSUE #2450: Segfault on outer `crate_type` attribute
gccrs segfaults, when using `#[crate_type]` as an outer attribute. This attribute doesn't do anything as an outer attribute, but gccrs should emit a warning instead of segfaulting.

I didn't try this with other attributes, so not sure if it is `crate_type` specific.

### Code

```rust
#[crate_type = "bin"]
fn main() {}
```

https://godbolt.org/z/n4G4TE8Wj

### Meta

```
gccrs (Compiler-Explorer-Build-gcc-235dbf0b572a51ed19cc313ed19e696b5eb51566-binutils-2.40) 13.0.1 20230417 (experimental)
```

### Error output

```
0x2388b4e internal_error(char const*, ...)
	???:0
0xab9d15 void std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_construct<char*>(char*, char*, std::forward_iterator_tag)
	???:0
0xc7b10b Rust::AST::Fragment Rust::MacroExpander::expand_attribute_proc_macro<Rust::AST::Item>(Rust::AST::Item&, Rust::AST::SimplePath&)
	???:0
0xc772b6 Rust::ExpandVisitor::expand_inner_items(std::vector<std::unique_ptr<Rust::AST::Item, std::default_delete<Rust::AST::Item> >, std::allocator<std::unique_ptr<Rust::AST::Item, std::default_delete<Rust::AST::Item> > > >&)
	???:0
0xbaa936 Rust::Session::expansion(Rust::AST::Crate&)
	???:0
0xbae603 Rust::Session::compile_crate(char const*)
	???:0
0xbaebaf Rust::Session::handle_input_files(int, char const**)
	???:0
Compiler returned: 1
```

P-E-P:
This PR does **NOT** name resolve proc macros properly, as stated initially since we're changing our name resolver.  Instead it brings proc macro mappings that will be used with the new name resolver.

Fixes #2410 
Fixes #2450


gcc/rust/ast/rust-ast.h
```diff
@@ -44,8 +45,11 @@ class Identifier
   Identifier &operator= (const Identifier &) = default;
   Identifier &operator= (Identifier &&) = default;
 
-  NodeId get_node_id () const { return node_id; }
-  const std::string &as_string () const { return ident; }
+  NodeId get_node_id () const override { return node_id; }
+
+  const std::string as_string () const override { return ident; }
+
+  const Location get_locus () const override { return Location (); }
```
P-E-P, line 52: This will change once Identifier location is merged.

P-E-P: Merged and rebased, therefore fixed.

gcc/rust/ast/rust-ast.h
```diff
@@ -36,36 +36,28 @@ class Identifier
...
 
   Identifier (const Identifier &) = default;
   Identifier (Identifier &&) = default;
   Identifier &operator= (const Identifier &) = default;
   Identifier &operator= (Identifier &&) = default;
 
-  NodeId get_node_id () const { return node_id; }
   location_t get_locus () const { return loc; }
-  const std::string &as_string () const { return ident; }
+  const std::string as_string () const { return ident; }
```
CohenArthur, line 55: why change to returning a copy of the string here?

gcc/rust/metadata/rust-imports.cc
```diff
@@ -160,21 +161,26 @@ Import::try_package_in_directory (const std::string &filename,
 
       fd = Import::try_suffixes (&found_filename);
       if (fd < 0)
-	return NULL;
+	{
+	  return std::make_pair (nullptr, std::vector<ProcMacro::Procmacro>{});
+	}
```
CohenArthur, line 166:
```suggestion
	  return std::make_pair (nullptr, std::vector<ProcMacro::Procmacro>{});
```

gcc/rust/rust-session-manager.cc
```diff
@@ -971,16 +971,32 @@ Session::load_extern_crate (const std::string &crate_name, location_t locus)
...
+  auto cli_extern_crate = extern_crates.find (crate_name);
+
+  std::pair<std::unique_ptr<Import::Stream>, std::vector<ProcMacro::Procmacro>>
+    s;
+  if (cli_extern_crate != extern_crates.end ())
+    {
+      auto path = cli_extern_crate->second;
+      s = Import::try_package_in_directory (path, locus);
+    }
+
```
CohenArthur, line 987:
```suggestion
```

CohenArthur (approved):
Looks good to me! Lots of mappings but I think we need them

gcc/rust/ast/rust-ast.h
P-E-P: This was an artifact from some old code, fixed.

gcc/rust/rust-session-manager.cc
P-E-P: Fixed.

PR #2305, opened 2023-06-17
Prevent invalid iterator dereference

### LINKED ISSUE #2302: Out of bounds read in rust-lex.cc
The compiler fails to build with the adress sanitizer enabled, since our lexer reads a pointer which might be null.

https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/lex/rust-lex.cc#L258-L261

The condition and variable declaration needs to be switched. The check for `keyword` to be valid needs to happen *before* the creation of the `id` variable.

powerboat9:
Fixes #2302


CohenArthur (approved):
Thanks!

PR #2312, opened 2023-06-21
Fix rebinding imports

powerboat9:
This fixes rebinding imports, and should also get us closer to working glob imports.

### DIFF
--- a/gcc/rust/resolve/rust-ast-resolve-item.cc
+++ b/gcc/rust/resolve/rust-ast-resolve-item.cc
@@ -891,33 +891,30 @@ ResolveItem::resolve_extern_item (AST::ExternalItem *item)
 }
 
 static void
-flatten_glob (const AST::UseTreeGlob &glob,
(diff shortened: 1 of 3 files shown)


CohenArthur (changes requested):
Import resolution needs to happen during early name resolution and macro expansion, so not within the "late" name resolver that is `rust-ast-resolve-item.cc`. At the moment the system is a bit of a mess and we are working on reimplementing our name resolution from scratch - I'll send the first PR soon, and it'd be great to have your contributions on it :)

CohenArthur:
@powerboat9 are you interested in updating/rebasing this PR? I think it would still be worth it to have rebinding imports fixed for the existing name resolution algorithm, since NR 2.0 is taking longer than I planned (due to me not having time to work on it at all, sorry about that)

powerboat9:
Did a quick rebase, this should be fine if tests pass

CohenArthur (approved):
thanks :)

PR #2313, opened 2023-06-21
Remove `AST::Dump` and replace it with `AST::TokenCollector`

### LINKED ISSUE #2294: Replace AST dump with call to `TokenCollector`
- The ast dump lacks several informations (missing keyword, missing field, missing attributes...)
- The `TokenCollector` shall transcribe back to a **perfect** Vector of token from the AST since it is used by procedural macros.
- Maintaining the token collector up to date is mandatory for procedural macros.

This means the token collector coupled with a call to `as_string` on resulting tokens would give us the exact rust text from the ast at a reduced cost (we would no longer have to maintain the ast dump code).

## Problem
The token collector lacks proper indentation (no indentation token exists) but some functions are already waiting for this formatting code (eg. `newline()`, `increment_indentation()`, `decrement_indentation()` and `indentation()`)

## Possible solution
These functions can be used by changing the TokenCollector internal buffer to accept some kind of c++11 `std::variant` containing either a token or a formatting character (`\n`, `\t`, ...).

- [ ] change the internal buffer of the token collector.
- [ ] ignore formatting char variants for procedural macro token collection.
- [ ] change ast dump call to `TokenCollector` visitor followed by calls to `as_string` on resulting tokens.
- [ ] Remove `dump_tokenstream` (`-frust-ast-tokenstream`) from the session manager.

P-E-P:
Remove tokenstream dump and replace current ast dump call with call to tokencollector. This will allow us to easily keep those very similar things accurate. Also they will stay in sync.

Fixed #2294


philberty (approved):
LGTM clever idea

philberty (changes requested):
Actually scratch that I just reasilied... Do we want to remove the old AST Dump because there is a difference.

This AST dump should really be the AST Dump of _everything_ since its got all the tokens, but the existing AST dump are the structures we are actually working for name-resolution, expansion and hir lowering so i think its important to keep the existing one around.

Not sure this is the right change to go for right now.

P-E-P:
For future reference here are the arguments we discussed during the call this morning and why I don't see any point in changing this PR:

The old `AST::Dump` class was doing the following conversion with a visitor
```mermaid
graph LR;
A[AST]-->|AST::Dump|S[String]
S-->O[Stream output]
```

When designing procedural macros we required to collect the tokens corresponding to a given ast subtree, this means we had to implement the same kind of visitor as `AST::Dump` but instead of a string output, we have a Token output.

This means we had two different visitors, doing essentially the same thing, given that a conversion from a token to a string is trivial.

Procedural macro token dump **SHALL** be correct for procedural macros to expand properly (this does not mean it hasn't any error though) whilst ast dump was in an incomplete state. Instead of maintaining two different visitors, this PR proposes to unify those two visitors.

```mermaid
graph LR;
A[AST]-->|AST::TokenCollector|C[CollectItem]
C-->T[Token]
C-->M[Format]
T-->S[String]
T-->P[Procedural macro]
M-->S
S-->O[Stream output]
```

Instead of collecting only tokens the ast collector can now also collect some formatting metadata (Indentation, newline) depending on the context, as well as some comments. When given to a procedural macro, those metadata are filtered out. When given to the dump process, those metadata are used to build a properly formatted rust code.

philberty (approved):
LGTM as we discussed early sorry i didnt realise that this was from the AST structures

CohenArthur:
As mentioned in private with @P-E-P, I think it would be worth running the AST dump testsuite on both implementations of the dump to compare them. This PR is a really good first start and we don't have a super strong dependency on the AST dump at the moment, so I'm okay with merging it. But it would be nice to see where it needs to improve

CohenArthur:
So from a first run it seems this new AST dump is already more correct than the old one:

old one:
635 tests: 250 tests passed, 385 tests failed
new one:
635 tests: 335 tests passes, 300 tests failed

The testsuite was as follows: Check that gccrs can compile the AST dump version of a valid Rust file in our testsuite

There are still a few things missing but we can work on them in the future. I'll review the PR but I'm in favor of merging it sooner rather than later

PR #2316, opened 2023-06-23
rust: Add -frust-name-resolution-2.0 option

CohenArthur:
This option enables an experimental name resolution algorithm. Disabled
by default.

gcc/rust/ChangeLog:

	* lang.opt: Add -frust-name-resolution-2.0 option


philberty (approved):
This is an interesting way to do it. I like it because it means we can merge the new resolver in sooner than later and default to the old resolver in the short term and we can all focus on the new one when you merge in some code.

philberty:
@CohenArthur merge whenever you like go for it

P-E-P (approved):

PR #2317, opened 2023-06-23
util: Add tl::expected type

CohenArthur:
Directly taken from https://github.com/TartanLlama/expected and adapted
slightly for our use case.

gcc/rust/ChangeLog:

	* util/expected.h: New file.

This type offers a type similar to `std::expected`, which is only available since C++23. It will be useful for cleaning up our error handling, chaining errors together, and easier control flow in general.

This will be used for the upcoming experimental name resolution algorithm.

### DIFF
--- a/gcc/rust/util/expected.h
+++ b/gcc/rust/util/expected.h
@@ -1,0 +1,2444 @@
+///
+// expected - An implementation of std::expected with extensions
+// Written in 2017 by Sy Brand (tartanllama@gmail.com, @TartanLlama)
+//
+// Documentation available at http://tl.tartanllama.xyz/
+//
+// To the extent possible under law, the author(s) have dedicated all
+// copyright and related and neighboring rights to this software to the
+// public domain worldwide. This software is distributed without any warranty.
+//
+// You should have received a copy of the CC0 Public Domain Dedication
+// along with this software. If not, see
+// <http://creativecommons.org/publicdomain/zero/1.0/>.
+///
+
+#ifndef TL_EXPECTED_HPP
+#define TL_EXPECTED_HPP
+
+#define TL_EXPECTED_VERSION_MAJOR 1
+#define TL_EXPECTED_VERSION_MINOR 1
+#define TL_EXPECTED_VERSION_PATCH 0
+
+#include <exception>
+#include <functional>
+#include <type_traits>
+#include <utility>
+
+#if defined(__EXCEPTIONS) || defined(_CPPUNWIND)
+#define TL_EXPECTED_EXCEPTIONS_ENABLED
+#endif
+
+#if (defined(_MSC_VER) && _MSC_VER == 1900)
+#define TL_EXPECTED_MSVC2015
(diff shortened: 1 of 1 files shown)


philberty (commented):
Added in other GCC folks, are we allowed to do this in GCC?

Another option might be to take your optional implementation and reuse it to bootstrap a Result<Value, Error> class helper.

philberty:
I am not sure how i feel about it my initial thought is we might have issues using this with GCC but on the other hand it could be a useful interface.

@dkm and @tschwinge any opinion?

tschwinge:
This has recently been discussed in GCC/Rust IRC, 2023-06-15: <https://github.com/Rust-GCC/gccrs/wiki/IRC-2023-06-15>.

tschwinge:
> This has recently been discussed in GCC/Rust IRC, 2023-06-15: https://github.com/Rust-GCC/gccrs/wiki/IRC-2023-06-15.

The outcome of that -- per my free interpretation -- was something like this:

Should talk to libstdc++ maintainers to determine whether doing this via existing (or future) libstdc++ code is feasible, whilst maintaining GCC 4.8 compatibility, and otherwise it's deemed OKish to import external code.

tschwinge:
One thing that's going to be useful for future maintenance: first import the external file(s) verbatim, and then do local changes in separate commit(s).  For some (?) external imports, `LOCAL_PATCHES` files are maintained -- not sure whether that'll be useful here, too?

```shell
$ find -name LOCAL_PATCHES
./libffi/LOCAL_PATCHES
./libsanitizer/LOCAL_PATCHES
./libstdc++-v3/src/c++17/fast_float/LOCAL_PATCHES
./libstdc++-v3/src/c++17/ryu/LOCAL_PATCHES
```

CohenArthur:
> One thing that's going to be useful for future maintenance: first import the external file(s) verbatim, and then do local changes in separate commit(s). For some (?) external imports, `LOCAL_PATCHES` files are maintained -- not sure whether that'll be useful here, too?
> 
> ```shell
> $ find -name LOCAL_PATCHES
> ./libffi/LOCAL_PATCHES
> ./libsanitizer/LOCAL_PATCHES
> ./libstdc++-v3/src/c++17/fast_float/LOCAL_PATCHES
> ./libstdc++-v3/src/c++17/ryu/LOCAL_PATCHES
> ```

@tschwinge this is going to create commits that do not compile, as the original implementation uses exceptions which we explicitly disable. So I'm not sure how needed this is?

tschwinge:
> this is going to create commits that do not compile, as the original implementation uses exceptions which we explicitly disable.

There's no such problem: the file will be unused initially.  :-)

CohenArthur:
> > this is going to create commits that do not compile, as the original implementation uses exceptions which we explicitly disable.
> 
> There's no such problem: the file will be unused initially. :-)

ah, that's a good point! :D

CohenArthur:
I'll split up the commit then

P-E-P (approved):

PR #2327, opened 2023-06-26
optional: Import tl's optional type

CohenArthur:
This is directly adapted from https://github.com/TartanLlama/optional.
This class is tested for C++11 and on GCC 4.8, which is our target mimimum
GCC version.

This optional class provides multiple advantages over our existing
implementation of Optional<T>:

- Easier API (no static constructors)
- Better method chaining (and_then, map...)
- Harsher semantics (no assign-through on Optional<T&>)

Over the std::optional type present in C++'s standard library, this type offers:

- Optional references without going through std::reference_wrapper
- Better method chaining

For these reasons, I am in favor of importing it directly rather than
adapt C++14's std::optional type to work with C++11 and GCC 4.8.

gcc/rust/ChangeLog:

	* util/optional.h: New file.


tschwinge:
<https://github.com/Rust-GCC/gccrs/pull/2317#issuecomment-1606937646> and following messages apply here, too.

philberty (approved):
Goooo for it....

gcc/rust/util/optional.h
```diff
@@ -0,0 +1,2460 @@
...
+  struct do_not_use
+  {
+  };
+  constexpr explicit nullopt_t (do_not_use, do_not_use) noexcept {}
+};
+/// Represents an empty optional
+static constexpr nullopt_t nullopt{nullopt_t::do_not_use{},
+				   nullopt_t::do_not_use{}};
+
+class bad_optional_access : public std::exception
```
philberty, line 801: can this be removed then

gcc/rust/util/optional.h
```diff
@@ -0,0 +1,2460 @@
...
+
+#ifndef TL_OPTIONAL_HPP
+#define TL_OPTIONAL_HPP
+
+#define TL_OPTIONAL_VERSION_MAJOR 1
+#define TL_OPTIONAL_VERSION_MINOR 1
+#define TL_OPTIONAL_VERSION_PATCH 0
+
+// TODO: Are these okay?
+#include <exception>
```
philberty, line 25: A bootstrap build will show if it works, I say we should include this into the rust-session-manager and merge it and see how MArks build farm copes with it before merging the other code so we can fix anything we find

dkm: IIRC, gcc is built without exceptions https://gcc.gnu.org/codingconventions.html#Exceptions

CohenArthur: I think it's okay to modify the file so it includes "rust-system.h" instead of the poisoned headers. As mentioned on the #2317 PR, I'll split up the commit into adding the original file and then modifying it.

tschwinge:
Re commit "optional: Adapt class to GCC's standards": "it also formats the file according to our .clang-format": that's generally frowned upon, as it makes it nearly impossible to see what actually has been changed.  Instead, how about making such files exempt from clang-format, via some magic in `.github/workflows/clang-format.yml` (is there an option to skip/exclude individual files -- or if only directories, put such things into an `gcc/rust/imported` or similar directory?), or use markers to have clang-format ignore this file, like `// clang-format off` on top of the file, <https://clang.llvm.org/docs/ClangFormatStyleOptions.html#disabling-formatting-on-a-piece-of-code>?  (That's however just my opinion/preference).

CohenArthur:
@tschwinge I would rather split this commit in two, then - one to format the file, and one to make the necessary changes. Any option other than adding `// clang-format off` at the top of the file will cause unaware contributors (like me) to autoformat the file regularly. I'm also in favor of having it formatted

CohenArthur:
Oh, I'm good with that too! But is it okay for GCC?

tschwinge:
> is it okay for GCC?

What problem would you anticipate?

CohenArthur:
just that the lack of respect of the coding style could be one of the review comments on the patch. but I'm good with dealing with that problem when it arrives :) I'll amend those commits

gcc/rust/util/optional.h
```diff
@@ -1289,24 +1280,28 @@ class optional : private detail::optional_move_assign_base<T>,
   TL_OPTIONAL_11_CONSTEXPR T &value() & {
     if (has_value())
       return this->m_value;
-    throw bad_optional_access();
+
+    gcc_unreachable();
```
tschwinge, line 1284: Just to verify: does `gcc_unreachable` here (and likewise elsewhere) have the correct semantics:
<https://github.com/Rust-GCC/gccrs/blob/7af86eaaffec8cab5ad80e242480fe9f6c1ffaa6/gcc/system.h#L828-L831>
..., that is: an unreachable code path (..., and the compiler should optimized based on that knowledge), or should the original `throw` rather map onto an `abort`:
<https://github.com/Rust-GCC/gccrs/blob/7af86eaaffec8cab5ad80e242480fe9f6c1ffaa6/gcc/system.h#L789-L795>
..., or thus `internal_error`?

CohenArthur: ah, I know I have used it in other places not as an optimization `unreachable` but more as a "should not reach here `unreachable`", so under the impression that it would always do the `fancy_abort`. I guess we could have a `rust_unreachable`, at least for now, because we're nowhere close to caring about optimization and always want those fancy aborts

CohenArthur: @tschwinge I've created a branch so the changes are made. I've added it to tomorrow's community call because I want to discuss it with you. I think there is value in having a macro that always `fancy_abort`s and in one that is used for optimization only in releases, but I don't think we are there yet

CohenArthur: In any way, these changes aren't specific to this PR, so for now I think `tl::optional` should stay similar to the rest of the gccrs codebase and use `gcc_unreachable` (even if I now think it is "invalid" and should be replaced by `fancy_abort`). I will fix it once we've reached a decision on the matter tomorrow :)

PR #2328, opened 2023-06-26
Remove old uses of Optional<T>

CohenArthur:
- optional: Import tl's optional type
- gccrs: Remove Rust::Optional in favor of tl::optional

Requires #2327 to be merged, and then this PR will be rebased


philberty (approved):
LGTM

PR #2329, opened 2023-06-26
Match tokens in macros more closely

### LINKED ISSUE #2264: Literals are not compared correctly in macros
I tried this code:

```rust
macro_rules! m {
    (1) => {};
}

m!(2);
```

I expected to see this happen: an error (rustc produces `error: no rules expected the token `2``)

Instead, this happened:  compiles successfully

### Meta

2d806801404fc7628a3d022da92b73cd06197f49

powerboat9:
Fixes #2264


gcc/testsuite/rust/compile/macro-issue2264.rs
```diff
@@ -0,0 +1,6 @@
+macro_rules! m {
```
matthewjasper, line 1: Can you add a similar test with identifiers instead of integer literals?

powerboat9: Modified the test

CohenArthur (approved):
LGTM, thank you!

PR #2332, opened 2023-06-27
ast: Change Identifier definition

P-E-P:
Change Identifier type definition from a simple typedef to a whole class with it's own node id.

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -28,10 +28,29 @@
 
 namespace Rust {
 // TODO: remove typedefs and make actual types for these
-typedef std::string Identifier;
 typedef int TupleIndex;
 struct Session;
 struct MacroExpander;
+
+class Identifier
+{
+public:
+  Identifier (std::string ident = "")
+    : ident (ident), node_id (Analysis::Mappings::get ()->get_next_node_id ())
+  {}
+
+  NodeId get_node_id () const { return node_id; }
+  const std::string &as_string () const { return ident; }
+
+  bool empty () const { return ident.empty (); }
+
+private:
+  std::string ident;
+  NodeId node_id;
+};
+
+std::ostream &
+operator<< (std::ostream &os, Identifier const &i);
 
 namespace AST {
 // foward decl: ast visitor
@@ -60,11 +79,11 @@ class Node : public Visitable
 public:
   /**
    * Get the kind of Node this is. This is used to differentiate various AST
-   * elements with very little overhead when extracting the derived type through
-   * static casting is not necessary.
+   * elements with very little overhead when extracting the derived type
+   * through static casting is not necessary.
    */
-  // FIXME: Mark this as `= 0` in the future to make sure every node implements
-  // it
+  // FIXME: Mark this as `= 0` in the future to make sure every node
+  // implements it
   virtual Kind get_ast_kind () const { return Kind::UNKNOWN; }
 };
 
@@ -93,8 +112,8 @@ public:
 
   virtual std::string as_string () const = 0;
 
-  /* Converts token tree to a flat token stream. Tokens must be pointer to avoid
-   * mutual dependency with Token. */
+  /* Converts token tree to a flat token stream. Tokens must be pointer to
+   * avoid mutual dependency with Token. */
   virtual std::vector<std::unique_ptr<Token>> to_token_stream () const = 0;
 
 protected:
@@ -151,9 +170,9 @@ class Token : public TokenTree, public MacroMatch
 
   const_TokenPtr tok_ref;
 
-  /* new idea: wrapper around const_TokenPtr used for heterogeneuous storage in
-   * token trees. rather than convert back and forth when parsing macros, just
-   * wrap it. */
+  /* new idea: wrapper around const_TokenPtr used for heterogeneuous storage
+   * in token trees. rather than convert back and forth when parsing macros,
+   * just wrap it. */
 
 public:
   // Unique pointer custom clone function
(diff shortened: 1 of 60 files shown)


gcc/rust/ast/rust-ast.h
```diff
@@ -28,11 +28,30 @@
...
 struct Session;
 struct MacroExpander;
 
+class Identifier
+{
+public:
+  Identifier (std::string ident = "")
+    : ident (ident), node_id (Analysis::Mappings::get ()->get_next_node_id ())
+  {}
+
```
philberty, line 41: I don't think we need a node_id for an Identifier as we do resolution to the NodeId of the parent item the identifier belongs to.

I think remove the node_id, or add a copy constructor and copy assignment constructor. We also wish to have location info but feel free to do this as a separate PR see: https://github.com/Rust-GCC/gccrs/issues/2333

philberty (changes requested):
I think remove the node_id is probably the best and add Location info in a separate PR

gcc/rust/ast/rust-ast.h
P-E-P: > I don't think we need a node_id for an Identifier as we do resolution to the NodeId of the parent item the identifier belongs to.
> 
> I think remove the node_id, or add a copy constructor and copy assignment constructor. We also wish to have location info but feel free to do this as a separate PR see: #2333

But I do need a node Id for procedural macro resolution. When I have such derive attribute:
```
#[derive(Toto, serde::Deserialize)]
```
`serde::Deserialize` is a `SimplePath` but `Toto` is an identifier.

philberty: Ah ok then go for it. But you will need to add the copy constructors otherwise you will get new node_id's

P-E-P: > Ah ok then go for it. But you will need to add the copy constructors otherwise you will get new node_id's

Good point, will do.

philberty: I am not completely versed in the derive macro and attribute syntax but worth double checking if rustc uses an Identifier structure there for Toto or if its using a PathExpression instead if it uses a PathExpression you might get some of the resolution steps for free

philberty (approved):
LGTM thanks for the changes

CohenArthur (approved):
Great stuff. Keep up the good work!

CohenArthur:
Will merge once the conflict is sorted out @P-E-P :)

P-E-P:
> Will merge once the conflict is sorted out @P-E-P :)

@CohenArthur done!

CohenArthur:
> > I don't think we need a node_id for an Identifier as we do resolution to the NodeId of the parent item the identifier belongs to.
> > I think remove the node_id, or add a copy constructor and copy assignment constructor. We also wish to have location info but feel free to do this as a separate PR see: #2333
> 
> But I do need a node Id for procedural macro resolution. When I have such derive attribute:
> 
> ```
> #[derive(Toto, serde::Deserialize)]
> ```
> 
> `serde::Deserialize` is a `SimplePath` but `Toto` is an identifier. They both share the same parent attribute with a common node id even though that's two different macros.

I feel like I've seen a lot of discussion around having `NodeId`s for `Identifier`s and the more I think about it, the more it seems like this makes sense *only* for proc macros. In this case, wouldn't it be better to have a different class specifically for these proc macro identifiers? E.g instead of having
```c++
class Identifier {
    std::string identifier;
    NodeId id;
    Location loc;
}
```
and ignore the `NodeId` in all cases *but* proc macro name resolution, why not have
```c++
class Identifier {
    std::string identifier;
    Location loc;
}

class SpecialMagicProcMacroIdentifier {
    Identifier ident;
    NodeId node;
}
```
?

@P-E-P @philberty not sure what you think of this?

P-E-P:
> > > I don't think we need a node_id for an Identifier as we do resolution to the NodeId of the parent item the identifier belongs to.
> > > I think remove the node_id, or add a copy constructor and copy assignment constructor. We also wish to have location info but feel free to do this as a separate PR see: #2333
> > 
> > 
> > But I do need a node Id for procedural macro resolution. When I have such derive attribute:
> > ```
> > #[derive(Toto, serde::Deserialize)]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `serde::Deserialize` is a `SimplePath` but `Toto` is an identifier. They both share the same parent attribute with a common node id even though that's two different macros.
> 
> I feel like I've seen a lot of discussion around having `NodeId`s for `Identifier`s and the more I think about it, the more it seems like this makes sense _only_ for proc macros. In this case, wouldn't it be better to have a different class specifically for these proc macro identifiers? E.g instead of having
> 
> ```c++
> class Identifier {
>     std::string identifier;
>     NodeId id;
>     Location loc;
> }
> ```
> 
> and ignore the `NodeId` in all cases _but_ proc macro name resolution, why not have
> 
> ```c++
> class Identifier {
>     std::string identifier;
>     Location loc;
> }
> 
> class SpecialMagicProcMacroIdentifier {
>     Identifier ident;
>     NodeId node;
> }
> ```
> 
> ?
> 
> @P-E-P @philberty not sure what you think of this?

This would mean we should change the AST structure to accept some `SpecialMagicProcMacroIdentifier` in attributes am I right ?

Those things are already a mess (tokenstream -> late parse -> MetaItems -> Identifier/SimplePath). This would also mean we would be moving away from the reference. Whilst I'm not entirely against it, I'm not keen of this direction. There might be an underlying problem, this situation does not feel right. Could we rely on something else in order to identify the macro match ?

I did investigate a little as suggested by @philberty on `PathExpression` but those structures are not used at all in rustc for this part of the name resolution.

philberty:
Sorry for the delay in reply. This was my initial reaction that it seemed off to have node_id attached to identifiers. Generally this only applies to Paths like simple, path-expr, type-path or qualified path or qualified type path. so yeah i am in favor of getting rid of this.

PR #2342, opened 2023-06-28
nr2.0: Add complex testcases

CohenArthur:
gcc/testsuite/ChangeLog:

	* rust/compile/name_resolution1.rs: New test.
	* rust/compile/name_resolution2.rs: New test.
	* rust/compile/name_resolution3.rs: New test.
	* rust/compile/name_resolution4.rs: New test.
	* rust/compile/name_resolution5.rs: New test.

Co-authored-by: @GuillaumeGomez

This PR adds some interesting test cases that our name resolution already handles properly, but that the reimplementation must handle too

### DIFF
--- a/gcc/testsuite/rust/compile/name_resolution4.rs
+++ b/gcc/testsuite/rust/compile/name_resolution4.rs
@@ -1,0 +1,13 @@
+trait Foo {
+    fn foo(&self) {} // { dg-warning "unused name" }
+}
+
+struct Bar;
+
+pub fn bar() {
+    impl Foo for Bar {}
+}
+
+fn main() {
+    Bar.foo();
+}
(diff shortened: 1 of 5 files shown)


P-E-P (approved):

gcc/testsuite/rust/compile/name_resolution4.rs
```diff
@@ -0,0 +1,13 @@
...
+}
+
+struct Bar;
+
+pub fn bar() {
+    impl Foo for Bar {}
+}
+
+fn main() {
+    Bar.foo();
```
philberty, line 12: name-resolution does not resolve the method it only resolves the expression path of Bar.

philberty (approved):
LGTM i think we have some quite good test cases already but these are nice an explicit.

gcc/testsuite/rust/compile/name_resolution4.rs
CohenArthur: right, and that's something we need to handle with the new name resolution too (deferring to type-checking or however we call it at the moment)

PR #2343, opened 2023-06-28
nr2.0: Add basic Rib class

CohenArthur:
This class adds a basic Rib class for the new name resolution algorithm.
It uses `optional` and `expected` return types in order to try and
improve error handling in these new passes.

gcc/rust/ChangeLog:

	* Make-lang.in: Add `rust-rib.cc` object.
	* resolve/rust-rib.cc: New file.
	* resolve/rust-rib.h: New file.

Needs #2317 and #2327

### DIFF
--- a/gcc/rust/resolve/rust-rib.h
+++ b/gcc/rust/resolve/rust-rib.h
@@ -1,0 +1,135 @@
+// Copyright (C) 2020-2023 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#ifndef RUST_RIB_H
+#define RUST_RIB_H
+
+#include "rust-system.h"
+#include "rust-ast.h"
+#include "optional.h"
+#include "expected.h"
+
+namespace Rust {
+namespace Resolver2_0 {
+
+/**
+ * All namespaces that Rust's name resolution needs to handle
+ */
+// TODO: Move to `rust-forever-stack.h`?
+enum class Namespace
+{
+  Values,
+  Types,
+  Labels,
+  Macros,
+  // TODO: Which namespaces are we missing?
+};
+
+/**
+ * Error returned by `Rib::insert` when the key was already present in the Rib's
+ * map. The class contains the previously-inserted NodeId as well as the name of
+ * the node.
+ */
+struct DuplicateNameError
+{
+  // TODO: We might need multiple kinds of errors later down the line
+  DuplicateNameError (std::string name, NodeId existing);
+
+  std::string name;
+  NodeId existing;
+};
+
+/**
+ * A rib is a container of nodes, either declaration or usages, as well as the
+ * identifier each node uses. They are used to delimit lexical scopes, and have
+ * an impact on name resolution - they restrict certain name accesses and serve
+ * as boundaries between scopes.
+
+ * For example, if we are resolving the following *variable* use:
+ *
+ * ```rust
+ * fn outer() {
+ *     let a = 15; // decl
+ *     fn inner() -> i32 {
+ *         a // use
+ *     }
+ * }
+ * ```
+ *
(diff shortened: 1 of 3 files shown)


gcc/rust/resolve/rust-rib.h
```diff
@@ -0,0 +1,133 @@
...
+    Normal,
+    Module,
+    Function,
+    Item, // TODO: what's an item scope? Is that the original crate's scope?
+    TraitOrImpl,
+    Closure,
+    ConstantItem, // -> this variant has a boolean
+    MacroDefinition,
+    ForwardTypeParam, // TODO: What is this one for?
+    ConstParamType,   // TODO: Const parameter ? Like... Constant item? Or Const
```
matthewjasper, line 92: const generic, as in the use of `T` here:

```rust
fn f<T, const X: T>() { ... }
                 ^
```

gcc/rust/resolve/rust-rib.h
```diff
@@ -0,0 +1,133 @@
...
+ */
+class Rib
+{
+public:
+  enum class Kind
+  {
+    Normal,
+    Module,
+    Function,
+    Item, // TODO: what's an item scope? Is that the original crate's scope?
```
matthewjasper, line 86: Any item that doesn't correspond to some other Kind (for example if instead of a function the above example was a `static`).

gcc/rust/resolve/rust-rib.h
```diff
@@ -0,0 +1,133 @@
...
+    TraitOrImpl,
+    Closure,
+    ConstantItem, // -> this variant has a boolean
+    MacroDefinition,
+    ForwardTypeParam, // TODO: What is this one for?
```
matthewjasper, line 91: For this:

```rust
struct A<T = S, S = ()>(T, S);
             ^ Cannot use S here
```

gcc/rust/resolve/rust-rib.h
```diff
@@ -0,0 +1,133 @@
...
+    ConstParamType,   // TODO: Const parameter ? Like... Constant item? Or Const
+		      // generic? */
+  };
+
+  Rib (Kind kind);
+  Rib (Kind kind, Identifier identifier, NodeId id);
+  Rib (Kind kind, std::unordered_map<Identifier, NodeId> values);
+
+  // TODO: What's the correctbehavior if the key already exists? What if a decl
+  // and use are in the same rib? Is that possible? Okay based on RibKind?
```
matthewjasper, line 101: If the key exists it's generally an error, except in cases where shadowing is allowed (local variables, loop labels).

Decls generally have to be visited before uses.

gcc/rust/resolve/rust-rib.h
```diff
@@ -0,0 +1,133 @@
...
+ * All namespaces that Rust's name resolution needs to handle
+ */
+// TODO: Move to `rust-forever-stack.h`?
+enum class Namespace
+{
+  Values,
+  Types,
+  Labels,
+  Macros,
+  // TODO: Which namespaces are we missing?
```
philberty, line 40: There are the only ones i know of. Though possibly const generics are put in their own namespace but i doubt it.

gcc/rust/resolve/rust-rib.h
```diff
@@ -0,0 +1,133 @@
...
+};
+
+/**
+ * Error returned by `Rib::insert` when the key was already present in the Rib's
+ * map. The class contains the previously-inserted NodeId as well as the name of
+ * the node.
+ */
+struct AlreadyInserted
```
philberty, line 48: Maybe DuplicateNameError is a better name.

gcc/rust/resolve/rust-rib.h
```diff
@@ -0,0 +1,133 @@
...
+
+  /**
+   * Access an inserted NodeId.
+   *
+   * @return tl::nullopt if the key does not exist, the NodeId otherwise
+   */
+  tl::optional<NodeId> get (const Identifier &name);
+
+  /* View all the values stored in the rib */
+  const std::unordered_map<Identifier, NodeId> &get_values () const;
```
philberty, line 123: Good idea to start using Identifier but we need to make sure it has a comparator function

philberty (changes requested):
Nice one just style nits

gcc/rust/resolve/rust-rib.h
CohenArthur: This set should only be "name-based" so I guess maybe it would be better to store the Identifier's string as a key with the new changes being done to that class. Because the `values` unordered_map contains the Identifier's NodeId as a value. Or maybe it would be better to turn this into an `std::set<Identifier>`. I don't know

CohenArthur: So for this set I think we should go with an `unordered_map<string, NodeId>`. Let's say we're inserting a new function in the rib: we care about the function's name, so we can use its Identifier, but we care about the function's node ID - not the identifier's node ID. So I don't think having a `set<Identifier>` makes sense for the rib. The one place where we will be using the Identifier's NodeId for resolution is for resolving proc macros, as @P-E-P pointed it out

Issue #1638, opened 2022-11-15
Remove C-style comments in frontend and prefer C++-style


CohenArthur:

jdupak:
What is the difference?

dkm:
C uses `/* ... */` and C++ uses `// ......`. But maybe that's not the question?

jdupak:
> C uses `/* ... */` and C++ uses `// ......`. But maybe that's not the question?

Yes, thanks, that was the question. But this is the first time I hear of such a distinction.

CohenArthur:
Yeah, sorry about that. I made these issues when reading the reviews on the patches and should have explained some of them a little better :)

anuragcs:
For removing the c-style comments in frontend and prefer c++ style we can switch to "  // "

CohenArthur:
Yep exactly @anuragcs :) are you interested in working on this?

anuragcs:
@CohenArthur .. Yes I would love to contribute my share and work on following issue

TieWay59:
@CohenArthur I wish to learn about the expected file range of this issue, because I searched current code base and got a  result of 20015 hits in 1688 files within `*.c, *.cc, *.h`.

tschwinge:
Only `gcc/rust/`, `libgrust/`, plus a few GCC/Rust-specific files elsewhere.

But I'm not convinced that such a mass-change is really a good use of everyone's time: to prepare the changes, and also the conflicts that arise when cherry-picking patches, including preparation of patches for upstream submission, etc.?  I mean, there's no actual problem with `/* [...] */`-style comments?

TieWay59:
Here is an example of `gcc/rust/` search result: https://github.com/search?q=repo%3ARust-GCC%2Fgccrs+%22%2F*%22+%28path%3A*.c+OR+path%3A*.cc+OR+path%3A*.h%29+path%3A%2F%5Egcc%5C%2Frust%5C%2F%2F+&type=code

There are 122  files in the result with more than 2K `/*`. 

I saw some difficult comment pattern what would not be easy to convert:

https://github.com/Rust-GCC/gccrs/blob/b028c7ece2ad67236e55a38b776c0bf70408a53e/gcc/rust/lex/rust-token.h#L159-L162

CohenArthur:
[...]

PR #2344, opened 2023-06-28
nr2.0: Add `ForeverStack` data structure.

CohenArthur:
This data structure replaces our existing scopes and allows for multiple
name resolution passes to insert definitions or look up names at different
stages of the pipeline. The documentation goes into detail about how the
data structure works, and how the source's hierarchy is preserved
throughout the pipeline.

The class is templated to enforce specific behavior based on the namespace
the scope map targets.

gcc/rust/ChangeLog:

	* resolve/rust-forever-stack.h: New file.
	* resolve/rust-forever-stack.hxx: New file.

Needs #2327 and #2317

### DIFF
--- a/gcc/rust/resolve/rust-forever-stack.h
+++ b/gcc/rust/resolve/rust-forever-stack.h
@@ -1,0 +1,533 @@
+// Copyright (C) 2020-2023 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#ifndef RUST_FOREVER_STACK_H
+#define RUST_FOREVER_STACK_H
+
+#include "rust-system.h"
+#include "rust-rib.h"
+#include "rust-ast.h"
+#include "rust-path.h"
+#include "optional.h"
+#include "expected.h"
+
+namespace Rust {
+namespace Resolver2_0 {
+
+/**
+
+Let's look at our stack for resolving and traversing the following Rust code:
+
+```rust
+mod foo {
+    mod bar {
+	fn outer() {
+	    fn inner() {}
+	}
+
+	fn another() {}
+    }
+}
+```
+
+We start by creating the stack, which contains only one rib - the crate's. We
+won't look in details on how different namespaces end up with different stacks,
+and will only consider the "value" namespace for this example. Modules do not
+get added to the value namespace, but functions do:
+
+```rust
+let _ = foo;   // foo is a module, invalid Rust code
+let _ = outer; // outer is a function, ok!
+```
+
+So passing each module will create a new Rib, but not add that module's node to
+the Rib.
+
+The current cursor of the stack will be denoted with `-->`: an arrow pointing to
+the current rib.
+
+When we start the `TopLevel` pass on the crate we are compiling, we only see the
+top rib, which is empty at first:
+
+      ┌───────────────┐
+      │               │
+  --> │               │
+      │               │
+      └───────────────┘
+
+We pass through our first module, and emplace another Rib: Another "scope" is
+created, and it impacts name resolution rules.
+
+      ┌───────────────┐
+      │               │
+      │               │
+      │               │
+      └───────┬───────┘
+	      │
+	  foo │
+	      │
+	      ▼
+      ┌───────────────┐
+      │               │
+  --> │               │
+      │               │
+      └───────────────┘
+
+Notice that we have moved the cursor to the newly-created Rib, and that we have
+added a path between the two ribs - this is a `Link`. A link contains
+information such as the scope's NodeId, as well as an optional path - present
+only when the scope is named. This allows us to easily fetch AST nodes based on
+their canonical path, or build a canonical path from a NodeId. It also makes it
+really easy to do complex path name resolution, such as `super::super::<item>`.
+As mentioned earlier, modules are not present in the value namespace, so our new
+rib is also empty. Let's pass through the second module:
+
+      ┌───────────────┐
+      │               │
+      │               │
+      │               │
+      └───────┬───────┘
+	      │
+	  foo │
+	      │
+	      ▼
+      ┌───────────────┐
+      │               │
+      │               │
+      │               │
+      └───────┬───────┘
+	      │
+	  bar │
+	      │
+	      ▼
+      ┌───────────────┐
+      │               │
+  --> │               │
+      │               │
+      └───────────────┘
+
+Once again, the new rib is empty, and we have a link with a path. We now go
+through each item in the `bar` module and visit them. The first item is a
+function, `outer` - upon being visited, it adds itself to the current rib.
+
+      ┌───────────────┐
+      │               │
+      │               │
+      │               │
+      └───────┬───────┘
+	      │
+	  foo │
+	      │
+	      ▼
+      ┌───────────────┐
+      │               │
+      │               │
+      │               │
+      └───────┬───────┘
+	      │
+	  bar │
+	      │
+	      ▼
+      ┌───────────────┐
+      │outer          │
+  --> │               │
+      │               │
+      └───────────────┘
+
+We now visit `outer`'s definition. This creates a new Rib, as functions can have
+arguments, whose declaration only lives for the function's scope.
+
+      ┌───────────────┐
+      │               │
+      │               │
+      │               │
+      └───────┬───────┘
+	      │
+	  foo │
+	      │
+	      ▼
+      ┌───────────────┐
+      │               │
+      │               │
+      │               │
+      └───────┬───────┘
+	      │
+	  bar │
+	      │
+	      ▼
+      ┌───────────────┐
+      │outer          │
+      │               │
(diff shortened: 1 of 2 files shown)


gcc/rust/resolve/rust-forever-stack.h
```diff
@@ -0,0 +1,531 @@
...
+  };
+
+  /* Should we keep going upon seeing a Rib? */
+  enum class KeepGoing
+  {
+    Yes,
+    No,
+  };
+
+  /* Add a new Rib to the stack. This is an internal method */
```
P-E-P, line 509: >   /* Add a new Rib to the stack. This is an internal method */

Shouldn't it be private then ?

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -0,0 +1,227 @@
...
+  // If the link does not exist, we create it and emplace a new `Node` with the
+  // current node as its parent. `unordered_map::emplace` returns a pair with
+  // the iterator and a boolean. If the value already exists, the iterator
+  // points to it. Otherwise, it points to the newly emplaced value, so we can
+  // just update our cursor().
+  auto emplace
+    = cursor ().children.emplace (std::make_pair (link, Node (rib, cursor ())));
+
+  auto it = emplace.first;
+  auto didnt_exist = emplace.second;
```
P-E-P, line 71: Inverting the value here might be a little clearer in the reader's mind ?

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -0,0 +1,227 @@
...
+template <Namespace N>
+tl::expected<NodeId, AlreadyInserted>
+ForeverStack<N>::insert (Identifier name, NodeId node)
+{
+  auto &innermost_rib = peek ();
+
+  // So what do we do here - if the Rib has already been pushed in an earlier
+  // pass, we might end up in a situation where it is okay to re-add new names.
+  // Do we just ignore that here? Do we keep track of if the Rib is new or not?
+  // should our cursor have info on the current node like "is it newly pushed"?
```
P-E-P, line 99: We could encode such information in the type system, mutating a YoungNode into an Grownode, this would prevent state in the cursor as well as state in the node, once grown there would be no way back.

Just throwing random ideas around, don't mind me :laughing:

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+  std::cerr << next << "rib: {\n";
+
+  for (const auto &kv : rib.get_values ())
+    std::cerr << next_next << kv.first << ": " << kv.second << "\n";
+
+  std::cerr << next << "},\n";
+
+  return "";
```
P-E-P, line 181: I'm not sure about the intent of this function. Are you dumping the rib to `std::cerr` or a string ? The result string seems kinda useless in the current state. I would construct the string and let the caller send it to `cerr`.

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -0,0 +1,227 @@
...
+      auto link = kv.first;
+      auto child = kv.second;
+      std::cerr << indent << "Link (" << link.id << ", "
+		<< (link.path.has_value () ? link.path.value () : "<anon>")
+		<< "):\n";
+
+      std::cerr << node_str (indentation + 4, child) << '\n';
+    }
+
+  return "";
```
P-E-P, line 214: Same question as earlier.

gcc/rust/resolve/rust-forever-stack.h
CohenArthur: Oh, I thought it was!

gcc/rust/resolve/rust-forever-stack.hxx
CohenArthur: yeah, the dump is unfinished - but this is what it should be doing. keep a `stringstream` and then return it, and the session manager can dump it to `std::cerr` or to a file. forgot I hadn't fixed that!

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -0,0 +1,227 @@
...
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
```
philberty, line 17: Why do we have two headers? Not sure i like the .hxx extension. Would prefer a different name or combine the header.

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -0,0 +1,227 @@
...
+// inline template <>
+// tl::optional<NodeId>
+// ForeverStack<Namespace::Values>::get (const Identifier &name)
+// {
+//   return {};
+// }
+
+template <Namespace N>
+std::string
+ForeverStack<N>::rib_str (const Rib &rib, const std::string &next,
```
philberty, line 165: the convention we have at the moment is as_string or debug_str i think is better. Another useful one is a debug function which calls as_string or debug_str and calls rust_debug on it.

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -0,0 +1,227 @@
...
+template <Namespace N>
+std::string
+ForeverStack<N>::node_str (unsigned indentation,
+			   const ForeverStack<N>::Node &node)
+{
+  auto indent = std::string (indentation, ' ');
+  auto next = std::string (indentation + 4, ' ');
+  auto next_next = std::string (indentation + 8, ' ');
+
+  std::cerr << indent << "Node {\n"
```
philberty, line 193: I think logging is really useful but it should be rust_debug lines.

gcc/rust/resolve/rust-forever-stack.h
```diff
@@ -0,0 +1,531 @@
...
+   *        scope
+   *
+   * @return a valid option with the NodeId if the identifier is present in the
+   *         current map, an empty one otherwise.
+   */
+  inline tl::optional<NodeId> get (const Identifier &name);
+
+  // TODO: Should we dump to a file instead, like the other dumps? Or just print
+  // the string to a file in the SessionManager?
+  std::string debug_str ();
```
philberty, line 457: This is the right thing to do because you can do assertions and call rust_debug in dev. Writing to file should be something where we change rust_debug to be a logger of some kind.

philberty (changes requested):
I like it, just style nits. I think keeping logging in is a good idea. just need to change them to rust_debug lines

gcc/rust/resolve/rust-forever-stack.h
```diff
@@ -0,0 +1,531 @@
...
+   *        lexical scope, like a module's.
+   */
+  void push (Rib rib, NodeId id, tl::optional<Identifier> path = {});
+
+  /**
+   * Pop the innermost Rib from the stack
+   */
+  void pop ();
+
+  /**
```
dkm, line 423: There was #1638 , but it's getting killed for good reasons. But do we want to enforce some kind of comments for new comments? Using `/** ... */` with doxygen style tags are fine. I was wondering how to properly document other changes.

gcc/rust/resolve/rust-forever-stack.hxx
P-E-P: Tbh hxx for template only implementations is pretty standard. We've discussed them a bit in the past with @CohenArthur and agreed on not using them when there is a non templated implementation, no `.hxx` file should exist along a `.cc` file for a same given header file.

I believe that when handled correctly those might reduce compile time, do not quote me on that though as they're included anyway with the header ?.

gcc/rust/resolve/rust-forever-stack.h
CohenArthur: I checked, and it is an internal method :D

gcc/rust/resolve/rust-forever-stack.hxx
CohenArthur: Let's talk about this tomorrow during the community call :)

gcc/rust/resolve/rust-forever-stack.h
CohenArthur: This is the way I prefer, but I'm fine with other styles. As long as I can write `@return` and see it get highlighted in my editor, I'm happy haha

CohenArthur:
For debugging, I went with an `as_debug_string` method which returns a string and uses a `stringstream` internally. We can dump it using `rust_debug` or write it to a file in the session manager

gcc/rust/resolve/rust-forever-stack.h
powerboat9: What do you use for an editor?

CohenArthur: https://helix-editor.com/ :)

gcc/rust/resolve/rust-forever-stack.hxx
tschwinge: By chance, I've just found that @rsandifo-arm's commit 73b7582775254b764fd92ddb252a33dc15872c69 "Add rtl-ssa" did add `*.inl` files for what appears to be a similar purpose (from a quick look)?

<https://github.com/Rust-GCC/gccrs/blob/73b7582775254b764fd92ddb252a33dc15872c69/gcc/rtl-ssa/internals.inl#L1>
<https://github.com/Rust-GCC/gccrs/blob/73b7582775254b764fd92ddb252a33dc15872c69/gcc/rtl-ssa/is-a.inl#L1>
<https://github.com/Rust-GCC/gccrs/blob/73b7582775254b764fd92ddb252a33dc15872c69/gcc/rtl-ssa/member-fns.inl#L20-L26>

(I have no opinion on `*.inl` vs `*.hxx`.)

PR #2337, opened 2023-06-27
Combine {,Proc}MacroInvocLexer

powerboat9:
Addresses/Solves #2276

PR #2351, opened 2023-06-30
Create MacroInvocLexerBase

powerboat9:
Split from #2337

### DIFF
--- a/gcc/rust/expand/rust-macro-invoc-lexer.h
+++ b/gcc/rust/expand/rust-macro-invoc-lexer.h
@@ -22,6 +22,33 @@
 #include "rust-ast.h"
 
 namespace Rust {
+template <class T> class MacroInvocLexerBase
+{
+public:
+  MacroInvocLexerBase (std::vector<T> stream)
+    : offs (0), token_stream (std::move (stream))
+  {}
+
+  // Advances current token to n + 1 tokens ahead of current position.
+  void skip_token (int n) { offs += (n + 1); }
+
+  // Skips the current token.
+  void skip_token () { skip_token (0); }
+
+  std::string get_filename () const
+  {
+    // FIXME
+    gcc_unreachable ();
+    return "FIXME";
+  }
+
+  size_t get_offs () const { return offs; }
+
+protected:
+  size_t offs;
+  std::vector<T> token_stream;
+};
+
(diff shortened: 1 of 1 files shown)


gcc/rust/expand/rust-macro-invoc-lexer.h
```diff
@@ -22,6 +22,33 @@
 #include "rust-ast.h"
 
 namespace Rust {
+template <class T> class MacroInvocLexerBase
+{
+public:
+  MacroInvocLexerBase (std::vector<T> stream)
```
CohenArthur, line 28:
```suggestion
  MacroInvocLexerBase (std::vector<T> &&stream)
```

gcc/rust/expand/rust-macro-invoc-lexer.h
```diff
@@ -22,6 +22,33 @@
...
+
+  // Skips the current token.
+  void skip_token () { skip_token (0); }
+
+  std::string get_filename () const
+  {
+    // FIXME
+    gcc_unreachable ();
+    return "FIXME";
+  }
```
CohenArthur, line 43: Do we need that function in the base class?

CohenArthur (approved):
Nice :D this will clean things up nicely!

gcc/rust/expand/rust-macro-invoc-lexer.h
powerboat9: https://stackoverflow.com/a/1849280

gcc/rust/expand/rust-macro-invoc-lexer.h
powerboat9: Whatever mechanism we end up using for ```get_filename``` would probably (?) involve an extra ```filename``` field, which would be handled the same way for all child classes

gcc/rust/expand/rust-macro-invoc-lexer.h
CohenArthur: But does the `MacroInvocLexer` need to store a copy? Or should it own its stream? I'm thinking it should own the vector, which might be created on the fly in the case of a `ProcMacroLexer` and thus needs to be moved to avoid unecessary copies

PR #2340, opened 2023-06-28
Ast collector fixes

P-E-P:
Fixes several issues with current ast collector.

Requires #2313 

Fixes #7

PR #2352, opened 2023-06-30
Fix type hint propagation

P-E-P:
Type hint was not propagated properly in literals, being replaced by a default `str` instead.

Needs #2340


CohenArthur (approved):
Nice fix thank you :)

PR #2353, opened 2023-06-30
abi: Add testcase for using "Rust" as ABI string

CohenArthur:
gcc/testsuite/ChangeLog:

	* rust/compile/rust_abi.rs: New test.


philberty (approved):
LGTM

PR #2356, opened 2023-07-01
Allow instances of Identifier to be compared and hashed

### LINKED ISSUE #2354: Add comparator to new AST identifier
To take advantage of the new Identifier class we should add a comparator to the identifer so it can be used in std::set and std::map.

```c++
bool operator< (const Identifier& rhs) const {
  return identifier.compare (rhs.identifier);
}
```

https://github.com/Rust-GCC/gccrs/blob/251348f3c8bc218360877a3bb9f16d9d5c7a685f/gcc/rust/ast/rust-ast.h#L35

powerboat9:
Fixes #2354

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -49,6 +49,32 @@ public:
 
   bool empty () const { return ident.empty (); }
 
+  bool operator<(const Identifier &oth) const { return node_id < oth.node_id; }
+
+  bool operator<= (const Identifier &oth) const
+  {
+    return node_id <= oth.node_id;
+  }
+
+  bool operator> (const Identifier &oth) const { return node_id > oth.node_id; }
+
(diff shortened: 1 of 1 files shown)


CohenArthur (changes requested):
I think we should look at removing the `NodeId` from the `Identifier` class as we talked about with @P-E-P in https://github.com/Rust-GCC/gccrs/pull/2332#issuecomment-1627724200. So I don't think we want to go through with comparisons on the id

P-E-P:
> I think we should look at removing the `NodeId` from the `Identifier` class as we talked about with @P-E-P in [#2332 (comment)](https://github.com/Rust-GCC/gccrs/pull/2332#issuecomment-1627724200). So I don't think we want to go through with comparisons on the id

The NodeId has been removed in #2300

philberty:
As this has been remvoed lets close this and the issue

Issue #2361, opened 2023-07-02
Get rid of Location wrappers


philberty:
We should get rid of rust-location.h and move to gcc's location_t. This wrapper does not serve a purpose anymore and should be removed. It will allow us to integrate more closely with the normal GCC diagnostic's API which keeps getting better every year.

1. Remove rust-location.h
2. remove rust-location.cc
3. update Make-lang.in to remove rust-location.o
4. refactor the code base to use location_t instead of Location

PR #2362, opened 2023-07-03
Change class Location into typedef

powerboat9:
Addresses #2361


philberty (approved):
Awesome this looks like a good starting point. LGTM

philberty:
Want to let @CohenArthur review this after i just saw https://github.com/Rust-GCC/gccrs/pull/2363

I think this PR, https://github.com/Rust-GCC/gccrs/pull/2370 and https://github.com/Rust-GCC/gccrs/pull/2371 are a good way to get rid of the location wrappers

PR #2363, opened 2023-07-03
Replace default initialization ```Location ()``` with ```UNDEF_LOCATION```

powerboat9:
Depends on #2362, addresses #2361

### DIFF
--- a/gcc/rust/ast/rust-ast-collector.cc
+++ b/gcc/rust/ast/rust-ast-collector.cc
@@ -73,7 +73,7 @@ TokenCollector::visit_items_joined_by_separator (T &collection,
       auto size = collection.size () - end_offset;
       for (size_t i = start_offset + 1; i < size; i++)
 	{
-	  push (Rust::Token::make (separator, Location ()));
+	  push (Rust::Token::make (separator, UNDEF_LOCATION));
 	  visit (collection.at (i));
 	}
     }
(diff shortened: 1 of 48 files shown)


CohenArthur:
@powerboat9 did you just do a search and replace? If yes, I think those changes are a bit dangerous because `Location()` and `UNKNOWN_LOCATION` should be a bit different. Some places use `Location()` because they do not have a location *yet* and should have one, and some should not have a real location (like error AST fragments for example). So we have to be a bit careful. Keeping the default constructor is nice because we can then delete it and review all places that break, and figure out if they should have no location or a real location

powerboat9:
In hindsight ```location_t ()``` will probably work, since it should be zero-initialization and ```UNKNOWN_LOCATION == 0```

powerboat9:
Ok, I've replaced ```Location ()``` with ```UNDEF_LOCATION```, so we can differentiate between the two cases more easily.

philberty:
Yeah i havent fully reviewed this yet but i think having UNKNOWN_LOCATION or UNDEF_LOCATION will be great so we can easily search the code base for missing location info which improves out GDB output and diagnostics.

Off the top of my head anything in the typecheck and backend feel free to change it over to this but i think yeah in the expansion code its a little more tricky to figure out the distinctions but i think this is a great way to move forward to improve locations in general.

CohenArthur (approved):
LGTM! Thank you :) Is this ready to be merged?

powerboat9:
This should be ready now

PR #2370, opened 2023-07-03
Reduce Linemap/Gcc_linemap abstraction

powerboat9:
Depends on #2362

PR #2371, opened 2023-07-03
Reduce Linemap/Gcc_linemap abstraction further

powerboat9:
Depends on #2370


philberty (approved):
LGTM

CohenArthur (approved):
I think this cleanup makes sense because having these functions as part of the `SessionManager` was annoying. But I'm a bit conflicted since I'm not a fan of all the magic C macros in GCC and they make the life of contributors harder. I think it might be nice to still keep a small abstraction over a `location_t` and have methods such as `.column()` and `.file()` which would call the C macros. But I'm okay with this cleanup

CohenArthur:
@powerboat9 once you've rebased this I'll merge it

PR #2390, opened 2023-07-07
[E0070] invalid left-hand side of assignment

MahadMuhammad:
## [`E0070`](https://doc.rust-lang.org/error_codes/E0070.html) invalid left-hand side of assignment
-  Added error code support for an assignment operator which was used on a non-place expression.

---

### Code Tested:
```rust
fn foo() {
    1 = 3; // { dg-error "invalid left-hand side of assignment" }
}

fn main() {
    foo();
}
```

---

### Output:
```rust
gccrs/gcc/testsuite/rust/compile/wrong_lhs_assignment.rs:4:5: error: invalid left-hand side of assignment [E0070]
compiler exited with status 1
PASS: rust/compile/wrong_lhs_assignment.rs  (test for errors, line 4)
```

---

**gcc/rust/ChangeLog:**

	* resolve/rust-ast-verify-assignee.h: called error function.

**gcc/testsuite/ChangeLog:**

	* rust/compile/wrong_lhs_assignment.rs: New test.

---

### DIFF
--- a/gcc/rust/resolve/rust-ast-verify-assignee.h
+++ b/gcc/rust/resolve/rust-ast-verify-assignee.h
@@ -35,7 +35,7 @@ public:
     VerifyAsignee checker;
     assignee->accept_vis (checker);
     if (!checker.ok)
-      rust_error_at (assignee->get_locus (),
+      rust_error_at (assignee->get_locus (), ErrorCode ("E0070"),
 		     "invalid left-hand side of assignment");
     return checker.ok;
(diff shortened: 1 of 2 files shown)


CohenArthur (approved):
Nice!

philberty (approved):
LGTM not sure this rust-ast-verify-assignee.h will be kept around like this there is code in GCC we can reuse to get this so it might change in the future.

PR #2399, opened 2023-07-10
dump: Output escaped character values

### LINKED ISSUE #2398: String literal dump is raw
Dumped string literals keep their raw value

I tried this code:

```rust
const t: &str = "\tlol\nhello\"this is ' in string";
```

`cat gccrs.ast-pretty.dump`

```rust
const t : &str = "\tlol\nhello\"this is ' in string";
```

Instead, this happened:

`cat gccrs.ast-pretty.dump`

```rust
const t : &str = "	lol
hello"this is ' in string";
```

### Meta

- a06cec181de

P-E-P:
Output escaped character values instead of their raw values in string.

Fixes #2398

### DIFF
--- a/gcc/rust/lex/rust-token.cc
+++ b/gcc/rust/lex/rust-token.cc
@@ -16,6 +16,7 @@
 // along with GCC; see the file COPYING3.  If not see
 // <http://www.gnu.org/licenses/>.
 
+#include "rust-system.h"
 #include "rust-token.h"
 #include "rust-diagnostics.h"
 
@@ -172,6 +173,41 @@ Token::get_str () const
   return *str;
 }
 
+namespace {
+enum class Context
+{
+  String,
+  Char
+};
+
+const std::map<char, std::string> matches = {
+  {'\t', "\\t"}, {'\n', "\\n"},	 {'\r', "\\r"},
+  {'\0', "\\0"}, {'\\', "\\\\"}, {'\v', "\\v"},
+};
+
+std::string
+escape_special_chars (const std::string &source, Context ctx)
+{
+  std::stringstream stream;
+  decltype (matches)::const_iterator result;
+  for (char c : source)
+    {
+      // FIXME: #2411 Also replace escaped unicode values and \x digits
+      if ((result = matches.find (c)) != matches.end ())
+	stream << result->second;
+      else if (c == '\'' && ctx == Context::Char)
+	stream << "\\'";
+      else if (c == '"' && ctx == Context::String)
+	stream << "\\\"";
+      else
+	stream << c;
+    }
+
+  return stream.str ();
+}
+
+} // namespace
+
 std::string
 Token::as_string () const
 {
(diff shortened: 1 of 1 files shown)


gcc/rust/lex/rust-token.cc
```diff
@@ -180,13 +229,13 @@ Token::as_string () const
       switch (get_id ())
 	{
 	case STRING_LITERAL:
-	  return "\"" + get_str () + "\"";
+	  return "\"" + escape_special_chars (get_str (), true) + "\"";
```
philberty, line 232: tiny style thing, no need to do anything but if you have multiple flags passed to a function like this i would tend to write it as:

```escape_special_chars (get_str(), true /*in_string_flag*/)```

Comment (philberty) [reply gcc/rust/lex/rust-token.cc:232]: Just makes it a little more readable without having to look at the function declaration you're calling

**Review by philberty — APPROVED** [review]
LGTM

File: gcc/rust/lex/rust-token.cc [hunk 2]
```diff
@@ -172,6 +173,54 @@ Token::get_str () const
...
+	  else
+	    ss << "\\'";
+	  break;
+	case '"':
+	  if (in_string)
+	    ss << "\\\"";
+	  else
+	    ss << c;
+	  break;
+	// Should we also replace unicode escape or \x digits ?
```
CohenArthur, line 214: @tamaroning what do you think?

CohenArthur (approved):
Looks good!

gcc/rust/lex/rust-token.cc
tamaroning: I think we should also convert characters back into unicode escapes and byte literal (\x).
If string literal contains other newlines than \n, dump file is broken. But it requires utf-8 decoding.

gcc/rust/lex/rust-token.cc
P-E-P: Actually I don't know why I kept it like this, even better than a comment I would have used an enum.

P-E-P: Rebased and improved.

gcc/rust/lex/rust-token.cc
```diff
@@ -172,6 +173,43 @@ Token::get_str () const
...
+{
+  std::stringstream ss;
+  decltype (matches)::const_iterator result;
+  for (char c : source)
+    {
+      // FIXME: Also replace escaped unicode values and \x digits
+      if ((result = matches.find (c)) != matches.end ())
+	{
+	  ss << result->second;
+	}
```
CohenArthur, line 199:
```suggestion
	  ss << result->second;
```
nit

gcc/rust/lex/rust-token.cc
```diff
@@ -172,6 +173,43 @@ Token::get_str () const
...
+
+const std::map<char, std::string> matches = {
+  {'\t', "\\t"}, {'\n', "\\n"},	 {'\r', "\\r"},
+  {'\0', "\\0"}, {'\\', "\\\\"}, {'\v', "\\v"},
+};
+
+std::string
+escape_special_chars (const std::string &source, Context ctx)
+{
+  std::stringstream ss;
```
CohenArthur, line 191:
```suggestion
  std::stringstream stream;
```

very small nit, but I always hvae trouble understanding what `ss` refers to while `stream` is clearer IMO

CohenArthur (approved):
I like the enum in the anonymous namespace :+1: LGTM

gcc/rust/lex/rust-token.cc
P-E-P: Fixed!

gcc/rust/lex/rust-token.cc
P-E-P: Fixed!

gcc/rust/lex/rust-token.cc
```diff
@@ -172,6 +173,41 @@ Token::get_str () const
...
+};
+
+std::string
+escape_special_chars (const std::string &source, Context ctx)
+{
+  std::stringstream stream;
+  decltype (matches)::const_iterator result;
+  for (char c : source)
+    {
+      // FIXME: Also replace escaped unicode values and \x digits
```
CohenArthur, line 195: Put in the issue number? last nit and I'll merge this haha

CohenArthur (approved):

gcc/rust/lex/rust-token.cc
P-E-P: Done

PR #2393, opened 2023-07-07
Merge ```Linemap``` and ```Gcc_linemap```

powerboat9:
Depends on #2392

PR #2400, opened 2023-07-10
Merge ```Linemap::to_string``` into ```Linemap::location_to_string```

powerboat9:
Depends on #2393


philberty:
conflicts @powerboat9  otherwise this will be good to go

PR #2403, opened 2023-07-10
Change class RichLocation into typedef

powerboat9:


philberty (approved):
LGTM are you considering getting rid of the typedef to eventually use rich_location directly?

PR #2409, opened 2023-07-11
proc_macro: Add from string implementation

### LINKED ISSUE #973: Allow gccrs to be used as a proc_macro server
https://github.com/rust-lang/rust/blob/c651ba8a542c7d89b271efbf024a31091c824f4b/compiler/rustc_builtin_macros/src/lib.rs#L118-L119

rustc has a special builtin macro (quote) which resolves, if I'm not mistaken, to proc_macro's quote (https://doc.rust-lang.org/proc_macro/fn.quote.html)

We can probably look into the interface required for such a server and add the functionality to gccrs. This probably requires being able to compile the proc_macro crate however

P-E-P:
Add a callback registration symbol into the proc macro library so the compiler can register it's own lexing/parsing functions on load. Register a conversion from string callback in order to implement the `FromStr` trait on `TokenStream`.

Fixes #973

### DIFF
--- a/gcc/rust/expand/rust-proc-macro.cc
+++ b/gcc/rust/expand/rust-proc-macro.cc
@@ -16,6 +16,8 @@
 
 #include "rust-diagnostics.h"
 #include "rust-proc-macro.h"
+#include "rust-lex.h"
+#include "rust-token-converter.h"
 #ifndef _WIN32
 #include <dlfcn.h>
 #endif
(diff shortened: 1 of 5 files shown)


P-E-P:
I've made some changes between fd3026b and bb5f1c6  that I'm not completely happy with. I didn't like the idea of having a global, and now we've got a macro as well. It doesn't seem completely sane.

I was forced to change the from a registration function approach to this direct variable access due to some cast constraint being conditionally supported.

philberty (approved):
Nice one

libgrust/libproc_macro/registration.h
```diff
@@ -0,0 +1,37 @@
...
+
+#ifndef REGISTRATION_H
+#define REGISTRATION_H
+
+#include <string>
+#include "tokenstream.h"
+
+namespace ProcMacro {
+
+using from_str_function_t = ProcMacro::TokenStream (*) (std::string &, bool &);
```
philberty, line 31: Actually is using std::string and references ok here?

I am probably wrong so please correct me, is this the C API that people use? Because std::string and references wont be available in the C ABI

P-E-P: No, we're using the C++ ABI here but I admit it is a little difficult to understand, we're juggling with 3 different ABI at the same time with this library:

- The callback function uses the C++ ABI, We're "communicating" with the C++ libprocmacro internals, not rust, hence why we can use the C++ ABI.
- The symbol containing the address of the callback function uses the C ABI (no mangling)

Does it make sense ? Did I missed a detail ?

philberty: no thats cool

gcc/rust/expand/rust-proc-macro.cc
```diff
@@ -24,6 +26,60 @@ namespace Rust {
...
+    }
+
+  lex_error = false;
+  return convert (tokens);
+}
+
+static_assert (
+  std::is_same<decltype (tokenstream_from_string) *,
+	       ProcMacro::from_str_function_t>::value,
+  "Registration callback signature not synced, check proc macro internals.");
```
CohenArthur, line 57: so does this work in C++11? do we need to add any `#ifdef` guards?

gcc/rust/expand/rust-proc-macro.cc
```diff
@@ -24,6 +26,60 @@ namespace Rust {
...
+
+template <typename Symbol, typename Callback>
+bool
+register_callback (void *handle, Symbol, std::string symbol_name,
```
CohenArthur, line 61: what do we need the `Symbol` variable for?

gcc/rust/expand/rust-proc-macro.cc
P-E-P: Yes! We don't need any `ifdef` guards to make it works.

PR #2412, opened 2023-07-11
Fix several warnings

P-E-P:
Several lines add some warning, it clutters the compiler output, makes error message harder to find/read when compiling with a high number of cores.


philberty (approved):
nice one

PR #2414, opened 2023-07-12
mklog: handle Signed-Off-By, minor cleanup

dkm:
(Has been submitted for review upstream: https://inbox.sourceware.org/gcc-patches/20230707070727.658031-1-dkm@kataplop.net/)

Consider Signed-Off-By lines as part of the ending of the initial
commit to avoid having these in the middle of the log when the
changelog part is injected after.

This is particularly usefull with:

 $ git gcc-commit-mklog --amend -s

that can be used to create the changelog and add the Signed-Off-By line.

Also applies most of the shellcheck suggestions on the
prepare-commit-msg hook.

### DIFF
--- a/contrib/mklog.py
+++ b/contrib/mklog.py
@@ -41,7 +41,34 @@ from unidiff import PatchSet
 
 LINE_LIMIT = 100
 TAB_WIDTH = 8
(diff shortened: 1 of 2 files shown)


CohenArthur (approved):
Thank you!

PR #2416, opened 2023-07-12
rust-unreachable: Add specific behavior for rust_unreachable

CohenArthur:
As discussed during the community call, rust_unreachable() should
always trigger an internal compiler error and should not offer
the behavior of __builtin_unreachable when asserts are disabled, unlike
gcc_unreachable().


philberty (approved):
LGTM nice one :D

P-E-P (approved):

PR #2422, opened 2023-07-12
gccrs: better handling of incorrect file loading in include_* builtin macros

### LINKED ISSUE #2421: Invalid use of empty vector
Enabling libstdc++ assertions shows an invalid use of empty vector:

https://github.com/Rust-GCC/gccrs/blob/b75357f4a1835dcd288df67402340b4cb6b96c12/gcc/rust/expand/rust-macro-builtins.cc#L572

Only 2 tests blowup:
rust/compile/builtin_macro_include_str_location_info.rs
rust/compile/issue-1830_str.rs

on the same assertion:

```
1325   │ 0x7fc7b822b471 __GI_abort
1326   │     ./stdlib/abort.c:79
1327   │ 0x21120a3 std::vector<unsigned char, std::allocator<unsigned char> >::operator[](unsigned long)
1328   │     /usr/include/c++/12/bits/stl_vector.h:1123
1329   │ 0x210d685 Rust::MacroBuiltin::include_str_handler(unsigned int, Rust::AST::MacroInvocData&)
1330   │     ../../gcc/rust/expand/rust-macro-builtins.cc:572
```

This patch fixes the issue, but I'll let @philberty decide:

```diff
diff --git a/gcc/rust/expand/rust-macro-builtins.cc b/gcc/rust/expand/rust-macro-builtins.cc
index 0423e516997..d06056b0134 100644
--- a/gcc/rust/expand/rust-macro-builtins.cc
+++ b/gcc/rust/expand/rust-macro-builtins.cc
@@ -565,7 +565,7 @@ MacroBuiltin::include_str_handler (Location invoc_locus,
     }
 
   std::string str;
-  if (expect_single)
+  if (expect_single || bytes.empty ())
     rust_error_at (invoc_locus, "%s was not a valid utf-8 file",
                   target_filename.c_str ());
   else
```

dkm:
When the load_file_bytes() fails to open the file, it emits an error but
returns an empty vector (as if the file is empty).

Instead, return a tl::optional nullopt in case of error and an empty
vector in case of an empty file.

fixes #2421

gcc/rust/ChangeLog:

	* expand/rust-macro-builtins.cc
	(load_file_bytes): Return a tl::optional.
	(MacroBuiltin::include_byte_handler): Adjust and check file was
	correctly opened.
	(MacroBuiltin::include_str_handler): Likewise.

gcc/testsuite/ChangeLog:

	* rust/compile/empty_file: New test.
	* rust/compile/issue-2421_str.rs: New test.

### DIFF
--- a/gcc/rust/expand/rust-macro-builtins.cc
+++ b/gcc/rust/expand/rust-macro-builtins.cc
@@ -370,14 +370,14 @@ source_relative_path (std::string path, location_t locus)
 /* Read the full contents of the file FILENAME and return them in a vector.
    FIXME: platform specific.  */
 
-std::vector<uint8_t>
+tl::optional<std::vector<uint8_t>>
 load_file_bytes (location_t invoc_locus, const char *filename)
 {
   RAIIFile file_wrap (filename);
   if (file_wrap.get_raw () == nullptr)
     {
       rust_error_at (invoc_locus, "cannot open filename %s: %m", filename);
-      return std::vector<uint8_t> ();
+      return tl::nullopt;
     }
 
   FILE *f = file_wrap.get_raw ();
@@ -387,7 +387,7 @@ load_file_bytes (location_t invoc_locus, const char *filename)
 
   std::vector<uint8_t> buf (fsize);
 
-  if (fread (&buf[0], fsize, 1, f) != 1)
+  if (fsize > 0 && fread (&buf[0], fsize, 1, f) != 1)
     {
       rust_error_at (invoc_locus, "error reading file %s: %m", filename);
       return std::vector<uint8_t> ();
@@ -453,8 +453,12 @@ MacroBuiltin::include_bytes_handler (location_t invoc_locus,
   std::string target_filename
     = source_relative_path (lit_expr->as_string (), invoc_locus);
 
-  std::vector<uint8_t> bytes
-    = load_file_bytes (invoc_locus, target_filename.c_str ());
+  auto maybe_bytes = load_file_bytes (invoc_locus, target_filename.c_str ());
+
+  if (!maybe_bytes.has_value ())
+    return AST::Fragment::create_error ();
+
+  std::vector<uint8_t> bytes = maybe_bytes.value ();
 
   /* Is there a more efficient way to do this?  */
   std::vector<std::unique_ptr<AST::Expr>> elts;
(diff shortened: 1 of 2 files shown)


powerboat9:
Does including an empty file produce an error?

dkm:
> Does including an empty file produce an error?

`rustc` seems to be opening `$CWD` in that case, which is a directory, not a file:

https://rust.godbolt.org/z/odnEqhaWf

Not really sure what is the best error: complaining that a `$CWD` is not a file, or that "" is not a valid file.

I guess rejecting the empty string here would be best...?

powerboat9:
I'm pretty sure ```bytes``` stores the file data, not the file name

dkm:
This fixes the case of including "", but an empty file is badly handled. I'll fix it here, so let's defer this PR until I find a little time to fix it correctly, should not be long.

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -568,8 +576,10 @@ MacroBuiltin::include_str_handler (Location invoc_locus,
   if (expect_single)
     rust_error_at (invoc_locus, "%s was not a valid utf-8 file",
 		   target_filename.c_str ());
-  else
+  else if (!bytes.empty ())
     str = std::string ((const char *) &bytes[0], bytes.size ());
+  else
+    str = "";
```
dkm, line 582: @CohenArthur is that the correct way to "do nothing"? Is there a better way to create an empty Fragment?

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -453,8 +453,12 @@ MacroBuiltin::include_bytes_handler (Location invoc_locus,
...
     = source_relative_path (lit_expr->as_string (), invoc_locus);
 
-  std::vector<uint8_t> bytes
-    = load_file_bytes (invoc_locus, target_filename.c_str ());
+  auto maybe_bytes = load_file_bytes (invoc_locus, target_filename.c_str ());
+
+  if (!maybe_bytes.has_value ())
+    return AST::Fragment::create_error ();
+
+  std::vector<uint8_t> bytes = maybe_bytes.value ();
```
CohenArthur, line 461: it's going to be really cool if this function can return a `tl::optional<AST::Fragment>` because then we'll be able to just `map` it, like *proper functional programmers*!

gcc/rust/expand/rust-macro-builtins.cc
CohenArthur: I think this is the best way. I'll have a think about it but it seems good to me

CohenArthur (approved):

gcc/rust/expand/rust-macro-builtins.cc
dkm: I've experimented with the optional approach and have refactored a bit. Tests are all OK.

gcc/rust/expand/rust-macro-builtins.cc
dkm: hmm, I'll have to read a bit more here, as I'm not sure to follow what changes this would allow

gcc/rust/expand/rust-macro-builtins.cc
dkm: Note that the initial question of "how to create empty fragment" had still to be answered. I've created a `create_empty` next to the error one.

powerboat9: You could try replacing ```(const char *) &bytes[0]``` with ```(const char *) bytes.data()```

CohenArthur (approved):
LGTM!

PR #2435, opened 2023-07-17
cli: Add frust-type option

P-E-P:
Add a new option to crab1 cli to accept crate type. This version of the argument only accept a single crate type. Rustc accepts a comma separated list of crate types but this might require a litle more work for gcc.

### DIFF
--- a/gcc/rust/lang.opt
+++ b/gcc/rust/lang.opt
@@ -82,6 +82,34 @@ frust-max-recursion-depth=
 Rust RejectNegative Type(int) Var(rust_max_recursion_depth) Init(64)
 -frust-max-recursion-depth=<integer>
 
+frust-crate-type=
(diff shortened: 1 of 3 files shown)


gcc/rust/lang.opt
```diff
@@ -82,6 +82,34 @@ frust-max-recursion-depth=
 Rust RejectNegative Type(int) Var(rust_max_recursion_depth) Init(64)
 -frust-max-recursion-depth=<integer>
 
+frust-type=
+Rust Joined RejectNegative Enum(frust_type) Var(flag_rust_type)
```
CohenArthur, line 86:
```suggestion
frust-crate-type=
Rust Joined RejectNegative Enum(frust_crate_type) Var(flag_rust_crate_type)
```

would be clearer I think?

P-E-P: Yeah I agree. I will change that.

PR #2445, opened 2023-07-18
[E0323] Implemented associated const, expected another trait

MahadMuhammad:
### Const is not member of trait [`E0323`](https://doc.rust-lang.org/error_codes/E0323.html) 

- Refactored Error message similiar to rustc.

---

### Code tested from [`E0323`](https://doc.rust-lang.org/error_codes/E0323.html):

```rust
// https://doc.rust-lang.org/error_codes/E0323.html
#![allow(unused)]
fn main() {
trait Foo {
    type N;
}

struct Bar;

impl Foo for Bar {
    const N : u32 = 0; // { dg-error "item .N. is an associated const, which does not match its trait .Foo." }
    // error: item `N` is an associated const, which doesn't match its
    //        trait `<Bar as Foo>`
}
}
```

---

### Output:

```rust
/home/mahad/Desktop/mahad/gccrs/gcc/testsuite/rust/compile/non_member_const.rs:11:5: error: item 'N' is an associated const, which does not match its trait 'Foo' [E0323]
compiler exited with status 1
PASS: rust/compile/non_member_const.rs  (test for errors, line 11)
PASS: rust/compile/non_member_const.rs (test for excess errors)
```

---

gcc/rust/ChangeLog:

	* typecheck/rust-hir-type-check-implitem.cc (TypeCheckImplItemWithTrait::visit): called error function.

gcc/testsuite/ChangeLog:

	* rust/compile/non_member_const.rs: New test.

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-implitem.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-implitem.cc
@@ -389,12 +389,14 @@ TypeCheckImplItemWithTrait::visit (HIR::ConstantItem &constant)
 				       TraitItemReference::TraitItemType::CONST,
 				       &raw_trait_item);
 
-  // unknown trait item
+  // unknown trait item - https://doc.rust-lang.org/error_codes/E0323.html
   if (!found || raw_trait_item->is_error ())
     {
       rich_location r (line_table, constant.get_locus ());
       r.add_range (trait_reference.get_locus ());
-      rust_error_at (r, "constant %<%s%> is not a member of trait %<%s%>",
+      rust_error_at (r, ErrorCode ("E0323"),
+		     "item %qs is an associated const, which does not match "
+		     "its trait %qs",
 		     constant.get_identifier ().as_string ().c_str (),
 		     trait_reference.get_name ().c_str ());
       return;
--- a/gcc/testsuite/rust/compile/non_member_const.rs
+++ b/gcc/testsuite/rust/compile/non_member_const.rs
@@ -1,0 +1,15 @@
+// https://doc.rust-lang.org/error_codes/E0323.html
+#![allow(unused)]
+fn main() {
+trait Foo {
+    type N;
+}
+
+struct Bar;
+
+impl Foo for Bar {
+    const N : u32 = 0; // { dg-error "item .N. is an associated const, which does not match its trait .Foo." }
+    // error: item `N` is an associated const, which doesn't match its
+    //        trait `<Bar as Foo>`
+}
+}


gcc/rust/typecheck/rust-hir-type-check-implitem.cc
```diff
@@ -389,12 +389,14 @@ TypeCheckImplItemWithTrait::visit (HIR::ConstantItem &constant)
...
 
-  // unknown trait item
+  // unknown trait item - https://doc.rust-lang.org/error_codes/E0323.html
   if (!found || raw_trait_item->is_error ())
     {
       rich_location r (line_table, constant.get_locus ());
       r.add_range (trait_reference.get_locus ());
-      rust_error_at (r, "constant %<%s%> is not a member of trait %<%s%>",
+      rust_error_at (r, ErrorCode ("E0323"),
+		     "item %<%s%> is an associated const, which does not match "
```
MahadMuhammad, line 398: - Refactored Error message, similiar to rustc.
- With a little difference, 
- The diff of rustc and gccrs error is:

```diff
# gccrs
+ item ‘N’ is an associated const, which does not match its trait ‘Foo’  

- item `N` is an associated const, which doesn't match its trait `Foo`
```

Is there a way to add single apostrophe, in the error message:

```bash
../../gccrs/gcc/rust/typecheck/rust-hir-type-check-implitem.cc:398:68: warning: bare apostrophe ‘'’ in format [-Wformat-diag]
  398 |                      "item %<%s%> is an associated const, which doe's not match "
      |                                                                    ^
../../gccrs/gcc/rust/typecheck/rust-hir-type-check-implitem.cc:398:22: note: if avoiding the apostrophe is not feasible, enclose it in a pair of ‘%<’ and ‘%>’ directives instead
  398 |                      "item %<%s%> is an associated const, which doe's not match "
      |                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  399 |                      "its trait %<%s%>",
      |                      ~~~~~~~~~~~~~~~~~~

```

CohenArthur: the grammar you used is correct, no need to add a single apostrophe?

CohenArthur, line 398:
```suggestion
		     "item %qs is an associated const, which does not match "
```
same formatting string and easier to use and read

gcc/rust/typecheck/rust-hir-type-check-implitem.cc
```diff
@@ -389,12 +389,14 @@ TypeCheckImplItemWithTrait::visit (HIR::ConstantItem &constant)
...
+		     "its trait %<%s%>",
```
CohenArthur, line 399:
```suggestion
		     "its trait %qs",
```
same here

gcc/rust/typecheck/rust-hir-type-check-implitem.cc
```diff
@@ -389,12 +389,14 @@ TypeCheckImplItemWithTrait::visit (HIR::ConstantItem &constant)
 				       TraitItemReference::TraitItemType::CONST,
 				       &raw_trait_item);
 
-  // unknown trait item
+  // unknown trait item - https://doc.rust-lang.org/error_codes/E0323.html
   if (!found || raw_trait_item->is_error ())
     {
       rich_location r (line_table, constant.get_locus ());
       r.add_range (trait_reference.get_locus ());
-      rust_error_at (r, "constant %<%s%> is not a member of trait %<%s%>",
```
CohenArthur, line 397: I liked this error message a little more personally :D but that's just nitpicking haha

gcc/rust/typecheck/rust-hir-type-check-implitem.cc
MahadMuhammad: @CohenArthur, Yeah there is no need to add a single apostrophe. But it would be better to add, then the whole message will be exactly similiar to rustc..

CohenArthur (approved):

MahadMuhammad:
@CohenArthur We accidently merged wrong error message & error code in this PR :(

The error code should be [`E0438`](https://doc.rust-lang.org/error_codes/E0438.html) and previous message was also wrong. For this error code [`E0323`](https://doc.rust-lang.org/error_codes/E0323.html), we need to do some extra work.

CohenArthur:
> @CohenArthur We accidently merged wrong error message & error code in this PR :(
> 
> The error code should be [`E0438`](https://doc.rust-lang.org/error_codes/E0438.html) and previous message was also wrong. For this error code [`E0323`](https://doc.rust-lang.org/error_codes/E0323.html), we need to do some extra work. I'll fix this in some other PR :)

that's okay, good catch! can you open an issue so we don't forget to address it?

PR #2456, opened 2023-07-20
name resolution 2.0: Add base Resolver and TopLevel pass

CohenArthur:
- name resolution 2.0: Add base for Resolver data structure
- top-level: Add base `TopLevel` visitor

This commit adds a first simple `Resolver` data structure, which will
be used by all passes of the name resolution to store definitions and access
them. Namely, it contains a few instances of `ForeverStack`, for each
namespace we will be using.

The `TopLevel` pass takes care of collecting definitions, placing them
in the proper namespaces, and making them accessible for later resolution
passes like `Early` and `Late`. It is meant to be run in a fixed point
fashion, as import resolution, macro resolution and macro expansion
may generate multiple new definitions.

### DIFF
--- a/gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
@@ -1,0 +1,212 @@
+// Copyright (C) 2020-2023 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#include "rust-toplevel-name-resolver-2.0.h"
+#include "rust-ast-full.h"
+#include "rust-hir-map.h"
+
+namespace Rust {
+namespace Resolver2_0 {
+
+TopLevel::TopLevel (NameResolutionContext &resolver) : ctx (resolver) {}
+
+template <typename T>
+void
+TopLevel::insert_or_error_out (const Identifier &identifier, const T &node,
+			       Namespace ns)
+{
+  auto loc = node.get_locus ();
+  auto node_id = node.get_node_id ();
+
+  // keep track of each node's location to provide useful errors
+  node_locations.emplace (node_id, loc);
+
+  auto result = ctx.insert (identifier, node_id, ns);
+
+  if (!result)
+    {
+      rich_location rich_loc (line_table, loc);
+      rich_loc.add_range (node_locations[result.error ().existing]);
+
+      rust_error_at (rich_loc, "already defined");
+    }
+}
+
+void
+TopLevel::go (AST::Crate &crate)
+{
+  for (auto &item : crate.items)
+    item->accept_vis (*this);
+}
+
+void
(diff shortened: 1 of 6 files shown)


CohenArthur:
Noticed a couple issues already so putting it in draft, but reviews welcome!

gcc/rust/resolve/rust-name-resolver-2.0.h
```diff
@@ -0,0 +1,181 @@
...
+  // scoped lambda?
+  void scoped (Rib rib, NodeId scope_id, std::function<void (void)> lambda,
+	       tl::optional<Identifier> path = {});
+  void scoped (Rib rib, Namespace ns, NodeId scope_id,
+	       std::function<void (void)> lambda,
+	       tl::optional<Identifier> path = {});
+
+  ForeverStack<Namespace::Values> values;
+  ForeverStack<Namespace::Types> types;
+  ForeverStack<Namespace::Macros> macros;
```
P-E-P, line 175: Should'nt those be private ? I thought the `insert` function would ensure a correct usage.

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
```diff
@@ -0,0 +1,215 @@
...
+
+  resolver.scoped (Rib::Kind::Function, function.get_node_id (), def_fn);
+}
+
+void
+TopLevel::visit (AST::Method &method)
+{
+  insert_or_error_out (method.get_method_name (), method, Namespace::Values);
+
+  method.get_definition ()->accept_vis (*this);
```
CohenArthur, line 97:
```suggestion
  auto def_fn
    = [this, &method] () { method.get_definition ()->accept_vis (*this); };

  resolver.scoped (Rib::Kind::Function, method.get_node_id (), def_fn);
```

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
```diff
@@ -0,0 +1,215 @@
...
+  node_locations.emplace (node_id, loc);
+
+  auto result = resolver.insert (identifier, node_id, ns);
+
+  if (!result)
+    {
+      rich_location rich_loc (line_table, loc);
+      rich_loc.add_range (node_locations[result.error ().existing]);
+
+      rust_error_at (rich_loc, "already defined", result.error ().name.c_str (),
```
P-E-P, line 46: missing format string arguments ?

gcc/rust/resolve/rust-name-resolver-2.0.h
CohenArthur: ah, yes they should be private - I had taken them out for debugging purposes

CohenArthur: after adding more to the Early name resolver I wonder if it doesn't make sense to keep these as public

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
CohenArthur: Fixed in #2469 so I'll leave it here

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
CohenArthur: Fixed

gcc/rust/resolve/rust-name-resolver-2.0.h
CohenArthur: we'll investigate in following PRs, but for now I think we should keep it public

gcc/rust/resolve/rust-name-resolver-2.0.cc
```diff
@@ -0,0 +1,98 @@
...
+      break;
+    case Namespace::Types:
+      types.push (rib, scope_id, path);
+      break;
+    case Namespace::Labels:
+    case Namespace::Macros:
+      gcc_unreachable ();
+    }
+
+  // for (auto &ns : namespaces)
```
P-E-P, line 75: Deadcode ?

PR #2462, opened 2023-07-20
Prevent multiple bootstrap builds from running at once

powerboat9:
This should prevent the github-provided runners from spending all their time on bootstrap builds


CohenArthur (approved):
Nice! I think this is a good idea

PR #2465, opened 2023-07-21
Proc macro non function errors

### LINKED ISSUE #2459: Emit error with `#[proc_macro]`, `#[proc_macro_derive]` or `#[proc_macro_attribute]` on non function items
Emit errors when such attribute is found on every kind of item/statement that is not a function.

```
the `#[proc_macro]` attribute may only be used on bare functions
```

P-E-P:
Emit an error when a procedural macro attribute is placed on a non function item.

Fixes #2459


CohenArthur (approved):
LGTM!

PR #2470, opened 2023-07-26
nr2.0: Add base for Early name resolution

CohenArthur:
This adds a base visitor for resolving macro invocations in the new name resolution 2.0 algorithm. Needs #2469 and #2456

### DIFF
--- a/gcc/rust/resolve/rust-default-resolver.cc
+++ b/gcc/rust/resolve/rust-default-resolver.cc
@@ -303,15 +303,25 @@ DefaultResolver::visit (AST::StructExprFieldIndexValue &)
 
 void
 DefaultResolver::visit (AST::CallExpr &expr)
-{}
+{
+  expr.get_function_expr ()->accept_vis (*this);
+  for (auto &param : expr.get_params ())
+    param->accept_vis (*this);
+}
 
 void
 DefaultResolver::visit (AST::MethodCallExpr &expr)
-{}
+{
+  expr.get_receiver_expr ()->accept_vis (*this);
+  for (auto &param : expr.get_params ())
+    param->accept_vis (*this);
+}
 
 void
 DefaultResolver::visit (AST::FieldAccessExpr &expr)
-{}
+{
+  expr.get_receiver_expr ()->accept_vis (*this);
+}
 
 void
 DefaultResolver::visit (AST::ClosureExprInner &)
--- a/gcc/rust/resolve/rust-default-resolver.h
+++ b/gcc/rust/resolve/rust-default-resolver.h
@@ -56,7 +56,7 @@ public:
   void visit (AST::Union &);
   void visit (AST::Enum &);
 
-  // Visitors that visit their expression node
+  // Visitors that visit their expression node(s)
   void visit (AST::BorrowExpr &);
   void visit (AST::DereferenceExpr &);
   void visit (AST::ErrorPropagationExpr &);
--- a/gcc/rust/resolve/rust-early-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-early-name-resolver-2.0.cc
@@ -1,0 +1,164 @@
+// Copyright (C) 2020-2023 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#include "rust-early-name-resolver-2.0.h"
+#include "rust-ast-full.h"
+#include "rust-toplevel-name-resolver-2.0.h"
+
+namespace Rust {
+namespace Resolver2_0 {
+
+Early::Early (NameResolutionContext &ctx) : DefaultResolver (ctx) {}
+
+void
+Early::go (AST::Crate &crate)
+{
+  // First we go through TopLevel resolution to get all our declared items
+  auto toplevel = TopLevel (ctx);
+  toplevel.go (crate);
+
+  textual_scope.push ();
+
+  // Then we proceed to the proper "early" name resolution: Import and macro
+  // name resolution
+  for (auto &item : crate.items)
+    item->accept_vis (*this);
+
+  textual_scope.pop ();
+}
+
+void
+Early::TextualScope::push ()
+{
+  // push a new empty scope
+  scopes.emplace_back ();
+}
+
+void
+Early::TextualScope::pop ()
+{
+  rust_assert (!scopes.empty ());
+
+  scopes.pop_back ();
+}
+
+void
+Early::TextualScope::insert (std::string name, NodeId id)
+{
+  rust_assert (!scopes.empty ());
+
+  // we can ignore the return value as we always want the latest defined macro
+  // to shadow a previous one - so if two macros have the same name and get
+  // inserted with the same key, it's not a bug
+  scopes.back ().insert ({name, id});
+}
+
+tl::optional<NodeId>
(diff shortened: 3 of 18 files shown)


CohenArthur:
will drop annoying commits from this PR and #2469 tomorrow, just running builds and testsuite on all commits here and can't really modify the branch while doing a `git rebase upstream/master -x ...`

gcc/rust/resolve/rust-default-resolver.cc
```diff
@@ -0,0 +1,805 @@
...
+
+void
+DefaultResolver::visit (AST::StructStruct &type)
+{
+  // do we need to scope anything here? no, right?
+
+  // we also can't visit `StructField`s by default, so there's nothing to do -
+  // correct? or should we do something like
+
+  for (auto &field : type.get_fields ())
```
P-E-P, line 150: Should we really visit those struct ?
```rust
struct Thing {
    a : usize,
}
```
`a` being resolved like any other name make sense in some way but it feels weird ?

gcc/rust/resolve/rust-default-resolver.h
```diff
@@ -0,0 +1,214 @@
...
+namespace Resolver2_0 {
+
+// FIXME: Documentation
+// does nothing for most nodes
+// apart from "scope" nodes for which it visits each member, in a name
+// resolution fashion
+class DefaultResolver : public AST::ASTVisitor
+{
+public:
+  virtual ~DefaultResolver () {} // FIXME: do we need this?
```
P-E-P, line 36: Yes, as it is not final.

gcc/rust/resolve/rust-default-resolver.h
```diff
@@ -0,0 +1,214 @@
...
+
+  // FIXME: Do we need this?
+  // is there an easy way to overload the visitors we don't need in a derived
```
P-E-P, line 39: I'm not sure I understand this comment. Could you clarify please ?

CohenArthur: that was me thinking while writing hoping that we could do something like `using visit(UnusedNode&) = do_nothing;` but it's stupid. I'll remove the comment and function

gcc/rust/resolve/rust-default-resolver.cc
```diff
@@ -0,0 +1,805 @@
...
+void
+DefaultResolver::visit (AST::ClosureExprInnerTyped &)
+{}
+
+void
+DefaultResolver::visit (AST::ContinueExpr &expr)
+{}
+
+void
+DefaultResolver::visit (AST::BreakExpr &expr)
```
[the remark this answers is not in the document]
CohenArthur: yes!

gcc/rust/resolve/rust-default-resolver.cc
```diff
@@ -0,0 +1,805 @@
...
+DefaultResolver::visit (AST::StructExprFieldIdentifierValue &)
+{}
+
+void
+DefaultResolver::visit (AST::StructExprFieldIndexValue &)
+{}
+
+void
+DefaultResolver::visit (AST::CallExpr &expr)
+{}
```
matthewjasper, line 306: Should everything be visiting children here? The classes derived from this are presumably also going to be handling uses,

gcc/rust/resolve/rust-early-name-resolver-2.0.cc
```diff
@@ -0,0 +1,168 @@
...
+    }
+  else
+    {
+      definition = resolver.macros.resolve_path (path);
+    }
+
+  if (!definition.has_value ())
+    {
+      rust_error_at (invoc.get_locus (), ErrorCode ("E0433"),
+		     "could not resolve macro invocation");
```
matthewjasper, line 143: This will need to be changed for macros that expand to macro definitions.

gcc/rust/resolve/rust-early-name-resolver-2.0.cc
```diff
@@ -0,0 +1,168 @@
...
+void
+Early::visit (AST::MacroRulesDefinition &def)
+{
+  DefaultResolver::visit (def);
+
+  textual_scope.insert (def.get_rule_name ().as_string (), def.get_node_id ());
+}
+
+void
+Early::visit (AST::BlockExpr &block)
```
matthewjasper, line 101: Am I missing where modules are handled?

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
```diff
@@ -0,0 +1,245 @@
...
+      return true;
+
+  return false;
+}
+
+void
+TopLevel::visit (AST::MacroRulesDefinition &macro)
+{
+  // we do not insert macros in the current rib as that needs to be done in the
+  // textual scope of the Early pass. we only insert them in the root of the
```
matthewjasper, line 87: `macro` definitions (which libcore uses for some built in macros) are inserted in module scopes,

gcc/rust/resolve/rust-name-resolver-2.0.h
```diff
@@ -0,0 +1,181 @@
...
+    TODO: Mention macro hygiene and that it is the same
+    TODO: How does this affect our data structures?
+    TODO: Last challenge - reuse the same APIs to allow the typechecker to not
+change?
+    TODO: Mention that ForeverStack is templated to make sure that behavior is
+correct
+*/
+
+// Now our resolver, which keeps track of all the `ForeverStack`s we could want
+class Resolver
```
matthewjasper, line 136: Having a class called `Resolver` that does not inherit from `DefaultResolver` is a bit confusing.

gcc/rust/resolve/rust-early-name-resolver-2.0.cc
CohenArthur: Modules are handled directly in the `TopLevel` visitor - I don't think we have anything to do in the `Early` visitor other than resolve `use` statements to the proper node

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
CohenArthur: grrrrr you're right. I hate macro name resolution rules lol. thank you! I'll take care of this

gcc/rust/resolve/rust-name-resolver-2.0.h
CohenArthur: That's a good point. I'll rename it to `ResolverContext`, which is different from what we had but makes more sense IMO

gcc/rust/resolve/rust-early-name-resolver-2.0.cc
matthewjasper: They need their own textual scopes, for cases like this

```rust
mod a {         // No #[macro_use]
    macro_rules! mac {
        () => {}
    }
    mac!();     // OK
}
mac!();         // ERROR - `mac` not in scope
```

CohenArthur: just realized that's the line you were pointing to lol. sorry I'm super tired haha

gcc/testsuite/rust/compile/name_resolution6.rs
```diff
@@ -0,0 +1,42 @@
+// { dg-options "-frust-name-resolution-2.0" }
```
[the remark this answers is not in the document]
CohenArthur: expansion should not change after this new name resolution algorithm, so I plan to just reuse the existing expansion testsuite when we enable name resolution 2.0 by default. I will split up the test in simpler ones though, you are completely right

gcc/rust/resolve/rust-default-resolver.cc
CohenArthur: yes, this should visit the called expression as well as each given argument and generic arguments if any - good catch

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -97,6 +99,12 @@ ForeverStack<N>::pop ()
   update_cursor (cursor ().parent.value ());
 }
 
+static tl::expected<NodeId, DuplicateNameError>
+insert_inner (Rib &rib, std::string name, NodeId node, bool can_shadow)
+{
```
P-E-P, line 104: I'm a bit confused by this. Is this function missing anything? Should'nt we keep the name somewhere ? If no, why keeping it in the parameters ?

P-E-P (approved):
LGTM (once CI succeed)

matthewjasper (approved):

PR #2471, opened 2023-07-26
Proc macro non root level verification

### LINKED ISSUE #2460: Emit errors when a `#[proc_macro]` is not declared at root level.
Emit error when a proc macro is not declared at the crate's root level.

```
functions tagged with `#[proc_macro]` must currently reside in the root of the crate
```

P-E-P:
This pr add a verification for proc macro attributes on non root level functions.

Requires #2465

Fixes #2460


CohenArthur (approved):
Looks good!

PR #2472, opened 2023-07-26
Increase FFIString usage

P-E-P:
Two remaining structures in the rust interface were still using raw string pointer and length to communicate with the C++ library throught extern C functions. Using FFIString instead allow us to reduce the scope of potential errors using those raw pointers.


CohenArthur (approved):
LGTM!

PR #2475, opened 2023-07-26
Add `is_available` function implementation

### LINKED ISSUE #2474: Implement proc macros' `is_available` function
is available function is not implement in C++ library. Now that we have callback capabilities we may update a value on load.

P-E-P:
Add is_available function implementation to assert the bridge is connected and the proc macro was not wildly linked to the current program.

Fixes #2474

### DIFF
--- a/libgrust/libproc_macro/registration.h
+++ b/libgrust/libproc_macro/registration.h
@@ -30,11 +30,10 @@
 namespace ProcMacro {
 
 using from_str_function_t = ProcMacro::TokenStream (*) (std::string &, bool &);
-using is_available_t = ProcMacro::BridgeState;
 
 } // namespace ProcMacro
 
 extern "C" ProcMacro::from_str_function_t __gccrs_proc_macro_from_str_fn;
-extern "C" ProcMacro::is_available_t __gccrs_proc_macro_is_available_fn;
+extern "C" ProcMacro::BridgeState __gccrs_proc_macro_is_available_fn;
 
 #endif /* !REGISTRATION_H */
--- a/libgrust/libproc_macro/proc_macro.cc
+++ b/libgrust/libproc_macro/proc_macro.cc
@@ -60,5 +60,5 @@ bridge_is_available ()
 } // namespace ProcMacro
 
 ProcMacro::from_str_function_t __gccrs_proc_macro_from_str_fn = nullptr;
-ProcMacro::is_available_t __gccrs_proc_macro_is_available_fn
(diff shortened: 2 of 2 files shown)


libgrust/libproc_macro/bridge.h
```diff
@@ -0,0 +1,35 @@
...
+// 3.1, as published by the Free Software Foundation.
+
+// You should have received a copy of the GNU General Public License and
+// a copy of the GCC Runtime Library Exception along with this program;
+// see the files COPYING3 and COPYING.RUNTIME respectively.  If not, see
+// <http://www.gnu.org/licenses/>.
+#ifndef BRIDGE_H
+#define BRIDGE_H
+
+namespace ProcMacro {
```
CohenArthur, line 25:
```suggestion

namespace Rust {
namespace ProcMacro {
```

or no? I can't remember haha

libgrust/libproc_macro/registration.h
```diff
@@ -25,13 +25,16 @@
 
 #include <string>
 #include "tokenstream.h"
+#include "bridge.h"
 
 namespace ProcMacro {
 
 using from_str_function_t = ProcMacro::TokenStream (*) (std::string &, bool &);
+using is_available_t = ProcMacro::BridgeState;
```
CohenArthur, line 33: I'm not sure we really need this `using` directive. Why not directly use `ProcMacro::BridgeState`?

CohenArthur (approved):
tiny nit but looks good otherwise!

libgrust/libproc_macro/registration.h
P-E-P: You're right, this line is a remnant from a first version of the code which used a function pointer much like from_str callback instead of a value.

libgrust/libproc_macro/bridge.h
P-E-P: No, proc macros never used any `Rust` parent namespace. Do you think it would be better with one ?

CohenArthur: nope, that;s all good for me, just checking!

libgrust/libproc_macro/registration.h
P-E-P: Fixed in [093ac28](https://github.com/Rust-GCC/gccrs/pull/2475/commits/093ac288152f11ed18391b19d820bc7ee5ef0414).

CohenArthur (approved):

PR #2476, opened 2023-07-27
Proc macro literal from string

### LINKED ISSUE #2473: Implement `Literal::from_str`
The implementation of this function is missing in the C++ proc_macro library.

P-E-P:
This function did not have an implementation, it was aborting instead. The new callback system allow us to implement this function easily.

Fixes #2473


CohenArthur (approved):
LGTM

PR #2480, opened 2023-07-28
error codes: Add rust_error_codes.def, refactor ErrorCode enum

CohenArthur:
gcc/rust/ChangeLog:

	* rust-diagnostics.cc: Rework `make_description` and `make_url` functions.
	* rust-diagnostics.h: Specify ErrorCode's underlying type,
	remove error_codes_strings table.
	* rust_error_codes.def: New file.

@mahadmuhammad please review and let me know what you think :)


philberty (approved):
Neat defining them like that

gcc/rust/rust-diagnostics.h
```diff
@@ -48,1047 +48,33 @@
...
+// E0009 = (10009 - 10000)
+//
+// which gets folded to the result of the substraction... 9. A valid decimal
+// literal which corresponds to E0009.
+#define ERROR(N) E##N = (1##N - 10000)
+enum class ErrorCode : unsigned int
+{
+#include "rust_error_codes.def"
 };
+#undef ERROR
```
MahadMuhammad, line 75: Sorry for late review, LGTM , that's a great logic  for converting `enum` to string by assigning them the same non-negative numbers. 

I am a bit sad for one thing that we no longer get the code completion for enum class, but that would be fine :)

Issue #2426, opened 2023-07-14
Get rid of GCC Backend abstraction


philberty:
As we are currently getting rid of the Location, RichLocation and Linemap GCC abstractions we should continue this by getting rid of the backend abstracts in:

See:

https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/rust-backend.h
https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/rust-gcc.cc

This is pointless now and we should create a new file of helpers in:

rust/backend/rust-gimple.cc

They should just be static functions but this will be easier to do with a series of patches

philberty:
@powerboat9 you will probably be interested in this one its been a pet peeve of the community for quite some time this abstraction

PR #2485, opened 2023-07-29
Move ```Gcc_backend``` declaration to header file

powerboat9:
Addresses #2426


philberty (approved):
Good idea to start doing this

PR #2440, opened 2023-07-18
Remove ```Gcc_backend```

powerboat9:

PR #2491, opened 2023-07-31
Move ```Backend::typed_identifier``` out of class

powerboat9:


CohenArthur (changes requested):
Why move the struct outside of the `Backend` class? The only difference is that it is no longer namespaced, correct? I think using `Backend::typed_identifier` is clearer, personally

powerboat9:
I was thinking I would fold ```Backend``` into ```Gcc_backend```, but [folding ```Gcc_backend``` into ```Backend```](https://github.com/Rust-GCC/gccrs/pull/2440#pullrequestreview-1557253542) seems like it works much better. As such, this pull request is obsolete.

PR #2496, opened 2023-07-31
Name resolve proc macros

P-E-P:
Requires #2470

### DIFF
--- a/gcc/rust/expand/rust-proc-macro.cc
+++ b/gcc/rust/expand/rust-proc-macro.cc
@@ -18,11 +18,31 @@
 #include "rust-proc-macro.h"
 #include "rust-lex.h"
 #include "rust-token-converter.h"
+#include "rust-attributes.h"
 #ifndef _WIN32
 #include <dlfcn.h>
 #endif
 
 namespace Rust {
+
+BangProcMacro::BangProcMacro (ProcMacro::Bang macro)
+  : name (macro.name),
+    node_id (Analysis::Mappings::get ()->get_next_node_id ()),
+    macro (macro.macro)
+{}
+
+AttributeProcMacro::AttributeProcMacro (ProcMacro::Attribute macro)
+  : name (macro.name),
+    node_id (Analysis::Mappings::get ()->get_next_node_id ()),
+    macro (macro.macro)
+{}
+
+CustomDeriveProcMacro::CustomDeriveProcMacro (ProcMacro::CustomDerive macro)
+  : trait_name (macro.trait_name),
+    attributes (macro.attributes, macro.attributes + macro.attr_size),
+    node_id (Analysis::Mappings::get ()->get_next_node_id ()),
+    macro (macro.macro)
+{}
 
 const std::string PROC_MACRO_DECL_PREFIX = "__gccrs_proc_macro_decls_";
 
--- a/gcc/rust/expand/rust-proc-macro.h
+++ b/gcc/rust/expand/rust-proc-macro.h
@@ -18,8 +18,65 @@
 #define RUST_PROC_MACRO_H
 
 #include "libproc_macro_internal/proc_macro.h"
+#include "rust-mapping-common.h"
 
 namespace Rust {
+
+class BangProcMacro
+{
+private:
+  std::string name;
+  NodeId node_id;
+  ProcMacro::BangMacro macro;
+
+public:
+  BangProcMacro (ProcMacro::Bang macro);
+  BangProcMacro () = default;
+
+  const std::string &get_name () const { return name; }
+
+  NodeId get_node_id () const { return node_id; }
+
+  ProcMacro::BangMacro get_handle () const { return macro; }
+};
+
+class AttributeProcMacro
+{
+private:
+  std::string name;
+  NodeId node_id;
+  ProcMacro::AttributeMacro macro;
+
+public:
+  AttributeProcMacro (ProcMacro::Attribute macro);
+  AttributeProcMacro () = default;
+
+  const std::string &get_name () const { return name; }
+
+  NodeId get_node_id () const { return node_id; }
+
+  ProcMacro::AttributeMacro get_handle () const { return macro; }
+};
+
+class CustomDeriveProcMacro
+{
+private:
+  std::string trait_name;
+  std::vector<std::string> attributes;
+  NodeId node_id;
+  ProcMacro::CustomDeriveMacro macro;
+
+public:
+  CustomDeriveProcMacro (ProcMacro::CustomDerive macro);
+  CustomDeriveProcMacro () = default;
+
+  const std::string &get_trait_name () const { return trait_name; }
+
+  NodeId get_node_id () const { return node_id; }
+
+  ProcMacro::CustomDeriveMacro get_handle () const { return macro; }
+};
+
 /**
  * Load a procedural macro library and collect its entrypoints.
  *
--- a/gcc/rust/expand/rust-proc-macro.h
+++ b/gcc/rust/expand/rust-proc-macro.h
@@ -70,7 +70,7 @@ public:
   CustomDeriveProcMacro (ProcMacro::CustomDerive macro);
   CustomDeriveProcMacro () = default;
 
-  const std::string &get_trait_name () const { return trait_name; }
+  const std::string &get_name () const { return trait_name; }
 
   NodeId get_node_id () const { return node_id; }
 
--- a/gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
@@ -71,6 +71,23 @@ TopLevel::visit (AST::Module &module)
 	      module.get_name ());
 }
 
+template <typename PROC_MACRO>
+static void
+insert_macros (std::vector<PROC_MACRO> &macros, NameResolutionContext &ctx)
+{
+  for (auto &macro : macros)
+    {
+      auto res = ctx.macros.insert (macro.get_name (), macro.get_node_id ());
+
+      if (!res)
+	{
+	  rust_error_at (UNKNOWN_LOCATION, ErrorCode::E0428,
+			 "macro %qs defined multiple times",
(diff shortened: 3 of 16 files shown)


gcc/rust/expand/rust-proc-macro.cc
```diff
@@ -18,12 +18,32 @@
...
+
+AttributeProcMacro::AttributeProcMacro (ProcMacro::Attribute macro)
+  : name (macro.name),
+    node_id (Analysis::Mappings::get ()->get_next_node_id ()),
+    macro (macro.macro)
+{}
+
+CustomDeriveProcMacro::CustomDeriveProcMacro (ProcMacro::CustomDerive macro)
+  : trait_name (macro.trait_name),
+    attributes (macro.attributes, macro.attributes + macro.attr_size),
```
CohenArthur, line 42: what does this addition do?

gcc/rust/expand/rust-proc-macro.h
```diff
@@ -18,8 +18,66 @@
...
+{
+private:
+  std::string name;
+  NodeId node_id;
+  ProcMacro::BangMacro macro;
+
+public:
+  BangProcMacro (ProcMacro::Bang macro);
+  BangProcMacro () = default;
+  ;
```
CohenArthur, line 35: extra semicolon

CohenArthur (commented):
I'll review the PR in details once the Early name resolver will have been merged, it's a lot of changes at the moment (my fault)

gcc/rust/expand/rust-proc-macro.cc
CohenArthur: no, that's okay - I get why you added these classes. I meant, what does `macro.attributes, macro.attributes + macro.attr_size` do when creating the vector? :D

P-E-P: Oh, that's for creating a vector from a pointer array, it copies every element into the vector.

Constructor (5) from the [cpp reference documentation](https://en.cppreference.com/w/cpp/container/vector/vector)

```
template< class InputIt >
vector( InputIt first, InputIt last,
        const Allocator& alloc = Allocator() );
```

gcc/rust/expand/rust-proc-macro.h
P-E-P: Rebased and fixed.

gcc/rust/util/rust-hir-map.cc
```diff
@@ -963,52 +963,44 @@ Mappings::insert_bang_proc_macros (CrateNum num,
...
+Mappings::lookup_bang_proc_macros (CrateNum num)
 {
   auto it = procmacrosBangMappings.find (num);
   if (it == procmacrosBangMappings.end ())
-    return false;
-
-  macros = it->second;
-  return true;
+    return tl::nullopt;
+  return it->second;
```
CohenArthur, line 990: nit
```suggestion
    return tl::nullopt;
    
  return it->second;
```

gcc/rust/util/rust-hir-map.cc
```diff
@@ -963,52 +963,44 @@ Mappings::insert_bang_proc_macros (CrateNum num,
...
+Mappings::lookup_attribute_proc_macros (CrateNum num)
 {
   auto it = procmacrosAttributeMappings.find (num);
   if (it == procmacrosAttributeMappings.end ())
-    return false;
-
-  macros = it->second;
-  return true;
+    return tl::nullopt;
+  return it->second;
```
CohenArthur, line 999: same here

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
```diff
@@ -71,6 +71,68 @@ TopLevel::visit (AST::Module &module)
 	      module.get_name ());
 }
 
+void
+TopLevel::visit (AST::ExternCrate &crate)
+{
+  CrateNum num;
+  if (!Analysis::Mappings::get ()->lookup_crate_name (
+	crate.get_referenced_crate (), num))
+    rust_unreachable ();
```
CohenArthur, line 80:
```suggestion
  rust_assert (Analysis::Mappings::get ()->lookup_crate_name (
	crate.get_referenced_crate (), num));
```

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
```diff
@@ -71,6 +71,68 @@ TopLevel::visit (AST::Module &module)
...
+
+  auto attribute_macros
+    = Analysis::Mappings::get ()->lookup_attribute_proc_macros (num);
+
+  auto bang_macros = Analysis::Mappings::get ()->lookup_bang_proc_macros (num);
+
+  auto derive_macros
+    = Analysis::Mappings::get ()->lookup_derive_proc_macros (num);
```
CohenArthur, line 88: I think we'll keep a `mappings` member in the `TopLevel` resolver too, so this might end up being simpler soon

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
```diff
@@ -71,6 +71,68 @@ TopLevel::visit (AST::Module &module)
...
+	    }
+	}
+  };
+
+  if (crate.has_as_clause ())
+    ctx.scoped (Rib::Kind::Module, crate.get_node_id (), sub_visitor,
+		crate.get_as_clause ());
+  else
+    ctx.scoped (Rib::Kind::Module, crate.get_node_id (), sub_visitor,
+		crate.get_referenced_crate ());
```
CohenArthur, line 133: very cool haha

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
```diff
@@ -71,6 +71,68 @@ TopLevel::visit (AST::Module &module)
...
+      for (auto &bang : bang_macros.value ())
+	{
+	  auto res = ctx.macros.insert (bang.get_name (), bang.get_node_id ());
+	  if (!res)
+	    {
+	      rust_error_at (UNKNOWN_LOCATION, ErrorCode::E0428,
+			     "macro %qs defined multiple times",
+			     bang.get_name ().c_str ());
+	    }
+	}
```
CohenArthur, line 125: I think we can factor this a little better since all the error messages are the same. Maybe a generic function which you call on `derive_macros`, `attribute_macros` and `bang_macros`?

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
P-E-P: We probably should use another PR for this change right ?

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
P-E-P: I'm not sure what you mean here. Should I create a function specifically to emit those errors ?

gcc/rust/util/rust-hir-map.cc
P-E-P: Fixed in 05159dbe21c

gcc/rust/util/rust-hir-map.cc
P-E-P: Fixed in 05159dbe21c.

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
P-E-P: Fixed in 05159dbe21c.

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
P-E-P: Fixed in c085260734e.

gcc/rust/rust-session-manager.cc
```diff
@@ -1010,19 +1010,24 @@ Session::load_extern_crate (const std::string &crate_name, location_t locus)
...
+  if (s.first != nullptr)
     {
-      rust_error_at (locus, "failed to load crate metadata");
-      return UNKNOWN_NODEID;
+      bool ok = extern_crate.load (locus);
+      if (!ok)
+	{
+	  rust_error_at (locus, "failed to load crate metadata");
+	  return UNKNOWN_NODEID;
+	}
```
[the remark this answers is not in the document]
P-E-P: What is bothering you in this part  ?

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
CohenArthur: yes, correct

gcc/rust/rust-session-manager.cc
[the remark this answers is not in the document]
CohenArthur: nothing, just asking for Philip's opinion since he's the one who wrote the original code :D seems perfectly fine to me!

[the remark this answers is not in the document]
philberty: I think the logic is ok but i dont like the variable name s. It took me a few times to read, so s is like the proc macro crate you're trying to load. So i would say something like proc_macro_import_name or something.

philberty (changes requested):
LGTM just change that variable name.

P-E-P:
> LGTM just change that variable name.
> I think the logic is ok but i dont like the variable name s. It took me a few times to read, so s is like the proc macro crate you're trying to load. So i would say something like proc_macro_import_name or something.

Actually it is a bit different, the package opening function knows where to look for an extern crate. For now proc macros are simple share libraries, thus without metadata.

If the function finds a `.so` it tries to open it and load the proc macros, if it's a `.rox` (?) it opens it and return a stream. That's why we have a pair with two components: some proc macros or a stream.

I think these changes highlight another underlying issue: Package opening should probably be decoupled from package search. The search function should probably return a list of path found, package loading should occur later depending on the extension and magic bytes. But this should probably be done in another PR as this refactor is not related to procedural macros.

PR #2506, opened 2023-08-02
Remove ```Backend::bool_type``` and ```Backend::char_type```

powerboat9:


CohenArthur (approved):
Nice cleanup, very GCC like and global but I think it is better

CohenArthur:
I mean, I prefer methods to global variables, but since we're simplifying that class more and more this change makes sense to me

philberty (approved):

PR #2507, opened 2023-08-02
Improve byte vector to string conversion

powerboat9:


gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -576,10 +576,8 @@ MacroBuiltin::include_str_handler (location_t invoc_locus,
   if (expect_single)
     rust_error_at (invoc_locus, "%s was not a valid utf-8 file",
 		   target_filename.c_str ());
-  else if (!bytes.empty ())
-    str = std::string ((const char *) &bytes[0], bytes.size ());
   else
-    return tl::nullopt;
```
dkm, line 582: It's not strictly equivalent, as `nullopt` can be used to signal that the transcriber didn't produce anything. Maybe that doesn't really matter, or it's better to have an empty fragment created here with sloc info. @CohenArthur ?

philberty (commented):
Do we not need to keep a check for bytes.empty() ?

Maybe if we get bytes.empty we can make str = empty string? and return to::nullopt?

dkm:
> Do we not need to keep a check for bytes.empty() ?
> 
> Maybe if we get bytes.empty we can make str = empty string? and return to::nullopt?

You mean replacing the `if (bytes.empty()) return nullopt;` by a later `if str.empty() return nullopt`?

powerboat9:
Can you not ```include_str!``` an empty file?

philberty:
Good point @powerboat9 I am just thinking its probably not safe to do .data on the bytes vector if its empty. How about we do:

```
if (!bytes.empty ())
else
  str = std::string ();
```

With no null so like you say we handle empty includes

powerboat9:
https://en.cppreference.com/w/cpp/container/vector/data seems to suggest that calling ```data``` on a vector should produce a valid result even if the vector is empty.

philberty:
Good find @powerboat9 lets go for it :D

philberty (approved):
LGTM

philberty:
lets try this again

Issue #2385, opened 2023-07-06
Add method of binding an ErrorCode to parser errors


philberty:
we use an Error struct in rust-diagnostics.h https://github.com/Rust-GCC/gccrs/blob/5406b633a1f5e6e3ae8da02589adf4ea003dec12/gcc/rust/rust-diagnostics.h#L133

This does not have the ability to store an Error code or Rich location so we cannot emit error codes for parser errors or expansion errors

philberty:
@MahadMuhammad ^^

PR #2527, opened 2023-08-05
[E0380] Use of auto trait with method or associated item

MahadMuhammad:
## Use of auto trait with a method or an associated item - [`E0380`](https://doc.rust-lang.org/error_codes/E0380.html)

 - Updated `add_error` to `rust_error_at`

---

### Running testcase:
- [`gcc/testsuite/rust/compile/auto_trait_invalid.rs`](https://github.com/Rust-GCC/gccrs/blob/master/gcc/testsuite/rust/compile/auto_trait_invalid.rs)
---

gcc/rust/ChangeLog:

	* parse/rust-parse-impl.h (Parser::parse_trait): added errorcode.

gcc/testsuite/ChangeLog:

	* rust/compile/auto_trait_invalid.rs: New comment.

---

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -4922,8 +4922,8 @@ Parser<ManagedTokenSource>::parse_trait (AST::Visibility vis,
 
   if (is_auto_trait && !trait_items.empty ())
     {
-      add_error (
-	Error (locus, "associated items are forbidden within auto traits"));
+      add_error (Error (locus, ErrorCode::E0380,
+			"auto traits cannot have associated items"));
 
       // FIXME: unsure if this should be done at parsing time or not
       for (const auto &item : trait_items)
--- a/gcc/testsuite/rust/compile/auto_trait_invalid.rs
+++ b/gcc/testsuite/rust/compile/auto_trait_invalid.rs
@@ -2,7 +2,7 @@
 
 #![feature(optin_builtin_traits)]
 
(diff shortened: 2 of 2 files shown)


philberty (approved):

philberty:
@MahadMuhammad is this ready to go? It looks good

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -4922,8 +4922,8 @@ Parser<ManagedTokenSource>::parse_trait (AST::Visibility vis,
 
   if (is_auto_trait && !trait_items.empty ())
     {
-      add_error (
-	Error (locus, "associated items are forbidden within auto traits"));
+      rust_error_at (locus, ErrorCode::E0380,
+		     "auto traits cannot have methods or associated items");
```
MahadMuhammad, line 4926: @philberty Parsing error needs to be collected using `Error` struct. I am working on this issue

- https://github.com/Rust-GCC/gccrs/issues/2385

Once, this issue get fixed. I'll then use the `Error` Struct to pass the `errorcode` instead of `rust_error_at`. If this doesn't change the behaviour of the code, then I am happy to remove this from draft.

CohenArthur, line 4926: yes, this does change the behavior of the code. `add_error` adds the created error to an internal table in the parser, which can choose to emit them later on or not. so we need to keep using `add_error` instead of `rust_error_at`

philberty (approved):
LGTM

CohenArthur (approved):

PR #2425, opened 2023-07-14
Add header file generator for Unicode normalization and alphanumeric check

tamaroning:
Addresses https://github.com/Rust-GCC/gccrs/issues/2379

This PR adds a header file generator written in python, which creates rust-unicode-data.h.
Also this PR adds initial implementation of Unicode normalization and is_numeric and is_alphabetic functions.

```
gcc/rust/ChangeLog:

	* Make-lang.in: Add rust-unicode.o
	* rust-lang.cc (run_rust_tests): Add test.
	* util/make-rust-unicode.py: Generater of rust-unicode-data.h.
	* util/rust-unicode-data.h: Auto-generated file.
	* util/rust-unicode.cc: New file.
	* util/rust-unicode.h: New file.
```

Unicode normalization is defined in https://unicode.org/reports/tr15/ (UAX15)
UAX15's implementation notes: https://unicode.org/reports/tr15/#Implementation_Notes
I looked at https://www.w3.org/International/charlint/ as reference implementation

PR #2529, opened 2023-08-06
Add type annotation to make-rust-unicode-data.py

tamaroning:
Addresses this comment by @CohenArthur 

> I think this is a very good first implementation of the script :) For future maintainability, I think it would be helpful to add types to the script so we can run it with mypy. But this is already very good, don't change it in this PR

https://github.com/Rust-GCC/gccrs/pull/2425#discussion_r1275951153

Python code was typechecked with mypy 1.4

```
gccrs: Type annotation for make-unicode-data.py
gcc/rust/ChangeLog:

	* util/make-rust-unicode.py: Add type annotation
```

### DIFF
--- a/gcc/rust/util/make-rust-unicode.py
+++ b/gcc/rust/util/make-rust-unicode.py
@@ -22,6 +22,10 @@
 #       > rust-unicode-data.h
 
 import sys
+from typing import Tuple
+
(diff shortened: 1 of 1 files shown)


philberty (approved):
LGTM

PR #2531, opened 2023-08-06
[E0753] Use of inner doc comment in invalid context

MahadMuhammad:
##  Use of inner doc comment in invalid context - [`E0753`](https://doc.rust-lang.org/error_codes/E0753.html)

---

### Code Tested:
- [`gcc/testsuite/rust/compile/bad_inner_doc.rs`](https://github.com/Rust-GCC/gccrs/blob/master/gcc/testsuite/rust/compile/bad_inner_doc.rs)

```rust
// ErrorCode - E0753
pub fn main ()
{
  //! inner doc allowed
  let _x = 42;
  // { dg-error "expected outer doc comment, inner doc" "" { target *-*-* } .+1 }
  //! inner doc disallowed
  mod module
  {
    /*! inner doc allowed */
    /// outer doc allowed
    // { dg-error "expected outer doc comment, inner doc" "" { target *-*-* } .+1 }
    /*! but inner doc not here */
    mod x { }
  }
}
```

### Output:

```bash
/home/mahad/Desktop/mahad/gccrs/gcc/testsuite/rust/compile/bad_inner_doc.rs:6:3: error: expected outer doc comment, inner doc ('//!' or '/*!') only allowed at start of item and before any outer attribute or doc ('#[', '///' or '/**') [E0753]
/home/mahad/Desktop/mahad/gccrs/gcc/testsuite/rust/compile/bad_inner_doc.rs:12:5: error: expected outer doc comment, inner doc ('//!' or '/*!') only allowed at start of item and before any outer attribute or doc ('#[', '///' or '/**') [E0753]
compiler exited with status 1
PASS: rust/compile/bad_inner_doc.rs  at line 5 (test for errors, line 6)
PASS: rust/compile/bad_inner_doc.rs  at line 11 (test for errors, line 12)
PASS: rust/compile/bad_inner_doc.rs (test for excess errors)
```

---

gcc/rust/ChangeLog:

	* parse/rust-parse-impl.h (Parser::parse_outer_attribute): Added errorcode & updated error function.

gcc/testsuite/ChangeLog:

	* rust/compile/bad_inner_doc.rs:

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -1218,8 +1218,9 @@ Parser<ManagedTokenSource>::parse_outer_attribute ()
   if (lexer.peek_token ()->get_id () == INNER_DOC_COMMENT)
     {
       Error error (
-	lexer.peek_token ()->get_locus (),
-	"inner doc (%<//!%> or %</*!%>) only allowed at start of item "
+	lexer.peek_token ()->get_locus (), ErrorCode::E0753,
+	"expected outer doc comment, inner doc (%<//!%> or %</*!%>) only "
+	"allowed at start of item "
 	"and before any outer attribute or doc (%<#[%>, %<///%> or %</**%>)");
       add_error (std::move (error));
       lexer.skip_token ();
--- a/gcc/testsuite/rust/compile/bad_inner_doc.rs
+++ b/gcc/testsuite/rust/compile/bad_inner_doc.rs
@@ -2,13 +2,13 @@ pub fn main ()
 {
   //! inner doc allowed
   let _x = 42;
-  // { dg-error "inner doc" "" { target *-*-* } .+1 }
+  // { dg-error "expected outer doc comment, inner doc" "" { target *-*-* } .+1 }
   //! inner doc disallowed
   mod module
   {
     /*! inner doc allowed */
     /// outer doc allowed
-    // { dg-error "inner doc" "" { target *-*-* } .+1 }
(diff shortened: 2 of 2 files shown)


gcc/rust/parse/rust-parse-impl.h
```diff
@@ -1217,11 +1217,11 @@ Parser<ManagedTokenSource>::parse_outer_attribute ()
 
   if (lexer.peek_token ()->get_id () == INNER_DOC_COMMENT)
     {
-      Error error (
-	lexer.peek_token ()->get_locus (),
-	"inner doc (%<//!%> or %</*!%>) only allowed at start of item "
+      rust_error_at (
```
P-E-P, line 1220: This changes the behavior of the code, the error is emitted immediately instead of being collected. Sometimes in some context we wish to collect those errors and discard them. I'm not too familiar with the parser's error collection. There should be a valid reason this was done before @CohenArthur ?

MahadMuhammad: Yes, @P-E-P you are right. For `parser` & `expansion` errors, we need to use `Error` struct instead of `rust_error_at` This shouldn't be there. I need to fix this issue

- [ ] https://github.com/Rust-GCC/gccrs/issues/2385


& then update this https://github.com/Rust-GCC/gccrs/pull/2531 and this PR



- [ ]  https://github.com/Rust-GCC/gccrs/pull/2527


Thanks for reminding, I completely forgot about this issue :)

philberty (approved):
LGTM

PR #2489, opened 2023-07-30
Normalize all identifiers to NFC

tamaroning:
Addresses https://github.com/Rust-GCC/gccrs/issues/2287
~~depends on https://github.com/Rust-GCC/gccrs/pull/2467~~

Normalize all identifiers (tokens) to their NFC form.
Normalization must be done before any macro expansion.

See https://doc.rust-lang.org/reference/identifiers.html#normalization for details

### Changelog
```
gccrs: Normalize all identifier tokens
gcc/rust/ChangeLog:

	* lex/rust-lex.cc (assert_source_content): Fix namespace specifier
	(test_buffer_input_source): Likewise.
	(test_file_input_source): Likewise.
	* lex/rust-lex.h: Move InputSource ...
	* lex/rust-input-source.h: ... to here. (New file)
	* lex/rust-token.cc (nfc_normalize_token_string): New function
	* lex/rust-token.h (nfc_normalize_token_string): New function
	* rust-lang.cc (run_rust_tests): Modify order of selftests.
	* rust-session-manager.cc (validate_crate_name): Modify interface of Utf8String.
	* util/rust-unicode.cc (lookup_cc): Modify codepoint_t typedef.
	(lookup_recomp): Likewise.
	(recursive_decomp_cano): Likewise.
	(decomp_cano): Likewise.
	(sort_cano): Likewise.
	(compose_hangul): Likewise.
	(assert_normalize): Likewise.
	(Utf8String::nfc_normalize): New function.
	* util/rust-unicode.h: Modify interface of Utf8String.

gcc/testsuite/ChangeLog:

	* rust/compile/unicode_norm1.rs: New test.
```

Issue #2287, opened 2023-06-12
Tokenize Unicode identifiers, whitespaces, etc.


tamaroning:
- [x] add util function to check XID_Start/XID_Continue to libcpp
  -  https://github.com/Rust-GCC/gccrs/pull/2284
- [x] identifiers
  -  https://github.com/Rust-GCC/gccrs/pull/2284
- [x] lifetime label
  -  https://github.com/Rust-GCC/gccrs/pull/2284
- [ ] raw identifiers
- [ ] string literal
  - already implemented?
- [ ] char literal
  - already implemented
- loop label 
  - not implemented yet 
- [ ] parse whitespaces
- [ ] add unit testing

CohenArthur:
> Unicode identifiers were stabilized in [rust-lang/rust#83799](https://github.com/rust-lang/rust/pull/83799). There is a PR ([rust-lang/reference#1022](https://github.com/rust-lang/reference/pull/1022)) on the reference to clarify the rules regarding unicode identifiers. There are also some lints to detect confusing cases like mixing different scripts.

*originally posted by @bjorn3*

Issue #305, opened 2021-03-26
Symbol Mangling


philberty:
Compiling Generics requires us to use symbol mangling to avoid the duplicate symbol error in linking.
This could also be part of a cleanup effort for the backend's handling for mangling in general.

- [X]  Legacy
- [x]  Legacy with actual sip-128 hash from https://github.com/rust-lang/rust/blob/506e75cbf8cb5305e49a41326307004ca3976029/compiler/rustc_data_structures/src/sip128.rs
- [ ] V0 https://rust-lang.github.io/rfcs/2603-rust-symbol-name-mangling-v0.html

bjorn3:
The hash that is part of the symbol name includes both the crate name and value passed in using `-Cmetadata` if any (always done by cargo). This metadata value allows for multiple versions of the same crate without symbol conflicts. Rustc will error when trying to link multiple versions of the same crate with the same `-Cmetadata` but different SVH (semantic version hash, derived from all inputs of the compilation process) in a single crate graph. The crate types bin, staticlib and cdylib are always the root of a crate graph. The crate types rlib and dylib are not, but rustc still checks for conflicts as far as I know as any conflicts present there must also be present in the final crate graph.

philberty:
Thanks for the heads up I didn't realise about -Cmetadata which is a really neat idea.

alex:
This is putting the cart way way way before the horse, but FYI: The Rust-for-Linux project uses the unstable v0 (though I have no recollection of why :-)). We're very interested in the progress of gccrs, for hopefully obvious reasons, so I figured I'd let you know.

philberty:
Moving this ticket out of the generics milestone, since we have a basic implementation of the legacy form will be enough for that milestone to be completed for now.

ojeda:
> The Rust-for-Linux project uses the unstable v0 (though I have no recollection of why :-)).

The reason is that the "legacy" mangling scheme uses invalid characters for C identifiers (like `$`) which is a problem when reusing the `EXPORT_SYMBOL_*` macros. Since I noticed v0 does not, I simply went for that. :)

PR #2530, opened 2023-08-06
Move Codepoint to gcc/rust/util/

tamaroning:
~~depends on https://github.com/Rust-GCC/gccrs/pull/2489~~

```
gccrs: Move Codepoint to gcc/rust/util/
gcc/rust/ChangeLog:

	* lex/rust-codepoint.h: Moved to...
	* util/rust-codepoint.h: ...here.
	* lex/rust-input-source.h: Add missing license
	* util/rust-unicode.cc: Add missing license
```

PR #2533, opened 2023-08-07
Implement punycode encoding

tamaroning:
Addresses https://github.com/Rust-GCC/gccrs/issues/2287 , https://github.com/Rust-GCC/gccrs/issues/305
Depends on https://github.com/Rust-GCC/gccrs/pull/2489

Added an punycode encoding function, which will be used in the v0 mangling scheme.
Decoder is not necessary because we do not need to demangle symbols.

```
gcc/rust/ChangeLog:

	* Make-lang.in: Add rust-punycode.o.
	* rust-lang.cc (run_rust_tests): Add selftest.
	* util/rust-punycode.cc: New file.
	* util/rust-punycode.h: New file.

Signed-off-by: Raiki Tamura <tamaron1203@gmail.com>
```

### DIFF
--- a/gcc/rust/util/rust-punycode.cc
+++ b/gcc/rust/util/rust-punycode.cc
@@ -1,0 +1,180 @@
+// Copyright (C) 2020-2023 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+// This file provides functions for punycode conversion
+// See https://datatracker.ietf.org/doc/html/rfc3492
+
+#include "rust-system.h"
+#include "rust-unicode.h"
+#include "optional.h"
+#include "selftest.h"
+
+namespace Rust {
+
+// https://tools.ietf.org/html/rfc3492#section-4.
+constexpr uint32_t BASE = 36;
+constexpr uint32_t TMIN = 1;
+constexpr uint32_t TMAX = 26;
+constexpr uint32_t SKEW = 38;
+constexpr uint32_t DAMP = 700;
+constexpr uint32_t INITIAL_BIAS = 72;
+constexpr uint32_t INITIAL_N = 128;
+constexpr char DELIMITER = '-';
+
+constexpr uint32_t MAX_ASCII_CODEPOINT = 0x7F;
+
+std::string
+extract_basic_string (const std::vector<Codepoint> &src)
+{
+  std::string basic_string;
+  for (auto c : src)
+    {
+      if (c.value <= MAX_ASCII_CODEPOINT)
+	basic_string += c.as_string ();
+    }
+  return basic_string;
+}
+
+uint32_t
+adapt_bias (uint32_t delta, const uint32_t n_points, const bool is_first)
+{
+  delta /= is_first ? DAMP : 2;
+  delta += delta / n_points;
+  uint32_t k = 0;
+
+  while (delta > (BASE - TMIN) * TMAX / 2)
+    {
+      delta /= BASE - TMIN;
+      k += BASE;
+    }
+  return k + (BASE - TMIN + 1) * delta / (delta + SKEW);
+}
+
+uint32_t
+clamped_sub (const uint32_t min, const uint32_t lhs, const uint32_t rhs,
+	     const uint32_t max)
+{
+  if (min + rhs >= lhs)
+    return min;
+  else if (max + rhs <= lhs)
+    return max;
+  else
+    return lhs - rhs;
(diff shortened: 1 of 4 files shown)


gcc/rust/lex/rust-input-source.h
```diff
@@ -0,0 +1,193 @@
...
+    uint32_t input = next_byte ();
+
+    if ((int32_t) input == EOF)
+      return Codepoint::eof ();
+    else if (input < 128)
+      {
+	// ascii -- 1 byte
+	return {input};
+      }
+    else if ((input & 0xC0) == 0x80)
```
P-E-P, line 31: There are *a lot* of hardcoded values. Would it be possible to give those values a name ? This could be achieved throught `#define`, or even better, with typed const values.

Here, someone new to unicode (and this code) could not grasp what this condition even test.

Actually you may even want to put those constant in their own separate header ?

gcc/rust/lex/rust-input-source.h
```diff
@@ -0,0 +1,193 @@
...
+    else
+      {
+	return {0xFFFE};
+      }
+  }
+
+protected:
+  // Check if the input source is valid as utf-8 and copy all characters to
+  // `chars`.
+  void init ()
```
P-E-P, line 103: Why not changing this to an `InputSource` constructor ? This would allow you to avoid calling it in any derived class.

gcc/rust/lex/rust-input-source.h
```diff
@@ -0,0 +1,193 @@
...
+{
+private:
+  const std::string &buffer;
+  size_t offs;
+
+  int next_byte () override
+  {
+    if (offs >= buffer.size ())
+      return EOF;
+    return (uint8_t) buffer.at (offs++);
```
P-E-P, line 178: I'm not sure about this cast. I mean, it doesn't even correspond to the return type. Are you trying to get rid of the upper bits ? Also, I would probably change this to a `static_cast`.

gcc/rust/lex/rust-input-source.h
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+public:
+  // Create new input source from file.
+  BufferInputSource (const std::string &b, size_t offset)
+    : InputSource (), buffer (b), offs (offset)
```
P-E-P, line 184: `InputSource` default constructor is called implicitely.

gcc/rust/lex/rust-input-source.h
tamaroning: Yes, this cast avoids a value from beign sign-extended.
static_cast is better.

gcc/rust/lex/rust-input-source.h
tamaroning: I just tried  code inside `void init()` to the ctor of `InputSource`, but segfault happens before or during the ctor.
`next_byte()` is defined as virtual so, I think, `next_byte` is not initialized when it is called.
I will leave `init()` as is.
https://stackoverflow.com/questions/962132/calling-virtual-functions-inside-constructors

(But I do not have knownledge about this so I might be completely wrong)

P-E-P: Oh ok, I didn't see that. keep it as is then. Could you please add a comment documenting this behavior ? I'm almost sure someone will ask the same thing in the upcoming years :laughing:

gcc/rust/lex/rust-input-source.h
tamaroning: Thanks. Fixed in https://github.com/Rust-GCC/gccrs/pull/2530

gcc/rust/lex/rust-input-source.h
tamaroning: Yes! Added comments in https://github.com/Rust-GCC/gccrs/pull/2530

gcc/rust/util/rust-punycode.cc
```diff
@@ -0,0 +1,180 @@
...
+  for (auto c : l)
+    if (c.value >= threshold && c.value < min)
+      min = c.value;
+  return min;
+}
+
+char
+encode_digit (const uint32_t d)
+{
+  return d + 22 + (d < 26 ? 75 : 0);
```
P-E-P, line 93: I'd like some comments here to describe the behavior of this function. Newcomers should be able to understand why and where do these values come from.

P-E-P (approved):
Small nit. I'll merge this PR once it is fixed and the conflicts are resolved.

gcc/rust/util/rust-punycode.cc
tamaroning: This function comes from the reference implementation in the RFC
https://datatracker.ietf.org/doc/html/rfc3492

```
/* encode_digit(d,flag) returns the basic code point whose value      */
/* (when used for representing integers) is d, which needs to be in   */
/* the range 0 to base-1.  The lowercase form is used unless flag is  */
/* nonzero, in which case the uppercase form is used.  The behavior   */
/* is undefined if flag is nonzero and digit d has no uppercase form. */

static char encode_digit(punycode_uint d, int flag)
{
  return d + 22 + 75 * (d < 26) - ((flag != 0) << 5);
  /*  0..25 map to ASCII a..z or A..Z */
  /* 26..35 map to ASCII 0..9         */
}
```

These value comes from a bit complicated calculation on parameters specific to Punycode.
I do not completely understand it but I think it is fine as is.

tamaroning:
Rebased on master.

gcc/rust/util/rust-punycode.cc
P-E-P: Oh ok. Given you've already provided the link at the top of the file I don't think there's much we can do.

PR #2537, opened 2023-08-08
Handle nested macro definition (resolver)

### LINKED ISSUE #2516: Handle macro invocations expanding to new macro definitions
https://github.com/rust-lang/rfcs/pull/1560#issuecomment-217683514

```rust
mod a {
    macro_rules! m { () => {
        macro_rules! m { () => {
            fn f() {}
        } }
    } }
}
mod b {
    use a::*;
    c::m! {} // If we expand this first, it would resolve to `a::m!` and then
             // `b::m!` would expand to `fn f() {}`, ensuring correctness.
}
mod c {
    use a::*;
    b::m! {} // If we expand this first, it would resolve to `a::m!` and then
             // `c::m!` would expand to `fn f() {}`, ensuring correctness.
}
```

One way to go about it is to run the Early name resolver and expander in a fixed point, and store errors in a container instead of emitting them immediately. Then, once the fixed point is reached, emit the errors if there are any

P-E-P:
Fixes #2516 

This PR clashes with #2541  we need to decide which one get merged. Per

> // now do we need to keep mappings or something? or insert "uses" into our
> // ForeverStack? can we do that? are mappings simpler?

This is still an early draft, I don't really know where I am going with this one. @CohenArthur could you give me more details about what you had in mind with this comment ?

### DIFF
--- a/gcc/rust/resolve/rust-early-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-early-name-resolver-2.0.cc
@@ -135,8 +135,10 @@ Early::visit (AST::MacroInvocation &invoc)
   // if the definition still does not have a value, then it's an error
   if (!definition.has_value ())
     {
-      rust_error_at (invoc.get_locus (), ErrorCode::E0433,
-		     "could not resolve macro invocation");
+      collect_error ([&] () {
+	rust_error_at (invoc.get_locus (), ErrorCode::E0433,
+		       "could not resolve macro invocation");
+      });
       return;
     }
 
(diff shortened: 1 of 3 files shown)


P-E-P:
Not sure wether we should go with mappings or a new system per:
>   // now do we need to keep mappings or something? or insert "uses" into our
>  // ForeverStack? can we do that? are mappings simpler?

A working version with mappings is available at https://github.com/P-E-P/gccrs/tree/nested_macro_definition_mappings

```rust
macro_rules! titi {
    () => {
        let a = 0;
    };
}
macro_rules! toto {
    () => {
        macro_rules! tata {
            () => {
                let i = 0;
            };
        }
    };
}

pub fn main() {
    toto!();
    tata!();
}
```

CohenArthur (approved):
This is a good starting point! Thank you for working on this :)

P-E-P:
We'll use mapping instead. Check #2541

PR #2541, opened 2023-08-10
Handle nested macro definition (mappings)

### LINKED ISSUE #2233: Macro-expanded macro definitions are not resolved
I tried this code:

```rust
macro_rules! one {
    () => {
        macro_rules! two {
            () => {
                
            }
        }
    }
}

one!();
two!();
```

I expected to see this happen: compiles

Instead, this happened:

```text
<source>:12:1: error: unknown macro: [two]
   12 | two!();
      | ^~~
```

### Meta

gccrs (Compiler-Explorer-Build-gcc-c7b7e297ea4b688ef9fb51a8aee2a8b2af39ac1c-binutils-2.40) 13.0.1 20230417 (experimental)

### LINKED ISSUE #2516: Handle macro invocations expanding to new macro definitions

P-E-P:
Fixes #2233 

This PR clashes with #2537 we need to decide which one get merged. Per

> // now do we need to keep mappings or something? or insert "uses" into our
> // ForeverStack? can we do that? are mappings simpler?

### DIFF
--- a/gcc/rust/resolve/rust-early-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-early-name-resolver-2.0.cc
@@ -135,10 +135,8 @@ Early::visit (AST::MacroInvocation &invoc)
   // if the definition still does not have a value, then it's an error
   if (!definition.has_value ())
     {
-      collect_error ([&] () {
-	rust_error_at (invoc.get_locus (), ErrorCode::E0433,
-		       "could not resolve macro invocation");
-      });
+      collect_error (Error (invoc.get_locus (), ErrorCode::E0433,
+			    "could not resolve macro invocation"));
       return;
     }
 
--- a/gcc/rust/resolve/rust-early-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-early-name-resolver-2.0.cc
@@ -135,13 +135,28 @@ Early::visit (AST::MacroInvocation &invoc)
   // if the definition still does not have a value, then it's an error
   if (!definition.has_value ())
     {
-      rust_error_at (invoc.get_locus (), ErrorCode::E0433,
-		     "could not resolve macro invocation");
+      collect_error ([&] () {
+	rust_error_at (invoc.get_locus (), ErrorCode::E0433,
+		       "could not resolve macro invocation");
+      });
       return;
     }
 
   // now do we need to keep mappings or something? or insert "uses" into our
   // ForeverStack? can we do that? are mappings simpler?
+  auto mappings = Analysis::Mappings::get ();
+  AST::MacroRulesDefinition *rules_def = nullptr;
+  if (!mappings->lookup_macro_def (definition.value (), &rules_def))
+    {
+      // Macro definition not found, maybe it is not expanded yet.
+      return;
+    }
+
+  AST::MacroRulesDefinition *tmp_def = nullptr;
+  if (mappings->lookup_macro_invocation (invoc, &tmp_def))
+    return;
+
+  mappings->insert_macro_invocation (invoc, rules_def);
 }
 
 void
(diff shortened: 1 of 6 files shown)


gcc/rust/resolve/rust-early-name-resolver-2.0.h
```diff
@@ -28,6 +28,8 @@
 namespace Rust {
 namespace Resolver2_0 {
 
+using ResolveError = std::function<void ()>;
```
CohenArthur, line 31: I think we should simply store errors in an `Error` container like the parser does. I like the lambda idea, but they are unwieldy in C++

gcc/rust/resolve/rust-early-name-resolver-2.0.cc
```diff
@@ -135,13 +135,28 @@ Early::visit (AST::MacroInvocation &invoc)
   // if the definition still does not have a value, then it's an error
   if (!definition.has_value ())
     {
-      rust_error_at (invoc.get_locus (), ErrorCode::E0433,
-		     "could not resolve macro invocation");
+      collect_error ([&] () {
+	rust_error_at (invoc.get_locus (), ErrorCode::E0433,
+		       "could not resolve macro invocation");
+      });
```
CohenArthur, line 141: so this would become something like
```suggestion
      add_error (Error (invoc.get_locus (), ErrorCode::E0433,
		       "could not resolve macro invocation"));
```

gcc/rust/rust-session-manager.cc
```diff
@@ -883,7 +884,11 @@ Session::expansion (AST::Crate &crate)
       auto ctx = Resolver2_0::NameResolutionContext ();
 
       if (flag_name_resolution_2_0)
-	Resolver2_0::Early (ctx).go (crate);
+	{
+	  Resolver2_0::Early early (ctx);
+	  early.go (crate);
+	  macro_errors = early.get_macro_resolve_errors ();
```
CohenArthur, line 890: we need to clear out the macro resolve errors each time we run through the fixed point, otherwise we are going to emit errors which would have been resolved in a later run, aren't we?

CohenArthur (approved):

gcc/rust/rust-session-manager.cc
P-E-P: Mmmh no? From what I remember I tested this edge case. I may be wrong though.

`macro_errors` get cleared by the assignation and the internal vector from `get_macro_resolve_errors` is cleared anyway because of the new instance of the early resolver in the loop. If we change this instantiation to put it out of the loop then yes we should provide a function to clear the vector.

gcc/rust/resolve/rust-early-name-resolver-2.0.h
P-E-P: Fixed in fdccd48aff5.

gcc/rust/resolve/rust-early-name-resolver-2.0.cc
P-E-P: Fixed in fdccd48aff5.

CohenArthur (approved):
Looks good! Great work :D will start using this straight away!

PR #2542, opened 2023-08-10
Support for `rich-loc` & `errorcode` in parser & expansion errors

MahadMuhammad:
## Binding `rich-loc` & `errorcode` in parser & expansion errors

Added method of binding ErrorCode & rich location to parser and expansion errors.
- Fixes https://github.com/Rust-GCC/gccrs/issues/2385
- Depends on https://github.com/Rust-GCC/gccrs/pull/2531 & https://github.com/Rust-GCC/gccrs/pull/2527

Currently, the `Error` struct only supports simple location to `parser` and `expansion` errors. Updated the code to handle all cases for `errorcode` and `parsing` errors. Now, we can rich-loation to `Error` struct by using ampersand `&` operator.

The `rust_error_at` function immediately emits the error. But in some contexts we wish to collect them and discard them. Now, we can use errorcode and rich location in `Error` struct's context.

--- 

**gcc/rust/ChangeLog:**

	* rust-diagnostics.cc (va_constructor): Added constructor for all possible cases. (Error::Error): Updated error struct for all possible cases.
	* rust-diagnostics.h (struct Error): Updated error struct to support error code & rich location support.

### DIFF
--- a/gcc/rust/rust-diagnostics.h
+++ b/gcc/rust/rust-diagnostics.h
@@ -23,6 +23,10 @@
 
 #include "rust-linemap.h"
 
+// This macro is used to specify the position of format string & it's
+// arguments within the function's paramter list.
+// 'm' specifies the position of the format string parameter.
+// 'n' specifies the position of the first argument for the format string.
 #if __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 1)
 #define RUST_ATTRIBUTE_GCC_DIAG(m, n)                                          \
   __attribute__ ((__format__ (__gcc_tdiag__, m, n)))                           \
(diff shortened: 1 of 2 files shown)


gcc/rust/rust-diagnostics.h
```diff
@@ -23,6 +23,10 @@
 
 #include "rust-linemap.h"
 
+// This macro is used to specify the position of format string & it's
+// arguments within the function's paramter list.
+// 'n' specifies the position of the format string parameter.
+// 'm' specifies the position of the first argument for the format string.
```
MahadMuhammad, line 29: It should be:
```suggestion
// 'm' specifies the position of the format string parameter.
// 'n' specifies the position of the first argument for the format string.
```

MahadMuhammad: Updated :)

gcc/rust/rust-diagnostics.h
```diff
@@ -185,9 +230,22 @@ struct Error
   }
 
   // TODO: the attribute part might be incorrect
+  // simple location
   Error (location_t locus, const char *fmt,
 	 ...) /*RUST_ATTRIBUTE_GCC_DIAG (2, 3)*/ RUST_ATTRIBUTE_GCC_DIAG (3, 4);
```
MahadMuhammad, line 235: Just eager to know, I tried using the commented version of `RUST_ATTRIBUTE_GCC_DIAG` macro (they are seem to correct). But don't know why are not working

Maybe they are linked with these constructor definitions:
https://github.com/Rust-GCC/gccrs/blob/5712d7d383b0597cd00466e5995c1e620c2fc809/gcc/rust/rust-diagnostics.cc#L387-L390

Or is there any other logic or reason?

philberty (approved):
LGTM

Issue #2548, opened 2023-08-12
`#[no_mangle]` should not be applied to non-ASCII items


tamaroning:
Items with an ASCII name and `#[no_mangle]` should not compile.

input:
```rust
#[no_mangle]
pub fn ぽ() {}

pub fn main() {}
```

output: compiled.

expected: rustc's output 
```bash
error[[E0754]](https://doc.rust-lang.org/stable/error_codes/E0754.html): `#[no_mangle]` requires ASCII identifier
 --> src/main.rs:2:1
  |
2 | pub fn ぽ() {}
  | ^^^^^^^^^^^
```

PR #2552, opened 2023-08-13
`#[no_mangle]` should not be applied to non-ASCII items

tamaroning:
Fixes https://github.com/Rust-GCC/gccrs/issues/2548

```
gccrs: Check function names with no_mangle.
gcc/rust/ChangeLog:

	* backend/rust-compile-base.cc (HIRCompileBase::setup_fndecl): Add parameter.
	* backend/rust-compile-base.h: Likewise.
	* backend/rust-compile-implitem.cc (CompileTraitItem::visit): Change type of parameter.
	* backend/rust-compile-item.cc (CompileItem::visit): Likewise.
	* lex/rust-input-source.h: Move constants to rust-codepoint.h
	* util/rust-codepoint.h (struct Codepoint): Add is_ascii method.
	* util/rust-punycode.cc (extract_basic_string): Use it.
	* lex/rust-lex.cc (Lexer::parse_byte_char): Likewise.
	* backend/rust-mangle.cc (legacy_mangle_name): Likewise.
	* util/rust-unicode.cc (is_ascii_only): New function.
	* util/rust-unicode.h (is_ascii_only): Likewise.
```

Error diagnostics looks like this:
```rust
#[no_mangle]
pub fn ascii() {}

#[no_mangle]
pub fn notーascii() {}
```

```bash
./gccrs-build/gcc/crab1 a.rs -frust-incomplete-and-experimental-compiler-do-not-use -fdump-tree-gimple
a.rs:5:8: error: attribute ‘no_mangle’ requires ASCII identifier [E0754]
    5 | pub fn notーascii() {}
      |        ^~~~~~~~

```

### DIFF
--- a/gcc/rust/util/rust-attributes.cc
+++ b/gcc/rust/util/rust-attributes.cc
@@ -22,6 +22,7 @@
 #include "rust-ast.h"
 #include "rust-ast-full.h"
 #include "rust-diagnostics.h"
+#include "rust-unicode.h"
 
 namespace Rust {
 namespace Analysis {
@@ -612,6 +613,22 @@ AttributeChecker::visit (AST::UseDeclaration &declaration)
   check_proc_macro_non_function (declaration.get_outer_attrs ());
 }
 
+static void
+check_no_mangle_function (const AST::Attribute &attribute,
+			  const AST::Function &fun)
+{
+  if (attribute.has_attr_input ())
+    {
+      rust_error_at (attribute.get_locus (), ErrorCode::E0754,
+		     "malformed %<no_mangle%> attribute input");
+      rust_inform (attribute.get_locus (),
+		   "must be of the form: %<#[no_mangle]%>");
+    }
+  if (!is_ascii_only (fun.get_function_name ().as_string ()))
+    rust_error_at (fun.get_function_name ().get_locus (),
+		   "the %<#[no_mangle]%> attribute requires ASCII identifier");
+}
+
 void
 AttributeChecker::visit (AST::Function &fun)
 {
@@ -649,6 +666,8 @@ AttributeChecker::visit (AST::Function &fun)
 	{
 	  check_crate_type (name, attribute);
 	}
+      else if (result.name == "no_mangle")
+	check_no_mangle_function (attribute, fun);
     }
   fun.get_definition ()->accept_vis (*this);
 }
--- a/gcc/rust/backend/rust-mangle.cc
+++ b/gcc/rust/backend/rust-mangle.cc
@@ -87,7 +87,7 @@ legacy_mangle_name (const std::string &name)
 	  i++;
 	  m = "..";
 	}
-      else if (c.value < 0x80)
+      else if (c.is_ascii ())
 	// ASCII
 	m.push_back (c.value);
       else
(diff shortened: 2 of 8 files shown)


tamaroning:
> Hm, I'm not sure having that check so late in the backend is the right way to go about it. We can add a check to the attribute checker in rust-attributes.h and just check if a function or method has the attribute and a non-ascii name. I think that will be easier

rustc does somethinig like it. It will be better.

tamaroning:
@CohenArthur 
Fixed code. Could you please review again?

gcc/rust/util/rust-attributes.cc
```diff
@@ -623,6 +624,12 @@ AttributeChecker::visit (AST::Function &fun)
 		     name);
   };
 
+  auto check_function_name = [] (const char *name, const Identifier &ident) {
+    if (!is_ascii_only (ident.as_string ()))
+      rust_error_at (ident.get_locus (),
+		     "the %<#[%s]%> attribute requires ASCII identifier", name);
+  };
```
CohenArthur, line 631: will this function be used for anything other than the `#[no_mangle]` attribute? If no, we can skip having the `const char *name` argument. I think it would also be cleaner to have it as a `static void` function in the file rather than a lambda, as C++ lambdas are not super easy to read

CohenArthur (approved):
Looks really good! And yeah, it is much simpler. Thanks for making the changes! Great work!

gcc/rust/util/rust-attributes.cc
tamaroning: Okay. I will fix it later.

tamaroning:
@CohenArthur Everything is fixed now :)

gcc/rust/util/rust-attributes.cc
```diff
@@ -612,6 +613,22 @@ AttributeChecker::visit (AST::UseDeclaration &declaration)
...
+  if (attribute.has_attr_input ())
+    {
+      rust_error_at (attribute.get_locus (), ErrorCode::E0754,
+		     "malformed %<no_mangle%> attribute input");
+      rust_inform (attribute.get_locus (),
+		   "must be of the form: %<#[no_mangle]%>");
+    }
+  if (!is_ascii_only (fun.get_function_name ().as_string ()))
+    rust_error_at (fun.get_function_name ().get_locus (),
+		   "the %<#[no_mangle]%> attribute requires ASCII identifier");
```
CohenArthur, line 629:
```suggestion
		   "the %<#[no_mangle]%> attribute requires an ASCII identifier");
```

CohenArthur (approved):
Looks great! Tiny nitpick and I'll merge this. Thank you!

gcc/rust/util/rust-attributes.cc
tamaroning: Sorry for the late reply! I was busy last week.
This message comes from this actual rustc's message.
```
Compiling playground v0.0.1 (/playground)
error[[E0754]](https://doc.rust-lang.org/stable/error_codes/E0754.html): `#[no_mangle]` requires ASCII identifier
 --> src/main.rs:2:1
  |
2 | pub fn ああ() {}
  | ^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0754`.
error: could not compile `playground` (bin "playground") due to previous error
```

CohenArthur: oh, alright!

PR #2554, opened 2023-08-13
Handle gengtype annotations in ```rust-constexpr.cc```

powerboat9:


powerboat9:
@philberty @CohenArthur I'm not sure why this patch isn't compiling, any ideas?

powerboat9:
Fixed it, it was some sort of type name collision.

CohenArthur (commented):
I don't know if that is really needed, I'm wondering if this isn't a remnant of the Go implementation that we ported over a long time ago. We don't use the garbage collecting routines anywhere else in the compiler

powerboat9:
It looks like it might be rather important for constant evaulation

philberty (approved):
LGTM nice one this stuff is a pain I tried to get it working last year but now I realise yes we needed the using namespace that you found.

Nice work

PR #2555, opened 2023-08-14
[E0592] method or associated functions already defined with same names

MahadMuhammad:
## Multiple candidated for `method` or `function` - [`E0592`](https://doc.rust-lang.org/error_codes/E0592.html)

- Added errorcode & rich location.

### Running testcases:
- [`gcc/testsuite/rust/compile/generics7.rs`](https://github.com/Rust-GCC/gccrs/blob/master/gcc/testsuite/rust/compile/generics7.rs)
- [`gcc/testsuite/rust/compile/issue-925.rs`](https://github.com/Rust-GCC/gccrs/blob/master/gcc/testsuite/rust/compile/issue-925.rs)

### Output:

```rust
../mahad-testsuite/E0592.rs:28:7: error: duplicate definitions with name ‘bar’ [E0592]
    9 |     fn bar(self) -> isize {
      |     ~~ 
      |     duplicate definitions for bar
......
   21 |     fn bar(self) -> T {
      |     ~~ 
      |     duplicate definitions for bar
......
   28 |     a.bar();
      |       ^~~

```

gcc/rust/ChangeLog:

	* typecheck/rust-hir-type-check-expr.cc (TypeCheckExpr::visit): Added rich location and errorcode.

gcc/testsuite/ChangeLog:

	* rust/compile/generics7.rs: Updated comment.
	* rust/compile/issue-925.rs: likewise.

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-expr.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-expr.cc
@@ -1097,8 +1097,7 @@ TypeCheckExpr::visit (HIR::MethodCallExpr &expr)
       r.add_fixit_replace (rich_msg.c_str ());
 
       rust_error_at (
-	r, ErrorCode::E0034,
-	"multiple applicable items in scope for method %qs",
+	r, ErrorCode::E0592, "duplicate definitions with name %qs",
 	expr.get_method_name ().get_segment ().as_string ().c_str ());
       return;
     }
--- a/gcc/rust/typecheck/rust-hir-inherent-impl-overlap.h
+++ b/gcc/rust/typecheck/rust-hir-inherent-impl-overlap.h
@@ -129,8 +129,11 @@ public:
 			   const std::string &name)
   {
     rich_location r (line_table, dup->get_locus ());
+    std::string msg = "duplicate definitions for " + name;
+    r.add_fixit_replace (query->get_locus (), msg.c_str ());
     r.add_range (query->get_locus ());
-    rust_error_at (r, "duplicate definitions with name %s", name.c_str ());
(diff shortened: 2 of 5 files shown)


gcc/rust/typecheck/rust-hir-type-check-expr.cc
```diff
@@ -1065,11 +1065,16 @@ TypeCheckExpr::visit (HIR::MethodCallExpr &expr)
...
       for (auto &c : candidates)
-	r.add_range (c.candidate.locus);
+	{
+	  r.add_fixit_replace (c.candidate.locus, msg.c_str ());
+	  r.add_range (c.candidate.locus);
+	}
 
       rust_error_at (
-	r, "multiple candidates found for method %<%s%>",
+	r, ErrorCode::E0592, "duplicate definitions with name %<%s%>",
```
CohenArthur, line 1077: hm, not sure which message we should keep here and which error code we should use

CohenArthur: I think we should keep the original error message, as this one could confuse the user since the error will appear on the use site rather than the definition site

CohenArthur (changes requested):

gcc/rust/typecheck/rust-hir-type-check-expr.cc
MahadMuhammad: Done :)

CohenArthur (approved):

philberty:
@MahadMuhammad this has merge conflicts can you fix it up?

MahadMuhammad:
@philberty, Sure thing. I'll do it :)

P-E-P (approved):

PR #2556, opened 2023-08-14
Non-allowed default type paramters

MahadMuhammad:
### Non-allowed default `type` paramters

- Added user friendly message & `fixit` hint.

gcc/rust/ChangeLog:

	* hir/rust-ast-lower-item.cc (ASTLoweringItem::visit): Added more helpful error message.

gcc/testsuite/ChangeLog:

	* rust/compile/generics10.rs: for dejagnu.

### DIFF
--- a/gcc/rust/hir/rust-ast-lower-item.cc
+++ b/gcc/rust/hir/rust-ast-lower-item.cc
@@ -499,10 +499,15 @@ ASTLoweringItem::visit (AST::InherentImpl &impl_block)
 
 		if (t.has_type ())
 		  {
-		    // see https://github.com/rust-lang/rust/issues/36887
-		    rust_error_at (
+		    rich_location rich_locus (line_table, t.get_locus ());
+		    rich_locus.add_fixit_replace (
 		      t.get_locus (),
-		      "defaults for type parameters are not allowed here");
+		      "for more information, see issue #36887 "
+		      "<https://github.com/rust-lang/rust/issues/36887>");
+		    rust_error_at (rich_locus,
+				   "defaults for type parameters are only "
+				   "allowed in %<struct%>, %<enum%>, %<type%>, "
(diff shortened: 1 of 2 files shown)


philberty (approved):
LGTM

gcc/rust/hir/rust-ast-lower-item.cc
```diff
@@ -645,10 +650,15 @@ ASTLoweringItem::visit (AST::TraitImpl &impl_block)
...
-		      "defaults for type parameters are not allowed here");
+		    rich_location rich_locus (line_table, t.get_locus ());
+		    rich_locus.add_fixit_replace (
+		      t.get_locus (), "for more information, see issue #36887 "
+				      "<https://github.com/rust-lang/rust/"
+				      "issues/36887>");
+		    rust_error_at (rich_locus,
+				   "defaults for type parameters are only "
+				   "allowed in %<struct%>, %<enum%>, %<type%>, "
+				   "or %<trait%> definitions");
```
CohenArthur, line 661: nice, it looks good! in a different PR, could you factor these lines in a separate function? this way we can modify the error message without changing it in two places

CohenArthur (approved):

gcc/rust/hir/rust-ast-lower-item.cc
MahadMuhammad: @philberty  left this issue to me to decide, whether to create a separate class for refactoring these errors.

But that seems excessive. What do you think, @CohenArthur? Should I create a separate class or add a refactor function when we encounter these cases?

CohenArthur: A separate class or a separate function is a good idea. I like both :)

MahadMuhammad: Okay :)

PR #2557, opened 2023-08-14
Invalid order of generic parameters

MahadMuhammad:
### Invalid order of generic parameters

- Added more user-friendly message & hint for fixing ordering

### Output:
- [`gcc/testsuite/rust/compile/generics13.rs`](https://github.com/Rust-GCC/gccrs/blob/master/gcc/testsuite/rust/compile/generics13.rs)
```rust
➜  gccrs-build gcc/crab1 /home/mahad/Desktop/mahad/gccrs/gcc/testsuite/rust/compile/generics13.rs
/home/mahad/Desktop/mahad/gccrs/gcc/testsuite/rust/compile/generics13.rs:1:12: error: invalid order for generic parameters: lifetime parameters must be declared prior to type and const parameters
    1 | struct Foo<A, 'a>; // { dg-error "invalid order for generic parameters: lifetimes should always come before types" }
      |            ^
      |            reorder the parameters: lifetimes, then consts and types

Analyzing compilation unit

 TOTAL                              :   0.00          0.00          0.00          146k

```
gcc/rust/ChangeLog:

	* parse/rust-parse-impl.h (Parser::parse_generic_params): Added fixit_hint and more user friendly message.

gcc/testsuite/ChangeLog:

	* rust/compile/generics13.rs: for dejagnu,

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -3142,9 +3142,12 @@ Parser<ManagedTokenSource>::parse_generic_params (EndTokenPred is_end_token)
 
   // FIXME: Add reordering hint
   if (order_error)
-    rust_error_at (generic_params.front ()->get_locus (),
-		   "invalid order for generic parameters: lifetimes should "
-		   "always come before types");
+    {
+      Error error (generic_params.front ()->get_locus (),
+		   "invalid order for generic parameters: lifetime parameters "
+		   "must be declared prior to type and const parameters");
+      add_error (std::move (error));
+    }
 
   generic_params.shrink_to_fit ();
   return generic_params;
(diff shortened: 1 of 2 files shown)


gcc/rust/parse/rust-parse-impl.h
```diff
@@ -3139,11 +3139,17 @@ Parser<ManagedTokenSource>::parse_generic_params (EndTokenPred is_end_token)
...
-    rust_error_at (generic_params.front ()->get_locus (),
-		   "invalid order for generic parameters: lifetimes should "
-		   "always come before types");
+    {
+      rich_location rich_locus (line_table,
+				generic_params.front ()->get_locus ());
+      rich_locus.add_fixit_replace (
+	generic_params.front ()->get_locus (),
+	"reorder the parameters: lifetimes, then consts and types");
+      rust_error_at (
```
philberty, line 3149: We can't do a rust_error_at like this in the parser. You need to write a wrapper for a an error like this with the Error class you have worked with

philberty (changes requested):

gcc/rust/parse/rust-parse-impl.h
MahadMuhammad: @philberty, currently we cannot bind `rich_location` to these parser errors. See this
- https://github.com/Rust-GCC/gccrs/issues/2566

So, I only added the updated the wrapper function, but didn't add the `fixit` hint :(

CohenArthur (approved):

PR #2563, opened 2023-08-16
[E0532] Pattern arm did not match expected kind.

MahadMuhammad:
### Code Tested:

- [`gcc/testsuite/rust/compile/issue-2029.rs`](https://github.com/Rust-GCC/gccrs/blob/master/gcc/testsuite/rust/compile/issue-2029.rs)

```rust
~/gccrs/gcc/testsuite/rust/compile/issue-2029.rs:8:9: error: expected tuple struct or tuple variant, found function ‘fn (_ usize,) -> Foo{Foo (0:bool)}’ [E0532]
    8 |         foo(x)
      |         ^~~
      |         a tuple struct with a similar name exists

```

---

**gcc/rust/ChangeLog:**

	* typecheck/rust-hir-type-check-pattern.cc (TypeCheckPattern::visit): Added rich location & error code.

**gcc/testsuite/ChangeLog:**

	* rust/compile/issue-2029.rs: Updated for dejagnu testcase.

---

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-pattern.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-pattern.cc
@@ -53,9 +53,10 @@ TypeCheckPattern::visit (HIR::TupleStructPattern &pattern)
   TyTy::BaseType *pattern_ty = TypeCheckExpr::Resolve (&pattern.get_path ());
   if (pattern_ty->get_kind () != TyTy::TypeKind::ADT)
     {
-      rust_error_at (pattern.get_locus (),
-		     "expected tuple struct/variant, found: %s",
-		     pattern_ty->get_name ().c_str ());
+      rust_error_at (
+	pattern.get_locus (), ErrorCode::E0532,
+	"expected tuple struct or tuple variant, found function %qs",
+	pattern_ty->get_name ().c_str ());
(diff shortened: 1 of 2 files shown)


gcc/rust/typecheck/rust-hir-type-check-pattern.cc
```diff
@@ -53,9 +53,14 @@ TypeCheckPattern::visit (HIR::TupleStructPattern &pattern)
   TyTy::BaseType *pattern_ty = TypeCheckExpr::Resolve (&pattern.get_path ());
   if (pattern_ty->get_kind () != TyTy::TypeKind::ADT)
     {
-      rust_error_at (pattern.get_locus (),
-		     "expected tuple struct/variant, found: %s",
-		     pattern_ty->get_name ().c_str ());
+      rich_location rich_locus (line_table, pattern.get_locus ());
+      rich_locus.add_fixit_replace (
+	"a tuple struct with a similar name exists");
```
CohenArthur, line 58: this should only be emitted if a tuple struct with a similar name exists, otherwise that help is misleading. I would remove this for now

CohenArthur (changes requested):

philberty (approved):

Issue #2430, opened 2023-07-15
Neither tuple struct nor a tuple variant was used as a pattern - Internal Compiler Error


MahadMuhammad:
## Neither tuple struct nor a tuple variant was used as a pattern  - [`E0164`](https://doc.rust-lang.org/error_codes/E0164.html)

- Fixing this issue, help `gccrs` to emit error code [`E0164`](https://doc.rust-lang.org/error_codes/E0164.html) similiar to rustc.
[...]

PR #2565, opened 2023-08-16
[E0164] Neither tuple struct nor tuple variant used as a pattern

MahadMuhammad:
## Neither a tuple struct nor a tuple variant was used as a pattern - [`E0164`](https://doc.rust-lang.org/error_codes/E0164.html)

Checking if pattern has items, before returing solves ICE. Added error code and rich location.
- Also, fixes https://github.com/Rust-GCC/gccrs/issues/2430

---

**gcc/rust/ChangeLog:**

	* ast/rust-pattern.h: No need of assertion, we are handling it.
	* resolve/rust-early-name-resolver.cc (EarlyNameResolver::visit): Added check which emits error instead of using assertion.
	* typecheck/rust-hir-type-check-pattern.cc (TypeCheckPattern::visit): Added rich location and error code.

**gcc/testsuite/ChangeLog:**

	* rust/compile/match5.rs: Updated comment for dejagnu.
	* rust/compile/pattern-struct.rs: New test for ICE.

### DIFF
--- a/gcc/rust/ast/rust-pattern.h
+++ b/gcc/rust/ast/rust-pattern.h
@@ -1176,12 +1176,7 @@ public:
 
   void accept_vis (ASTVisitor &vis) override;
 
-  // TODO: seems kinda dodgy. Think of better way.
-  std::unique_ptr<TupleStructItems> &get_items ()
-  {
-    rust_assert (has_items ());
-    return items;
-  }
+  std::unique_ptr<TupleStructItems> &get_items () { return items; }
 
   PathInExpression &get_path () { return path; }
   const PathInExpression &get_path () const { return path; }
--- a/gcc/rust/typecheck/rust-hir-type-check-pattern.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-pattern.cc
@@ -76,14 +76,16 @@ TypeCheckPattern::visit (HIR::TupleStructPattern &pattern)
     }
 
   // error[E0532]: expected tuple struct or tuple variant, found struct variant
-  // `Foo::D`
+  // `Foo::D`, E0532 by rustc 1.49.0 , E0164 by rustc 1.71.0
   if (variant->get_variant_type () != TyTy::VariantDef::VariantType::TUPLE)
     {
       std::string variant_type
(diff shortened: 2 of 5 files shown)


gcc/rust/typecheck/rust-hir-type-check-pattern.cc
```diff
@@ -76,14 +76,16 @@ TypeCheckPattern::visit (HIR::TupleStructPattern &pattern)
...
   if (variant->get_variant_type () != TyTy::VariantDef::VariantType::TUPLE)
     {
       std::string variant_type
 	= TyTy::VariantDef::variant_type_string (variant->get_variant_type ());
 
+      rich_location rich_locus (line_table, pattern.get_locus ());
+      rich_locus.add_fixit_replace ("not a tuple struct or tuple variant");
       rust_error_at (
-	pattern.get_locus (),
+	rich_locus, ErrorCode::E0164,
```
MahadMuhammad, line 88: Here, rustc 1.49.0 emits `E0532` &  rustc 1.71.0 emits `E0164`

philberty: Good catch. Go with the latest one

gcc/rust/ast/rust-pattern.h
```diff
@@ -1176,12 +1176,7 @@ class TupleStructPattern : public Pattern
 
   void accept_vis (ASTVisitor &vis) override;
 
-  // TODO: seems kinda dodgy. Think of better way.
-  std::unique_ptr<TupleStructItems> &get_items ()
-  {
-    rust_assert (has_items ());
-    return items;
-  }
+  std::unique_ptr<TupleStructItems> &get_items () { return items; }
```
philberty, line 1179: Good find @dkm has been fixing these too. Keep your eye out for them :D

CohenArthur (approved):

gcc/rust/typecheck/rust-hir-type-check-pattern.cc
MahadMuhammad: Added the latest one :)

PR #2570, opened 2023-08-19
Fork ```c++``` resorting methods

powerboat9:
This should help towards integrating ```backend/rust-tree.c``` and ```backend/rust-tree.h``` with gengtype, as an eventual step towards stripping down some of the data structures we have vestigially from the c++ frontend.


powerboat9:
Now depends on #2572

philberty (approved):
LGTM

PR #2575, opened 2023-08-21
gccrs: Fix compilation of types which hold onto dangling infer vars

philberty:
There is a case where some generic types are holding onto inference variable pointers directly. So this gives the backend a chance to do one final lookup to resolve the type.

This now allows us to compile a full test case for iterators but there is still one miscompilation in here which results in a segv on O2 and bad result on -O0.

Addresses #1895

gcc/rust/ChangeLog:

	* backend/rust-compile-type.cc (TyTyResolveCompile::visit): do a final lookup

gcc/testsuite/ChangeLog:

	* rust/compile/iterators1.rs: New test.

### DIFF
--- a/gcc/testsuite/rust/compile/iterators1.rs
+++ b/gcc/testsuite/rust/compile/iterators1.rs
@@ -1,0 +1,551 @@
+#![feature(intrinsics)]
+
+pub use option::Option::{self, None, Some};
+pub use result::Result::{self, Err, Ok};
+
+extern "C" {
+    fn printf(s: *const i8, ...);
+}
+
+mod option {
+    pub enum Option<T> {
+        None,
+        Some(T),
+    }
+}
+
+mod result {
+    enum Result<T, E> {
+        Ok(T),
+        Err(E),
+    }
+}
+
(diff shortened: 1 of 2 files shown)


gcc/testsuite/rust/compile/iterators1.rs
```diff
@@ -0,0 +1,551 @@
...
+    fn next(&mut self) -> Option<A> {
+        if self.start < self.end {
+            // We check for overflow here, even though it can't actually
+            // happen. Adding this check does however help llvm vectorize loops
+            // for some ranges that don't get vectorized otherwise,
+            // and this won't actually result in an extra check in an optimized build.
+            match self.start.add_usize(1) {
+                Option::Some(mut n) => {
+                    mem::swap(&mut n, &mut self.start);
+                    Option::Some(n)
```
CohenArthur, line 504: the original code uses `mem::replace` to get the "old value" before updating `self.start`. so this should be changed to something like:

```rust
let tmp = self.start;
self.start = n;
Option::Some(tmp)
```

Issue #2431, opened 2023-07-16
Missing AltPattern typechecking


powerboat9:
I tried this code:

```rust
pub fn main() {
    match 12 {
        12 | 13 | 14 => {}
        _ => {}
    }
}
```

I expected to see this happen: no error

Instead, this happened:

```
<source>:3:9: sorry, unimplemented: type checking alternate patterns not supported
    3 |         12 | 13 | 14 => {}
      |         ^~
```

PR #2581, opened 2023-08-24
AltPattern typechecking [WIP]

dme2:
Fix typechecking (and compilation) error for alt patterns in match expressions.

Addresses #2431


gcc/rust/typecheck/rust-hir-type-check-pattern.cc
```diff
@@ -466,8 +466,15 @@ TypeCheckPattern::typecheck_range_pattern_bound (
...
 TypeCheckPattern::visit (HIR::AltPattern &pattern)
 {
-  rust_sorry_at (pattern.get_locus (),
-		 "type checking alternate patterns not supported");
+  const auto &alts = pattern.get_alts ();
+
+  for (auto &alt_pattern : alts)
+    {
+      // do something with the result?
+      auto result = TypeCheckPattern::Resolve (alt_pattern.get (), parent);
```
philberty, line 474: What are the rules with alt patterns are the types all meant to match? We do need a least upper bound coercion helper which is missing at the moment but you could copy what we do for array expressions at least for now and raise an issue and a TODO comment.

Ideally it should look like this:
```
std::vector<TyTy::BaseType*> alts;

for each pattern {
auto result = TypeCheckPattern::Resolve (alt_pattern.get (), parent);
  alts.push(result);

}

lub_coercion(alts, pattern.get_locus (), pattern.get_mappings().get_hirid())
```

But yeah that lub_coercion helper doesnt exist right now. See:  https://github.com/Rust-GCC/gccrs/blob/8e2e7fa0211f49fe8f4c8b156468320ccfa711a0/gcc/rust/typecheck/rust-hir-type-check-expr.cc#L933C9-L943

its not a perfect lub, it should really be coercion site but its ok for now.

philberty:
@dme2 good job with this PR automation is only failing due to the unused result. Do you want to have a go at copying over the poor mans least upper bound coercion from our array expression code. Or do you want to fix the warning merge now and raise the issue?

philberty:
Also you will need to rebase your PR

dme2:
> @dme2 good job with this PR automation is only failing due to the unused result. Do you want to have a go at copying over the poor mans least upper bound coercion from our array expression code. Or do you want to fix the warning merge now and raise the issue?

I think I'll try copying over that array expression code.

philberty:
@dme2 this patch looks good to me, can you rebase your branch and mark it ready for review when your ready? As far as I can tell from a glance it looks good to merge

PR #2584, opened 2023-08-28
Fix iterator intrinsics

### LINKED ISSUE #2583: Bad uninit intrinsic
I tried this code:

```rust
#[lang = "sized"]
pub trait Sized {}

mod intrinsics {
    extern "rust-intrinsic" {
        pub fn uninit<T>() -> T;
    }
}

pub fn main() -> usize {
    unsafe { intrinsics::uninit() }
}
```

I expected to see this happen: main returns 1

Instead, this happened: segv on optimized builds and random return values
### Meta

- What version of Rust GCC were you using, git sha if possible.c727e23596ad62b441b9dea655b13f62caf74092

philberty:
This patch set it to fix the bad intrinsic's causing issues with the iterator code.

Fixes https://github.com/Rust-GCC/gccrs/issues/2583

### DIFF
--- a/gcc/testsuite/rust/execute/torture/issue-2583.rs
+++ b/gcc/testsuite/rust/execute/torture/issue-2583.rs
@@ -1,0 +1,13 @@
+#[lang = "sized"]
+pub trait Sized {}
+
+mod intrinsics {
+    extern "rust-intrinsic" {
+        pub fn uninit<T>() -> T;
+    }
+}
+
+pub fn main() -> i32 {
+    let _val: usize = unsafe { intrinsics::uninit() };
+    0
+}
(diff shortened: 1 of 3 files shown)


gcc/testsuite/rust/execute/torture/issue-2583.rs
```diff
@@ -0,0 +1,13 @@
...
+
+mod intrinsics {
+    extern "rust-intrinsic" {
+        pub fn uninit<T>() -> T;
+    }
+}
+
+pub fn main() -> usize {
+    let val: usize = unsafe { intrinsics::uninit() };
+    val - 1
```
CohenArthur, line 12: should we guarantee that `val` will be initialized to `1` here? is that something that Rust guarantees?

CohenArthur (approved):

gcc/testsuite/rust/execute/torture/issue-2583.rs
philberty: Good point

philberty: because actually all i care about is that this test case doesn't segv at runtime

PR #2585, opened 2023-08-28
Rename libproc_macro to libproc_macro_internal

P-E-P:
The name was a bit misleading since libproc_macro is a rust library that should we should be able to link against some rust code. This cpp library was the layer below the libproc_macro library, the "real" gcc libproc_macro act as a translation layer for the internal library.


CohenArthur (approved):
LGTM!

PR #2587, opened 2023-08-29
Retrieve proc macro definition

P-E-P:
Proc macro are resolved but not expanded. This PR allows the code to properly retrieve the macro definition in order for it to be expanded.

Requires #2496


CohenArthur (approved):
Looks great! Thank you!

PR #2589, opened 2023-09-01
Move type-related functions into base class ```Backend```

powerboat9:


philberty (approved):
LGTM thanks for continuing to do this!!!

PR #2591, opened 2023-09-01
Unify raw attribute values

P-E-P:
Maybe controversial changes, review carefully and consider all implications, this is meant to remove raw attribute values. Hopefully this will make the code less prone to typos.

### DIFF
--- a/gcc/rust/ast/rust-ast.cc
+++ b/gcc/rust/ast/rust-ast.cc
@@ -27,6 +27,7 @@ along with GCC; see the file COPYING3.  If not see
 #include "rust-parse.h"
 #include "rust-operators.h"
 #include "rust-dir-owner.h"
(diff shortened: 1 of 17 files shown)


philberty (approved):
I like this change so long as we have a way to represent unknow attributes for the case of new ones being added

P-E-P:
> I like this change so long as we have a way to represent unknow attributes for the case of new ones being added

Raw values are still supported, simply discouraged. Keeping those values in constexpr allows multiple things:

- No more typos in the attribute name: you target it using its variable name
- Your linter will help you (if you have one):
    - It knows the variables and suggest the names
    - It can rename a variable everywhere
- We can easily rename an attribute without having to grep & replace through the whole project

CohenArthur (approved):

PR #2597, opened 2023-09-02
[E0603] Use of private item outside scope

MahadMuhammad:
## Use of private item outside scope - [`E0603`](https://doc.rust-lang.org/error_codes/E0603.html) 
- Use of private item outside scope
- Updated error message.

gcc/rust/ChangeLog:

	* checks/errors/privacy/rust-privacy-reporter.cc (PrivacyReporter::check_for_privacy_violation): Added errorcode & richlocation.

gcc/testsuite/ChangeLog:

	* rust/compile/privacy1.rs: Updated comment for dejagnu.
	* rust/compile/privacy3.rs: likewise.
	* rust/compile/privacy4.rs: likewise.
	* rust/compile/privacy5.rs: likewise.

### DIFF
--- a/gcc/rust/checks/errors/privacy/rust-privacy-reporter.cc
+++ b/gcc/rust/checks/errors/privacy/rust-privacy-reporter.cc
@@ -166,7 +166,12 @@ PrivacyReporter::check_for_privacy_violation (const NodeId &use_id,
     }
 
   if (!valid)
-    rust_error_at (locus, "definition is private in this context");
+    {
+      rich_location richloc (line_table, locus);
+      richloc.add_fixit_replace ("item is private");
+      rust_error_at (richloc, ErrorCode::E0603,
+		     "definition is private in this context");
+    }
 }
 
 void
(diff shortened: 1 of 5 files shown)


gcc/rust/checks/errors/privacy/rust-privacy-reporter.cc
```diff
@@ -166,7 +166,12 @@ PrivacyReporter::check_for_privacy_violation (const NodeId &use_id,
     }
 
   if (!valid)
-    rust_error_at (locus, "definition is private in this context");
+    {
+      rich_location richloc (line_table, locus);
+      richloc.add_fixit_replace ("function is private");
```
philberty, line 171: I think the error message should be more general and say item is private from glancing at the testcase changes they are not always function item privacy voilations

philberty (changes requested):

gcc/rust/checks/errors/privacy/rust-privacy-reporter.cc
MahadMuhammad: Updated the `error message` to the previous message, which was added by @CohenArthur.

philberty (approved):

Issue #2461, opened 2023-07-20
Emit error on incorrect proc macro function signature.


P-E-P:
```
function-like proc macro has incorrect signature
```

PR #2603, opened 2023-09-04
Export proc macros in proc macro crates

P-E-P:
In order to fully support procedural macros we need to be able to compile those down to their binary representation (shared library). This means we should :

- [x]  Export procedural macro functions with the correct ABI
- [x] Mark those function as procedural macros
- [x] Collect procedural macros with another backend pass
- [x] Export the final array into gimple from collected macros

Note that #2461 should be handled separately in another PR.


gcc/rust/backend/rust-compile.cc
```diff
@@ -45,6 +49,391 @@ CompileCrate::go ()
...
+// structures that needs to get exposed in the final shared library for
+// procedural macro crates.
+//
+// The compiler needs some additional metadata to find which function correspond
+// to the desired macro. The library shall expose one entrypoint symbol leading
+// to those metadata which in turn lead to the correct function.
+// This namespace describes how to build and initialize those metadata
+// structures. Those structure should be kept in sync with the structures in
+// libproc_macro_internal/proc_macro.h describing how they should be read.
+namespace {
```
philberty, line 68: why the anon namespace?

gcc/rust/backend/rust-compile.cc
```diff
@@ -45,6 +49,391 @@ CompileCrate::go ()
...
+
+// Namespace containing all functions to build the different types.
+namespace build {
```
philberty, line 71: Not sure i like the namespace build i feel like these functions could be in a seperate file like rust-compile-proc-macro.cc

philberty (changes requested):
This looks good but i think those functions should be in a seperate file

gcc/rust/backend/rust-compile.cc
P-E-P: I didn't want to export all those functions outside of this TU.

gcc/rust/backend/rust-compile.cc
P-E-P: Yeah, I'm not entirely convinced myself. Do you have anything to suggest ?

P-E-P:
> This looks good but i think those functions should be in a seperate file

Done!

gcc/rust/backend/rust-compile-base.cc
```diff
@@ -108,7 +115,94 @@ HIRCompileBase::setup_fndecl (tree fndecl, bool is_main_entry_point,
...
+
+  // TODO: Should we rely on fixed index ? Should we search for the
+  // attribute tokentree instead ?
+  if (tt.get_token_trees ().size () > 3)
+    {
+      rust_assert (tt.get_token_trees ()[3]->as_string () == "attributes");
+
+      auto attributes = static_cast<const AST::DelimTokenTree *> (
+	tt.get_token_trees ()[4].get ());
+      auto &token_trees = attributes->get_token_trees ();
```
CohenArthur, line 174: I think it'd be nice to have a little bit of documentation here explaining why we choose 3 and 4. something showing what the attribute looks like and what the token vector looks like for example

gcc/rust/backend/rust-compile-base.cc
```diff
@@ -108,7 +115,94 @@ HIRCompileBase::setup_fndecl (tree fndecl, bool is_main_entry_point,
...
+}
+
+static std::string
+get_trait_name (const AST::Attribute &attr)
+{
+  rust_assert (attr.get_attr_input ().get_attr_input_type ()
+	       == Rust::AST::AttrInput::TOKEN_TREE);
+  const auto &tt
+    = static_cast<const AST::DelimTokenTree &> (attr.get_attr_input ());
+  return tt.get_token_trees ()[1]->as_string ();
```
CohenArthur, line 193: same here

CohenArthur (approved):
looks great! I'm excited to see this land haha

gcc/rust/backend/rust-compile-base.cc
P-E-P: Fixed in [b5ca1e4](https://github.com/Rust-GCC/gccrs/pull/2603/commits/b5ca1e4d9d85c991283fd1425af36fa628c1eb19).

gcc/rust/backend/rust-compile-proc-macro.cc
```diff
@@ -0,0 +1,393 @@
...
+  auto total_macros = ctx->get_attribute_proc_macros ().size ()
+		      + ctx->get_bang_proc_macros ().size ()
+		      + ctx->get_derive_proc_macros ().size ();
+
+  tree pm_type = build::proc_macro ();
+  tree pm_buffer_type = build::proc_macro_buffer (pm_type, total_macros);
+  tree entrypoint_type = build::entrypoint (pm_buffer_type);
+
+  std::string decl_symbol_name = generate_proc_macro_decls_symbol (
+    0 /* FIXME: Change to stable crate id */);
```
philberty, line 364: You should be able to get mappings->get_current_crate for that.

P-E-P: Are your sure we're talking about the same crate id ? IIRC the stable crate id is made from the hash of the crate's metadata. I opened #2696 for this a while ago, but since you wanted to rework the metadata format I left it as is while we wait for this rework.

PR #2609, opened 2023-09-06
Convert class ```Backend``` into namespace

powerboat9:
Depends on #2440


CohenArthur (approved):

philberty (approved):
Awesome work!!!! I am not a fan of the name backend for the namespace but its whats been there for ages now. 

Can you think of any new names to call this?

Also I think it would be nice if we can move the rust-gcc.cc and rust-backend.h code into the backend pass folder.

Also does it make sense to rename rust-backend.h into rust-gcc.h now?

Some other front-end use the name genericify or gimplify might be a name there we could use.

PR #2614, opened 2023-09-09
Add `try` intrinsics

liushuyu:
This pull request adds the `try` intrinsic, making it possible to handle `panic`s.

A new option, `-frust-panic` is added to specify the panic handling strategy.

Note that currently, only `try`-`catch` mechanism is added in this pull request.
To correctly handle the `panic` structures, an unwinder must also be implemented (in the runtime library).

### DIFF
--- a/gcc/rust/backend/rust-compile-intrinsic.cc
+++ b/gcc/rust/backend/rust-compile-intrinsic.cc
@@ -23,6 +23,7 @@
 #include "rust-diagnostics.h"
 #include "rust-location.h"
 #include "rust-constexpr.h"
+#include "rust-session-manager.h"
 #include "rust-tree.h"
 #include "tree-core.h"
 #include "rust-gcc.h"
@@ -194,6 +195,9 @@ expect_handler (bool likely)
   };
 }
 
+static tree
+try_handler (Context *ctx, TyTy::FnType *fntype);
+
 inline tree
 sorry_handler (Context *ctx, TyTy::FnType *fntype)
 {
@@ -241,6 +245,7 @@ static const std::map<std::string,
     {"likely", expect_handler (true)},
     {"unlikely", expect_handler (false)},
     {"assume", assume_handler},
+    {"try", try_handler},
 };
 
 Intrinsics::Intrinsics (Context *ctx) : ctx (ctx) {}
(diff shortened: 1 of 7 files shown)


dkm:
You are also adding support for `__builtin_expect`

gcc/rust/backend/rust-compile-intrinsic.cc
```diff
@@ -1104,5 +1135,116 @@ move_val_init_handler (Context *ctx, TyTy::FnType *fntype)
...
+
+  bool panic_is_abort = Session::get_instance ().options.get_panic_strategy ()
+			== CompileOptions::PanicStrategy::Abort;
+  tree try_fn = Backend::var_expression (param_vars[0], UNDEF_LOCATION);
+  tree user_data = Backend::var_expression (param_vars[1], UNDEF_LOCATION);
+  tree catch_fn = Backend::var_expression (param_vars[2], UNDEF_LOCATION);
+  tree normal_return_stmt
+    = Backend::return_statement (fndecl, integer_zero_node, BUILTINS_LOCATION);
+  tree error_return_stmt
+    = Backend::return_statement (fndecl, integer_one_node, BUILTINS_LOCATION);
```
CohenArthur, line 1210: did you find these values in the rustc source? could it be possible to add the link to the source here? :)

gcc/rust/backend/rust-compile-intrinsic.cc
```diff
@@ -1104,5 +1135,116 @@ move_val_init_handler (Context *ctx, TyTy::FnType *fntype)
...
+
+  tree catch_block = Backend::block (fndecl, enclosing_scope, {},
+				     UNDEF_LOCATION, UNDEF_LOCATION);
+  Backend::block_add_statements (catch_block,
+				 std::vector<tree>{catch_call,
+						   error_return_stmt});
+  // TODO(liushuyu): eh_personality needs to be implemented as a runtime thing
+  auto eh_construct
+    = Backend::exception_handler_statement (try_block, catch_block, NULL_TREE,
+					    BUILTINS_LOCATION);
```
CohenArthur, line 1241: this looks good to me, but I have no real idea of what the `try` intrinsic does or how it is implemented. I also wonder if this is how `rustc_codegen_gcc` handles it? @antoyo @GuillaumeGomez

gcc/rust/backend/rust-compile-intrinsic.cc
```diff
@@ -717,7 +746,9 @@ copy_nonoverlapping_handler (Context *ctx, TyTy::FnType *fntype)
     = build2 (MULT_EXPR, size_type_node, TYPE_SIZE_UNIT (param_type), count);
 
   tree memcpy_raw = nullptr;
-  BuiltinsContext::get ().lookup_simple_builtin ("memcpy", &memcpy_raw);
+  BuiltinsContext::get ().lookup_simple_builtin (overlaps ? "memmove"
+							  : "memcpy",
+						 &memcpy_raw);
```
CohenArthur, line 751:
```suggestion
						 &memcpy_raw);
```

nit but this variable should probably be renamed since it can be either `memmove` or `memcpy` :)

CohenArthur (approved):
this PR looks great to me, thank you @liushuyu! unfortunately I'm not super well versed in our backend so I'll let someone else review it before merging it :)

gcc/rust/backend/rust-compile-intrinsic.cc
GuillaumeGomez: We do it [here](https://github.com/antoyo/gcc/blob/master/gcc/jit/jit-playback.cc#L2343) if you want to take a look.

antoyo: Our implementation of panics doesn't work correctly with optimizations enabled, though. Not sure why: I still need to investigate.

philberty (approved):
I am in favor of merging this sooner than later its a great starting point

philberty:
@liushuyu are you able to rebase this work i think we should merge this as a starting point.

liushuyu:
Hi @philberty @CohenArthur:
You might want to re-review this pull request as I have made additional changes to account for Rust standard library API changes since 1.78.

I have also added two unit tests for testing the code generation; since the fine people in the `rustc_cg_gcc` community had figured out how to prevent GCC from optimizing away the `try {...} catch {...}` blocks, I just copied and pasted their entire homework here.

philberty (approved):
LGTM great work I am happy with this

PR #2617, opened 2023-09-11
Remove unused `ctx` argument in `make_unsigned_long_tree` function

GuillaumeGomez:
As discussed in https://github.com/Rust-GCC/gccrs/pull/2612. The `ctx` argument was unused so I removed it.


CohenArthur (approved):

CohenArthur:
Will merge this before #2612 since the code is still valid and `ctx` was already unused. Thanks a lot @GuillaumeGomez !

PR #2619, opened 2023-09-12
Add ```assume``` intrinsic

powerboat9:
I've been having trouble getting this to work so far, optimizations don't seem to be taking the assumption into account. This *might* change after gcc upstream is merged in, but there's probably something off with my code.

### DIFF
--- a/gcc/rust/backend/rust-compile-intrinsic.cc
+++ b/gcc/rust/backend/rust-compile-intrinsic.cc
@@ -32,6 +32,11 @@
 
 #include "print-tree.h"
 
+// declaration taken from "stringpool.h"
+// the get_identifier macro causes compilation issues
+extern tree
+get_identifier (const char *);
+
 namespace Rust {
 namespace Compile {
 
@@ -83,6 +88,8 @@ static tree
 uninit_handler (Context *ctx, TyTy::FnType *fntype);
 static tree
 move_val_init_handler (Context *ctx, TyTy::FnType *fntype);
+static tree
+assume_handler (Context *ctx, TyTy::FnType *fntype);
 
 enum class Prefetch
 {
@@ -231,6 +238,7 @@ static const std::map<std::string,
     {"move_val_init", move_val_init_handler},
     {"likely", expect_handler (true)},
     {"unlikely", expect_handler (false)},
+    {"assume", assume_handler},
 };
 
 Intrinsics::Intrinsics (Context *ctx) : ctx (ctx) {}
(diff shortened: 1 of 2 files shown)


dkm:
You can see if the assume hint is propagated by looking at the dump. Have you tried? For example with C: https://c.godbolt.org/z/fha6cxjGq
you can see the split for the branch is 90/10 instead of 50/50 without the builtin call.

powerboat9:
I'm not using ```__builtin_expect```, I'm using ```.ASSUME``` (```__attribute__((assume(x)))```) which gets converted to an if statement and ```__builtin_unreachable``` call.

powerboat9:
Looking at the dumps, the function call to the function containing ```.ASSUME``` isn't being inlined right.

dkm:
> I'm not using `__builtin_expect`, I'm using `.ASSUME` (`__attribute__((assume(x)))`) which gets converted to an if statement and `__builtin_unreachable` call.

oh right, of course you're not using `expect`... I mixed your changes and the one adding `likely` based on `expect` in another PR, sorry.

powerboat9:
I'm not really sure how I should even expect the compiler to handle assume statements during optimization, so I think I'll keep this as-is with a few ```TODO``` statements added.

dkm:
Isn't it possible to have something similar to this: https://c.godbolt.org/z/hf538n1qf to check it is working as expected?
This simple rust example also seems to show that dead code elimination correctly uses the hint in rustc: https://rust.godbolt.org/z/qrh65a4Ex

philberty:
I think lets merge it and see how we go as it will give us a point of reference. I am worried that because we don't auto inline intrinsic maybe gcc isn't going to optimize as aggressively unless you use -O2.

I have a patch to experiment with inlining some intrinsic but its messy, because semanticly they are HIR::CallExpr's, keeping it this way is more semanticaly accurate version of the program.

Its personally a pet peeve of mine that Rustc is so eagerly optimizing things at the front-end level just to keep stack trace info nicer when its not an accurate representation of whats going on and GCC will be able to inline these calls for us anyway.

philberty (approved):
LGTM

gcc/rust/backend/rust-compile-intrinsic.cc
```diff
@@ -1174,5 +1182,50 @@ expect_handler_inner (Context *ctx, TyTy::FnType *fntype, bool likely)
...
+  tree lookup = NULL_TREE;
+  if (check_for_cached_intrinsic (ctx, fntype, &lookup))
+    return lookup;
+
+  auto fndecl = compile_intrinsic_function (ctx, fntype);
+
+  // TODO: make sure these are necessary
+  TREE_READONLY (fndecl) = 0;
+  DECL_DISREGARD_INLINE_LIMITS (fndecl) = 1;
+  DECL_ATTRIBUTES (fndecl) = tree_cons (get_identifier ("always_inline"),
```
philberty, line 1203: we should be doing this for all our intrinsics can  you add this to thee compile_intrinsic_function in another PR. We probably should avoid setting disregard inline limits for all of them though.

philberty: the always inline attrib is what i mean should be added for all of them

PR #2620, opened 2023-09-12
Add late name resolution pass

CohenArthur:
This is a massive draft PR which I'll split in multiple tinier ones

- expected: nodiscard
- optional: Add [[nodiscard]]
- late
- sesh: Add late name resolution 2.0
- session manager
- ast: change SimplePath, PathInExpression etc's API
- default: Add `visit_function_param` helper function
- rib
- toplevel
- ctx: Add labels foreverstack
- forever-stack: Fix basic get logic
- late
- foreverstack: Specialize `get` for Namespace::Labels
- forever stack: Fix resolve_path signature
- late: resolve path
- early: resolve path
- late: Add test
- nameresolutionctx: Store mappings in ctx
- early: [f | d]: Add macro invocations to mappings
- late: Start setting up builtin types
- late: builtins
- late: Start storing mappings properly in the resolver
- early
- nameresolutionctx: Keep map of resolved usages
- toplevel: Add comment about running the collector twice
- ast: Add NodeId to UseTree base class
- default resolver: Visit `UseDeclaration`s
- forever stack: Do not copy segment when dereferencing iterator in `find_starting_point` [other functions too?]
- early: Move usedec resolving to TopLevel
- toplevel: Start usedec resolution
- default: [f] it to [d] the other
- toplevel: Handle use declarations properly
- session-manager: [f] STart dumping name reoslution pass
- toplevel: Resolve use decls more properly
- toplevel: Start cleanup of import handling
- add more name res test casaes
- immutable nrctx
- session manager: Init Immutable name resolver
- nrctx: add resolved lookup
- typecheck: Fetch ImmutableNRCtx
- foreverstack: Add to_canonical_path method
- foreverstack: Add to_rib method
- typecheck: Start using nr2.0 properly
- backend: Use new name resolver where necessary
- nr2.0: Start using newtype pattern for Usage and Definition
- rib: Add base for storing shadowable ids
- typecheck path: wip
- late: Setup builtin types properly
- rib: Add shadowable definitions on a per definition level
- forever stack: Use new Rib::Definition API
- toplevel: Add note about resolving glob imports
- default: Visit external functions properly
- compile: use new rib API
- type-check-type: Smol cleanup in format string


gcc/rust/backend/rust-compile-base.cc
```diff
@@ -624,13 +625,41 @@ HIRCompileBase::compile_function (
...
+  auto local_ids = std::vector<NodeId> ();
+
+  if (flag_name_resolution_2_0)
+    {
+      auto nr_ctx
+	= Resolver2_0::ImmutableNameResolutionContext::get ().resolver ();
+
+      auto candidate = nr_ctx.values.to_rib (body_mappings.get_nodeid ());
+
+      // FIXME: Ugly
```
dkm, line 641: :laughing:

P-E-P:
> optional: Add [[nodiscard]]
Isn't this C++17 ? Won't it break with 4.8 ?

CohenArthur:
> > optional: Add [[nodiscard]]
> 
> Isn't this C++17 ? Won't it break with 4.8 ?

yeah, this does break it. I'll remove the commits. it's very useful for testing however and works in C++11, the compiler just complains that this attribute is in C++17 and still applies it

CohenArthur:
@matthewjasper I opened issues for all your comments, but I don't think I'll be addressing them in this PR as the goal for this one is to not add any regressions, and thus it can be in an incomplete state. thanks a lot for having taken the time to review it :)

P-E-P:
@CohenArthur Should we close this PR ?

PR #2622, opened 2023-09-13
Remove unused complex number support

powerboat9:
Since rust doesn't have complex types, these functions can be removed


powerboat9:
This should probably wait for after #2610

PR #2627, opened 2023-09-17
[E0617] Attempt for invalid type variable in `variadic` function

MahadMuhammad:
## Attempt to pass `invalid type` variable into a `variadic` function - [`E0617`](https://doc.rust-lang.org/error_codes/E0617.html)

There are some certain rust types must be cast before passing them to a variadic function, because of arcane ABI rules dictated by the C standard. To fix the error, cast the value to the type specified by the error message.

---

gcc/rust/ChangeLog:

	* typecheck/rust-tyty-call.cc (TypeCheckCallExpr::visit): Added ErrorCode & more fixit hints.

gcc/testsuite/ChangeLog:

	* rust/compile/variadic.rs: Added new checks.

---

### DIFF
--- a/gcc/rust/typecheck/rust-tyty-call.cc
+++ b/gcc/rust/typecheck/rust-tyty-call.cc
@@ -184,7 +184,10 @@ TypeCheckCallExpr::visit (FnType &type)
 		if ((int_ty.get_int_kind () == TyTy::IntType::IntKind::I8)
 		    || (int_ty.get_int_kind () == TyTy::IntType::IntKind::I16))
 		  {
-		    rust_error_at (arg_locus,
+		    rich_location richloc (line_table, arg_locus);
+		    richloc.add_fixit_replace (
+		      "cast the value to c_int: as c_int");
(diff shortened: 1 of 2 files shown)


gcc/rust/typecheck/rust-tyty-call.cc
```diff
@@ -184,7 +184,10 @@ TypeCheckCallExpr::visit (FnType &type)
 		if ((int_ty.get_int_kind () == TyTy::IntType::IntKind::I8)
 		    || (int_ty.get_int_kind () == TyTy::IntType::IntKind::I16))
 		  {
-		    rust_error_at (arg_locus,
+		    rich_location richloc (line_table, arg_locus);
+		    richloc.add_fixit_replace (
+		      "cast the value to c_int: as c_int");
+		    rust_error_at (richloc, ErrorCode::E0617,
```
MahadMuhammad, line 190: Combining all of these into one function with different messages might seem like a bit overkill, but if it works better, I'll do that.

philberty (approved):
LGTM

PR #2633, opened 2023-09-20
Fix CanonicalPath for impl

### LINKED ISSUE #2631: Improper `CanonicalPath` for impl
Gccrs treats `a::S::impl_func` as `Resolver::CanonicalPath` for `impl_func` in the code below,
but it is not correct according to https://doc.rust-lang.org/reference/paths.html#canonical-paths .
It should be `<a::S>::impl_func` instead. 

It becomes a problem when the v0 mangling is implemented  because 
the v0 encodes path segments for types (`T`) and ones for inherent impl (`<T>`) in a different way.

```rust
// a.rs
struct S;

impl S {
    fn impl_func() {}
}
```

tamaroning:
Fix #2631 

```
gcc/rust/ChangeLog:

	* util/rust-canonical-path.h: Add new segment kind for inherent impl.
	* resolve/rust-ast-resolve-item.cc (ResolveItem::visit): Use it.
	* resolve/rust-ast-resolve-toplevel.h: Use it.
```

### DIFF
--- a/gcc/rust/resolve/rust-ast-resolve-item.cc
+++ b/gcc/rust/resolve/rust-ast-resolve-item.cc
@@ -565,6 +565,7 @@ ResolveItem::visit (AST::InherentImpl &impl_block)
 
   // FIXME this needs to be protected behind nominal type-checks see:
   // rustc --explain E0118
+  // issue #2634
   ResolveType::go (impl_block.get_type ().get ());
 
   // Setup paths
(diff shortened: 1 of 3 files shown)


tamaroning:
I noticed problems remain. please do not review now.

tamaroning:
> Done.

P-E-P (approved):
LGTM!

Issue #2358, opened 2023-07-02
Unnecessary "empty token vector" ```FIXME```


powerboat9:
https://github.com/Rust-GCC/gccrs/blob/cfc51d46fcdb70bde84f030557c96866f3b18e23/gcc/rust/expand/rust-macro-builtins.cc#L580-L581

There are a few more ```FIXME``` comments similar to the above in the same file

rizan21:
Hey @CohenArthur, can I clean this up?

powerboat9:
I'm alright with it, but @CohenArthur might have more details à la Chesterton's Fence

rizan21:
Sure, I'll wait for his response

CohenArthur:
@rizan21 sure, go ahead! I've assigned the issue to you :)

rizan21:
Hey, @CohenArthur, Ive started a PR #2644, Please review

officialhemant511:
return AST::Fragment(node, std::move(str_tok));

By removing the curly braces around node and std::move(str_tok), the arguments will be passed properly to the AST::Fragment constructor

CohenArthur:
Closed by #2644

PR #2640, opened 2023-10-02
Remove unecessary "empty vector" comments

rizan21:
Addresses #2358
- Not sure if checks should be added in place of the comments 

---

- \[X] GCC development requires copyright assignment or the Developer's Certificate of Origin sign-off, see https://gcc.gnu.org/contribute.html or https://gcc.gnu.org/dco.html
- \[X] Read contributing guidlines
- \[X] `make check-rust` passes locally
- \[ ] Run `clang-format`

---

Removed unnecessary comments

### DIFF
--- a/gcc/rust/backend/rust-compile.cc
+++ b/gcc/rust/backend/rust-compile.cc
@@ -115,8 +115,8 @@ HIRCompileBase::coercion_site1 (tree rvalue, TyTy::BaseType *rval,
       if (!valid_coercion)
 	return error_mark_node;
 
-      const TyTy::ReferenceType *exp
-	= static_cast<const TyTy::ReferenceType *> (expected);
+      const TyTy::PointerType *exp
+	= static_cast<const TyTy::PointerType *> (expected);
 
       TyTy::BaseType *actual_base = nullptr;
       if (actual->get_kind () == TyTy::TypeKind::REF)


dkm:
You don't need to add more test, as the change are unlikely to break any code checks.
BUT, you still need to come with a correct git commit log message. 
You can check for other commit doing similar changes to comments (look for typo check as a good example).

You can see above that everything is fine, except for the commit message that lacks a GNU ChangeLog.

You also need to state that you have read and understood the https://gcc.gnu.org/dco.html in the description of this Pull Request, and add a Signed-off-by line to your commit message. Given the nature of your change, it's not mandatory, but it will make our life easier to have it.

Thanks!

powerboat9:
Try taking a look at, and running all/part of, [gcc-git-customization.sh](https://github.com/Rust-GCC/gccrs/blob/master/contrib/gcc-git-customization.sh). There's an alias called ```gcc-commit-mklog``` that should automate part of changelog generation. You should probably also read https://www.gnu.org/prep/standards/standards.html#Change-Logs and https://gcc.gnu.org/codingconventions.html#ChangeLogs.

CohenArthur (commented):
thanks! the changes look good, but I agree with what Marc has said - the formatting of commits is important as we send our commits to GCC afterwards without modification. if you need help getting your commit formatted properly let me know on Zulip!

rizan21:
I'll get working on that asap 

P.S. i will need help

rizan21:
mb .. took longer than expected

PR #2642, opened 2023-10-02
Fix type confusion in coercion

### LINKED ISSUE #2641: Type confusion error
https://github.com/Rust-GCC/gccrs/blob/8da5a49a0ee5842c4abd0b079ee837290c420814/gcc/rust/backend/rust-compile.cc#L119  Checks that TyTy is a pointer but casts it to reference.

jdupak:
Fixes #2641

There was a mismatch between a manual discriminant test and the static cast.

gcc/rust/ChangeLog:

	* backend/rust-compile.cc (HIRCompileBase::coercion_site1): Fix wrong cast

- \[x] Run `clang-format`


philberty (approved):
LGTM nice find

PR #2674, opened 2023-10-10
Fix macro "default" call at item level

### LINKED ISSUE #2655: Macro call named `default`
The compiler refutes a macro call named `default` at file scope.

I tried this code:

```rust
macro_rules! default {
    ($($x:tt)*) => { $($x)* }
}

default! {
    struct A;
}
```

I expected to see this happen: Compiles correctly.

Instead, this happened:

```
test.rs:5:1: error: ‘default’ is only allowed on items within ‘impl’ blocks
    5 | default! {
      | ^~~~~~~
test.rs:5:1: error: failed to parse item in crate
```

### Meta

c7728c8020334fecb47789163ba94326aab6208c

P-E-P:
A macro `default` called at item level was previously rejected.

Fixes #2655


CohenArthur (approved):
good catch

PR #2675, opened 2023-10-10
Accept `macro_rules` name for a macro rule.

### LINKED ISSUE #2651: Macro rule with macro_rules name
Macro rule macro rule is not parsed correctly

I tried this code:

```rust
macro_rules! macro_rules { () => { struct S; } }
macro_rules! {} // calls the macro defined above

fn main() {
    let s = S;
}
```

I expected to see this happen: Compile correctly

Instead, this happened:

```
./test.rs:5:14: error: expecting ‘identifier’ but ‘{’ found
    5 | macro_rules! {} // OK, calls the macro defined above
      |              ^
./test.rs:5:14: error: failed to parse item in crate
```

### Meta

https://github.com/Rust-GCC/gccrs/commit/007166ca6ceecd429f7033336af25dd679424ca2

P-E-P:
This fix allows the user to use the name `macro_rules` for it's macros.

Fixes #2651

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -1070,6 +1070,18 @@ Parser<ManagedTokenSource>::parse_token_tree ()
     }
 }
 
+template <typename ManagedTokenSource>
+bool
+Parser<ManagedTokenSource>::is_macro_rules_def (const_TokenPtr t)
+{
+  auto macro_name = lexer.peek_token (2)->get_id ();
+
+  bool allowed_macro_name = (macro_name == IDENTIFIER || macro_name == TRY);
+
+  return t->get_str () == "macro_rules"
+	 && lexer.peek_token (1)->get_id () == EXCLAM && allowed_macro_name;
+}
+
 // Parses a single item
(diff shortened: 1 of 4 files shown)


CohenArthur (approved):
this is good, I am a bit scared about this test case which is valid Rust:

```rust
macro_rules! macro_rules {
    () => {}
}

macro_rules! foo {
    () => {}
}

foo!();

fn main() {}
```
but I don't know if we handle it properly

P-E-P:
> ```rust
> macro_rules! macro_rules {
>     () => {}
> }
> 
> macro_rules! foo {
>     () => {}
> }
> 
> foo!();
> 
> fn main() {}
> ```

We do. I can add your test to the testsuite it it puts your mind at rest.

CohenArthur:
> > ```
> > macro_rules! macro_rules {
> >     () => {}
> > }
> > 
> > macro_rules! foo {
> >     () => {}
> > }
> > 
> > foo!();
> > 
> > fn main() {}
> > ```
> 
> We do. I can add your test to the testsuite it it puts your mind at rest.

yeah I think that'd be good :) thanks!

PR #2676, opened 2023-10-10
Quick fix for ```MacroRuleDefinition``` use-after-free

powerboat9:
Macro expansion seems to destruct instances of MacroRulesDefinition in modules. This patch should prevent a use-after-free by cloning and then leaking MacroRulesDefinition instances.

### DIFF
--- a/gcc/rust/ast/rust-macro.h
+++ b/gcc/rust/ast/rust-macro.h
@@ -578,6 +578,11 @@ public:
 
   MacroKind get_kind () const { return kind; }
 
+  std::unique_ptr<MacroRulesDefinition> clone_macro_rules_def () const
+  {
+    return std::unique_ptr<MacroRulesDefinition> (clone_item_impl ());
+  }
+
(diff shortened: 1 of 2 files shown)


powerboat9:
Converting this to a draft for a bit, just to make sure I understand the issue

CohenArthur:
do you have a Rust example where this causes a use-after-free? ASAN will probably be helpful with this, if you'd like I can send in my configure line and `make` command to use ASAN without the build crashing at the first memory leak in g++

powerboat9:
> do you have a Rust example where this causes a use-after-free?

Not yet, that's why I converted this to a draft :P

powerboat9:
This seems to be an issue with `macroInvocations` entries pointing to freed `MacroRulesDefinition` instances, which doesn't currently seem to cause problems, as long as you're very careful with `#[macro_use]` handling

powerboat9:
I think I've figured this out

PR #2670, opened 2023-10-10
Fix #2667

powerboat9:
This also involves some changes to ```get_location``` and ```get_node_id``` method implementation, but I'll probably split those off into separate pull requests.

Fixes #2667

PR #2679, opened 2023-10-11
Create base class ```AssociatedItem```

powerboat9:
Taken from #2670

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -1446,53 +1446,18 @@ protected:
   }
 };
 
-// Item used in trait declarations - abstract base class
-class TraitItem : public Visitable
-{
-protected:
-  TraitItem (location_t locus)
-    : node_id (Analysis::Mappings::get ()->get_next_node_id ()), locus (locus)
-  {}
-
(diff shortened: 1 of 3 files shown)


dkm:
Is "deconstructor" a reference to something precise, or is it supposed to read "destructor"?

powerboat9:
It refers to [member functions which are called when the lifetime of an object ends](https://en.cppreference.com/w/cpp/language/destructor) :P

dkm:
> It refers to [member functions which are called when the lifetime of an object ends](https://en.cppreference.com/w/cpp/language/destructor) :P

Ok, so destructor, not deconstructor as you had in the commit log :) I was wondering if this was yet-another C++ concept I didn't know :)

philberty (changes requested):
I can see this being useful can you do the same for the HIR?

Need @CohenArthur sign off here too for this one. I like this idea

powerboat9:
I'd probably want to modify the HIR in another pull request -- it looks like the HIR representation of associated items is already slightly different from the AST representation

CohenArthur (approved):
looks great! thank you :D I like these changes

PR #2694, opened 2023-10-17
Fix token lexed as a float literal

### LINKED ISSUE #2659: Tuple index not parsed correctly
GCCRS finds a float literal instead of a tuple index

I tried this code:

```rust
fn main () {
    (1, (2, 3)).1.1;
}
```

I expected to see this happen: Code compiles

Instead, this happened:

```
./test.rs:2:16: error: failed to parse path expr segment of method call expr
    2 |     (1, (2, 3)).1.1;
      |                ^
./test.rs:2:5: error: failed to parse statement or expression in block expression
    2 |     (1, (2, 3)).1.1;
      |     ^
./test.rs:2:17: error: unrecognised token ‘float literal’ for start of item
    2 |     (1, (2, 3)).1.1;
      |                 ^~~
./test.rs:2:17: error: failed to parse item in crate
```

### Meta

P-E-P:
The lexer cannot distinguish the difference between a float literal and a tuple index in some cases. This means we should fix this while parsing depending on the context.

Fixes #2659


gcc/rust/expand/rust-macro-invoc-lexer.cc
```diff
@@ -49,6 +49,22 @@ MacroInvocLexer::split_current_token (TokenId new_left, TokenId new_right)
...
+{
+  rust_assert (new_tokens.size () > 0);
+
+  auto current_pos = token_stream.begin () + offs;
+
+  token_stream.erase (current_pos);
+
+  for (size_t i = 1; i < new_tokens.size (); i++)
+    {
+      token_stream.insert (current_pos + i, std::unique_ptr<AST::Token> (
```
CohenArthur, line 63: this is gonna start inserting at `current_pos + 1`, but we erase the one at `current_pos`. is that okay and what we want? I can never remember how insertion works in C++

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -12862,6 +12862,23 @@ Parser<ManagedTokenSource>::left_denotation (const_TokenPtr tok,
...
+	    lexer.split_current_token (
+	      {Token::make_int (current_loc, std::move (prefix),
+				CORETYPE_PURE_DECIMAL),
+	       Token::make (DOT, current_loc + 1),
+	       Token::make_int (current_loc + 2, std::move (suffix),
+				CORETYPE_PURE_DECIMAL)});
+	    return parse_tuple_index_expr (tok, std::move (left),
+					   std::move (outer_attrs),
+					   restrictions);
+	  }
```
CohenArthur, line 12881: can you add some documentation here and an explanation as to why this is needed?

PR #2701, opened 2023-10-18
Parse variadic functions

### LINKED ISSUE #2664: Cannot parse variadic functions
Cannot parse variadic functions even if those are disabled.

I tried this code:

```rust
fn main() {}

#[cfg(FALSE)]
fn variadic(x: isize, ...) {}
```

Instead, this happened: Failure on variadic ellipsis parsing.

```
./test.rs:4:23: error: unexpected token ‘...’ in pattern
    4 | fn variadic(x: isize, ...) {}
      |                       ^~~
./test.rs:4:23: error: failed to parse function param (in function params)
./test.rs:4:23: error: expecting ‘)’ but ‘...’ found
./test.rs:4:23: error: function declaration missing closing parentheses after parameter list
./test.rs:5:1: error: failed to parse item in crate
```

### Meta

### LINKED ISSUE #2700: Cannot parse named variadic on extern functions
Named variadics are not parsed correctly.

I tried this code:

```rust
extern "C" {
    fn variadic(x: isize, args: ...);
}

fn main() {}
```

I expected to see this happen: Compiles with a waning (unused function).

Instead, this happened:

Emit an error

```
./test.rs:2:33: error: unrecognised token ‘...’ in type
    2 |     fn variadic(x: isize, args: ...);
      |                                 ^~~
./test.rs:2:33: error: could not parse param type in extern block function declaration
./test.rs:2:27: error: could not parse named function parameter in external function
    2 |     fn variadic(x: isize, args: ...);
      |                           ^~~~
./test.rs:2:5: error: failed to parse external item despite not reaching end of extern block
    2 |     fn variadic(x: isize, args: ...);
      |     ^~
./test.rs:9:1: error: failed to parse item in crate
```

### Meta

- b3f02d9c8cb

P-E-P:
Parse variadic functions.

This version of the PR modifies the way variadic are stored in the ast. Variadic were previously handled at function level, but this makes the parser a bit ugly. This PR suggest keeping variadic informations as a special parameter instead. It requires a lot of work to reflect this change at all level of the compiler. Many passes consider the number of arguments without checking for variadicity, therefore, this PR creates a big pile of ICE as of [c6d9843](https://github.com/Rust-GCC/gccrs/pull/2701/commits/c6d984301f88b315137b61cb91a0053d5e3e0f46).

Alternative to #2704

Fixes #2664 
Fixes #2700

### DIFF
--- a/gcc/rust/ast/rust-item.h
+++ b/gcc/rust/ast/rust-item.h
@@ -4166,6 +4166,7 @@ class NamedFunctionParam
 
   NodeId node_id;
   location_t locus;
+  bool variadic;
 
 public:
   /* Returns whether the named function parameter has a name (i.e. name is not
@@ -4178,7 +4179,7 @@ public:
   bool is_error () const
   {
     // also if identifier is "" but that is probably more costly to compute
-    return param_type == nullptr;
+    return param_type == nullptr && !variadic;
   }
 
   std::string get_name () const { return name; }
@@ -4195,17 +4196,35 @@ public:
 		      std::vector<Attribute> outer_attrs, location_t locus)
     : name (std::move (name)), param_type (std::move (param_type)),
       outer_attrs (std::move (outer_attrs)),
-      node_id (Analysis::Mappings::get ()->get_next_node_id ()), locus (locus)
+      node_id (Analysis::Mappings::get ()->get_next_node_id ()), locus (locus),
+      variadic (false)
+  {}
+
+  NamedFunctionParam (std::string name, std::vector<Attribute> outer_attrs,
+		      location_t locus)
+    : name (std::move (name)), param_type (nullptr),
+      outer_attrs (std::move (outer_attrs)),
+      node_id (Analysis::Mappings::get ()->get_next_node_id ()), locus (locus),
+      variadic (true)
+  {}
+
+  NamedFunctionParam (std::vector<Attribute> outer_attrs, location_t locus)
+    : name (""), param_type (nullptr), outer_attrs (std::move (outer_attrs)),
+      node_id (Analysis::Mappings::get ()->get_next_node_id ()), locus (locus),
+      variadic (true)
   {}
 
   // Copy constructor
   NamedFunctionParam (NamedFunctionParam const &other)
-    : name (other.name), outer_attrs (other.outer_attrs)
+    : name (other.name), outer_attrs (other.outer_attrs),
+      variadic (other.variadic)
   {
     node_id = other.node_id;
     // guard to prevent null dereference (only required if error state)
     if (other.param_type != nullptr)
       param_type = other.param_type->clone_type ();
+    else
+      param_type = nullptr;
   }
 
   ~NamedFunctionParam () = default;
--- a/gcc/rust/ast/rust-item.h
+++ b/gcc/rust/ast/rust-item.h
@@ -1682,6 +1682,12 @@ public:
 
   void accept_vis (ASTVisitor &vis) override;
 
+  bool is_variadic () const
+  {
+    return function_params.size () != 0
+	   && function_params.back ().is_variadic ();
+  }
+
   // Invalid if block is null, so base stripping on that.
   void mark_for_strip () override { function_body = nullptr; }
   bool is_marked_for_strip () const override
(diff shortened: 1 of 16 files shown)


P-E-P:
Although these changes pass the test, they are not entirely complete. We should error out when the variadic argument is incorrect. This should be handled in another PR.

Those bugs are described in issues #2711 and #2710

philberty (changes requested):
In theory it should be possibe to add a hello world test here by adding a parameter name to the printf varadic function lets make sure that works and go for it

gcc/rust/expand/rust-expand-visitor.cc
```diff
@@ -1264,12 +1264,13 @@ ExpandVisitor::visit (AST::ExternalFunctionItem &item)
   for (auto &param : item.get_generic_params ())
     visit (param);
 
-  // FIXME: Should this work? What is the difference between NamedFunctionParam
-  // and FunctionParam?
-  // expand_function_params (item.get_function_params ());
+  auto end = item.is_variadic () ? item.get_function_params ().end () - 1
```
philberty, line 1267: This seems sus... are you trying to get the last argument?

gcc/rust/expand/rust-cfg-strip.cc
```diff
@@ -2588,7 +2588,8 @@ CfgStrip::visit (AST::ExternalFunctionItem &item)
   /* strip function parameters if required - this is specifically
    * allowed by spec */
   auto &params = item.get_function_params ();
-  for (auto it = params.begin (); it != params.end ();)
+  auto end = item.is_variadic () ? params.end () - 1 : params.end ();
```
philberty, line 2591: same here

gcc/rust/ast/rust-item.h
```diff
@@ -570,7 +586,10 @@ class FunctionParam
   FunctionParam &operator= (FunctionParam &&other) = default;
 
   // Returns whether FunctionParam is in an invalid state.
-  bool is_error () const { return param_name == nullptr || type == nullptr; }
+  bool is_error () const
+  {
+    return (param_name == nullptr || type == nullptr) && !variadic;
```
philberty, line 591: this is hard to read to me... I would have a if (varadic) { return error_condition } else param_name == nullptr || type ==nullptr

gcc/rust/expand/rust-expand-visitor.cc
P-E-P: I'm trying to avoid the last argument in the for loop if the function is variadic because this argument has no type. I recognize it is not as easy to read as I would like but I haven't found a better way to express it.

P-E-P: Now that I've worked on #2713 I know that AST validation pass is done after the expansion dump. This means there may be some invalid variadic function as shown in issues #2710 and #2711 going in that loop whose type cannot be fetched. Instead of avoiding the last parameter when the function is variadic, we should check for the presence of a type in the loop.

CohenArthur (approved):
thank you! nice work :)

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -22,7 +22,9 @@
 /* DO NOT INCLUDE ANYWHERE - this is automatically included with rust-parse.h
  * This is also the reason why there are no include guards. */
 
+#include "rust-ast-full-decls.h"
 #include "rust-token.h"
+#include <vector>
```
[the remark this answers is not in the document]
P-E-P: Yeah, I didn't catch this while reviewing my diff sorry. My editor has this new behavior that I need to disable.

gcc/rust/expand/rust-expand-visitor.cc
P-E-P: Rebased and fixed!

gcc/rust/expand/rust-cfg-strip.cc
P-E-P: Fixed!

gcc/rust/ast/rust-item.h
P-E-P: Fixed!

P-E-P:
> In theory it should be possibe to add a hello world test here by adding a parameter name to the printf varadic function lets make sure that works and go for it

I did as you suggested, with two new tests in [171e3ae](https://github.com/Rust-GCC/gccrs/pull/2701/commits/171e3ae7071792d3907217a0ad26d02a594c67de), one execute with a printf as well as another test to ensure named variadic with a wildcard pattern are parsed correctly.

philberty (approved):
LGTM go for it

PR #2697, opened 2023-10-17
HIR cleanup and refactor

jdupak:
Minor changed done during the borrowck work. Tested after each  commit (but before the last rebase).

PR #2702, opened 2023-10-19
BIR

jdupak:
Depends on umberged PR's (hence the noise). #2689 #2697

Initial structure for borrowchecking IR.

### DIFF
--- a/gcc/rust/checks/errors/borrowck/bir-design-notes.md
+++ b/gcc/rust/checks/errors/borrowck/bir-design-notes.md
@@ -1,0 +1,188 @@
+# Borrow-checker IR (BIR) design notes
+
+## Design goals
+
+Rust GCC project aims to use the [Polonius project](https://github.com/rust-lang/polonius) as its borrow-checker.
+Polonius operates on a set of [facts](https://github.com/rust-lang/polonius/blob/master/polonius-engine/src/facts.rs) about the program to determine
+the actual lifetimes of borrows.
+As Polonius's primary analysis is location sensitive, the facts are tied to the program's control flow graph (CFG).
+Unlike rustc, which has
+its
+own three address language specific representation called [MIR](https://rustc-dev-guide.rust-lang.org/mir/index.html), gccrs uses gcc's AST based
+representation GENERIC.
+GIMPLE (the three address IR) is generated from GENERIC inside GCC and can carry no language-specific information.
+Therefore, we
+need to generate our own representation of the program's CFG.
+Since gccrs has no intention of having a MIR-like IR, the BIR is not to be used for
+code generation.
+Therefore, BIR carries only minimal information that is necessary for the borrow-checker and for BIR debugging.
+
+Since BIR is in fact a dead branch of the compilation pipeline, the only way to verify its generations is through manual inspection.
+To have some frame of reference for testing, BIR build and dump are carefully designed to resemble the textual dump of rustc's MIR as much as
+possible.
+This includes the style of the output, numbering of locals and order of basic blocks (when possible).
+
+## BIR Dump Example
+
+An example program calculating the i-th fibonacci number:
+
+```rust
+
+fn fib(i: usize) -> i32 {
+    if i == 0 || i == 1 {
+        1
+    } else {
+        fib(i - 1) + fib(i - 2)
+    }
+}
+```
+
+Here is an example of BIR dump (note: this needs to be updated regularly):
+
+```
+fn fib(_1: usize) -> i32 {
+    let _0: i32;
+    let _2: i32;
+    let _3: bool;
+    let _4: bool;
+    let _5: bool;
+    let _6: usize;
+    let _7: i32;
+    let _8: usize;
+    let _9: i32;
+    let _10: i32;
+
+    bb0: {
+        _4 = Operator(_1, const usize);
+        switchInt(_4) -> [bb1, bb2];
+    }
+
+    bb1: {
+        _3 = const bool;
+        goto -> bb3;
+    }
+
+    bb2: {
+        _5 = Operator(_1, const usize);
+        _3 = _5;
+        goto -> bb3;
+    }
+
+    bb3: {
+        switchInt(_3) -> [bb4, bb7];
+    }
+
+    bb4: {
+        _2 = const i32;
+        goto -> bb8;
+    }
+
+    bb5: {
+        _6 = Operator(_1, const usize);
+        _7 = Call(fib)(_6, ) -> [bb6];
+    }
+
+    bb6: {
+        _8 = Operator(_1, const usize);
+        _9 = Call(fib)(_8, ) -> [bb7];
+    }
+
+    bb7: {
+        _10 = Operator(_7, _9);
+        _2 = _10;
+        goto -> bb8;
+    }
+
+    bb8: {
+        _0 = _2;
+        return;
+    }
+}
+
+
+```
+
+The dump consists of:
+
+- A function header with arguments: `fn fib(_1: usize) -> i32 { ... }`.
+- Declaration of locals: `let _0: i32;`, where `_0` is the return value (even if it is of the unit type). Arguments are not listed here, they are
+  listed in the function header.
+- A list of basic blocks: `bb0: { ... }`. The basic block name is the `bb` prefix followed by a number.
+- Each basic block consists of a list of BIR nodes (instructions). Instruction can be either assigned to a local (place) or be a statement.
+  Instructions take locals (places) as arguments.
+- Each basic block is terminated with a control flow instruction followed by a list of destinations:
+    - `goto -> bb3;` - a goto instruction with a single destination.
+    - `switchInt(_3) -> [bb4, bb7];` - a switch instruction with multiple destinations.
+    - `return;` - a return instruction with no destinations.
+    - `Call(fib)(_6, ) -> [bb6];` - a call instruction with a single destination. This section is prepared for panic handling.
+
+## BIR Structure
+
+BIR structure is defined in `gcc/rust/checks/errors/borrowck/rust-bir.h`. It is heavily inspired by rustc's MIR. The main difference is that BIR
+drastically reduces the amount of information carried to only borrow-checking relevant information.
+
+As borrow-checking is performed on each function independently, BIR represents a single function (`struct Function`). A `Function` consists of a list
+of basic blocks, list of arguments (for dump only) and place database, which keeps track of locals.
+
+### Basic Blocks
+
+A basic block is identified by its index in the function's basic block list. It contains a list of BIR nodes (instructions) and a list of successor
+basic block indices in CFG.
+
+### BIR Nodes (Instructions)
+
+BIR nodes are of three categories:
+
+- An assignment of an expression to a local (place).
+- A control flow operation (switch, return).
+- A special node (not executable) node, which carries additional information for borrow-checking (`StorageDead`, `StorageLive`).
+
+#### Expressions
+
+Expressions represent the executable parts of the rust code. Many different Rust contracts are represented by a single expression, as only data (and
+lifetime) flow needs to be tracked.
+
+- `InitializerExpr` represents any kind of struct initialization. It can be either explicit (struct expression) or implicit (range expression,
+  e.g. `0..=5`).
+- `Operator<ARITY>` represents any kind of operation, except the following, where special information is needed either for borrow-checking or for
+  better debugging.
+- `BorrowExpr` represents a borrow operation.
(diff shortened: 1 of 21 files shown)


jdupak:
@CohenArthur

jdupak:
Fixed all commits to compile with gcc48.

gcc/rust/checks/errors/borrowck/bir-design-notes.md
```diff
@@ -0,0 +1,188 @@
+# Borrow-checker IR (BIR) design notes
```
dkm, line 1: should we have all these doc files in a `doc/` subdir?

dkm (commented):
minor question

gcc/rust/checks/errors/borrowck/bir-design-notes.md
jdupak: I am fine with either variant. If `docs/` is more consistent, we can go with that.

gcc/rust/checks/errors/borrowck/rust-bir.h
```diff
@@ -0,0 +1,200 @@
...
+private:
+  Kind kind;
+  // ASSIGNMENT: lhs
+  // SWITCH: switch_val
+  // StorageDead/StorageLive: place
+  // otherwise: <unused>
+  PlaceId place;
+  // ASSIGNMENT: rhs
+  // otherwise: <unused>
+  std::unique_ptr<AbstractExpr> expr;
```
CohenArthur, line 68: so in Rust this would be something like this, correct?

```rust
enum Kind {
    Assignment { lhs: PlaceId, rhs: AbstractExpr },
    Switch { switch_val: PlaceId },
    StorageDead(PlaceId),
    StorageAlive(PlaceId),
    Goto,
    Return,
}
```

if that is the case, then I think a better API would be to have methods like `get_lhs()`, `get_rhs()` or `get_switch_value()` which would `rust_assert()` that we have the proper kind. does that make sense?

gcc/rust/checks/errors/borrowck/rust-bir.h
```diff
@@ -0,0 +1,200 @@
...
+
+  explicit Node (Kind kind, PlaceId place = INVALID_PLACE,
+		 AbstractExpr *expr = nullptr)
+    : kind (kind), place (place), expr (expr)
+  {}
+
+public:
+  [[nodiscard]] Kind get_kind () const { return kind; }
+  [[nodiscard]] PlaceId get_place () const { return place; }
+  [[nodiscard]] AbstractExpr &get_expr () const { return *expr; }
```
CohenArthur, line 83: we need to handle `[[nodiscard]]` nicely in our CI and in the GCC build system for this to work properly, otherwise the compiler will complain that it is a C++17 extension (despite having the proper behavior in C++11 mode). @tschwinge @dkm do you know if this is used in other frontends/how to deal with it?

gcc/rust/checks/errors/borrowck/rust-function-collector.h
```diff
@@ -0,0 +1,198 @@
...
+  {
+    functions.push_back (&function);
+    function.get_definition ()->accept_vis (*this);
+  }
+
+  void visit (HIR::ClosureExpr &closure) override
+  {
+    closures.push_back (&closure);
+    closure.get_expr ()->accept_vis (*this);
+  }
```
CohenArthur, line 68: the implementation for these functions should live in a `.cc` file. This can be done in a separate PR

gcc/rust/checks/errors/borrowck/rust-function-collector.h
```diff
@@ -0,0 +1,198 @@
...
+
+protected:
+  template <typename T> void visit_all (std::vector<std::unique_ptr<T>> &items)
+  {
+    for (std::unique_ptr<T> &item : items)
+      {
+	item->accept_vis (*this);
+      }
+  }
```
CohenArthur, line 56: IIUC you also need to call this on all the statements of a block, and all the items of a module right? this can be done in a later PR

CohenArthur (commented):
since the PR is very big, I will split my reviewing in a couple of passes as well. I've left some comments but I haven't finished reading through all of it - which I'll get to later in the day :) thank you, this is very impressive work! looking forward to understanding all of it!

gcc/po/gcc.pot
```diff
@@ -1288,7 +1288,7 @@ msgstr ""
 #, no-c-format
 msgid ""
 "-frust-compile-until=[ast|attributecheck|expansion|nameresolution|lowering|"
-"typecheck|privacy|unsafety|const|copimlation|end]             When to stop "
+"typecheck|privacy|unsafety|const|borrowcheck|copimlation|end]             When to stop "
```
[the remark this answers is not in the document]
jdupak: Hmm, it is generated by gettext but tracked by git. So it should be a part of the PR, but not manually changed. Do you know how?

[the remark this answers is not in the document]
jdupak: I did change lang.opt as well.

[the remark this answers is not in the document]
CohenArthur: ah nice. I hadn't got to that part of the PR yet. I think that file is generated by GCC on a nightly basis or something, so it shouldn't be part of the PR - in fact, it still contains the `comiplation` typo or whatever, which was fixed in lang.opt. It will be fixed and re-generated once we do a full merge upstream I assume

gcc/rust/checks/errors/borrowck/rust-bir.h
```diff
@@ -0,0 +1,200 @@
...
+    : operands (operands)
+  {}
+
+public:
+  template <size_t I>[[nodiscard]] PlaceId get_operand () const
+  {
+    static_assert (I < ARITY, "Index out of bounds");
+    return operands[I];
+  }
+};
```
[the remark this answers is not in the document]
jdupak: There is not use for it as long as we keep binary and unary operands separated in the BIR. Think of it as `get_lhs` and `get_rhs` for binary operands.

gcc/rust/checks/errors/borrowck/rust-borrow-checker.cc
```diff
@@ -0,0 +1,99 @@
...
+void
+mkdir_wrapped (const std::string &dirname)
+{
+  int ret;
+#ifdef _WIN32
+  ret = _mkdir (dirname.c_str ());
+#elif unix
+  ret = mkdir (dirname.c_str (), 0775);
+#elif __APPLE__
+  ret = mkdir (dirname.c_str (), 0775);
```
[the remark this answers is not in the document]
jdupak: Actually not really, since failing to open a file will handle that error. I thought about it but I don't care about failure for example if it already exists - which might get annoying to track for different platforms.

> Also, should that function live in rust-borrow-checker.cc and not in rust-bir-dump.cc?
From my original point of view, the caller (rust-borrow-checker.cc) handles the saving of the dump. But now thinking about it, it might be better there, cause it is not interesting for borrowchecking.

gcc/rust/checks/errors/borrowck/rust-function-collector.h
jdupak: Yes, as mentioned bellow, that is a TODO.

gcc/rust/checks/errors/borrowck/rust-function-collector.h
jdupak: The header file is only used in one place, so it seamed like useless separations. I only created a cc file if there was a reason. Most of my implementation is in header files. I like it that way more, but it is up to you.

gcc/po/gcc.pot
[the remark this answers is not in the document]
jdupak: From looking at the git log, I don't feel it is automatic. But I don't know.

gcc/rust/checks/errors/borrowck/rust-bir.h
jdupak: I take that citation back, that requirement is C++17.

gcc/rust/checks/errors/borrowck/rust-bir.h
jdupak: >so in Rust this would be something like this, correct?

Yes

> if that is the case, then I think a better API would be to have methods like get_lhs(), get_rhs() or get_switch_value() which would rust_assert() that we have the proper kind. does that make sense?

Sure, sounds good. I am not convinced it is actually useful here, but it might as the checks get more complicated.

gcc/rust/checks/errors/borrowck/rust-bir.h
```diff
@@ -0,0 +1,200 @@
...
+public:
+  enum Kind
+  {
+    ASSIGNMENT,	  // <place> = <expr>
+    SWITCH,	  // switch <place>
+    RETURN,	  // return
+    GOTO,	  // goto
+    STORAGE_DEAD, // StorageDead(<place>)
+    STORAGE_LIVE, // StorageLive(<place>)
+  };
```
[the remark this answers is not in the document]
jdupak: [fixed] enum class

gcc/po/gcc.pot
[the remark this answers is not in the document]
tschwinge: Regeneration of `gcc/po/gcc.pot` is semi-automatic: it's a (scripted) manual process.  It's correct that changes to `[...].pot` files are not to be included in general commits.  (They won't do any harm, though; these files will be regenerated before being sent to the translation teams.)

gcc/rust/checks/errors/borrowck/rust-bir.h
tschwinge: The standard way would be to define a version/feature-dependent macro in `gcc/system.h`.  For example, like the one that already exists in `libstdc++-v3/include/bits/c++config`:

```c++
// Macro to warn about unused results.
#if __cplusplus >= 201703L
# define _GLIBCXX_NODISCARD [[__nodiscard__]]
#else
# define _GLIBCXX_NODISCARD
#endif
```

However, can't you just use the existing `gcc/system.h:WARN_UNUSED_RESULT` (see <https://gcc.gnu.org/onlinedocs/gcc-4.8.5/gcc/Function-Attributes.html#index-g_t_0040code_007bwarn_005funused_005fresult_007d-attribute-2785>):

```c++
#if GCC_VERSION >= 3004
#define WARN_UNUSED_RESULT __attribute__ ((__warn_unused_result__))
#else
#define WARN_UNUSED_RESULT
#endif
```

Or does that have different semantics than `nodiscard`?

gcc/po/gcc.pot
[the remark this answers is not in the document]
jdupak: Removed from PR.

gcc/rust/checks/errors/borrowck/rust-function-collector.h
CohenArthur: putting it in headers slows down the compilation a lot sadly, since we constantly do a lot of includes. it does reflect on our compile time if the implementation lives in headers vs source files, so we've been trying to keep it that way. especially for "complex" functions, which could get modified later on like these. a simple getter or constructor can live in a header, that's no problem

gcc/rust/checks/errors/borrowck/rust-borrow-checker.cc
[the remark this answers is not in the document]
jdupak: It would be probably better to use here the mkdir from `gcc/system.h` (as Thomas suggested on Zulip) then the ifdefs which I took from some other part of gcc. One problem there is that `mkdir` symbol is deprecated on windows and `_mkdir` should be used. But that should probably be handled for the whole gcc in the system header once that becomes an issue.s Regarding the error handling, we need to ignore error when directory exists, but I found out that errno value macro EEXISTS is defined in C++11 (not just POSIX), so we should be safe there.

Regarding the location, I will need the mkdir also in other parts then bir dump (facts dump), so I think it should stay in `rust-borrow-checker.cc`

@CohenArthur ?

gcc/rust/checks/errors/borrowck/rust-bir.h
jdupak: Replaced for WARN_UNUSED_RESULT.

gcc/rust/checks/errors/borrowck/rust-borrow-checker.cc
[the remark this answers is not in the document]
CohenArthur: sounds like a good explanation :) I'm okay with all of it and will mark this as resolved

CohenArthur (approved):
amazing work, thank you! I'm excited to merge this :)

PR #2708, opened 2023-10-23
Parse const with no value

### LINKED ISSUE #2665: Cannot parse disabled const in impl
The following snippet is rejected by GCCRS even though the code is disabled and should be accepted.

I tried this code:

```rust
fn main() {}

#[cfg(FALSE)]
impl X {
    const Y: u8;
}
```

I expected to see this happen: Compiles correctly

Instead, this happened:

```
./test.rs:5:16: error: expecting ‘=’ but ‘;’ found
    5 |     const Y: u8;
      |                ^
./test.rs:6:1: error: failed to parse inherent impl item in inherent impl
    6 | }
      | ^
./test.rs:6:1: error: failed to parse item in crate
```

### Meta

P-E-P:
Fixes #2665.

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -4783,6 +4783,16 @@ Parser<ManagedTokenSource>::parse_const_item (AST::Visibility vis,
   // parse constant type (required)
   std::unique_ptr<AST::Type> type = parse_type ();
 
+  // A const with no given expression value
+  if (lexer.peek_token ()->get_id () == SEMICOLON)
+    {
+      lexer.skip_token ();
+      return std::unique_ptr<AST::ConstantItem> (
+	new AST::ConstantItem (std::move (ident), std::move (vis),
+			       std::move (type), std::move (outer_attrs),
+			       locus));
+    }
+
   if (!skip_token (EQUAL))
     {
       skip_after_semicolon ();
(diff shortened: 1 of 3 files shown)


gcc/rust/parse/rust-parse-impl.h
```diff
@@ -4783,6 +4783,16 @@ Parser<ManagedTokenSource>::parse_const_item (AST::Visibility vis,
...
 
+  // A const with no given expression value
+  if (lexer.peek_token ()->get_id () == SEMICOLON)
+    {
+      lexer.skip_token ();
+      return std::unique_ptr<AST::ConstantItem> (
+	new AST::ConstantItem (std::move (ident), std::move (vis),
+			       std::move (type), std::move (outer_attrs),
+			       locus));
+    }
```
CohenArthur, line 4794: I think this is good, but we need to handle it somewhere else, right? e.g. the rest of the AST and HIR are probably expecting that an `impl`'s `const` always has a value, right?

CohenArthur: yeah, this ICEs:

```rust
struct S;

impl S {
    const D: u8;
}
```

while it should produce a nice error: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9b8643a863d5350e8b226e8bc9789b88

this should be adressed in a later PR however :)

PR #2713, opened 2023-10-24
Reject const item with no value in impl context

### LINKED ISSUE #2709: ICE in impl const items without value
```rust
struct S;

impl S {
    const D: u8;
}
```

while it should produce a nice error: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9b8643a863d5350e8b226e8bc9789b88

this should be adressed in a later PR however :)

_Originally posted by @CohenArthur in https://github.com/Rust-GCC/gccrs/pull/2708#discussion_r1368934653_

P-E-P:
Add an AST validation visitor pass rejecting invalid rust code.

Fixes #2709


gcc/rust/checks/errors/rust-ast-validation.h
CohenArthur, line 1: I like this, we should definitely have more AST validation passes. however, I think it might be time to *finally* port over one of the default visitors, because fuck reimplementing the same empty visitors over and over and over and *OVER AGAIN*. I think a lot of the functions here are not going to do anything besides visiting their children nodes, so we should definitely do this. what do you think?

gcc/rust/checks/errors/rust-ast-validation.h
```diff
@@ -0,0 +1,208 @@
...
+    INHERENT_IMPL,
+    TRAIT_IMPL,
+    CRATE,
+  };
+
+  std::vector<Context> context;
+
+  void push_context (Context ctx) { context.push_back (ctx); }
+
+  void pop_context () { context.pop_back (); }
```
CohenArthur, line 203: I think we have a similar class in many other visitors, might be worth factoring it into a common thing. I think the macro expansion code definitely uses something like that

CohenArthur (approved):
Looks good to me! Even if this is a big PR and a big change I think this is the way to go

gcc/rust/checks/errors/rust-ast-validation.h
P-E-P: This was what I had in mind too. Please check the new commits.

gcc/rust/checks/errors/rust-ast-validation.h
P-E-P: I've made a whole new class for this, I'm still unsure it fits the macro visitors since the latter miss some information. In either case I'm not modifying the macro visitors in this PR, if it fits then I'll refactor it in another one.

PR #2715, opened 2023-10-29
Replace ```AST::Method``` with existing ```AST::Function```

powerboat9:


P-E-P:
With these changes it will be possible to add the right error message in the ast validation pass.

```
`self` parameter is only allowed in associated functions
```

powerboat9:
@philberty

P-E-P (approved):
I was very reluctant about these changes and was waiting for other people's opinions, but when I went to see rustc I realised that they had something similar, altough the `self` type is not optional but rather a parameter variant with it's own type (`ImplicitSelf`). Our AST is massive and there is many duplicated structures, I'm in favor of merging this.

P-E-P:
`rustc` consider the self parameter to be a plain parameter, a detail was still bothering me with this: Does rustc reject a wrong self parameter during the AST validation pass ? If so, we couldn't take this approach because our optional does not keep the `self` parameter's position. Hopefully [it does not](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7cad8867720ba99f1b2bded1ec941103), this code is rejected during parsing.

powerboat9:
Judging by [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7c69b0e65170f45d3ccdf2d60f4b4b99) the position of a self parameter is checked during parsing, but whether a method/function should have a self parameter is checked afterwards.

philberty (approved):
LGTM lets go for it.. I did this in the HIR because that's the dungeon where I live but was worried how this might impact higher level stuff like macro expansion so given that we already have loads of tests on the parsing and macros I think we should just go for it since it reduces a lot of duplication!

powerboat9:
@philberty conflicts resolved

Issue #2650, opened 2023-10-06
Fail to parse async function


P-E-P:
GCCRS fails to parse an async function

I tried this code:

```rust
pub async fn a() -> u32 {
    1
}
```

I expected to see this happen: Compile correctly

Instead, this happened: 
```
./test.rs:1:5: error: failed to parse item in crate
    1 | pub async fn a() -> u32 {
      |     ^~~~~
```

### Meta

[007166c](https://github.com/Rust-GCC/gccrs/commit/007166ca6ceecd429f7033336af25dd679424ca2)

mvvsmk:
Hey , @P-E-P could you assign this to me, I'll start working on it. :)

officialhemant511:
i think GCCRS is a tool that aims to provide compatibility with the C language and may not support all Rust language features, such as async/await. Therefore, if you're trying to use async/await in your code, GCCRS may not be the appropriate tool for that specific task.i think we work with c code here...

P-E-P:
> i think GCCRS is a tool that aims to provide compatibility with the C language and may not support all Rust language features, such as async/await. Therefore, if you're trying to use async/await in your code, GCCRS may not be the appropriate tool for that specific task.i think we work with c code here...

We're trying to support rust version 1.49 for now. Async/await syntax were introduced in version 1.38 of the language, this means we should support those. Even if asynchronous code may not be the highest priority, this issue reduce the gap between rust and our parser. This is mandatory to be able to parse all rustc's test suite, we're only a few percentage away.

P-E-P:
@mvvsmk have you made any progress so far ? I think we'll need this earlier than expected as I'm working on AST validation. If you feel you won't be able to work on this soon I could reassign the issue, don't put any pressure on yourself or anything.

Also if you need any help to solve this, tell me.

mvvsmk:
Hey @P-E-P ,  I made the changes two weeks ago and forgot to make a PR. Apologies for the delay; it completely slipped my mind, been a bit busy last month. I have made the PR now. Could you code review my PR, if I missed something do let me know.

PR #2560, opened 2023-08-15
[E0034] found more than one items for method

MahadMuhammad:
### Multiple Applicable items found for method `foo` - [`E0034`](https://doc.rust-lang.org/error_codes/E0034.html) 

Added rich location and error code for [`E0034`](https://doc.rust-lang.org/error_codes/E0034.html) 

Also, Fixes https://github.com/Rust-GCC/gccrs/issues/2366

### Code tested:
- [`gcc/testsuite/rust/compile/generics6.rs`](https://github.com/Rust-GCC/gccrs/blob/master/gcc/testsuite/rust/compile/generics6.rs)

```rust
/home/mahad/Desktop/mahad/gccrs/gcc/testsuite/rust/compile/generics6.rs:29:23: error: multiple applicable items in scope for: ‘test’ [E0034]
    9 |     fn test() -> i32 { // { dg-error "possible candidate" "TODO" { xfail *-*-* } }
      |     ~~                 
......
   19 |     fn test() -> i32 { // { dg-error "possible candidate" "TODO" { xfail *-*-* } }
      |     ~~                 
......
   29 |     let a: i32 = Foo::test(); // { dg-error "multiple applicable items in scope for: test" }
      |                       ^~~~
      |                       multiple test found
/home/mahad/Desktop/mahad/gccrs/gcc/testsuite/rust/compile/generics6.rs:29:18: error: Failed to resolve expression of function call
   29 |     let a: i32 = Foo::test(); // { dg-error "multiple applicable items in scope for: test" }
      |                  ^~~


```

---

**gcc/rust/ChangeLog:**

	* typecheck/rust-hir-type-check-expr.cc (TypeCheckExpr::visit): Added rich location and error code.

**gcc/testsuite/ChangeLog:**

	* rust/compile/generics7.rs: Updated dejagnu comment.
	* rust/compile/issue-925.rs: likewise.

PR #2736, opened 2023-11-10
Added support to Parse ASYNC function

mvvsmk:
Fixes issue #2650 
The parser now parses ASYNC functions.
gcc/rust/ChangeLog:

	* parse/rust-parse-impl.h (Parser::parse_item):Added ASYNC case.
	(Parser::parse_vis_item):Added a switch case to handle ASYNC.

gcc/testsuite/ChangeLog:

	* rust/compile/issue-2650.rs: New test.

Signed-off-by : M V V S Manoj Kumar <mvvsmanojkumar@gmail.com>

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -30,6 +30,7 @@
 #include "rust-dir-owner.h"
 #include "rust-attribute-values.h"
 #include "rust-keyword-values.h"
+#include "rust-session-manager.h"
 
 #include "optional.h"
 
@@ -1113,6 +1114,8 @@ Parser<ManagedTokenSource>::parse_item (bool called_from_statement)
 	  add_error (std::move (error));
 	}
       return nullptr;
+
+    case ASYNC:
     case PUB:
     case MOD:
     case EXTERN_KW:
@@ -1389,6 +1392,10 @@ Parser<ManagedTokenSource>::parse_vis_item (AST::AttrVec outer_attrs)
 	  lexer.skip_token (1); // TODO: is this right thing to do?
 	  return nullptr;
 	}
+    // for async functions
+    case ASYNC:
+      return parse_async_item (std::move (vis), std::move (outer_attrs));
+
     case STATIC_KW:
       return parse_static_item (std::move (vis), std::move (outer_attrs));
     case AUTO:
@@ -1427,6 +1434,39 @@ Parser<ManagedTokenSource>::parse_vis_item (AST::AttrVec outer_attrs)
       break;
     }
   return nullptr;
+}
+
+template <typename ManagedTokenSource>
+std::unique_ptr<AST::Function>
+Parser<ManagedTokenSource>::parse_async_item (AST::Visibility vis,
+					      AST::AttrVec outer_attrs)
+{
+  const_TokenPtr t = lexer.peek_token ();
+  if (Session::get_instance ().options.get_edition ()
+      == CompileOptions::Edition::E2015)
+    {
+      add_error (Error (t->get_locus (), ErrorCode::E0670,
+			"%<async fn%> is not permitted in Rust 2015"));
+      add_error (
+	Error::Hint (t->get_locus (),
+		     "to use %<async fn%>, switch to Rust 2018 or later"));
+    }
+
+  t = lexer.peek_token (1);
+
+  switch (t->get_id ())
+    {
+    case UNSAFE:
+    case FN_KW:
+      return parse_function (std::move (vis), std::move (outer_attrs));
+
+    default:
+      add_error (
+	Error (t->get_locus (), "expected item, found keyword %<async%>"));
+
+      lexer.skip_token (1);
+      return nullptr;
+    }
 }
 
 // Parses a macro rules definition syntax extension whatever thing.
--- a/gcc/testsuite/rust/compile/issue-2650-2.rs
+++ b/gcc/testsuite/rust/compile/issue-2650-2.rs
@@ -1,0 +1,5 @@
+// { dg-additional-options "-frust-edition=2015" }
+
+pub async fn a() -> u32 { // { dg-error "'async fn' is not permitted in Rust 2015" }
+    1
+}
--- a/gcc/rust/parse/rust-parse.h
+++ b/gcc/rust/parse/rust-parse.h
@@ -357,6 +357,8 @@ private:
   std::unique_ptr<AST::ExternBlock>
   parse_extern_block (AST::Visibility vis, AST::AttrVec outer_attrs);
   std::unique_ptr<AST::Function> parse_method ();
+  std::unique_ptr<AST::Function> parse_async_item (AST::Visibility vis,
+						   AST::AttrVec outer_attrs);
 
   // Expression-related (Pratt parsed)
   std::unique_ptr<AST::Expr>
--- a/gcc/testsuite/rust/compile/issue-2650-1.rs
+++ b/gcc/testsuite/rust/compile/issue-2650-1.rs
@@ -1,0 +1,5 @@
+// { dg-additional-options "-frust-edition=2018" }
+
+pub async fn a() -> u32 {
+    1
+}


dkm:
Typo in commit log (handle).
Also, you probably want to refer to #2650 instead of #2560

mvvsmk:
Thanks for pointing out the typo and the link error. I have reworded my commit to reflect it.

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -1385,6 +1387,25 @@ Parser<ManagedTokenSource>::parse_vis_item (AST::AttrVec outer_attrs)
...
+	default:
+	  add_error (
+	    Error (t->get_locus (),
+		   "unexpected token %qs in some sort of async production",
+		   t->get_token_description ()));
+
+	  lexer.skip_token (1);
+	  return nullptr;
+	}
+
```
CohenArthur, line 1408: async is not allowed if the Rust edition is set to 2015, which is the default. I think keeping the parser like that is fine, but we should add that to the checks in the AST validation pass

gcc/testsuite/rust/compile/issue-2650.rs
```diff
@@ -0,0 +1,3 @@
+pub async fn a() -> u32 {
+    1
+}
```
CohenArthur, line 3: so this test should eventually fail, unless we are in an edition >= 2018. so please do the following:
```suggestion
// { dg-additional-options "-frust-edition=2018" }

pub async fn a() -> u32 {
    1
}
```

CohenArthur (approved):

CohenArthur:
example of the error if the Rust edition is set to 2015: https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=9696f6f21bd504217a3b9dab4947f507

gcc/rust/parse/rust-parse-impl.h
P-E-P: Rustc error out on `async` during parsing. We probably want to do the same to ensure macro compatibility.

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -1385,6 +1387,25 @@ Parser<ManagedTokenSource>::parse_vis_item (AST::AttrVec outer_attrs)
 	  lexer.skip_token (1); // TODO: is this right thing to do?
 	  return nullptr;
 	}
+    case ASYNC:
```
P-E-P, line 1390: I feel like we could have a new function here, something along `parse_async_item`, we already have `parse_const_item`.

P-E-P (approved):
I'll be happy to merge this once the edition thing will be sorted out.

gcc/rust/parse/rust-parse-impl.h
mvvsmk: Could you tell me where or how I can find which version of Rust is set ? I know the magic happens in the lang.opt file, but how do we know what parameters are passed?

gcc/testsuite/rust/compile/issue-2650.rs
mvvsmk: I see, I'll do that. I am not familiar with dejagnu , but wouldn't it be "-frust-edition>=2018" or does it handle it on its own?

gcc/rust/parse/rust-parse-impl.h
mvvsmk: Understood, I'll make this change.

gcc/testsuite/rust/compile/issue-2650.rs
P-E-P: This dejagnu directive pass the argument to the compiler, this means the test code will be compiled using gccrs with the flag `-frust-edition=2018` hence why we're not setting `>=`, because we're setting the exact rust edition, we're not checking against all editions greater than 2018.

mvvsmk: Oh , I got it now, thanks for the help :D

gcc/rust/parse/rust-parse-impl.h
P-E-P: AFAIK there is a singleton in the rust session manager.

gcc/rust/parse/rust-parse-impl.h
mvvsmk: Hey @P-E-P , in our parser we have a function called `parse_async_block`, which has just has a defination and no caller in the entire codebase, would you like me to make a new `parse_async_item` function or just ammend my code to `parse_async_block`

P-E-P: I think `parse_async_block` was made exclusively for async block (`async {}`) you should probably create a new function instead of using this one.

mvvsmk: Oh ok, got it :)

mvvsmk:
I have made the requested changes, if I missed anything do let me know :). 

I also added another test that fails, so we have both cases covered. I also noticed that our `parse_async_block` has no such check, would you like me to add it in this PR itself or make a separate one?

also a minor change but in rustc error says :
```
`async fn` is not permitted in Rust 2015
```
but the cpp11 compiler just kept giving me warnings about the " ` " symbol don't know what that was about, spent half the day figuring it out but it would still give me a warning no matter what I did so, changed it to :
```
'async fn' is not permitted in Rust 2015
```
if you guys got any hints about it do let me know :)

P-E-P:
> I have made the requested changes, if I missed anything do let me know :).
> 
> I also added another test that fails, so we have both cases covered. I also noticed that our `parse_async_block` has no such check, would you like me to add it in this PR itself or make a separate one?
> 
> also a minor change but in rustc error says :
> 
> ```
> `async fn` is not permitted in Rust 2015
> ```
> 
> but the cpp11 compiler just kept giving me warnings about the " ` " symbol don't know what that was about, spent half the day figuring it out but it would still give me a warning no matter what I did so, changed it to :
> 
> ```
> 'async fn' is not permitted in Rust 2015
> ```
> 
> if you guys got any hints about it do let me know :)

That's because of quoted string. If the quoted thing lives in a variable you might use `%qX` with X being the correct format argument. For `*char` that would be `%qs`. Otherwise you can use `%<` and `%>`.

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -1427,6 +1434,40 @@ Parser<ManagedTokenSource>::parse_vis_item (AST::AttrVec outer_attrs)
...
+std::unique_ptr<AST::Function>
+Parser<ManagedTokenSource>::parse_async_item (AST::Visibility vis,
+					      AST::AttrVec outer_attrs)
+{
+  const_TokenPtr t = lexer.peek_token ();
+  if (Session::get_instance ().options.get_edition ()
+      == CompileOptions::Edition::E2015)
+    {
+      add_error (Error (Error::Kind::Err, t->get_locus (), ErrorCode::E0670,
+			" %<`async fn`%> is not permitted in Rust 2015"));
```
P-E-P, line 1447: `%<` already puts a ` in your output text, you do not need to keep those `

gcc/testsuite/rust/compile/issue-2650-2.rs
```diff
@@ -0,0 +1,6 @@
+// { dg-additional-options "-frust-edition=2015" }
+
+// { dg-excess-errors "'async fn' is not permitted in Rust 2015" }
```
P-E-P, line 3: You won't be able to match the backtick token using dejaGNU because AFAIK it is based on the locale. User a dot `.` to match any character instead.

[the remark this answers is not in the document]
mvvsmk: Sure, I'll get it done.

gcc/rust/parse/rust-parse-impl.h
mvvsmk: Done :D

CohenArthur (approved):
LGTM :) nice work!

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -1427,6 +1434,40 @@ Parser<ManagedTokenSource>::parse_vis_item (AST::AttrVec outer_attrs)
...
+
+  switch (t->get_id ())
+    {
+    case UNSAFE:
+    case FN_TOK:
+      return parse_function (std::move (vis), std::move (outer_attrs));
+
+    default:
+      add_error (Error (t->get_locus (),
+			"unexpected token %qs in some sort of async production",
```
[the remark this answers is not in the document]
CohenArthur: once this is addressed we should merge the PR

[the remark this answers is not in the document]
mvvsmk: I have made this change but I was experimeting a bit with rustc and found this error code 
```
error: expected one of `extern`, `fn`, or `unsafe`, found keyword `const`
 --> src/lib.rs:1:7
  |
1 | async const fn foo() {
  | ------^^^^^
  | |     |
  | |     expected one of `extern`, `fn`, or `unsafe`
  | help: `const` must come before `async`: `const async`
  ```
  would you like me to go with that and make the relavent changes by adding extern to the next token list and change the error message to the one emmited by rustc?
  
  this is the code I was testing : https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=38fc87fda8f9debc717354d3af2ef863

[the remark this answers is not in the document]
P-E-P: @mvvsmk I think it would fit better in another PR

mvvsmk:
@P-E-P Could you add this back to the merge queue when you get time, there was a merge conflict so it was removed from the queue.

PR #2739, opened 2023-11-15
Add ForeverStack APIs for Name Resolution 2.0

CohenArthur:
- ast: Change *Path nodes API
- rib: Add Namespace enum
- forever-stack: Fix basic get logic
- foreverstack: Specialize `get` for Namespace::Labels
- forever stack: Fix resolve_path signature
- forever stack: Improve resolve_path implementation
- foreverstack: Add `to_canonical_path` method
- foreverstack: Add `to_rib` method

This PR introduces the necessary changes from #2620 in order to make late name resolution possible. Cherry-picked, cleaned up and all changelog'ed up.

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -399,7 +399,7 @@ public:
   {
     return as_string ().compare ("crate") == 0;
   }
-  bool is_lower_self () const { return as_string ().compare ("self") == 0; }
+  bool is_lower_self_seg () const { return as_string ().compare ("self") == 0; }
   bool is_big_self () const { return as_string ().compare ("Self") == 0; }
 };
 
--- a/gcc/rust/resolve/rust-ast-resolve-path.cc
+++ b/gcc/rust/resolve/rust-ast-resolve-path.cc
@@ -367,7 +367,7 @@ ResolvePath::resolve_path (AST::SimplePath *expr)
       //   	      is_first_segment ? "true" : "false",
       //   	      is_final_segment ? "true" : "false");
       if (resolved_node_id == UNKNOWN_NODEID && !is_first_segment
-	  && is_final_segment && segment.is_lower_self ())
+	  && is_final_segment && segment.is_lower_self_seg ())
 	{
 	  resolved_node_id = previous_resolved_node_id;
 	}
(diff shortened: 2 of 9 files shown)


gcc/rust/ast/rust-ast.h
```diff
@@ -399,7 +399,7 @@ class SimplePathSegment : public PathSegment
   {
     return as_string ().compare ("crate") == 0;
   }
-  bool is_lower_self () const { return as_string ().compare ("self") == 0; }
+  bool is_lower_self_seg () const { return as_string ().compare ("self") == 0; }
   bool is_big_self () const { return as_string ().compare ("Self") == 0; }
```
P-E-P, line 403: Not entirely sure why `is_lower_self` got renamed, shouldn't `is_big_self` be renamed too ?

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -406,20 +429,130 @@ ForeverStack<N>::resolve_segments (
...
+      return KeepGoing::Yes;
     });
+
+    auto path = Resolver::CanonicalPath::create_empty ();
+    for (const auto &segment : segments)
+      path = path.append (segment);
+
+    // Finally, append the name
+    path = path.append (Resolver::CanonicalPath::new_seg (id, name));
+    rust_debug ("[ARTHUR] found path: %s. Size: %lu", path.get ().c_str (),
```
P-E-P, line 526: Should we keep this debug information ?

P-E-P (approved):
Some minor comments, otherwise LGTM

gcc/rust/resolve/rust-ast-resolve-path.cc
```diff
@@ -367,7 +367,7 @@ ResolvePath::resolve_path (AST::SimplePath *expr)
       //   	      is_first_segment ? "true" : "false",
       //   	      is_final_segment ? "true" : "false");
       if (resolved_node_id == UNKNOWN_NODEID && !is_first_segment
-	  && is_final_segment && segment.is_lower_self ())
+	  && is_final_segment && segment.is_lower_self_seg ())
 	{
 	  resolved_node_id = previous_resolved_node_id;
 	}
```
P-E-P, line 373:
```suggestion
	  && is_final_segment && segment.is_lower_self_seg ())
	  resolved_node_id = previous_resolved_node_id;
```

gcc/rust/ast/rust-ast.h
CohenArthur: this is for using them in a templated function, so that the API is consistent with other Path-like nodes. I think changing `is_big_self` also makes sense

gcc/rust/resolve/rust-forever-stack.hxx
CohenArthur: nope, this can go away you're right :)

PR #2741, opened 2023-11-15
Prevent const self ptr

P-E-P:
Fix self pointer rejection during parsing.


CohenArthur (commented):
quick comment since this is still a draft, but should this be a parsing error or an AST validation error?

P-E-P:
> quick comment since this is still a draft, but should this be a parsing error or an AST validation error?

It was a parser error in [rustc v1.49](https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_parse/src/parser/item.rs#L1785)

CohenArthur (approved):
alright, looks great :) nice work

PR #2740, opened 2023-11-15
Add base for late name resolution 2.0

CohenArthur:
- ctx: Add Labels ForeverStack to the resolver.
- nr2.0: Add base for late name resolution
- toplevel: Use DefaultResolver for Function
- nr2.0: Store mappings in NameResolutionContext
- late: Start setting up builtin types
- late: Start storing mappings properly in the resolver

Needs #2739

PR #2742, opened 2023-11-15
Add `use` declaration resolution

CohenArthur:
- early: Resolve paths properly
- toplevel: Add comment about running the collector twice
- ast: Add NodeId to UseTree base class
- early: Move `use` declaration resolving to TopLevel
- toplevel: Resolve `use` declarations

Needs #2740
Needs #2739

### DIFF
--- a/gcc/rust/resolve/rust-early-name-resolver-2.0.h
+++ b/gcc/rust/resolve/rust-early-name-resolver-2.0.h
@@ -50,10 +50,7 @@ public:
   void visit (AST::Module &) override;
 
   void visit (AST::MacroInvocation &) override;
-  void visit (AST::UseDeclaration &) override;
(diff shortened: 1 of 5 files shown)


gcc/rust/resolve/rust-early-name-resolver-2.0.h
```diff
@@ -50,16 +50,23 @@ class Early : public DefaultResolver
...
 
 private:
   void visit_attributes (std::vector<AST::Attribute> &attrs);
 
+  /**
+   * Insert a resolved macro invocation into the mappings once, meaning that we
+   * can call this function each time the early name resolution pass is underway
+   * and it will not trigger assertions for already resolved invocations.
+   */
+  // TODO: Rename
```
P-E-P, line 65: Not sure what you want to rename here ?

PR #2743, opened 2023-11-15
Keyword list rework

### LINKED ISSUE #2735: Add missing reserved keyword
Multiple keywords are missing

- [x] ~~`$crate` (update the parser, currently treated as two tokens)~~ rustc threat this as two tokens
- [x] `_` (underscore)
- [x] `{{root}}`
- [x]  other missing keyword ? ...

- Change token key prefix from `TOK` to `KW` ?

P-E-P:
Some keyword are missing, the list lack edition separation and some suffix do not fit every context very well.

- [x]  Add missing keyword
- [x] Change keyword suffixes
- [x] Add some edition  mechanism
- [x] Replace raw keyword call with they constexpr counterpart

Fixes #2735


gcc/rust/lex/rust-token.h
```diff
@@ -147,69 +145,71 @@ enum PrimitiveCoreType
...
+  RS_TOKEN_KEYWORD_2015 (YIELD, "yield") /* unused */                          \
   RS_TOKEN (LAST_TOKEN, "<last-token-marker>")
 
 // Contains all token types. Crappy implementation via x-macros.
 enum TokenId
 {
 #define RS_TOKEN(name, _) name,
-#define RS_TOKEN_KEYWORD(x, y) RS_TOKEN (x, y)
+#define RS_TOKEN_KEYWORD_2015(x, y) RS_TOKEN (x, y)
+#define RS_TOKEN_KEYWORD_2018 RS_TOKEN_KEYWORD_2015
```
CohenArthur, line 209: :D

CohenArthur (approved):
that looks like an improvement, thanks :) I guess later down the line we can think about adding more specific behavior to 2018 vs 2015 keywords, and other editions

Issue #2725, opened 2023-11-07
`?Trait` should not be used in supertraits


tamaroning:
I tried this code:

```rust
// comment out below when compiling with rustc
#[lang = "sized"]
pub trait Sized {}

trait Trait: ?Sized {}

```

I expected to see this happen: 
```bash
   Compiling playground v0.0.1 (/playground)
error: `?Trait` is not permitted in supertraits
 --> src/lib.rs:3:14
  |
3 | trait Trait: ?Sized {}
  |              ^^^^^^
  |
  = note: traits are `?Sized` by default

error: could not compile `playground` (lib) due to previous error
```

Instead, this happened: compiled.

### Meta
godbolt

PR #2745, opened 2023-11-15
Fix optional trait parsing

dme2:
Addresses #2725

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-item.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-item.cc
@@ -518,6 +518,24 @@ TypeCheckItem::visit (HIR::Module &module)
 void
 TypeCheckItem::visit (HIR::Trait &trait)
 {
+  if (trait.has_type_param_bounds ())
+    {
+      for (auto &tp_bound : trait.get_type_param_bounds ())
+        {
+          if (tp_bound.get ()->get_bound_type ()
+              == HIR::TypeParamBound::BoundType::TRAITBOUND)
+            {
+              HIR::TraitBound &tb
+                  = static_cast<HIR::TraitBound &> (*tp_bound.get ());
+              if (tb.get_polarity () == BoundPolarity::AntiBound)
+                {
+                  rust_error_at (tb.get_locus (),
+                                 "%<?Trait%> is not permitted in supertraits");
+                }
+            }
+        }
+    }
+
   TraitReference *trait_ref = TraitResolver::Resolve (trait);
   if (trait_ref->is_error ())
     {
--- a/gcc/testsuite/rust/compile/issue-2725.rs
+++ b/gcc/testsuite/rust/compile/issue-2725.rs
@@ -1,0 +1,3 @@
+#[lang = "sized"]
+pub trait Sized {}
+trait Trait: ?Sized {} // { dg-error ".?Trait. is not permitted in supertraits" }


powerboat9:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0363d352bef03bd91009d58c499ebf95

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -4939,6 +4939,13 @@ Parser<ManagedTokenSource>::parse_trait (AST::Visibility vis,
     {
       lexer.skip_token ();
 
+      if (lexer.peek_token()->get_id() == QUESTION_MARK) {
+            lexer.skip_token();
+            rust_error_at(lexer.peek_token()->get_locus(),
+                          "%<?Trait%> is not permitted in supertraits");
+            return nullptr;
+      }
```
CohenArthur, line 4947: yes, as @powerboat9 showed we should not be emitting this error at the parsing stage, but probably in the typechecker. you can see that it parses properly in the playground link, or with a nightly `rustc` with the `-Z parse-only` CLI option

CohenArthur: if you'd like any guidance for adding that check to the type checker please do say so :)

dme2: Thanks for the feedback. I'll go ahead and try to add the check to the typechecker.

dme2:
Hi @CohenArthur, I know the formatting isn't passing yet, but does that latest commit look ok?

gcc/rust/typecheck/rust-hir-type-check-item.cc
```diff
@@ -518,6 +518,20 @@ TypeCheckItem::visit (HIR::Module &module)
 void
 TypeCheckItem::visit (HIR::Trait &trait)
 {
+  if (trait.has_type_param_bounds ())
+    {
+      for (auto &tp_bound : trait.get_type_param_bounds ())
+        {
+          HIR::TraitBound &tb
+              = static_cast<HIR::TraitBound &> (*tp_bound.get ());
+          if (tb.get_polarity () == BoundPolarity::AntiBound)
```
CohenArthur, line 527: this is good, but the static cast might fail if the trait bound is a lifetime.

you can switch on `tp_bound.get_bound_type()` and check if it is a lifetime or a trait, and proceed to cast safely then. so something like:
```suggestion
          if (tp_bound.get_bound_type() == LIFETIME) {
          HIR::TraitBound &tb
              = static_cast<HIR::TraitBound &> (*tp_bound.get ());
          if (tb.get_polarity () == BoundPolarity::AntiBound)
          /* snip */
          }
```

etc. does that make sense?

dme2: @CohenArthur, thank you for the review. Do  you have an example of the when the static cast might fail?

This program seems to compile fine with the code as is:

```rust
#[lang = "sized"]
pub trait Sized {}
trait Trait<'a>: 'a {}
```

CohenArthur: ah, my bad - the static cast won't fail, but the value will be undefined. so we still need to do the check.

P-E-P:
@dme2 What is the current state of this PR ? From what I've read it seems like there is only a minor format issue ?

P-E-P:
Closed in favor of #3055 (This is exactly the same PR with proper code format). Thank you for the initial work!

PR #2749, opened 2023-11-20
Auto trait AST validation

P-E-P:
Some constructs are forbidden in rust with auto traits such as generic arguments and super traits. Rust code containing such construct should be rejected and the compiler should emit an error.


CohenArthur (approved):
this is great. note that auto traits might get removed from Rust entirely since they're only available in very specific cases, and I believe there is ongoing discussion to only allow them for compiler builtins such as `Send`, `Sync` and `Sized`

PR #2751, opened 2023-11-21
Split async and const function qualifiers

P-E-P:
Function cannot be both async and const but this needs to be rejected at a later stage than parsing, during the ast validation pass.

Probably requires #2736


gcc/rust/util/rust-common.h
```diff
@@ -25,18 +25,30 @@
...
+{
+  Yes,
+  No,
+};
+
+enum class Async
+{
+  Yes,
+  No
+};
```
CohenArthur, line 50: Great stuff, these enums are much better than booleans for use cases like these

PR #2752, opened 2023-11-21
Unsafe module

P-E-P:
GCCRS does not allow unsafe module with it's parser even though the code shall be rejected at a later stage.


philberty (approved):
LGTM i like the new errors pass

PR #2759, opened 2023-11-24
Change feature gate visitor to use default visitor

P-E-P:
The feature gating behavior may be shortened and kept cleaner using the default visitor. This means less maintenance on visit functions as the traversal is shared by multiple visitors.


CohenArthur (approved):
yup, good idea

PR #2761, opened 2023-11-27
Make expand visitor inherit from default visitor

P-E-P:
Many visit functions in the expand visitor simply visit their components like the default visitor. Making the expand visitor inherit from the default visitor allows us to keep all visitor in sync without having to change every visitor.


philberty (approved):
LGTM

PR #2762, opened 2023-11-27
Make early name resolver inherit from default one

P-E-P:
Many visit functions in the early name resolver are just plain old traversal visit functions like the ones from the default visitor. Makes early name visitor inherit from the default visitor in order to avoid traversal duplication.


CohenArthur (approved):
hopefully this class gets removed soon anyway!

PR #2763, opened 2023-11-27
Make default resolver inherit from default visitor

P-E-P:
The default resolver put some scope in place but mostly has traversal functions similar to the default ast visitor, making it inherit from the default visitor allows us to avoid code duplication.


P-E-P:
> maybe we could remove the entire class and use the DefaultVisitor everywhere instead?

Not entirely sure, as AFAIK the Default resolver owns some member attributes ? Should be done in another PR anyway.

philberty (approved):
LGTM

PR #2764, opened 2023-11-28
Remove class ```AST::InherentImplItem```

powerboat9:


philberty (approved):
This seems lefit to me in the HIR i think we turn things into an ImplBlock so there is no distinction between items in a plain impl block vs a trait impl block.

P-E-P (approved):
LGTM

Issue #2379, opened 2023-07-05
Unicode Normalization of Identifiers


tamaroning:
All identifiers must be normalized to the NFC form.
Normalization is used to treat equally strings which consists of different codepoints but practiaclly the same.
It is useful when a program is linked to other Rust libraries,

NFC Normalization takes two steps:
1. Some kinds of characters are decomposed to one or multiple chatacters. (This is called NFD form)
2. Then, some kinds of chatacters are composed to a single character.  (This is called NFC form)

For example, 
https://unicode.org/reports/tr15/images/UAX15-NormFig4.jpg

## Implementation
libcpp already contains a executable which parses Unicode database (UnicodeData.txt, etc.) and generate ucnid.h, which can be used in C/C++. But uncid.h has an arrays each element of which represents a range of codepoint. Therefore, this is not usable for the normalization process. 

```cpp
static const struct ucnrange ucnranges[] = {
{   0|  0|  0|  0|  0|    0|    0|CID|NFC|NKC|  0,   0, 0x00a7 }, // codepoint U+0000 to U+00A7
{   0|  0|  0|C11|  0|    0|    0|CID|NFC|  0|  0,   0, 0x00a8 }, // codepoint U+00A8 to U+00A8
{   0|  0|  0|  0|  0|    0|    0|CID|NFC|NKC|  0,   0, 0x00a9 }, // codepoint U+00A9 to U+00A9
...
```

So I am going to create a huge switch statement to lookup.
Considering the future use, I think, normalization functions are put in a gcc shared folder but are no longer be put in libcpp/.

## Ref
- https://doc.rust-lang.org/reference/identifiers.html#normalization
- https://unicode.org/reports/tr15/

PR #2769, opened 2023-12-02
Implement Quick Check

tamaroning:
Addresses #2379 

I implemented Unicode Quick Check.
This checks if strings (usually identifiers) should be normalized or not before performing Unicode normalization.
QuickCheckResult::Yes shows the given string is already normalized, saving time and memory.
Strings should be normalized if the result is QuickCheckResult::No or QuickCheckResult::Maybe.

Also, I added Unicode data files which are compatible with GPL.
they are used to generate rust-unicode-data.h
See https://www.gnu.org/licenses/license-list.en.html#Unicode

```
gcc/rust/ChangeLog:

	* rust-system.h: Add <algorithm>.
	* util/make-rust-unicode.py: Output NFC_Quick_Check table.
	* util/rust-codepoint.h (struct Codepoint): Add is_supplementary
	method.
	* util/rust-unicode-data.h: Generated.
	* util/rust-unicode.cc (lookup_cc): Modified to use
	std::lower_bound.
	(is_alphabetic): Likewise.
	(nfc_quick_check): New function.
	(nfc_normalize): Use nfc_quick_check.
	(is_nfc_qc_maybe): New function.
	(is_nfc_qc_no): New function.
	(rust_nfc_qc_test): New test.
	* util/rust-unicode.h (is_nfc_qc_no): New function.
	(is_nfc_qc_maybe): New function.
	(enum class): New enum class.
	(nfc_quick_check): New function.
	(rust_nfc_qc_test): New test.
	* util/DerivedCoreProperties.txt: New file.
	* util/DerivedNormalizationProps.txt: New file.
	* util/UnicodeData.txt: New file.

Signed-off-by: Raiki Tamura <tamaron1203@gmail.com>
```


gcc/rust/util/make-rust-unicode.py
```diff
@@ -16,7 +16,15 @@
 # along with GCC; see the file COPYING3.  If not see
 # <http://www.gnu.org/licenses/>.
 
-# Run this program as
+# First, download the folloqing files from unicode.org
```
dkm, line 19:
```suggestion
# First, download the following files from unicode.org
```

tamaroning: Oh thanks😆

gcc/rust/util/DerivedCoreProperties.txt
CohenArthur, line 1: so these files are not present in any other GCC frontend? this is surprising. I guess it'll be a nice thing to bring higher in the GCC file hierarchy, so that other languages can benefit from this

CohenArthur (approved):
this is great, thank you :) is it possible to maybe add some unit testing? otherwise this looks good to me

gcc/rust/util/DerivedCoreProperties.txt
tamaroning: Omg, I completely missed the fact that contrib/unicode contains these files 😮😮
https://github.com/tamaroning/gccrs/tree/2a72592cf111f8f3dcc024d1f34ccd7fe6d36878/contrib/unicode

CohenArthur: ah, no problem! better late than never :D well done finding them

tamaroning: removed them and pushed just now

CohenArthur (approved):
this looks great :) thanks @tamaroning, good work!

gcc/rust/util/rust-unicode.cc
```diff
@@ -39,75 +39,35 @@ const uint32_t L_COUNT = 19, V_COUNT = 21, T_COUNT = 28;
...
+  if (it == ranges.end ())
+    return false;
   else
-    return -1;
+    return it->first <= target_cp && target_cp < it->second;
 }
 
 template <std::size_t SIZE>
-int64_t
+bool
```
P-E-P, line 59: Should we replace this function with a direct call to std::binary_search instead ?

tamaroning: Oh yes. i
I will fix it later

tamaroning:
@P-E-P 
I've busy for my university exams
Sorry for super late response!
i fixed everything.

P-E-P (approved):
LGTM, thank you!

PR #2770, opened 2023-12-03
Minor changes needed for borrowck

jdupak:
Sending them separatelly to not clutter the later bigger PRs.

For all commits:
- \[x] Run `clang-format`

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-type.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-type.cc
@@ -442,14 +442,9 @@ TypeCheckType::resolve_root_path (HIR::TypePath &path, size_t *offset,
       root_tyty = lookup;
 
       // this enforces the proper get_segments checks to take place
-      bool is_adt = root_tyty->get_kind () == TyTy::TypeKind::ADT;
-      if (is_adt)
-	{
-	  const TyTy::ADTType &adt
-	    = *static_cast<const TyTy::ADTType *> (root_tyty);
-	  if (adt.is_enum ())
-	    return root_tyty;
-	}
+      auto *maybe_adt = root_tyty->try_as<const TyTy::ADTType> ();
+      if (maybe_adt && maybe_adt->is_enum ())
+	return root_tyty;
     }
 
   return root_tyty;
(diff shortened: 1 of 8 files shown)


gcc/rust/typecheck/rust-hir-type-check-type.cc
```diff
@@ -442,12 +442,9 @@ TypeCheckType::resolve_root_path (HIR::TypePath &path, size_t *offset,
       root_tyty = lookup;
 
       // this enforces the proper get_segments checks to take place
-      bool is_adt = root_tyty->get_kind () == TyTy::TypeKind::ADT;
-      if (is_adt)
+      if (auto adt = root_tyty->try_as<const TyTy::ADTType> ())
```
philberty, line 445: Hmm i am not sure this is safe to do because try_as will return null if it is not an ADT Type so its not required to turn this into an ADT here.

philberty (changes requested):
This callable type interface could be useful but we may end up changing its name to FunctionItemTypes or some other but for not this is fine.

I think that for loop over the ADT Type check just needs changed and then this is good to be merged

gcc/rust/typecheck/rust-hir-type-check-type.cc
philberty: actually i am sorry i somehow read that as a for loop. But i would still prefer the old check, this is hard to read

jdupak: @philberty  This is something we should discuss more in depth, because you liked the API when I introduced it and now I am using it heavily in unpublished borrowchecker code.

jdupak:
Notice: I have added some small changed I missed before. Plase see the last force push diff.

gcc/rust/typecheck/rust-hir-type-check-type.cc
philberty: I like the API i just think in that specific if statement I don't like introducing variables like that. It looks too much like a for loop.

philberty: I don't have any problem with the try_as, is and as API I think its great.

CohenArthur: I agree that it would be more readable to initialize the variable above the `if` and then check for its validity in the `if`'s condition

CohenArthur:
I think this is great and am in favor of merging it if you do that `if` change :) thanks for the work!

philberty (approved):
LGTM

PR #2771, opened 2023-12-04
Borrowck ast lifetimes

### LINKED ISSUE #2756: Special lifetime indentifiers name not matched
Parser parses lifetime name `'static` as `static`, but `Parser<ManagedTokenSource>::lifetime_from_token (const_TokenPtr tok)` checks against the name with `'`. As a result, `'static` and `'_` are considerted to be ordinary named lifetimes.

Assign to me.

jdupak:
Depends on #2770

Fixes  #2756

### DIFF
--- a/gcc/rust/hir/rust-ast-lower-base.cc
+++ b/gcc/rust/hir/rust-ast-lower-base.cc
@@ -530,16 +530,24 @@ ASTLoweringBase::visit (AST::SelfParam &param)
 {}
 
 HIR::Lifetime
-ASTLoweringBase::lower_lifetime (AST::Lifetime &lifetime)
+ASTLoweringBase::lower_lifetime (AST::Lifetime &lifetime,
+				 bool default_to_static_lifetime)
 {
+  auto lifetime_type = lifetime.get_lifetime_type ();
+  if (lifetime_type == AST::Lifetime::WILDCARD && default_to_static_lifetime)
+    {
+      // If compiling in a static context.
+      lifetime_type = AST::Lifetime::STATIC;
+    }
+
   auto crate_num = mappings->get_current_crate ();
   Analysis::NodeMapping mapping (crate_num, lifetime.get_node_id (),
 				 mappings->get_next_hir_id (crate_num),
 				 UNKNOWN_LOCAL_DEFID);
   mappings->insert_node_to_hir (mapping.get_nodeid (), mapping.get_hirid ());
 
-  return HIR::Lifetime (mapping, lifetime.get_lifetime_type (),
-			lifetime.get_lifetime_name (), lifetime.get_locus ());
+  return HIR::Lifetime (mapping, lifetime_type, lifetime.get_lifetime_name (),
+			lifetime.get_locus ());
 }
 
 HIR::LoopLabel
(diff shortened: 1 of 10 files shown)


gcc/rust/hir/rust-ast-lower-base.cc
```diff
@@ -530,16 +530,24 @@ ASTLoweringBase::visit (AST::SelfParam &param)
...
-ASTLoweringBase::lower_lifetime (AST::Lifetime &lifetime)
+ASTLoweringBase::lower_lifetime (AST::Lifetime &lifetime,
+				 bool default_to_static_lifetime)
 {
+  auto lifetime_type = lifetime.get_lifetime_type ();
+  if (lifetime_type == AST::Lifetime::WILDCARD && default_to_static_lifetime)
+    {
+      // If compiling in a static context.
+      lifetime_type = AST::Lifetime::STATIC;
+    }
```
CohenArthur, line 541:
```suggestion
  // If compiling in a static context.
  if (lifetime_type == AST::Lifetime::WILDCARD && default_to_static_lifetime)
      lifetime_type = AST::Lifetime::STATIC;
```

GNU style nit, sorry

gcc/rust/hir/rust-ast-lower-type.h
```diff
@@ -79,7 +80,14 @@ class ASTLoweringType : public ASTLoweringBase
...
 private:
-  ASTLoweringType () : ASTLoweringBase (), translated (nullptr) {}
+  ASTLoweringType (bool default_to_static_lifetime)
+    : ASTLoweringBase (),
+      default_to_static_lifetime (default_to_static_lifetime),
+      translated (nullptr)
+  {}
+
+  /** Used when compiling const and static items. */
+  bool default_to_static_lifetime;
```
CohenArthur, line 90: I don't get how this flag is used or set. If we are lowering an `AST::ConstantItem` or `AST::StaticItem`, we know we are doing so and can pass an explicit `true` flag to `ASTLoweringType` (or even better, a `LifetimeDefault::Static` enum variant which is clearer). and we could have a default value to the bool argument to the `ASTLoweringType` argument?

does that make sense? or is it important that we keep it this way for future pull-requests?

CohenArthur (approved):
looks good :) thanks for all the work :heart:

gcc/rust/hir/rust-ast-lower-type.h
jdupak: This flag is used to get the state across the visitor API and propagate further `ASTLoweringType::translate` calls since the visitor API does not allow more parameters. I think it makes sense. Does it make sense now?

CohenArthur: ah, yeah I see what you mean. so we really can't do without a default parameter then?

jdupak: Do you mean the member variable or the parameter? We need the member variable, the parameter is useful because we don't need to change it everywhere.

CohenArthur: I meant the parameter - but I'm fine with keeping it this way, your explanation makes sense. I'll merge this now

Issue #2767, opened 2023-12-01
Emit error on async trait functions


P-E-P:
you can add the same check for `async` functions, even if that is about to change soon.

_Originally posted by @CohenArthur in https://github.com/Rust-GCC/gccrs/pull/2753#discussion_r1411796349_
            

```
error[E0706]: functions in traits cannot be declared `async`
 --> src/lib.rs:2:5
  |
2 |     async fn titi();
  |     -----^^^^^^^^^^^
  |     |
  |     `async` because of this
  |
  = note: `async` trait functions are not currently supported
  = note: consider using the `async-trait` crate: https://crates.io/crates/async-trait
  = note: see issue #91611 <https://github.com/rust-lang/rust/issues/91611> for more information
```

braw-lee:
I want to fix this

braw-lee:
I added a check and test for `async` function inside a trait but gccrs fails to recognize `async` as a token inside trait impl
`test.rs:6:5: error: unrecognised token ‘async’ for item in trait impl`

`    6 |     async fn f() -> u32 {`
`      |     ^~~~~`
`test.rs:6:5: error: failed to parse trait impl item in trait impl`
`test.rs:6:5: error: failed to parse item in crate`

braw-lee:
is `async` token not fully supported yet? I was trying to step through the lexer with gdb to find out but for some reason it skips over some code even with -O0 flag enabled

P-E-P:
> is `async` token not fully supported yet? I was trying to step through the lexer with gdb to find out but for some reason it skips over some code even with -O0 flag enabled

AFAIK `async` token is supported, we already have multiple async nodes in our ast. For example the `parse_async_item` function [make use of it indirectly](https://godbolt.org/z/sv6a8z5K8).

braw-lee:
yes, looks like `async` is supported

[this](https://godbolt.org/z/4d3abT1Pc) makes me think there is an unhandled case for `async` token when we are inside a trait
I will look into it in a few days

tschwinge:
@P-E-P, @braw-lee, @CohenArthur, is it correct that this issue got closed via #2944 "Merge upstream, 2024-03-09" (why wasn't it then already closed via a GCC/Rust master branch commit?), or remains there work to be done here?

braw-lee:
https://github.com/Rust-GCC/gccrs/pull/2779
the second commit of this PR fixed this issue

braw-lee:
[...]

PR #2779, opened 2023-12-15
Handle `async` qualifier inside trait

### LINKED ISSUE #2778: Unhandled `async` while parsing items inside trait
[error](https://godbolt.org/z/5Kor8j5M9)
compiler does not recognize `async` as a keyword which is bug

braw-lee:
Fixes #2778

gcc/rust/ChangeLog:

	* parse/rust-parse-impl.h (Parser::parse_trait_impl_item): Handled `async` similar to `const` (Parser::parse_pattern): clang-format.

in the git commit message.

- \[x] Run `clang-format`

order to get feedback.
---

### DIFF
--- a/gcc/rust/checks/errors/rust-ast-validation.cc
+++ b/gcc/rust/checks/errors/rust-ast-validation.cc
@@ -107,6 +107,11 @@ ASTValidation::visit (AST::Function &function)
     rust_error_at (function.get_locus (), ErrorCode::E0379,
 		   "functions in traits cannot be declared const");
 
+  // may change soon
+  if (qualifiers.is_async () && context.back () == Context::TRAIT_IMPL)
+    rust_error_at (function.get_locus (), ErrorCode::E0706,
+		   "functions in traits cannot be declared %<async%>");
+
   if (valid_context.find (context.back ()) == valid_context.end ()
       && function.has_self_param ())
     rust_error_at (
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -5784,6 +5784,8 @@ Parser<ManagedTokenSource>::parse_trait_impl_item ()
       // function or method
       return parse_trait_impl_function_or_method (visibility,
 						  std::move (outer_attrs));
+    case ASYNC:
+      return parse_async_item (visibility, std::move (outer_attrs));
     case CONST:
       // lookahead to resolve production - could be function/method or const
       // item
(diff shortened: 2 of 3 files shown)


braw-lee:
second commit fixes #2767

braw-lee:
@philberty review pls

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -5784,6 +5785,7 @@ Parser<ManagedTokenSource>::parse_trait_impl_item ()
       // function or method
       return parse_trait_impl_function_or_method (visibility,
 						  std::move (outer_attrs));
+    case ASYNC:
```
CohenArthur, line 5788: `const` is handled this way because inside a trait, you could have a const variable or a constant function:

```rust
impl Foo for u32 {
    const TRAIT_ASSOC_CST: u32 = 15;
    
    const fn constant_fn() -> u32 { 14 }
}
```

for async, we can only have `async` functions - so we don't need to bundle the cases for `CONST` and `ASYNC` together.

Instead, this is what I would like to see:

```suggestion
    case ASYNC:
        return parse_async_item (visibility, std::move(outer_attrs))
    case CONST:
         /* snip */
```
```

File: gcc/rust/checks/errors/rust-ast-validation.cc [hunk 2]
```diff
@@ -107,6 +107,11 @@ ASTValidation::visit (AST::Function &function)
     rust_error_at (function.get_locus (), ErrorCode::E0379,
 		   "functions in traits cannot be declared const");
 
+  // may change soon
+  if (qualifiers.is_async () && context.back () == Context::TRAIT_IMPL)
+    rust_error_at (function.get_locus (), ErrorCode::E0706,
+		   "functions in traits cannot be declared %<async%>");
```
CohenArthur, line 113: this is great :)

CohenArthur (commented):
the testcase is great, and if you fix the comment I've added this looks good to me for merging :) thanks for working on this!

braw-lee:
fixed suggested changes, thanks for review @CohenArthur

CohenArthur (approved):
LGTM! If you'd like @braw-lee you could implement the same sort of pattern for [trait functions](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=305e9b412509616bfbd50274b9afddd2) and parsing of async functions for [regular impls](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f2c6f9f4651c31d500066bb0d1c80370)

PR #2784, opened 2023-12-30
Added newline to get more readable lexdump

### LINKED ISSUE #2783: Lex dump is hard to read
I was using the flag -frust-dump-lex to see the parsed tokens but the output file `gccrs.lex.dump` was very hard to read because all tokens are written onto the same line.
Instead of ending the token information line with " " we can end it with '\n' to make it more readable

braw-lee:
Fixes #2783

The line giving information about token now ends with '\n' rather than " " to make the gccrs.lex.dump file more readable. This file is generated using flag -frust-dump-lex

### DIFF
--- a/gcc/rust/lex/rust-lex.cc
+++ b/gcc/rust/lex/rust-lex.cc
@@ -240,7 +240,7 @@ Lexer::dump_and_skip (int n)
 				     + std::string (tok->get_type_hint_str ()))
 				  : "")
 	      << " ";
-	  out << Linemap::location_to_string (loc) << " ";
+	  out << Linemap::location_to_string (loc) << '\n';
 	}
 
       token_queue.skip (0);


CohenArthur (approved):
looks good to me, thanks :) asking for reviews from the others, I can't think of anything this would break so it should be good to go

gcc/rust/lex/rust-lex.cc
```diff
@@ -240,7 +240,7 @@ Lexer::dump_and_skip (int n)
 				     + std::string (tok->get_type_hint_str ()))
 				  : "")
 	      << " ";
-	  out << Linemap::location_to_string (loc) << " ";
+	  out << Linemap::location_to_string (loc) << '\n';
```
dkm, line 243: is `std::endl` better for this?

P-E-P: `std::endl` [will flush the buffer](https://en.cppreference.com/w/cpp/io/manip/endl) AFAIK we don't need to ensure the output buffer is flushed at this point and we should keep the chosen flushing strategy (IIRC it is linemode by default, so it should not change that much in most situations).

philberty (approved):

PR #2786, opened 2024-01-03
Handle `async` functions in traits

### LINKED ISSUE #2785: Unhandled `async` for trait items
[error](https://godbolt.org/z/d6jvc16G6)
compiler does not handle `async` keyword for trait item

braw-lee:
Fixes #2785 
I opened this pull request because I have some doubts.
So I created a switch-case for `async` keyword when reading trait items. I generate an error (I know its not the right way) and return nullptr.
The returning nullptr generates additional errors that results in the test case failing.
The correct way to generate error would be to add check in `ASTValidation::visit (AST::TraitFunctionDecl &decl)` function but it won't work right now because we never made the AST node.
To solve this problem I think we need to implement `parse_trait_async()` and to implement this function we need to make an AST node `AST::TraitItemAsync`.
But this node will be pretty much useless as `async` isn't allowed inside traits so I don't know if this is the right way to approach this problem
@CohenArthur Is there any other way to solve this problem, I am sorry if I failed to reach an obvious solution, lemme know.

### DIFF
--- a/gcc/rust/checks/errors/rust-ast-validation.cc
+++ b/gcc/rust/checks/errors/rust-ast-validation.cc
@@ -149,7 +149,7 @@ ASTValidation::visit (AST::TraitFunctionDecl &decl)
 		       "functions in traits cannot be declared %<async%>");
       if (qualifiers.is_const ())
 	rust_error_at (decl.get_identifier ().get_locus (), ErrorCode::E0379,
-		       "functions in traits cannot be declared const");
+		       "functions in traits cannot be declared %<const%>");
     }
 }
 
--- a/gcc/rust/checks/errors/rust-ast-validation.cc
+++ b/gcc/rust/checks/errors/rust-ast-validation.cc
@@ -141,9 +141,16 @@ ASTValidation::visit (AST::TraitFunctionDecl &decl)
 {
   const auto &qualifiers = decl.get_qualifiers ();
 
-  if (context.back () == Context::TRAIT && qualifiers.is_const ())
-    rust_error_at (decl.get_identifier ().get_locus (), ErrorCode::E0379,
-		   "functions in traits cannot be declared const");
+  if (context.back () == Context::TRAIT)
+    {
+      // may change soon
+      if (qualifiers.is_async ())
+	rust_error_at (decl.get_identifier ().get_locus (), ErrorCode::E0706,
+		       "functions in traits cannot be declared %<async%>");
+      if (qualifiers.is_const ())
+	rust_error_at (decl.get_identifier ().get_locus (), ErrorCode::E0379,
+		       "functions in traits cannot be declared const");
+    }
 }
 
 void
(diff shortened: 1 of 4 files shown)


P-E-P:
> function but it won't work right now because we never made the AST node. To solve this problem I think we need to implement `parse_trait_async()` and to implement this function we need to make an AST node `AST::TraitItemAsync`.

We usually only create a node for a given syntax tree component, the `async` keyword gives us information but cannot have any "child", it contains the asyncness information. [Function declarations](https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/ast/rust-item.h#L2542) store the asyncness in [the function qualifiers](https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/ast/rust-item.h#L536).

Even though the `async` should probably be rejected in the ast validation stage, the compiler struggles at an earlier stage (during parsing) per [your example](https://godbolt.org/z/d6jvc16G6). It expects some token and find another, it already [covered all possible cases](https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/parse/rust-parse-impl.h#L5227) and expect a macro call.

This means we should modify the parsing stage to accept the `async` token, store the asyncness information in the function qualifiers and reject an async function declaration in traits at a later stage during ast validation [probably around here](https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/checks/errors/rust-ast-validation.cc#L143) like it has been done with `const` functions.

Hopefully this is clearer ?

braw-lee:
yes, this is clearer. Thanks for help.

braw-lee:
@P-E-P is this PR good to merge?

gcc/rust/checks/errors/rust-ast-validation.cc
```diff
@@ -141,9 +141,16 @@ ASTValidation::visit (AST::TraitFunctionDecl &decl)
...
-		   "functions in traits cannot be declared const");
+  if (context.back () == Context::TRAIT)
+    {
+      // may change soon
+      if (qualifiers.is_async ())
+	rust_error_at (decl.get_identifier ().get_locus (), ErrorCode::E0706,
+		       "functions in traits cannot be declared %<async%>");
+      if (qualifiers.is_const ())
+	rust_error_at (decl.get_identifier ().get_locus (), ErrorCode::E0379,
+		       "functions in traits cannot be declared const");
```
CohenArthur, line 152:
```suggestion
		       "functions in traits cannot be declared %<const%>");
```

since you're changing this line anyway, let's have consistent formatting here :)

CohenArthur (approved):
looks great to me! thanks!

gcc/rust/checks/errors/rust-ast-validation.cc
braw-lee: will need to fix the corresponding test case too, so will create another commit.

CohenArthur: yes, that's all good :) thanks a lot

CohenArthur (approved):
thanks a lot!

PR #2789, opened 2024-01-06
Handle `async` keyword for regular implementations

### LINKED ISSUE #2788: Unhandled `async` for regular implementations.
[error](https://godbolt.org/z/reKjavnEd)
compiler does not handle `async` keyword for regular implementations.

braw-lee:
Fixes #2788 
I am not sure how to test this as we don't have to check for error. I was reading up on `async` functions usage and I think Tokio is required to write small code snippets, which isn't supported by gccrs yet.

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -5600,6 +5600,7 @@ Parser<ManagedTokenSource>::parse_inherent_impl_item ()
 	    return nullptr;
 	  }
       }
+    case ASYNC:
     case EXTERN_KW:
     case UNSAFE:
     case FN_KW:
--- a/gcc/testsuite/rust/compile/issue-2788.rs
+++ b/gcc/testsuite/rust/compile/issue-2788.rs
@@ -1,0 +1,10 @@
+// { dg-additional-options "-frust-compile-until=lowering" }
+struct Foo {
+    arg_1: u32,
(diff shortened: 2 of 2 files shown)


P-E-P:
> Fixes #2788 I am not sure how to test this as we don't have to check for error. I was reading up on `async` functions usage and I think Tokio is required to write small code snippets, which isn't supported by gccrs yet.

@braw-lee You don't need `tokio` tu use async in rust, there are a lot of different executors (`async-std`, `smol`, IIRC even `embassy` makes use of the async keyword) and you could even do without those.

You're right though when saying that it won't be easy to test as we cannot lower async yet. We only want to make sure the syntax is correct. Here is what you can do, since you modified the parser you could test that it behaves correctly. To do so you need to create an async rust snippet and add it to the testsuite under `gcc/testsuite/rust/compile` with the following first line:

```rust
// { dg-additional-options "-frust-compile-until=lowering" }
```

This tells the testsuite to pass a flag to the compiler in order to stop before the lowering step. There are a multiple other compile steps, tune the value to your needs but keep in mind the compiler will stop *before* the selected step.

braw-lee:
thanks, I will try this soon

braw-lee:
@P-E-P as the compiler will stop before lowering step, I made the test case as minimal as possible. Let me know if you were expecting a more verbose test case.

CohenArthur (approved):
LGTM

PR #2791, opened 2024-01-12
expand: C++ify proc macro decls generation

CohenArthur:
This uses a stringstream instead of the older, better and more readable
`sprintf` version as the format overflow warning in current GCC is
overzealous and triggered on that code, despite the non-zero allocation.

Even using an unsigned value instead of a signed one for the `size`
variable caused issues, so switching to this is simpler.

gcc/rust/ChangeLog:

	* expand/rust-proc-macro.cc: Use stringstream instead of sprintf.
	* rust-system.h: Include <iomanip>.


CohenArthur:
the code was too simple anyway

P-E-P (approved):
LGTM.

PR #2795, opened 2024-01-15
Emit error when range pattern is empty

braw-lee:
I couldn't find a cleaner way to get values of upper and lower bound. Tell me if there is a better way to get those values.
@CohenArthur @P-E-P

### DIFF
--- a/gcc/rust/checks/errors/rust-ast-validation.cc
+++ b/gcc/rust/checks/errors/rust-ast-validation.cc
@@ -21,6 +21,8 @@
 #include "rust-diagnostics.h"
 #include "rust-item.h"
 #include "rust-keyword-values.h"
+#include "rust-pattern.h"
+#include "safe-ctype.h"
 
 namespace Rust {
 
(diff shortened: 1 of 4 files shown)


P-E-P:
I'm unsure this should be done at this stage of the compiler, this surely is simpler after a bit of desugaring. Have you taken a look at rustc's way of doing this ? Is it doing it this early ?

powerboat9:
https://godbolt.org/z/sThP53WbG

It looks like the error only happens in match-like statements

braw-lee:
I wonder if that is the intended behaviour. I will ask in rustc if this is what they want or a missed case.

CohenArthur:
@braw-lee do you want some help on moving this to the typechecking phase?

philberty:
Hi @braw-lee this branch needs rebased before it can be merged

philberty:
This is a year old now. Please feel free to rebase and reopen this PR

powerboat9:
Remember to reopen it *before* rebasing -- github doesn't like it when you modify a branch between when a PR is closed and when it's reopened

PR #2801, opened 2024-01-17
rust_debug: Cast size_t values to unsigned long before printing.

CohenArthur:
Using %lu to format size_t values breaks 32 bit targets, and %zu is not
supported by one of the hosts GCC aims to support - HPUX

gcc/rust/ChangeLog:

	* backend/rust-compile-base.cc (HIRCompileBase::resolve_method_address):
	Cast size_t value to unsigned long.
	* expand/rust-proc-macro.cc (load_macros): Likewise.
	* typecheck/rust-hir-type-check-expr.cc (TypeCheckExpr::visit): Likewise.

in the git commit message.

- \[ ] Run `clang-format`

order to get feedback.
---

### DIFF
--- a/gcc/rust/backend/rust-compile-base.cc
+++ b/gcc/rust/backend/rust-compile-base.cc
@@ -918,7 +918,8 @@ HIRCompileBase::resolve_method_address (TyTy::FnType *fntype,
     }
 
   const Resolver::PathProbeCandidate *selectedCandidate = nullptr;
-  rust_debug_loc (expr_locus, "resolved to %lu candidates", candidates.size ());
+  rust_debug_loc (expr_locus, "resolved to %lu candidates",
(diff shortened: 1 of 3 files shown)


dkm (approved):

7Z0t99:
I see that this was already merged, but I still wonder whether this solution is sound on all platforms, because sizeof(unsigned long) < sizeof(size_t) on some platforms, e.g. on win64.

CohenArthur:
@7Z0t99 you're right that it is unsound, the correct fix would be to use the `PRIu64` format specifier. this is a quick and dirty fix as this breaks the build on some GCC targets. it is unsound in that it would display incorrect length values if we have more than `UINT32_MAX` proc macros at once for example, which is a much bigger issue and shouldn't happen :) these are debug prints anyway and will probably be removed soon.

PR #2805, opened 2024-01-18
testsuite: Fix missing handling of little endian.

CohenArthur:
Some failures occur in the testsuite because we
did not account for the big-endian case.

gcc/testsuite/ChangeLog:

	* rust/compile/issue-1446.rs: Add swap_bytes function.
	* rust/compile/iterators1.rs: Remove unused {to, from}_le functions.

Before merging this I would like to run our testsuite on the compile farm on a SPARC machine to see that they're really fixed. Will report here once it's done.

### DIFF
--- a/gcc/testsuite/rust/compile/issue-1446.rs
+++ b/gcc/testsuite/rust/compile/issue-1446.rs
@@ -1,3 +1,11 @@
+// fake function
+pub fn swap_bytes(this: u32) -> u32 {
+    (((this) & 0xff000000) >> 24)
+        | (((this) & 0x00ff0000) >> 8)
+        | (((this) & 0x0000ff00) << 8)
+        | (((this) & 0x000000ff) << 24)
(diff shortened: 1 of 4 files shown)


thesamesam:
I can give you SSH to our beefy sparc machine (just email me an SSH key + desired username). It needs kicking for a reboot every so often but it's fine otherwise.

It's the same machine used for the sourceware buildbot.

CohenArthur:
thanks @thesamesam I appreciate it a lot :) I've tested these changes on one of the machines of the GCC compile farm but I'll definitely take you up on that offer :D I'll email you!

CohenArthur:
not sure why this causes a regression on MacOS. I'll ssh into the machine and investigate later

philberty (approved):
LGTM lets ignore macos its going to be too hard now that the new m2 chip is out

PR #2806, opened 2024-01-19
Set the default ABI to C for extern blocks and extern functions

nobel-sh:
Previously, the default ABI was set to Rust, which is incorrect for extern blocks and extern functions. This patch changes the default ABI to C for these cases.

Reference: https://doc.rust-lang.org/reference/items/functions.html#extern-function-qualifier
Zulip discussion:  https://gcc-rust.zulipchat.com/#narrow/stream/266897-general/topic/extern.20blocks/near/346429045

gcc/rust/ChangeLog:

	* hir/rust-ast-lower-base.cc (ASTLoweringBase::lower_qualifiers): (ASTLoweringBase::lower_extern_block): Change ABI to C for extern

### DIFF
--- a/gcc/rust/hir/rust-ast-lower-base.cc
+++ b/gcc/rust/hir/rust-ast-lower-base.cc
@@ -718,8 +718,8 @@ ASTLoweringBase::lower_qualifiers (const AST::FunctionQualifiers &qualifiers)
   Unsafety unsafety
     = qualifiers.is_unsafe () ? Unsafety::Unsafe : Unsafety::Normal;
   bool has_extern = qualifiers.is_extern ();
+  ABI abi = has_extern ? ABI::C : ABI::RUST;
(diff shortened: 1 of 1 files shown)


CohenArthur (approved):
LGTM, waiting on @philberty as he owns that part of the code. thanks for looking into this :)

philberty (commented):
Is this the right thing? iirc we can do:

extern {
}

I assumed this would default to Rust ABI. Do you have a reference for this it would be good to capture that.

Otherwise the code looks good.

philberty:
apologies i see it now: 
"Just as with [external block](https://doc.rust-lang.org/reference/items/external-blocks.html), when the extern keyword is used and the "ABI" is omitted, the ABI used defaults to "C". That is, this:"

Thanks for linking this.

PR #2810, opened 2024-01-22
Remove ```TraitImplItem```

powerboat9:
This moves us closer to unifying handling of associated items


P-E-P (approved):

braw-lee:
I was working on another PR [link](https://github.com/Rust-GCC/gccrs/pull/2811/commits/d5ab6c5c8ba46bd3a305d7bb1d0f926c00a06417), I want the Function class to inherit from TraitItem so we can remove TraitItemFunc and TraitItemMethod classes. Do you think it is a good idea? @powerboat9

Issue #2787, opened 2024-01-04
Unify `TraitItemMethod` & `TraitItemFunc` with `Function`


P-E-P:
All those node share the same attribute members, `TraitItemMethod` used to be separated from `TraitItemFunc` because of the `Self` param but it is now handled as a regular parameter, this means the distinction between those two should not be required anymore.

Furthermore, `TraitItemMethod` and `TraitItemFunc` were used because a `Function` could not have an optional body. This is no longer the case since the introduction of the AST validation pass. This means those former classes should have no reason to exist anymore.

Solving this issue should greatly reduce the amount of "duplicated" code in the parser.

- [ ] Change the parser to use a function declaration instead of trait specific function classes
- [ ] Remove trait specific function classes
- [ ] Remove deadcode & fix ast passes on trait items
  - [ ] AST dump
  - [ ] AST validation
  - [ ] Lowering
  - [ ] ... ?

braw-lee:
can I take this?

PR #2811, opened 2024-01-23
Unify TraitItemMethod & TraitItemFunc with Function

braw-lee:
Addresses #2787


braw-lee:
This PR is getting a bit bigger, but I am not sure if I can divide it into seperate buildable commits.

PR #2813, opened 2024-01-23
Parse normal functions with `self` parameter correctly

### LINKED ISSUE #2812: Parser cannot handle normal functions with self parameter
[error](https://godbolt.org/z/YdYzxYfa1)
Normal functions are not supposed to have self parameter but compiler should be able to still parse this and generate error [here](https://github.com/Rust-GCC/gccrs/blob/1586de85db580fa6fe5a3698c148d6b36209f8ff/gcc/rust/checks/errors/rust-ast-validation.cc#L119)

braw-lee:
Fixes #2812 
Might use same function to parse normal function and trait functions so should be a useful fix.


braw-lee:
will need this soon for #2787 @P-E-P @CohenArthur

CohenArthur (approved):
Great stuff, thank you :)

PR #2814, opened 2024-01-23
Borrowck: add support for regions (lifetimes) in TyTy

jdupak:
@philberty

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check.h
+++ b/gcc/rust/typecheck/rust-hir-type-check.h
@@ -23,6 +23,7 @@
 #include "rust-tyty.h"
 #include "rust-hir-trait-reference.h"
 #include "rust-autoderef.h"
+#include "rust-tyty-region.h"
 
 #include <stack>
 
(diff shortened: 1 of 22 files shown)


gcc/rust/typecheck/rust-hir-type-check.h
```diff
@@ -79,6 +82,49 @@ class TypeCheckContextItem
   Item item;
 };
 
+/**
+ * Interned lifetime representation in TyTy
+ *
+ * On the HIR->TyTy boundary HIR::Lifetime is interned into this struct.
+ */
+class Lifetime
```
philberty, line 90: I dont think you need to change this now but we already have AST::Lifetime and HIR::Lifetime, I wonder should this really be moved to its own namespace at some point Resolver is ok but not sure its the right name.

philberty (approved):
LGTM nice work maybe just have a think if you want this stuff to live inside the Resolver namespace it might be a good idea to have its own namespace but I think this is fine for now.

gcc/rust/typecheck/rust-hir-type-check.h
jdupak: My thinking for the future was that the interning might be pushed up to AST to avoid passing the strings around. Then this class would be pointless.
Right now it has only very short life - it is used to resolve lifetimes (syntactic objects) into regions (semantic objects). All of it use happens in the resolver.

PR #2815, opened 2024-01-26
Unify TraitItemMethod & TraitItemFunc with Function

### LINKED ISSUE #2787: Unify `TraitItemMethod` & `TraitItemFunc` with `Function`

braw-lee:
Fixes #2787


braw-lee:
Gets rid of duplicated classes in AST and related functions.
Can further reduce duplicated code by getting rid of classes that inherit from AST::TraitItem, but that would be for another PR.
Is it good for merge? @P-E-P

P-E-P:
> Gets rid of duplicated classes in AST and related functions. Can further reduce duplicated code by getting rid of classes that inherit from AST::TraitItem, but that would be for another PR.
You're right, there already is a lot of deletion in this PR, let's not introduce further changes in this PR, you may do this in another one.

> Is it good for merge?
I've assigned some additional reviewers as this PR "deletes" a lot of code. I don't have much time for now and will review it later.

gcc/rust/hir/rust-ast-lower-implitem.h
```diff
@@ -240,28 +240,32 @@ class ASTLowerTraitItem : public ASTLoweringBase
...
+
+    // set self parameter to error if this is a method
+    // else lower to hir
+    HIR::SelfParam self_param = func.has_self_param ()
+				  ? lower_self (func.get_self_param ())
+				  : HIR::SelfParam::error ();
 
     std::vector<HIR::FunctionParam> function_params;
-    for (auto &p : ref.get_function_params ())
+    for (auto &p : func.get_function_params ())
```
P-E-P, line 268: Maybe we should introduce a member function in AST to retrieve regular (non self|variadic) params ? I'm almost sure the current API might lead to some errors in the future. Would it be really useful ?

Should not be done in this PR anyway.

P-E-P (approved):
Looks good to me. We even have a proper error when parsing an unexpected token in traits! Great work.

gcc/rust/hir/rust-ast-lower-implitem.h
```diff
@@ -240,28 +240,32 @@ class ASTLowerTraitItem : public ASTLoweringBase
...
     std::vector<std::unique_ptr<HIR::WhereClauseItem> > where_clause_items;
     HIR::WhereClause where_clause (std::move (where_clause_items));
     HIR::FunctionQualifiers qualifiers
-      = lower_qualifiers (func.get_trait_function_decl ().get_qualifiers ());
+      = lower_qualifiers (func.get_qualifiers ());
 
     std::vector<std::unique_ptr<HIR::GenericParam> > generic_params;
-    if (ref.has_generics ())
+    if (func.has_generics ())
       {
```
P-E-P, line 252: Nit: We could get rid of those curly braces.

braw-lee:
also, is it ok to make commits like "run clang-format" or do you prefer editing each commit to satisfy clang-format? @P-E-P

P-E-P:
> also, is it ok to make commits like "run clang-format" or do you prefer editing each commit to satisfy clang-format? @P-E-P

Clang format shall be applied on all commits (IIRC we have a CI step to check that `Clang format Lint / clang-format`). If you're mentioning this because of the curly braces on the `if`, it is not detected by clang-format, that's why we have to manually check. But don't bother too much about that, if you miss one someday it's fine.

braw-lee:
actually I missed clang-format on earlier commits, so the 7th commit is currently fixing the format. Do you want me to fix previous commits and get rid of the 7th commit? 
I think the CI does not verify clang-format for each commit individually, rather it runs clang-format after all changes are done.

braw-lee:
this can't be merged anymore due to merge conflicts with #2821 @P-E-P

philberty (approved):
LGTM

philberty:
looks like this needs fixed up @braw-lee

braw-lee:
> looks like this needs fixed up @braw-lee

actually it was good to merge before #2821 was merged. I would suggest reverting #2821 and merging this first as #2821 involves separating function definitions into .cc files which will be easier to redo than fixing this one. 
Let me know if don't want to revert #2821 , I will try to fix this one then. @philberty

braw-lee:
@CohenArthur resolved merge conflicts, still working somehow ;)

PR #2823, opened 2024-02-02
Borrowck: variance analysis

jdupak:
Depends on #2814 (merged)

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-item.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-item.cc
@@ -25,6 +25,7 @@
 #include "rust-hir-trait-resolve.h"
 #include "rust-substitution-mapper.h"
 #include "rust-type-util.h"
+#include "rust-tyty-variance-analysis.h"
 
 namespace Rust {
 namespace Resolver {
@@ -204,7 +205,7 @@ TypeCheckItem::visit (HIR::TupleStruct &struct_decl)
   TyTy::ADTType::ReprOptions repr
     = parse_repr_options (attrs, struct_decl.get_locus ());
 
-  TyTy::BaseType *type = new TyTy::ADTType (
+  auto *type = new TyTy::ADTType (
     struct_decl.get_mappings ().get_hirid (), mappings->get_next_hir_id (),
     struct_decl.get_identifier ().as_string (), ident,
     TyTy::ADTType::ADTKind::TUPLE_STRUCT, std::move (variants),
@@ -215,6 +216,8 @@ TypeCheckItem::visit (HIR::TupleStruct &struct_decl)
 
   context->insert_type (struct_decl.get_mappings (), type);
   infered = type;
+
+  context->get_variance_analysis_ctx ().add_type_constraints (*type);
 }
 
 void
@@ -266,7 +269,7 @@ TypeCheckItem::visit (HIR::StructStruct &struct_decl)
   TyTy::ADTType::ReprOptions repr
     = parse_repr_options (attrs, struct_decl.get_locus ());
 
-  TyTy::BaseType *type = new TyTy::ADTType (
+  auto *type = new TyTy::ADTType (
     struct_decl.get_mappings ().get_hirid (), mappings->get_next_hir_id (),
     struct_decl.get_identifier ().as_string (), ident,
     TyTy::ADTType::ADTKind::STRUCT_STRUCT, std::move (variants),
@@ -277,6 +280,8 @@ TypeCheckItem::visit (HIR::StructStruct &struct_decl)
 
   context->insert_type (struct_decl.get_mappings (), type);
   infered = type;
+
+  context->get_variance_analysis_ctx ().add_type_constraints (*type);
 }
 
 void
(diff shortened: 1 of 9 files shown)


gcc/rust/typecheck/rust-hir-type-check-item.cc
```diff
@@ -26,6 +26,8 @@
 #include "rust-substitution-mapper.h"
 #include "rust-type-util.h"
 
+#include <rust-tyty-variance-analysis.h>
```
philberty, line 29: I think stick to single quotes for these includes like the others

gcc/rust/typecheck/rust-tyty-subst.cc
```diff
@@ -588,6 +588,16 @@ SubstitutionRef::get_used_arguments () const
 {
   return used_arguments;
 }
+tl::optional<SubstitutionArg>
```
philberty, line 591: can you add a space here between the previous block and this one

philberty (approved):
just some minor style nits. Otherwise this is pretty advanced stuff its going to take a while before I can follow it all properly but it looks very consistent which is a very good sign

gcc/rust/typecheck/rust-hir-type-check-item.cc
jdupak: Fixed

gcc/rust/typecheck/rust-tyty-subst.cc
jdupak: Fixed

gcc/rust/typecheck/rust-tyty-variance-analysis.cc
```diff
@@ -0,0 +1,540 @@
...
+void
+add_type_constraints (ADTType &type)
+{
+  GenericTyVisitorCtx::get ().process_type (type);
+}
+
+void
+solve ()
+{
+  auto ctx = GenericTyVisitorCtx::get ();
```
CohenArthur, line 27: I'm not a big fan of adding more global context to the frontend, but if it can't be helped them I'm also fine with it. note that this might cause hurt down the line with the compilation of multiple crates, proc macros, possible refactors for a more re-entrant frontend, etc

gcc/rust/typecheck/rust-tyty-variance-analysis.cc
```diff
@@ -0,0 +1,540 @@
...
+	{
+	  for (const auto &field : variant->get_fields ())
+	    add_constraints_from_ty (field->get_field_type (), V::covariant ());
+	}
+    }
+
+  // Cleanup
+  first_lifetime = 0;
+  first_type = 0;
+  param_names.clear ();
```
CohenArthur, line 163: I think we could possibly avoid this with a non-global GenericTyVisitorCtx, but again it might cause more harm than good to make it non-global - so I'm trusting you

gcc/rust/typecheck/rust-tyty-variance-analysis.cc
```diff
@@ -0,0 +1,540 @@
...
+	     + to_string (*term.transform.rhs) + ")";
+    }
+  rust_unreachable ();
+}
+
+std::string
+GenericTyVisitorCtx::to_string (SolutionIndex index) const
+{
+  // Search all values in def_id_to_solution_index_start and find key for
+  // largest value smalled than index
```
CohenArthur, line 285:
```suggestion
  // largest value smaller than index
```
?

gcc/rust/typecheck/rust-tyty-variance-analysis.cc
```diff
@@ -0,0 +1,540 @@
...
+	      add_constraints_from_ty (arg.value ().get_tyty (), variance_i);
+	    }
+	}
+    }
+}
+
+Variance
+TyVisitorCtx::contra (Variance variance)
+{
+  return V::transform (variance, V::contravariant ());
```
CohenArthur, line 511: not a big fan of having the `using V` for this line - it would make it clearer to use the full qualified name please

CohenArthur (commented):
a couple of comments but this looks very good and clean :) thanks a lot for this work

gcc/rust/typecheck/rust-tyty-variance-analysis.cc
jdupak: I guess, we could tie it to the typecheck context and pass a pointer to this method. Does that look good to you?

jdupak:
@CohenArthur I have split the context for one type processing from the global context to make it reentrant and integrated the global one in the typecheck context.

gcc/rust/typecheck/rust-hir-type-check.h
```diff
@@ -24,6 +24,7 @@
 #include "rust-hir-trait-reference.h"
 #include "rust-autoderef.h"
 #include "rust-tyty-region.h"
+#include "rust-tyty-variance-analysis.h"
 
 #include <stack>
```
CohenArthur, line 29: just noticed this, from d42498b69e27. I had missed it! we should not be including system headers like this directly, we have to be careful with this. I can open an issue but you don't have to be the one to fix it if you don't want to

CohenArthur: #2920

jdupak: It will be easier to fix in next PR, than to do all the rebasing and testing. Especially since you already made the issue.

CohenArthur (approved):
thanks! that's a lot of very good work :)

PR #2824, opened 2024-02-03
Remove horrible path expression generic hack

### LINKED ISSUE #2723: Cannot compile generic method on generic struct
I encountered this bug when investigating rustc's mangling scheme.

I tried this code:

```rust
#[lang = "sized"]
pub trait Sized {}

struct S<T>(T);

impl S<i32> {
    fn f<S>(t: S) -> S {
        t
    }
}

pub fn main() {
    S::<i32>::f::<i32>(0);
}

```

rustc compiles this code.

Instead, this happened: 

```bash
a.rs:13:19: error: generic item takes at most 1 type arguments but 1 were supplied
    7 |     fn f<S>(t: S) -> S {
      |          ~         
......
   13 |     S::<i32>::f::<i32>(0);
      |                   ^~~
a.rs:13:5: error: Failed to resolve expression of function call
   13 |     S::<i32>::f::<i32>(0);
      |     ^
```

### Meta

philberty:
This patch introduces one regression because generics are getting better
understood over time. The code here used to apply generics with the same
symbol from previous segments which was a bit of a hack with out limited
inference variable support. The regression looks like it will be related
to another issue which needs to default integer inference variables much
more aggresivly to default integer.

Fixes #2723


philberty:
I have reopening #1773 its only working by accident as far as I can tell.

Issue #2800, opened 2024-01-17
Update copyright to 2024


CohenArthur:
make sure libgrust is handled properly

dkm:
note for whoever is going to look into this: This is in `contrib/update-copyright.py`

sahilyeole:
I would like to work on it :v:

PR #2842, opened 2024-02-09
Update copyright years to 2024

sahilyeole:
Fixes: #2800
Made separate commits for:
1) Changing the update-copyright.py script to include `libgrust`.
2) Updating copyright years in `gcc/rust` and `libgrust`.

### DIFF
--- a/contrib/update-copyright.py
+++ b/contrib/update-copyright.py
@@ -775,6 +775,7 @@ class GCCCmdLine (CmdLine):
         self.add_dir ('libgfortran')
         # libgo is imported from upstream.
         self.add_dir ('libgomp')
+        self.add_dir ('libgrust')
         self.add_dir ('libiberty')
         self.add_dir ('libitm')
         self.add_dir ('libobjc')
@@ -802,6 +803,7 @@ class GCCCmdLine (CmdLine):
             'libgcc',
             'libgfortran',
             'libgomp',
+            'libgrust',
             'libiberty',
             'libitm',
             'libobjc',
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -1,6 +1,6 @@
 # Make-lang.in -- Top level -*- makefile -*- fragment for GCC Rust frontend.
 
-# Copyright (C) 2009-2023 Free Software Foundation, Inc.
+# Copyright (C) 2009-2024 Free Software Foundation, Inc.
 
 # This file is part of GCC.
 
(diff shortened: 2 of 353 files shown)


dkm:
Thank you for this. Some notes...

"3300 files changes" => This alone should be a hint that maybe, something's odd. And indeed, you are modifying many many files outside of the rust frontend. These files need to be "bumped", but in fact, they already have... in the upstream repository. Our version is still in 2023 because we haven't merged upstream for some times now.

You changes will cause trouble when we actually merge upstream, as both upstream and us are modifying the same locations.

So, bottom line: don't touch any files outside of `gcc/rust` and `libgrust/` for bumping the copyright date.

The only change you have to do outside of these 2 dir is for modifying the script responsible for the bump:
- 1 commit for the script
- 1 commit for bumping rust (and only rust) copyright years

If you see that some other files need to be bumped outside the rust directories, that could be part of a followup change.

Also, we need commit messages. For the script, a regular GNU changelog part is needed. For the commit that only bumps the year, you can use the "Update copyright years" comment (see previous year) that is recognized by the checkers (and you don't need to add a gazilion lines of GNU changelog for every files).

gcc/rust/ChangeLog
```diff
@@ -3238,7 +3238,7 @@
 	* operator.h: New.
 
 
-Copyright (C) 2022-2023 Free Software Foundation, Inc.
+Copyright (C) 2022-2024 Free Software Foundation, Inc.
```
dkm, line 3241: Never touch ChangeLog file, these automatically modified by various scripts :)

dkm (changes requested):

gcc/rust/ChangeLog
sahilyeole: Sure 👍 
Are there any other changes you would recommend?

CohenArthur (commented):
this looks good, thanks :) but you'll need to fix the formatting for both commits, which the commit format checker complains about:

```
Checking 14545605e35ecc9be022efa89cd6f28079e7d36c: FAILED
ERR: cannot find a ChangeLog location in message
Checking 017758e8eca260e5443ba8c6afee56fe38d17d92: FAILED
ERR: line should start with a tab: "* update-copyright.py: Add libgrust folder."
```

For the first issue, you need to add a period to the end of your commit message, so `Update copyright years.`. This is what the commit format script checks for:

```python
        # Skip Update copyright years commits
        if self.info.lines and self.info.lines[0] == 'Update copyright years.':
            return
```

(note the period at the end).

for the second commit, as the error says you need to use a tabulation instead of spaces characters before the mentioned line. sorry about this!

CohenArthur (approved):
thanks @sahilyeole, great job :)

PR #2849, opened 2024-02-13
Unify `AST::ExternalFunctionItem` with `AST::Function`

0xn4utilus:
Added support for external function in AST::Function. It can be verified it in AST validation as follows. But in line 6113 of rust-parse-impl.h, `parse_external_item()` needs return type ExternalItem from parse_function() which returns `Function`, thus we cannot directly use `parse_function()`. 
How should I proceed ?

### DIFF
--- a/gcc/rust/ast/rust-item.h
+++ b/gcc/rust/ast/rust-item.h
@@ -1289,7 +1289,7 @@ protected:
 class LetStmt;
 
 // Rust function declaration AST node
-class Function : public VisItem, public AssociatedItem
+class Function : public VisItem, public AssociatedItem, public ExternalItem
 {
   FunctionQualifiers qualifiers;
   Identifier function_name;
@@ -1300,6 +1300,7 @@ class Function : public VisItem, public AssociatedItem
   tl::optional<std::unique_ptr<BlockExpr>> function_body;
   location_t locus;
   bool is_default;
+  bool is_external_function;
 
 public:
   std::string as_string () const override;
@@ -1330,16 +1331,17 @@ public:
 	    std::unique_ptr<Type> return_type, WhereClause where_clause,
 	    tl::optional<std::unique_ptr<BlockExpr>> function_body,
 	    Visibility vis, std::vector<Attribute> outer_attrs,
-	    location_t locus, bool is_default = false)
+	    location_t locus, bool is_default = false,
+	    bool is_external_function = false)
     : VisItem (std::move (vis), std::move (outer_attrs)),
-      qualifiers (std::move (qualifiers)),
+      ExternalItem (Stmt::node_id), qualifiers (std::move (qualifiers)),
       function_name (std::move (function_name)),
       generic_params (std::move (generic_params)),
       function_params (std::move (function_params)),
       return_type (std::move (return_type)),
       where_clause (std::move (where_clause)),
       function_body (std::move (function_body)), locus (locus),
-      is_default (is_default)
+      is_default (is_default), is_external_function (is_external_function)
   {}
 
   // TODO: add constructor with less fields
@@ -1363,6 +1365,8 @@ public:
     return function_params.size () != 0
 	   && function_params.back ()->is_variadic ();
   }
+
+  bool is_external () const { return is_external_function; }
 
   // Invalid if block is null, so base stripping on that.
   void mark_for_strip () override { function_body = nullptr; }
@@ -1422,6 +1426,9 @@ public:
     return function_params[0];
   }
 
+  // ExternalItem::node_id is same as Stmt::node_id
+  NodeId get_node_id () const override { return Stmt::node_id; }
+
 protected:
   /* Use covariance to implement clone function as returning this object
    * rather than base */
@@ -1430,6 +1437,13 @@ protected:
   /* Use covariance to implement clone function as returning this object
    * rather than base */
   Function *clone_associated_item_impl () const override
+  {
+    return new Function (*this);
+  }
+
+  /* Use covariance to implement clone function as returning this object
+   * rather than base */
+  Function *clone_external_item_impl () const override
   {
     return new Function (*this);
   }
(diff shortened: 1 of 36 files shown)


powerboat9:
You'd probably want to make ```Function``` inherit from ```ExternalItem```

0xn4utilus:
Wanted to do that, but will it be a good practice? Also it generates ambiguous functions, that can be probably tackled with virtual and override.

P-E-P:
You're right. We'll need to inherit from `ExternalItem` in order to return a function from this function but it'll be probably harder than expected.

At first I thought this would cause a lot of troubles, turns out we have a dead `ExternalItem` class in `rust-item.h` that should probably be deleted.

The real `ExternalItem` lies in `rust-ast.h`. It duplicates `node_id` with `Stmt` (`Function` -> `VisItem` -> `Stmt`). We could make `ExternalItem` inherit from item instead of `Visitable` but this surely have some unwanted outcome I can't see right now.

0xn4utilus:
Yes, will try out some things and see if it works.

powerboat9:
I'm pretty sure you can fix ambiguous functions by overriding them in ```Function```, which wouldn't be ideal, but might be the best option.

0xn4utilus:
Yes overriding works, now I have to make corresponding modifications in `Rust::HIR::ASTLoweringExternItem`.

0xn4utilus:
I am getting an unexpected bug. Where should I start looking?

```rust
extern "C"{
        fn fun(a:i32);
}
```

test.rs:2:25: error: unknown reference for resolved name: ‘i32’
    2 |         fn fun(a:i32);
      |                         ^

powerboat9:
It doesn't look like the types of external function parameters are being resolved. Either that, or external items in the ```extern "C" {}``` block aren't being resolved. What happens if you have an external function that has no parameters but returns an ```i32```?

0xn4utilus:
```rust
extern "C"{
        fn fun()->i32;
}
```
has the same error. So probably external function items are not being resolved

gcc/rust/ast/rust-item.h
```diff
@@ -1422,6 +1426,14 @@ class Function : public VisItem, public AssociatedItem
     return function_params[0];
   }
 
+  NodeId get_node_id () const override
```
powerboat9, line 1429: You'd probably want to create a new constructor for ```ExternalItem``` that allows you to pass in a ```NodeId```, just to make sure the ```ExternalItem``` isn't storing a different node id. It's not too big an issue, but it might be a footgun in the future.

0xn4utilus: Thanks for the review :) I'll make the necessary changes. I have university exams this week, so my progress on this might be a bit slower.

gcc/rust/checks/errors/rust-feature-gate.cc
```diff
@@ -131,7 +131,8 @@ FeatureGate::visit (AST::MacroRulesDefinition &rules_def)
 void
 FeatureGate::visit (AST::Function &function)
 {
-  check_rustc_attri (function.get_outer_attrs ());
+  if (!function.is_external ())
+    check_rustc_attri (function.get_outer_attrs ());
```
powerboat9, line 135: Why disable this check?

gcc/rust/resolve/rust-ast-resolve-item.cc
```diff
@@ -1038,9 +1039,12 @@ ResolveExternItem::visit (AST::ExternalFunctionItem &function)
...
   // correctly
-  for (auto &param : function.get_function_params ())
-    if (!param.is_variadic ())
-      ResolveType::go (param.get_type ().get ());
+  for (auto &it : function.get_function_params ())
+    if (!it.get ()->is_variadic ())
+      {
+	auto param = static_cast<AST::FunctionParam *> (it.get ());
+	ResolveType::go (param->get_type ().get ());
+      }
```
powerboat9, line 1047: You should be able to replace ```it.get ()->``` with ```it->```

powerboat9 (commented):
Looks good, just a few nitpicks

powerboat9:
You'd also need to fix the build errors and format the commit messages

gcc/rust/checks/errors/rust-feature-gate.cc
0xn4utilus: I believe this check is for Function only and not External Functions,  without  `if (!function.is_external ())` compiler was not able detect `#![feature(intrinsics)]` in case of external function attributes as in `issue-1031.rs`.

P-E-P (commented):
I'm not a fan of the NodeId duplication, I'll take a deeper look in the upcoming days to be sure there's no way around.

Otherwise, it looks good.

gcc/rust/ast/rust-item.h
```diff
@@ -1422,6 +1426,14 @@ class Function : public VisItem, public AssociatedItem
...
+  {
+    if (is_external_function)
```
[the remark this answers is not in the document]
0xn4utilus: yes they are same, but I wanted to keep it explicit just in case for clarity, maybe a simple comment will be fine too

0xn4utilus:
Requested changes made and removed `AST::ExternalFunctionItem` dead code.

powerboat9:
Looks good. You might want to split off some of these commits into separate pull requests, so that they can get reviewed/merged faster -- that's what I tend to do

0xn4utilus:
> Looks good. You might want to split off some of these commits into separate pull requests, so that they can get reviewed/merged faster -- that's what I tend to do

Sure

P-E-P (approved):

0xn4utilus:
Rebased, now it should be able to merge

P-E-P:
> Rebased, now it should be able to merge

Not entirely sure, sometimes the queue is doing weird things and we have to merge by hand. Let's try again.

PR #2851, opened 2024-02-13
Fix variadic check on all function parameters

### LINKED ISSUE #2850: Variadic check only on last parameter
I tried this code: https://godbolt.org/z/4r5KGPYeE

```rust
// test case 1
fn myfun0(...,_:i32) {}

// test case 2
fn myfun1(a:i32,...,_:i32) {}

// test case 3
struct z {
    x: f64,
    y: f64,
}

impl z {
    fn new(x: f64, ..., y: f64) -> z {
        z { x: x, y: y }
    }
}

```

I expected to see compilation error as seen when the below code is run.

```rust
fn myfun0(_:i32,...) {}
```

Instead, this happened: 

```<source>:5:4: warning: function is never used: 'myfun0'
    5 | fn myfun0(...,_:i32) {}
      |    ^~~~~~
<source>:7:4: warning: function is never used: 'myfun1'
    7 | fn myfun1(a:i32,...,_:i32) {}
      |    ^~~~~~
<source>:9:1: warning: struct is never constructed: 'z'
    9 | struct z {
      | ^~~~~~
<source>:15:8: warning: associated function is never used: 'new'
   15 |     fn new(x: f64, ..., y: f64) -> z {
      |        ^~~
<source>:7:11: warning: unused name 'a' [-Wunused-variable]
    7 | fn myfun1(a:i32,...,_:i32) {}
      |           ^
Compiler returned: 0
```

0xn4utilus:
Fixes #2850.

### DIFF
--- a/gcc/rust/checks/errors/rust-ast-validation.cc
+++ b/gcc/rust/checks/errors/rust-ast-validation.cc
@@ -132,10 +132,14 @@ ASTValidation::visit (AST::Function &function)
 	rust_error_at (function.get_locus (), "free function without a body");
     }
 
-  if (function.is_variadic ())
-    rust_error_at (
-      function.get_function_params ().back ()->get_locus (),
-      "only foreign or %<unsafe extern \"C\"%> functions may be C-variadic");
+  auto &function_params = function.get_function_params ();
+  for (auto it = function_params.begin (); it != function_params.end (); it++)
+    {
+      if (it->get ()->is_variadic ())
+	rust_error_at (it->get ()->get_locus (),
+		       "only foreign or %<unsafe extern \"C\"%> functions may "
+		       "be C-variadic");
+    }
 
   AST::ContextualASTVisitor::visit (function);
 }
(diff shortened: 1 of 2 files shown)


CohenArthur (approved):
LGTM, I agree that we need a test case :) Have a look at `gcc/testsuite/rust/compile/invalid_variadics.rs` to see what failing testcases should look like, and feel free to ask for help if you need! One last thing, could you change the title of your PR to something more descriptive please? Thanks!

P-E-P:
I'm not entirely sure whether we should consider an invalid variadic function that will be rejected by the AST validation pass as "variadic". I can't think of any example where we would want to know whether an invalid function is variadic, macros don't care about this.
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5e3b59bc82ced0dd9c81fc72dc1da441

We could keep this as it is for now until someone find an issue with the loop being in the ast check instead of being inside the `is_variadic`.

Either way, we should add this message about the variadic function being incorrect in the AST validation pass.

dkm:
You can squash both commit, so the test comes with the actual code :)

PR #2802, opened 2024-01-17
Merge gcc/trunk into us

CohenArthur:
Strategy used was checking out *our* changes for `gcc/rust`, `libgrust` and `gcc/testsuite/rust`, but checking out *theirs* for the remaining folders.

Two commits were added on top of things to conform to new APIs that were introduced by trunk in the meantime. These were fixed in #2744 but immediately squashed with the offending commits... I think it might be worth doing the same here but not sure. Should have published them as separate PRs when I did them however :) sorry about that

PR #2853, opened 2024-02-15
libgrust: Add missed commits for Changelog creation

CohenArthur:
- libgrust: Add ChangeLog file
- libgrust: Add entry for maintainers


tschwinge (approved):
That's OK if necessary to unblock #2822, as I understand.

There are, however, further changes that need to be backported from upstream GCC master branch into GCC/Rust master branch, to account for rework during upstreaming.  I'll (eventually...) look into that in context of #2802.

(Should've used `git cherry-pick -x` here.)  ;-P

CohenArthur:
@tschwinge re-cherry-picked them with the -x option :) thanks! didn't know it existed

P-E-P (approved):

PR #2861, opened 2024-02-19
Fix false positive for top-level ```AltPattern```

powerboat9:


CohenArthur:
@powerboat9 do you have a testcase where this is shown? the fix looks good to me

powerboat9:
Had to fiddle with the test a bit, but it works now.

CohenArthur (approved):
Thanks :D

PR #2862, opened 2024-02-19
gccrs: fix powif64 intrinsic

dkm:
The signature was incorrectly using an i64 for the integer power,
instead of an i32.


CohenArthur (approved):
oh, good catch. thanks

PR #2863, opened 2024-02-19
gccrs: minor cleanup in langhook.type_for_mode

dkm:
gcc/rust/ChangeLog:

	* rust-lang.cc (grs_langhook_type_for_mode): simplify code for
        *Imode. Add missing long_double_type_node.

### DIFF
--- a/gcc/rust/rust-lang.cc
+++ b/gcc/rust/rust-lang.cc
@@ -172,38 +172,29 @@ static tree
 grs_langhook_type_for_mode (machine_mode mode, int unsignedp)
 {
   // TODO: change all this later to match rustc types
+  if (mode == QImode)
+    return unsignedp ? unsigned_intQI_type_node : intQI_type_node;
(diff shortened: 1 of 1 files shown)


gcc/rust/rust-lang.cc
```diff
@@ -172,38 +172,29 @@ static tree
 grs_langhook_type_for_mode (machine_mode mode, int unsignedp)
 {
   // TODO: change all this later to match rustc types
```
CohenArthur, line 174: hehe, we should definitely do this at some point

CohenArthur (approved):
LGTM, thanks!

gcc/rust/rust-lang.cc
dkm: Yes... At first I wanted to see if I could DTRW, but then figured it's a bit more involved than hacking a few lines...

PR #2867, opened 2024-02-22
Remove references to smart pointers in the AST API

P-E-P:
Remove references to smart pointer as those could be replaced with references to the underlying object.

### DIFF
--- a/gcc/rust/expand/rust-expand-visitor.cc
+++ b/gcc/rust/expand/rust-expand-visitor.cc
@@ -360,7 +360,7 @@ ExpandVisitor::expand_struct_fields (std::vector<AST::StructField> &fields)
 {
   for (auto &field : fields)
     {
-      maybe_expand_type (field.get_field_type ());
+      maybe_expand_type (field.get_field_type_ptr ());
     }
 }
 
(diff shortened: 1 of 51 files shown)


gcc/rust/expand/rust-expand-visitor.cc
```diff
@@ -328,27 +328,27 @@ ExpandVisitor::expand_inner_stmts (AST::BlockExpr &expr)
...
 {
   expander.push_context (MacroExpander::ContextType::EXPR);
-  expr->accept_vis (*this);
+  expr.accept_vis (*this);
   expander.pop_context ();
 
   auto final_fragment = expander.take_expanded_fragment ();
   if (final_fragment.should_expand ()
       && final_fragment.is_expression_fragment ())
-    expr = final_fragment.take_expression_fragment ();
```
powerboat9, line 340: It doesn't seem like this would work, since instances of derived classes of ```Expr``` can have different sizes

P-E-P: You're right. I believe the only way out is to keep reference to the unique pointer for type/expr expansion.

PR #2871, opened 2024-02-23
Fix FixMe in changing return type of builtin_macro_from_string() from BuiltinMacro to tl::optional<BuiltinMacro>

badumbatish:
```bash
                === rust Summary ===
# of expected passes            8402
# of expected failures          69
# of unsupported tests          1
```
- \[x] Run `clang-format`

---

I noticed a FixMe in `rust-macro-builtins.cc` that asks to change the return type of a function from BuiltInMacro to be wrapped in an optional so I changed it.

I then ran the codebase through `make -s -j16` to make sure there isn't any compiler error.
I then ran `make check-rust` to make sure there isn't any broken test-case. Results are as above.

I then run `clang-format`, add all the files edited and do a `git commit --signoff` and push them, creating a pull-request

### DIFF
--- a/gcc/rust/expand/rust-macro-builtins.cc
+++ b/gcc/rust/expand/rust-macro-builtins.cc
@@ -126,8 +126,7 @@ std::unordered_map<std::string, AST::MacroTranscriberFunc>
     {"Hash", MacroBuiltin::proc_macro_builtin},
 };
 
-// FIXME: This should return an tl::optional
-BuiltinMacro
+tl::optional<BuiltinMacro>
 builtin_macro_from_string (const std::string &identifier)
 {
   auto macro = MacroBuiltin::builtins.lookup (identifier);
--- a/gcc/rust/expand/rust-macro-builtins.h
+++ b/gcc/rust/expand/rust-macro-builtins.h
@@ -75,7 +75,7 @@ enum class BuiltinMacro
   Hash,
 };
 
-BuiltinMacro
+tl::optional<BuiltinMacro>
 builtin_macro_from_string (const std::string &identifier);
 
 /**
--- a/gcc/rust/resolve/rust-early-name-resolver.cc
+++ b/gcc/rust/resolve/rust-early-name-resolver.cc
@@ -497,7 +497,7 @@ EarlyNameResolver::visit (AST::MacroInvocation &invoc)
     {
       auto builtin_kind
 	= builtin_macro_from_string (rules_def->get_rule_name ().as_string ());
-      invoc.map_to_builtin (builtin_kind);
+      invoc.map_to_builtin (builtin_kind.value ());
     }
 
   auto attributes = rules_def->get_outer_attrs ();


powerboat9:
You should be using ```git clang-format master``` instead of directly using ```clang-format```. You'd have to copy https://github.com/Rust-GCC/gccrs/blob/master/contrib/clang-format to a new file named ```.clang-format``` in the root of your local git repo -- that'll ensure ```git clang-format``` uses the correct formatting style.

badumbatish:
I see, I was wondering why clang-format change a bunch of places

badumbatish:
@powerboat9 hi there, i'm not sure about the guildline that my commit messages should adhere to? Any pointers?

dkm:
Hello @badumbatish 
The commit message should contain a valid GNU ChangeLog. See https://gcc.gnu.org/codingconventions.html#ChangeLogs
The easiest is probably to look at existing logs :) Also, all our commits should be "correct", in the sense that they build, pass the test, etc. Before merging the final change, you should fixup/squash all commits where this applies. In your case, the reindent commit should be squashed with the original change for example :)

To contribute to GCC, you can assign your copyright to the FSF (the historic way, still the most used) or use the DCO. For the latter, you'll need to add a Signed-off-by line at the end of your commit (use `git commit -s`) with your real name and email.

dkm:
> I tried adding tab on each of my changelog but it still fails with the same message.
> 
> ```
> python3 contrib/gcc-changelog/git_check_commit.py
> Checking fdcf88ac52c3ec7e0a2342048ecd6a0eddcf0477: FAILED
> ERR: line should start with a tab: "    * expand/rust-macro-builtins.cc: Change BuiltinMacro in"
> ERR: line should start with a tab: "    builtin_macro_from_string to tl::optional<>"
> ERR: line should start with a tab: "    * expand/rust-macro-builtins.h (enum class): Change BuiltinMacro"
> ERR: line should start with a tab: "    in builtin_macro_from_string to tl::optional<>"
> ERR: line should start with a tab: "    * resolve/rust-early-name-resolver.cc (EarlyNameResolver::visit):"
> ERR: line should start with a tab: "    Resolved wrong type dependency of builtin_macro_from_string"
> ```

Usually caused by editor replacing tabs with spaces when saving... How did you fix the issue? :)

CohenArthur:
@badumbatish you probably have options in your neovim config about replacing tab characters with spaces, something like `expandtab` or `set tabstop=2` etc (sorry, it's been a while since I used neovim).

an easy way to make sure you're inputting a tab character is to do `Ctrl+V` and then press `tab` in Insert mode.

also, have a look at running the script called `contrib/gcc-git-customization.sh` - it will unlock a git subcommand called `git gcc-commit-mklog` which acts the same as `git commit` but adds a "skeleton" for your Changelog. so it'll already have the nice formatting and tabs and everything

CohenArthur (approved):
fix looks good, thank you!

dkm:
> also, have a look at running the script called `contrib/gcc-git-customization.sh` - it will unlock a git subcommand called `git gcc-commit-mklog` which acts the same as `git commit` but adds a "skeleton" for your Changelog. so it'll already have the nice formatting and tabs and everything

Another solution if the git customization is not wanted/possible (it assumes you have a git access to upstream gcc...) is to call the script by hand, works nicely. For example:
```
$ git show | ~/git/gcc/contrib/mklog.py | wl-copy
```
Is the command I sometimes use to scaffold the changelog part and copy it in the copy buffer :). You can send this python script any diff and it spits some changelog :)

badumbatish:
> > I tried adding tab on each of my changelog but it still fails with the same message.
> > ```
> > python3 contrib/gcc-changelog/git_check_commit.py
> > Checking fdcf88ac52c3ec7e0a2342048ecd6a0eddcf0477: FAILED
> > ERR: line should start with a tab: "    * expand/rust-macro-builtins.cc: Change BuiltinMacro in"
> > ERR: line should start with a tab: "    builtin_macro_from_string to tl::optional<>"
> > ERR: line should start with a tab: "    * expand/rust-macro-builtins.h (enum class): Change BuiltinMacro"
> > ERR: line should start with a tab: "    in builtin_macro_from_string to tl::optional<>"
> > ERR: line should start with a tab: "    * resolve/rust-early-name-resolver.cc (EarlyNameResolver::visit):"
> > ERR: line should start with a tab: "    Resolved wrong type dependency of builtin_macro_from_string"
> > ```
> 
> Usually caused by editor replacing tabs with spaces when saving... How did you fix the issue? :)

i edited all my changelogs and then I go into nvim and do set noet, retab! i think :) It turns every space to a tab, then i can hit save and recheck it again!

badumbatish:
> git show | ~/git/gcc/contrib/mklog.py | wl-copy

thank you this is really helpful!

PR #2873, opened 2024-02-25
Improve error [E0164] handling

### LINKED ISSUE #2846: Empty ```TupleStructPattern``` assumed to be function call
I tried this code:

```rust
struct S();

pub fn main() {
    match S() {
        S() => {}
    }
}
```

Instead, this happened:
```
<source>:5:9: error: expected tuple struct or tuple variant, found associated function [E0164]
    5 |         S() => {}
      |         ^
      |         function calls are not allowed in patterns
```

sahilyeole:
This PR is an improvement over #2565.
It checks for path segment size to ensure error [E0164] is correctly emitted in case of associated functions in a pattern.

Fixes #2846.

gcc/rust/ChangeLog:

	* resolve/rust-early-name-resolver.cc (EarlyNameResolver::visit):
	Use segment size to identify associated functions for [E0164].

### DIFF
--- a/gcc/rust/resolve/rust-early-name-resolver.cc
+++ b/gcc/rust/resolve/rust-early-name-resolver.cc
@@ -558,6 +558,17 @@ EarlyNameResolver::visit (AST::StructPattern &)
 void
 EarlyNameResolver::visit (AST::TupleStructPattern &pattern)
 {
+  auto &segments = pattern.get_path ().get_segments ();
+  if (segments.size () > 1 && !pattern.has_items ())
+    {
+      rich_location rich_locus (line_table, pattern.get_locus ());
+      rich_locus.add_fixit_replace (
+	"function calls are not allowed in patterns");
+      rust_error_at (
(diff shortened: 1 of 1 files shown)


CohenArthur (commented):
@sahilyeole could you add a testcase for this? to ensure we don't introduce regressions down the line. thanks :)

sahilyeole:
> @sahilyeole could you add a testcase for this? to ensure we don't introduce regressions down the line. thanks :)

I'm not sure what testcase I can add since testcase for [E0164] was already added in #2565 and the code mentioned in #2846 now gives ICE (because it seems tuple struct patterns are not implemented yet), instead of [E0164].

CohenArthur:
@sahilyeole you just need to rebase this on the latest master and we can merge it :)

CohenArthur:
@sahilyeole the CI is failing because the API changed a little bit for the code you're using :) if you adapt it we can merge it

PR #2874, opened 2024-02-25
First stab at issue 2855 by spltting the two maps

badumbatish:
@CohenArthur 
Hi there, I'm looking for some review/feedback on this issue. I just finished splitting the BiMap and unordered_map of the MacroBuiltin class. 

I'm planning on putting the non-member functions' declaration in `gcc/rust/expand/rust-macro-builtins.h` but ran into some error with the compiler not knowing Parser as a class, etc etc.


gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -0,0 +1,21 @@
...
+  {"llvm_asm", BuiltinMacro::LlvmAsm},
+  {"global_asm", BuiltinMacro::GlobalAsm},
+}};
+
+const std::unordered_map<std::string, AST::MacroTranscriberFunc>
+  asm_builtin_transcribers = {
+    {"asm", MacroBuiltin::sorry},
+    {"llvm_asm", MacroBuiltin::sorry},
+    {"global_asm", MacroBuiltin::sorry},
+};
```
CohenArthur, line 19: I like this idea of splitting the maps, but I think it might also make code harder to read and follow. I think it would be better to keep the maps all in one place if possible - either in `rust-macro-builtins.h` or `rust-macro-builtins.cc`

CohenArthur (commented):
what error did you run into specifically? I saw your comment about circular inclusion on Zulip, I think we should be able to find a solution

gcc/rust/expand/rust-macro-builtins-asm.cc
CohenArthur: so keep both maps as they originally were, but split off the handlers into files as you've done

badumbatish:
> what error did you run into specifically? I saw your comment about circular inclusion on Zulip, I think we should be able to find a solution

So initially i want to move the declaration of all non-member functions inside the header and add the necessary headers to make the declaration complete, one of the files was rust-parse.h. When i added it, it creates a circular dependency like this

`rust-parse.h -> rust-item.h -> rust-expr.h -> rust-macro.h -> rust-macro--builtins.h -> rust-parse.h`

I was a bit impatient and decided to put all the non-member functions declaration inside a new header to avoid this circular dependency.
The logs are in this chain of messages if you want to take a look. It's about 20000 lines long (jeez) https://gcc-rust.zulipchat.com/#narrow/stream/327528-GSoC/topic/issue.202855

gcc/rust/expand/rust-macro-builtins-asm.cc
badumbatish: gotcha

CohenArthur:
> > what error did you run into specifically? I saw your comment about circular inclusion on Zulip, I think we should be able to find a solution
> 
> So initially i want to move the declaration of all non-member functions inside the header and add the necessary headers to make the declaration complete, one of the files was rust-parse.h. When i added it, it creates a circular dependency like this
> 
> `rust-parse.h -> rust-item.h -> rust-expr.h -> rust-macro.h -> rust-macro--builtins.h -> rust-parse.h`
> 
> I was a bit impatient and decided to put all the non-member functions declaration inside a new header to avoid this circular dependency. The logs are in this chain of messages if you want to take a look. It's about 20000 lines long (jeez) https://gcc-rust.zulipchat.com/#narrow/stream/327528-GSoC/topic/issue.202855

ah, I see! so I think one way to do it would indeed be to put the "parsing" function declarations in `rust-macro-builtins.h`. these functions should not have any structs/enums from `rust-parse.h` in their declarations, as they are supposed to create instances of parsers unless they are helper functions.

another possibility is to indeed split the parsing functions into their own header + source, which includes `rust-parse.h` but isn't included by `rust-macro-builtins.h`. then, the builtin macro handlers (in `rust-macro-builtin-asm` for example) can include that header to gain access to the parsing functions.

badumbatish:
i must have missed these errors on my local build, i'll try running the same command as via github actions

CohenArthur:
@badumbatish is this ready for review? if not, could you put it in Draft please? :) thanks

badumbatish:
> @badumbatish is this ready for review? if not, could you put it in Draft please? :) thanks

hi! it is ready for review

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -0,0 +1,21 @@
+//
+// Created by jjsm on 2/25/24.
```
dkm, line 2: Please use the regular header with the license & boilerplate text instead. I guess it's coming from your editor?

gcc/rust/expand/rust-macro-builtins-asm.h
```diff
@@ -0,0 +1,20 @@
+//
+// Created by jjsm on 2/25/24.
```
dkm, line 2: ditto

gcc/rust/expand/rust-macro-builtins-helpers.cc
```diff
@@ -0,0 +1,256 @@
...
+
+std::vector<std::unique_ptr<AST::Expr>>
+try_expand_many_expr (Parser<MacroInvocLexer> &parser,
+		      const TokenId last_token_id, MacroExpander *expander,
+		      bool &has_error)
+{
+  auto restrictions = Rust::ParseRestrictions ();
+  // stop parsing when encountered a braces/brackets
+  restrictions.expr_can_be_null = true;
+  // we can't use std::optional, so...
```
dkm, line 115: we have `tl:optionnal`, so you can have a "clean" option :)
EDIT: and I see you are using it in other parts. So maybe there's something else preventing the use of tl::? If so, maybe adjust the comment. Or maybe it's an oversight and you can refactor... Your call :)

gcc/rust/expand/rust-macro-builtins-helpers.cc
```diff
@@ -0,0 +1,256 @@
...
+}
+
+std::string
+source_relative_path (std::string path, location_t locus)
+{
+  std::string compile_fname = LOCATION_FILE (locus);
+
+  auto dir_separator_pos = compile_fname.rfind (file_separator);
+
+  /* If there is no file_separator in the path, use current dir ('.').  */
```
dkm, line 221: nit: you're mixing both C and C++ comment style. I don't think we have a strict rule, but I think it's best to be a bit uniform in the same file, even more when everything is new :)

dkm (commented):
minor remarks. I'll let the real devs do a real review :)

gcc/rust/expand/rust-macro-builtins-helpers.cc
badumbatish: oops i think this is just me copying-verbatim from before haha. I'll see what I can do about this

gcc/rust/expand/rust-macro-builtins-asm.h
badumbatish: I'll change these soon

gcc/rust/expand/rust-macro-builtins-helpers.cc
badumbatish: I'll resolve to the C++ // style

badumbatish: I did a bit of grepping in the gcc/rust/expand folder and saw that there is a lot of C style commenting

```
jjasmine@jjasmine ~/C/g/g/r/expand (split_up_rust_macro_builtsin)> grep -F "/*" * -l
rust-cfg-strip.cc
rust-cfg-strip.h
rust-derive-clone.cc
rust-derive-clone.h
rust-derive-copy.cc
rust-derive-copy.h
rust-derive.h
rust-expand-visitor.cc
rust-expand-visitor.h
rust-macro-builtins-helpers.cc
rust-macro-builtins-helpers.h
rust-macro-builtins-trait.cc
rust-macro-builtins-utility.cc
rust-macro-builtins.cc
rust-macro-builtins.h
rust-macro-expand.cc
rust-macro-expand.h
rust-macro-substitute-ctx.h
rust-proc-macro-invoc-lexer.h
rust-proc-macro.cc
rust-proc-macro.h
```
I'm not sure what i should do in the case of these.

gcc/rust/expand/rust-macro-builtins-helpers.cc
badumbatish: image

Looks like there is a bunch of related flags in addition to `expr_can_be_null`. This must be what the prior dev was referring to. I can adjust the comment for clearer meaning.

badumbatish:
Not ready for review yet there seems to be some problem with rebasing

badumbatish:
@CohenArthur hi Arthur, I just rebased the master branch into this branch and its ready for review.

gcc/rust/expand/rust-macro-builtins-asm.cc
CohenArthur, line 1: I think it's fine to keep these empty files, as they'll be filled out soon anyway

gcc/rust/expand/rust-macro-builtins-utility.cc
CohenArthur, line 1: I think this file is still a bit too big. I'd like to see something like:

- one file for format_args_handler, because it's going to become even bigger
- one file for the `include*` macros, so `include_str!()`, `include!()` and `include_bytes!()` if I'm not forgetting anything
- one file for the `line!()`, `file!()` and `column!()` macros. maybe this one can be called `rust-macro-builtins-location.{cc, h}`?

gcc/rust/util/bi-map.h
```diff
@@ -48,7 +48,7 @@ template <typename K, typename V> class BiMap
     return itr->second;
   }
 
-private:
+public:
```
CohenArthur, line 51: typo or is that change supposed to be here?

CohenArthur (commented):
thanks for working on this :) I think only a couple nitpicks and then we can merge it!

gcc/rust/util/bi-map.h
badumbatish: definitely a typo, I added an insert() method when we were trying the splitting of the BiMap, i'll fix it soon

gcc/rust/expand/rust-macro-builtins-utility.cc
badumbatish: on it

badumbatish:
@CohenArthur everything should be good to go

gcc/rust/expand/rust-macro-builtins-format-args.h
CohenArthur, line 1: I'm noticing that a lot of the headers do not contain any declarations - in that case, could you remove them? and only leave the headers which have declarations?

CohenArthur (approved):
last nitpick and we'll merge this (sorry :D)

gcc/rust/expand/rust-macro-builtins-format-args.h
badumbatish: Sounds good to me, I created them just in case we need to declare some helper functions to use

PR #2880, opened 2024-02-27
Cleanup creation of lang items

CohenArthur:
- lang-items: Cleanup parsing and lookups of lang items.
- lang-items: Make lang items enum stronger, rename class, cleanup ns.

This cleans up the handling and creation of lang items. The second commit is more of a nitpick, which renames the class and enum to make it easier to use - but we can definitely drop it if it's too many changes at once.

- [ ] Need an implementation of `std::less/std::hash` for `LangItem::Kind` for the C++almost11 standard


CohenArthur:
we could also rename the methods in `LangItem` to be lowercase instead, like the rest of the compiler? or instead rename the rest of our static methods to use this case as well since I do think it makes sense, even though I'm not a huge fan

gcc/rust/hir/rust-ast-lower-base.cc
```diff
@@ -788,14 +788,13 @@ ASTLoweringBase::handle_lang_item_attribute (const ItemWrapper &item,
...
-  if (lang_item_type == Analysis::RustLangItem::ItemType::UNKNOWN)
-    {
-      rust_error_at (attr.get_locus (), "unknown lang item");
-      return;
-    }
-  mappings->insert_lang_item (lang_item_type,
-			      item.get_mappings ().get_defid ());
+  auto lang_item_type = LangItem::Parse (lang_item_type_str);
+
+  if (lang_item_type)
```
P-E-P, line 793: I'd prefer we keep the condition explicit instead of relying on the underlying value.

P-E-P (approved):

gcc/rust/hir/rust-ast-lower-base.cc
P-E-P: No sorry, I ready too fast and thought we kept the Unknown item kind. Given it is an optional we can keep this.

PR #2887, opened 2024-02-28
Ensure ```TupleStructPattern``` and ```TuplePattern``` have items

powerboat9:


CohenArthur (commented):
I don't know if this is actually correct, unless I'm misunderstanding? it seems to me that empty patterns are allowed? https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=755341695f14b8bd2d51f095778ec222

powerboat9:
This just ensures that an empty pattern is represented with a non-null ```TupleStructItems```.

CohenArthur (approved):

CohenArthur:
> This just ensures that an empty pattern is represented with a non-null `TupleStructItems`.

thanks for the explanation :)

PR #2892, opened 2024-02-29
Implemented method translation to BIR.

### LINKED ISSUE #2888: borrowck: Implement method translation to BIR
See: https://github.com:Rust-GCC/gccrs/blob/borrowck-stage2-cleaned/gcc/rust/checks/errors/borrowck/rust-bir-builder-expr-stmt.cc#L263-L263

And implement: https://github.com:Rust-GCC/gccrs/blob/borrowck-stage2-cleaned/gcc/rust/checks/errors/borrowck/rust-bir-builder-expr-stmt.cc#L283-L283

braw-lee:
Fixes #2888 
Depends on #2889

### DIFF
--- a/gcc/rust/checks/errors/borrowck/rust-bir-builder-expr-stmt.cc
+++ b/gcc/rust/checks/errors/borrowck/rust-bir-builder-expr-stmt.cc
@@ -21,6 +21,8 @@
 #include "rust-bir-builder-lazyboolexpr.h"
 #include "rust-bir-builder-pattern.h"
 #include "rust-bir-builder-struct.h"
+#include "rust-hir-path-probe.h"
+#include "rust-type-util.h"
 
 namespace Rust {
 namespace BIR {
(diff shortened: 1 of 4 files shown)


gcc/rust/checks/errors/borrowck/rust-bir-builder-expr-stmt.cc
```diff
@@ -281,7 +281,46 @@ ExprStmtBuilder::visit (HIR::CallExpr &expr)
...
+  PlaceId fn = visit_expr (expr.get_method_name ());
+
+  std::vector<PlaceId> arguments = visit_list (expr.get_arguments ());
+
+  const auto fn_type = ctx.place_db[fn].tyty->as<const TyTy::FnType> ();
+
+  // compile `self` and handle autoref
+  PlaceId self = visit_expr (*expr.get_receiver ());
+  auto ty = ctx.place_db[self].tyty;
+  if (ty->get_kind () == TyTy::REF)
```
jdupak, line 295: This seems to be inverted. Autoref needs to make reference to values that are not references.

jdupak: Autoderef may be needed here. What you want to achieve is to have exactly one reference (in terms of levels).

gcc/rust/checks/errors/borrowck/rust-bir-builder-expr-stmt.cc
```diff
@@ -281,7 +281,46 @@ ExprStmtBuilder::visit (HIR::CallExpr &expr)
 
 void
 ExprStmtBuilder::visit (HIR::MethodCallExpr &expr)
-{}
+{
+  // TODO : need ExprStmtBuilder::visit_expr (HIR::PathExprSegment)
+  PlaceId fn = visit_expr (expr.get_method_name ());
```
jdupak, line 286: The trick for this is to look into the gccrs backend and check how Philip is already doing it. That you need here is at: gcc\rust\backend\rust-compile-expr.cc:1281

Note that the back-end needs to handle way more then we need to do here. He only care about the flow of data and its types.

jdupak:
TLDR: @braw-lee I cannot push to your branch, so this is my suggested change: https://github.com/jdupak/gccrs/tree/borrowck-wip

jdupak:
If you have time, I suggest adding test for some convoluted testcases.

philberty:
This needs rebased and is almost a year old please feel free to reopen at a later point

PR #2903, opened 2024-03-05
Desugar AST::ForLoopExprs as part of HIR lowering

CohenArthur:


CohenArthur:
This is causing a lot of issues and they are really annoying to fix... we can probably experiment with definning `println!()` with an iterator rather than a for-loop

powerboat9:
Still applicable?

CohenArthur:
> Still applicable?

Yeah, I think I'll use that branch as the basis for the remaining `for-loop` work which depends on #3290 being merged. So hopefully we can close this out relatively soon

PR #2904, opened 2024-03-05
Check derive compatibility

### LINKED ISSUE #2154: Report invalid uses of `#[derive]`
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=393cdbd04e6e32648c23d1816ab1c056

https://doc.rust-lang.org/stable/error_codes/E0774.html

`#[derive]` attributes should only be allowed on structs, enums and unions.

GuillaumeGomez:
Fixes #2154.

Second commit is a small typo I encountered in the function I was working on.


GuillaumeGomez:
So, seems like I'm stuck: I think the attribute visitor is overloading the methods wrongly, making them not go through the code recursively. If so, the simplest way would be for the original ASTVisitor to have two sets of methods:
1. the current one which can be overloaded
2. a new set which would call each visit method as it should and that could not be overloaded

Again, maybe I'm completely misunderstanding how it works and sorry for the noise.

P-E-P:
> So, seems like I'm stuck: I think the attribute visitor is overloading the methods wrongly, making them not go through the code recursively. If so, the simplest way would be for the original ASTVisitor to have two sets of methods:
> 
>     1. the current one which can be overloaded
> 
>     2. a new set which would call each visit method as it should and that could not be overloaded

Not entirely sure I understand what you mean. Right now we use `ASTVisitor` to provide the visitor interface, as well as `DefaultASTVisitor` to automatically recurse into the nodes. The latter can be overloaded on desired nodes to either stop the recursion, change the behavior or process whatever you want. Would this description fix your problem ?

> Again, maybe I'm completely misunderstanding how it works and sorry for the noise.

don't worry, I'm glad we've got people challenging the implementation

GuillaumeGomez:
If it sounds good to you, I can implement it.

PR #2911, opened 2024-03-09
Store visibility properly in ExternalTypeItem: Fixes #2897

### LINKED ISSUE #2897: Allow visibility modifiers for extern types
Currently we do not lower/handle the visibility modifiers properly for `ExternalTypeItem` nodes.

badumbatish:
Ran through git-clang-format, make check-rust passes as expected. 

Check-list:

(x) Make sure that visibility modifier are parsed for extern type nodes: I made sure that `parse_external_type_item` has AST::Visibility within its declaration
```c++
// in rust-parse-impl.h
std::unique_ptr<AST::ExternalTypeItem>
Parser<ManagedTokenSource>::parse_external_type_item (AST::Visibility vis,
						      AST::AttrVec outer_attrs);
```

(x) Make sure that visibility is assigned to `ExternalTypeItem` I made sure that in in the function mentioned above, its constructor, ExternalTypeItem(), handled visibility correctly via its parent's constructor
```c++
// In rust-item.h
ExternalTypeItem (Identifier item_name, Visibility vis,
		  std::vector<Attribute> outer_attrs, location_t locus)
  : ExternalItem (), outer_attrs (std::move (outer_attrs)), visibility (vis),
    item_name (std::move (item_name)), locus (locus), marked_for_strip (false)
{}

```

(x) Make sure that the AST visibility is correctly lowered to an HIR visibility I added a translate_visibility call via
`HIR::Visibility vis = translate_visibility (type.get_visibility ());` in the visit() function.

I also changed the constructor of HIR::ExternalTypeItem to accept a visibility and change the way it delegate visibility to its parent's constructor.

gcc/rust/ChangeLog:

	* hir/rust-ast-lower-extern.h: Add translate_visiblity
	* hir/tree/rust-hir-item.h: Fix constructor of ExternalTypeItem

Closes #2897

### DIFF
--- a/gcc/rust/hir/rust-ast-lower-extern.h
+++ b/gcc/rust/hir/rust-ast-lower-extern.h
@@ -133,8 +133,10 @@ public:
 				   mappings->get_next_hir_id (crate_num),
 				   mappings->get_next_localdef_id (crate_num));
 
+    HIR::Visibility vis = translate_visibility (type.get_visibility ());
+
     translated = new HIR::ExternalTypeItem (mapping, type.get_identifier (),
-					    type.get_locus ());
+					    vis, type.get_locus ());
   }
 
 private:
--- a/gcc/rust/hir/tree/rust-hir-item.h
+++ b/gcc/rust/hir/tree/rust-hir-item.h
@@ -3154,9 +3154,9 @@ class ExternalTypeItem : public ExternalItem
 {
 public:
   ExternalTypeItem (Analysis::NodeMapping mappings, Identifier item_name,
-		    location_t locus)
+		    Visibility vis, location_t locus)
     : ExternalItem (std::move (mappings), std::move (item_name),
-		    Visibility (Visibility::PRIVATE),
+		    Visibility (std::move (vis)),
 		    /* FIXME: Is that correct? */
 		    {}, locus)
   {}


dkm:
Personal taste, but not my call. I would avoid using github issues as the main part of the commit title. It's harder to guess what is the commit about when doing "git log", for example. Something like "fix visibility for XXX" or "change visibility for XXX".

Or, feel free to ignore if @CohenArthur feels this is too nitpicky :)

badumbatish:
> Personal taste, but not my call. I would avoid using github issues as the main part of the commit title. It's harder to guess what is the commit about when doing "git log", for example. Something like "fix visibility for XXX" or "change visibility for XXX".
> 
> Or, feel free to ignore if @CohenArthur feels this is too nitpicky :)

good call, i think i'll change it tomorrow on my pc when i got a chance, tysm!

dkm:
Thank you for all your contributions :+1: ! I'm merely trying to help with the "packaging"

CohenArthur (approved):
this looks great! thanks @badumbatish. what else is missing for this to close issue #2897? I notice you've said "initial steps" but it seems to me this is complete?

CohenArthur:
also regarding the commit - I agree with @dkm that this can be confusing because the hash/pound sign is also used for other bug trackers etc. If you want to mention an issue in a commit I'd go for something like "issue 2897 on github".

For this specific PR I think we can simplify the commit a little bit so that it makes the life of people who read Changelogs easier, but first I want to make sure we got all the pieces for closing out #2897

CohenArthur:
ah, I think I get it - is this about me saying we should ensure visibility works properly for external types in a later PR? If that is the case, then this PR is enough to close the issue and we should change its title + the commit to something more descriptive like "Store visibility properly in ExternalTypeItem"

Then in the PR's body it would be great if you could add "Fixes #2897" so that the issue gets automatically closed when we merge this.

For the later bits (checking that the visibility constraints are actually respected) we'll see much later, this isn't really important for now :) good work!

badumbatish:
> this looks great! thanks @badumbatish. what else is missing for this to close issue #2897? I notice you've said "initial steps" but it seems to me this is complete?

yep this is complete haha I just never changed the title, I think the PR is pretty complete as well.

badumbatish:
Let me actually fix the commit title, I forgot to do that

badumbatish:
@CohenArthur commit is fixed

badumbatish:
> also regarding the commit - I agree with @dkm that this can be confusing because the hash/pound sign is also used for other bug trackers etc. If you want to mention an issue in a commit I'd go for something like "issue 2897 on github".
> 
> For this specific PR I think we can simplify the commit a little bit so that it makes the life of people who read Changelogs easier, but first I want to make sure we got all the pieces for closing out #2897

i realized I missed this point, I deleted all check-list and only leave the change log now.

CohenArthur:
thanks! :D good work!

PR #2917, opened 2024-03-18
ci: Install Rust directly from rustup instead of apt

CohenArthur:
ChangeLog:

	* .github/workflows/ccpp.yml: Install Rust manually.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -37,7 +37,9 @@ jobs:
                   build-essential \
                   gcc-multilib \
                   g++-multilib \
-                  dejagnu
+                  dejagnu;
+          # install Rust directly using rustup
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain=1.72.0;
 
     - name: Configure
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ccpp.yml
```diff
@@ -37,7 +37,9 @@ jobs:
                   build-essential \
                   gcc-multilib \
                   g++-multilib \
-                  dejagnu
+                  dejagnu;
+          # install Rust directly using rustup
+          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y;
```
P-E-P, line 42: I think we do want to pin the versions of rustup and rustc. `sh.rustup.rs` does not provides a way to fetch an older version of rustup but the resulting script provides a way to select a given toolchain version, I guess that's better than nothing.

```suggestion
          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain=1.49.0;
```

P-E-P (approved):

.github/workflows/ccpp.yml
P-E-P: Here is the url for a pinned rustup version:

```
https://static.rust-lang.org/rustup/archive/1.27.0/x86_64-unknown-linux-gnu/rustup-init
```

P-E-P (approved):

PR #2918, opened 2024-03-19
macro: Use MacroInvocation's node_id in ExternalItem constructor.

CohenArthur:
gcc/rust/ChangeLog:

	* ast/rust-macro.h: Use proper node id instead of the one in the base
	Expr class - which is uninitialized.

I think the `MacroInvocation` class would benefit from a cleanup on that part. It inherits from a ton of base classes, and its full constructor does not even call into the `ExternalItem` constructor. I'll open an issue.


CohenArthur:
Opened #2919

PR #2928, opened 2024-03-26
Avoid parsing const unsafe/extern functions as async

powerboat9:


P-E-P:
Would'nt this mean async cannot be extern nor unsafe ? Don't forget we need to validate some part of the ast later.
https://godbolt.org/z/fvz6TosaG

powerboat9:
https://doc.rust-lang.org/reference/items/functions.html
https://godbolt.org/z/eeKr7c889

```async``` must come before ```unsafe``` or ```extern```, so if we parse ```const unsafe``` or ```const extern``` we know that the current function can't be async. Splitting qualifier parsing off into a separate function could be done, but this seems like a simple and robust fix for the issue at hand.

P-E-P:
> https://doc.rust-lang.org/reference/items/functions.html https://godbolt.org/z/eeKr7c889
> 
> `async` must come before `unsafe` or `extern`, so if we parse `const unsafe` or `const extern` we know that the current function can't be async.

Oh yes, you're right. Unsafe and extern shall be handled within parse_async to be correct, good call.

> Splitting qualifier parsing off into a separate function could be done
Now I remember another PR where this was discussed. IIRC we postponed this because function qualifier order detection came after 1.49, the compiler simply emitted a parsing error before that without mentioning the required order.

> but this seems like a simple and robust fix for the issue at hand.
Indeed it is!

P-E-P (approved):
LGTM, good call as usual !

PR #2929, opened 2024-03-26
Recognize unstable as a builtin attribute

powerboat9:
In the future we'll have to produce [E0658](https://doc.rust-lang.org/error_codes/E0658.html), but this should allow us to compile code containing the ```unsafe``` attribute.


CohenArthur:
> In the future we'll have to produce [E0658](https://doc.rust-lang.org/error_codes/E0658.html), but this should allow us to compile code containing the `unsafe` attribute.

the `unstable` attribute? or am I confusing this with something else?

powerboat9:
Fixed commit message

PR #2930, opened 2024-03-27
Improve parsing of raw byte string literals

powerboat9:
Ideally we would combine the string literal parsing functions somehow or use some abstraction to more easily track locations, but this should improve our current situation.


CohenArthur (approved):
thanks, these are great :)

PR #2936, opened 2024-03-29
raw-strings: Remove dg-excess-error directive

CohenArthur:
The error is actually expected and uses the correct location.

gcc/testsuite/ChangeLog:

	* rust/compile/raw-byte-string-loc.rs: Use dg-error instead of
	dg-excess-error.
	* rust/compile/raw-string-loc.rs: Likewise.

@powerboat9 how do you feel about this?


P-E-P (approved):

powerboat9:
Not sure what was going on there, but if it passes the tests now, it should be fine

PR #2938, opened 2024-04-03
Check for cargo during libgrust configuration

### LINKED ISSUE #2881: Add checking that `cargo` exists when building gccrs
Add checking for the existence of `cargo` when building gccrs, erroring out nice and early if it is not present - this will improve the user experience now that we rely on Rust components.

P-E-P:
`libgrust` configuration script did not check for cargo, which is now required in order to build `libformat_parser`.

Fixes #2881


CohenArthur (approved):

tschwinge:
The should be done in the top-level `configure.ac`, similar to "Disable Ada if no preexisting GNAT is available", or "Disable D if no preexisting GDC is available".

P-E-P:
> The should be done in the top-level `configure.ac`, similar to "Disable Ada if no preexisting GNAT is available", or "Disable D if no preexisting GDC is available".

@tschwinge AFAIK we can't modify top level configure since common directories are already frozen ?

P-E-P:
Close in favor of #2942

PR #2940, opened 2024-04-04
Allow `use` statements to happen before the items they reference

CohenArthur:
- nr2.0: default-visitor: Conditionally visit type in self parameters.
- [fixup wip] notes: Add them for how to resolve this
- toplevel: Build list of imports for Early to resolve
- early: Resolve imports and create import mappings
- imports: Add FinalizeImports class

This is not ready to review. There is currently a regression compared to master which I'm investigating:

```rust
pub mod module {
    pub fn f() {}

    pub fn returns_i32() -> i32 {
        15
    }

    pub fn bar() -> bool {
        true
    }
}

use module::bar;
use module::f;
use module::returns_i32;

fn main() {
    let foo = f();
    let bar = f();

    let a: bool = f();
    let b: bool = returns_i32();

    let c = module::bar();
}
```

The above code should compile, and does when using `-frust-name-resolution-2.0` on master, but not here. This is probably due to `FinalizeImports` not being implemented properly.

This also needs testing and cleaning up


CohenArthur:
we're now storing the namespaces and inserting definitions properly - the issue is inserting them in the right Rib. `toplevel.insert_or_error_out(...)` inserts them in the topmost Rib, which is not correct. we can store the Rib in which to insert data in `ImportKind` - maybe via a reference/pointer? we need to ensure these pointers will stay live for the duration of `FinalizeImports`, which I think is a fair assumption

CohenArthur:
but we can't know which namespace to get the Rib from before `Early` so we're hitting the same problem :( maybe FinalizeImports should be a fully fledged visitor as well? urgh

CohenArthur:
Storing all ribs gets messy and unwieldy. I guess having a `map<UseTree, ImportKind>` would make more sense, and we could just visit all `UseTree`s in `FinalizeImports` and get the import data from this

CohenArthur:
last step is to declare imports in each namespace where they resolved properly. this is highlighted by name_resolution16.rs in the testsuite.

the current implementation bails as soon as it finds one namespace an import was resolved in, but we actually need to build a list of namespaces + node Ids for each import

P-E-P:
I will kindly close this PR because it will be replaced soon by another one :smiley: Thank you for your work, it will  be merged as part of another PR soon :smiley:

Issue #2899, opened 2024-03-01
Pass configuration flags to `cargo`


tschwinge:
As of #2822 "Start reusing rustc's format_args parser", we invoke `cargo` to produce object files to be linked into GCC.  We currently don't specify any compilation flags and linker to use, so it's easy to run into mismatches between what GCC is `configure`d for (for example, specific linker), and what `cargo` uses by default (for example, system `cc` for linking).

For example, the linker may be specified: <https://doc.rust-lang.org/cargo/reference/config.html#targettriplelinker> -- but do we need the Rust target triple for that (see  #2898)?  Trying to use a `cfg()` expressions à la `--config "target.'cfg(all())'.linker=[...]"` only produces:
```
warning: unused key `linker` in [target] config table `cfg(all())`
```
(Maybe I've misunderstood how that is to be used.)

tschwinge:
Next problem: the GCC `$(LINKER)` defaults to `$(CXX)`, and doesn't just point to an executable, but may have "wrapper" executables prepended (`ccache`, for example), and also contain flags.

For a simple `--config target.x86_64-unknown-linux-gnu.linker=\"'$(LINKER)'\"`, and `LINKER = $(CXX)`, and `CXX = ccache g++ [flags]`, we then get:
```
error: linker `[...]/build-gcc/gcc/ccache g++ [flags] ` not found
```

powerboat9:
It looks like [```target.<triple>.rustflags```](https://doc.rust-lang.org/cargo/reference/config.html#targettriplerustflags) and [```link-args```](https://doc.rust-lang.org/rustc/codegen-options/index.html#link-args) could be used to pass arguments to the linker

Issue #2898, opened 2024-03-01
`cargo` should build for the _host_


tschwinge:
As of #2822 "Start reusing rustc's format_args parser", we invoke `cargo` to produce object files to be linked into GCC.  I've not yet run into the following -- but I suspect that `cargo` by default builds for the system it's currently running on: the GCC _build_ system.  However, these object files are to be linked into the compiler we're building, the GCC _host_ system.

Therefore, I suspect, we should pass:

```
      --target <TRIPLE>         Build for the target triple
```

In the `Makefile`, we do have the GCC _host_ system available in `$(host)` (or is `$(host_noncanonical)` applicable here?) -- but how to programmatically translate that into the corresponding Rust target triple, <https://doc.rust-lang.org/cargo/appendix/glossary.html#target>, <https://doc.rust-lang.org/nightly/rustc/platform-support.html>?

PR #2942, opened 2024-04-05
Check for cargo when building rust language

P-E-P:
Prevent rust language from building when cargo is missing.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1411,6 +1411,7 @@ fi
 
 ACX_PROG_GNAT
 ACX_PROG_GDC
+ACX_PROG_CARGO
 ACX_PROG_CMP_IGNORE_INITIAL
 
 AC_ARG_ENABLE([bootstrap],
@@ -2346,6 +2347,23 @@ directories, to avoid imposing the performance cost of
 	        esac
           ;;
         esac
+
+        # Disable Rust if cargo is unavailable.
+        case ${add_this_lang}:${language}:${have_cargo} in
+          yes:rust:no)
+            # Specifically requested language; tell them.
+            AC_MSG_ERROR([cargo is required to build $language])
+            ;;
+          all:rust:no)
+            AC_MSG_WARN([cargo is required to build $language])
+            add_this_lang=unsupported
+            ;;
+          *:rust:no)
+            # Silently disable.
+            add_this_lang=unsupported
+            ;;
+        esac
+
 
         # Disable a language that is unsupported by the target.
 	case "${add_this_lang}: $unsupported_languages " in
--- a/gcc/doc/install.texi
+++ b/gcc/doc/install.texi
@@ -313,6 +313,13 @@ documentation including the target @code{SYSTEM} definition module.
 If Python3 is unavailable Modula-2 documentation will include a target
 independent version of the SYSTEM modules.
 
+@item @anchor{GCCRS-prerequisite}Rust
+
+In order to build GCCRS you need a working rust compiler, for now we
+are using cargo and rustc but in the future GCC will be able to compile
+it's own rust components and recompile itself with those components
+included.
+
 @item A ``working'' POSIX compatible shell, or GNU bash
 
 Necessary when running @command{configure} because some
@@ -3225,6 +3232,15 @@ and network filesystems.
 @end ifnothtml
 @ifhtml
 @uref{prerequisites.html#GDC-prerequisite,,GDC prerequisites}.
+@end ifhtml
+
+@section Building the Rust compiler
+
+@ifnothtml
+@ref{GCCRS-prerequisite}.
+@end ifnothtml
+@ifhtml
+@uref{prerequisites.html#GCCRS-prerequisite,,GCCRS prerequisites}.
 @end ifhtml
 
 @section Building with profile feedback
--- a/config/acx.m4
+++ b/config/acx.m4
@@ -424,6 +424,17 @@ else
 fi
 ])
 
+# Test for Rust
+# We require cargo and rustc for some parts of the rust compiler.
+AC_DEFUN([ACX_PROG_CARGO],
+[AC_REQUIRE([AC_CHECK_TOOL_PREFIX])
+AC_CHECK_TOOL(CARGO, cargo, no)
+if test "x$CARGO" != xno; then
+  have_cargo=yes
+else
+  have_cargo=no
+fi])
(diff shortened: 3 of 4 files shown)


configure.ac
```diff
@@ -2305,6 +2306,23 @@ directories, to avoid imposing the performance cost of
           ;;
         esac
 
+        # Disable Rust if cargo is unavailable.
+        case ${add_this_lang}:${language}:${have_cargo} in
+          yes:rust:no)
+            # Specifically requested language; tell them.
+            AC_MSG_ERROR([Cargo is required to build $language])
```
CohenArthur, line 2313:
```suggestion
            AC_MSG_ERROR([cargo is required to build $language])
```
nit

configure.ac
```diff
@@ -2305,6 +2306,23 @@ directories, to avoid imposing the performance cost of
...
+            ;;
+          all:rust:no)
+            AC_MSG_WARN([Cargo is required to build $language])
```
CohenArthur, line 2316:
```suggestion
            AC_MSG_WARN([cargo is required to build $language])
```
same here

CohenArthur (approved):

tschwinge:
Is it expected that _GCC Rust build and test / build-and-check-gcc-48_ failed here: `configure: error: Cargo is required to build rust` -- per the _Install Deps_ step, Rust/Cargo is being installed?  Is this hopefully just some kind of scripting issue?

tschwinge:
Per a quick look, these changes look like (the first part of) how I think they should look like, thanks!  In case there are requests for changes, I suggest to first get this reviewed on <gcc-patches@gcc.gnu.org> and committed upstream, before pushing to GCC/Rust (here).

tschwinge:
And then ("the next part") we should add a check that the object files that Cargo produces for a simple (dummy) crate can actually be linked into the GCC compiler we're about to build.  This is to detect early cases like described in #2898 "`cargo` should build for the host system", or #2899 "Pass configuration flags to `cargo`", and still be able to react (disable Rust language, or sensible error message, as applicable), instead of later running into a build error.

P-E-P:
> And then ("the next part") we should add a check that the object files that Cargo produces for a simple (dummy) crate can actually be linked into the GCC compiler we're about to build. This is to detect early cases like described in #2898 "`cargo` should build for the host system", or #2899 "Pass configuration flags to `cargo`", and still be able to react (disable Rust language, or sensible error message, as applicable), instead of later running into a build error.

In fact, I've been hesitating for a long time to also check which edition of rust can be compiled but @CohenArthur convinced me to keep it simple.

configure.ac
P-E-P: Rebased and fixed.

configure.ac
P-E-P: Rebased and fixed.

CohenArthur:
> Per a quick look, these changes look like (the first part of) how I think they should look like, thanks! In case there are requests for changes, I suggest to first get this reviewed on [gcc-patches@gcc.gnu.org](mailto:gcc-patches@gcc.gnu.org) and committed upstream, before pushing to GCC/Rust (here).

patch has been reviewed and approved by Richard Biener, so I think we can merge it here as well. @P-E-P do you want me to take care of the CI issue?

P-E-P:
> do you want me to take care of the CI issue?

@CohenArthur Yes please. I won't be able to work on it for a while.

tschwinge:
So, are you going to `git push` the patch into GCC upstream?

CohenArthur:
@tschwinge I was thinking of pushing it as part of our next upstream? would you rather it was pushed beforehand?

tschwinge:
I prefer these generic non-trivial pieces to be integrated incrementally in GCC upstream: one after the other (in a logically meaningful order) instead of in bulk, so that there's some time for testing/stabilization between the individual parts.

The commit log then be updated to

CohenArthur:
alright :) that commit is good to upstream then I think. I'll fix the CI so we can merge it here

P-E-P:
@tschwinge I'm a bit lost, should we merge this ? IIRC we already sent this upstream, so we should technically close this PR right ?

CohenArthur:
> @tschwinge I'm a bit lost, should we merge this ? IIRC we already sent this upstream, so we should technically close this PR right ?

I think yes, we'll get the content from upstream when we update our fork. or we can also merge it if the commits match exactly so that it gets skipped once we rebase on gcc's trunk. @P-E-P can you check if there are any differences between what was merged on trunk and this?

P-E-P:
> > @tschwinge I'm a bit lost, should we merge this ? IIRC we already sent this upstream, so we should technically close this PR right ?
> 
> I think yes, we'll get the content from upstream when we update our fork. or we can also merge it if the commits match exactly so that it gets skipped once we rebase on gcc's trunk. @P-E-P can you check if there are any differences between what was merged on trunk and this?

There are some differences, this explains my confusion. This PR adds some documentation and modifies the CI, upstream content only has the cargo checking in the autoconf configuration file.

CohenArthur:
@P-E-P then we should merge this too, but make sure that the commit which touches the build system is exactly the same here as it was pushed on trunk

P-E-P:
I'll wait for #2979 to be merged first.

gcc/doc/install.texi
```diff
@@ -313,6 +313,13 @@ documentation including the target @code{SYSTEM} definition module.
 If Python3 is unavailable Modula-2 documentation will include a target
 independent version of the SYSTEM modules.
 
+@item @anchor{GCCRS-prerequisite}Rust
+
+In order to build GCCRS you need a working rust compiler, for now we
+are using cargo and rustc but in the future GCC will be able to compile
+it's own rust components and recompile itself with those components
```
thesamesam, line 320:
```suggestion
its own rust components and recompile itself with those components
```

gcc/doc/install.texi
```diff
@@ -313,6 +313,13 @@ documentation including the target @code{SYSTEM} definition module.
 If Python3 is unavailable Modula-2 documentation will include a target
 independent version of the SYSTEM modules.
 
+@item @anchor{GCCRS-prerequisite}Rust
+
+In order to build GCCRS you need a working rust compiler, for now we
```
thesamesam, line 318:
```suggestion
In order to build GCCRS, you need a working Rust compiler. For now, we
```

PR #2943, opened 2024-04-09
libgrust: Vendor Rust dependencies

CohenArthur:
This commits vendors and locks our dependencies so that our Rust components
do not need internet access to be built anymore. Tested in a docker container
with no network access.

We can think about merging the vendor directory once we have more than one
Rust component, e.g. when we add polonius to the mix.

Fixes #2882

PR #2945, opened 2024-04-10
libgrust: Vendor Rust dependencies: Fix configuration file

tschwinge:
If I run `cargo vendor`, it tells me the configuration file is `.cargo/config.toml`, not `.cargo/config`.  And indeed, with `rm -rf ~/.cargo`, the build then fails:

    [...]
    cd [...]/gcc/../libgrust/libformat_parser && cargo build --offline  # FIXME: Not always release, right?
    error: no matching package named `libc` found
    location searched: registry `crates-io`
    required by package `libformat_parser v0.1.0 ([...]/libgrust/libformat_parser)`
    As a reminder, you're using offline mode (--offline) which can sometimes cause surprising resolution failures, if this error is too confusing you may wish to retry without the offline flag.
    make[2]: *** [[...]/gcc/rust/Make-lang.in:426: rust/libformat_parser.a] Error 101
    [...]

Fixup for #2943 "libgrust: Vendor Rust dependencies".


tschwinge:
<https://doc.rust-lang.org/cargo/reference/config.html>:

> Note: Cargo also reads config files without the `.toml` extension, such as `.cargo/config`. Support for the `.toml` extension was added in version 1.39 and is the preferred form. If both files exist, Cargo will use the file without the extension.

..., and indeed I'm now no longer able to reproduce the issue from last night...  (That is, "all is good".)

tschwinge:
I'm therefore closing this issue -- as I understand, `.cargo/config` is compatible with Cargo pre-1.39 (2019-11-07), if that's useful still.

PR #2949, opened 2024-04-12
Do not link -ldl and -lpthread unconditionally

CohenArthur:


tschwinge:
Yes, that's going in the right direction.  :+1:

At least partially, this again (like #2942 "Check for cargo when building rust language") needs to go into the top-level `configure`, in order to be able to gracefully disable GCC/Rust for `--enable-languages=all` configurations.

CohenArthur:
> Yes, that's going in the right direction. 👍
> 
> At least partially, this again (like #2942 "Check for cargo when building rust language") needs to go into the top-level `configure`, in order to be able to gracefully disable GCC/Rust for `--enable-languages=all` configurations.

yeah, I saw that Jakub had mentioned this but it does seem a bit too late to make such a change, no? Jakub had also mentioned erroring out, which is what I chose to do here

tschwinge:
It's late no matter whether in the top-level `configure`, or `gcc/configure`.  ;-|

tschwinge:
Gracefully disable for implicit "rust" in `--enable-languages=all`, error out for explicit `--enable-languages=rust`.

tschwinge:
Only `crab1` should then be linked against these libraries, not all GCC things generally.

CohenArthur:
> Only `crab1` should then be linked against these libraries, not all GCC things generally.

right, so add `-ldl -lpthread` to something like `CRAB1_LIBS` instead? I think we have `RUST_LIBS` for this already, I'll check.

> Gracefully disable for implicit "rust" in --enable-languages=all, error out for explicit --enable-languages=rust.

so you're forcing me to do more autotools work, do you like seeing me suffer or something? :P will work on this next

tschwinge:
The implicit vs. explicit thing should be really easy ;-) in the top-level `configure`: just like how #2942 "Check for cargo when building rust language" does this, and similarly for GCC/D if `gdc` or GCC/Ada if `gnat` is not available in a suitable version.

CohenArthur:
@tschwinge I've made the required changes and pushed them :) tested locally and it seems to work fine on a system where glibc < 2.34. tested with `--enable-languages=all` and `--enable-languages=rust`

P-E-P (approved):

P-E-P:
Should we merge this ? Has it already been merged upstream ?

CohenArthur:
I'm going to merge this PR because it has been merged upstream, sadly it introduce a regression because I'm too stupid for autotools and shell scripting in general, so I'll open a follow-up PR shortly

CohenArthur:
turns out that the bug/issue does not actually prevent the build from completing, it's just noisy but the build completes. I'll still push the fix ASAP

PR #2957, opened 2024-04-19
Add mini standard library.

P-E-P:
Add a new mini standard library with standard type. This is intended to change and be later replaced with the real standard library.

### DIFF
--- a/libgrust/libministd/lib.rs
+++ b/libgrust/libministd/lib.rs
@@ -1,0 +1,1100 @@
+//! # Mini standard library
+//!
+//! This library is intended to replace the standard library during the
+//! experimental phase of GCC's rust frontend.
+//! When GCC will be able to compile the whole standard library this will
+//! be deprecated and deleted.
+//!
+//! The overall goal of this library is to bring as many features as
+//! possible from the rust standard library whilst staying compilable using
+//! GCC.
+//!
+//! The API should be kept as close as possible to the standard library but
+//! some things may slightly change to accomodate the capabilities of GCC.
+//! The API may break without any warning, this library is only intended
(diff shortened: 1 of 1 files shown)


P-E-P:
Compile time is awful and it lacks a bunch of things.

P-E-P:
Also, we can't even make modules because there is a bug during typechecking with modules. Will open an issue later with a minimized example.

dkm:
IIUC, this is mostly some copy/paste + adjustment from `rustc`'s `libstd`? We should probably have matching copyright/license info ?

P-E-P:
> IIUC, this is mostly some copy/paste + adjustment from `rustc`'s `libstd`? We should probably have matching copyright/license info ?

I was wondering what should be the appropriate header. It has mostly been copied from rust standard library with some minor adjustments to make `crab1` happy. Probably the original license ?

CohenArthur (approved):
this looks good to me so far :) we need to discuss how this will be "distributed" within the compiler, and then I think we can merge it and keep adding to it until 14.1

dkm:
> > IIUC, this is mostly some copy/paste + adjustment from `rustc`'s `libstd`? We should probably have matching copyright/license info ?
> 
> I was wondering what should be the appropriate header. It has mostly been copied from rust standard library with some minor adjustments to make `crab1` happy. Probably the original license ?

Maybe add the first part of https://github.com/rust-lang/rust/blob/master/COPYRIGHT in the header. Not sure where to put MIT/APACHE text, but there must be similar case in GCC already

philberty:
@P-E-P  can this be added as a test case please? If some of it fails just comment it out it will be really helpful for me to work on. I've been doing this a bunch of times then starting from scratch loads.

PR #2958, opened 2024-04-21
Handle structs as scrutinee for match expressions

### LINKED ISSUE #2906: Rebinding match expressions on structs are unhandled
I tried this code:

```rust
struct Foo { a: i32 }

fn main() {
    let a = Foo { a: 15 };
    
    match a {
        b => { }
    }
}
```

this is very ugly Rust code, but it's used by rustc in a number of places for desugaring purposes, for example for loops - which get desugared as something of the form:

```rust
for <pat> in <head> <body>
```

becomes

```rust
{
    let result = match <head>.into_iter() {
        mut iter => loop { match iter.next() { ... } }
    }
}
```

I expected to see this happen: no error: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=eedc6ff5cd556e9e703c52cf601d2f4c

The issue lies in here:

```c++
  // rust-compile-expr.cc:952
  if (scrutinee_kind == TyTy::TypeKind::ADT)
    {
      // this will need to change but for now the first pass implementation,
      // lets assert this is the case
      TyTy::ADTType *adt = static_cast<TyTy::ADTType *> (scrutinee_expr_tyty);
      rust_assert (adt->is_enum ());
      rust_assert (adt->number_of_variants () > 0);
    }
```

we are asserting that the scrutinee we're matching on is an enum, which is not the case here

Instead, this happened: *explanation*

### Meta

nobel-sh:
gcc/rust/ChangeLog:

	* backend/rust-compile-expr.cc (check_match_scrutinee): Handle structs

fixes #2906

We should now be able to use struct as scrutinee. I tested the code locally and it ran without issues but I was wondering if this patch also required a test case along with it.

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -954,8 +954,8 @@ check_match_scrutinee (HIR::MatchExpr &expr, Context *ctx)
       // this will need to change but for now the first pass implementation,
       // lets assert this is the case
       TyTy::ADTType *adt = static_cast<TyTy::ADTType *> (scrutinee_expr_tyty);
-      rust_assert (adt->is_enum ());
-      rust_assert (adt->number_of_variants () > 0);
+      if (adt->is_enum ())
+	rust_assert (adt->number_of_variants () > 0);
     }
   else if (scrutinee_kind == TyTy::TypeKind::FLOAT)
     {
--- a/gcc/testsuite/rust/compile/issue-2906.rs
+++ b/gcc/testsuite/rust/compile/issue-2906.rs
@@ -1,0 +1,10 @@
+// { dg-warning "field is never read: .a." "" { target *-*-* } .-1 }
+struct Foo { a: i32 }
+
+fn main() {
+    let a = Foo { a: 15 };
+    
+    match a {
+        b => { }
+    }
+}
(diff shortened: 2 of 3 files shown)


dkm:
You're right, adding a matching test would be great. Even more if you already have one at hand! Thanks!

CohenArthur (approved):
This looks correct, thanks! I agree with @dkm that we would require a testcase in that instance, so that we can ensure we don't introduce regressions later on. I'd also like to see an execution testcase to check if we do match properly on structures (let me know if you need help with that)

nobel-sh:
Sure, I'm on it! Also, @CohenArthur, I'm not very familiar with execution test cases (are these supposed to differ from other test cases?). Any help you could offer would be greatly appreciated!

CohenArthur:
> Sure, I'm on it! Also, @CohenArthur, I'm not very familiar with execution test cases (are these supposed to differ from other test cases?). Any help you could offer would be greatly appreciated!

they differ in that they're in a different folder in the testsuite, and they will usually have an extra `dg-output` directive to make sure the printed output is correct :) but the best way is to create a main function which returns an `i32`, and to make it return `0` if the test works and something else if the test fails. check out `gcc/testsuite/rust/execute/torture/matches_macro.rs` for a good example

nobel-sh:
@CohenArthur please LMK if these test cases are enough or not.

Also, a question not related to this thread. Have we implemented immutability/mutability for variables yet? because while writing these tests I was able to mutate a variable without declaring it mutable.

P-E-P (approved):
LGTM

PR #2968, opened 2024-04-24
Removed obsolete objects

### LINKED ISSUE #2967: Obsolete objects, can be removed safely
https://github.com/Rust-GCC/gccrs/blob/611d0bdc55a858736bd5cfabfc1ed4fef800d81a/gcc/rust/backend/rust-compile-expr.cc#L1320C7-L1321C71

the objects in above 2 lines were used once, but are obsolete now and can be removed safely

braw-lee:
Fixes #2967 

Created this PR to check if my new setup was working for PRs or not,
I can delete this if you want to save this issue as `good-first-pr` for first contributions.


gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -1315,12 +1315,8 @@ CompileExpr::visit (HIR::MethodCallExpr &expr)
...
     }
   else
-    {
-      // lookup compiled functions since it may have already been compiled
-      HIR::PathExprSegment method_name = expr.get_method_name ();
-      HIR::PathIdentSegment segment_name = method_name.get_segment ();
-      fn_expr = resolve_method_address (fntype, receiver, expr.get_locus ());
-    }
+    // lookup compiled functions since it may have already been compiled
+    fn_expr = resolve_method_address (fntype, receiver, expr.get_locus ());
```
CohenArthur, line 1319: hmmm, I feel like there used to be/should be a method cache lookup here which does not exist for some reason? maybe it would be better to comment out the two objects and add a `FIXME` comment?

braw-lee: oh ok, fixme comment something like this?
```
// FIXME : method cache lookup
// comment deleted lines
// here
code...
```

CohenArthur: yes something like this would be good :) but could you investigate that piece of code? was there method lookup at some point here that was removed, or was it just never added?

braw-lee: I checked with git blame, what I found was these 2 objects were passed to the below function first, but later function didn't need these parameters, and the above 2 lines were left unchanged

braw-lee: i asked @philberty about these 2 lines, he told me to go for it
Philip, do you remember what Arthur is talking about?

CohenArthur: that's fine then, we can go ahead and remove them :D thanks for checking!

CohenArthur (approved):

PR #2969, opened 2024-04-24
Rework hir map to make it's API more confortable to work with.

P-E-P:
These are old changes I just rebased onto master. More commits are coming with some few other changes, mainly to make lookup function return and optional instead of a boolean.


P-E-P:
Given the size of this PR I should probably split it into multiple PRs.

CohenArthur (approved):
LGTM! well done :)

PR #2966, opened 2024-04-23
gccrs: Add base documentation for using the Rust frontend.

CohenArthur:
gcc/rust/ChangeLog:

	* Make-lang.in: Add documentation targets.
	* gccrs.texi: New file.

PR #2975, opened 2024-04-29
Add "Using gccrs" section

### LINKED ISSUE #2050: Add required documentation for GCC release
- [ ] Add changes to `changes.html` (https://gcc.gnu.org/gcc-13/changes.html)
- [ ] Contribute to gcc.gnu.org/wiki for some explanations on how to use the compiler
- [ ] Mention the state of the compiler in `changes.html`
- [ ] Add documentation about our compiler's internal
- More documentation needed?

CohenArthur:
Needs #2966, so only review the last commit

Fixes #2050

### DIFF
--- a/gcc/rust/gccrs.texi
+++ b/gcc/rust/gccrs.texi
@@ -85,7 +85,8 @@ language, see @uref{https//rust-lang.org}.
 * Copying::                     The GNU General Public License.
 * GNU Free Documentation License::
                                 How you can share and copy this manual.
-* Invoking gccrs::              How to run gccrs.
+* Invoking @code{gccrs}::       How to run @code{gccrs}.
+* Using @code{gccrs}::          Compiling Rust code with @code{gccrs}.
 * Index::                       Index.
 @end menu
 
@@ -94,8 +95,8 @@ language, see @uref{https//rust-lang.org}.
 
 @include fdl.texi
 
-@node Invoking gccrs
-@chapter Invoking gccrs
+@node Invoking @code{gccrs}
+@chapter Invoking @code{gccrs}
 
 @c man title gccrs A GCC-based compiler for the Rust programming language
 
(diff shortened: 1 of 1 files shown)


P-E-P (approved):
I've taken a look at the last commit and the text looks fine.

CohenArthur:
Will send that patch upstream for review before merging it, in the hopes it makes it to 14.2

P-E-P:
@CohenArthur We need to merge this before stage 3 right ?

CohenArthur:
> @CohenArthur Has this been sent (From what I recall this did not get accepted) ? Do we need to merge this before stage 3 ?

It's been sent for 14.2 but not reviewed, so we'll add it in 15.1. We don't need to merge it before stage 3, as the changes only concern the `gcc/rust/` folder. We can upstream it whenever. I do need to find why it fails though

thesamesam:
> It's been sent for 14.2 but not reviewed,

Backports don't tend to be reviewed / need review, but isn't all of this really Rust-specific anyway (so would barely need it to begin with)? That said, yes, I'd just leave it for 15.x at this point.

CohenArthur:
> > It's been sent for 14.2 but not reviewed,
> 
> Backports don't tend to be reviewed / need review, but isn't all of this really Rust-specific anyway (so would barely need it to begin with)? That said, yes, I'd just leave it for 15.x at this point. Things can only be backported if they're on trunk to begin with though.

I wasn't meant to be backported, it was sent in time for 14.2 but not reviewed. I had chatted with Jakub who told me we couldn't make "technical" changes freely for gcc/rust/ between 14.1 and 14.2, so things like this had to be reviewed before being merged. I didn't ping people for this as I assumed they had more important things to review :D it's fine if it only makes it in 15.1

thesamesam:
Ah, I see! Thanks for explaining.

powerboat9:
@CohenArthur is this still relevant? I think GCC is still in stage 1 (according to https://gcc.gnu.org/develop.html), so changes to common code should be ok.

PR #2978, opened 2024-05-03
Allow matching on non-enum ADTs

powerboat9:


powerboat9:
Couldn't find #2906 until just now, so I didn't notice that this was a duplicate of #2958

P-E-P:
> Couldn't find #2906 until just now, so I didn't notice that this duplicates much of #2958

We merged the other PR however it misses the non enum variant number test. Would you rebase this PR on master so we could integrate your test as well as the missing changes ?

powerboat9:
Rebased

PR #2979, opened 2024-05-04
Add a new CI step for windows environments

P-E-P:
Add a new CI step on Windows to ensure the project compiles on this platform with mingw.

Motivation: During upstreaming process we discovered some `dlsym` code was incorrectly disabled, the project does not compile using mingw anymore.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -495,7 +495,7 @@ jobs:
       matrix:
         include:
           - { sys: mingw64, env: x86_64 }
-          - { sys: mingw32, env: i686 }
+          # - { sys: mingw32, env: i686 }
     name: 'Windows ${{ matrix.env }} ${{ matrix.sys }}'
     defaults:
       run:
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -488,3 +488,50 @@ jobs:
             else \
               exit 0; \
             fi
+
+  build-and-check-windows:
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        include:
+          - { sys: mingw64, env: x86_64 }
+          - { sys: mingw32, env: i686 }
+    name: 'Windows ${{ matrix.env }} ${{ matrix.sys }}'
+    defaults:
+      run:
+        shell: msys2 {0}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: '${{ matrix.sys }} Install deps'
+        uses: msys2/setup-msys2@v2
+        with:
+          msystem: ${{matrix.sys}}
+          update: true
+          install: >-
+            git
+            dejagnu
+            base-devel
+          pacboy: >-
+            toolchain:p
+            rust:p
+
+      - run: |
+          find . -name .cargo-checksum.json -exec sed -i.uncheck -e 's/"files":{[^}]*}/"files":{ }/' '{}' '+'
+
+
+      - name: Configure
+        run: |
+           mkdir -p gccrs-build;
+           cd gccrs-build;
+           ../configure \
+               --enable-languages=rust \
+               --disable-bootstrap
+      - name: Build
+        run: |
+           make -C gccrs-build -j1 # $(nproc)
+
+      - name: Run Tests
+        run: |
+           cd gccrs-build; \
+           make check-rust
--- a/configure
+++ b/configure
@@ -9202,20 +9202,15 @@ $as_echo "$as_me: WARNING: GDC is required to build $language" >&2;}
         esac
 
         # Disable Rust if we are missing some required C libraries for the Rust runtime.
-        case ${add_this_lang}:${language}:${missing_rust_dynlibs} in
-          *:rust:none)
-            # Nothing to do - we're not missing any C libraries
+        case ${add_this_lang}:${language}:${missing_rust_dynlibs}:${target} in
+          *:rust:none:* | *:rust:*:*-*-mingw*)
+            # Nothing to do - we're not missing any C libraries OR we're using win32
             ;;
-          yes:rust:*)
+          yes:rust:*:* | all:rust:*:*)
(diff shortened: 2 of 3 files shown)


P-E-P:
```
error: the listed checksum of `D:\a\gccrs\gccrs\gccrs-build\libgrust\libformat_parser\..\..\..\libgrust\libformat_parser\vendor\unicode-xid\src/tables.rs` has changed:
expected: e4c2e9b70f45d262b78fde8435a969eb2478ca6d9dacd061953f54cf344f0f4d
actual:   7325938d169daf6c9052ef45fe75ce873c9f2cee6c57b0ee12ed159a7ec2a314

directory sources are not intended to be edited, if modifications are required then it is recommended that `[patch]` is used with a forked copy of the source
make[3]: *** [Makefile:431: debug/liblibformat_parser.a] Error 101
make[3]: Leaving directory '/d/a/gccrs/gccrs/gccrs-build/libgrust/libformat_parser'
make[2]: *** [Makefile:422: all-recursive] Error 1
make[2]: Leaving directory '/d/a/gccrs/gccrs/gccrs-build/libgrust'
make[1]: *** [Makefile:12394: all-libgrust] Error 2
```

.github/workflows/ccpp.yml
```diff
@@ -301,3 +301,59 @@ jobs:
...
+      matrix:
+        include:
+          - { sys: mingw64, env: x86_64 }
+          - { sys: mingw32, env: i686 }
+    name: 'Windows ${{ matrix.env }} ${{ matrix.sys }}'
+    defaults:
+      run:
+        shell: msys2 {0}
+    steps:
+      - uses: actions/checkout@v3
```
MehdiChinoune, line 316:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/ccpp.yml
```diff
@@ -301,3 +301,59 @@ jobs:
...
+
+
+      - name: Configure
+        run: |
+           mkdir -p gccrs-build;
+           cd gccrs-build;
+           ../configure \
+               --enable-languages=rust \
+               --disable-bootstrap \
+               --enable-multilib
```
MehdiChinoune, line 351: I don't think multilib is support on mingw-w64, We don't build mingw-w64-gcc with it.

.github/workflows/ccpp.yml
```diff
@@ -301,3 +301,59 @@ jobs:
...
+      - name: Build
+        run: |
+           make -C gccrs-build -j1 # $(nproc)
```
MehdiChinoune, line 354: gcc could be built in parallel on mingw-w64.

P-E-P: This is only because I don't have any windows machine, it is longer but build output are not merged, this was not meant to be merged as is.

P-E-P:
We'll need #2949

P-E-P:
@MehdiChinoune Any idea about what is going on here ?

From what I understand the standard library requires some win32 functions, I tought linking with `-lkernel32` would do the trick but I was proven wrong.

MehdiChinoune:
Looks like those functions come from ws2_32 library.

mati865:
In case you need a list of Win32 libraries that official compiler links check https://github.com/rust-lang/rust/blob/360f7d7b731c508c88b7b56921a92d281c05dbf7/compiler/rustc_target/src/spec/base/windows_gnu.rs#L25 and https://github.com/rust-lang/rust/blob/360f7d7b731c508c88b7b56921a92d281c05dbf7/library/windows_targets/src/lib.rs#L37-L41

configure.ac
```diff
@@ -2042,12 +2042,19 @@ AC_SUBST(PICFLAG)
...
+      CRAB1_LIBS="-lkernel32"
+	;;
+  *)
+    save_LIBS="$LIBS"
+    LIBS=
+    AC_SEARCH_LIBS([dlopen], [dl])
+    AC_SEARCH_LIBS([pthread_create], [pthread])
+    CRAB1_LIBS="$LIBS"
+    LIBS="$save_LIBS"
+  ;;
```
thesamesam, line 2056: The `;;` indentation is inconsistent with the block above.

thesamesam (commented):
Looks good otherwise from that nit.

configure.ac
P-E-P: I wasn't too sure how to indent that. Should the semicolon be aligned with the case ?

```
case "${target}" in
LIBS=
  *-*-mingw*)
AC_SEARCH_LIBS([dlopen], [dl])
      CRAB1_LIBS="-lkernel32 -luser32 -lmsvcrt -lmingwex -lmingw32 -lmingwex"
AC_SEARCH_LIBS([pthread_create], [pthread])
;;
CRAB1_LIBS="$LIBS"
  *)
LIBS="$save_LIBS"
    save_LIBS="$LIBS"
    LIBS=
    AC_SEARCH_LIBS([dlopen], [dl])
    AC_SEARCH_LIBS([pthread_create], [pthread])
    CRAB1_LIBS="$LIBS"
    LIBS="$save_LIBS"
;;
```

thesamesam: Honestly, either style is fine (people do a mix) but just have it be consistent. I personally have it aligned with the last statement inside the case usually (so the 1st one).

GCC seems to usually do that too (first variant).

configure
```diff
@@ -8891,7 +8896,7 @@ if test "$ac_res" != no; then :
 
 fi
 
-{ $as_echo "$as_me:${as_lineno-$LINENO}: checking for library containing pthread_create" >&5
+    { $as_echo "$as_me:${as_lineno-$LINENO}: checking for library containing pthread_create" >&5
```
powerboat9, line 8899: Was this intentional?

powerboat9: Oh, it's in the `case` statement

powerboat9: nvm, this file is autogenerated

Issue #2919, opened 2024-03-19
Cleanup `AST::MacroInvocation` class


CohenArthur:
The current implementation inherits from a lot of base classes, which is necessary, but does not always constructs all of them. Both of its constructors do not call into the same base class constructors for example. Similarly, this class has a lot of duplicate members with its base classes (`locus`, `node_id`) which we need to sort out

badumbatish:
i think i can do this

v1bh475u:
Hello, @CohenArthur . I would like to work on this issue. Can you please assign it to me?

PR #2981, opened 2024-05-08
Cleanup macro invoc

badumbatish:
MacroInvoc Refactor for #2919 
    
    Remove get_macro_node_id in favor of get_node_id in MacroInvocation
    Add ExteralItem node_id constructor to avoid extra get_node_id()
    
    gcc/rust/ChangeLog:
    
            * ast/rust-macro.h: Likewise.
            * resolve/rust-early-name-resolver.cc (EarlyNameResolver::visit): Likewise.
            * util/rust-hir-map.cc (Mappings::insert_macro_invocation): Likewise.
            (Mappings::lookup_macro_invocation): Likewise.
            (Mappings::insert_bang_proc_macro_invocation): Likewise.

### DIFF
--- a/gcc/rust/resolve/rust-early-name-resolver.cc
+++ b/gcc/rust/resolve/rust-early-name-resolver.cc
@@ -466,7 +466,7 @@ EarlyNameResolver::visit (AST::MacroInvocation &invoc)
   NodeId resolved_node = UNKNOWN_NODEID;
   NodeId source_node = UNKNOWN_NODEID;
   if (has_semicolon)
-    source_node = invoc.get_macro_node_id ();
+    source_node = invoc.get_node_id ();
   else
     source_node = invoc.get_node_id ();
   auto seg
--- a/gcc/rust/ast/rust-macro.h
+++ b/gcc/rust/ast/rust-macro.h
@@ -672,17 +672,12 @@ public:
     outer_attrs = std::move (new_attrs);
   }
 
-  NodeId get_node_id () const override final
-  {
-    return ExprWithoutBlock::get_node_id ();
-  }
+  NodeId get_node_id () const override final { return node_id; }
 
   AST::Kind get_ast_kind () const override
   {
     return AST::Kind::MACRO_INVOCATION;
   }
-
-  NodeId get_macro_node_id () const { return node_id; }
 
   MacroInvocData &get_invoc_data () { return invoc_data; }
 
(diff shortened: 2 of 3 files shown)


badumbatish:
ok i appear to have developed this from another branch other than master

gcc/rust/resolve/rust-early-name-resolver.cc
```diff
@@ -466,7 +466,7 @@ EarlyNameResolver::visit (AST::MacroInvocation &invoc)
   NodeId resolved_node = UNKNOWN_NODEID;
   NodeId source_node = UNKNOWN_NODEID;
   if (has_semicolon)
-    source_node = invoc.get_macro_node_id ();
+    source_node = invoc.get_node_id ();
   else
     source_node = invoc.get_node_id ();
```
CohenArthur, line 471: see for example here we're no longer doing anything different if the macro has a semicolon or not, so we can remove the `if` block

CohenArthur (commented):
I think this is already a good cleanup - I need to have a think about how exactly we can take it further. I'll come back to this later :) thank you!

P-E-P:
I don't get your changelog message:

```
        * ast/rust-macro.h: Likewise.
        * resolve/rust-early-name-resolver.cc (EarlyNameResolver::visit): Likewise.
        * util/rust-hir-map.cc (Mappings::insert_macro_invocation): Likewise.
        (Mappings::lookup_macro_invocation): Likewise.
        (Mappings::insert_bang_proc_macro_invocation): Likewise.
```

There is nothing to be like, how could the first statement be `likewise` ? In case you meant "Like stated in the commit message" I would suggest duplicating the message because IIRC changelog may be cut from the commit message to be put in the changelog file.

badumbatish:
> I don't get your changelog message:
> 
> ```
>         * ast/rust-macro.h: Likewise.
>         * resolve/rust-early-name-resolver.cc (EarlyNameResolver::visit): Likewise.
>         * util/rust-hir-map.cc (Mappings::insert_macro_invocation): Likewise.
>         (Mappings::lookup_macro_invocation): Likewise.
>         (Mappings::insert_bang_proc_macro_invocation): Likewise.
> ```
> 
> There is nothing to be like, how could the first statement be `likewise` ? In case you meant "Like stated in the commit message" I would suggest duplicating the message because IIRC changelog may be cut from the commit message to be put in the changelog file.

Purely an oversight from my end. I'll update it once the PR needs work again.

badumbatish:
sorry about dragging this out, I think I ought to close out these stale PRs now since I'm more focused on my full time job now. Thanks for everything :)

PR #2982, opened 2024-05-08
Inline asm

badumbatish:
Draft PR for my mentors


gcc/rust/ast/rust-expr.h
```diff
@@ -4729,15 +4726,47 @@ struct TupleTemplateStr
...
+
 public:
   std::vector<InlineAsmTemplatePiece> template_;
   std::vector<TupleTemplateStr> template_strs;
   std::vector<InlineAsmOperand> operands;
-  TupleClobber clobber_abi;
-  InlineAsmOptions options;
+  std::vector<TupleClobber> clobber_abi;
+  // std::set<InlineAsmOptions> options;
+  std::set<InlineAsmOptions> options;
```
CohenArthur, line 4743: I think the `InlineAsmOptions` should be renamed to `InlineAsmOption` (no 's'), especially since we are keeping a set of them. this is very nitpicky but reading that member I thought we might be dealing with a set of sets of asm options, and I got confused

```suggestion
  std::set<InlineAsmOptions> options;
```

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -109,13 +108,12 @@ std::unordered_map<std::string, AST::MacroTranscriberFunc>
     {"include", MacroBuiltin::include_handler},
     {"format_args", format_args_maker (AST::FormatArgs::Newline::No)},
     {"format_args_nl", format_args_maker (AST::FormatArgs::Newline::Yes)},
+    {"asm", MacroBuiltin::nonglobal_asm_handler},
+    {"global_asm", MacroBuiltin::global_asm_handler},
```
CohenArthur, line 112: here you can use the same technique as we did for `format_args_maker` - because both `nonglobal_asm_handler` and `global_asm_handler` call into `parse_asm`, simply with a different last parameter, it's cleaner to have just one function/lambda and it's easier to maintain

gcc/rust/ast/rust-expr.h
badumbatish: resolved

gcc/rust/ast/rust-expr.h
```diff
@@ -4729,15 +4726,47 @@ struct TupleTemplateStr
...
+  bool is_marked_for_strip () const override { return false; }
+
+  std::vector<Attribute> &get_outer_attrs () override { return outer_attrs; }
+
+  void set_outer_attrs (std::vector<Attribute> v) override { outer_attrs = v; }
+
+  ExprWithoutBlock *clone_expr_without_block_impl () const override
+  {
+    return nullptr;
+  }
```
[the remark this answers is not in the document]
badumbatish: resolved

gcc/rust/expand/rust-macro-builtins.cc
badumbatish: resolved

gcc/rust/ast/rust-expr.h
```diff
@@ -5,6 +5,7 @@
 #include "rust-path.h"
 #include "rust-macro.h"
 #include "rust-operators.h"
+#include <memory>
```
CohenArthur, line 8: this system include should be replaced by `#include "rust-system.h"` which should provide the content of `<memory>`
```suggestion
#include "rust-system.h"
```

gcc/rust/ast/rust-expr.h
```diff
@@ -4620,6 +4621,23 @@ struct AnonConst
 {
   NodeId id;
   std::unique_ptr<Expr> value;
+  AnonConst () {}
+  AnonConst (const AnonConst &other)
+  {
+    id = other.id;
+    value = other.value == nullptr
+	      ? nullptr
+	      : std::unique_ptr<Expr> (other.value->clone_expr ());
```
CohenArthur, line 4630: I think it'd be clearer to use a regular `if` here 
```suggestion
    if (other.value)
        value = other.value->clone_expr ();
```

and `value` should be set to `nullptr` automatically anyway

gcc/rust/ast/rust-expr.h
```diff
@@ -4620,6 +4621,23 @@ struct AnonConst
...
+  }
+
+  AnonConst operator= (const AnonConst &other)
+  {
+    id = other.id;
+    value = other.value == nullptr
+	      ? nullptr
+	      : std::unique_ptr<Expr> (other.value->clone_expr ());
+    return *this;
```
CohenArthur, line 4639: same here

gcc/rust/ast/rust-expr.h
```diff
@@ -4640,6 +4658,10 @@ struct InlineAsmRegOrRegClass
     std::string Symbol;
   };
 
+  Type type;
+  struct Reg reg;
+  struct RegClass regClass;
```
CohenArthur, line 4663: nit: we usually go for snake case in class members
```suggestion
  struct RegClass reg_class;
```

gcc/rust/ast/rust-expr.h
```diff
@@ -4660,20 +4682,79 @@ struct InlineAsmOperand
...
     InlineAsmRegOrRegClass reg;
     std::unique_ptr<Expr> expr;
+
+    In () {}
+    In (const struct In &other)
+    {
+      reg = other.reg;
+      expr = other.expr == nullptr
+	       ? nullptr
+	       : std::unique_ptr<Expr> (other.expr->clone_expr ());
```
CohenArthur, line 4692: same here:

```suggestion
      reg = other.reg;
      if (other.expr)
          expr = other.expr->clone_expr ();
```

gcc/rust/ast/rust-expr.h
```diff
@@ -4660,20 +4682,79 @@ struct InlineAsmOperand
...
+    }
+
+    In operator= (const struct In &other)
+    {
+      reg = other.reg;
+      expr = other.expr == nullptr
+	       ? nullptr
+	       : std::unique_ptr<Expr> (other.expr->clone_expr ());
+
+      return *this;
```
CohenArthur, line 4702: same here

gcc/rust/ast/rust-expr.h
```diff
@@ -4692,7 +4800,34 @@ struct InlineAsmOperand
...
+      sym = std::unique_ptr<Expr> (other.sym->clone_expr ());
+      return *this;
+    }
   };
+  RegisterType registerType;
+
+  struct In in;
+  struct Out out;
+  struct InOut inOut;
+  struct SplitInOut splitInOut;
```
CohenArthur, line 4821:
```suggestion
  struct InOut in_out;
  struct SplitInOut split_in_out;
```

gcc/rust/ast/rust-expr.h
```diff
@@ -4731,13 +4863,44 @@ struct TupleTemplateStr
 // Inline Assembly Node
 class InlineAsm : public ExprWithoutBlock
 {
+private:
+  location_t locus;
+  // TODO: Not sure how outer_attrs plays with InlineAsm, I put it here in order
+  // to override, very hacky.
+  std::vector<Attribute> outer_attrs;
```
CohenArthur, line 4870: what exactly was the issue here? parsing?

gcc/rust/ast/rust-expr.h
```diff
@@ -4731,13 +4863,44 @@ struct TupleTemplateStr
...
+  bool is_marked_for_strip () const override { return false; }
+
+  std::vector<Attribute> &get_outer_attrs () override { return outer_attrs; }
+
+  void set_outer_attrs (std::vector<Attribute> v) override { outer_attrs = v; }
+
+  InlineAsm *clone_expr_without_block_impl () const override
+  {
+    return new InlineAsm (*this);
+  }
```
CohenArthur, line 4903: this is correct!

gcc/rust/checks/errors/borrowck/ffi-polonius/Cargo.lock
CohenArthur, line 1: This file shouldn't be part of this PR

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -16,5 +16,528 @@
...
+
+int
+parseDirSpec (Parser<MacroInvocLexer> &parser, TokenId last_token_id)
+{
+  return 0;
+}
+
+int
+parse_clobber_abi (Parser<MacroInvocLexer> &parser, TokenId last_token_id,
+		   InlineAsmContext &inlineAsmCtx)
```
CohenArthur, line 42: one thing we can look at here is improving error handling by returning a `tl::expected<T, E>` - this way, errors can be properly chained and emitted whenever we'd like

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -16,5 +16,528 @@
...
+    {
+      // TODO: Raise error exactly like rustc if left parenthesis is not
+      // encountered.
+      token = parser.peek_current_token ();
+
+      // TODO: Error reporting shifted to the left 1 character, I'm not sure
+      // why.
+      if (token->get_id () == last_token_id)
+	{
+	  rust_error_at (parser.peek_current_token ()->get_locus (),
```
CohenArthur, line 59:
```suggestion
	  rust_error_at (token->get_locus (),
```

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -16,5 +16,528 @@
...
+			 "expected `(`, found end of macro arguments");
+	  return -1;
+	}
+      else
+	{
+	  rust_error_at (
+	    parser.peek_current_token ()->get_locus (),
```
CohenArthur, line 66:
```suggestion
	    token->get_locus (),
```

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -16,5 +16,528 @@
...
+	    "expected `(`, found `%s`",
+	    parser.peek_current_token ()->get_token_description ());
```
CohenArthur, line 68:
```suggestion
	    token->get_token_description ());
```

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -16,5 +16,528 @@
...
+      if (token->get_id () == STRING_LITERAL)
+	{
+	  // TODO: Caring for span in here.
+	  new_abis.push_back ({token->as_string (), token->get_locus ()});
+	}
+      else
+	{
+	  // TODO: We encountered something that is not string literal, which
+	  // should be illegal, please emit the correct error
+	  // https://github.com/rust-lang/rust/blob/b92758a9aef1cef7b79e2b72c3d8ba113e547f89/compiler/rustc_builtin_macros/src/asm.rs#L387
```
CohenArthur, line 99: this is good - add an extra `rust_unreachable` here so that the parser doesn't just keep going silently :)
```suggestion
	{
	  // TODO: We encountered something that is not string literal, which
	  // should be illegal, please emit the correct error
	  // https://github.com/rust-lang/rust/blob/b92758a9aef1cef7b79e2b72c3d8ba113e547f89/compiler/rustc_builtin_macros/src/asm.rs#L387
	  rust_unreachable ();
```

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -16,5 +16,528 @@
...
+
+      if (parser.skip_token (RIGHT_PAREN))
+	{
+	  break;
+	}
+
+      if (!parser.skip_token (COMMA))
+	{
+	  // TODO: If the skip of comma is unsuccessful, which should be
+	  // illegal, pleaes emit the correct error.
```
CohenArthur, line 110: same here
```suggestion
	{
	  // TODO: If the skip of comma is unsuccessful, which should be
	  // illegal, pleaes emit the correct error.
	  rust_unreachable ();
```

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -16,5 +16,528 @@
...
+      regClass.regClass.Symbol = token->as_string ();
+    }
+  else
+    {
+      // TODO: This should emit error
+      //  return
+      //  Err(p.dcx().create_err(errors::ExpectedRegisterClassOrExplicitRegister
+      //  {
+      //           span: p.token.span,
+      //       }));
```
CohenArthur, line 169:
```suggestion
      // TODO: This should emit error
      //  return
      //  Err(p.dcx().create_err(errors::ExpectedRegisterClassOrExplicitRegister
      //  {
      //           span: p.token.span,
      //       }));
      rust_unreachable ();
```

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -16,5 +16,528 @@
...
+  //           Some(ident.name)
+  //       } else {
+  //           None
+  //       };
+
+  using RegisterType = AST::InlineAsmOperand::RegisterType;
+  AST::InlineAsmOperand reg_operand;
+  auto token = parser.peek_current_token ();
+  auto iden_token = parser.peek_current_token ();
+  auto &inlineAsm = inlineAsmCtx.inlineAsm;
```
CohenArthur, line 207:
```suggestion
  auto &inline_asm = inlineAsmCtx.inlineAsm;
```

CohenArthur: (the field should also be named `inline_asm` haha)

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -16,5 +16,528 @@
...
+      return tl::nullopt;
+    }
+  else if (!is_global_asm && check_identifier (parser, "inout"))
+    {
+      return tl::nullopt;
+    }
+  else if (!is_global_asm && check_identifier (parser, "inlateout"))
+    {
+      return tl::nullopt;
+    }
```
CohenArthur, line 241: I think this might be cleaner if it was turned into an array of strings to check for, and if any matches, return `tl::nullopt`. what do we check here exactly?

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -16,5 +16,528 @@
...
+void
+check_and_set (Parser<MacroInvocLexer> &parser, InlineAsmContext &inlineAsmCtx,
+	       AST::InlineAsmOption option)
+{
+  auto &inlineAsm = inlineAsmCtx.inlineAsm;
+  if (inlineAsm.options.count (option) != 0)
+    {
+      // TODO: report an error of duplication
+      rust_error_at (parser.peek_current_token ()->get_locus (),
+		     "the `%s` option was already provided",
```
CohenArthur, line 273:
```suggestion
		     "the %qs option was already provided",
```

use `%qs` for quoted strings so that we have the nice unicode quotes if our terminals support them :D

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -16,5 +16,528 @@
...
+	{
+	  rust_error_at (parser.peek_current_token ()->get_locus (),
+			 "expected `(`, found end of macro arguments");
+	  return -1;
+	}
+      else
+	{
+	  rust_error_at (
+	    parser.peek_current_token ()->get_locus (),
+	    "expected `(`, found `%s`",
```
CohenArthur, line 67: same here
```suggestion
	    "expected `(`, found %qs",
```

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -16,5 +16,528 @@
...
+	}
+
+      // Parse comma as optional
+      if (parser.skip_token (COMMA))
+	{
+	  continue;
+	}
+      else
+	{
+	  std::cout << "Sum wrong here, check it out" << std::endl;
```
CohenArthur, line 356: :smile: 

I think we can keep that in a `rust_debug` but it should be removed from the PR :P

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -93,6 +92,22 @@ format_args_maker (AST::FormatArgs::Newline nl)
   };
 }
 
+enum class isGlobalAsm
+{
+  Yes,
+  No,
+};
```
CohenArthur, line 99:
```suggestion
enum class AsmType
{
  Global,
  Inline,
};
```

CohenArthur (approved):
this is great progress :DD well done! a couple nitpicks and things to fix but we can merge this soon

gcc/rust/ast/rust-expr.h
badumbatish: resolved

gcc/rust/expand/rust-macro-builtins-asm.cc
badumbatish: this is will eventually constructor instances of RegisterOperand class and here i'm just setting up the scaffolding to match rust's asm.rs for easier matching right now. Different lines of code should come out from these if statements

badumbatish: we're checking if the identifier matches each of the name, which stands for `dir_spec := "in" / "out" / "lateout" / "inout" / "inlateout"` in the ABNF. i guess it also goes with whether the asm call is global or not too so we also need to check for that

gcc/rust/expand/rust-macro-builtins-asm.cc
badumbatish: 💀💀💀💀 i have no idea how i let this slipped through the post commit stage 💀💀💀

gcc/rust/ast/rust-expr.h
badumbatish: i'm just not sure what outer_attrs mean in general and how it plays out with asm!

badumbatish:
Pr's looking good, i let some quoted strings warnings slipped through but it should get fixed soon. I'll look into tl::expected soon

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -16,5 +16,709 @@
...
+  {AST::InlineAsmOption::PRESERVES_FLAGS, "preserves_flags"},
+  {AST::InlineAsmOption::NORETURN, "noreturn"},
+  {AST::InlineAsmOption::NOSTACK, "nostack"},
+  {AST::InlineAsmOption::MAY_UNWIND, "may_unwind"},
+  {AST::InlineAsmOption::ATT_SYNTAX, "att_syntax"},
+  {AST::InlineAsmOption::RAW, "raw"},
+};
+
+tl::expected<InlineAsmContext, std::string>
+parse_clobber_abi (InlineAsmContext inline_asm_ctx)
```
CohenArthur, line 35: this is good! in another PR, I'd like you to start working on a "proper" error type instead of strings, something like

```c++
enum class AsmParseError {
    Reason1,
    Reason2,
    MissingAbiArgument,
    ...
}
```

it'll clean up the code a little bit and we can then see how we'll transform these into proper `Error`s to emit for the user

CohenArthur (approved):

gcc/rust/ast/rust-expr.h
```diff
@@ -6,6 +6,8 @@
 #include "rust-path.h"
 #include "rust-macro.h"
 #include "rust-operators.h"
+#include "rust-system.h"
+#include <memory>
```
CohenArthur, line 10: ah, the system include is still here. you should be able to remove it entirely

gcc/rust/ast/rust-expr.h
```diff
@@ -4757,34 +4778,138 @@ struct InlineAsmOperand
...
   };
 
   struct InOut
   {
-    InlineAsmRegOrRegClass reg;
+    tl::optional<InlineAsmRegOrRegClass> reg;
     bool late;
     std::unique_ptr<Expr> expr; // this can't be null
+
+    InOut () {}
```
CohenArthur, line 4849: if `expr` cannot be null then it would be nice to add more assertions to this struct (but in another PR)

1. remove the empty constructor, as that will set `expr` to `nullptr` and thus create an invalid `InOut`
2. in the full constructor on line 4850, `rust_assert(expr)` to ensure that `expr` is not null
3. because `expr` is not null, remove the `if (other.expr)` checks in both the copy constructor and `operator=` overload

gcc/rust/ast/rust-expr.h
CohenArthur: ah, alright. can you open an issue for this please? titled something like `Handle outer attributes properly for inline assembly` and we'll have a look later during the project :)

gcc/rust/ast/rust-expr.h
badumbatish: gotcha

Issue #2962, opened 2024-04-23
Missing feature gate `negative_impls`


P-E-P:
This feature gate is missing, even though it is required in order to compile `core`.

https://doc.rust-lang.org/beta/unstable-book/language-features/negative-impls.html

GCCRS shall be able to understand the following attribute and reject negative impls when it is missing.
```rust
#![feature(negative_impls)]
```

badumbatish:
where should I perform rust_error() for missing feature gates, I'm not sure which file i should look into for this ?

P-E-P:
> where should I perform rust_error() for missing feature gates, I'm not sure which file i should look into for this ?

@badumbatish IIRC you should look into `gcc/rust/checks/errors/rust-feature-gate.*`

PR #3002, opened 2024-05-15
Make gccrs recognize negative_impls

badumbatish:
just a draft pr for #2962


badumbatish:
I think i got gccrs to recognize the feature negative_impl but I'm not sure how to make it so that when user do sth like this
```
trait ExampleTrait {}

impl !ExampleTrait for i32 {}
```

it fails to compile instead of continuing with compilation because we haven't implemented them

P-E-P:
> it fails to compile instead of continuing with compilation because we haven't implemented them or if the user didn't do `#![feature(negative_impls)]` in the beginning of the .rs file
> 
@badumbatish You are in the right direction! There is a visitor in `gcc/rust/checks/error/rust-feature-gate.cc` which match the AST against available features. There is a member `as_exclam` in `TraitImpl`, if you meet one in the visitor, you need to error out and that's it.

badumbatish:
hmm i think i might be cooking something

badumbatish:
hmm i got down the case where if a user attempts to do `impl !ExampleTrait for i32 {}` we error out with `// { dg-error "negative_impls are not yet implemented" }` but I don't think the rustc compiler is doing this https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b63a98a5146d438ded89b9dac762be48

Also with the case that we did use `#![feature(negative_impls)]`, we should also error out sth? https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1f3c2bd54f30a5931c02a2a4abcac08b

I'm also not sure how this is relate to core mentioned in #2962

CohenArthur:
> hmm i got down the case where if a user attempts to do `impl !ExampleTrait for i32 {}` we error out with `// { dg-error "negative_impls are not yet implemented" }` but I don't think the rustc compiler is doing this https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b63a98a5146d438ded89b9dac762be48
> 

the error is not exactly the same but that's okay - it erros out and that's what matters :D

> Also with the case that we did use `#![feature(negative_impls)]`, we should also error out sth? https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1f3c2bd54f30a5931c02a2a4abcac08b
> 

we don't have a nightly and stable versions of the compiler at the moment, so you can just assume that everything is "nightly". so we handle having `#![feature]` at all times, and you don't need to do any particular things regarding nightly vs stable. so the right behavior is to not error out if the feature gate was enabled, like rustc does when run in nightly

> I'm also not sure how this is relate to core mentioned in #2962, there seems to not have an enum state with "nightly" or sth for me to use on this

gcc/testsuite/rust/compile/negative_impls.rs
```diff
@@ -0,0 +1,10 @@
+// { dg-output "" }
```
CohenArthur, line 1: since this is a `compile` test, and not an `execute` one (it's in `gcc/testsuite/rust/compile/` rather than `gcc/testsuite/rust/execute`) there is no need to check the `stdout` output with `dg-output` - so you can remove that line :)

gcc/testsuite/rust/compile/negative_impls_2.rs
```diff
@@ -0,0 +1,19 @@
...
+// 8 | impl !ExampleTrait for i32 {}//
+//   |      ^^^^^^^^^^^^^
+//   |
+//   = note: see issue #68318 <https://github.com/rust-lang/rust/issues/68318> for more information
+
+// For more information about this error, try `rustc --explain E0658`.
+// error: could not compile `playground` (bin "playground") due to 1 previous error
+trait ExampleTrait {}
+
+impl !ExampleTrait for i32 {} // { dg-error "negative_impls are not yet implemented" }
```
CohenArthur, line 16: perfect!

gcc/rust/checks/errors/rust-feature-gate.cc
```diff
@@ -144,4 +145,14 @@ FeatureGate::visit (AST::ExternalTypeItem &item)
...
 }
 
+void
+FeatureGate::visit (AST::TraitImpl &impl)
+{
+  if (impl.is_exclam ())
+    {
+      gate (Feature::Name::NEGATIVE_IMPLS, impl.get_locus (),
+	    "negative_impls are not yet implemented");
+    }
```
CohenArthur, line 155: nitpick:
```suggestion
      gate (Feature::Name::NEGATIVE_IMPLS, impl.get_locus (),
	    "negative_impls are not yet implemented");
```

gcc/rust/checks/errors/rust-feature.cc
```diff
@@ -54,6 +57,7 @@ const std::map<std::string, Feature::Name> Feature::name_hash_map = {
   {"decl_macro", Feature::Name::DECL_MACRO},
   // TODO: Rename to "auto_traits" when supporting
   // later Rust versions
+  {"negative_impls", Feature::Name::NEGATIVE_IMPLS},
```
CohenArthur, line 60:
```suggestion
  {"negative_impls", Feature::Name::NEGATIVE_IMPLS},
  // TODO: Rename to "auto_traits" when supporting
  // later Rust versions
```

so that the comment still concerns the right feature gate

CohenArthur (approved):
nice job!

gcc/testsuite/rust/compile/negative_impls_2.rs
```diff
@@ -0,0 +1,19 @@
...
+
+
+fn main() {}
```
P-E-P, line 19: `main` functions are not mandatory

gcc/testsuite/rust/compile/negative_impls.rs
badumbatish: rebase edited

gcc/rust/checks/errors/rust-feature-gate.cc
badumbatish: rebase edited

badumbatish:
Ill squash everything down to 1 commit after this rebase from master passes

P-E-P:
@badumbatish Could you rewrite your commit message please ?

CohenArthur (approved):
looks great! well done!

PR #3004, opened 2024-05-16
Add call and method call default visitors

powerboat9:


P-E-P:
Could you add some ast traversals test to highlight this change ? I'd like to keep the amount of regressions low and this should have made a test fail when I refactored the visitors.

P-E-P (approved):
Looks good, thank you!

PR #3005, opened 2024-05-16
Visit constant items without expressions properly

powerboat9:
This should improve handling of constant expressions for name resolution 2.0


P-E-P (approved):
This will probably fix some ICE in the resolver, nice!

PR #3010, opened 2024-05-18
Create new test system for name resolution 2.0

powerboat9:
This should allow us to more easily test and record progress on name resolution 2.0


powerboat9:
Note that a decent fraction of new tests after this is merged (?) will have to be entered into the ```exclude``` file.

CohenArthur:
@powerboat9 I like the implementation and I think this feature makes sense, but I'm not sure if it's going to make our life easier in the long run. I'm in favor of merging it if the rest of the community thinks it is a good idea. it does add a burden which is maintaining and revisiting that `exclude` file. but the alternative is also annoying, as we'd need to add a copy of existing testcases with `-frust-name-resolution-2.0` added as an argument... 

maybe we can do something like keep the `nr2` folder, run it with different dejagnu options but only keep symlinks to testcases that pass under NR2.0?

P-E-P:
> maybe we can do something like keep the `nr2` folder, run it with different dejagnu options but only keep symlinks to testcases that pass under NR2.0?

AFAIK symlink wouldn't work when running tests on windows.


On one hand I really like this PR's idea but maintaining this list of file looks horrible. I'm in favor of merging this for now, we could maintain that list until the nr2 is stable and working as expected, we could then come back to this decision.

powerboat9:
Thoughts on merging this? Should I make some tweaks?

powerboat9:
Where is the development on nr2.0 taking place? I feel like I might be a bit out of the loop

P-E-P:
> Where is the development on nr2.0 taking place? I think I might be a bit out of the loop

I need to open a PR but I've been working on top of #2940. If you want to get an exact idea on what is working/missing you could compile my branch https://github.com/P-E-P/gccrs/tree/allow-use-before-items-being-used-continued. Right now one test is still failing (use duplication, name_resolution21.rs).

powerboat9:
When I try to force-enable nr2.0 on `allow-use-before-items-being-used-continued` (by replacing `flag_name_resolution_2_0` references with `1`) I get a bunch of test errors. Are you sure there's only one non-working test?

P-E-P:
> When I try to force-enable nr2.0 on `allow-use-before-items-being-used-continued` (by replacing `flag_name_resolution_2_0` references with `1`) I get a bunch of test errors. Are you sure there's only one non-working test?

You're right, I didn't apply the correct patch. Looks like most of those fails are due to an ICE (but not all of them):
- rust/resolve/rust-late-name-resolver-2.0.cc:206
- rust/resolve/rust-late-name-resolver-2.0.cc:157
- rust/hir/rust-ast-lower-base.cc:655
- rust/resolve/rust-name-resolver.cc:661
- We also have a segfault somewhere

The situation is worse than I initially thought.

P-E-P (approved):
I can't fix those in the upcoming PR as it is becoming too big. And there is a lot of work remaining, we should merge this PR.

P-E-P:
Wait I thought the error was due to some github queue shenanigans but it doesn't. @powerboat9 would you mind rebasing this PR on master ?

powerboat9:
Yep, looks like it needs some tweaks after some tests were moved around. Will need to be adjusted to handle sub directories properly.

powerboat9:
Alright, this should do it. We should also now get `XPASS` reports when tests in the exclude list no longer need to be excluded.

PR #3011, opened 2024-05-19
Remove cstddef header from rust-fmt

### LINKED ISSUE #3006: Remove `<cstddef>` include in `rust-fmt.h`
https://github.com/Rust-GCC/gccrs/blob/c7687ef1f7a9d2683bb45dcbf547b42aae856075/gcc/rust/ast/rust-fmt.h#L23

badumbatish:
A short pr, this fixes #3006

gcc/rust/ChangeLog:

	* ast/rust-fmt.h: title.


dkm:
I would rephrase the Changelog part. Referencing the title is not a good idea, as it will not be kept during the daily bump that gathers the last commit and append the changelog part to the different ChangeLog files.
Best to have something like "* ast/rust-fmt.h: Remove useless cstddef include."

:)

badumbatish:
> I would rephrase the Changelog part. Referencing the title is not a good idea, as it will not be kept during the daily bump that gathers the last commits and appends the changelog parts to the different ChangeLog files. Best to have something like "* ast/rust-fmt.h: Remove useless cstddef include."
> 
> :)

ahh i think i made this mistake before, i'll force push it right now

dkm:
> > I would rephrase the Changelog part. Referencing the title is not a good idea, as it will not be kept during the daily bump that gathers the last commits and appends the changelog parts to the different ChangeLog files. Best to have something like "* ast/rust-fmt.h: Remove useless cstddef include."
> > :)
> 
> ahh i think i made this mistake before, i'll force push it right now

:+1: No big deal of course...
(I see 2 force push, but I don't think you've changed anything to the commit or the PR description (the former can stay, we don't even need the changelog part here))

badumbatish:
hmm i definitely did not read your comments too fast and thought you were talking about my commit title hahaha.

i have a question tho. So when we change a lot of files, what should we write if they are all related to the same issue? Do we just reference the issue number itself?

P-E-P (approved):
LGTM :+1:

PR #3012, opened 2024-05-20
Parse box keyword

P-E-P:
Add ability to parse the box syntax but this PR does not lower it. Add the corresponding feature gate as well as two tests.


CohenArthur (commented):
great stuff :+1:

PR #3013, opened 2024-05-20
borrowck: Added location support to BIR nodes

braw-lee:
depends on #3076 
Added `location` field to BIR::Statement and BIR::Place, updated the BIR builders to take the source location from HIR and pass it to BIR node.

### DIFF
--- a/gcc/rust/checks/errors/borrowck/rust-bir-builder-internal.h
+++ b/gcc/rust/checks/errors/borrowck/rust-bir-builder-internal.h
@@ -237,33 +237,35 @@ protected:
   }
 
 protected: // Helpers to add BIR statements
-  void push_assignment (PlaceId lhs, AbstractExpr *rhs)
+  void push_assignment (PlaceId lhs, AbstractExpr *rhs, location_t location)
   {
-    ctx.get_current_bb ().statements.emplace_back (lhs, rhs);
+    ctx.get_current_bb ().statements.emplace_back (lhs, rhs, location);
     translated = lhs;
   }
 
-  void push_assignment (PlaceId lhs, PlaceId rhs)
+  void push_assignment (PlaceId lhs, PlaceId rhs, location_t location)
   {
-    push_assignment (lhs, new Assignment (rhs));
+    push_assignment (lhs, new Assignment (rhs), location);
   }
 
-  void push_tmp_assignment (AbstractExpr *rhs, TyTy::BaseType *tyty)
+  void push_tmp_assignment (AbstractExpr *rhs, TyTy::BaseType *tyty,
+			    location_t location)
   {
     PlaceId tmp = ctx.place_db.add_temporary (tyty);
     push_storage_live (tmp);
-    push_assignment (tmp, rhs);
+    push_assignment (tmp, rhs, location);
   }
 
-  void push_tmp_assignment (PlaceId rhs)
+  void push_tmp_assignment (PlaceId rhs, location_t location)
   {
-    push_tmp_assignment (new Assignment (rhs), ctx.place_db[rhs].tyty);
+    push_tmp_assignment (new Assignment (rhs), ctx.place_db[rhs].tyty,
+			 location);
   }
 
-  void push_switch (PlaceId switch_val,
+  void push_switch (PlaceId switch_val, location_t location,
 		    std::initializer_list<BasicBlockId> destinations = {})
   {
-    auto copy = move_place (switch_val);
+    auto copy = move_place (switch_val, location);
     ctx.get_current_bb ().statements.emplace_back (Statement::Kind::SWITCH,
 						   copy);
     ctx.get_current_bb ().successors.insert (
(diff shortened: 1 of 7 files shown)


braw-lee:
should I work on my branch the whole time, and you can merge it after 2-3 months?
or should I just create new PRs whenever I get something done and you will merge those from time to time? @CohenArthur @P-E-P

gcc/rust/checks/errors/borrowck/rust-bir-builder-internal.h
```diff
@@ -237,25 +237,25 @@ class AbstractBuilder
...
     PlaceId tmp = ctx.place_db.add_temporary (tyty);
     push_storage_live (tmp);
-    push_assignment (tmp, rhs);
+    push_assignment (tmp, rhs, tyty->get_locus ());
   }
 
-  void push_tmp_assignment (PlaceId rhs)
+  void push_tmp_assignment (PlaceId rhs, location_t location)
   {
     push_tmp_assignment (new Assignment (rhs), ctx.place_db[rhs].tyty);
```
CohenArthur, line 260: we're not using the `location` argument here yet - I assume this is intentional?

gcc/rust/checks/errors/borrowck/rust-bir.h
```diff
@@ -77,26 +77,40 @@ class Statement
   // otherwise: <unused>
   std::unique_ptr<AbstractExpr> expr;
   TyTy::BaseType *type;
+  // stores location of the actual expression from source code
+  // currently only available when kind == Kind::ASSIGNMENT
+  tl::optional<location_t> location;
```
CohenArthur, line 82: do we want to make this optional? if this is just a temporary measure, then we should add a `FIXME` comment

gcc/rust/checks/errors/borrowck/rust-bir-builder-pattern.h
```diff
@@ -82,7 +82,9 @@ class PatternBindingBuilder : protected AbstractBuilder,
 
     if (init.has_value ())
       {
-	push_assignment (translated, init.value ());
+	push_assignment (
+	  translated, init.value (),
```
jdupak, line 86: Maybe, it would make more sense here to pass the location to the visit identifier method.

gcc/rust/checks/errors/borrowck/rust-bir-place.h
```diff
@@ -78,12 +78,13 @@ struct Place
   TyTy::BaseType *tyty;
   FreeRegions regions{{}};
   std::vector<LoanId> borrowed_by{};
+  location_t location;
```
jdupak, line 81: I am not sure, if this is needed. We might have just locations for Origin.

gcc/rust/checks/errors/borrowck/rust-bir-place.h
```diff
@@ -255,7 +256,9 @@ class PlaceDB
 
   PlaceId add_variable (NodeId id, TyTy::BaseType *tyty)
   {
-    return add_place ({Place::VARIABLE, id, {}, is_type_copy (tyty), tyty}, 0);
+    return add_place (
+      {Place::VARIABLE, id, {}, is_type_copy (tyty), tyty, tyty->get_locus ()},
```
jdupak, line 260: Using location of type does not seem right. As I mentioned before, we might not oven need it here. Alternative could be to extract the location from NodeId. But that can always be done on demand.

gcc/rust/checks/errors/borrowck/rust-bir.h
jdupak: If this is a TODO, it would make more sense to use just location and init it to UNKNOWN_LOCATION.

jdupak (changes requested):
Mostly looks good. There are some issues around Places.

gcc/rust/checks/errors/borrowck/rust-bir.h
braw-lee: ok, will fix that

gcc/rust/checks/errors/borrowck/rust-bir-place.h
braw-lee: actually I was first trying to use get_ast_item() method to retrieve the location, but it seems like the location is copied from AST->HIR->TyTy, so getting locus from NodeId or the TyTy should be same, unless we plan to change the location information of TyTy in future

CohenArthur (approved):

CohenArthur:
This needs to be rebased as well @braw-lee

gcc/rust/checks/errors/borrowck/rust-bir-builder-internal.h
```diff
@@ -301,45 +303,50 @@ class AbstractBuilder
...
+  template <typename T>
+  void move_all (T &args, std::vector<location_t> capture_locations)
   {
-    std::transform (args.begin (), args.end (), args.begin (),
-		    [this] (PlaceId arg) { return move_place (arg); });
+    rust_assert (args.size () == capture_locations.size ());
+    std::transform (args.begin (), args.end (), capture_locations.begin (),
+		    args.begin (), [this] (PlaceId arg, location_t location) {
+		      return move_place (arg, location);
+		    });
```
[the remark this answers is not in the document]
jdupak: Well , I think that captures are the only place, where you have multiple diff­erent loc­ations. But In tha­t case this function should be renamed. I'm not sure if such function makes sense in this context.

CohenArthur (approved):
LGTM!

PR #3020, opened 2024-05-21
Parse raw borrow

P-E-P:
Raw borrow syntax was not parsed correctly.

Requires  #3014


gcc/rust/parse/rust-parse-impl.h
```diff
@@ -12372,33 +12372,64 @@ Parser<ManagedTokenSource>::null_denotation_not_path (
...
+	      {
+	      case MUT:
+		mutability = Mutability::Mut;
+		break;
+	      case CONST:
+		mutability = Mutability::Imm;
+		break;
+	      default:
+		rust_error_at (lexer.peek_token ()->get_locus (),
+			       "raw borrow should be either const or mut");
```
CohenArthur, line 12405: I can't remember so feel free to disregard, but shouldn't we do something like `add_error(Error(...))` here? so that parser errors are only emitted if we want them to be?

CohenArthur (commented):
great work, the testcases are good, edge cases are handled, well done :heart:

PR #3043, opened 2024-05-27
Corrected access specifiers

braw-lee:
`locus` should be private, we also have a getter for it
looks like the access specifier was misstyped


P-E-P:
@braw-lee there is no changelog in your commit message

P-E-P (approved):
Nice catch!

PR #3044, opened 2024-06-06
Convert Polonius output in rust to C++ format

braw-lee:
1. DONE: Added FFIVector<T>, it will be used to convert Polonius output in rust to C++ format.
2. DONE: Provide helper implementations for all required types.
3. DONE: Construct the output and pass to C++.
4. DONE: Use the newly converted data to trigger borrow checking errors.

### DIFF
--- a/gcc/rust/checks/errors/borrowck/ffi-polonius/src/lib.rs
+++ b/gcc/rust/checks/errors/borrowck/ffi-polonius/src/lib.rs
@@ -25,24 +25,18 @@ use std::fmt::Debug;
 use std::hash::Hash;
 
 extern "C" {
-    #[allow(dead_code)]
-    fn FFIVector__new(capacity: usize) -> FFIVector<usize>;
-    #[allow(dead_code)]
+    fn FFIVector__new(capacity: usize) -> *mut FFIVector<usize>;
     fn FFIVector__new_vec_pair(
         capacity: usize,
-    ) -> FFIVector<gccrs_ffi::Pair<usize, *mut FFIVector<usize>>>;
-    #[allow(dead_code)]
+    ) -> *mut FFIVector<gccrs_ffi::Pair<usize, *mut FFIVector<usize>>>;
     fn FFIVector__new_vec_triple(
         capacity: usize,
-    ) -> FFIVector<gccrs_ffi::Triple<usize, usize, usize>>;
-    #[allow(dead_code)]
+    ) -> *mut FFIVector<gccrs_ffi::Triple<usize, usize, usize>>;
     fn FFIVector__push(vector: *mut FFIVector<usize>, element: usize);
-    #[allow(dead_code)]
     fn FFIVector__push_vec_pair(
         vector: *mut FFIVector<gccrs_ffi::Pair<usize, *mut FFIVector<usize>>>,
         element: gccrs_ffi::Pair<usize, *mut FFIVector<usize>>,
     );
-    #[allow(dead_code)]
     fn FFIVector__push_vec_triple(
         vector: *mut FFIVector<gccrs_ffi::Triple<usize, usize, usize>>,
         element: gccrs_ffi::Triple<usize, usize, usize>,
@@ -68,6 +62,13 @@ impl From<usize> for GccrsAtom {
 
 impl From<GccrsAtom> for usize {
     fn from(atom: GccrsAtom) -> Self {
+        atom.index()
+    }
+}
+
+
+impl From<&GccrsAtom> for usize {
+    fn from(atom: &GccrsAtom) -> Self {
         atom.index()
     }
 }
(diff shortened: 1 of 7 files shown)


braw-lee:
I am currently generating 2 dump files
`rust_side_polonius_output.txt`
`cpp_side_polonius_output.txt`
and making sure that the diff command gives no output

this is just to verify that the data was correctly transfered from rust side to cpp side
I wrote that stuff just to test while writing code, 
should I get rid of the dump lines ?
@CohenArthur @P-E-P @jdupak

P-E-P:
> I am currently generating 2 dump files `rust_side_polonius_output.txt` `cpp_side_polonius_output.txt` and making sure that the diff command gives no output
> 
> this is just to verify that the data was correctly transfered from rust side to cpp side I wrote that stuff just to test while writing code, should I get rid of the dump lines ? @CohenArthur @P-E-P @jdupak

If you think this is worthwhile and could help you later you could add them behind a flag.

gcc/rust/checks/errors/borrowck/ffi-polonius/src/lib.rs
```diff
@@ -172,9 +198,78 @@ pub unsafe extern "C" fn polonius_run(
...
+                second: origin_1,
+                third: origin_2,
+            };
+            FFIVector__push_vec_triple(subset_errors, triple);
+        }
+    }
+
+    // write a dump for polonius output to verify data is correctly moved to C++ side
+    use std::fs;
+    fs::write("rust_side_polonius_output.txt", buffer_string).expect("Error writing to file!");
```
CohenArthur, line 269: hm, I would remove the file output here and all of the operations on `buffer_string`. I think it is a good way of debugging this (and I'm glad you thought about it, I think it was the correct move) but it makes the final code very hard to read. I think we should be able to find memory issues by running the compiler with ASAN, so I'm not sure on how useful this will be.

if you think it is worth keeping, then I would put it in a completely different function. both functions could iterate on the errors, and one could build the `buffer_string` while the other could build the resulting `loan_errors`, `move_errors` and `subset_errors` vectors

gcc/rust/checks/errors/borrowck/ffi-polonius/src/lib.rs
```diff
@@ -19,10 +19,36 @@
...
+        capacity: usize,
+    ) -> *mut FFIVector<gccrs_ffi::Triple<usize, usize, usize>>;
+    #[allow(dead_code)]
+    fn FFIVector__push(vector: *mut FFIVector<usize>, element: usize);
+    #[allow(dead_code)]
+    fn FFIVector__push_vec_pair(
+        vector: *mut FFIVector<gccrs_ffi::Pair<usize, *mut FFIVector<usize>>>,
+        element: gccrs_ffi::Pair<usize, *mut FFIVector<usize>>,
+    );
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
braw-lee: yes, were generating errors in initial commits, will remove in later commits

gcc/rust/checks/errors/borrowck/ffi-polonius/src/lib.rs
braw-lee: I think I won't need really need this, will get rid of the debugging code

gcc/rust/checks/errors/borrowck/ffi-polonius/src/lib.rs
```diff
@@ -172,9 +192,54 @@ pub unsafe extern "C" fn polonius_run(
         }
     }
 
+    let loan_errors = FFIVector__new_vec_pair(output.errors.len());
+    for (keys, values) in output.errors.iter() {
+        let loans_vec = FFIVector__new(values.len());
+        for loan in values.iter() {
+            let loan: usize = (*loan).into();
+            FFIVector__push(loans_vec, loan);
+        }
```
CohenArthur, line 201: I'm not a fan of that `(*loan).into()` syntax we have to use. I would like you to add this extra implementation at the top of the file, alongside the impl for `From<GccrsAtom> for usize`:

```rust
impl From<GccrsAtom> for usize {
    fn from(atom: GccrsAtom) -> Self {
        atom.index()
    }
}

impl From<&GccrsAtom> for usize {
    fn from(atom: &GccrsAtom) -> Self {
        atom.index()
    }
}
```

which will then enable us to do this instead:

```suggestion
        for loan in values.iter() {
            FFIVector__push(loans_vec, loan.into());
        }
```

CohenArthur: this will also help in removing all of the other `(*<something>).into()`

gcc/rust/checks/errors/borrowck/polonius/rust-polonius-ffi.h
```diff
@@ -102,11 +104,122 @@ struct FactsView
...
+  for (size_t i = 0; i < size; ++i)
+    {
+      data[i].second->drop ();
+    }
+  delete[] data;
+  size = 0;
+  capacity = 0;
+  delete this;
+}
+
```
CohenArthur, line 217: I have a feeling that this might be too complicated for what we want. I think having a simple wrapper around `std::vector` might be a little easier to deal with and would produce less code? is there something I'm not thinking of? if we have proper destructors for all of our FFI classes then we wouldn't need to call `->drop()` manually either, the `std::vector`'s destructor would take care of that

CohenArthur: I think the `extern "C"` functions around this class are really useful though

braw-lee: > I have a feeling that this might be too complicated for what we want. I think having a simple wrapper around `std::vector` might be a little easier to deal with and would produce less code? is there something I'm not thinking of? if we have proper destructors for all of our FFI classes then we wouldn't need to call `->drop()` manually either, the `std::vector`'s destructor would take care of that

these functions dealing with raw pointers are inspired from libproc_macro_internals dir
[file](https://github.com/Rust-GCC/gccrs/blob/master/libgrust/libproc_macro_internal/tokenstream.cc) 

we need to pass raw pointers to rust and our struct must be compatible with C ABI
not sure if all that will be possible with std::vector

braw-lee: > I have a feeling that this might be too complicated for what we want. I think having a simple wrapper around `std::vector` might be a little easier to deal with and would produce less code? is there something I'm not thinking of? if we have proper destructors for all of our FFI classes then we wouldn't need to call `->drop()` manually either, the `std::vector`'s destructor would take care of that

I tried the wrapper approach and ended up with this #3058

Issue #3052, opened 2024-06-13
Fully incorporate tl::expected into InlineAsm parsing


badumbatish:
.

badumbatish:
resolved in #3053

PR #3053, opened 2024-06-14
incorporate tl::expected into InlineAsm parsing

badumbatish:
Draft issue for my mentors, relates to #3052

### DIFF
--- a/gcc/rust/expand/rust-macro-builtins-asm.cc
+++ b/gcc/rust/expand/rust-macro-builtins-asm.cc
@@ -740,11 +740,9 @@ parse_asm (location_t invoc_locus, AST::MacroInvocData &invoc,
 			     is_global_asm == AST::AsmKind::Global);
   auto inline_asm_ctx = InlineAsmContext (inline_asm, parser, last_token_id);
 
-  tl::expected<InlineAsmContext, InlineAsmParseError> resulting_context
-    = tl::expected<InlineAsmContext, InlineAsmParseError> (inline_asm_ctx);
-  resulting_context.and_then (parse_format_strings)
-    .and_then (parse_asm_arg)
-    .and_then (validate);
+  auto resulting_context = parse_format_strings (inline_asm_ctx)
+			     .and_then (parse_asm_arg)
+			     .and_then (validate);
 
   // TODO: I'm putting the validation here because the rust reference put it
   // here Per Arthur's advice we would actually do the validation in a different
--- a/gcc/rust/expand/rust-macro-builtins-asm.cc
+++ b/gcc/rust/expand/rust-macro-builtins-asm.cc
@@ -208,52 +208,34 @@ parse_reg_operand (InlineAsmContext inline_asm_ctx)
 	}
     }
 
-  // For the keyword IN, currently we count it as a seperate keyword called
-  // Rust::IN search for #define RS_TOKEN_LIST in code base.
   tl::expected<InlineAsmContext, InlineAsmParseError> parsing_operand
     = tl::expected<InlineAsmContext, InlineAsmParseError> (inline_asm_ctx);
 
-  // PARSING WITH IN
-  parsing_operand.map (parse_reg_operand_in);
-  if (parsing_operand || parsing_operand.error () == COMMITTED)
-    return parsing_operand;
+  // Here is all parse_reg_operand functions we're using in a for loop
+  auto parse_funcs = {parse_reg_operand_in,	   parse_reg_operand_out,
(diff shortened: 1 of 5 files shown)


badumbatish:
while i was doing this i ran into some failed logic for parse_asm_arg where I just either parse options, clobber_abi, or 1 line of reg_operand and then it just bubbles up to the first layer of parsing. 

This prevents the asm! macro to be parse correctly. I didn't end up fixing it (maybe it should be in another PR?) but I just want to note here that it is happening

CohenArthur:
> while i was doing this i ran into some failed logic for parse_asm_arg where I just either parse options, clobber_abi, or 1 line of reg_operand and then it just bubbles up to the first layer of parsing.
> 
> This prevents the asm! macro to be parse exhaustively. I didn't end up fixing it (maybe it should be in another PR?) but I just want to note here that it is happening

yes, it would be better if it was fixed in a different PR. can you open up a new issue with a code example which causes that bug? thanks :)

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -637,9 +740,8 @@ parse_asm (location_t invoc_locus, AST::MacroInvocData &invoc,
 			     is_global_asm == AST::AsmKind::Global);
   auto inline_asm_ctx = InlineAsmContext (inline_asm, parser, last_token_id);
 
-  // operands stream, also handles the optional ","
-  tl::expected<InlineAsmContext, std::string> resulting_context
-    = tl::expected<InlineAsmContext, std::string> (inline_asm_ctx);
+  tl::expected<InlineAsmContext, InlineAsmParseError> resulting_context
+    = tl::expected<InlineAsmContext, InlineAsmParseError> (inline_asm_ctx);
```
CohenArthur, line 744: so AFAICT this is necessary purely to do the monadic operations on line 745, with `resulting_context.and_then(parse_format_strings)`

I think it would be better to have the following:

```c++
parse_format_strings(inline_asm_ctx).and_then(parse_asm_arg).and_then(validate);
```
what do you think?

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -201,22 +196,102 @@ parse_reg_operand (InlineAsmContext inline_asm_ctx)
...
-	  rust_unreachable ();
+	  name = token->as_string ();
+	}
+      else
+	{
+	  rust_error_at (token->get_locus (),
+			 "expected operand, %s, options, or "
+			 "additional template string",
+			 "clobber_abi");
+	  return tl::unexpected<InlineAsmParseError> (COMMITTED);
```
CohenArthur, line 215: in all of these you can simplify the code a little bit as C++ is able to infer the type of the return parameter

```suggestion
	  return tl::unexpected (COMMITTED);
```

but that's a nit and not required

CohenArthur (approved):
looks good to me! good work :)

gcc/rust/expand/rust-macro-builtins-asm.cc
badumbatish: nice catch Arthur, I guess i was afraid of doing that because i ran into a lot of type errors when beginning using tl::expected so i wanted to take it slow. I'm fixing it now

gcc/rust/expand/rust-macro-builtins-asm.cc
badumbatish: fixing this now

badumbatish: ok i see now it's not possible to do that, idk if it's because it's C++11 or sth
Screenshot 2024-06-26 at 9 37 45 AM

CohenArthur: uh, weird, I tried it out yesterday when reviewing the PR and didn't have any errors! but that's okay let's leave it with the template argument

CohenArthur (approved):

CohenArthur:
oh right btw @badumbatish thanks to @P-E-P's comment on the other PR, I notice that your commits aren't signed - could you rebase them and sign all of them?

badumbatish:
> oh right btw @badumbatish thanks to @P-E-P's comment on the other PR, I notice that your commits aren't signed - could you rebase them and sign all of them?

hmm thats weird, yes let me rebase and sign them :) thanks for the heads up

badumbatish:
@CohenArthur everything should be signed off now :)

PR #3056, opened 2024-06-17
gccrs: configure.ac: Fix quoting around ldl/lpthread checks

CohenArthur:
ChangeLog:

	* configure.ac: Add quoting around both variable checks performed, which
	will fix the noise reported when libc contains both ldl and lpthread. A
	typo around `pthread_create` being typed `pthread_crate` is also fixed.
	* configure: Regenerate.

Addresses [Bugzilla report #115453](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=115453)

### DIFF
--- a/configure
+++ b/configure
@@ -8827,7 +8827,7 @@ fi
 
 
 
-# Rust requires -ldl and -lpthread if you are using an old glibc that does not include them by
+# Rust requires -ldl and -lpthread if you are using a libc which does not include them by
 # default, so we check for them here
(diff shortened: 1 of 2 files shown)


CohenArthur:
will merge once reviewed upstream, I also submitted the patch there

CohenArthur:
This has been fixed upstream by a different set of people, so we'll just need to update our fork for the changes to be reflected. Closing

PR #3059, opened 2024-06-22
Add test case for using asm! outside of unsafe {}

badumbatish:
I've added the unsafe guard check in #2982 but I haven't added it in since then


CohenArthur (approved):
LGTM!

PR #3074, opened 2024-07-01
Safe-guard InlineAsm structs

### LINKED ISSUE #3050: Safe guard InlineAsm-related structs
The current InlineAsm structs have a lot of constructors where we permit the intialization of nullptr and then we would have to test to see if the pointer is nullptr before performing clone_expr (and not to mention other pointer related operation)

Please preview Arthur's review in this PR.

> if `expr` cannot be null then it would be nice to add more assertions to this struct (but in another PR)

1. remove the empty constructor, as that will set `expr` to `nullptr` and thus create an invalid `InOut`
2. in the full constructor on line 4850, `rust_assert(expr)` to ensure that `expr` is not null
3. because `expr` is not null, remove the `if (other.expr)` checks in both the copy constructor and `operator=` overload

_Originally posted by @CohenArthur in https://github.com/Rust-GCC/gccrs/pull/2982#discussion_r1636731459_

badumbatish:
Fixed #3050 by safeguarding constructors and using tl::optional in member of first layer's construction of InlineAsm

### DIFF
--- a/gcc/rust/expand/rust-macro-builtins-asm.cc
+++ b/gcc/rust/expand/rust-macro-builtins-asm.cc
@@ -307,9 +307,9 @@ parse_reg_operand_in (InlineAsmContext inline_asm_ctx)
 
       // TODO: When we've succesfully parse an expr, remember to clone_expr()
       // instead of nullptr
-      struct AST::InlineAsmOperand::In in (reg, nullptr);
-      reg_operand.set_in (in);
-      inline_asm_ctx.inline_asm.operands.push_back (reg_operand);
+      // struct AST::InlineAsmOperand::In in (reg, nullptr);
+      // reg_operand.set_in (in);
+      // inline_asm_ctx.inline_asm.operands.push_back (reg_operand);
       return inline_asm_ctx;
     }
   return tl::unexpected<InlineAsmParseError> (NONCOMMITED);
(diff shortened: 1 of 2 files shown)


badumbatish:
force-pushing again since i seemed to have forgotten to sign off again ...

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -307,9 +307,9 @@ parse_reg_operand_in (InlineAsmContext inline_asm_ctx)
 
       // TODO: When we've succesfully parse an expr, remember to clone_expr()
       // instead of nullptr
-      struct AST::InlineAsmOperand::In in (reg, nullptr);
-      reg_operand.set_in (in);
-      inline_asm_ctx.inline_asm.operands.push_back (reg_operand);
+      // struct AST::InlineAsmOperand::In in (reg, nullptr);
+      // reg_operand.set_in (in);
+      // inline_asm_ctx.inline_asm.operands.push_back (reg_operand);
```
CohenArthur, line 312: so we no longer create any of those structs? is that intentional?

badumbatish: in the original code, I created those structs there regardless if i managed to parse expr successfully or not (hence me passing in the nullptr). With the new safeguard, these line will trigger rust_asserts() on nullptr (which is what we want since we don't want to pass around nullptr)

The construction of these structs doesn't matter for now since there isn't any information being passed into the structs (we parse expr via parse_format_string() for now)

CohenArthur: that makes sense, thanks for the explanation!

PR #3078, opened 2024-07-07
#[may_dangle] in safe impl

### LINKED ISSUE #3045: Prevent `#[may_dangle]` attribute from being used in safe impl
`#[may_dangle]` attributes should not be used from safe impl.

https://godbolt.org/z/6eKe5sKdo

liamnaddell:
```
    gcc/rust/ChangeLog:
            * ast/rust-ast.cc:
            Fix Attribute constructors to copy inner_attribute
            * checks/errors/rust-unsafe-checker.cc:
            Add pass for #[may_dangle] in safe impl's
            * hir/rust-ast-lower-item.cc:
            Add support for unsafe impl's
            * hir/rust-ast-lower-type.cc:
            Lower attributes in impl's from AST to HIR
            * hir/rust-hir-dump.cc:
            Change single attribute to AttrVec
            * hir/tree/rust-hir-item.h:
            Add unsafe support to Impl blocks in HIR
            * hir/tree/rust-hir.cc:
            Change single attribute to AttrVec
            * hir/tree/rust-hir.h:
            Add has/get_outer_attribute to GenericParam

    gcc/testsuite/ChangeLog:
            * rust/compile/issue-3045-1.rs:
            Add test for #[may_dangle] Generic Type triggering error
            * rust/compile/issue-3045-2.rs:
            Add test for #[may_dangle] Lifetime triggering error
```

Fixes #3045 

- \[x] Run `clang-format`

---

Added support for rejecting safe Impl's that use #[may_dangle] for either lifetime parameters or generic type parameters

### DIFF
--- a/gcc/rust/checks/errors/rust-unsafe-checker.cc
+++ b/gcc/rust/checks/errors/rust-unsafe-checker.cc
@@ -784,7 +784,23 @@ UnsafeChecker::visit (Trait &trait)
 void
 UnsafeChecker::visit (ImplBlock &impl)
 {
-  // FIXME: Handle unsafe impls
+  bool safe = !impl.is_unsafe ();
+  // Check for unsafe-only attributes on generics and lifetimes
+  if (safe)
+    for (auto &parm : impl.get_generic_params ())
+      {
+	for (auto o_attr : parm->get_outer_attrs ())
+	  {
+	    rust_assert (!o_attr.is_inner_attribute ());
+
+	    Rust::AST::SimplePath path = o_attr.get_path ();
+	    if (path == Values::Attributes::MAY_DANGLE)
+	      rust_error_at (
+		o_attr.get_locus (), ErrorCode::E0569,
+		"use of %<may_dangle%> is unsafe and requires unsafe impl");
+	  }
+      }
+
   for (auto &item : impl.get_impl_items ())
     item->accept_vis (*this);
 }
--- a/gcc/rust/hir/rust-ast-lower-type.cc
+++ b/gcc/rust/hir/rust-ast-lower-type.cc
@@ -17,6 +17,7 @@
 // <http://www.gnu.org/licenses/>.
 
 #include "rust-ast-lower-type.h"
+#include "rust-attribute-values.h"
 
 namespace Rust {
 namespace HIR {
@@ -446,16 +447,17 @@ void
 ASTLowerGenericParam::visit (AST::LifetimeParam &param)
 {
   auto crate_num = mappings.get_current_crate ();
+  AST::Lifetime lifetime = param.get_lifetime ();
   Analysis::NodeMapping mapping (crate_num, param.get_node_id (),
 				 mappings.get_next_hir_id (crate_num),
 				 mappings.get_next_localdef_id (crate_num));
 
-  HIR::Lifetime lt (mapping, param.get_lifetime ().get_lifetime_type (),
-		    param.get_lifetime ().get_lifetime_name (),
-		    param.get_lifetime ().get_locus ());
+  HIR::Lifetime lt (mapping, lifetime.get_lifetime_type (),
+		    lifetime.get_lifetime_name (), lifetime.get_locus ());
 
   translated = new HIR::LifetimeParam (mapping, lt, param.get_locus (),
-				       std::vector<Lifetime> ());
+				       std::vector<Lifetime> (),
+				       param.get_outer_attrs ());
 }
 
 void
@@ -482,7 +484,6 @@ ASTLowerGenericParam::visit (AST::ConstGenericParam &param)
 void
 ASTLowerGenericParam::visit (AST::TypeParam &param)
 {
-  AST::Attribute outer_attr = AST::Attribute::create_empty ();
   std::vector<std::unique_ptr<HIR::TypeParamBound>> type_param_bounds;
   if (param.has_type_param_bounds ())
     {
@@ -506,7 +507,8 @@ ASTLowerGenericParam::visit (AST::TypeParam &param)
   translated
     = new HIR::TypeParam (mapping, param.get_type_representation (),
 			  param.get_locus (), std::move (type_param_bounds),
-			  std::unique_ptr<Type> (type), std::move (outer_attr));
+			  std::unique_ptr<Type> (type),
+			  param.get_outer_attrs ());
 }
 
 HIR::TypeParamBound *
--- a/gcc/rust/hir/rust-hir-dump.cc
+++ b/gcc/rust/hir/rust-hir-dump.cc
@@ -1508,7 +1508,8 @@ void
 Dump::visit (TypeParam &e)
 {
   begin ("TypeParam");
-  put_field ("outer_attr", e.get_outer_attribute ().as_string ());
+  auto &outer_attrs = e.get_outer_attrs ();
+  do_outer_attrs (outer_attrs);
 
   put_field ("type_representation", e.get_type_representation ().as_string ());
 
--- a/gcc/rust/hir/tree/rust-hir-item.h
+++ b/gcc/rust/hir/tree/rust-hir-item.h
@@ -35,9 +35,7 @@ class TypePath;
 // A type generic parameter (as opposed to a lifetime generic parameter)
 class TypeParam : public GenericParam
 {
-  // bool has_outer_attribute;
-  // std::unique_ptr<Attribute> outer_attr;
-  AST::Attribute outer_attr;
+  AST::AttrVec outer_attrs;
 
   Identifier type_representation;
 
@@ -59,16 +57,16 @@ public:
   bool has_type_param_bounds () const { return !type_param_bounds.empty (); }
 
   // Returns whether the type param has an outer attribute.
-  bool has_outer_attribute () const { return !outer_attr.is_empty (); }
-  AST::Attribute &get_outer_attribute () { return outer_attr; }
+  bool has_outer_attribute () const override { return outer_attrs.size () > 0; }
+  AST::AttrVec &get_outer_attrs () override { return outer_attrs; }
 
   TypeParam (Analysis::NodeMapping mappings, Identifier type_representation,
 	     location_t locus = UNDEF_LOCATION,
 	     std::vector<std::unique_ptr<TypeParamBound>> type_param_bounds
 	     = std::vector<std::unique_ptr<TypeParamBound>> (),
 	     std::unique_ptr<Type> type = nullptr,
-	     AST::Attribute outer_attr = AST::Attribute::create_empty ())
-    : GenericParam (mappings), outer_attr (std::move (outer_attr)),
+	     AST::AttrVec outer_attrs = std::vector<AST::Attribute> ())
+    : GenericParam (mappings), outer_attrs (std::move (outer_attrs)),
       type_representation (std::move (type_representation)),
       type_param_bounds (std::move (type_param_bounds)),
       type (std::move (type)), locus (locus)
@@ -76,7 +74,7 @@ public:
 
   // Copy constructor uses clone
   TypeParam (TypeParam const &other)
-    : GenericParam (other.mappings), outer_attr (other.outer_attr),
+    : GenericParam (other.mappings), outer_attrs (other.outer_attrs),
       type_representation (other.type_representation), locus (other.locus)
   {
     // guard to prevent null pointer dereference
@@ -92,7 +90,7 @@ public:
   TypeParam &operator= (TypeParam const &other)
   {
     type_representation = other.type_representation;
-    outer_attr = other.outer_attr;
+    outer_attrs = other.outer_attrs;
     locus = other.locus;
     mappings = other.mappings;
 
@@ -2741,6 +2739,7 @@ class ImplBlock : public VisItem, public WithInnerAttrs
   BoundPolarity polarity;
   location_t locus;
   std::vector<std::unique_ptr<ImplItem>> impl_items;
+  bool unsafe;
 
 public:
   ImplBlock (Analysis::NodeMapping mappings,
@@ -2749,20 +2748,20 @@ public:
 	     std::unique_ptr<Type> impl_type,
 	     std::unique_ptr<TypePath> trait_ref, WhereClause where_clause,
 	     BoundPolarity polarity, Visibility vis, AST::AttrVec inner_attrs,
-	     AST::AttrVec outer_attrs, location_t locus)
+	     AST::AttrVec outer_attrs, location_t locus, bool unsafe = false)
     : VisItem (std::move (mappings), std::move (vis), std::move (outer_attrs)),
       WithInnerAttrs (std::move (inner_attrs)),
       generic_params (std::move (generic_params)),
       impl_type (std::move (impl_type)), trait_ref (std::move (trait_ref)),
       where_clause (std::move (where_clause)), polarity (polarity),
-      locus (locus), impl_items (std::move (impl_items))
+      locus (locus), impl_items (std::move (impl_items)), unsafe (unsafe)
   {}
 
   ImplBlock (ImplBlock const &other)
     : VisItem (other), WithInnerAttrs (other.inner_attrs),
       impl_type (other.impl_type->clone_type ()),
       where_clause (other.where_clause), polarity (other.polarity),
-      locus (other.locus)
+      locus (other.locus), unsafe (other.unsafe)
   {
     generic_params.reserve (other.generic_params.size ());
     for (const auto &e : other.generic_params)
@@ -2781,6 +2780,7 @@ public:
     polarity = other.polarity;
     inner_attrs = other.inner_attrs;
     locus = other.locus;
+    unsafe = other.unsafe;
 
     generic_params.reserve (other.generic_params.size ());
     for (const auto &e : other.generic_params)
@@ -2800,6 +2800,8 @@ public:
 
   // Returns whether inherent impl block has inherent impl items.
   bool has_impl_items () const { return !impl_items.empty (); }
+
+  bool is_unsafe () const { return unsafe; }
 
   void accept_vis (HIRFullVisitor &vis) override;
   void accept_vis (HIRStmtVisitor &vis) override;
--- a/gcc/rust/ast/rust-ast.cc
+++ b/gcc/rust/ast/rust-ast.cc
@@ -311,7 +311,8 @@ Attribute::get_traits_to_derive ()
 
 // Copy constructor must deep copy attr_input as unique pointer
 Attribute::Attribute (Attribute const &other)
-  : path (other.path), locus (other.locus)
+  : path (other.path), locus (other.locus),
+    inner_attribute (other.inner_attribute)
 {
   // guard to protect from null pointer dereference
   if (other.attr_input != nullptr)
@@ -324,6 +325,7 @@ Attribute::operator= (Attribute const &other)
 {
   path = other.path;
   locus = other.locus;
+  inner_attribute = other.inner_attribute;
   // guard to protect from null pointer dereference
   if (other.attr_input != nullptr)
     attr_input = other.attr_input->clone_attr_input ();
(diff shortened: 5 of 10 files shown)


gcc/rust/checks/errors/rust-unsafe-checker.cc
```diff
@@ -784,7 +784,25 @@ UnsafeChecker::visit (Trait &trait)
 void
 UnsafeChecker::visit (ImplBlock &impl)
 {
-  // FIXME: Handle unsafe impls
+  bool safe = !impl.is_unsafe ();
```
badumbatish, line 787: I'm not sure if this is relevant, i might be talking nonsense, but i think there is something in the Unsafe class to detect unsafe context for you. You can use it like this to avoid populuting every new class that needs to be check for unsafe usage with a boolean unsafe

```
  if (unsafe_context.is_in_context ()) { /* code */ }

File: gcc/rust/hir/tree/rust-hir-item.h [hunk 2]
```diff
@@ -2801,6 +2802,8 @@ class ImplBlock : public VisItem, public WithInnerAttrs
   // Returns whether inherent impl block has inherent impl items.
   bool has_impl_items () const { return !impl_items.empty (); }
 
+  bool is_unsafe () const { return unsafe; }
```
badumbatish, line 2805: might not be necesary due to `if (unsafe_context.is_in_context ())` ?

badumbatish:
hi! i left some reviews :)

gcc/rust/checks/errors/rust-unsafe-checker.cc
liamnaddell: These functions do exist, but I think we are in a different sort of "unsafe context" here that needs to be handled separately. an "unsafe impl" is used when implementing "unsafe traits", the functions inside the unsafe impl can be safe, and need "unsafe" contexts of their own to implement unsafe behavior.

From https://doc.rust-lang.org/reference/unsafe-keyword.html , "An unsafe trait is a trait that comes with extra safety conditions that must be upheld by implementations of the trait. The unsafe trait should come with documentation explaining what those extra safety conditions are. Such a trait must be prefixed with the keyword unsafe and can only be implemented by unsafe impl blocks."

Originally I tried doing what you suggested, however, the issue I ran into is that by entering the unsafe context, unsafe functions in unsafe impl's would no longer be required, which I think is out-of-spec.

As a final note, currently we don't enter an unsafe context when parsing unsafe impl's. I believe the "unsafe" token was simply ignored before I made changes to gcc/rust/hir/rust-ast-lower-item.cc.

gcc/rust/hir/tree/rust-hir-item.h
liamnaddell: I think the same comment left https://github.com/Rust-GCC/gccrs/pull/3078#discussion_r1669367590 would apply to this as well

liamnaddell:
> hi! i left some reviews :)

Thank you!

gcc/rust/checks/errors/rust-unsafe-checker.cc
badumbatish: oh wow i didn't know there were multiple usage of unsafe, usually i thought there's only one way to use which is unsafe {}, thank you for your explanation

badumbatish:
Hi Liam, the reviews will probably come next week, hope you got a great weekend:)

gcc/rust/checks/errors/rust-unsafe-checker.cc
badumbatish: @liamnaddell i'm curious, how did you find out if your `unsafe` boolean will be turned to true if there is an unsafe word before impl? Is a parse somewhere changing it ?

badumbatish:
i left some reviews but i think i might need another pair of eyes

gcc/rust/hir/tree/rust-hir-item.h
```diff
@@ -2749,13 +2750,13 @@ class ImplBlock : public VisItem, public WithInnerAttrs
...
     : VisItem (std::move (mappings), std::move (vis), std::move (outer_attrs)),
       WithInnerAttrs (std::move (inner_attrs)),
       generic_params (std::move (generic_params)),
       impl_type (std::move (impl_type)), trait_ref (std::move (trait_ref)),
       where_clause (std::move (where_clause)), polarity (polarity),
-      locus (locus), impl_items (std::move (impl_items))
+      locus (locus), impl_items (std::move (impl_items)), unsafe (unsafe)
   {}
 
   ImplBlock (ImplBlock const &other)
```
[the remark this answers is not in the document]
liamnaddell: I'm not quite sure what you are asking, does this answer your question?
```
Breakpoint 2, Rust::HIR::ImplBlock::ImplBlock (this=0x4fe1bb0, mappings=..., impl_items=..., generic_params=...,
    impl_type=..., trait_ref=..., where_clause=..., polarity=Rust::NegativeBound, vis=..., inner_attrs=..., outer_attrs=...,
    locus=67301, unsafe=true) at ../../gccrs/gcc/rust/hir/tree/rust-hir-item.h:2747
2747      ImplBlock (Analysis::NodeMapping mappings,
(gdb) n
2759          locus (locus), impl_items (std::move (impl_items)), unsafe (unsafe)
(gdb) p unsafe
$19 = true
(gdb) p this.unsafe
$20 = false
(gdb) n
2754        : VisItem (std::move (mappings), std::move (vis), std::move (outer_attrs)),
(gdb)
2759          locus (locus), impl_items (std::move (impl_items)), unsafe (unsafe)
(gdb)
2754        : VisItem (std::move (mappings), std::move (vis), std::move (outer_attrs)),
(gdb)
2759          locus (locus), impl_items (std::move (impl_items)), unsafe (unsafe)
(gdb)
2755          WithInnerAttrs (std::move (inner_attrs)),
(gdb)
2759          locus (locus), impl_items (std::move (impl_items)), unsafe (unsafe)
(gdb)
2756          generic_params (std::move (generic_params)),
(gdb)
2757          impl_type (std::move (impl_type)), trait_ref (std::move (trait_ref)),
(gdb)
2758          where_clause (std::move (where_clause)), polarity (polarity),
(gdb)
2759          locus (locus), impl_items (std::move (impl_items)), unsafe (unsafe)
(gdb)
2760      {}
(gdb) p this.unsafe
$21 = true
(gdb)
```

gcc/rust/hir/rust-ast-lower-type.cc
```diff
@@ -445,17 +446,28 @@ ASTLowerGenericParam::translate (AST::GenericParam &param)
...
+  for (auto o_attr : param.get_outer_attrs ())
+    {
+      rust_assert (!o_attr.is_inner_attribute ());
+
+      Rust::AST::SimplePath path = o_attr.get_path ();
+      if (path == Values::Attributes::MAY_DANGLE)
+	outer_attr = o_attr;
+    }
+
   HIR::Lifetime lt (mapping, param.get_lifetime ().get_lifetime_type (),
```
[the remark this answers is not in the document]
liamnaddell: Sure, I can make this change while I'm modifying this section.

gcc/rust/hir/tree/rust-hir-item.h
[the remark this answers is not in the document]
liamnaddell: This was a mistake. I fixed the copy constructor to copy unsafe, the assignment operator. The move constructor is default so I don't think there's anything to change there

liamnaddell:
@badumbatish On the same note as the "unsafe constructor" issue, it appears a similar issue was in the codebase introduced by https://github.com/Rust-GCC/gccrs/commit/e44f127cdb12a28536fe21983dfad20570bceda0 . When I set inner_attribute to false in rust-ast.h, I set it because I was observing UB (I think?) with inner_attribute being set to bogus values like "254".

Does this look like the same mistake as the one I made, or is there something different in this case?

gcc/rust/checks/errors/rust-unsafe-checker.cc
liamnaddell: I think the AST already had a get_unsafe (gcc/rust/hir/rust-ast-lower-item.cc:626). Then I checked using GDB and using the two tests I attached to the review

gcc/rust/ast/rust-ast.h
```diff
@@ -634,7 +634,7 @@ struct Attribute
 
   location_t locus;
 
-  bool inner_attribute;
+  bool inner_attribute = false;
```
CohenArthur, line 637: I don't think this is required, since the full constructor of that class has a default parameter for `inner_attribute` which sets the field to `false`

gcc/rust/checks/errors/rust-unsafe-checker.cc
```diff
@@ -784,7 +784,25 @@ UnsafeChecker::visit (Trait &trait)
...
+
+	AST::Attribute &attr = parm->get_outer_attribute ();
+
+	AST::SimplePath path = attr.get_path ();
+
+	if (path != Values::Attributes::MAY_DANGLE)
+	  continue;
+
+	rust_error_at (attr.get_locus (), ErrorCode::E0133,
+		       "use of %<may_dangle%> unsafe and requires unsafe impl");
```
CohenArthur, line 803: I would switch the logic around to better show the intent, but this is completely valid :)

```suggestion
	if (path == Values::Attributes::MAY_DANGLE)
	rust_error_at (attr.get_locus (), ErrorCode::E0133,
		       "use of %<may_dangle%> is unsafe and requires unsafe impl");
```

gcc/rust/hir/rust-ast-lower-type.cc
```diff
@@ -445,17 +446,28 @@ ASTLowerGenericParam::translate (AST::GenericParam &param)
...
+    }
+
+  HIR::Lifetime lt (mapping, lifetime.get_lifetime_type (),
+		    lifetime.get_lifetime_name (), lifetime.get_locus ());
 
-  translated = new HIR::LifetimeParam (mapping, lt, param.get_locus (),
-				       std::vector<Lifetime> ());
+  translated
+    = new HIR::LifetimeParam (mapping, lt, param.get_locus (),
+			      std::vector<Lifetime> (), std::move (outer_attr));
```
CohenArthur, line 470: hm, I'm not sure I understand the purpose of this code. we only set an outer attribute to lifetime parameters if they have a `#[may_dangle]` one? can they not have multiple attributes? I don't know exactly how attributes work on lifetimes so I'm genuinely curious haha

CohenArthur (commented):
overall this looks like great work and it's a nice implementation. just a few nits on my part but that PR is close to being merged :)

badumbatish:
> @badumbatish On the same note as the "unsafe constructor" issue, it appears a similar issue was in the codebase introduced by [e44f127](https://github.com/Rust-GCC/gccrs/commit/e44f127cdb12a28536fe21983dfad20570bceda0) . When I set inner_attribute to false in rust-ast.h, I set it because I was observing UB (I think?) with inner_attribute being set to bogus values like "254".
> 
> Does this look like the same mistake as the one I made, or is there something different in this case? If you agree it's an issue, I have a fix locally I can push to this PR
> 
> (Also, thanks again for the reviews!)

hmm i think if

> @badumbatish On the same note as the "unsafe constructor" issue, it appears a similar issue was in the codebase introduced by [e44f127](https://github.com/Rust-GCC/gccrs/commit/e44f127cdb12a28536fe21983dfad20570bceda0) . When I set inner_attribute to false in rust-ast.h, I set it because I was observing UB (I think?) with inner_attribute being set to bogus values like "254".
> 
> Does this look like the same mistake as the one I made, or is there something different in this case? If you agree it's an issue, I have a fix locally I can push to this PR
> 
> (Also, thanks again for the reviews!)

i think it should be fine if it's not set since the member attribute inner_attribute will be set through the constructor and we don't have a default constructor generated from the compiler (since we defined our own constructor already). The garbage value you see might have been because if we don't specify a default value like `bool inner_attributes = false` then it just uses whatever value is at that stack pointer, 254 > 0 so `inner_attributes` accidentally turns to true until the constructor does its job

gcc/rust/ast/rust-ast.h
liamnaddell: Yea, you are correct here, I think I got spooked looking at -O2 code in GDB

gcc/rust/hir/rust-ast-lower-type.cc
liamnaddell: This was because of a mismatch between AST and HIR. AST has an AttrVec, but HIR has:
```
651 // A lifetime generic parameter (as opposed to a type generic parameter)
652 class LifetimeParam : public GenericParam
653 {
654   Lifetime lifetime;
655
656   // bool has_lifetime_bounds;
657   // LifetimeBounds lifetime_bounds;
658   std::vector<Lifetime> lifetime_bounds; // inlined LifetimeBounds
659
660   // bool has_outer_attribute;
661   // std::unique_ptr<Attribute> outer_attr;
662   AST::Attribute outer_attr;
```

I have absolutely no idea whether a lifetime (or type param) can have multiple attributes. If I had to guess, the answer is yes, but I can't find an attribute other than may_dangle to test it out on. On another note, it seems like rust is doing the pass for may_dangle on the AST instead of in HIR.

liamnaddell: The best evidence I have that rustc supports multiple attributes is this:
```
liam@gentoo ~/dangle $ rustc  main2.rs
error[E0517]: attribute should be applied to a struct, enum, or union
 --> main2.rs:7:20
  |
7 | unsafe impl<#[repr(C)] #[may_dangle] 'a> Action for Inspector2<'a> {}
  |                    ^                 -- not a struct, enum, or union

error: aborting due to 1 previous error
```
Note, it thinks repr(C) is applied to 'a not #[may_dangle])

gcc/rust/ast/rust-ast.h
liamnaddell: Nevermind, I undid this change, and now I get an ICE because inner_attribute is set to true on an outer attribute.

liamnaddell:
```c++
diff --git a/gcc/rust/ast/rust-ast.cc b/gcc/rust/ast/rust-ast.cc
index 9c0b1bf8245..f4732bfd1fd 100644
--- a/gcc/rust/ast/rust-ast.cc
+++ b/gcc/rust/ast/rust-ast.cc
@@ -311,7 +311,7 @@ Attribute::get_traits_to_derive ()

 // Copy constructor must deep copy attr_input as unique pointer
 Attribute::Attribute (Attribute const &other)
-  : path (other.path), locus (other.locus)
+  : path (other.path), locus (other.locus), inner_attribute (other.inner_attribute)
 {
   // guard to protect from null pointer dereference
   if (other.attr_input != nullptr)
@@ -324,6 +324,7 @@ Attribute::operator= (Attribute const &other)
 {
   path = other.path;
   locus = other.locus;
+  inner_attribute = other.inner_attribute;
   // guard to protect from null pointer dereference
   if (other.attr_input != nullptr)
     attr_input = other.attr_input->clone_attr_input ();
     ```

Comment (CohenArthur) [reply gcc/rust/ast/rust-ast.h]: ah yeah didn't catch that... that was indeed missing from the copy constructors. this looks very good and correct, I'd like you to use it yeah :D thanks!

Comment (liamnaddell) [reply gcc/rust/ast/rust-ast.h]: I pushed this patch, the testing should be fixed now

File: gcc/rust/hir/rust-ast-lower-type.cc (hunk 7 above)
Comment (liamnaddell) [reply gcc/rust/hir/rust-ast-lower-type.cc:470]: I'll write up a patch for this soon, it's a reasonable change to make.

Comment (liamnaddell) [reply gcc/rust/hir/rust-ast-lower-type.cc:470]: Fixing the AttrVec stuff was somewhat involved. I've attached a changeset to make it easier to review

Comment (liamnaddell) [reply gcc/rust/hir/rust-ast-lower-type.cc:470]: [0001-squash-and-post-later.patch](https://github.com/user-attachments/files/16312084/0001-squash-and-post-later.patch)

File: gcc/rust/hir/rust-hir-dump.cc [hunk 8]
```diff
@@ -1508,7 +1508,8 @@ void
 Dump::visit (TypeParam &e)
 {
   begin ("TypeParam");
-  put_field ("outer_attr", e.get_outer_attribute ().as_string ());
+  auto oa = e.get_outer_attrs ();
+  do_outer_attrs (oa);
```
CohenArthur, line 1512:
```suggestion
  auto &outer_attrs = e.get_outer_attrs ();
  do_outer_attrs (outer_attrs);
```

CohenArthur: I think the extra `&` is necessary to avoid making a copy, not that it really matters here

gcc/rust/checks/errors/rust-unsafe-checker.cc
```diff
@@ -784,7 +784,23 @@ UnsafeChecker::visit (Trait &trait)
 void
 UnsafeChecker::visit (ImplBlock &impl)
 {
-  // FIXME: Handle unsafe impls
+  bool safe = !impl.is_unsafe ();
+  // Check for unsafe-only attributes on generics and lifetimes
+  if (safe)
+    for (auto &parm : impl.get_generic_params ())
+      {
+	for (auto o_attr : (*parm).get_outer_attrs ())
```
CohenArthur, line 792:
```suggestion
	for (auto o_attr : parm->get_outer_attrs ())
```

any reason for the deref + method call instead of this?

gcc/rust/checks/errors/rust-unsafe-checker.cc
```diff
@@ -784,7 +784,23 @@ UnsafeChecker::visit (Trait &trait)
...
+	  {
+	    rust_assert (!o_attr.is_inner_attribute ());
+
+	    Rust::AST::SimplePath path = o_attr.get_path ();
+	    if (path == Values::Attributes::MAY_DANGLE)
+	      rust_error_at (
+		o_attr.get_locus (), ErrorCode::E0569,
+		"use of %<may_dangle%> unsafe and requires unsafe impl");
```
CohenArthur, line 800:
```suggestion
		"use of %<may_dangle%> is unsafe and requires unsafe impl");
```

gcc/rust/hir/rust-ast-lower-type.cc
```diff
@@ -445,17 +446,19 @@ ASTLowerGenericParam::translate (AST::GenericParam &param)
 void
 ASTLowerGenericParam::visit (AST::LifetimeParam &param)
 {
+  AST::Attribute outer_attr = AST::Attribute::create_empty ();
```
CohenArthur, line 449: I think we can remove this since we are not using it? or maybe I don't see the use

CohenArthur (approved):
great work, thank you! a couple nits and we can merge this :D well done!

gcc/rust/hir/rust-ast-lower-type.cc
liamnaddell: Fixed,

I thought the compiler would emit warnings for this, no?

gcc/rust/checks/errors/rust-unsafe-checker.cc
liamnaddell: Fixed

gcc/rust/checks/errors/rust-unsafe-checker.cc
liamnaddell: Fixed

gcc/rust/hir/rust-hir-dump.cc
liamnaddell: Good catch, I think this was a holdover from before we were copying vectors. I fixed this

P-E-P (approved):

gcc/rust/hir/rust-ast-lower-type.cc
P-E-P: I think it does not emit a warning because the constructor has a side effect, it increments the node id.

PR #3085, opened 2024-07-18
Fix CI bootstrap build with a nightly rust compiler

P-E-P:
The CI failed to compile some new code that requires a nightly feature. This change allows the CI to use a nightly compiler instead.


CohenArthur (approved):
thanks!

PR #3088, opened 2024-07-19
Better loan errors

braw-lee:
depends on #3013

### DIFF
--- a/gcc/rust/checks/errors/borrowck/rust-bir.h
+++ b/gcc/rust/checks/errors/borrowck/rust-bir.h
@@ -78,8 +78,8 @@ private:
   std::unique_ptr<AbstractExpr> expr;
   TyTy::BaseType *type;
   // stores location of the actual expression from source code
-  // currently only available when kind == Kind::ASSIGNMENT
-  // FIXME: Add location for Statements other than ASSIGNMENT
+  // currently only available when kind is ASSIGNMENT | RETURN
+  // FIXME: Add location for other statement kinds
   location_t location;
 
 public:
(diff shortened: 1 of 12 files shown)


gcc/rust/checks/errors/borrowck/rust-bir.h
```diff
@@ -94,6 +94,7 @@ class Statement
   {}
 
   explicit Statement (Kind kind, PlaceId place, TyTy::BaseType *type,
+		      AbstractExpr *expr = nullptr,
 		      location_t location = UNKNOWN_LOCATION)
```
CohenArthur: try and check if putting the `location` optional parameter before `*type` and before `*expr` helps keeping the instantiations simpler

CohenArthur (approved):

gcc/rust/checks/errors/borrowck/rust-bir.h
braw-lee: made a helper function, `push_return` [here](https://github.com/Rust-GCC/gccrs/pull/3088/commits/24070d2e884b568b7cfd662e76a01d41af631822#diff-33f99b533fa23e311e13441446152c7001fbd17d8579a846bfe67a022586cab0R306-R310)
we still have to pass INVALID_PLACE as C++ cannot differ between location_t and PlaceId as both are of `unsigned int`, but we do this at once place only i.e this helper function

to get rid of INVALID_PLACE, we need these 2 constructors
```cpp
Statement (Kind kind, PlaceId place);
Statement (Kind king, location_t location);
```
which is possible if PlaceId is strongly typed, we can do this in future PRs

let me know what do you think about this

PR #3090, opened 2024-07-20
Added options for ParseMode

badumbatish:
gcc/rust/ChangeLog:

	* ast/rust-fmt.cc (Pieces::collect): Added options for ParseMode
	* ast/rust-fmt.h (collect_pieces): Likewise. (struct Pieces): Likewise.
	* expand/rust-macro-builtins-format-args.cc (MacroBuiltin::format_args_handler): Likewise.

libgrust/ChangeLog:

	* libformat_parser/generic_format_parser/src/lib.rs: Likewise.
	* libformat_parser/src/bin.rs: Likewise.
	* libformat_parser/src/lib.rs: Likewise.

### DIFF
--- a/gcc/rust/ast/rust-fmt.h
+++ b/gcc/rust/ast/rust-fmt.h
@@ -258,10 +258,16 @@ struct FormatArgsHandle
   RustString rust_string;
 };
 
+typedef enum
+{
+  Format,
+  InlineAsm,
+} ParseMode;
+
 extern "C" {
 
 FormatArgsHandle
-collect_pieces (const char *input, bool append_newline);
+collect_pieces (const char *input, bool append_newline, ParseMode parse_mode);
 
 FormatArgsHandle
 clone_pieces (const FormatArgsHandle &);
@@ -274,7 +280,8 @@ void destroy_pieces (FormatArgsHandle);
 
 struct Pieces
 {
-  static Pieces collect (const std::string &to_parse, bool append_newline);
+  static Pieces collect (const std::string &to_parse, bool append_newline,
+			 ffi::ParseMode parse_mode);
   ~Pieces ();
 
   Pieces (const Pieces &other);
--- a/gcc/rust/ast/rust-fmt.h
+++ b/gcc/rust/ast/rust-fmt.h
@@ -258,11 +258,11 @@ struct FormatArgsHandle
   RustString rust_string;
 };
 
-typedef enum
+enum ParseMode
 {
-  Format,
+  Format = 0,
   InlineAsm,
-} ParseMode;
+};
 
 extern "C" {
 
--- a/gcc/rust/expand/rust-macro-builtins-format-args.cc
+++ b/gcc/rust/expand/rust-macro-builtins-format-args.cc
@@ -16,6 +16,7 @@
 // along with GCC; see the file COPYING3.  If not see
 // <http://www.gnu.org/licenses/>.
 #include "rust-ast-fragment.h"
+#include "rust-fmt.h"
 #include "rust-macro-builtins-helpers.h"
 #include "rust-expand-format-args.h"
 
@@ -162,7 +163,8 @@ MacroBuiltin::format_args_handler (location_t invoc_locus,
   if (append_newline)
     fmt_str += '\n';
 
-  auto pieces = Fmt::Pieces::collect (fmt_str, append_newline);
+  auto pieces = Fmt::Pieces::collect (fmt_str, append_newline,
+				      Fmt::ffi::ParseMode::Format);
 
   // TODO:
   // do the transformation into an AST::FormatArgs node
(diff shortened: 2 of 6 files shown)


badumbatish:
@CohenArthur  hi Arthur, when i refactored the code (in rust) and ran the tests, it didn't have any errors. I was wondering if we have any tests for these format-args?

gcc/rust/ast/rust-fmt.h
```diff
@@ -258,10 +258,16 @@ struct FormatArgsHandle
   RustString rust_string;
 };
 
+typedef enum
+{
+  Format,
+  InlineAsm,
+} ParseMode;
```
CohenArthur, line 265:
```suggestion
enum ParseMode
{
  Format,
  InlineAsm,
};
```

there's no need to typedef it in C++ for an enum

libgrust/libformat_parser/generic_format_parser/src/lib.rs
```diff
@@ -78,6 +78,7 @@ enum InputStringKind {
 }
 
 /// The type of format string that we are parsing.
+#[repr(C)]
```
CohenArthur, line 81: I'd like to not change this, because ideally this should be a straight up copy of rustc's `rustc_format_parser` crate. I think the best thing would be to add an extra enum on our side of the Rust crate, e.g. in `src/lib.rs` directly, which is `#[repr(C)]` and whose initial value is `= 0`. then, add a quick conversion function to `generic_format_parser::ParseMode`

gcc/rust/ast/rust-fmt.h
CohenArthur: and then I would change the enum to this, per my other comment on the FFI enum thing
```suggestion
enum ParseMode
{
  Format = 0,
  InlineAsm,
};
```

CohenArthur (approved):
one nitpick because I'm a hardass on FFI but that looks good!! well done! exactly what I meant :)

libgrust/libformat_parser/generic_format_parser/src/lib.rs
badumbatish: hmm i see, thanks for clearing up my confusion haha, i was just changing stuff up, good to know src/lib.rs is our side

CohenArthur: yeah no worries it's not super explicit!

libgrust/libformat_parser/src/lib.rs
```diff
@@ -327,16 +327,20 @@ mod ffi {
...
         style: Option<usize>,
         snippet: Option<String>,
         append_newline: bool,
+        parse_mode: ParseMode,
     ) -> Vec<Piece<'_>> {
-        let parser = Parser::new(input, style, snippet, append_newline, ParseMode::Format);
-
+        let converted_parse_mode = match parse_mode {
+            ParseMode::Format => generic_format_parser::ParseMode::Format,
+            ParseMode::InlineAsm => generic_format_parser::ParseMode::InlineAsm,
```
CohenArthur, line 341: this is good, if you want one thing you could do to avoid the nameclash is the following:

```
use generic_format_parser::ParseMode as RustcParseMode;
```

and then use it like so:


```suggestion
        let converted_parse_mode = match parse_mode {
            ParseMode::Format => RustcParseMode::Format,
            ParseMode::InlineAsm => RustcParseMode::InlineAsm,
```

or we could also add an implementation for `Into<RustcParseMode> for ParseMode`.

but this is definitely just a nit/suggestion and not a requirement

gcc/rust/expand/rust-macro-builtins-format-args.cc
```diff
@@ -162,7 +163,8 @@ MacroBuiltin::format_args_handler (location_t invoc_locus,
   if (append_newline)
     fmt_str += '\n';
 
-  auto pieces = Fmt::Pieces::collect (fmt_str, append_newline);
+  auto pieces = Fmt::Pieces::collect (fmt_str, append_newline,
+				      Fmt::ffi::ParseMode::Format);
```
CohenArthur, line 167: another small nitpick - on the C++ side of things I'd like to avoid using the `ffi` namespace, as it's meant more for FFI related stuff. so ideally we'd have two functions, something like `collect_asm_pieces` and `collect_fmt_pieces` (or other names). but again, this is a suggestion and not a requirement :)

CohenArthur (approved):
looks great! all the suggestions are not requirements, so if you feel like adding them we should do that in another PR. I'll merge this. thanks a lot, well done!

libgrust/libformat_parser/src/lib.rs
badumbatish: :) thank you for reviewing this

Issue #3069, opened 2024-06-25
Make asm parser stores parse result


badumbatish:
Potential Long standing issue. 

This is required for asm code gen of generic IL. Past PR about asm from me only focuses on parsing the work right instead of recording the result of the parse. 

Every successful parse needs to be added to `InlineAsm&` from `InlineAsmCtx`

badumbatish:
resolve in #3073

PR #3093, opened 2024-07-21
Change assertion of asm operand constructor

badumbatish:
Related to #3069, needed for asm generic il to work

I realized I was asserting expr instead of this->expr and the code kept failing from assertion for no reason ...

Change the assert of (expr != nullptr) to
(this->expr != nullptr) because we assigned (std::move(expr)) to this->expr, no need to assert expr

gcc/rust/ChangeLog:

	* ast/rust-expr.h (struct AnonConst): Change assertion of constructor 
	* (struct InlineAsmOperand): Change assertion of constructor


CohenArthur (approved):
LGTM!

Issue #1791, opened 2023-02-01
Expand necessary builtin macros eagerly


CohenArthur:
The following list of macros is allowed to expand eagerly as of Rust 1.49. In our current implementation, this will mean tweaking that builtin's handler in `rust-macro-builtins.cc` in order to return a new fragment containing the unexpanded macro expansion. The list is as follows:

- [x] `concat!()`
- [ ] `cfg!()`
- [ ] `assert!()`
- [ ] `format!()`
- [ ] `env!()`
- [ ] `asm!()`
- [ ] `global_asm!()`
- [ ] `llvm_asm!()`
- [ ] `line!()`
- [ ] `column!()`
- [ ] `file!()`
- [ ] `stringify!()`
- [ ] `module_path!()`
- [ ] `include_str!()`
- [ ] `include_bytes!()`

I'll create associated issues for each of these tasks. They should all be fairly easy to deal with and can be thought of as one big good first issue :)

CohenArthur:
Updated as most of the macros on this list did not need to be expanded eagerly :)

PR #3094, opened 2024-07-22
Add option_env support

### LINKED ISSUE #1806: Implement eager expansion for `option_env!()`

liamnaddell:
```

    gcc/rust/ChangeLog:
            * expand/rust-macro-builtins-utility.cc: Add macro expansion for
            option_env with eager expansion
            * expand/rust-macro-builtins.cc: Add option_env to builtin list
            * expand/rust-macro-builtins.h: Add option_env handler to header
            file

    gcc/testsuite/ChangeLog:
            * rust/compile/macros/builtin/option_env1.rs: Add success case for option_env
            * rust/compile/macros/builtin/option_env2.rs: Add failure case for option_env
            * rust/execute/torture/builtin_macro_option_env.rs: Add
            execution case for option_env
            * rust/compile/nr2/exclude: Some issues with nr2 need to be
            resolved before allowing option_env with NR2.0.
```

Fixes #1806 

Addresses #927, #1791 

- \[x] Run `clang-format`

---

Adds compiler support for option_env, adds eager expansion for option_env, adds tests for option_env

### DIFF
--- a/gcc/rust/expand/rust-macro-builtins-utility.cc
+++ b/gcc/rust/expand/rust-macro-builtins-utility.cc
@@ -17,6 +17,7 @@
 // <http://www.gnu.org/licenses/>.
 
 #include "rust-fmt.h"
+#include "rust-ast-builder.h"
 #include "rust-macro-builtins.h"
 #include "rust-macro-builtins-helpers.h"
 
@@ -224,6 +225,83 @@ MacroBuiltin::env_handler (location_t invoc_locus, AST::MacroInvocData &invoc,
     = make_token (Token::make_string (invoc_locus, std::move (env_value)));
 
   return AST::Fragment ({node}, std::move (tok));
+}
+
+/* Expand builtin macro option_env!(), which inspects an environment variable at
+   compile time. */
+tl::optional<AST::Fragment>
+MacroBuiltin::option_env_handler (location_t invoc_locus,
+				  AST::MacroInvocData &invoc,
+				  AST::InvocKind semicolon)
+{
+  auto invoc_token_tree = invoc.get_delim_tok_tree ();
+  MacroInvocLexer lex (invoc_token_tree.to_token_stream ());
+  Parser<MacroInvocLexer> parser (lex);
+
+  auto last_token_id = macro_end_token (invoc_token_tree, parser);
+  std::unique_ptr<AST::LiteralExpr> lit_expr = nullptr;
+  bool has_error = false;
+
+  auto start = lex.get_offs ();
+  auto expanded_expr = try_expand_many_expr (parser, last_token_id,
+					     invoc.get_expander (), has_error);
+  auto end = lex.get_offs ();
+
+  auto tokens = lex.get_token_slice (start, end);
+
+  if (has_error)
+    return AST::Fragment::create_error ();
+
+  auto pending = check_for_eager_invocations (expanded_expr);
+  if (!pending.empty ())
+    return make_eager_builtin_invocation (BuiltinMacro::OptionEnv, invoc_locus,
+					  invoc_token_tree,
+					  std::move (pending));
+
+  if (expanded_expr.size () != 1)
+    {
+      rust_error_at (invoc_locus, "%<option_env!%> takes 1 argument");
+      return AST::Fragment::create_error ();
+    }
+
+  if (expanded_expr.size () > 0)
+    if (!(lit_expr
+	  = try_extract_string_literal_from_fragment (invoc_locus,
+						      expanded_expr[0])))
+      return AST::Fragment::create_error ();
+
+  parser.skip_token (last_token_id);
+
+  auto env_value = getenv (lit_expr->as_string ().c_str ());
+  AST::Builder b (invoc_locus);
+
+  if (env_value == nullptr)
+    {
+      auto none_expr = std::unique_ptr<AST::Expr> (
+	new AST::PathInExpression (LangItem::Kind::OPTION_NONE, {},
+				   invoc_locus));
+
+      auto node = AST::SingleASTNode (std::move (none_expr));
+      std::vector<AST::SingleASTNode> nodes;
+      nodes.push_back (node);
+
+      return AST::Fragment (nodes, std::vector<std::unique_ptr<AST::Token>> ());
+    }
+  std::vector<std::unique_ptr<AST::Expr>> args;
+  args.push_back (b.literal_string (env_value));
+
+  std::unique_ptr<AST::Expr> some_expr
+    = b.call (std::unique_ptr<AST::Expr> (
+		new AST::PathInExpression (LangItem::Kind::OPTION_SOME, {},
+					   invoc_locus)),
+	      std::move (args));
+
+  auto node = AST::SingleASTNode (std::move (some_expr));
+
+  std::vector<AST::SingleASTNode> nodes;
+  nodes.push_back (node);
+
+  return AST::Fragment (nodes, std::vector<std::unique_ptr<AST::Token>> ());
 }
 
 tl::optional<AST::Fragment>
--- a/gcc/rust/resolve/rust-late-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-late-name-resolver-2.0.cc
@@ -254,6 +254,8 @@ Late::visit (AST::PathInExpression &expr)
   // TODO: How do we have a nice error with `can't capture dynamic environment
   // in a function item` error here?
   // do we emit it in `get<Namespace::Labels>`?
+  if (expr.is_lang_item ())
+    return;
 
   auto resolved
     = ctx.values.resolve_path (expr.get_segments ()).or_else ([&] () {
--- a/gcc/rust/expand/rust-macro-builtins.cc
+++ b/gcc/rust/expand/rust-macro-builtins.cc
@@ -120,8 +120,8 @@ std::unordered_map<std::string, AST::MacroTranscriberFunc>
     {"format_args_nl", format_args_maker (AST::FormatArgs::Newline::Yes)},
     {"asm", inline_asm_maker (AST::AsmKind::Inline)},
     {"global_asm", inline_asm_maker (AST::AsmKind::Global)},
+    {"option_env", MacroBuiltin::option_env_handler},
     /* Unimplemented macro builtins */
-    {"option_env", MacroBuiltin::sorry},
     {"concat_idents", MacroBuiltin::sorry},
     {"module_path", MacroBuiltin::sorry},
     {"log_syntax", MacroBuiltin::sorry},
--- a/gcc/rust/expand/rust-macro-builtins.h
+++ b/gcc/rust/expand/rust-macro-builtins.h
@@ -159,6 +159,10 @@ public:
 						  AST::MacroInvocData &invoc,
 						  AST::InvocKind semicolon);
 
+  static tl::optional<AST::Fragment>
+  option_env_handler (location_t invoc_locus, AST::MacroInvocData &invoc,
+		      AST::InvocKind semicolon);
+
   static tl::optional<AST::Fragment> cfg_handler (location_t invoc_locus,
 						  AST::MacroInvocData &invoc,
 						  AST::InvocKind semicolon);
--- a/gcc/testsuite/rust/compile/macros/builtin/option_env1.rs
+++ b/gcc/testsuite/rust/compile/macros/builtin/option_env1.rs
@@ -1,0 +1,29 @@
+#![feature(rustc_attrs)]
+
+#[rustc_builtin_macro]
+macro_rules! option_env {
+    () => {}
+}
+
+#[lang = "sized"]
+trait Sized {}
+
+pub mod core {
+    pub mod option {
+        pub enum Option<T> {
+            #[lang = "Some"]
+            Some(T),
+            #[lang = "None"]
+            None,
+        }
+    }
+}
+
+use core::option::Option;
+
+
+fn main() {
+    // Both a guaranteed-to-exist variable and a failed find should compile
+    let _: Option<&str> = option_env!("PWD");
+    let _: Option<&str> = option_env!("PROBABLY_DOESNT_EXIST");
+}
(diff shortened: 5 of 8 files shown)


gcc/rust/expand/rust-macro-builtins-utility.cc
```diff
@@ -226,6 +227,86 @@ MacroBuiltin::env_handler (location_t invoc_locus, AST::MacroInvocData &invoc,
...
+  auto env_value = getenv (lit_expr->as_string ().c_str ());
+  AST::Builder b (invoc_locus);
+  std::string option = "Option";
+
+  if (env_value == nullptr)
+    {
+      std::string none = "None";
+      std::unique_ptr<AST::Expr> none_expr
+	= std::unique_ptr<AST::Expr> (new AST::PathInExpression (
+	  b.path_in_expression (std::vector<std::string> ({option, none}))));
```
CohenArthur, line 284: so, the issue with this is that this won't necessarily pick up the "right" `Option` type, especially at the moment since gccrs does not handle the Rust prelude. a way to counter this would be to use this path in expression instead:

```suggestion
	  b.path_in_expression ({"core", "option", "Option", "None"})));
```

but this has an effect on your testcases, as you now need to add two modules to each of them

gcc/testsuite/rust/compile/option_env1.rs
```diff
@@ -0,0 +1,21 @@
...
+    () => {}
+}
+
+#[lang = "sized"]
+trait Sized {}
+
+enum Option<T> {
+    Some(T),
+    None,
+}
```
CohenArthur, line 14: ...something like this:

```suggestion
mod core {
mod option {

enum Option<T> {
    Some(T),
    None,
}

}
}
```

CohenArthur (commented):
honestly, well done - this isn't an easy part of the codebase, and you did a really good job :) don't worry about the tokenstream stuff, we can think about fixing this later. for the path issue, you're hitting a very frustrating issue which I'm trying to fix in #3068. ideally, we should be able to create paths to lang items (such as `Option::Some` and `Option::None`) so that we don't care about the module structure, or the name, or anything - we just care about hitting the proper type from the standard library. good work!

gcc/rust/expand/rust-macro-builtins-utility.cc
liamnaddell: Sure, I didn't realize it was better to use the full path. 
(Also I didn't know {"a"} => vector<string>, super useful)

gcc/testsuite/rust/compile/option_env1.rs
liamnaddell: I fixed this for the tests. However, to get the compilation to work, I needed an extra
`use core::option::Option;`

Is that expected behavior? I would assume so.

liamnaddell:
> honestly, well done - this isn't an easy part of the codebase, and you did a really good job :)

Thank you :)

gcc/testsuite/rust/compile/option_env1.rs
CohenArthur: hmmm... this is weird. you can investigate the generated code by using `-frust-dump-all` and looking at the generated `gccrs.ast-pretty-expanded.dump` file. it should show exactly the code you've generated with your macro expansion

liamnaddell: This seems like the expected behavior, since we don't have a prelude at the moment:
```rust
liam@gentoo ~/option_env2 $ rustc main.rs
error[E0412]: cannot find type `Option2` in this scope
  --> main.rs:23:13
   |
23 |     let _ : Option2< &'_ str > = core2::option2::Option2::Some("/home/liam/option2_env2");
   |             ^^^^^^^
  --> /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/option.rs:571:1
   |
   = note: similarly named enum `Option` defined here
   |
help: an enum with a similar name exists
   |
23 |     let _ : Option< &'_ str > = core2::option2::Option2::Some("/home/liam/option2_env2");
   |             ~~~~~~
help: consider importing this enum
   |
9  + use core2::option2::Option2;
   |

error[E0412]: cannot find type `Option2` in this scope
  --> main.rs:24:13
   |
24 |     let _ : Option2< &'_ str > = core2::option2::Option2::None;
   |             ^^^^^^^
  --> /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/option.rs:571:1
   |
   = note: similarly named enum `Option` defined here
   |
help: an enum with a similar name exists
   |
24 |     let _ : Option< &'_ str > = core2::option2::Option2::None;
   |             ~~~~~~
help: consider importing this enum
   |
9  + use core2::option2::Option2;
   |

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0412`.
liam@gentoo ~/option_env2 $ cat main.rs
//#![feature(rustc_attrs)]
//#[rustc_builtin_macro] macro_rules! option2_env {
//    () => { };
//}

//#[lang = "sized"]
//trait Sized< > { }

pub mod core2 {
    pub mod option2 {
        pub enum Option2< T > {
            Some(T),
            None,
        }

}

}

//use core2::option2::Option2;

fn main() {
    let _ : Option2< &'_ str > = core2::option2::Option2::Some("/home/liam/option2_env2");
    let _ : Option2< &'_ str > = core2::option2::Option2::None;
}
```

liamnaddell: I did mess up the "pub" on the module names in the test though

liamnaddell:
@CohenArthur Should I update this PR to use the lang-item paths from https://github.com/Rust-GCC/gccrs/pull/3068 ?

P-E-P:
> @CohenArthur Should I update this PR to use the lang-item paths from #3068 ?

I've been trying to catch up on that issue (sorry for the delay!). We should probably rely on lang item path because right now the tests make use of a `core` module where it should be the `core` library (aka technically the file itself since we're defining the builtin macro from within the crate).

liamnaddell:
> > @CohenArthur Should I update this PR to use the lang-item paths from #3068 ?
> 
> I've been trying to catch up on that issue (sorry for the delay!). We should probably rely on lang item path because right now the tests make use of a `core` module where it should be the `core` library (aka technically the file itself since we're defining the builtin macro from within the crate).

Ok, I can wait until the lang items path stuff is usable. If appropriate, we can also revisit some of the other macros then as well to integrate lang items.

liamnaddell:
Likely waiting on #3343 to be merged

liamnaddell:
Marking as "ready for review" since most of the code should be there for option_env support

liamnaddell:
Should be ready after #3366 is integrated

gcc/rust/hir/tree/rust-hir-path.h
```diff
@@ -283,6 +284,7 @@ class PathInExpression : public PathPattern, public PathExpr
   // Constructor
   PathInExpression (Analysis::NodeMapping mappings,
 		    std::vector<PathExprSegment> path_segments,
+		    tl::optional<LangItem::Kind> lang_item = tl::nullopt,
```
CohenArthur, line 287: I'd like to see two constructors, one with segments and one with the lang item. Is that code I wrote or is it something you added for this PR?

gcc/rust/hir/rust-ast-lower.cc
```diff
@@ -495,6 +495,18 @@ void
...
   std::vector<HIR::PathExprSegment> path_segments;
+  auto crate_num = mappings.get_current_crate ();
+  Analysis::NodeMapping mapping (crate_num, expr.get_node_id (),
+				 mappings.get_next_hir_id (crate_num),
+				 UNKNOWN_LOCAL_DEFID);
+  if (expr.get_path_kind () == AST::Path::Kind::LangItem)
+    {
+      translated
+	= new HIR::PathInExpression (mapping, std::move (path_segments),
+				     expr.get_lang_item (), expr.get_locus (),
```
CohenArthur, line 506: so we'd get rid of the segments here for example

```suggestion
	= new HIR::PathInExpression (mappin, expr.get_lang_item (), expr.get_locus (),
```

gcc/rust/typecheck/rust-hir-type-check-path.cc
```diff
@@ -175,12 +176,53 @@ TypeCheckExpr::visit (HIR::QualifiedPathInExpression &expr)
...
+  context->insert_variant_definition (expr.get_mappings ().get_hirid (),
+				      vde->get_id ());
+  bl = SubstMapper::InferSubst (bl, expr.get_locus ());
+  resolver->insert_resolved_misc (expr.get_mappings ().get_nodeid (),
+				  expr.get_mappings ().get_nodeid ());
+
+  infered = bl;
+  rust_assert (bl != nullptr);
+  return;
+}
```
CohenArthur, line 211: @philberty how does that look? you know the typechecker better than me lol

gcc/rust/ast/rust-path.h
```diff
@@ -622,15 +622,14 @@ class Path : public Pattern
 
   std::string as_string () const override;
 
-  // TODO: this seems kinda dodgy
   std::vector<PathExprSegment> &get_segments ()
   {
-    rust_assert (kind == Kind::Regular);
+    // For lang-items, this returns an empty path.
     return segments;
```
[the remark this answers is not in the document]
liamnaddell: Ok, I can change this, but it will require adding a lot of code like:

```
if (path.is_lang_item())
   return;
```

to various AST passes

gcc/rust/hir/tree/rust-hir-path.h
liamnaddell: I added all lang item handling at/below HIR level for PathInExpression.

I can fix this, this was a hack I forgot to remove

gcc/rust/typecheck/rust-hir-type-check-path.cc
liamnaddell: As a note, I'm looking at this code again, and it seems like some of the assert's might want to be compile errors instead maybe. I'm thinking of the case where somebody puts `#[lang = 'Some']` on something that's not an enum variant

CohenArthur:
> Also I see the build-and-check with glibcxx assertions turned on ICE's. I'll try to figure out what's going on there. (gcc is giving me trouble with compiling :/ )

might be because we're doing something like `.get_segment().front()` somewhere, which triggers the assertions (and is UB). this is also why I think we should separate lang items and segmented paths and always make `get_segments()` assert that we're dealing with a segmented path

liamnaddell:
> > Also I see the build-and-check with glibcxx assertions turned on ICE's. I'll try to figure out what's going on there. (gcc is giving me trouble with compiling :/ )
> 
> might be because we're doing something like `.get_segment().front()` somewhere, which triggers the assertions (and is UB). this is also why I think we should separate lang items and segmented paths and always make `get_segments()` assert that we're dealing with a segmented path

Because of my env, I was compiling gccrs with gccrs, which didn't work, because I think gccrs' gcc wasn't properly configured. I changed to use system gcc and it compiles perfectly. 

Your guess was exactly what it was. the ICE was rust-hir-path.cc:266 segments.back ()

gcc/rust/backend/rust-compile-resolve-path.cc
```diff
@@ -42,8 +42,23 @@ ResolvePathRef::visit (HIR::QualifiedPathInExpression &expr)
...
+      bool ok
+	= ctx->get_tyctx ()->lookup_type (expr.get_mappings ().get_hirid (),
+					  &lookup);
+      rust_assert (ok);
+
+      tree t = attempt_constructor_expression_lookup (lookup, ctx,
+						      expr.get_mappings (),
+						      expr.get_locus ());
+      TREE_USED (t) = 1;
+      resolved = t;
```
liamnaddell, line 57: @CohenArthur That GLIBCXX assert CI/CD job caught some undefined behavior, where I pushed a random address into the ResolvePathRef::resolve function as the "final segment". I added code here to properly handle lang-item paths.

(Still working on fixing the AST::PathInExpression, I should be able to get that done soon)

gcc/rust/ast/rust-path.h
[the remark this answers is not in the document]
liamnaddell: Fixed

gcc/rust/ast/rust-path.h
```diff
@@ -622,15 +622,14 @@ class Path : public Pattern
...
   }
   const std::vector<PathExprSegment> &get_segments () const
   {
-    rust_assert (kind == Kind::Regular);
+    // For lang-items, this returns an empty path.
```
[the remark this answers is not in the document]
liamnaddell: Fixed

gcc/rust/hir/tree/rust-hir-path.h
liamnaddell: Fixed

gcc/rust/hir/rust-ast-lower.cc
liamnaddell: Fixed

gcc/rust/typecheck/rust-hir-type-check-path.cc
CohenArthur: that's fine, because having "wrong lang items" is such a corner case that we don't really need to handle it. we know that `core` is correct code, if someone wants to reimplement their own `core` then the burden falls on them rather than on the compiler. asserts are better IMO

gcc/rust/resolve/rust-late-name-resolver-2.0.cc
```diff
@@ -254,6 +254,8 @@ Late::visit (AST::PathInExpression &expr)
   // TODO: How do we have a nice error with `can't capture dynamic environment
   // in a function item` error here?
   // do we emit it in `get<Namespace::Labels>`?
+  if (expr.is_lang_item ())
+    return;
```
CohenArthur, line 258: I think resolution here should be pretty easy but we'll do this in another PR - I'm happy with that fix

CohenArthur (approved):
that's great work, looks perfect to me :D thank you so much @liamnaddell !

PR #3100, opened 2024-07-26
Fixed bitwise operation in `extract_stmt`

braw-lee:
1 << 16 gives : 10000000000000000
~(1 << 16) gives : 11111111111111101111111111111111
i think the assumption was output of ~(1 << 16) will be 1111111111111111, which results in extraction of wrong statement

this bug can be verified by checking nll_dump of this function
```rust
fn complex_cfg_subset<'a, 'b>(b: bool, x: &'a u32, y: &'b u32) -> &'a u32 {
    if b { 
        y
    } else {
        x
    }   
}
```

### DIFF
--- a/gcc/rust/checks/errors/borrowck/polonius/rust-polonius.h
+++ b/gcc/rust/checks/errors/borrowck/polonius/rust-polonius.h
@@ -45,7 +45,7 @@ struct FullPoint
   static uint32_t extract_bb (Point point) { return point >> 16; }
   static uint32_t extract_stmt (Point point)
   {
-    return (point & ~(1 << 16)) >> 1;
+    return (point >> 1) & ((1 << 15) - 1);
   }
   static bool extract_mid (Point point) { return point & 1; }
 


CohenArthur (approved):
LGTM

jdupak (approved):
😅

gcc/rust/checks/errors/borrowck/polonius/rust-polonius.h
```diff
@@ -45,7 +45,7 @@ struct FullPoint
   static uint32_t extract_bb (Point point) { return point >> 16; }
   static uint32_t extract_stmt (Point point)
   {
-    return (point & ~(1 << 16)) >> 1;
+    return (point >> 1) & ((1 << 15) - 1);
```
P-E-P, line 48: To be fair I would have not written it like this. We lack any documentation/intent hint with such expression. In such case I would:
- Use many constexpr or define to describe the intent of the operation (What does the 1 correspond to ? What about the mask ?) We should give a name to every hard coded constant.
- I would also (probably) change `((1 << 15) - 1)` to `0x7fff`, I think it better hints about the binary representation ? To be fair I'd like to hear from you about this.

P-E-P (approved):
Approving this because you need those changes quickly and it is indeed a correct fix, but I'd like to discuss the shape of this.

jdupak:
`Personally I need to think really hard to figure out what it is and how many bits it masks for 0x7fff.`

At that time I was thinking about function 'extract_bits' but it did not seem justified. Because it is used just for this one thing, where I need to serialize 3 numbers into one.

Making constants for the masks, etc sounds good.

PR #3109, opened 2024-07-31
Inline asm resolve expr

badumbatish:
This should resolve InlineAsm through its operand, depends on https://github.com/Rust-GCC/gccrs/pull/3103

Relevant file for review is rust-ast-resolve-expr.h/cc

### DIFF
--- a/gcc/rust/resolve/rust-ast-resolve-expr.cc
+++ b/gcc/rust/resolve/rust-ast-resolve-expr.cc
@@ -347,6 +347,60 @@ ResolveExpr::visit (AST::BlockExpr &expr)
 }
 
 void
+translate_operand (AST::InlineAsm &expr, const CanonicalPath &prefix,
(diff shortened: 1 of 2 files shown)


badumbatish:
For HIR::TypeCheckExpr, this weirdly depends on https://github.com/Rust-GCC/gccrs/pull/3060 since asm_generic_il implements the hir visitor pattern for inline asm. To break dependency I think i'll implement hir type check inside asm_generic_il.

@CohenArthur thoughts?

CohenArthur (approved):
I think this is exactly what you need to do :) looks great to me

CohenArthur:
> For HIR::TypeCheckExpr, this weirdly depends on #3060 since asm_generic_il implements the hir visitor pattern for inline asm. To break dependency I think i'll implement hir type check inside asm_generic_il.
> 
> @CohenArthur thoughts?

that's fine with me!

Issue #3070, opened 2024-06-26
Union pattern matching expecting all fields


P-E-P:
I tried this code:
https://godbolt.org/z/3GhT7Ms4z
```rust
#[repr(C)]
union MyUnion {
    f1: u32,
    f2: f32,
}

fn f(u: MyUnion) -> i32 {
    unsafe {
        match u {
            MyUnion { f1: 10 } => { 0 }
            MyUnion { f2 } => { 1 }
        }
    }
}

pub fn main() {}
```

I expected to see this happen: Compiles correctly.

Instead, this happened: Compiler emits an error

```
<source>:10:13: error: pattern does not mention fields f2 [E0027]
   10 |             MyUnion { f1: 10 } => { 0 }
      |             ^~~~~~~
<source>:11:13: error: pattern does not mention fields f1 [E0027]
   11 |             MyUnion { f2 } => { 1 }
      |             ^~~~~~~
Compiler returned: 1
```

### Meta

- [40d5292](https://github.com/Rust-GCC/gccrs/commit/40d52922469989d16626fd29d1943596af919d92)

PR #3110, opened 2024-07-31
Add check for union patterns

tamaroning:
Fixes https://github.com/Rust-GCC/gccrs/issues/3070

gcc/rust/ChangeLog:

	* typecheck/rust-hir-type-check-pattern.cc (TypeCheckPattern::visit): Add check for union patterns.

gcc/testsuite/ChangeLog:

	* rust/compile/match8.rs: New test.

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-pattern.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-pattern.cc
@@ -197,10 +197,6 @@ TypeCheckPattern::visit (HIR::StructPattern &pattern)
       return;
     }
 
-  // check the elements
-  // error[E0027]: pattern does not mention fields `x`, `y`
-  // error[E0026]: variant `Foo::D` does not have a field named `b`
-
   std::vector<std::string> named_fields;
   auto &struct_pattern_elems = pattern.get_struct_pattern_elems ();
   for (auto &field : struct_pattern_elems.get_struct_pattern_fields ())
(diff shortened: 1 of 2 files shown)


gcc/testsuite/rust/compile/match8.rs
badumbatish, line 1: hi! just passing by to learn new stuff, i think the test case is missing this case?
```
rust_error_at (pattern.get_locus (),
			       "'%s' cannot be used in union patterns",
			       first_elem.c_str ());
```

tamaroning: Good pick.
I think test cases for this should be added later since many important patterns to be tested (e.g. `..`) don't seem to be implemented now.

tamaroning: OK, that's because access to union fields is an unsafe operation.
Different union fields can be placed on the same address so accessing a field which is of type `T1` but is initialized as `T2` may cause a bug. It is known as type confusion and is often used by attackers.

https://doc.rust-lang.org/reference/items/unions.html

gcc/rust/typecheck/rust-hir-type-check-pattern.cc
```diff
@@ -254,31 +250,67 @@ TypeCheckPattern::visit (HIR::StructPattern &pattern)
...
 
-  if (named_fields.size () != variant->num_fields ())
+  // check the elements
+  if (adt->is_union ())
     {
-      std::map<std::string, bool> missing_names;
+      auto &struct_pattern_elems = pattern.get_struct_pattern_elems ();
+      if (struct_pattern_elems.get_struct_pattern_fields ().size () != 1)
+	rust_error_at (pattern.get_locus (),
+		       "union patterns does should have exactly one field");
```
CohenArthur, line 259:
```suggestion
		       "union patterns should have exactly one field");
```

typo

gcc/rust/typecheck/rust-hir-type-check-pattern.cc
```diff
@@ -254,31 +250,67 @@ TypeCheckPattern::visit (HIR::StructPattern &pattern)
...
+	    case HIR::StructPatternField::ItemType::IDENT:
+	    case HIR::StructPatternField::ItemType::IDENT_PAT:
+	      break;
+	      default: {
+		auto first_elem
+		  = struct_pattern_elems.get_struct_pattern_fields ()
+		      .at (0)
+		      ->as_string ();
+		rust_error_at (pattern.get_locus (),
+			       "'%s' cannot be used in union patterns",
```
CohenArthur, line 276:
```suggestion
			       "%qs cannot be used in union patterns",
```
to make use of GCC's quoting system

CohenArthur (commented):
nice work :)

gcc/rust/typecheck/rust-hir-type-check-pattern.cc
tamaroning: thanks

gcc/rust/typecheck/rust-hir-type-check-pattern.cc
tamaroning: thanks! I found it is documented in https://gcc.gnu.org/codingconventions.html  ☺️

tamaroning:
Sorry for late response. I fixed everything

P-E-P (approved):

PR #3112, opened 2024-07-31
Improve compressed point bit manipulation

### LINKED ISSUE #3111: Improve compressed point bit manipulation
To be fair I would have not written it like this. We lack any documentation/intent hint with such expression. In such case I would:
- Use many constexpr or define to describe the intent of the operation (What does the 1 correspond to ? What about the mask ?) We should give a name to every hard coded constant.
- I would also (probably) change `((1 << 15) - 1)` to `0x7fff`, I think it better hints about the binary representation ? To be fair I'd like to hear from you about this.

_Originally posted by @P-E-P in https://github.com/Rust-GCC/gccrs/pull/3100#discussion_r1697998526_

braw-lee:
fixes #3111

added some comments on both rust and cpp side
i am not sure about the terminology used in bit manipulation, pls suggest improvements if any
i didn't name the hard-coded constants (1 and 0x7FFF) yet
do we need to name them or the explanation in comments seems enough?
@P-E-P

### DIFF
--- a/gcc/rust/checks/errors/borrowck/ffi-polonius/src/lib.rs
+++ b/gcc/rust/checks/errors/borrowck/ffi-polonius/src/lib.rs
@@ -105,9 +105,24 @@ impl From<gccrs_ffi::FactsView> for AllFacts<GccrsFacts> {
 
 fn print_point(point: GccrsAtom) {
     let val: usize = point.into();
+    // Point is a 32 bit unsigned integer
+    // 16               15              1
+    // xxxxxxxxxxxxxxxx xxxxxxxxxxxxxxx x
+    // ^~~~~~~~~~~~~~~~ ^~~~~~~~~~~~~~~ ^
+    // |                |               |
+    // basic_block      |               start/mid
+    //                  statement
+    // the left most 16 bits store the basic block number
+    // the right most bit, represents the start/mid status
+    // the remaining 15 bits between these two represent the statement
     let mid = val % 2 == 1;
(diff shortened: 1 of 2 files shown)


P-E-P:
> added some comments on both rust and cpp side

Nice, it really lacked some
 > i am not sure about the terminology used in bit manipulation, pls suggest improvements if any

Looks fine, probably a tiny teensy overly verbose but I still like it like that.
> i didn't name the hard-coded constants (1 and 0x7FFF) yet do we need to name them or the explanation in comments seems enough? @P-E-P

It is already clear enough with the comments, I'll put here what I would have written myself but this is meant as an example for you to get a rough idea, not because I want you to change your code. This is a community project, not @P-E-P 's personal project :smile: 

```cpp
// probably in a namespace or something so those values don't slip everywhere in the project
constexpr uint32_t BASIC_BLOCK_OFFSET = 16;
constexpr uint32_t BASIC_BLOCK_MASK = 0xFFFF;

constexpr uint32_t STATEMENT_OFFSET = 1;
constexpr uint32_t STATEMENT_MASK = 0x7FFF;

// [...]

return (point >> STATEMENT_OFFSET) & STATEMENT_MASK;
```

What you've written is fine, that's just a different way of doing it. Some people would probably have written a union over `Point` and a bitfield struct or something.

P-E-P (approved):

braw-lee:
ah that looks much better, now I feel like copy pasting this 😄

PR #3114, opened 2024-08-02
Fix 32-bits builds

CohenArthur:
Fixes this bugzilla PR: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=116192

We are once again hitting issues for formatting `size_t` values in our printf-like macros. We should think about having a 32bit build CI for cases like these. Also, on architectures where `size_t` is `unsigned int`, such as 32bit x86, we encounter an issue with `PlaceId` and `FreeRegion` being aliases to the same types. This poses an issue for overloading functions for these two types, such as `push_subset` in that case. This commit renames one of these `push_subset` functions to avoid the issue, but this should be fixed with a newtype pattern for these two types, as is currently being discussed on Zulip

### DIFF
--- a/gcc/rust/checks/errors/borrowck/rust-bir-builder.h
+++ b/gcc/rust/checks/errors/borrowck/rust-bir-builder.h
@@ -91,7 +91,8 @@ private:
 	  ctx.fn_free_regions[bound.second.get_index ()]);
 
 	auto last_bound = universal_region_bounds.back ();
-	rust_debug ("\t\t %ld: %ld", last_bound.first, last_bound.second);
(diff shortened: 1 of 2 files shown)


CohenArthur:
@braw-lee also asking you for a review but I can't add you to the github list :)

jdupak (approved):
LGTM

Eventually, I would like strong types here. I mainly avoided it to move faster with the FFI bindings.

But regardless, I have no problem with the named method.

braw-lee:
LGTM!
I will be adding strong types soon
agree with Jakub, method name change is good

P-E-P (approved):
LGTM

PR #3121, opened 2024-08-05
rust: avoid clobbering LIBS

dkm:
Save LIBS around calls to AC_SEARCH_LIBS to avoid clobbering $LIBS.

### DIFF
--- a/configure
+++ b/configure
@@ -8829,9 +8829,12 @@ fi
 
 # Rust requires -ldl and -lpthread if you are using an old glibc that does not include them by
 # default, so we check for them here
-
+# We are doing the test here and not in the gcc/configure to be able to nicely disable the
+# build of the Rust frontend in case a dep is missing.
 missing_rust_dynlibs=none
 
+save_LIBS="$LIBS"
+LIBS=
 { $as_echo "$as_me:${as_lineno-$LINENO}: checking for library containing dlopen" >&5
 $as_echo_n "checking for library containing dlopen... " >&6; }
 if ${ac_cv_search_dlopen+:} false; then :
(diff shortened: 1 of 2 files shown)


tschwinge (approved):
Oh, are the `AC_SEARCH_LIBS`s currently affecting the general `LIBS` variable?  In that case, your proposed changes seems correct to me (which files regenerated).


> Also, do you know why this is in the top-level .ac and not in the gcc/ one?

So that we can still silently/verbosely "Disable Rust if we are missing some required C libraries for the Rust runtime", see further down in `configure.ac`, where `missing_rust_dynlibs` is used.

tschwinge:
Indeed with your change, on a Ubuntu 20.04 system, the top-level `config.status` changes as follows:

```
@@ -951 +951 @@
-S["CRAB1_LIBS"]=" -ldl -lpthread"
+S["CRAB1_LIBS"]="-lpthread -ldl "
@@ -1021 +1021 @@
-S["LIBS"]="-lpthread -ldl "
+S["LIBS"]=""
```

..., which seems correct to me.

```
Reviewed-by: Thomas Schwinge <tschwinge@baylibre.com>
Tested-by: Thomas Schwinge <tschwinge@baylibre.com>
```

dkm:
Thanks. This was directed by discussion on IRC. This is a common pattern in gcc/configure.ac.

> So that we can still silently/verbosely "Disable Rust if we are missing some required C libraries for the Rust runtime", see further down in configure.ac, where missing_rust_dynlibs is used.

oh, ok, thanks. I'll forward this reply on IRC where the question was asked. I'll send the patch on gcc-patches@ later today

dkm:
I've cleaned the log, generated the `configure` file, added a small comment so that we don't have again the question "why is it in the toplevel makefile?" and I've added the `Reviewed-by/Tested-by` footer. I'll wait for your review before merging it and sending the same change to gcc-patches@.

tschwinge (approved):
Looks good, just one boring item: please re-regenerate `configure`, so that "added a small comment" also gets propagated there.

tschwinge (approved):

dkm:
Looks like I should also update the commit checker, it doesn't like `Tested-By` footer.

dkm:
This has been merged upstream as r15-3443-gda3a2985fff39e

CohenArthur:
well done :) thanks

PR #3124, opened 2024-08-10
Dynamic dispatch with supertraits

### LINKED ISSUE #914: Dynamic objects support inheritance

liamnaddell:
There was only one function that needed modification, the rest is testing for this new feature :)
```
    gcc/rust/ChangeLog:

            * backend/rust-compile.cc:
            Modify compute_address_for_trait_item to support supertraits
            * typecheck/rust-tyty.cc:
            Remove auto

    gcc/testsuite/ChangeLog:

            * rust/compile/trait13.rs:
            Add test for supertraits of supertraits
            * rust/compile/trait14.rs:
            Diamond problem with supertraits test
            * rust/execute/torture/trait14.rs:
            Add test for dynamic dispatch with supertraits
            * rust/execute/torture/trait15.rs:
            Add test for dynamic dispatch with generics
            * rust/execute/torture/trait16.rs:
            Add test for dynamic dispatch with lifetime params 1
            * rust/execute/torture/trait17.rs:
            Add test for dynamic dispatch with lifetime params 2
            * rust/execute/torture/trait18.rs:
            Add test for default implementations with dynamic dispatch and
            supertraits
```

Fixes #914 

- \[x] Run `clang-format`

---

Modifies compute_address_for_trait_item to support supertraits. The algorithm works by looping over receiver_bounds for all relevant traits, and attempting to find one that matches the type, and contains the correct function implementation.

### DIFF
--- a/gcc/rust/backend/rust-compile.cc
+++ b/gcc/rust/backend/rust-compile.cc
@@ -241,100 +241,84 @@ HIRCompileBase::compute_address_for_trait_item (
     &receiver_bounds,
   const TyTy::BaseType *receiver, const TyTy::BaseType *root, location_t locus)
 {
-  // There are two cases here one where its an item which has an implementation
-  // within a trait-impl-block. Then there is the case where there is a default
-  // implementation for this within the trait.
-  //
-  // The awkward part here is that this might be a generic trait and we need to
-  // figure out the correct monomorphized type for this so we can resolve the
-  // address of the function , this is stored as part of the
-  // type-bound-predicate
-  //
-  // Algo:
-  // check if there is an impl-item for this trait-item-ref first
-  // else assert that the trait-item-ref has an implementation
-  //
-  // FIXME this does not support super traits
-
   TyTy::TypeBoundPredicateItem predicate_item
     = predicate->lookup_associated_item (ref->get_identifier ());
   rust_assert (!predicate_item.is_error ());
 
-  // this is the expected end type
+  // This is the expected end type
   TyTy::BaseType *trait_item_type = predicate_item.get_tyty_for_receiver (root);
   rust_assert (trait_item_type->get_kind () == TyTy::TypeKind::FNDEF);
   TyTy::FnType *trait_item_fntype
     = static_cast<TyTy::FnType *> (trait_item_type);
 
-  // find impl-block for this trait-item-ref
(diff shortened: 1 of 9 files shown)


liamnaddell:
NOTE: I found a bug with the following code sample:
```rust
extern "C" {
    fn printf(s: *const i8, ...);
}

struct Foo {
    my_int: u32,
}

trait Parent {
    fn parent(&self) -> bool;
}

trait Child : Parent {
    fn child(&self);
}

impl Parent for Foo {
    fn parent(&self) -> bool {
        // Call supertrait method
        return true;
    }
}

impl Child for Foo {
    fn child(&self) {
        let _ = self;
    }
}

pub fn main() {
    let a = Foo{ my_int: 0xf00dfeed};
    let b: &dyn Child = &a;

    let c: &dyn Parent = b;

    c.parent();
}
```

I will fix this case soon. Right now it ICE's

P-E-P (approved):
Looks good, great work. I like all those new tests.

PR #3133, opened 2024-08-20
Fix the disorder struct and class in inline asm

badumbatish:
gcc/rust/ChangeLog:

	* ast/rust-ast-full-decls.h (struct InlineAsmOperand): Change to class (class InlineAsmOperand): Change from struct
	* hir/tree/rust-hir-full-decls.h (struct InlineAsmRegOrRegClass): Removed from decl, used from AST (struct AnonConst): new decl from rust-hir-expr.h (class InlineAsmOperand): new decl from rust-hir-expr.h

CI doesn't warn about ast decl having InlineAsmOperand as struct and in rust-expr.h having InlineAsmOperand as class.

Same for rust-hir-full-decls declaring HIR version of InlineAsmRegOrRegClass but doesn't define them in rust-hir-expr.h

### DIFF
--- a/gcc/rust/ast/rust-ast-full-decls.h
+++ b/gcc/rust/ast/rust-ast-full-decls.h
@@ -148,7 +148,7 @@ class AsyncBlockExpr;
 enum class InlineAsmOption;
 struct AnonConst;
 struct InlineAsmRegOrRegClass;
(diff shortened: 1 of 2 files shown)


CohenArthur (approved):
LGTM! thanks!

PR #3135, opened 2024-08-21
attributes: Start handling prelude_import properly

### LINKED ISSUE #2985: Handle `#[prelude_import]` attribute

CohenArthur:
This commit adds basic handling for the `#[prelude_import]` attribute,
without doing anything functionality wise.

gcc/rust/ChangeLog:

	* checks/errors/rust-feature-gate.cc (FeatureGate::visit): Add base
	feature gating for `#[feature(prelude_import)]`.
	* checks/errors/rust-feature-gate.h: Likewise.
	* checks/errors/rust-feature.cc (Feature::create): Likewise.
	* checks/errors/rust-feature.h: Likewise.
	* util/rust-attribute-values.h: Add base handling for `#[prelude_import]`
	attribute.
	* util/rust-attributes.cc: Likewise.

gcc/testsuite/ChangeLog:

	* rust/compile/prelude_import.rs: New test.

Fixes #2985

### DIFF
--- a/gcc/testsuite/rust/compile/prelude_import.rs
+++ b/gcc/testsuite/rust/compile/prelude_import.rs
@@ -1,0 +1,12 @@
+#![feature(prelude_import)]
+
+mod core {
+    mod prelude {
+        mod v1 {
+            // hehe
+        }
+    }
+}
+
+#[prelude_import]
+use core::prelude::v1::*;
--- a/gcc/rust/checks/errors/rust-feature-gate.cc
+++ b/gcc/rust/checks/errors/rust-feature-gate.cc
@@ -217,4 +217,12 @@ FeatureGate::visit (AST::RangePattern &pattern)
 	  "exclusive range pattern syntax is experimental");
 }
 
+void
+FeatureGate::visit (AST::UseTreeGlob &use)
+{
+  // At the moment, UseTrees do not have outer attributes, but they should. we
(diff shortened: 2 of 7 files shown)


P-E-P (approved):

gcc/testsuite/rust/compile/prelude_import.rs
```diff
@@ -0,0 +1,12 @@
+#![feature(prelude_import)]
+
+mod core {
+    mod prelude {
+        mod v1 {
```
liamnaddell, line 5: Shouldn't these be pub mods?

CohenArthur: so this is just a silly testcase for the attribute itself, and we need a valid import path so it doesn't error out. it's the same as doing something like

```rust
#[prelude_import]
use foo;

mod foo {}
```

just making sure we can parse the attribute itself :)

but yes, you are correct that in the real prelude import we need those mods to be public, and the prelude use to be public as well

liamnaddell: Ah ok. AFAIK currently we don't check the pub status of submodules, so I was worried something like this would later error out. That was my only concern.

CohenArthur: ah yeah, this is eventually going to break. I think we'll fix it once we get to this. we should definitely have one more look at our privacy checker because it's not great

PR #3138, opened 2024-08-23
Add glibcxx_assertion check

badumbatish:
ChangeLog:

	* .github/workflows/ccpp.yml: Add glibcxx_assertion check


badumbatish:
will try again tmr

badumbatish:
fixed by adding a boolean check of binder.empty()

.github/workflows/ccpp.yml
```diff
@@ -94,6 +94,88 @@ jobs:
...
+    - name: Check for new warnings
+      run: |
+           cd gccrs-build
+           < log grep 'warning: ' | sort > log_warnings
+           if diff -U0 ../.github/glibcxx_ubuntu64b_log_expected_warnings log_warnings; then
+               :
+           else
+               echo 'See <https://github.com/Rust-GCC/gccrs/pull/1026>.'
+               exit 1
+           fi >&2
```
CohenArthur, line 155: so do we need a different expected warnings file? are there more warnings when `GLIBCXX_ASSERTIONS` is defined?

CohenArthur (approved):
thanks!

.github/workflows/ccpp.yml
badumbatish: the default ubuntu 64 bit has a few more warnings than when we run it with GLIBCXX_ASSERTIONS, thus triggering failure when CI is run. Should we use a different diff at line 150?

PR #3142, opened 2024-08-27
Compile using ASAN on the github CI

powerboat9:
This adds a job for running GCC with ASAN to the github CI

### DIFF
--- a/.github/no-bootstrap-asan.mk
+++ b/.github/no-bootstrap-asan.mk
@@ -1,0 +1,7 @@
+# This option enables -fsanitize=address for stage1.
+
+# Suppress LeakSanitizer in bootstrap.
+export ASAN_OPTIONS=detect_leaks=0:use_odr_indicator=1
+
+STAGE1_CFLAGS += -fsanitize=address
+STAGE1_LDFLAGS += -fsanitize=address -static-libasan
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -386,3 +386,87 @@ jobs:
             else \
               exit 0; \
             fi
+
+  build-and-check-asan:
+
+    env:
+      # Force locale, in particular for reproducible results re '.github/bors_log_expected_warnings' (see below).
+      LC_ALL: C.UTF-8
+
+    runs-on: ubuntu-22.04
+
+    steps:
+    - uses: actions/checkout@v3
+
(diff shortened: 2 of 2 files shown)


CohenArthur (approved):
great stuff! thanks a lot!

powerboat9:
Not working yet, having trouble with linker during link tests

powerboat9:
Alright, looks like things are working now

powerboat9:
It looks like the execute tests are failing because runtime libraries compiled with ASAN require that the programs using them are compiled with ASAN.

.github/configure-san
```diff
@@ -0,0 +1,20 @@
...
+
+# This should be run by github CI
+# However, running it locally (if it works on your PC) might be useful
+
+export GCC_LIBS=$(gcc -print-search-dirs | head -n1 | sed 's/install: //')
+export COMMON="-fsanitize=address"
+
+CONFIGURE=$(dirname $(dirname $0))/configure
+
+$CONFIGURE CFLAGS="$COMMON -B$GCC_LIBS" \
```
thesamesam, line 15: Consider using `--with-build-config="bootstrap-asan"` (or whatever it is) instead, but not sure how to handle tests (some DejaGnu hackery may be needed there).

powerboat9: `bootstrap-asan` doesn't work with `--disable-bootstrap`, as far as I can tell

powerboat9:
Should be ready for review again

.github/configure-san
thesamesam: Ah, you're right, but I see you've figured it out ;)

thesamesam (approved):
LGTM. Thanks!

CohenArthur:
Ah, I think the test failures might be due to the execute testsuite running but not linking with ASAN? this is something that happens to me when I do local ASAN builds. If that is the case, then a solution would be to either run only the compile.exp testsuite with `RUNTESTFLAGS` or figure out how to run the `execute.exp` testsuite in compile-only mode, without linking/running the final binary. but I don't know if dejagnu supports this

powerboat9:
I managed to get the execute tests working, and it seems like the remaining test failures are down to bugs in GCCRS

powerboat9:
The ASAN check *should* pass now

P-E-P (approved):

PR #3147, opened 2024-09-03
Add virtual destructor to `AbstractExpr`

powerboat9:
This allows `std::unique_ptr<AbstractExpr>` to be properly destructed, discovered during development of #3142.


P-E-P (approved):
I stumbled upon the same warning, thank you!

PR #3148, opened 2024-09-04
Fix `v0-mangle1.rs` test when run with ASAN

powerboat9:
Fixes a bug with the test case itself, and one with `crate_num` initialization in `CanonicalPath`


tamaroning:
thanks! lgtm

P-E-P (approved):

PR #3155, opened 2024-09-06
Avoid accidental insertion into map

powerboat9:
[Indexing a map inserts an entry if one does not already exist](https://en.cppreference.com/w/cpp/container/map/operator_at). This should fix the issue with test `alt_patterns1.rs` in #3142


P-E-P:
I couldn't believe it at first but once again this language should be thrown into the fire :rofl: Thank you for pointing at this, I'm 99% sure we have the same pattern in some other place in the codebase, we might have more to fix.

P-E-P (approved):
Great job, this highlights some other potential problems within the codebase, thank you!

powerboat9:
It really does seem like C++ should have an `operator[]=` overload, for write-indexing

Issue #3141, opened 2024-08-24
Incorrect handling of overflow in numeric types


nobel-sh:
There appears to be a bug in the handling of overflows for signed integer types and floating-point types. While unsigned integers are unaffected, assigning the minimum value (`TYPE::MIN`) to signed integer variables results in a compiler error. But in case of floats, error occurs on both `TYPE::MIN` and `TYPE::MAX` and also on values which are smaller than `TYPE::MAX` or larger than `TYPE::MIN`.

## Signed Integer
```rust
// integer types
fn main() {
    let _: i8 = -128;
    let _: i16 = -32768;
    let _: i32 = -2147483648;
    let _: i64 = -9223372036854775808;
}
```
### Expected Behavior:
The code should compile without errors.

### Actual Behavior:
```
<source>:2:18: error: integer overflows the respective type 'i8'
    2 |     let _: i8 = -128;
      |                  ^~~
<source>:3:19: error: integer overflows the respective type 'i16'
    3 |     let _: i16 = -32768;
      |                   ^~~~~
<source>:4:19: error: integer overflows the respective type 'i32'
    4 |     let _: i32 = -2147483648;
      |                   ^~~~~~~~~~
<source>:5:19: error: integer overflows the respective type 'i64'
    5 |     let _: i64 = -9223372036854775808;
      |                   ^~~~~~~~~~~~~~~~~~~
```
## Floating point
``` rust
// floating types
fn main() {
    let _f32_min:f32 = -3.40282347E+38f32;
    let _f32_max:f32 = 3.40282347E+38f32;
    let _f64_min:f64 = 1.7976931348623157E+308f64;
    let _f64_max:f64 = -1.7976931348623157E+308f64;
    // and other float types.

    // Some values although not on the limit also seem to throw
    // compiler error.
    let _f32_random_fail_1:f32 = 1.40282347E+30f32;
    let _f32_random_fail_2:f32 = 1.40282347E+10f32;
    let _f32_random_pass:f32 = 1.40282347E+9f32; // this passes
}
```
### Expected Behavior:
The code should compile without errors.

### Actual Behavior:

```
<source>:9:25: error: decimal overflows the respective type 'f32'
    9 |     let _f32_min:f32 = -3.40282347E+38f32;
      |                         ^~~~~~~~~~~~~~~~~
<source>:10:24: error: decimal overflows the respective type 'f32'
   10 |     let _f32_max:f32 = 3.40282347E+38f32;
[...]

PR #3158, opened 2024-09-10
[#3141] Fix incorrect handling of overflow in numeric types

JoanVC100:
Fixes [#3141]

gcc/rust/ChangeLog:

	* backend/rust-compile-expr.cc: Fix range checking for both integers and floats.
	* hir/tree/rust-hir-expr.h: Add "negative_number" boolean to LiteralExpr class.

gcc/testsuite/ChangeLog:

	* rust/compile/issue-3141.rs: New test.

This pull request aims to fix an incorrect handling of literal "overflows".

To fix the handling of integer literals I added a boolean attribute to the HIR LiteralExpr class to signal if the literal is negative, and compare it accordingly.

As for float literals, I added checks on the float exponent in order to check for overflows.

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -237,6 +237,8 @@ CompileExpr::visit (HIR::NegationExpr &expr)
   auto op = expr.get_expr_type ();
 
   const auto literal_expr = expr.get_expr ().get ();
+
+  // If it's a negated integer/float literal, we can return early
   if (op == NegationOperator::NEGATE
       && literal_expr->get_expression_type () == HIR::Expr::ExprType::Lit)
     {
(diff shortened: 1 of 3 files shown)


gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -1506,13 +1520,22 @@ CompileExpr::compile_integer_literal (const HIR::LiteralExpr &expr,
...
       rust_error_at (expr.get_locus (),
 		     "integer overflows the respective type %<%s%>",
 		     tyty->get_name ().c_str ());
       return error_mark_node;
     }
+  // Other tests break if we don't reverse the negation
+  if (expr.is_negative ())
+    {
+      mpz_neg (ival, ival);
+    }
```
powerboat9, line 1538: That's probably why the negation has to be reversed here, since otherwise both the integer literal expression and negation expression would be performing negations

powerboat9:
Looks good

philberty (approved):
LGTM

PR #3164, opened 2024-09-16
mingw: Fix build with patch from Liu Hao

CohenArthur:
This commit adds Liu Hao's patch from
https://github.com/lhmouse/MINGW-packages/blob/5859d27b2b6101204a08ad9702cb2937f8797be9/mingw-w64-gcc/0100-rust-fix.patch

gcc/rust/ChangeLog:

	* checks/errors/borrowck/rust-borrow-checker.cc (mkdir_wrapped): Remove.
	(BorrowChecker::go): Use `mkdir` instead.
	* expand/rust-proc-macro.cc (register_callback): Use Windows APIs to
	open dynamic proc macro library.
	(load_macros_array): Likewise.
	* parse/rust-parse.cc (defined): Replace separators in paths using
	std::replace.

This adds the patch from https://github.com/lhmouse/MINGW-packages/commit/16dad6ba10a5680ad3ca6304480210cef8366985 but reworked for our new changes.

@lhmouse could you check it looks okay and builds on your side?

Sidenote, this might not build on Linux either - I don't have access to my build machine atm so I'm using the CI to test this patch.

### DIFF
--- a/gcc/rust/checks/errors/borrowck/rust-borrow-checker.cc
+++ b/gcc/rust/checks/errors/borrowck/rust-borrow-checker.cc
@@ -28,20 +28,6 @@ namespace Rust {
 namespace HIR {
 
 void
-mkdir_wrapped (const std::string &dirname)
-{
-  int ret;
-#ifdef _WIN32
-  ret = _mkdir (dirname.c_str ());
-#elif unix
-  ret = mkdir (dirname.c_str (), 0775);
-#elif __APPLE__
-  ret = mkdir (dirname.c_str (), 0775);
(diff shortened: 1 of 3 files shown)


lhmouse:
You may just take the patch and assume it builds. The patch is also available in MSYS2 (https://github.com/msys2/MINGW-packages/blob/e9b8bb22e94a577fd94ac1ded8519778a02499a5/mingw-w64-gcc/2001-fix-building-rust-on-mingw-w64.patch).

P-E-P (approved):

PR #3168, opened 2024-09-17
Avoid visiting nonexistant function parameter types during nr2.0

powerboat9:
This makes 3 more tests pass with name resolution 2.0

### DIFF
--- a/gcc/rust/ast/rust-item.h
+++ b/gcc/rust/ast/rust-item.h
@@ -528,10 +528,9 @@ public:
   NodeId get_node_id () const { return node_id; }
 
   // TODO: is this better? Or is a "vis_block" better?
-  Type &get_type ()
+  tl::optional<Type &> get_type ()
   {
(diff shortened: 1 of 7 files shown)


gcc/rust/ast/rust-item.h
```diff
@@ -3548,6 +3548,8 @@ class NamedFunctionParam
    * '_'). */
   bool has_name () const { return name != "_" && name != ""; }
 
+  bool has_type () const { return param_type != nullptr; }
```
P-E-P, line 3551: If a `NamedFunctionParam` cannot have a type then we should probably wrap the type pointer within an optional.

P-E-P (approved):
There are some errors on gcc 4.8 but otherwise it looks fine.

powerboat9:
It looks like this was mostly superseded by another PR, I'm on the fence with discarding this PR, combining it with something else, or just merging it with some tweaks.

PR #3169, opened 2024-09-18
Add extra assertions to `tl:optional`

powerboat9:
This isn't producing any extra errors on my end, but the extra checks should at least be good to have.


powerboat9:
Forgot the UB was only present in #3150, this is probably applicable there

CohenArthur (approved):
thanks! looks good!

PR #3181, opened 2024-09-23
Improve handling of types in `DefaultResolver`

powerboat9:


gcc/rust/resolve/rust-default-resolver.cc
```diff
@@ -645,16 +645,24 @@ DefaultResolver::visit (AST::TraitObjectTypeOneBound &)
 {}
 
 void
-DefaultResolver::visit (AST::TupleType &)
-{}
+DefaultResolver::visit (AST::TupleType &type)
```
P-E-P, line 648: If those types simply require a visit of their subcomponent we might simply remove those functions definition and make use of their parent DefaultVisitor::visit instead of overloading everything with empty function. Unless I'm missing anything ?

P-E-P (commented):
Correct me if I'm wrong but this file looks like it could be cut in half by removing all those empty definitions.

gcc/rust/resolve/rust-default-resolver.cc
powerboat9: Alright, yeah, that works a lot better

P-E-P (approved):
Wonderful, thank you!

powerboat9:
Looks like this failed due to CI issues, can you try submitting it again?

PR #3188, opened 2024-10-04
Improve `Rib::Definition` shadowing

powerboat9:
This modifies `Rib::Definition` in order to better handle shadowing. It splits shadowing declarations into `shadowing` and `globbed` declarations, as shadowing declarations must shadow non-shadowing declarations, but non-shadowing declarations must shadow "globbed" declarations.


P-E-P (approved):
Great work!

PR #3191, opened 2024-10-10
Improve handling of `InlineAsm` in `DefaultASTVisitor`

powerboat9:


P-E-P (approved):

gcc/rust/ast/rust-ast-visitor.cc
```diff
@@ -675,7 +675,46 @@ DefaultASTVisitor::visit (AST::AsyncBlockExpr &expr)
...
-{}
+{
+  visit_outer_attrs (expr);
+  using RegisterType = AST::InlineAsmOperand::RegisterType;
+  for (auto &operand : expr.get_operands ())
+    {
+      switch (operand.get_register_type ())
+	{
+	  case RegisterType::In: {
+	    operand.get_in ().expr->accept_vis (*this);
```
P-E-P, line 686: Using `visit` directly would be more in line with the rest of the file and shorter but I don't mind merging this as is.

P-E-P:
Argh, I didn't know `InlineAsm` did not expose a getter to a reference. We should probably add
```cpp
const Expr& In::get_expr() const {
    return *expr;
}
```

powerboat9:
This should work now that #3193 is merged

P-E-P (approved):

PR #3200, opened 2024-10-16
Disambiguate generic args during name resolution 2.0

powerboat9:


gcc/rust/resolve/rust-late-name-resolver-2.0.cc
```diff
@@ -262,5 +262,36 @@ Late::visit (AST::StructExprStructFields &s)
...
+{
+  for (auto &lifetime : args.get_lifetime_args ())
+    visit (lifetime);
+
+  for (auto &generic : args.get_generic_args ())
+    visit (generic);
+
+  for (auto &binding : args.get_binding_args ())
+    visit (binding);
+}
```
P-E-P, line 276: Doesn't the default visitor already visit generic args ?

powerboat9: Kind of, from what I can tell `GenericArg::accept_vis` just visits the properties of `GenericArg` and there isn't a virtual method for `visit (GenericArg &)`. Therefore, in order to get the `GenericArgs` visitor to call into the non-virtual `GenericArg` visitor, the former has to be overridden along with the latter.

powerboat9: I've rebased and added a comment to the `GenericArgs` visitor method.

P-E-P (approved):

PR #3203, opened 2024-10-17
Improve path handling while testing name resolution 2.0

powerboat9:


gcc/testsuite/rust/compile/nr2/compile.exp
```diff
@@ -44,23 +44,23 @@ namespace eval rust-nr2-ns {
...
-	    if {$test_dir == "."} {
-		set test_lbl $test
-	    } else {
-		set test_lbl "$test_dir/$test"
-            }
+        set directory [list {*}[file split $srcdir] {*}[file split $subdir]]
+	set directory [lreplace $directory end end]
+	set directory [list {*}$directory {*}$test_dir]
+        foreach test [lsort [glob -nocomplain -tails -directory [file join {*}$directory] *.rs]] {
+	    # use '/' as the path seperator here, for exclude file entries
```
CohenArthur, line 57:
```suggestion
	    # use '/' as the path seperator here, to exclude file entries
```

CohenArthur (approved):
one nit but otherwise looks good

gcc/testsuite/rust/compile/nr2/compile.exp
powerboat9: Ah, "exclude file entries" is the noun. I should probably quote `exclude` to make that more clear

CohenArthur: ah, gotcha! sorry I misread it

PR #3208, opened 2024-10-17
Modify test `cfg1.rs` for name resolution 2.0

powerboat9:
This PR tweaks `cfg1.rs` so that it passes with name resolution 2.0. As an alternative we could modify the error message emitted by name resolution 2.0 to match that emitted by name resolution 1.0, or to match that emitted by rustc.


powerboat9:
Going to turn this into a draft for a bit, for some tweaks

powerboat9:
Resolved by https://github.com/Rust-GCC/gccrs/pull/3763

PR #3211, opened 2024-10-21
Make `const` references to `ForeverStack` more useful

powerboat9:
Adds `const` qualified member function overloads to `ForeverStack`


CohenArthur (approved):
thanks @powerboat9, looks good!

PR #3212, opened 2024-10-22
Add dump option to show node

Kamiinarii78:
Add a new dump option to show node node as
internal comment with a blacklist to ignore some
node.

gcc/rust/ChangeLog:

	* ast/rust-ast-collector.cc (TokenCollector::begin_internal_comment): Add internal comment to print node beginning. (TokenCollector::end_internal_comment): Add internal comment to print node end. (TokenCollector::visit): Add the comments of the node visited. (TokenCollector::visit_closure_common): Likewise. (TokenCollector::visit_loop_common): Likewise.
	* ast/rust-ast-collector.h: Add internal comment as a nes Kind.
	* ast/rust-ast-dump.cc (Dump::Dump): add a Dump constructor to enable internal.
	* ast/rust-ast-dump.h: Add printing of internal comment in the dump
	* rust-session-manager.cc (Session::enable_dump): Activate ast dump and fill the blacklist. (Session::handle_internal_blacklist): Parse the flag to get node to be blacklisted. (Session::compile_crate): Launch ast dump internal when asked. (Session::dump_ast_pretty_internal): Call the visitor to dump the internals.
	* rust-session-manager.h (struct CompileOptions): add Interal in Dump option enum.


gcc/rust/ast/rust-ast-collector.h
```diff
@@ -23,6 +23,7 @@
 #include "rust-ast-visitor.h"
 #include "rust-ast.h"
 #include "rust-ast-full.h"
+#include <sstream>
```
P-E-P, line 26: You probably want to include "rust-system.h" instead.

gcc/rust/ast/rust-ast-collector.cc
```diff
@@ -125,22 +144,30 @@ TokenCollector::visit (FunctionParam &param)
...
+  begin_internal_comment ("VariadicParam");
+
   if (param.has_pattern ())
     {
       visit (param.get_pattern ());
       push (Rust::Token::make (COLON, UNDEF_LOCATION));
     }
   push (Rust::Token::make (ELLIPSIS, UNDEF_LOCATION));
+
+  end_internal_comment ("VariadicParam");
```
CohenArthur, line 163: This pattern is nice but if forces you to write the node's "name" twice - once when you begin the comment, and once when you end it. It would be nicer and easier to maintain if this was wrapped in a lambda of some sort, e.g.:

```c++
internal_comment("VariadicParam", []() {
      if (param.has_pattern ())
    {
      visit (param.get_pattern ());
      push (Rust::Token::make (COLON, UNDEF_LOCATION));
    }
  push (Rust::Token::make (ELLIPSIS, UNDEF_LOCATION));
  });
  ```
  
  and the code for `internal_comment` could be something like the following:
  
  ```c++
  template <typename T> internal_comment(std::string node_name, std::function<void(T& node)> visitor) {
      begin_internal_comment(node_name);
      
      visitor();
      
      end_internal_comment(node_name);
}
```

note that the ergonomics might not be amazing, in which case we can look at something else like a custom destructor like what @jdupak did recently (we talked about it a little bit on Zulip)

gcc/rust/ast/rust-ast-collector.h
```diff
@@ -33,13 +34,16 @@ class CollectItem
...
     Comment,
+    InternalComment,
     Newline,
     Indentation,
     Token,
   };
 
   CollectItem (TokenPtr token) : token (token), kind (Kind::Token) {}
-  CollectItem (std::string comment) : comment (comment), kind (Kind::Comment) {}
+  CollectItem (std::string comment, bool internal = false)
```
CohenArthur, line 44: I would avoid using the boolean here and expose a new enum instead, which would make the code clearer

```c++
enum class Comment {
    Regular,
    Internal,
};

CollectItem (std::string comment, Comment kind = Comment::Regular) : comment(comment), kind(kind == Comment::Internal ? Kind::InternalComment : Kind::Comment) { ... }
```

gcc/rust/ast/rust-ast-dump.h
```diff
@@ -32,6 +32,8 @@ class Dump
 {
 public:
   Dump (std::ostream &stream);
+  Dump (std::ostream &stream, bool print_internal,
+	std::vector<std::string> blacklist);
```
CohenArthur, line 36: we can use a `std::set<std::string>` here for faster lookup, and for slightly better APIs as well since we can do `blacklist.contains(...)`. also, I would call the set `forbidden` or `excluded` instead of `blacklist`

gcc/rust/rust-session-manager.cc
```diff
@@ -57,6 +58,8 @@
 #include "selftest.h"
 #include "tm.h"
 #include "rust-target.h"
+#include <sstream>
+#include <vector>
```
CohenArthur, line 63: These includes are prohibited here for obscure bootstrapping compiler reasons, but you can include `rust-system.h` instead

gcc/rust/rust-session-manager.cc
```diff
@@ -386,6 +411,20 @@ Session::enable_dump (std::string arg)
   return true;
 }
 
+/* Helper function to parse a string when dump internal to get node to blacklist
+ */
+
+void
+Session::handle_internal_blacklist (std::string arg)
+{
+  std::istringstream blist_str (arg.substr (arg.find (":") + 1, 50));
```
CohenArthur, line 420: what is 50 here? can we use a define or constant instead?

gcc/rust/rust-session-manager.cc
```diff
@@ -999,6 +1042,26 @@ Session::dump_ast_pretty (AST::Crate &crate, bool expanded) const
...
+  if (out.fail ())
+    {
+      rust_error_at (UNKNOWN_LOCATION, "cannot open %s:%m; ignored",
+		     kASTDumpFile);
+      return;
+    }
+
+  std::vector<std::string> str_tmp = options.get_blacklist ();
+
+  AST::Dump (out, true, str_tmp).go (crate);
```
CohenArthur, line 1060: same here, ideally we'd use something other than `true` like an enum

gcc/rust/rust-session-manager.h
```diff
@@ -223,10 +223,15 @@ struct CompileOptions
...
     HIR_DUMP_PRETTY,
     BIR_DUMP,
+    INTERNAL_DUMP,
   };
 
   std::set<DumpOption> dump_options;
 
+  /* List of node that is not print during the dump of the ast with internal
+   * comment */
+  std::vector<std::string> internal_blacklist;
```
CohenArthur, line 233:
```suggestion
  std::set<std::string> excluded_nodes;
```

CohenArthur (commented):
This is great work!! I've added a couple comments but they're mostly nitpick. The feature in itself is solid and you did good :) Thanks!!!!

gcc/rust/rust-session-manager.cc
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2020-2024 Free Software Foundation, Inc.
+// Copyright (C) 2020 - 2024 Free Software Foundation, Inc.
```
philberty, line 1: dont change this. This gets auto updated with a script

gcc/rust/rust-session-manager.cc
```diff
@@ -373,6 +376,27 @@ Session::enable_dump (std::string arg)
     {
       options.enable_dump_option (CompileOptions::BIR_DUMP);
     }
+  else if (!arg.compare (0, 8, "internal"))
+    {
+      if (arg.size () == 8)
+	{
```
philberty, line 382: no need for 2nd level of braces

gcc/rust/rust-session-manager.cc
```diff
@@ -373,6 +376,27 @@ Session::enable_dump (std::string arg)
...
+	  options.enable_dump_option (CompileOptions::INTERNAL_DUMP);
+	}
+      else
+	{
+	  if (arg[8] != ':')
```
philberty, line 387: that could just be an else if

gcc/rust/rust-session-manager.cc
```diff
@@ -387,6 +411,22 @@ Session::enable_dump (std::string arg)
   return true;
 }
 
+/* Helper function to parse a string when dump internal to get node to blacklist
```
philberty, line 414: do // style comments please

philberty (changes requested):
some changes still needed

P-E-P:
Superseded by  #4326

PR #3216, opened 2024-10-24
Add a newline to the end of `nr2/exclude`

powerboat9:


P-E-P (approved):
Should we also put a comment as in some autoconf file in gcc to tell people to not delete the last line ?

PR #3218, opened 2024-10-25
[fork update] 2024-10-24

CohenArthur:
`git merge gcc/trunk && git checkout --ours gcc/rust libgrust`

- [ ] There are new warnings about narrowing conversions in the borrow checker interface (C++)
- [x] Some APIs have been changed


tschwinge:
> `git merge gcc/trunk && git checkout --ours gcc/rust libgrust`

You can't just drop all upstream changes to `gcc/rust/`, `libgrust/`, but have to review them individually.

tschwinge:
I suppose it'd help if I wrote down how I'm doing these merges.  (..., and sorry for my recent non-activity...)

CohenArthur:
> I suppose it'd help if I wrote down how I'm doing these merges. (..., and sorry for my recent non-activity...)

no worries Thomas! I know you've been busy and gccrs shouldn't be a priority so it's perfectly understandable :) but yes, it'd be extremely helpful if you could write down your process for these.

and btw, I don't think this dropped all changes to gcc/rust - IIUC it only checked out our versions in the case of merge conflicts, which should be okay? this branch has changes that were made upstream but not merged here

CohenArthur:
so do you take all commits that affect `gcc/rust | libgrust | gcc/testsuite/rust` from `gcc/trunk` and merge them individually? how do you apply the rest of the non-Rust commits?

dkm:
> > `git merge gcc/trunk && git checkout --ours gcc/rust libgrust`
> 
> You can't just drop all upstream changes to `gcc/rust/`, `libgrust/`, but have to review them individually.

I may have suggested this to @CohenArthur sorry. If we track upstream changes and correctly apply them, that should work (and the traffic from upstream is rather low...).

Do you see other issues with this? We're using something similar (of course with some difference) for the Ada frontend.

dkm:
> and btw, I don't think this dropped all changes to gcc/rust - IIUC it only checked out our versions in the case of merge conflicts, which should be okay? this branch has changes that were made upstream but not merged here

I think it does drop all changes. It would only drop in case of conflict if you provide a path to a conflicting file. If you give it a path, it will checkout recursively from there.

CohenArthur:
I'm not sure there's an easy way to find all commits which were applied upstream but not submitted here... we need to do a better job of applying them here as soon as they come in upstream I guess. I'll try and find a list of them

dkm:
> I'm not sure there's an easy way to find all commits which were applied upstream but not submitted here... we need to do a better job of applying them here as soon as they come in upstream I guess. I'll try and find a list of them

If you have the last commit upstreamed, you should be able to compare the current upstream gcc tree with the old gccrs tree. Or is it too naive?

CohenArthur:
I think that would work, I was looking at all the differences between the two branches but yeah I should probably look at the differences *until* the last upstreamed commit

CohenArthur:
Alright, this is completely different now. I took all of the commits present upstream but not in GitHub, and cherry-picked them separately and fixed all the merge conflicts (by either `checkout`-ing upstream or our branch depending on the merge conflict's location). Then, I merged `gcc/trunk` into this branch, and added one commit to fix the very few issues remaining due to API changes upstream and slightly wrong merges. That last commit should probably instead be split up and `--fixup`ed into the original commits, as if they had been committed with a more updated GCC base.

Does that seem alright?

dkm:
> Alright, this is completely different now. I took all of the commits present upstream but not in GitHub, and cherry-picked them separately and fixed all the merge conflicts (by either `checkout`-ing upstream or our branch depending on the merge conflict's location). Then, I merged `gcc/trunk` into this branch, and added one commit to fix the very few issues remaining due to API changes upstream and slightly wrong merges. That last commit should probably instead be split up and `--fixup`ed into the original commits, as if they had been committed with a more updated GCC base.
> 
> Does that seem alright?

sounds good, I think. Yes, the last part should be fixedup if you don't want to resolve the conflict again when sending those back in GCC.... But it means rewriting history, which is probably not a good idea. Don't know how @tschwinge handled that ? Or maybe there was no such changes?

CohenArthur:
Yeah, I was thinking of rewriting our history slightly. It's only one or two commits so I'm thinking it's okay. I don't know if this is what we did as well, but it would avoid build failures upstream for these commits

CohenArthur:
But in order to autosquash the fixup I'd have to rebase on a veeeeeeeery old commit

tschwinge:
Yeah, rewriting history for commits that have already `git fetch`ed by others is going to cause a lot of pain, please don't do that.

tschwinge:
> Alright, this is completely different now. I took all of the commits present upstream but not in GitHub, and cherry-picked them separately and fixed all the merge conflicts (by either `checkout`-ing upstream or our branch depending on the merge conflict's location). Then, I merged `gcc/trunk` into this branch, and added one commit to fix the very few issues remaining due to API changes upstream and slightly wrong merges. That last commit should probably instead be split up and `--fixup`ed into the original commits, as if they had been committed with a more updated GCC base.
> 
> Does that seem alright?

No, I suggest to not do it in this way: the cherry-picking will cause every upstream GCC commit to get a new SHA-1 object name (Git commit ID), and therefore be "detached" from its upstream GCC commit, which is likely to cause confusion.

Instead, what I've been doing, is as follows, quickly:

On the GCC/Rust master branch:

    $ git log --reverse --stat -p ..upstream/trunk

..., and skip through all commits, until you note something that "requires attention" for GCC/Rust.  For example, global changes that require corresponding changes, or upstream commits to `gcc/rust/` etc.  For such a commit, then merge the commit before (which should merge cleanly, typically...), and then merge the commit itself, and as part of the merge commit, fix up things.  Then proceed.  Similarly, if running into commits that GCC/Rust has upstreamed: merge the commit before (which should merge cleanly, typically...), and then merge the upstreamed commits with `git merge --ours` (as discussed elsewhere).

Yes, that's quite a bit of work, and yes, that cannot really be automated very much.

I routinely do the "note" part during my every-day catch-up with GCC upstream for my GCC work branches.  (So, I have a text file with all the GCC upstream  commits that potentially require attention.)  I have failed, in the past several months, to execute this list...  :-/

CohenArthur:
> > Alright, this is completely different now. I took all of the commits present upstream but not in GitHub, and cherry-picked them separately and fixed all the merge conflicts (by either `checkout`-ing upstream or our branch depending on the merge conflict's location). Then, I merged `gcc/trunk` into this branch, and added one commit to fix the very few issues remaining due to API changes upstream and slightly wrong merges. That last commit should probably instead be split up and `--fixup`ed into the original commits, as if they had been committed with a more updated GCC base.
> > Does that seem alright?
> 
> No, I suggest to not do it in this way: the cherry-picking will cause every upstream GCC commit to get a new SHA-1 object name (Git commit ID), and therefore be "detached" from its upstream GCC commit, which is likely to cause confusion.
> 
> Instead, what I've been doing, is as follows, quickly:
> 
> On the GCC/Rust master branch:
> 
> ```
> $ git log --reverse --stat -p ..upstream/trunk
> ```
> 
> ..., and skip through all commits, until you note something that "requires attention" for GCC/Rust. For example, global changes that require corresponding changes, or upstream commits to `gcc/rust/` etc.

Ah, I guess that part I got right then, by looking at commits which touch `gcc/rust` | `libgrust/` | `gcc/testsuite/rust/` upstream that aren't present in our fork.

> For such a commit, then merge the commit before (which should merge cleanly, typically...), and then merge the commit itself, and as part of the merge commit, fix up things. Then proceed. Similarly, if running into commits that GCC/Rust has upstreamed: merge the commit before (which should merge cleanly, typically...), and then merge the upstreamed commits with `git merge --ours` (as discussed elsewhere).

Hmmm... that is a lot of work indeed :< I'm not sure what we could do to make it easier/scriptable and do it more often.

> 
> Yes, that's quite a bit of work, and yes, that cannot really be automated very much.
> 
> I routinely do the "note" part during my every-day catch-up with GCC upstream for my GCC work branches. (So, I have a text file with all the GCC upstream commits that potentially require attention.) I have failed, in the past several months, to execute this list... :-/

No worries. Honestly, it's a huge burden that shouldn't fall on just one person - hence we should try and improve the process. This isn't something that should rely on you, or occupy a lot of your time. It's quite unfair you've had to deal with this on your own all this time!!!

thesamesam:
> Yes, that's quite a bit of work, and yes, that cannot really be automated very much.

Unfortunately, I concur that I don't see a better way of doing it yet.

dkm:
> > Yes, that's quite a bit of work, and yes, that cannot really be automated very much.
> 
> Unfortunately, I concur with there not being a better way of doing this, really.
> 
> EDIT: Or, well, you could take the Go (or IIRC D) approach, where they have a README that essentially says "don't touch this, send changes to X", and I always interpreted that as a risk of any changes being clobbered on imports.

I must mention that at AdaCore, we are using automated scripts to update several branches from upstream gcc (usually 3), do nightly builds + testsuite runs, with the same goal as Arthur: use a frontend developed on a different gcc base with latest gcc while maintaining both changes from gcc and the main dev (gccrs github). But it's true that changes from upstream to the frontend don't keep their original git rev (but everything is tracked) and there's still some manual tracking to bring changes from gcc to our internal dev branches. And it probably only works smoothly because changes to the frontend coming from upstream gcc are rare (same as gccrs) (wouldn't work for a high traffic FE, or any other central part of the compiler).

powerboat9:
It looks like upstreaming [5f0db57567e147846cd0b7aa4cb2fc8bba9208a0](https://github.com/Rust-GCC/gccrs/commit/5f0db57567e147846cd0b7aa4cb2fc8bba9208a0) with any dependents might remove one of the early merge conflicts

powerboat9:
It might be a good idea to upstream as much as we can, before trying to merge from upstream, since the former should make the latter much easier

CohenArthur:
Updated with Thomas' method:

```
$ git log --format=%h --reverse gcc/trunk ^master -- gcc/rust/ libgrust/ gcc/testsuite/rust/ > log-commits
$ set COMMIT $(cat log-commits | head -$COUNTER | tail -1)
$ set BEFORE $(git log -1 $COMMIT^1 --format=%h)
$ git merge $BEFORE 
$ git merge $COMMIT
```

PR #3220, opened 2024-10-26
Resolve instances of `TypePath` more accurately

powerboat9:
Depends on #3219

### DIFF
--- a/gcc/rust/resolve/rust-forever-stack.hxx
+++ b/gcc/rust/resolve/rust-forever-stack.hxx
@@ -21,6 +21,7 @@
 #include "rust-diagnostics.h"
 #include "rust-forever-stack.h"
 #include "rust-rib.h"
+#include "rust-unwrap-segment.h"
 #include "optional.h"
 
 namespace Rust {
@@ -389,7 +390,8 @@ ForeverStack<N>::find_starting_point (
 
   for (; !is_last (iterator, segments); iterator++)
     {
-      auto &seg = *iterator;
+      auto &outer_seg = *iterator;
+      auto &seg = unwrap_type_segment (outer_seg);
       auto is_self_or_crate
 	= seg.is_crate_path_seg () || seg.is_lower_self_seg ();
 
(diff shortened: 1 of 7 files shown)


gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -365,6 +365,52 @@ check_leading_kw_at_start (const S &segment, bool condition)
...
+};
+
+template <class T> class unwrap_type_segment_inner
+{
+public:
+  using ret = AST::PathIdentSegment;
+
+  static ret &unwrap (T &x) { return x.get_ident_segment (); }
+  static const ret &unwrap (const T &x) { return x.get_ident_segment (); }
+};
```
CohenArthur, line 412: is this really needed? I can't really wrap my head around that code or what code it "replaces" if that makes sense. it seems a bit hard to maintain, but if there's a lot to gain from using it I'm happy with it.

powerboat9: It's supposed to help handle different segment types being passed to `resolve_path`, by converting segment references down into `PathIdentSegment` or `SimplePathSegment` references. I suppose I could try to reduce it down to a couple of templated function overloads

powerboat9: When I try to reduce it down to function overloads I keep getting linker errors, since some of the functions aren't templates, and everything's defined in an `.hxx` file (included in multiple places)

powerboat9: I've added some commits to try to make things more clear

powerboat9:
Should I try to edit this, or keep it as-is?

philberty (changes requested):
I feel like this class is unrelated to the forever stack. I think this should be in its own file

powerboat9:
> I feel like this class is unrelated to the forever stack. I think this should be in its own file

Fixed, and now depends on #3377

P-E-P (approved):
Looks great!

PR #3221, opened 2024-10-26
Resolve `SelfParam` in name resolution 2.0

powerboat9:


CohenArthur (approved):
thanks! I think the location being off is not a super big deal, we should eventually fix it obv but that's okay

powerboat9:
Should be fixed now

PR #3224, opened 2024-10-27
Use name resolver 2.0 for module descendance checks

powerboat9:


gcc/rust/resolve/rust-forever-stack.h
```diff
@@ -521,6 +521,12 @@ template <Namespace N> class ForeverStack
 
   std::string as_debug_string ();
 
+  /**
+   * Used to check if a module is a decendent of another module
+   * Intended for use in the privacy checker
+   */
+  bool is_module_decendent (NodeId parent, NodeId child) const;
```
CohenArthur, line 528: typo
```suggestion
  bool is_module_descendant (NodeId parent, NodeId child) const;
```

but also, I'd be happy with something like `is_module_child` which seems a bit simpler

CohenArthur (approved):

gcc/rust/resolve/rust-forever-stack.h
powerboat9: I thought I had it spelled right, but it looks like I [didn't](https://www.merriam-webster.com/dictionary/decedent)

powerboat9: Should be fixed now

CohenArthur (approved):
thanks!

PR #3225, opened 2024-10-27
Improve handling of struct expressions in nr2.0

powerboat9:


CohenArthur (commented):
this looks good - does it fix a testcase or something you had? If so, please add it as a testcase in this PR

powerboat9:
I've been grabbing changes from a development branch I've been working on, and this one came later in the commit history, so it doesn't fix any test cases as-is.

PR #3226, opened 2024-10-27
Fix variable shadowing in late resolution 2.0

powerboat9:


P-E-P (changes requested):
Ok, I had trouble understanding why the order was important here. For reference in the future here is an extract of what this PR is fixing:
`issue-1930.rs`:
```rust
// { dg-options "-w" }
#[lang = "sized"]
pub trait Sized {}

fn test<T>(x: *mut T) {
    let x = x as *mut u8;
}

```
`issue-1813.rs`:
```rust
fn main() {
    let a = 15u8;
    let a = &a;
    match a {
        &15 => {}
        &14 => {}
        _ => {}
    }
}
```

error:
```
issue-1813.rs:3:14: error: cannot find value ‘a’ in this scope [E0425]
    3 |     let a = &a;
      |              ^
crab1: internal compiler error: in TyVar, at rust/typecheck/rust-tyty-util.cc:31
```

Both those issues shadow a variable and visiting the pattern before the expression makes the new name resolution ICE because the identifier refers to the new binding being created and not the one being shadowed.

Whilst I do think it is indeed the correct fix, I'm unsure we should change the default visitor for this. This problem is solely tied to name resolution, I think we need to overload the default resolver in the name resolution visitor so no other part of the code will be affected by this change (I guess people expect the default visitor to visit the items in order as in the code ?) and put a comment to show future contributor the intents of this weird order so it isn't too confusing.

powerboat9:
Alright, should be addressed

P-E-P (approved):

PR #3232, opened 2024-10-28
Make `TyTy::TupleType::get_unit_type` cache its return value

powerboat9:
This removes a call to `Resolver::get_unit_type_node_id`, which relies on the old resolver calling `Resolver::set_unit_type_node_id`

### DIFF
--- a/gcc/rust/resolve/rust-late-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-late-name-resolver-2.0.cc
@@ -100,7 +100,7 @@ Late::setup_builtin_types ()
     }
 
   // ...here!
-  auto *unit_type = TyTy::TupleType::get_unit_type (next_hir_id ());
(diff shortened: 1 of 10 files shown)


P-E-P:
So here we've got the same HIR id for all unit types. It seems fins but I do not know all implications so I'll let @philberty review

philberty (approved):
LGTM yes this is actually the right thing to do.

Historically i wasnt sure how the type system was going to work at all from reading lots of rustc. But this now makes unit-type work like the other builtins like i32.

The idea of the reference is that we have an hir node that reflects this type at this node-id. But when it comes to referencing it we update the reference but the ty_ref references the original declaration. We use this to manage generic parameter types because we can then simply replace the ty-ref on the type to point to what the new parameter points to which is still super scary to me and still suprises me that it seems to be working at all but i do have sleepless nights over it...

CohenArthur (commented):
lgtm

PR #3236, opened 2024-11-01
Remove usage of `Resolver::get_builtin_types`

powerboat9:
This simplifies `Context::setup_builtins` and removes a direct dependency on the old name resolver


philberty (approved):
LGTM

PR #3237, opened 2024-11-04
gccrs: fix bad type inference on local patterns

### LINKED ISSUE #2323: Type inference on generic type field access
I tried this code:

```rust
pub struct S<T>(T);

pub fn foo<T>(x: T) {
    let y = S(x);
    y.0;
}
```

Instead, this happened:

```
test.rs:5:5: error: type annotations needed
```

### Meta

- git sha ```3ccf008b148455c4800f2a7c70be03e1dd708c59```

philberty:
We do not need to inject inference variables on generic patterns with generic blocks. This will just cause unconstrained inference variables as they may not unify against something.

Fixes #2323

gcc/rust/ChangeLog:

	* typecheck/rust-hir-type-check-path.cc (TypeCheckExpr::resolve_root_path): dont infer here

gcc/testsuite/ChangeLog:

	* rust/compile/issue-2323.rs: New test.


CohenArthur (commented):
LGTM!

Issue #2349, opened 2023-06-29
Self generic injection causes dump issue


P-E-P:
Dump process shall dump all generics in order to be accurate, but an implicit `Self` generic is injected during name-resolution step. This means, the dump used for metadata export does contain this implicit `Self` and cannot compile anymore.

### Code

See disabled test `trait_import0.rs`

### Meta

- 77279ee4a5a

### Error output

Failure to parse metadata.

<details><summary>Details</summary>
<p>

```
crab1: internal compiler error: in linemap_ordinary_map_lookup, at libcpp/line-map.cc:1101
0x395fae3 linemap_ordinary_map_lookup(line_maps const*, unsigned int)
	/home/patryp/Documents/gccrs/build/libcpp/../../libcpp/line-map.cc:1101
0x395f4fc linemap_lookup(line_maps const*, unsigned int)
	/home/patryp/Documents/gccrs/build/libcpp/../../libcpp/line-map.cc:1057
0x3961831 linemap_location_in_system_header_p(line_maps*, unsigned int)
	/home/patryp/Documents/gccrs/build/libcpp/../../libcpp/line-map.cc:1323
0x1609e8a in_system_header_at(unsigned int)
	../../gcc/input.h:202
0x1e094df set_inlining_locations(diagnostic_context*, diagnostic_info*)
	/home/patryp/Documents/gccrs/build/gcc/../../gcc/tree-diagnostic.cc:358
0x38e2372 get_any_inlining_info(diagnostic_context*, diagnostic_info*)
	/home/patryp/Documents/gccrs/build/gcc/../../gcc/diagnostic.cc:1222
0x38dd34c diagnostic_enabled(diagnostic_context*, diagnostic_info*)
	/home/patryp/Documents/gccrs/build/gcc/../../gcc/diagnostic.cc:1431
0x38dd606 diagnostic_report_diagnostic(diagnostic_context*, diagnostic_info*)
	/home/patryp/Documents/gccrs/build/gcc/../../gcc/diagnostic.cc:1545
0x38de7f1 diagnostic_impl(rich_location*, diagnostic_metadata const*, int, char const*, __va_list_tag (*) [1], diagnostic_t)
	/home/patryp/Documents/gccrs/build/gcc/../../gcc/diagnostic.cc:1756
0x38dfbfc error_at(unsigned int, char const*, ...)
	/home/patryp/Documents/gccrs/build/gcc/../../gcc/diagnostic.cc:2079
[...]

PR #3238, opened 2024-11-04
Improve handling of implicit `Self` parameter in AST

powerboat9:
This should partially address #2349, although I'm not sure why `trait_import_*.rs` tests are still failing


powerboat9:
Fixed the `trait_import_*` issue, should now address #2349.

P-E-P (commented):
This is how I would have done it, great changes. @philberty Do you mind taking a look ?

philberty (approved):
LGTM its been a while since i looked at the import stuff. But yeah there is a section on the Rust documentation internals that Traits used the hack that Self is an implicit generic so to make it just reuse all the normal plumping it was simpler to just inject this during name resolution at the time.

PR #3241, opened 2024-11-06
Handle `TypeParam` in late resolver 2.0

powerboat9:


powerboat9:
I think the top-level resolver *should* already be handling this, will have to look into it more

powerboat9:
I should be able to make this moot with a few different PRs aimed at the late resolver

PR #3246, opened 2024-11-10
Add ctpop intrinsics

c8ef:
Part of https://github.com/Rust-GCC/gccrs/issues/658.

This patch adds support for the `ctpop` intrinsic.

gcc/rust/ChangeLog:

        * backend/rust-compile-intrinsic.cc (ctpop_hander):

gcc/testsuite/ChangeLog:

        * rust/execute/torture/ctpop.rs: New test.

Signed-off-by: Yuao Ma <c8ef@outlook.com>

### DIFF
--- a/gcc/rust/backend/rust-compile-intrinsic.cc
+++ b/gcc/rust/backend/rust-compile-intrinsic.cc
@@ -92,6 +92,8 @@ static tree
 move_val_init_handler (Context *ctx, TyTy::FnType *fntype);
 static tree
 assume_handler (Context *ctx, TyTy::FnType *fntype);
+static tree
+ctpop_hander (Context *ctx, TyTy::FnType *fntype);
 
 enum class Prefetch
 {
@@ -241,6 +243,7 @@ static const std::map<std::string,
     {"likely", expect_handler (true)},
     {"unlikely", expect_handler (false)},
     {"assume", assume_handler},
+    {"ctpop", ctpop_hander},
 };
 
 Intrinsics::Intrinsics (Context *ctx) : ctx (ctx) {}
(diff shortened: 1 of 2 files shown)


dkm:
Hello!
Thanks for your contribution!

As we are contributing everything to GCC, we need to follow their rules wrt licensing/authorship. Can you please read https://gcc.gnu.org/dco.html and make sure you add the necessary `Signed-off-by` line in your commit log and clearly state in your PR description that you've read and agree to the conditions ? Unless you already have an entry in the dco file in upstream GCC or have filed the needed papers for the copyright assignment to the FSF.

Also, we'd need you to squash all 3 commits in a single one and add a commit description:
- what the change is doing at least
- the GNU ChangeLog part

gcc/testsuite/rust/compile/torture/intrinsics-6.rs
```diff
@@ -22,5 +24,10 @@ fn main() -> i32 {
     let zero5 = unsafe { (1 << 5) - unchecked_shl(1, 5) };
     let zero6 = unsafe { (1 >> 5) - unchecked_shr(1, 5) };
 
+    let _pop1 = ctpop(42i32);
+    let _pop2 = ctpop(42u32);
+    let _pop3 = ctpop(42i128);
+    let _pop4 = ctpop(42u128);
```
dkm, line 30: I think the test can be made to check the actual result. You could do something similar to the above, or you could even create a dedicated file that only checks ctpop.

dkm: "gcc 14" is a bit too vague, as we're really tracking gcc master branch. If the builtin is missing, then we can't really merge it now I guess.

powerboat9:
Try passing `-fdump-tree-gimple` to the compiler, and comparing the generated files with [these](https://gcc.godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1AB9U8lJL6yAngGVG6AMKpaAVxYMQAJlIOAMngMmABy7gBGmMQg0gAOqAqEtgzObh7epHEJNgIBQaEsEVHSlpjWSUIETMQEKe6ePiVlAhVVBLkh4ZHRFpXVtWkNvW2BHQVdkgCUFqiuxMjsHAD0AFQA1AAqAJ4xmKubM8SraFirCJGYpKskq7SoTOirhquYqqwx9AB0q8uLAKQaAEFAgRVgoAI6uKqYCCuBgJYBBB7A1YMdwTVa/ADsACF/gDVgTVsRMARZgxVqYwq46DYGMY4jE0LCCMAIKiWBNfgBmXGArEAETxHCmtE4AFZeJ4OFpSKhOI5QQd5hivFyeKQCJphVMANYgLlcj4G40m00ANn0nEkkq1ss4vAUIA0Gq1UzgsCQaBYMTokXIlC9PvoUWQwCkPhotAIkUdEDCtrCgSqm046sTzGImwA8mFtKVNdxeF62IIswxaCnpbwsFTgI4xLRHYXSFgWIZgOIqy28MSygA3TBNmUvUquaOp3jAzCiru0PBhYjJ5xYW0EYh4FgT0gD4hheKYfmYNtGOdGV18AzABQANTwmAA7lmdlL1fxBCIxOwpDJBIoVOou10Lh9HbEBTGMcw5zCR1ICmVAYmyOF7VlHd1ywGCICmRpEPsBgnBcOo9H8EZ8kKPRMkSAR%2Bk8YCKMQ9pSK6YDsPKIZqL0FjmiGBjOiiZi2IItJ%2BNaHixj4rClS/EVxRtLs5Q4VZVAADjNABaM1JFWYBkGQVYpA%2BLxVggXBCGuX5VS4CZeALLQJl1fVDVNJzjQtGdrVIKUZXkh0nRdKs3RgRAQG05B0HcGJ/Qge50GJBQFCfHDmwbBKki3MJbmQHV4oQ1Lm2LQx0HpYkaHUTDSEDX1iGCVgFmUtSNK0nS9MkAzqxJJg6GyxL1UwfAiDQvRI0YGrjAUVwqBKkBiA%2BF4YgKn9hFEcRvzfeQlDUW1gA3d5MAAMUwFK7GbPA5v5cLbSA/QRgYVBDpfSUasoKZF0fHKjvVapaDO71bWw3rlCYOKppm1Q5rwixp3zVKIAcdjJCU3w8NEsiuTFDJ4ko5JBM8eH0ayJJka6VGIasRCWj6bHogRziGHJ4Y8l4/U0YUATUhx6nuJIxnUawucWDurcWaYAguty9U10wA9zq7CXMDwdBfOpdABebPt4NF97hUtDgJQ8215IAcUcBV1mJTBFgAJXWPxVlvB9IkUlT1M0kLmoMoyTP6lU1UuZxvUq73LOs117INI1nKc7X3M83hvIsXybO1bWvFkrzkMTuztxjVLJCAA%3D%3D)

gcc/rust/backend/rust-compile-intrinsic.cc
```diff
@@ -611,6 +614,70 @@ wrapping_op_handler_inner (Context *ctx, TyTy::FnType *fntype, tree_code op)
...
+				   UNDEF_LOCATION, &tmp_stmt);
+
+  enter_intrinsic_block (ctx, fndecl, {result_variable});
+
+  // BUILTIN popcount FN BODY BEGIN
+  auto x = Backend::var_expression (x_param, UNDEF_LOCATION);
+
+  tree popcount_builtin = error_mark_node;
+
+  BuiltinsContext::get ().lookup_simple_builtin ("__builtin_popcountg",
```
dkm, line 657: why `popcountg` and not `popcount` here?

dkm: Maybe check that the builtin is correctly registered and found (rust-builtins.cc in define_builtins() and lookup_gcc_builtin.

powerboat9:
It looks like you're getting errors on the 32 bit testsuites and the gcc4.8 one

dkm:
> For the program below:
> 
> ```c
> int main() { return __builtin_popcountg((unsigned __int128)45); }
> ```
> 
> The current version of gccrs will produce the following error when using `-m32`:
> 
> ```
> demo.c: In function ‘main’:
> demo.c:1:51: error: ‘__int128’ is not supported on this target
>     1 | int main() { return __builtin_popcountg((unsigned __int128)45); }
>       |                                                   ^~~~~~~~
> ```

int128 are probably not supported for this target. You need to adjust your test. In C you'd test a macro to check support for int128, but not sure how that is supposed to work in rust.

powerboat9:
I think we should be checking `int_n_enabled_p` and calling `rust_sorry_at` if someone tries to use a `u128` when gcc doesn't support it

powerboat9:
And then once `libcore` compiles, try to get `i128`/`u128` working on more targets

powerboat9:
It's probably fine to just skip `u128`/`i128` support for now -- "perfect is the enemy of good" and all

dkm:
You don't need to close the PR everytime there's something going on. You can mark it as draft to make it clear you're still working on it.

PR #3247, opened 2024-11-11
Fix `ForeverStack::find_starting_point` output parameter

powerboat9:


gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -493,12 +494,12 @@ ForeverStack<N>::resolve_path (const std::vector<S> &segments)
   if (segments.size () == 1)
     return get (segments.back ().as_string ());
 
-  auto starting_point = cursor ();
+  std::reference_wrapper<Node> starting_point = cursor ();
```
CohenArthur, line 497: so we can't use `auto` anymore for getting the cursor?

CohenArthur (commented):
I'm not against this, but what is the issue with the current way of doing it?

gcc/rust/resolve/rust-forever-stack.hxx
powerboat9: Previously, `starting_point` was of type `Node` and not `Node &` -- so `find_starting_point` and `resolve_path` were copying nodes instead of adjusting a reference to a node

powerboat9: When I made this patch, I didn't notice that it was `auto starting_point` instead of `auto &starting_point`, so I thought that in some cases the code would clobber parts of the `ForeverStack` -- but it would probably be best to avoid the `Node` copies anyways

powerboat9: To sum it up for clarity, `find_starting_point` should probably take a reference to a reference to a `Node`, rather than a reference to a `Node`

powerboat9:
@CohenArthur @philberty thoughts on merging?

PR #3253, opened 2024-11-19
Use nr2.0 in typechecker

powerboat9:
I probably missed a few spots, but this should cover most of the typechecker


P-E-P (approved):
Looks good, missing spots will probably show up later with some failing tests.

PR #3255, opened 2024-11-20
Add a new CI step to detect raw issue references in commit messages

### LINKED ISSUE #3213: [meta] Enforce scoped `Rust-GCC/gccrs#NNN` instead of unadorned `#NNN` in Git commit logs
I had wanted to bring this up for a long time, but...  As is usually good practice (no question about that!), GCC/Rust Git commit logs sometimes use `#NNN` to refer to GitHub GCC/Rust Issue or Pull Request `#NNN`.  However, in upstream GCC, `#NNN` have a very specific meaning: to refer to GCC Bugzilla `#NNN`, so it's very confusing if GCC/Rust commits get upstreamed, where their Git commit logs contain `#NNN`, which don't refer to GCC Bugzilla, but rather to GCC/Rust Issue or Pull Request `#NNN`.  That means, for example, that GCC/Rust commit logs' `#NNN`s in several occasions got auto-linked to (totally unrelated!) upstream GCC bugs `#NNN`.  

In context of the "Sourceware Forgejo experiment", @jwakely again raised this in <https://inbox.sourceware.org/CAH6eHdTnB5Bqru=mZNzXmyu29D8PwCKj5Z1mbaAfA7rbrrDrYg@mail.gmail.com> "Linking commits to PRs/MRs":

> We have hundreds, if not thousands, of
> commits where the message refers to a Bugzilla bug as `#1234` [...]
> We should ask the gccrs devs to
> stop using Issue `#2514` to refer to issues in their external github repo.
> They could use `Rust-GCC/gccrs#1234` instead, which won't refer to some
> arbitrary `#1234` in an unrelated repo. e.g. for gccgo the commits include
> things like `"Fixes golang/go#33739"` which is fine, the ID is scoped so
> won't link to the wrong thing.

(I've added a few back-ticks to Jonathan's text, in order to stop auto-linking for those, here.  ;-)

See <https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls>.

This checking would (preferrably) go into `.github/workflows/commit-format.yml` (right?), and/or be verified when upstreaming commits.

It should highlight (or reject?) any unadorned `#N`, `PRNNN`, `PR COMPONENT/NNN`.  (Have to check the exact patterns that GCC upstream considers to "own".)  Probably is has to be "highlight" instead of generally "reject": there may be valid cases where GCC/Rust PRs contain Git commits with commit logs containing `#NNN`, such as, when cherry-picking commits from GCC upstream.

P-E-P:
Issue references shall now be used with the Rust GCC prefix in order to avoid mixing gccrs issues and GCC bugzilla PRs.

Fixes #3213.

### DIFF
--- a/.github/workflows/commit-format.yml
+++ b/.github/workflows/commit-format.yml
@@ -108,3 +108,33 @@ jobs:
           done < <(git rev-list --reverse "$rev_list" )
 
           exit $retval;
+
+  check-issue-reference:
+    runs-on: ubuntu-latest
+    continue-on-error: true # We do not want to block merge if it is a legitimate GCC bugzilla reference.
+    name: check-issue-reference
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+
+      - name: Check for issue number reference in commit messages
+        run: |
+          retval=0;
+          rev_list="origin/${{ github.event.pull_request.base.ref }}..${{ github.event.pull_request.head.sha }}"
+          for commit in $(git rev-list --reverse "$rev_list"); do
+            if [ "$(git log --format=%B -n 1 \"$commit\" | grep '#[0-9]*' | grep -v -i 'Rust-GCC/gccrs#[0-9]*' | wc -l)" -ne 0 ]; then
+              echo "$commit: KO"
+              retval=1
+            else
+              echo "$commit: OK"
+            fi
+          done;
+          if [ "$retval" -ne 0 ]; then
+            echo "Some raw issue references were found (eg. #4242)."
+            echo "You shall rewrite the faulty commit message with this format: Rust-GCC/gccrs#4242"
+            echo "You may ignore this CI step if it represents a valid GCC bugzilla or external repository reference instead."
+          fi
+          exit $retval;


.github/workflows/commit-format.yml
```diff
@@ -108,3 +108,33 @@ jobs:
...
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+
+      - name: Check for issue number reference in commit messages
+        run: |
+          retval=0;
+          rev_list="origin/${{ github.event.pull_request.base.ref }}..${{ github.event.pull_request.head.sha }}"
+          for commit in $(git rev-list --reverse "$rev_list"); do
+            if [ "$(git show "$commit" | grep '#[0-9]*' | grep -v -i 'Rust GCC#[0-9]*' | wc -l)" -ne 0 ]; then
```
dkm, line 128:
```suggestion
            if [ "$(git show "$commit" | grep '#[0-9]*' | grep -v -i 'Rust-GCC/gccrs#[0-9]*' | wc -l)" -ne 0 ]; then
```

dkm (changes requested):

.github/workflows/commit-format.yml
P-E-P: Should we enforce the case too ?

dkm: I don't think so, both are still matched by github. Rust-GCC/gccrs#3213 rust-gcc/GcCrS#3213

.github/workflows/commit-format.yml
```diff
@@ -108,3 +108,33 @@ jobs:
...
+              echo "$commit: KO"
+              retval=1
+            else
+              echo "$commit: OK"
+            fi
+          done;
+          if [ "$retval" -ne 0 ]; then
+            echo "Some raw issue references were found (eg. #4242)."
+            echo "You shall rewrite the faulty commit message with this format: Rust-GCC/gccrs#4242"
+            echo "You may ignore this CI step if it represents a valid GCC bugzilla reference instead."
```
dkm, line 138: I would rephrase the exception  to also make it explicitly OK to ignore failure for correctly adorned external reference (following your answer)

dkm (approved):
I think it's good now!

philberty:
Merge this so you can try this out one me in this pr https://github.com/Rust-GCC/gccrs/pull/3256

philberty (approved):
LGTM great job

.github/workflows/commit-format.yml
```diff
@@ -108,3 +108,33 @@ jobs:
...
+            if [ "$(git log --format=%B -n 1 \"$commit\" | grep '#[0-9]*' | grep -v -i 'Rust-GCC/gccrs#[0-9]*' | wc -l)" -ne 0 ]; then
+              echo "$commit: KO"
+              retval=1
+            else
+              echo "$commit: OK"
+            fi
+          done;
+          if [ "$retval" -ne 0 ]; then
+            echo "Some raw issue references were found (eg. #4242)."
+            echo "You shall rewrite the faulty commit message with this format: Rust-GCC/gccrs#4242"
```
CohenArthur, line 137:
```suggestion
            echo "Rewrite the faulty commit message with this format: Rust-GCC/gccrs#4242"
```

.github/workflows/commit-format.yml
```diff
@@ -108,3 +108,33 @@ jobs:
...
+            echo "You may ignore this CI step if it represents a valid GCC bugzilla or external repository reference instead."
```
CohenArthur, line 138:
```suggestion
            echo "You may ignore this CI step if it represents a valid GCC bugzilla reference instead."
```

CohenArthur (approved):
thanks! I think that looks good and I like it

PR #3257, opened 2024-11-21
Prevent execution of some nr1.0 functions with nr2.0

powerboat9:
This should make some common dependencies on the old resolver easier to debug


P-E-P (approved):
Great idea

CohenArthur (approved):
thanks, I like that idea!

PR #3262, opened 2024-11-22
CI: upgrade GCC baseline from 4.8 -> 5.4 (5.5)

thesamesam:
GCC's baseline was changed from C++11 -> C++14 and GCC 4.8 -> GCC 5.4
in r15-4719-ga9ec1bc06bd3cc, so adjust the baseline CI job accordingly.

One quirk here is that the Ubuntu image we're using in CI has 5.5.0
rather than 5.4.0. But I think we can eat that difference.

ChangeLog:
        * .github/workflows/ccpp.yml: Use GCC 5.5 instead of GCC 4.8.
        * .github/bors.toml: Update CI job name.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -266,7 +266,7 @@ jobs:
               exit 0; \
             fi
 
-  build-and-check-gcc-48:
+  build-and-check-gcc-5:
 
     runs-on: ubuntu-22.04
     container: ubuntu:18.04
@@ -291,10 +291,10 @@ jobs:
                   libmpfr-dev \
                   libmpc-dev \
                   build-essential \
-                  gcc-4.8 \
-                  g++-4.8 \
-                  gcc-4.8-multilib \
-                  g++-4.8-multilib \
+                  gcc-5 \
+                  g++-5 \
+                  gcc-5-multilib \
+                  g++-5-multilib \
                   dejagnu;
           # install Rust directly using rustup
           curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain=1.72.0;
@@ -307,8 +307,8 @@ jobs:
            mkdir -p gccrs-build;
            cd gccrs-build;
            ../configure \
-               CC='gcc-4.8' \
-               CXX='g++-4.8' \
+               CC='gcc-5' \
+               CXX='g++-5' \
                --enable-languages=rust \
                --disable-bootstrap \
                --enable-multilib
@@ -328,7 +328,7 @@ jobs:
     - name: Archive check-rust results
       uses: actions/upload-artifact@v3
       with:
-        name: check-rust-logs-4.8
+        name: check-rust-logs-5
         path: |
           gccrs-build/gcc/testsuite/rust/
 


tschwinge:
Great, thanks!  I was meaning to file an Issue for this -- now we've got a Pull Request already!  :-D

.github/workflows/ccpp.yml
```diff
@@ -266,7 +266,7 @@ jobs:
               exit 0; \
             fi
 
-  build-and-check-gcc-48:
+  build-and-check-gcc-54:
```
tschwinge, line 269: I thought that `gcc-48` already was a bit unclear; can we just make that `gcc-5` now?

Do we need to accordingly update `build-and-check-gcc-48` in `.github/bors.toml`?

thesamesam: Nice catch on bors!

.github/bors.toml
dkm, line 1: I think this file should go away, we're not using bors anymore. You can at least drop the change, as it may gives the false impression that it is maintained and used. I'll remove it in another PR.

.github/workflows/ccpp.yml
dkm: Ah, missed this comment. No sure how I'm managed this, sorry. I think bors.yaml can go away.

.github/bors.toml
dkm: ugh, I think the CONTRIBUTING.md should be updated as well...

dkm:
I've removed all bors mention/files in https://github.com/Rust-GCC/gccrs/pull/3263

.github/bors.toml
tschwinge: Aha, sorry for the back-and-forth work then, @thesamesam.

.github/workflows/ccpp.yml
```diff
@@ -266,7 +266,8 @@ jobs:
               exit 0; \
             fi
 
-  build-and-check-gcc-48:
+  # This job name must be kept in sync with .github/bors.toml.
```
powerboat9, line 269: Should probably remove the bors mention here

thesamesam: Yeah, was just waiting for the other PR to land so I could rebase.

powerboat9:
It's not auto-merging, seemingly because build-and-check-gcc-48 no longer exists

tschwinge:
@powerboat9: Yeah, probably.  Where does it get that from, do you (or anyone else, of course) happen to know?

I could easily push this manually, but not sure I should?  Maybe wait for the pros to be back on Monday.  ;-)

dkm:
> @powerboat9: Yeah, probably. Where does it get that from, do you (or anyone else, of course) happen to know?
> 
> I could easily push this manually, but not sure I should? Maybe wait for the pros to be back on Monday. ;-)

Looks like a known issue when renaming names in workflows. `build-and-check-gcc-5` is OK, so we can probably force the merge.

dkm:
We may need the super powers of @CohenArthur for really merging this...

tschwinge:
Given we all agree, I went ahead and used my superpowerz -- let's have Arthur use his time for more interesting things.  :+1:

philberty:
Here this doesnt seem to be working for me on new PR's see: https://github.com/Rust-GCC/gccrs/actions/runs/12019251579/job/33527105478?pr=3267

powerboat9:
I think I remember seeing the gcc-4.8 failing in a similar way earlier. It looks like, as part of an automatic upgrade from node16 to node20 (?), the version of glibc used by ubuntu 18.04 is too old for node to use.

https://github.com/nodejs/node/blob/c2e4b1fa9ad0b744616c4e4c13a5017772a630c4/BUILDING.md#platform-list

PR #3265, opened 2024-11-25
Prepend crate name to functions with nr2

P-E-P:
Some pre nr2 were relying on function name, those name are prefixed with the crate's name.

### DIFF
--- a/gcc/rust/backend/rust-compile-base.cc
+++ b/gcc/rust/backend/rust-compile-base.cc
@@ -25,7 +25,8 @@
 #include "rust-compile-type.h"
 #include "rust-constexpr.h"
 #include "rust-diagnostics.h"
-#include "rust-expr.h"	// for AST::AttrInputLiteral
+#include "rust-expr.h" // for AST::AttrInputLiteral
+#include "rust-hir-map.h"
 #include "rust-macro.h" // for AST::MetaNameValueStr
 #include "rust-hir-path-probe.h"
 #include "rust-type-util.h"
@@ -38,6 +39,9 @@
 #include "attribs.h"
 #include "tree.h"
 #include "print-tree.h"
+
+// rust-name-resolution-2.0
+#include "options.h"
 
 namespace Rust {
 namespace Compile {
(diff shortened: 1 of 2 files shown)


powerboat9:
I'm pretty sure this is an issue with how canonical paths are generated -- it should probably be fixed in `ForeverStack::to_canonical_path`

P-E-P:
> I'm pretty sure this is an issue with how canonical paths are generated -- it should probably be fixed in `ForeverStack::to_canonical_path`

I'm not entirely sure because we should not be able to resolve `crate_name::XXXX`

powerboat9:
We could have all lookups stop before they reach the parent of a `ForeverStack` node representing a crate -- as-is, the root `ForeverStack` node represents the current crate namespace, but it would probably make things easier with extern crates and preludes if crates had their own nodes. That would make lookups work correctly (all lookups would be done relative to `my_crate::` or one of its descendants, baring edition-specific leading `::` behavior) and make the current canonical path calculation code work properly (instead of `ROOT(my_crate)->foo->bar` corresponding to `foo::bar`, `ROOT->my_crate->foo->bar` would correspond to `my_crate::foo::bar`)

powerboat9:
It does look like we extend canonical paths to be global between crates and to exist for most (all? virtually all?) items, where in `rustc` canonical paths have meaning local to a crate and only cover some items.

https://doc.rust-lang.org/reference/paths.html

We should be fine, as long as we can still map our canonical paths to `rustc`-like canonical paths when necessary.

powerboat9:
Rereading `ResolveTopLevel::go`, it looks like two canonical paths are tracked for every item, although the one that doesn't have a prepended crate name seems to only be used inside the 1.0 resolver.

P-E-P:
> Rereading `ResolveTopLevel::go`, it looks like two canonical paths are tracked for every item, although the one that doesn't have a prepended crate name seems to only be used inside the 1.0 resolver.

This is probably why it was working as intended with nr1, I'll give it a look tomorrow

CohenArthur (approved):
I think those changes are fine. My gut feeling is we probably need to do something similar for exports? But I'm not sure. I think this is a fine solution though

P-E-P:
Tbh I don't really like my own solution, initially I intended to modify the name resolution, add a name to the root node.

Cons for this method are obvious as name resolution from crate name should be prevented. I'll take a few days to give this PR some thoughts.

PR #3268, opened 2024-11-26
ci: bump actions/checkout version

dkm:
v3 looks obsolete, bump it to the current v4.


philberty (approved):
I love this.

Only question we have is do we want to make this a requirement for merging? Like we had with the gcc 4.8 one i sort of dont think its worth making it a requirement. but 99.99% of the time people will see the CI job fail and just fix it anyway

dkm:
> I love this.
> 
> Only question we have is do we want to make this a requirement for merging? Like we had with the gcc 4.8 one i sort of dont think its worth making it a requirement. but 99.99% of the time people will see the CI job fail and just fix it anyway

oh, this, I don't know. I think you're more qualified than me to decide (along with the real devs, doing actual coding)

P-E-P:
> Only question we have is do we want to make this a requirement for merging?

Why wouldn't we make a it a requirement ? It shouldnt be more than a box to check in the settings right ? I feel like I'm missing something here.

dkm:
> > Only question we have is do we want to make this a requirement for merging?
> 
> Why wouldn't we make a it a requirement ? It shouldnt be more than a box to check in the settings right ? I feel like I'm missing something here.

I think that's a setting only available to the project's admin. Maybe you are? At least @philberty is and can confirm.... But IIUC, there's nothing more to do in my PR regarding the "required" aspect of gcc-5-build.

CohenArthur (approved):
amazing work, thank you @dkm!!! ❤️❤️❤️

P-E-P (approved):

PR #3269, opened 2024-11-26
gccrs: improve handling of Self Type paths

### LINKED ISSUE #2907: Unable to call trait method as function when implemented on bound
I tried this code:

```rust
#![feature(lang_items)]

#[lang = "sized"]
pub trait Sized {}

pub trait Bar {}

pub trait Foo {
    type Ty;

    fn foo(self) -> Self::Ty;
}

impl<B: Bar> Foo for B {
    type Ty = u32;

    fn foo(self) -> Self::Ty {
        14
    }
}

struct Qux;

impl Bar for Qux {}

fn main() {
    let a = Qux;
    a.foo();

    let b = Qux;
    Foo::foo(b);
}

```

I expected to see this happen: no error. https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=41805c77628c84f5e2ccb80c4be2bfe6

Instead, this happened:

```rust
trait_call.rs:17:30: error: mismatched types, expected ‘<placeholder:>’ but got ‘<integer>’ [E0308]
   17 |     fn foo(self) -> Self::Ty {
      |     ~~              ~~~~     ^
crab1: error: bounds not satisfied for dyn [Foo<Self>] ‘Sized, Bar’ is not satisfied [E0277]
trait_call.rs:31:14: error: mismatched types, expected ‘dyn [Foo<Self>]’ but got ‘Qux’ [E0308]
   17 |     fn foo(self) -> Self::Ty {
      |            ~~~~
......
   31 |     Foo::foo(b);
      |              ^
```
I'm guessing we're missing something simple? This shows up in the for-loop testcase when trying to call `IntoIterator::into_iter(1..3)` vs `(1..3).into_iter()`, since `IntoIterator` is implemented like so:

```rust
        impl<I: Iterator> IntoIterator for I {
            type Item = I::Item;
            type IntoIter = I;

            fn into_iter(self) -> I {
                self
            }
        }
```

philberty:
TypePaths have special handling for Self where we can look at the current ctx for more acurate TypeAlias information if required. We cant do this for Impl contexts but not for Traits as we might as well fall back to the TypePathProbe.

The other issue was the dyn type comming in because Foo::foo and Foo is a trait reference we represent this as a dyn type as the root resolved path but then find the associated impl block for this but we cannot do this when we have resolved to a Dyn because this is just a representation that we know we are talking about a trait not because we are actually working with a real Dyn type.

Fixes #2907

gcc/rust/ChangeLog:

	* typecheck/rust-hir-trait-resolve.cc (TraitResolver::resolve_trait): track trait
	* typecheck/rust-hir-type-check-implitem.cc: trait block
	* typecheck/rust-hir-type-check-path.cc (TypeCheckExpr::resolve_segments): dont when dyn
	* typecheck/rust-hir-type-check-type.cc (TypeCheckType::visit): look at Self contenxt (TypeCheckType::resolve_root_path): track if Self (TypeCheckType::resolve_associated_type): look at current context for associated types
	* typecheck/rust-hir-type-check-type.h: change prototype
	* typecheck/rust-hir-type-check.h (class TypeCheckBlockContextItem): new context system to track current state
	* typecheck/rust-typecheck-context.cc (TypeCheckContext::have_block_context): likewise (TypeCheckContext::peek_block_context): likewise (TypeCheckContext::push_block_context): likewise (TypeCheckContext::pop_block_context): likewise (TypeCheckBlockContextItem::Item::Item): likewise (TypeCheckBlockContextItem::TypeCheckBlockContextItem): likewise (TypeCheckBlockContextItem::is_impl_block): likewise (TypeCheckBlockContextItem::is_trait_block): likewise (TypeCheckBlockContextItem::get_impl_block): likewise (TypeCheckBlockContextItem::get_trait): likewise

gcc/testsuite/ChangeLog:

	* rust/compile/issue-2907.rs: New test.

### DIFF
--- a/gcc/rust/typecheck/rust-typecheck-context.cc
+++ b/gcc/rust/typecheck/rust-typecheck-context.cc
@@ -177,6 +177,32 @@ TypeCheckContext::peek_context ()
   return return_type_stack.back ().first;
 }
 
+bool
+TypeCheckContext::have_block_context () const
+{
+  return !block_stack.empty ();
+}
+
+TypeCheckBlockContextItem
+TypeCheckContext::peek_block_context ()
+{
+  rust_assert (!block_stack.empty ());
+  return block_stack.back ();
+}
+
+void
+TypeCheckContext::push_block_context (TypeCheckBlockContextItem block)
+{
+  block_stack.push_back (block);
+}
+
+void
+TypeCheckContext::pop_block_context ()
+{
+  rust_assert (!block_stack.empty ());
+  block_stack.pop_back ();
+}
+
 void
 TypeCheckContext::iterate (std::function<bool (HirId, TyTy::BaseType *)> cb)
 {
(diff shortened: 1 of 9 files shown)


gcc/rust/typecheck/rust-typecheck-context.cc
```diff
@@ -177,6 +177,32 @@ TypeCheckContext::peek_context ()
...
+  block_stack.push_back (block);
+}
+
+void
+TypeCheckContext::pop_block_context ()
+{
+  rust_assert (!block_stack.empty ());
+  block_stack.pop_back ();
+}
+
```
CohenArthur, line 205: Note that we can remove all of that code if we use the `StackedContexts` class we have: https://github.com/Rust-GCC/gccrs/blob/master/gcc/rust/util/rust-stacked-contexts.h

would remove some of the lines of code. I can turn this into a good-first-pr issue if you'd like

CohenArthur: #3284

PR #3271, opened 2024-11-26
Cleanup `ContextualASTVisitor`

CohenArthur:
As part of my cleanups for lang-items, which should land soon...

This uses the `StackedContexts` API which is safer than a simple vector for keeping track of context


P-E-P (approved):
LGTM

PR #3273, opened 2024-11-27
Cleanup lang items mappings

CohenArthur:
Small cleanup of the lang item mappings in preparation for the remaining lang item patches

PR #3274, opened 2024-11-27
Collect lang items during AST

CohenArthur:
In order to have lang items paths, we need to be able to resolve these lang items to their definition point which is impossible in the current system, as lang items are only saved in the mappings during lowering (and thus after name resolution). This small visitor collects lang items and stores their NodeIds in the mappings for an easy resolution.

Draft because I need to deduplicate some methods, but reviews welcome. Needs #3273 so ignore the first two commits


CohenArthur:
Closing this since I messed it up and will open another PR with the fixes

PR #3275, opened 2024-11-27
Add some additional tools to the CI

P-E-P:
Add [clang-tidy](https://clang.llvm.org/extra/clang-tidy/) and [iwyu](https://github.com/include-what-you-use/include-what-you-use)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,106 @@
+name: Additional lint
+
+on:
+  push:
+    branches:
+      - trying
+      - staging
+  pull_request:
+    branches: [ master ]
+  merge_group:
+
+jobs:
+  compilation-database:
+    env:
+      # Force locale, in particular for reproducible results re '.github/log_expected_warnings' (see below).
+      LC_ALL: C.UTF-8
+
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Install Deps
+      run: |
+          sudo apt-get update;
+          sudo apt-get install -y \
+                  automake \
+                  autoconf \
+                  libtool \
+                  autogen \
+                  bison \
+                  flex \
+                  libgmp3-dev \
+                  libmpfr-dev \
+                  libmpc-dev \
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -0,0 +1,106 @@
...
+      - trying
+      - staging
+  pull_request:
+    branches: [ master ]
+  merge_group:
+
+jobs:
+  compilation-database:
+    env:
+      # Force locale, in particular for reproducible results re '.github/log_expected_warnings' (see below).
```
dkm, line 15: this can probably go away

.github/workflows/lint.yml
```diff
@@ -0,0 +1,106 @@
...
+        with:
+          name: compilation-database
+          path: db
+      - run: |
+          run-clang-tidy -p db/compile-commands.json -style contrib/clang-format gcc/rust/*
+
+  include-what-you-use:
+    needs: compilation-database
+    env:
+      # Force locale, in particular for reproducible results re '.github/log_expected_warnings' (see below).
```
dkm, line 91: ditto :grimacing:

powerboat9:
The build probably has to be parallelized in order to complete in time

P-E-P:
> The build probably has to be parallelized in order to complete in time

Bear doesn't like parallel jobs, IIRC there is a race condition on the jobserver, compiling from scratch with only one job shouldn't take that much time, at most 1h.

Here we hit another bug in bear, which makes bear hang indefinitely even if the compilation was done.

I'll probably get rid of bear anyway, I thought it would be great to have compilation database artifacts around but they are not as useful as I initially thought and I do not want to fight bear indefinitely.

philberty:
Darn i think paid CI jobs allow us to run longer ones

P-E-P:
> Darn i think paid CI jobs allow us to run longer ones

Tbh the build is not supposed to be that long (bear waits for a job it isn't supposed to). I haven't modified this PR again since it was quite low priority compared to fixing nr2. I'll get back to it later without a compilation database this time.

PR #3277, opened 2024-12-01
Fix bug in type resolution of paths

powerboat9:
When name resolution 2.0 is enabled these sections of the typechecker should be performing resolution lookups using segment node ids, rather than path node ids, as is done when name resolution 1.0 is enabled. Since name resolution 2.0 doesn't currently insert resolutions for individual segments, lookups based on path node id are left as a fallback.


gcc/rust/typecheck/rust-hir-type-check-path.cc
```diff
@@ -218,8 +218,17 @@ TypeCheckExpr::resolve_root_path (HIR::PathInExpression &expr, size_t *offset,
...
+	    ref_node_id = resolved;
+	  });
+
+	  // HACK: on failure, redo lookup using path node id
+	  // This is incorrect, but will help some tests pass for now
+	  // TODO: remove this
+	  if (ref_node_id == UNKNOWN_NODEID)
+	    nr_ctx.lookup (expr.get_mappings ().get_nodeid ())
+	      .map (
+		[&ref_node_id] (NodeId resolved) { ref_node_id = resolved; });
```
CohenArthur, line 231: I think I would rather spend some time fixing the root cause for this. This seems a bit dodgy. what do you think is missing from nr2.0 in order to enable this? simply resolving each segment when we resolve a path?

CohenArthur (changes requested):

gcc/rust/typecheck/rust-hir-type-check-type.cc
```diff
@@ -395,10 +395,17 @@ TypeCheckType::resolve_root_path (HIR::TypePath &path, size_t *offset,
 	  auto nr_ctx
 	    = Resolver2_0::ImmutableNameResolutionContext::get ().resolver ();
 	  // assign the ref_node_id if we've found something
-	  nr_ctx.lookup (path.get_mappings ().get_nodeid ())
-	    .map ([&ref_node_id, &path] (NodeId resolved) {
-	      ref_node_id = resolved;
-	    });
+	  nr_ctx.lookup (ast_node_id).map ([&ref_node_id] (NodeId resolved) {
```
philberty, line 398: you need to be looking at the segment node_id.

The name resolution for Paths is a really huge pain in the ass. this is resolving the _ROOT_ of a path so for example you could have:

mod_a::Type::impl_item

the root path should just resolve this:

mod_a::Foo then stop then fall back to the resolve segments to then find the impl_item

So i think you should just be using the ast_node_id like i do in the old name resolver here.

gcc/rust/typecheck/rust-hir-type-check-path.cc
philberty: do your lookup on the ast_node_it which is the segment not the _path_ name resolution cant resolve a full path on its own.

philberty (approved):
this LGTM

powerboat9:
It looks like this might need something like #3220 to prevent test regressions

powerboat9:
This now removes a bunch of tests from the exclude file, but causes a regression with `issue-2070.rs`

powerboat9:
I've fixed the regression, and removed a few more tests from the exclude list

CohenArthur (approved):
LGTM, thanks a lot Owen!

P-E-P (approved):

PR #3278, opened 2024-12-02
Remove `Rust::make_unique`

powerboat9:
While minor, it looks like this fix was accidentally applied while upstreaming and was never performed downstream


powerboat9:
It looks like my `clang-format` might be misbehaving. Alternatively, we could push an inverted patch upstream, or leave upstream and downstream out of sync. A PR to remove our implementation of `std::make_unique` (now that C++14 is supported) could make the formatting moot, but would be non-trivial to upstream while we're out of sync.

philberty:
I think this is all my fault my desktop is a super old clang format 11 or something stupid iirc. What version of clang format are you using @CohenArthur and @P-E-P ? Maybe we look to upgrade and I will sort my life out lol

P-E-P:
> I think this is all my fault my desktop is a super old clang format 11 or something stupid iirc. What version of clang format are you using @CohenArthur and @P-E-P ? Maybe we look to upgrade and I will sort my life out lol

IIRC we should use clang format 10 because newer versions have some incompatibilities with their configuration.

CohenArthur (approved):
I'm happy with merging this. I think there are lots of grey areas when it comes to formatting C++ code for GCC. I'm happy as long as clang-format is happy, and variadics are the devil's weapon anyway so they don't deserve nice formatting.

CohenArthur:
Ah, but clang-format isn't happy... Hmmm. This is annoying since it's going to trigger on all future PRs, so we can't merge it. We can update our clang-format to use a more recent version, as Philip pointed out we're supposed to use 10 for the project but it's 9 versions behind.

powerboat9:
I could use `// clang-format `{`on`,`off`}, but I'd probably want to stick that in a separate comment for ease of upstreaming later

powerboat9:
I think it might just be better to remove this file, now that we have `std::make_unique` with C++14

powerboat9:
It looks like both of the failing builds are using `-std=gnu++11`

CohenArthur:
Yeah, let's wait for #3294 or for an update from upstream. I'm happy with removing our `make_unique`.

powerboat9:
Collecting some stats on gcc using a minimally efficient bash script, but I should be able to rebase in a bit

2024-12-11-162451_817x482_scrot

powerboat9:
This should work now

powerboat9:
It looks like there's an issue with the clang check

thesamesam:
```
 checking dependency style of clang++ -std=gnu++14... clang++ -std=gnu++14  -I../../libcpp -I. -I../../libcpp/../include -I/usr/local/include -I../../libcpp/include -I/usr/local/include -g -O2     -W -Wall -Wno-narrowing -Wwrite-strings -Wmissing-format-attribute -pedantic -Wno-long-long  -fno-exceptions -fno-rtti -I../../libcpp -I. -I../../libcpp/../include -I/usr/local/include -I../../libcpp/include -I/usr/local/include   -c -o charset.o -MT charset.o -MMD -MP -MF .deps/charset.Tpo ../../libcpp/charset.cc
gcc3
checking how to run the C++ preprocessor... In file included from ../../libcpp/charset.cc:21:
In file included from ../../libcpp/system.h:279:
/usr/local/include/libintl.h:1089:34: error: expected unqualified-id
char *setlocale (int __category, const char *__locale)
                                 ^
/usr/local/include/libintl.h:1089:34: error: expected ')'
/usr/local/include/libintl.h:1089:7: note: to match this '('
char *setlocale (int __category, const char *__locale)
      ^
../../libcpp/system.h:275:38: note: expanded from macro 'setlocale'
# define setlocale(category, locale) (locale)
   `                                  ^
```

We may need to install libintl from brew as well.

powerboat9:
@CohenArthur @philberty

PR #3286, opened 2024-12-03
gccrs: fix a compiler crash when path expression contains nothing

liushuyu:
gcc/rust/ChangeLog:

	* ast/rust-path.h: allows error nodes to specify the source
	location, so that it could be used in diagnostics later.
	* parse/rust-parse-impl.h: creates the PathInExpression error node
	with source location and prints proper diagnostics if the path
	is empty.

gcc/testsuite/ChangeLog:

	* rust/compile/paamayim-nekudotayim.rs: add a test for testing
	proper error recovery logic when there is no path names.

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -6713,9 +6713,9 @@ Parser<ManagedTokenSource>::parse_path_in_expression ()
   AST::PathExprSegment initial_segment = parse_path_expr_segment ();
   if (initial_segment.is_error ())
     {
-      // skip after somewhere?
-      // don't necessarily throw error but yeah
-      return AST::PathInExpression::create_error ();
+      // we can not throw an error here because if entered from macro expansion
+      // the macro expander can throw another error which might confuse the user
+      return AST::PathInExpression::create_error (locus);
     }
   segments.push_back (std::move (initial_segment));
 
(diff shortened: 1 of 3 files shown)


liushuyu:
> This should probably be handled by the parser, rather than the type checker, but good catch

I want to handle this in the parser, but there seems to be a curious comment in the parser code (https://github.com/Rust-GCC/gccrs/pull/3286/files#diff-0ec0ef9112d5a4a50636e9fde592f348f936f51b4139796976c7ad1608f935f8R6716-R6717) that states:

```c++
      // skip after somewhere?
      // don't necessarily throw error but yeah
```

powerboat9:
Yeah, that's a bit odd. It looks like that comment was introduced by [`a2ea9f7e7027a9945744f5f7c126a0618a30c334`](https://github.com/Rust-GCC/gccrs/commit/a2ea9f7e7027a9945744f5f7c126a0618a30c334) (at that point in file `gcc/rust/test3/parse/rust-parse.cc`) and it looks like according to https://doc.rust-lang.org/reference/paths.html#paths-in-expressions paths must have at least one segment. I'd think the comment could be removed -- @philberty @CohenArthur?

CohenArthur:
I agree with @powerboat9. I think this case should be handled in the parser, this is how `rustc` seems to do it at least. so I'm happy with removing the comment and handling it in the parser @liushuyu. Once this is done I think this PR is ready to go :thumbsup:

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -11637,6 +11637,11 @@ Parser<ManagedTokenSource>::parse_stmt_or_expr ()
 	AST::PathInExpression path = parse_path_in_expression ();
 	std::unique_ptr<AST::Expr> null_denotation;
 
+	if (path.is_error ())
+	  {
+	    rust_error_at (path.get_locus (), "expected identifier");
+	  }
```
CohenArthur, line 11643:
```suggestion
	    rust_error_at (path.get_locus (), "expected identifier");
```

small formatting nit :)

CohenArthur (approved):
I think we need for https://github.com/Rust-GCC/gccrs/pull/3295 to go through or this PR will also fail if it gets rebased. but I'm happy to merge it

PR #3290, opened 2024-12-04
Use lang items as paths

CohenArthur:
This PR makes it possible to use lang items as paths, for example when generating code for derives or for-loops. Instead of generating a complex path to something like `core::marker::Copy`, we can just use a `LangItemPath` to `LangItem::COPY`. This way, the code generation does not need to care about name resolution, about std's module hierarchy changing, or about duplicate definitions with the same module structure (e.g. if a user defines a copy trait in their own source file with their own `core` and `marker` module).

This is not going to build until #3289 is merged as it depends on those mappings - but it should still be reviewable

### DIFF
--- a/gcc/rust/ast/rust-item.h
+++ b/gcc/rust/ast/rust-item.h
@@ -3264,6 +3264,13 @@ public:
     return *trait_path;
   }
 
+  Type &get_trait_path_type ()
+  {
+    rust_assert (trait_path->get_path_kind () == Path::Kind::Type);
+
+    return (AST::Type &) static_cast<AST::TypePath &> (*trait_path);
+  }
+
 protected:
   /* Use covariance to implement clone function as returning this object
    * rather than base */
--- a/gcc/rust/ast/rust-item.h
+++ b/gcc/rust/ast/rust-item.h
@@ -3172,7 +3172,7 @@ class TraitImpl : public Impl
 {
   bool has_unsafe;
   bool has_exclam;
-  TypePath trait_path;
+  std::unique_ptr<Path> trait_path;
 
   // bool has_impl_items;
   std::vector<std::unique_ptr<AssociatedItem>> impl_items;
@@ -3184,7 +3184,7 @@ public:
   bool has_impl_items () const { return !impl_items.empty (); }
 
   // Mega-constructor
-  TraitImpl (TypePath trait_path, bool is_unsafe, bool has_exclam,
+  TraitImpl (std::unique_ptr<Path> trait_path, bool is_unsafe, bool has_exclam,
 	     std::vector<std::unique_ptr<AssociatedItem>> impl_items,
 	     std::vector<std::unique_ptr<GenericParam>> generic_params,
 	     std::unique_ptr<Type> trait_type, WhereClause where_clause,
@@ -3197,10 +3197,26 @@ public:
       trait_path (std::move (trait_path)), impl_items (std::move (impl_items))
   {}
 
+  // Helper constructor with a typepath
+  TraitImpl (TypePath trait_path, bool is_unsafe, bool has_exclam,
+	     std::vector<std::unique_ptr<AssociatedItem>> impl_items,
+	     std::vector<std::unique_ptr<GenericParam>> generic_params,
+	     std::unique_ptr<Type> trait_type, WhereClause where_clause,
+	     Visibility vis, std::vector<Attribute> inner_attrs,
+	     std::vector<Attribute> outer_attrs, location_t locus)
+    : Impl (std::move (generic_params), std::move (trait_type),
+	    std::move (where_clause), std::move (vis), std::move (inner_attrs),
+	    std::move (outer_attrs), locus),
+      has_unsafe (is_unsafe), has_exclam (has_exclam),
+      trait_path (std::unique_ptr<TypePath> (new TypePath (trait_path))),
+      impl_items (std::move (impl_items))
+  {}
+
   // Copy constructor with vector clone
   TraitImpl (TraitImpl const &other)
     : Impl (other), has_unsafe (other.has_unsafe),
-      has_exclam (other.has_exclam), trait_path (other.trait_path)
+      has_exclam (other.has_exclam),
+      trait_path (other.trait_path->clone_path ())
   {
     impl_items.reserve (other.impl_items.size ());
     for (const auto &e : other.impl_items)
@@ -3211,7 +3227,7 @@ public:
   TraitImpl &operator= (TraitImpl const &other)
   {
     Impl::operator= (other);
-    trait_path = other.trait_path;
+    trait_path = other.trait_path->clone_path ();
     has_unsafe = other.has_unsafe;
     has_exclam = other.has_exclam;
 
@@ -3242,10 +3258,10 @@ public:
   }
 
   // TODO: is this better? Or is a "vis_block" better?
-  TypePath &get_trait_path ()
+  Path &get_trait_path ()
   {
     // TODO: assert that trait path is not empty?
-    return trait_path;
+    return *trait_path;
   }
 
 protected:
--- a/gcc/rust/ast/rust-path.cc
+++ b/gcc/rust/ast/rust-path.cc
@@ -152,8 +152,7 @@ RegularPath::as_string () const
 std::string
 LangItemPath::as_string () const
 {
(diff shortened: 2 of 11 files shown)


gcc/rust/ast/rust-path.cc
```diff
@@ -152,8 +152,7 @@ RegularPath::as_string () const
 std::string
 LangItemPath::as_string () const
 {
-  // FIXME: Handle #[lang] paths
-  rust_unreachable ();
+  return LangItem::ToString (kind);
```
CohenArthur, line 155: This should be something like `"#[lang = { LangItem::ToString (kind) }"` instead, to make it clearer instead of just displaying `copy`

gcc/rust/ast/rust-path.h
```diff
@@ -736,10 +745,13 @@ class PathInExpression : public Pattern, public ExprWithoutBlock
...
-      return !static_cast<RegularPath &> (*path).has_segments ();
-
-    return false;
+    switch (path->get_path_kind ())
+      {
+      case Path::Kind::Regular:
+	return !static_cast<RegularPath &> (*path).has_segments ();
+      default:
+	rust_unreachable ();
+      }
```
CohenArthur, line 754: actually the old version is better, but we need to turn the `return false` into a `rust_unreachable`

gcc/rust/ast/rust-path.h
```diff
@@ -815,10 +827,14 @@ class PathInExpression : public Pattern, public ExprWithoutBlock
...
-
-    return false;
+    switch (path->get_path_kind ())
+      {
+      case Path::Kind::Regular:
+	return static_cast<RegularPath &> (*path).get_segments ().size () == 1;
+      case Path::Kind::Type:
+      case Path::Kind::LangItem:
+	rust_unreachable ();
+      }
```
CohenArthur, line 837: same here

gcc/rust/expand/rust-derive-copy.cc
```diff
@@ -44,7 +46,8 @@ DeriveCopy::copy_impl (
   // `$crate::core::marker::Copy` instead
   auto segments = std::vector<std::unique_ptr<TypePathSegment>> ();
   segments.emplace_back (builder.type_path_segment ("Copy"));
-  auto copy = TypePath (std::move (segments), loc);
+  auto copy = Rust::make_unique<LangItemPath> (LangItem::Kind::COPY, loc);
+  // auto copy = TypePath (std::move (segments), loc);
```
CohenArthur, line 50: dead code

gcc/rust/hir/rust-ast-lower-type.cc
```diff
@@ -17,14 +17,42 @@
...
+{
+  ASTLowerTypePath resolver;
+  type.accept_vis (resolver);
+  rust_assert (resolver.translated != nullptr);
+  return resolver.translated;
+}
+
+HIR::TypePath *
+ASTLowerTypePath::translate_lang_item_type_path (AST::LangItemPath &type)
 {
```
CohenArthur, line 56: I feel like this manual dispatch isn't necessary, and that we can just use dynamic dispatch here. I'll try and fix it

gcc/rust/hir/rust-ast-lower-type.cc
```diff
@@ -126,6 +154,30 @@ ASTLowerTypePath::visit (AST::TypePath &path)
...
+void
+ASTLowerTypePath::visit (AST::LangItemPath &path)
+{
+  auto crate_num = mappings.get_current_crate ();
+  auto hirid = mappings.get_next_hir_id (crate_num);
+  // FIXME: Here, do we want the path's node ID or the lang-item's node id? Same
+  // for the DefId? This is where we want to add our lang item to the mappings
+  // too no? -> no, cause this is a usage of the lang item, but not the
+  // definition itself e.g. impl #[lang = copy] for i32 {}. so we don't want to
+  // add to mappings
```
CohenArthur, line 166: Remove this comment

gcc/rust/resolve/rust-ast-resolve-item.cc
```diff
@@ -678,16 +678,33 @@ ResolveItem::visit (AST::TraitImpl &impl_block)
       return;
     }
 
+  // FIXME: Cleanup!
```
CohenArthur, line 681: Remove comment

gcc/rust/resolve/rust-ast-resolve-item.cc
```diff
@@ -678,16 +678,33 @@ ResolveItem::visit (AST::TraitImpl &impl_block)
...
-      resolver->get_label_scope ().pop ();
-      return;
+      auto &lang_item
+	= static_cast<AST::LangItemPath &> (impl_block.get_trait_path ());
+
+      canonical_trait_type
+	= CanonicalPath::new_seg (lang_item.get_node_id (),
+				  LangItem::ToString (
+				    lang_item.get_lang_item_kind ()));
+    }
```
CohenArthur, line 696: In this block the `ok` bool should be `true`, either set it manually or initialize it as `true`

gcc/rust/resolve/rust-ast-resolve-type.h
```diff
@@ -56,6 +60,40 @@ class ResolveType : public ResolverBase
...
+  }
+
+  static NodeId go (AST::Path &type_path)
+  {
+    if (type_path.get_path_kind () == AST::Path::Kind::LangItem)
+      {
+	auto &type = static_cast<AST::LangItemPath &> (type_path);
+
+	rust_debug ("[ARTHUR]: lang item kind: %s",
+		    LangItem::ToString (type.get_lang_item_kind ()).c_str ());
```
CohenArthur, line 75: Remove

gcc/rust/resolve/rust-ast-resolve-type.h
```diff
@@ -56,6 +60,40 @@ class ResolveType : public ResolverBase
...
+      }
+
+    rust_assert (type_path.get_path_kind () == AST::Path::Kind::Type);
+
+    // We have to do this dance to first downcast to a typepath, and then upcast
+    // to a Type. The altnernative is to split `go` into `go` and `go_inner` or
+    // something, but eventually this will need to change as we'll need
+    // `ResolveType::` to resolve other kinds of `Path`s as well.
+    return ResolveType::go (
+      (AST::Type &) static_cast<AST::TypePath &> (type_path));
```
CohenArthur, line 94: It would be nice to find a cleaner way to do this because this is *disgusting*

gcc/rust/ast/rust-item.h
```diff
@@ -3197,10 +3197,26 @@ class TraitImpl : public Impl
...
+	     Visibility vis, std::vector<Attribute> inner_attrs,
+	     std::vector<Attribute> outer_attrs, location_t locus)
+    : Impl (std::move (generic_params), std::move (trait_type),
+	    std::move (where_clause), std::move (vis), std::move (inner_attrs),
+	    std::move (outer_attrs), locus),
+      has_unsafe (is_unsafe), has_exclam (has_exclam),
+      trait_path (std::unique_ptr<TypePath> (new TypePath (trait_path))),
+      impl_items (std::move (impl_items))
+  {}
+
```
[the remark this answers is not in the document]
CohenArthur: I'm not against that, I think that's a good idea. What did you have in mind exactly?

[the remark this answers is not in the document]
powerboat9: [This](https://github.com/Rust-GCC/gccrs/pull/3290/files/c6bed257a7536e4b2869623399894e22f3d5803d#diff-55bfd29990acc94d21252c82fb6f930018765bc0bd4893467d3f343ecd5428faR3200-R3213) constructor could become a delegating constructor, with [this](https://github.com/Rust-GCC/gccrs/pull/3290/files/c6bed257a7536e4b2869623399894e22f3d5803d#diff-55bfd29990acc94d21252c82fb6f930018765bc0bd4893467d3f343ecd5428faL3187-L3198) constructor as the target constructor.

https://en.cppreference.com/w/cpp/language/constructor#Delegating_constructor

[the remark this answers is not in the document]
CohenArthur: I think we can probably turn this into a good first PR as it's a nice cleanup but not functionally any different

PR #3292, opened 2024-12-04
Changes made to ensure that '=>' token is parsed properly

### LINKED ISSUE #3099: parse_expr not stopping on =>
### Summary

From Jun 5th,

[badumbatish](https://gcc-rust.zulipchat.com/#narrow/stream/327528-GSoC/topic/Inline.20Assembly.20Support/near/442814292):

> does parsing expression in parser.parse_expr() act the same as parsing outside? I called parse_expr (with no arguments) with current token of the first num1 and it kept on parsing until the right parenthesis of asm!
```rust
    unsafe {
        asm!(
            "add {0}, {0}",
            inout(reg) num1 => num1,
            in(reg) num2,
        );
    }
```

### Reproducer

I tried this code:

```rust
    unsafe {
        asm!(
            "add {0}, {0}",
            inout(reg) num1 => num1,
            in(reg) num2,
        );
    }
```

- [ ] Nightly

### Godbolt link

_No response_

### Actual behavior

Currently the inline assembly uses 2 identifier token checks, one for num1 before the `=>`, one for num1 after the `=>` 

### Expected behavior

I expected to be able to use parse_expr for both identifier, meaning I would want to call parse_expr(), then check for `=>` , then call parse_expr() again without parse_expr eating up all token from the token stream

### GCC Version

https://github.com/Rust-GCC/gccrs/pull/3098

sriganeshres:
gcc/rust/ChangeLog:

	* parse/rust-parse-impl.h (Parser::parse_expr): Updated parsing logic to stop on
	encountering '=>' or ',' while having restrictions.stop_on_token turned on.
	(Parser::left_denotations): Updated to stop parsing on the special tokens mentioned above.
	(Parser::null_denotation): Updated to stop parsing on the special tokens mentioned above.
	* parse/rust-parse.h (struct ParseRestrictions): Added a field stop_on_token (default false)
	to maintain the above functionality.

Fixes #3099 

- \[x] Run `clang-format`

* I have made changes to 3 functions `Parser::parse_expr`, `Parser::left_denotations`, `Parser::null_denotation` So that they stop on special token like `=>` and `,`.  They would terminate the parsing to allow exatract expression.
* Also I have added a test to verify the running of this.

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -12023,6 +12023,11 @@ Parser<ManagedTokenSource>::parse_expr (int right_binding_power,
   if (restrictions.expr_can_be_null)
     {
       TokenId id = current_token->get_id ();
+      if (restrictions.stop_on_token && (id == MATCH_ARROW || id == COMMA))
+	{
+	  rust_debug ("Stopping Parsing at special token");
+	  return nullptr;
+	}
       if (id == SEMICOLON || id == RIGHT_PAREN || id == RIGHT_CURLY
 	  || id == RIGHT_SQUARE || id == COMMA || id == LEFT_CURLY)
 	return nullptr;
@@ -12078,6 +12083,14 @@ Parser<ManagedTokenSource>::left_denotations (std::unique_ptr<AST::Expr> expr,
   while (right_binding_power < left_binding_power (current_token))
     {
       lexer.skip_token ();
+      if (restrictions.stop_on_token
+	  && (current_token->get_id () == MATCH_ARROW
+	      || current_token->get_id () == COMMA))
+	{
+	  rust_debug (
+	    "Stopping parsing at restricted token in left_denotations");
+	  return expr;
+	}
 
       // FIXME attributes should generally be applied to the null denotation.
       expr = left_denotation (current_token, std::move (expr),
@@ -12115,6 +12128,13 @@ Parser<ManagedTokenSource>::null_denotation (const_TokenPtr tok,
 {
   /* note: tok is previous character in input stream, not current one, as
    * parse_expr skips it before passing it in */
+
+  if (restrictions.stop_on_token
+      && (tok->get_id () == MATCH_ARROW || tok->get_id () == COMMA))
+    {
+      rust_debug ("Stopping parsing at restricted token in null_denotation");
+      return nullptr;
+    }
 
   /* as a Pratt parser (which works by decomposing expressions into a null
    * denotation and then a left denotation), null denotations handle primaries
--- a/gcc/rust/parse/rust-parse.h
+++ b/gcc/rust/parse/rust-parse.h
@@ -89,6 +89,8 @@ struct ParseRestrictions
   bool entered_from_unary = false;
   bool expr_can_be_null = false;
   bool expr_can_be_stmt = false;
+  bool stop_on_token
+    = false; // This is for stopping parsing at a specific token
   bool consume_semi = true;
   /* Macro invocations that are statements can expand without a semicolon after
    * the final statement, if it's an expression statement. */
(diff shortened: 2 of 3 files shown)


CohenArthur (commented):
Thanks @sriganeshres for contributing!! Could you add a testcase to your PR to ensure we don't introduce a regression on this in the future?

sriganeshres:
> Thanks @sriganeshres for contributing!! Could you add a testcase to your PR to ensure we don't introduce a regression on this in the future?

Working on that. I have ran `make check-rust` it gave correct results. Can you give me an idea of what type of test to be written for this. 

Thank You.

P-E-P:
> Working on that. I have ran `make check-rust` it gave correct results. Can you give me an idea of what type of test to be written for this.

You could write a test from the snippet given in the fixed issue, it would pass with your changes

gcc/rust/parse/rust-parse.h
```diff
@@ -89,6 +89,8 @@ struct ParseRestrictions
   bool entered_from_unary = false;
   bool expr_can_be_null = false;
   bool expr_can_be_stmt = false;
+  bool stop_on_token
+    = false; // This is for stopping parsing at a specific token
```
philberty, line 93: What does this mean? It seems like a hack

philberty (changes requested):
This doesnt seem correct to me the stop_on_token doesn't seem to be set to true anywhere so to me this doesn't seem to solve anything

gcc/testsuite/rust/compile/issue-3099.rs
```diff
@@ -0,0 +1,16 @@
...
+
+pub fn main() {
+    unsafe {
+        let value = 10;
+        let result = match value {
+            10 => 15,
+            _ => 20,
+        };
+
+        let format = "Result: %d\n\0" as *const str as *const i8;
```
badumbatish, line 13: hi @sriganeshres, make sure that you also update your dejagnu regex to assert that you'll get out 15 from println!. 

Without asserting the printf, you're just guarantee that we at least parse it successfully

gcc/testsuite/rust/compile/issue-3099.rs
```diff
@@ -0,0 +1,16 @@
+extern "C" {
+    fn printf(s: *const i8, ...);
+}
+
+pub fn main() {
+    unsafe {
```
badumbatish, line 6: is the unsafe necessary?

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -12023,6 +12023,11 @@ Parser<ManagedTokenSource>::parse_expr (int right_binding_power,
   if (restrictions.expr_can_be_null)
     {
       TokenId id = current_token->get_id ();
+      if (restrictions.stop_on_token && (id == MATCH_ARROW || id == COMMA))
```
badumbatish, line 12026: you're not setting stop_on_token to true so you're not entering these conditional either.

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -12078,6 +12083,14 @@ Parser<ManagedTokenSource>::left_denotations (std::unique_ptr<AST::Expr> expr,
   while (right_binding_power < left_binding_power (current_token))
     {
       lexer.skip_token ();
+      if (restrictions.stop_on_token
+	  && (current_token->get_id () == MATCH_ARROW
+	      || current_token->get_id () == COMMA))
```
badumbatish, line 12088: this seems weird. you'd want to enter left denotation and create the infix with =>

badumbatish (changes requested):

gcc/testsuite/rust/compile/issue-3099.rs
CohenArthur: it's necessary for the `printf` call, but not before - so it could be moved down a few lines indeed

Issue #3291, opened 2024-12-04
Refactor helper functions around attributes


CohenArthur:
> We should introduce some helper function here instead, I'm sure it would be reused elsewhere and it would ease this function a bit.

> Probably within `rust-attributes.cc` since we already have some similar functions there ?

_Originally posted by @P-E-P in https://github.com/Rust-GCC/gccrs/pull/3289#discussion_r1869691014_

There are plenty of places where attributes are being checked or used and a lot of repetition is happening. We have just added an `Attributes` class to `rust-attributes.cc`, and it would be nice to collect functions related to attributes in that class.

Harishankar14:
" I completely agree that introducing a helper function within the Attributes class in rust-attributes.cc would significantly reduce repetition and make the code cleaner and more maintainable. I’d like to take up this task and refactor the existing logic for attribute checks and retrieval into reusable methods.

My plan would be to:

Identify all instances where attributes are being checked or manipulated across the codebase.
Abstract common patterns into a helper function, such as has_attribute or get_attribute_value, to encapsulate these operations.
Integrate the helper function into the Attributes class while ensuring compatibility with existing functionality.
Add thorough documentation and tests to validate the changes and cover edge cases.
Let me know if there are any specific use cases or design constraints I should consider while working on this!"

itoshi69sae:
Hi, I am interested in working on this issue, but I am new here and need some guidance. Could you please point me to where the implementation for the attributes can be found in the codebase? Any relevant files, functions, or documentation would be really helpful.

Thank you for your time and support, and apologies if this is a basic question!

itoshi69sae:
@CohenArthur

PR #3293, opened 2024-12-05
Refactor helper functions around attributes #3291

Harishankar14:
In the original codebase, attribute checks were scattered across different functions, leading to a lot of repetition and scattered logic. The goal was to centralize the functionality related to attribute checking and retrieval into a single, reusable Attributes class. This would help improve maintainability, reduce redundancy, and improve the clarity of the codebase.

Changes Made:
Created the Attributes Class:

Introduced a new static utility class Attributes to centralize common operations on attributes such as checking for the existence of an attribute and retrieving its value.
The class includes the following functions:
has_attribute: Checks if an attribute with a specified name exists in a list of attributes.
get_attribute_value: Retrieves the value of an attribute by its name, returning an empty string if the attribute is not found.
is_proc_macro_type: Determines if an attribute corresponds to a "proc-macro" type, which is used to check attributes that are related to procedural macros.
get_builtin_type: A new helper function that maps attribute names to an enumeration (BuiltinAttributeType). This centralizes the mapping logic, making it easier to extend in the future.
Refactored the AttributeChecker Class:

Updated the check_attribute method to use the new Attributes class instead of direct attribute checks.
Instead of using multiple checks for each attribute, the check_attribute function now uses the Attributes::get_builtin_type to map attribute names to their corresponding type. This enables more modular and maintainable handling of attribute types.
Used the centralized attribute checking functions (has_attribute, get_attribute_value, is_proc_macro_type) to streamline the code and remove redundancy in the attribute checks.
Simplified Attribute Checks in Various Functions:

Replaced repetitive checks for specific attribute types (like PROC_MACRO, DOC, etc.) with calls to Attributes::is_proc_macro_type and Attributes::get_builtin_type. This reduces code duplication and improves the clarity of the logic.
Simplified the check_proc_macro_non_function and check_proc_macro_non_root functions by using Attributes::is_proc_macro_type to handle the checking of proc-macro attributes centrally.
Improved Error Handling for Invalid Attributes:

In functions like check_proc_macro_non_function and check_proc_macro_non_root, used the new Attributes::is_proc_macro_type function to check for proc-macro types and handle errors, making the code cleaner and more maintainable.
The error handling now leverages the rust_error_at function to print clear error messages when attributes are used incorrectly (e.g., proc-macro attributes used on non-function items or outside the crate root).
Enum for Builtin Attribute Types:

Introduced the BuiltinAttributeType enum to represent known attribute types (e.g., DOC, PROC_MACRO). This provides a more structured way to manage attribute types and makes it easier to add new types in the future.
The mapping of attribute names to types is now centralized in the get_builtin_type function, avoiding repetitive string-based checks

### DIFF
--- a/gcc/rust/util/rust-attributes.cc
+++ b/gcc/rust/util/rust-attributes.cc
@@ -67,14 +67,18 @@ static const BuiltinAttrDefinition __definitions[]
      {Attrs::RUSTC_CONST_UNSTABLE, STATIC_ANALYSIS},
      {Attrs::PRELUDE_IMPORT, NAME_RESOLUTION}};
 
-BuiltinAttributeMappings *
-BuiltinAttributeMappings::get ()
-{
-  static BuiltinAttributeMappings *instance = nullptr;
-  if (instance == nullptr)
-    instance = new BuiltinAttributeMappings ();
+BuiltinAttributeMappings* BuiltinAttributeMappings::get() {
+    static std::mutex mtx;
+    static BuiltinAttributeMappings* instance = nullptr;
 
-  return instance;
+    if (instance == nullptr) {
+        std::lock_guard<std::mutex> lock(mtx);
+        if (instance == nullptr) {
+            instance = new BuiltinAttributeMappings();
+        }
+    }
+
+    return instance;
 }
 
 const BuiltinAttrDefinition &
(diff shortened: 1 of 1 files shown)


gcc/rust/util/rust-attributes.cc
```diff
@@ -105,39 +109,77 @@ BuiltinAttributeMappings::BuiltinAttributeMappings ()
       mappings.insert ({def.name, def});
     }
 }
+class Attributes {
```
CohenArthur, line 112: There is already an `Attributes` class in `rust-attributes.h`, which you should add functions to.

CohenArthur (changes requested):

CohenArthur:
@Harishankar14 your PR message mentions the creation of the `Attributes` class, which already exists in the codebase. did you use an LLM to write the code as well? It does not feel like you really understand what you modified, or why you modified it. A lot of the code you changed didn't need to be changed, and you missed multiple uses of the "old" attribute system which would have needed to be refactored as well. You only modified one file, which indicates you didn't spend a lot of time searching in the code base for what to replace with your new attribute system.

it's no problem to ask an LLM for help with understanding a codebase, but using one to generate code for a project you don't understand is not great. I would much rather spend time chatting with you on the issue and replying to your questions, and helping you get started on the project. if you need I'm happy to help by chatting on our Zulip or here on github.

PR #3294, opened 2024-12-05
build: update bootstrap req to C++14

thesamesam:
We moved to a bootstrap requirement of C++11 in GCC 11, 8 years after support was stable in GCC 4.8.

It is now 8 years since C++14 was the default mode in GCC 6 (and 9 years since support was complete in GCC 5), and we have a few bits of optional C++14 code in the compiler, so it seems a good time to update the bootstrap requirement again.

The big benefit of the change is the greater constexpr power, but C++14 also added variable templates, generic lambdas, lambda init-capture, binary literals, and numeric literal digit separators.

C++14 was feature-complete in GCC 5, and became the default in GCC 6.  5.4.0 bootstraps trunk correctly; trunk stage1 built with 5.3.0 breaks in eh_data_format_name due to PR69995.

gcc/ChangeLog:

	* doc/install.texi (Prerequisites): Update to C++14.

ChangeLog:

	* configure.ac: Update requirement to C++14.
	* configure: Regenerate.

----

This cherry-picks r15-4719-ga9ec1bc06bd3cc so the GCC 5 bot is happy when we actually use C++14 features (see https://github.com/Rust-GCC/gccrs/pull/3278).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1456,16 +1456,16 @@ case "$have_compiler:$host:$target:$enable_bootstrap" in
     ;;
 esac
 
-# When bootstrapping with GCC, build stage 1 in C++11 mode to ensure that a
-# C++11 compiler can still start the bootstrap.  Otherwise, if building GCC,
-# require C++11 (or higher).
+# When bootstrapping with GCC, build stage 1 in C++14 mode to ensure that a
+# C++14 compiler can still start the bootstrap.  Otherwise, if building GCC,
+# require C++14 (or higher).
 if test "$enable_bootstrap:$GXX" = "yes:yes"; then
-  CXX="$CXX -std=c++11"
+  CXX="$CXX -std=c++14"
 elif test "$have_compiler" = yes; then
-  AX_CXX_COMPILE_STDCXX(11)
+  AX_CXX_COMPILE_STDCXX(14)
 
   if test "${build}" != "${host}"; then
-    AX_CXX_COMPILE_STDCXX(11, [], [], [_FOR_BUILD])
+    AX_CXX_COMPILE_STDCXX(14, [], [], [_FOR_BUILD])
   fi
 fi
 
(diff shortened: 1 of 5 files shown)


thesamesam:
The warnings changes are harmless (most of them are texi offsets).

tschwinge:
Thanks.  Per "Some checks were not successful", "Details", please update the `.github/*log*_warnings` files, and then this is good to go.

thesamesam:
Should I update the ChangeLog in the cherry-pick, or do a followup for updating the expected warnings?

tschwinge:
> Should I update the ChangeLog in the cherry-pick, or do a followup for updating the expected warnings?

You mean "[...] update the [`.github/*log*_warnings` files] in [...]"?  Do as you like, no strong preference.

(Over here, we don't maintain the `ChangeLog` files.)

tschwinge:
Actually, it makes more sense here to put the `.github/*log*_warnings` files updates into a separate commit, given these are authored by you vs. Jason for the actual changes.

configure.ac
```diff
@@ -1456,16 +1456,16 @@ case "$have_compiler:$host:$target:$enable_bootstrap" in
...
 if test "$enable_bootstrap:$GXX" = "yes:yes"; then
-  CXX="$CXX -std=c++11"
+  CXX="$CXX -std=c++14"
 elif test "$have_compiler" = yes; then
-  AX_CXX_COMPILE_STDCXX(11)
+  AX_CXX_COMPILE_STDCXX(14)
 
   if test "${build}" != "${host}"; then
-    AX_CXX_COMPILE_STDCXX(11, [], [], [_FOR_BUILD])
+    AX_CXX_COMPILE_STDCXX(14, [], [], [_FOR_BUILD])
```
CohenArthur, line 1468: so I'm not sure I understand this bit - is that change upstream already? is it gccrs specific?

tschwinge:
@thesamesam, via #3295 "ci: Update expected warnings list for 32 bits build", `.github/alpine_32bit_log_warnings` has been updated, so you'll want to rebase on top of that.

thesamesam:
Thanks all, and sorry for forgetting this originally.

PR #3299, opened 2024-12-10
Fix nr2 unit struct

P-E-P:


powerboat9:
Label scope is for ex: `break 'a` right?

powerboat9:
Most of the edits here look like they only apply to nr1

powerboat9:
A lot of these commits seem like they're ready to be merged, if put in separate pull request(s)

P-E-P:
@philberty could you please take a look at b646b259e9d8b31e7ba9c9b1758369536d1b0273 ? There is a mention of `ResolvePathType::Compile` earlier in a comment but that type must have disappeared since I can't find it. Not 100% convinced this is the right way of fixing unit struct compile issue.

P-E-P:
If accepted as is I could technically remove [ac30ed1](https://github.com/Rust-GCC/gccrs/pull/3299/commits/ac30ed113e530078e848f68f16ee5abd67549fb4).

CohenArthur (approved):
LGTM but waiting on feedback from @philberty :)

PR #3300, opened 2024-12-11
gccrs: feat: Made changes to ensure no wrong assignments are done.

### LINKED ISSUE #3287: gccrs does not check if something is assignable
### Summary

gccrs does not check if something is assignable, and gccrs accepts many un-assignable expressions.

### Reproducer

I tried this code:

```rust
fn main() -> i32 {
    let mut x = 42;
    x + 1 = 2;

    x
}
```

- [ ] Nightly

### Godbolt link

_No response_

### Actual behavior

The current behavior is that gccrs accepts this kind of semantically invalid code and even successfully generates machine code for them (in this case, gccrs generates code to make the function return `42`).

### Expected behavior

I expected to see an error thrown like `rustc` does:

```
error[E0070]: invalid left-hand side of assignment
 --> test1.rs:3:11
  |
3 |     x + 1 = 2;
  |     ----- ^
  |     |
  |     cannot assign to this expression
```

### GCC Version

fa93e280ed1c11d9775cb9c3211e13853d39c51c

sriganeshres:
gcc/rust/ChangeLog:

	* backend/rust-compile-expr.cc (lvalue_p): Created a function that checks the lvalue. (CompileExpr::visit): Modified the function to check the lvalue using the above mentioned function and also checks the type of lvalue expression.

gcc/testsuite/ChangeLog:

	* rust/compile/issue-3287.rs: New test for testing wrong assignments.

Fixes #3287

- \[x] Run `clang-format`


gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -958,12 +958,71 @@ CompileExpr::visit (HIR::LiteralExpr &expr)
     }
 }
 
+bool
+lvalue_p (const_tree ref)
```
philberty, line 962: this should be renamed and moved into rust-compile-base.cc or into rust-tree.cc call it is_lvalue_tree

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -958,12 +958,71 @@ CompileExpr::visit (HIR::LiteralExpr &expr)
...
+  const enum tree_code code = TREE_CODE (ref);
+
+  switch (code)
+    {
+    case REALPART_EXPR:
+    case IMAGPART_EXPR:
+    case COMPONENT_REF:
+      return lvalue_p (TREE_OPERAND (ref, 0));
+
+    case C_MAYBE_CONST_EXPR:
```
philberty, line 973: c_ TREE's dont exists here remove this case

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -958,12 +958,71 @@ CompileExpr::visit (HIR::LiteralExpr &expr)
...
+      /* MEM_REFs can appear from -fgimple parsing or folding, so allow them
+	 here as well.  */
+    case INDIRECT_REF:
+    case ARRAY_REF:
+    case VAR_DECL:
+    case PARM_DECL:
+    case RESULT_DECL:
+    case ERROR_MARK:
+      return (TREE_CODE (TREE_TYPE (ref)) != FUNCTION_TYPE
+	      && TREE_CODE (TREE_TYPE (ref)) != METHOD_TYPE);
```
philberty, line 993: iirc METHOD_TYPE might  not be a common tree type either

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -958,12 +958,71 @@ CompileExpr::visit (HIR::LiteralExpr &expr)
...
+    case PLUS_EXPR:
+    case MINUS_EXPR:
+    case MULT_EXPR:
+    case POINTER_PLUS_EXPR:
+    case POINTER_DIFF_EXPR:
+    case MULT_HIGHPART_EXPR:
+    case TRUNC_DIV_EXPR:
+      return false;
+    default:
+      rust_debug ("unknown");
```
philberty, line 1006: remove this useless debug

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -958,12 +958,71 @@ CompileExpr::visit (HIR::LiteralExpr &expr)
...
+    }
+}
+
 void
 CompileExpr::visit (HIR::AssignmentExpr &expr)
 {
   auto lvalue = CompileExpr::Compile (expr.get_lhs (), ctx);
   auto rvalue = CompileExpr::Compile (expr.get_rhs (), ctx);
 
+  if (!lvalue_p (lvalue)
```
philberty, line 1017: I dont like if statements like this its hard to read:

it should be:

bool valid_lvalue = XXX
if (!valid_lvalue) {
 -- error diagnostic
}

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -958,12 +958,71 @@ CompileExpr::visit (HIR::LiteralExpr &expr)
...
+      || expr.get_lhs ().get_expression_type ()
+	   == HIR::Expr::ExprType::Operator)
+    {
+      rust_error_at (expr.get_lhs ().get_locus (),
```
philberty, line 1021: you need to add in the rust error code ussing the proper interface

philberty (changes requested):
There are changes required here i think we can get away wit this check but not totally convinced yet

Also it will need to be applied to the other assignments such as:

x +=1 etc

gcc/rust/backend/rust-compile-expr.cc
sriganeshres: Should i remove that

gcc/rust/backend/rust-compile-expr.cc
sriganeshres: Can you give me more clarity

philberty: you need to pass the rust error code into this function like we do in other places take a look through the other parts of the code to check

sriganeshres:
> There are changes required here i think we can get away wit this check but not totally convinced yet
> 
> Also it will need to be applied to the other assignments such as:
> 
> x +=1 etc

Can you give me more information about the case we are missing as x+= 1 is valid as long as x is mutable in rust.

powerboat9:
> > There are changes required here i think we can get away wit this check but not totally convinced yet
> > Also it will need to be applied to the other assignments such as:
> > x +=1 etc
> 
> Can you give me more information about the case we are missing as x+= 1 is valid as long as x is mutable in rust.

You have to adjust the visitor for `CompoundAssignmentExpr` as well as the visitor for `AssignmentExpr`

powerboat9:
Would probably be good to make `lvalue_p` operate on HIR nodes instead of trees

gcc/rust/backend/rust-compile-base.h
```diff
@@ -30,6 +30,7 @@ class HIRCompileBase
   virtual ~HIRCompileBase () {}
 
   static tree address_expression (tree expr, location_t locus);
+  bool lvalue_p (const_tree ref);
```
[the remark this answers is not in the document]
sriganeshres: Done

gcc/testsuite/rust/compile/issue-3297.rs
```diff
@@ -0,0 +1,12 @@
...
+}
+
+pub fn main() {
+    unsafe {
+        let mut x = 42;
+        x + 1= 2; // { dg-error "invalid left-hand side of assignment" }
+        let format = "Result: %d\n\0" as *const str as *const i8;
+        printf(format, x);
+    }
+}
```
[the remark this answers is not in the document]
sriganeshres: Done

gcc/testsuite/rust/compile/issue-3297-2.rs
```diff
@@ -0,0 +1,4 @@
+pub fn main() {
+    let mut x = 42;
+    x += 1;
```
powerboat9, line 3: You'd have to change
```rust
x += 1;
```
to
```rust
x + 1 += 1; // { dg-error "invalid left-hand side of assignment" }
```

sriganeshres: I thought the test was to check no regression is introduced. Any ways i will make another test case with the error.

philberty: yeah @powerboat9 is right here the whole point of this PR is to handle bad assignments so this is another case you need to handle. So you need to add your code to the compound assignment expression too.

philberty (approved):
LGTM but i think you need to change that test case

gcc/rust/backend/rust-compile-base.cc
```diff
@@ -1015,5 +1015,52 @@ HIRCompileBase::unit_expression (location_t locus)
...
+    case REALPART_EXPR:
+    case IMAGPART_EXPR:
+    case COMPONENT_REF:
+      return is_lvalue (TREE_OPERAND (ref, 0));
+
+    case COMPOUND_LITERAL_EXPR:
+    case STRING_CST:
+    case CONST_DECL:
+    case INTEGER_CST:
+      return true;
```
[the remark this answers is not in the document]
philberty: TREE_CODE (a) should be a VAR_DECL iirc

philberty (changes requested):
Yeah this is good but yeah arthur has some minor comments too which need done.

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -964,6 +975,18 @@ CompileExpr::visit (HIR::AssignmentExpr &expr)
   auto lvalue = CompileExpr::Compile (expr.get_lhs (), ctx);
   auto rvalue = CompileExpr::Compile (expr.get_rhs (), ctx);
 
+  bool validl_value = is_lvalue (lvalue);
+
+  rust_debug ("%d", expr.get_lhs ().get_expression_type ());
+  if (!validl_value
+      || expr.get_lhs ().get_expression_type ()
+	   == HIR::Expr::ExprType::Operator)
+    {
```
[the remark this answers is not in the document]
sriganeshres: Like some exceptions we there so to get those. Like validl_value catches half and the other catches the other half.

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -964,6 +975,18 @@ CompileExpr::visit (HIR::AssignmentExpr &expr)
   auto lvalue = CompileExpr::Compile (expr.get_lhs (), ctx);
   auto rvalue = CompileExpr::Compile (expr.get_rhs (), ctx);
 
+  bool validl_value = is_lvalue (lvalue);
+
+  rust_debug ("%d", expr.get_lhs ().get_expression_type ());
```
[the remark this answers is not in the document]
sriganeshres: We can remove that

gcc/rust/backend/rust-compile-base.cc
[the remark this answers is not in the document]
sriganeshres: There is already a test case with wrong_lhs_assignment.rs

sriganeshres:
I guess they failing now because of newer tests. If this this merged with upstream there will not be as issue-3147.rs file is added after this changes.

philberty:
You need to rebase you code and investigate why you have failing tests.

gcc/rust/backend/rust-compile-expr.cc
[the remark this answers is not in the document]
powerboat9: Since assignment and compound assignment operators should share a definition of [assignee expressions](https://doc.rust-lang.org/reference/expressions.html#place-expressions-and-value-expressions), the two conditions shouldn't differ. If both conditions *need* to be different to pass the testsuite, that suggests both conditions are subtly wrong in ways our testsuite isn't catching.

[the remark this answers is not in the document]
sriganeshres: Well i have checked some cases with rustc and added those. The thing is whenever references like *a are being used they are just targeted by compoundassigmentexpr so the check to pass them is seen there. That is the reason for different conditions.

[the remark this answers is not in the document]
powerboat9: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=048300fdee2d01c8d5267b30ca45ecc5

PR #3301, opened 2024-12-12
gccrs: fix bad not expression in rust

philberty:
Fixes Rust-GCC#3229

gcc/rust/ChangeLog:

	* rust-gcc.cc (operator_to_tree_code): ! expressions are BIT_NOT_EXPR


powerboat9:
I think it might actually be different depending on type (`bool` vs number)

https://doc.rust-lang.org/reference/expressions/operator-expr.html#negation-operators

Issue #3230, opened 2024-10-28
 Missing `else` Clause in `if` expr That Returns Should Produce Type Mismatch Error


nobel-sh:
If the `if` expr returns a type `T` then else is expected to produce the same type `T` but in absence of `else` branch it return a `()` type. However, gccrs does not enforce this check.
 
```rust
fn foo(pred: bool) -> u8 {
    if pred {
        1
    }
    3
}
fn main(){
}
```

rustc:
```bash
error[E0308]: mismatched types
 --> <source>:3:9
  |
2 | /     if pred {
3 | |         1
  | |         ^ expected `()`, found integer
4 | |     }
  | |_____- expected this to be `()`
  |
help: you might have meant to return this value
  |
3 |         return 1;
  |         ++++++  +

error: aborting due to 1 previous error
```

gccrs: Compiles sucessfully

godbolt: https://godbolt.org/z/xnczfjznT

PR #3306, opened 2024-12-16
Add type check on if-expr

Kamiinarii78:
Check if an if-expr return void type.

gcc/rust/ChangeLog:

	* typecheck/rust-hir-type-check-expr.cc (TypeCheckExpr::visit): Add check on if-expr.

gcc/testsuite/ChangeLog:

	* rust/compile/implicit_returns_err3.rs: Change test to be valid.
	* rust/compile/torture/if.rs: Likewise.
	* rust/compile/if-without-else.rs: New test.

Fixes: #3230

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-expr.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-expr.cc
@@ -526,9 +526,18 @@ TypeCheckExpr::visit (HIR::IfExpr &expr)
 				    expr.get_if_condition ().get_locus ()),
 	      expr.get_locus ());
 
-  TypeCheckExpr::Resolve (expr.get_if_block ());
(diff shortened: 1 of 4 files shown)


CohenArthur (approved):
I think this is good, thank you!! Just waiting on Philip to check it as he knows the typechecker :D

philberty (approved):
LGTM

philberty (changes requested):
This is now rebased and looks like there are failing tests can you investigate?

philberty (commented):
I think this LGTM but i would prefer if we use the result of the coercion_site if that works if not then let us know and we can go with this.

Also we really need to make an issue about making much nicer helpers for looking up those builtin types

Issue #3210, opened 2024-10-21
Tracking issue for experimental features in 1.49 implemented in gccrs


MajoraSans:
- [ ] abi_avr_interrupt (https://github.com/rust-lang/rust/issues/69664)
- [ ] abi_msp430_interrupt (https://github.com/rust-lang/rust/issues/38487)
- [ ] abi_ptx (https://github.com/rust-lang/rust/issues/38788)
- [ ] abi_x86_interrupt (https://github.com/rust-lang/rust/issues/40180)
- [ ] alloc_error_handler (https://github.com/rust-lang/rust/issues/51540)
- [ ] arbitrary_self_types (https://github.com/rust-lang/rust/issues/44874)
- [ ] associated_type_defaults (https://github.com/rust-lang/rust/issues/29661)
- [ ] async_closure (https://github.com/rust-lang/rust/issues/62290)
- [ ] auto_traits (https://github.com/rust-lang/rust/issues/13231)
- [ ] box_patterns (https://github.com/rust-lang/rust/issues/29641)
- [ ] c_variadic (https://github.com/rust-lang/rust/issues/44930)
- [ ] cfg_sanitize (https://github.com/rust-lang/rust/issues/39699)
- [ ] cfg_target_thread_local (https://github.com/rust-lang/rust/issues/29594)
- [ ] cfg_version (https://github.com/rust-lang/rust/issues/64796)
- [ ] cmse_nonsecure_entry (https://github.com/rust-lang/rust/issues/75835)
- [ ] const_extern_fn (https://github.com/rust-lang/rust/issues/64926)
- [ ] const_muts_ref (https://github.com/rust-lang/rust/issues/57349)
- [ ] const_precise_live_drops (https://github.com/rust-lang/rust/issues/73255)
- [ ] const_refs_to_cell (https://github.com/rust-lang/rust/issues/80384)
- [ ] const_trait_impl (https://github.com/rust-lang/rust/issues/67792)
- [ ] const_try (https://github.com/rust-lang/rust/issues/74935)
- [ ] coroutines (https://github.com/rust-lang/rust/issues/43122)
- [ ] custom_inner_attributes (https://github.com/rust-lang/rust/issues/54726)
- [ ] custom_test_frameworks (https://github.com/rust-lang/rust/issues/50297)
- [ ] decl_macro (https://github.com/rust-lang/rust/issues/39412)
- [ ] do_not_recommend (https://github.com/rust-lang/rust/issues/51992)
[...]

PR #3309, opened 2024-12-18
Required features for Rust For Linux

### LINKED ISSUE #3233: Implementation of the recognition of the required features for Rust for Linux
In the #3210 are listed the required features of the 1.49.0 version of Rust for Rust for Linux,
and this issue is here to give some code to the recognition of these features in the feature-gate,
and the implementation of the errors if these features are used without the attribute.
I will also provide some tests for all of these.

MajoraSans:
Fixes #3233.

This PR is here to show the work done on the recognition of the required features for Rust for Linux mentioned in the #3210, and aims to implement all of them.


MajoraSans:
As for now, all features marked as "Required" in #3210 are recognized, except `doc_cfg`, where I did not understand what was required or checked, and what the feature was actually providing. Furthermore, the `Nice to have` features and more precisely the `associated_type_defaults` one required me to change how the syntax was previously made to allow type aliases in trait to have a definition ("=" after the declaration) which currently cause a syntax error, so I preferred making this PR before continuing.

gcc/rust/checks/errors/rust-feature-gate.cc
```diff
@@ -17,11 +17,19 @@
...
+#include "rust-ast.h"
 #include "rust-attribute-values.h"
 #include "rust-ast-visitor.h"
 #include "rust-feature.h"
 #include "rust-ast-full.h"
+#include "rust-item.h"
+#include "rust-path.h"
+#include "rust-type.h"
+#include "rust-tyty.h"
+#include <iostream>
```
philberty, line 32: this should use rust-system.h i dont see why you need iostream and rust-tyty.h for the type abstractions

gcc/rust/ast/rust-ast-visitor.h
```diff
@@ -435,7 +435,7 @@ class DefaultASTVisitor : public ASTVisitor
   virtual void visit (AST::StructPatternElements &spe);
   virtual void visit (AST::MaybeNamedParam &param);
 
-  void visit (AST::Attribute &attribute) {}
+  virtual void visit (AST::Attribute &attribute) {}
```
philberty, line 438: does this need virtual i dont see where your overriding it

philberty: oh wait scratch that i see now sorry

philberty:
I think i would like to get this merged its a good idea. Ideally these new tests that fail should be like XFAIL so that we keep the build happy so when we fix things then they go XPASS i guess? @dkm  any suggestions?

dkm:
> I think i would like to get this merged its a good idea. Ideally these new tests that fail should be like XFAIL so that we keep the build happy so when we fix things then they go XPASS i guess? @dkm any suggestions?

Oh, sorry, missed that ping :bow: Sounds like a good plan yes!

gcc/rust/ast/rust-ast-visitor.cc
```diff
@@ -831,8 +831,6 @@ DefaultASTVisitor::visit (AST::Function &function)
   visit (function.get_qualifiers ());
   for (auto &generic : function.get_generic_params ())
     visit (generic);
-  if (function.has_self_param ())
-    visit (function.get_self_param ());
```
powerboat9, line 835: What breaks if this isn't removed?

Polygonalr:
The relevant feature gates are already implemented by @powerboat9 in https://github.com/Rust-GCC/gccrs/commit/23a20908e414a64d283d6b8e727c8f503e8c01b4, with the exception of `impl_trait_in_assoc_type` which is not actually used in RfL. This can be closed.

P-E-P:
> The relevant feature gates are already implemented by @powerboat9 in [23a2090](https://github.com/Rust-GCC/gccrs/commit/23a20908e414a64d283d6b8e727c8f503e8c01b4), with the exception of `impl_trait_in_assoc_type` which is not actually used in RfL. This can be closed.

Thank you.

PR #3319, opened 2024-12-21
Allow float type to be casted as integer type

### LINKED ISSUE #3261: Restricted type casting from integer types to float types and vice versa
`gccrs` does not allow conversion from integer type to float type and float type to integer type.

```rust
fn main(){
    let a: u8 = 13;
    let b:f64 = a as f64;
}
```
gccrs produces the following error:
```bash
<source>:3:17: error: casting 'u8' as 'f64' is invalid [E0606]
    3 |     let b:f64 = a as f64;
      |                 ^    ~~~
```

Godbolt link: https://godbolt.org/z/qorqTvfbx

nobel-sh:
#3270 added rules for `int -> float` casts but not the opposite .
This PR should add  support for `float -> int` casts and fully fix #3261.

### DIFF
--- a/gcc/rust/typecheck/rust-casts.cc
+++ b/gcc/rust/typecheck/rust-casts.cc
@@ -225,6 +225,12 @@ TypeCastRules::cast_rules ()
     case TyTy::TypeKind::FLOAT:
       switch (to.get_ty ()->get_kind ())
 	{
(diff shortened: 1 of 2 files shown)


CohenArthur (approved):
LGTM but waiting for Philip :)

philberty (approved):
LGTM

PR #3323, opened 2024-12-24
nr2.0: Handle `Self` properly in trait definitions

powerboat9:


gcc/rust/resolve/rust-late-name-resolver-2.0.cc
```diff
@@ -287,6 +287,16 @@ Late::visit (AST::TypePath &type)
   DefaultResolver::visit (type);
 }
 
+void
+Late::visit (AST::Trait &trait)
+{
+  // kind of weird how this is done
```
CohenArthur, line 293: very nitpicky but could you add a comment here explaining the weirdness? will be helpful for future contributors

CohenArthur (approved):
LGTM, well done!

PR #3330, opened 2024-12-28
nr2.0: Resolve type aliases inside trait definitions

powerboat9:
Requires #3323

### DIFF
--- a/gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
@@ -109,6 +109,15 @@ TopLevel::visit (AST::Trait &trait)
   DefaultResolver::visit (trait);
 }
 
+void
+TopLevel::visit (AST::TraitItemType &trait_item)
+{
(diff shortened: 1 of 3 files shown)


gcc/rust/resolve/rust-late-name-resolver-2.0.cc
```diff
@@ -300,6 +300,15 @@ Late::visit (AST::Trait &trait)
   DefaultResolver::visit (trait);
 }
 
+void
+Late::visit (AST::TraitItemType &trait_item)
+{
+  std::ignore = ctx.types.insert (trait_item.get_identifier (),
+				  trait_item.get_node_id ());
```
CohenArthur, line 307: any reason we're not asserting that we've inserted the trait item correctly?

CohenArthur (approved):
LGTM, just one question but happy to merge

gcc/rust/resolve/rust-late-name-resolver-2.0.cc
powerboat9: On second thought, the toplevel resolver really should be handling this

PR #3336, opened 2024-12-31
refactor all attribute functions in rust-compile-base.cc

saeitoshi-10:
gcc/rust/ChangeLog:
        * backend/rust-compile-base.cc (get_attributes):removed checker function (get_trait_name):removed checker function
	* util/rust-attributes.cc (Attributes::get_attributes):added checker function (Attributes::get_trait_name): added checker function
	* util/rust-attributes.h: added functions definitions in the header file for access

in the git commit message.

- \[ ] Run `clang-format`

order to get feedback.
---
Addresses : #3291 

### DIFF
--- a/gcc/rust/util/rust-attributes.h
+++ b/gcc/rust/util/rust-attributes.h
@@ -29,6 +29,8 @@ class Attributes
 {
 public:
   static bool is_known (const std::string &attribute_path);
+  static std::string get_trait_name (const AST::Attribute &attr);
+  static std::vector<std::string> get_attributes (const AST::Attribute &attr);
 };
 
 enum CompilerPass
--- a/gcc/rust/backend/rust-compile-base.cc
+++ b/gcc/rust/backend/rust-compile-base.cc
@@ -33,6 +33,7 @@
 #include "rust-compile-implitem.h"
 #include "rust-attribute-values.h"
 #include "rust-immutable-name-resolution-context.h"
+#include "rust-attributes.h"
 
 #include "fold-const.h"
 #include "stringpool.h"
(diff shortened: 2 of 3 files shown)


saeitoshi-10:
no i didn't see any other implementations till but if it exists i'll be sure to replace that

philberty (commented):
This seems fine but not 100% sure probably best to ask @P-E-P iirc did this code sometimes its useful to keep stuff thats associated local i am fine with it if others are

philberty:
@saeitoshi-10 merge conflicts this needs rebased

gcc/rust/util/rust-attributes.h
```diff
@@ -29,6 +29,8 @@ class Attributes
 {
 public:
   static bool is_known (const std::string &attribute_path);
+  static std::string get_trait_name (const AST::Attribute &attr);
```
P-E-P, line 32: I'm not keen to adding those functions in the attribute class as here attribute means two different things. The first one is the attribute as we all know it in rust `#[attribute]` (1), but the second one is specific to proc macros (2). On one hand moving it inside the attribute class seems appropriate because it is indeed part of proc macro declaration which are attributes(1) but on the other hand it might confuse people as it really only works for (2)

P-E-P (changes requested):
You need to amend your commit message. Right now you've kept the autogenerated commit message which is great. But a commit message should contain more than that.

GCC uses the usual convention that can be found in most projects, one short title (max 50 columns), it should contain keywords to easily identify the commit among others.

Then a blank line.

And then a description about why the commit has been introduced, why it exists and even why those changes are done this way. This section should be wrapped at 72 columns.

Then I usually leave another blank line and keep the generated changelog structure filled with what has been changed.

PR #3339, opened 2025-01-02
refactor all attribute functions in rust-privacy-reporter.cc

saeitoshi-10:
gcc/rust/ChangeLog:	
        * checks/errors/privacy/rust-privacy-reporter.cc (find_proc_macro_attribute):"Removed checker function" (proc_macro_privacy_check): "modified checker implementation"
	* util/rust-attributes.cc (is_proc_macro_type): "modifed fn into Attributes class" (Attributes::is_proc_macro_type): "modifed fn" (check_proc_macro_non_function): "modified checker implementation" (check_proc_macro_non_root): "modified checker implementation"
	* util/rust-attributes.h: "included definition"

in the git commit message.

- \[ ] Run `clang-format`

order to get feedback.
---
Addresses #3291 

### DIFF
--- a/gcc/rust/checks/errors/privacy/rust-privacy-reporter.cc
+++ b/gcc/rust/checks/errors/privacy/rust-privacy-reporter.cc
@@ -23,6 +23,7 @@
 #include "rust-hir-item.h"
 #include "rust-attribute-values.h"
 #include "rust-immutable-name-resolution-context.h"
+#include "rust-attributes.h"
 
 namespace Rust {
 namespace Privacy {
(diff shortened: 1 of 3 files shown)


philberty:
You need to fix your commit format before this can be merged

gcc/rust/checks/errors/privacy/rust-privacy-reporter.cc
```diff
@@ -34,33 +35,23 @@ PrivacyReporter::PrivacyReporter (
...
 {
   if (item->get_hir_kind () == HIR::Node::BaseKind::VIS_ITEM)
     {
-      auto attribute = find_proc_macro_attribute (item->get_outer_attrs ());
+      tl::optional<std::string> attribute = tl::nullopt;
+      // Find a proc_macro, proc_macro_derive or proc_macro_attribute
+      // attribute in a vector of attribute
+      for (const auto &attr : item->get_outer_attrs ())
+	{
+	  if (Analysis::Attributes::is_proc_macro_type (attr) != tl::nullopt)
```
philberty, line 49: this seems odd i think you should be storing the result here in a variable instead of calling it twice like this

gcc/rust/util/rust-attributes.cc
```diff
@@ -223,27 +223,30 @@ check_doc_attribute (const AST::Attribute &attribute)
...
-
 // Emit an error when one encountered attribute is either #[proc_macro],
 // #[proc_macro_attribute] or #[proc_macro_derive]
 static void
 check_proc_macro_non_function (const AST::AttrVec &attributes)
 {
   for (auto &attr : attributes)
     {
-      if (is_proc_macro_type (attr))
+      if (Attributes::is_proc_macro_type (attr) != tl::nullopt)
```
philberty, line 249: can this not just be if (Attributes::is_proc_macro_type (attr)) ?

gcc/rust/util/rust-attributes.cc
```diff
@@ -258,7 +261,7 @@ check_proc_macro_non_root (AST::AttrVec attributes, location_t loc)
...
```
philberty, line 264: same here

philberty (changes requested):
needs some changes

gcc/rust/util/rust-attributes.cc
```diff
@@ -223,27 +223,30 @@ check_doc_attribute (const AST::Attribute &attribute)
       }
     }
 }
-
-static bool
-is_proc_macro_type (const AST::Attribute &attribute)
+tl::optional<std::string>
```
P-E-P, line 226: Could we return an enum instead of a string here ? Enums are cheap, strings are not. and if I'm reading this correctly string values are known statically.

gcc/rust/util/rust-attributes.cc
```diff
@@ -223,27 +223,30 @@ check_doc_attribute (const AST::Attribute &attribute)
...
+Attributes::is_proc_macro_type (const AST::Attribute &attribute)
```
P-E-P, line 227: You're changing the semantic of the function, either you need a new one (two different functions) or you need to rename this one, we're not deciding if an attribute is or is not a proc macro anymore but returning the kind of proc macro with your new code.

I'd keep two different functions, even if one is calling the other and simply using `tl::optional::has_value()`

gcc/rust/util/rust-attributes.cc
P-E-P: You're right, it could be shortened.

PR #3343, opened 2025-01-03
Finish `derive(Clone)` for enums

### LINKED ISSUE #2996: Clone

CohenArthur:
This PR changes how lang item paths are handled and subsequently finishes the implementation of `derive(Clone)` for enums and their variants. Putting it as a draft since it's a mess and needs cleaning up. Addresses #2996, but we still need to properly implement generic Clone I think?

I might instead split this up in multiple PRs to first refactor the lang item paths and then finish implementing `derive(Clone)` on top of it - next week.

Needs #3366

### DIFF
--- a/gcc/rust/ast/rust-ast-builder.h
+++ b/gcc/rust/ast/rust-ast-builder.h
@@ -160,9 +160,14 @@ public:
 
   /**
    * Create an expression for struct instantiation with fields (`S { a, b: c }`)
+   * Tuple expressions are call expressions and can thus be constructed with
+   * `call`
    */
   std::unique_ptr<Expr>
   struct_expr (std::string struct_name,
+	       std::vector<std::unique_ptr<StructExprField>> &&fields) const;
+  std::unique_ptr<Expr>
+  struct_expr (PathInExpression struct_name,
 	       std::vector<std::unique_ptr<StructExprField>> &&fields) const;
 
   /* Create a field expression for struct instantiation (`field_name: value`) */
--- a/gcc/rust/ast/rust-ast-builder.h
+++ b/gcc/rust/ast/rust-ast-builder.h
@@ -114,16 +114,27 @@ public:
 
   /* And similarly for type path segments */
   std::unique_ptr<TypePathSegment> type_path_segment (std::string seg) const;
+  std::unique_ptr<TypePathSegment>
+  type_path_segment (LangItem::Kind lang_item) const;
 
   std::unique_ptr<TypePathSegment>
-  generic_type_path_segment (std::string seg, GenericArgs args) const;
+  type_path_segment_generic (std::string seg, GenericArgs args) const;
+  std::unique_ptr<TypePathSegment>
+  type_path_segment_generic (LangItem::Kind lang_item, GenericArgs args) const;
 
   /* Create a Type from a single string - the most basic kind of type in our AST
    */
   std::unique_ptr<Type> single_type_path (std::string type) const;
+  std::unique_ptr<Type> single_type_path (LangItem::Kind lang_item) const;
 
   std::unique_ptr<Type> single_generic_type_path (std::string type,
 						  GenericArgs args) const;
+  std::unique_ptr<Type> single_generic_type_path (LangItem::Kind lang_item,
+						  GenericArgs args) const;
+
+  TypePath type_path (std::unique_ptr<TypePathSegment> &&segment) const;
+  TypePath type_path (std::string type) const;
+  TypePath type_path (LangItem::Kind lang_item) const;
 
   /**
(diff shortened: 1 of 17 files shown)


gcc/rust/expand/rust-derive-clone.cc
```diff
@@ -28,6 +34,15 @@ DeriveClone::clone_call (std::unique_ptr<Expr> &&to_clone)
...
+  // Interestingly, later versions of Rust have a `clone_fn` lang item which
+  // corresponds to this. But because we are first targeting 1.49, we cannot use
+  // it yet. Once we target a new, more recent version of the language, we'll
+  // have figured out how to compile and distribute `core`, meaning we'll be
+  // able to directly call `::core::clone::Clone::clone()`
+
+  // Not sure how to call it properly in the meantime...
+
   auto path = std::unique_ptr<Expr> (
     new PathInExpression (builder.path_in_expression ({"Clone", "clone"})));
```
CohenArthur, line 47: In the meantime we could generate something like `<to_clone_type as #[lang = "clone"]>::clone(to_clone)` but this is extra work. It's not super difficult, so I'll mark it as a good-first-pr

gcc/rust/expand/rust-derive-clone.cc
```diff
@@ -227,36 +239,229 @@ DeriveClone::visit_struct (StructStruct &item)
...
   // FIXME: Should be $crate::core::clone::AssertParamIsCopy (or similar)
+  // (Rust-GCC#3329)
+
+  auto copy_path = TypePath (vec (builder.type_path_segment ("Copy")), loc);
+  auto sized_path = TypePath (vec (builder.type_path_segment ("Sized")), loc);
+
+  auto copy_bound = std::unique_ptr<TypeParamBound> (
+    new TraitBound (copy_path, item.get_locus ()));
+  auto sized_bound = std::unique_ptr<TypeParamBound> (
+    new TraitBound (sized_path, item.get_locus (), false, true));
```
CohenArthur, line 426: This needs to be a `?Sized` bound rather than a `Sized` bound

gcc/rust/expand/rust-derive-clone.cc
```diff
@@ -227,36 +239,229 @@ DeriveClone::visit_struct (StructStruct &item)
...
 }
 
 void
 DeriveClone::visit_union (Union &item)
 {
   // FIXME: Should be $crate::core::clone::AssertParamIsCopy (or similar)
+  // (Rust-GCC#3329)
+
+  auto copy_path = TypePath (vec (builder.type_path_segment ("Copy")), loc);
+  auto sized_path = TypePath (vec (builder.type_path_segment ("Sized")), loc);
```
CohenArthur, line 421: Both of these need to be lang items instead

gcc/rust/expand/rust-derive-clone.cc
```diff
@@ -227,36 +239,229 @@ DeriveClone::visit_struct (StructStruct &item)
...
+
+  auto copy_bound = std::unique_ptr<TypeParamBound> (
+    new TraitBound (copy_path, item.get_locus ()));
+  auto sized_bound = std::unique_ptr<TypeParamBound> (
+    new TraitBound (sized_path, item.get_locus (), false, true));
+
+  auto bounds = std::vector<std::unique_ptr<TypeParamBound>> ();
+  bounds.emplace_back (std::move (copy_bound));
+  bounds.emplace_back (std::move (sized_bound));
```
CohenArthur, line 430: We can use `vec(copy_bound, sized_bound)` here

gcc/rust/ast/rust-ast-builder.h
```diff
@@ -21,10 +21,35 @@
...
+std::vector<std::unique_ptr<T>>
+vec (std::unique_ptr<T> &&t1, std::unique_ptr<T> &&t2)
+{
+  auto v = std::vector<std::unique_ptr<T>> ();
+
+  v.emplace_back (std::move (t1));
+  v.emplace_back (std::move (t2));
+
+  return v;
+}
```
CohenArthur, line 51: this should probably be just one variadic function?

gcc/rust/expand/rust-derive-clone.cc
CohenArthur: nvm it's already the case

gcc/rust/expand/rust-derive-clone.cc
CohenArthur: Note that this is important for handling `core` and closing out the milestone though, so it does need to be done "shortly"

CohenArthur: https://github.com/Rust-GCC/gccrs/issues/3371

PR #3344, opened 2025-01-04
refactor some attribute functions in rust-hir-map.cc and rust-early-name-resolver.cc

saeitoshi-10:
gcc/rust/ChangeLog:
        * resolve/rust-early-name-resolver.cc (EarlyNameResolver::visit): "modified impl"
	* util/rust-attributes.cc (Attributes::is_rustc_builtin_macro): "helper function"
	* util/rust-attributes.h: "added"
	* util/rust-hir-map.cc (Mappings::insert_macro_def): "modified impl"

in the git commit message.

- \[ ] Run `clang-format`

order to get feedback.
---
Addresses #3291 

### DIFF
--- a/gcc/rust/util/rust-attributes.cc
+++ b/gcc/rust/util/rust-attributes.cc
@@ -37,6 +37,18 @@ Attributes::is_known (const std::string &attribute_path)
 
   return !lookup.is_error ();
 }
+bool
+Attributes::is_rustc_builtin_macro (AST::AttrVec outer_attrs)
+{
+  for (auto attr : outer_attrs)
+    {
+      if (attr.get_path () == Values::Attributes::RUSTC_BUILTIN_MACRO)
+	{
+	  return true;
+	}
+    }
+  return false;
+}
 
 using Attrs = Values::Attributes;
 
--- a/gcc/rust/resolve/rust-early-name-resolver.cc
+++ b/gcc/rust/resolve/rust-early-name-resolver.cc
@@ -21,6 +21,7 @@
 #include "rust-name-resolver.h"
 #include "rust-macro-builtins.h"
 #include "rust-attribute-values.h"
+#include "rust-attributes.h"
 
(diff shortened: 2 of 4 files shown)


gcc/rust/util/rust-attributes.cc
```diff
@@ -37,6 +37,18 @@ Attributes::is_known (const std::string &attribute_path)
 
   return !lookup.is_error ();
 }
+bool
+Attributes::is_rustc_builtin_macro (AST::AttrVec outer_attrs)
+{
+  for (auto attr : outer_attrs)
+    {
+      if (attr.get_path () == Values::Attributes::RUSTC_BUILTIN_MACRO)
+	{
```
philberty, line 46: You can just do return true here no need for the extra braces

gcc/rust/util/rust-attributes.cc
```diff
@@ -37,6 +37,18 @@ Attributes::is_known (const std::string &attribute_path)
 
   return !lookup.is_error ();
 }
+bool
+Attributes::is_rustc_builtin_macro (AST::AttrVec outer_attrs)
+{
+  for (auto attr : outer_attrs)
```
philberty, line 43: you should probably do const auto attr

gcc/rust/util/rust-attributes.cc
```diff
@@ -37,6 +37,18 @@ Attributes::is_known (const std::string &attribute_path)
 
   return !lookup.is_error ();
 }
+bool
+Attributes::is_rustc_builtin_macro (AST::AttrVec outer_attrs)
```
philberty, line 41: You should be using const AST::AttrVec& outer_attrs since this can be constified

philberty (changes requested):
some cleanup needed

P-E-P:
You need to rewrite those commit message, they lack a title and a commit description.

PR #3347, opened 2025-01-06
gccrs: add support for lang_item eq and PartialEq trait

philberty:
The Eq and Partial Ord are very similar to the operator overloads we support for add/sub/etc... but they differ in that usually the function call name matches the name of the lang item. This time we need to have support to send in a new path for the method call on the lang item we want instead of just the name of the lang item.

NOTE: this test case doesnt work correctly yet we need to support the derive of partial eq on enums to generate the correct comparison code for that.

Fixes Rust-GCC#3302

gcc/rust/ChangeLog:

	* backend/rust-compile-expr.cc (CompileExpr::visit): handle partial_eq possible call
	* backend/rust-compile-expr.h: handle case where lang item calls differ from name
	* hir/tree/rust-hir-expr.cc (OperatorExprMeta::OperatorExprMeta): new helper
	* hir/tree/rust-hir-expr.h: likewise
	* typecheck/rust-hir-type-check-expr.cc (TypeCheckExpr::visit): handle partial_eq
	(TypeCheckExpr::resolve_operator_overload): likewise
	* typecheck/rust-hir-type-check-expr.h: likewise
	* util/rust-lang-item.cc (LangItem::ComparisonToLangItem): map comparison to lang item
	(LangItem::ComparisonToSegment): likewise
	* util/rust-lang-item.h: new lang items PartialOrd and Eq
	* util/rust-operators.h (enum class): likewise

gcc/testsuite/ChangeLog:

	* rust/compile/cmp1.rs: New test.

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-expr.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-expr.cc
@@ -344,6 +344,21 @@ TypeCheckExpr::visit (HIR::ComparisonExpr &expr)
   auto lhs = TypeCheckExpr::Resolve (expr.get_lhs ());
   auto rhs = TypeCheckExpr::Resolve (expr.get_rhs ());
 
+  auto borrowed_rhs
+    = new TyTy::ReferenceType (mappings.get_next_hir_id (),
+			       TyTy::TyVar (rhs->get_ref ()), Mutability::Imm);
+  context->insert_implicit_type (borrowed_rhs->get_ref (), borrowed_rhs);
+
+  auto seg_name = LangItem::ComparisonToSegment (expr.get_expr_type ());
+  auto segment = HIR::PathIdentSegment (seg_name);
+  auto lang_item_type = LangItem::ComparisonToLangItem (expr.get_expr_type ());
+
+  bool operator_overloaded
+    = resolve_operator_overload (lang_item_type, expr, lhs, borrowed_rhs,
+				 segment);
+  if (operator_overloaded)
+    return;
+
   unify_site (expr.get_mappings ().get_hirid (),
 	      TyTy::TyWithLocation (lhs, expr.get_lhs ().get_locus ()),
 	      TyTy::TyWithLocation (rhs, expr.get_rhs ().get_locus ()),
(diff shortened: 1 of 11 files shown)


gcc/rust/util/rust-lang-item.cc
```diff
@@ -145,6 +148,48 @@ LangItem::OperatorToLangItem (ArithmeticOrLogicalOperator op)
...
+    case ComparisonOperator::LESS_THAN:
+      return "lt";
+    case ComparisonOperator::GREATER_OR_EQUAL:
+      return "ge";
+    case ComparisonOperator::LESS_OR_EQUAL:
+      return "le";
+    }
+
+  rust_unreachable ();
+  return std::string ();
```
CohenArthur, line 190:
```suggestion
```
not needed since we have the unreachable

CohenArthur (commented):
had a quick first look and it looks good, I'll review in more detail later

philberty:
For clarity this is whats getting generated at the end of the day:

```
  try          
    {
      _c1 = <test::Book as test::PartialEq::<test::BookFormat>>::eq (&b1, &D.501);  
      _c2 = <test::BookFormat as test::PartialEq::<test::Book>>::eq (&D.502, &b2); 
      _c3 = test::PartialEq::ne<Book, Book> (&b1, &b2); 
      return D.503;
    }
  finally                                               
    {                                                   
      b2 = {CLOBBER(eos)};       
      b1 = {CLOBBER(eos)};                                                                                                                                                                                                       
    }

```

gcc/rust/typecheck/rust-hir-type-check-expr.cc
```diff
@@ -344,6 +344,21 @@ TypeCheckExpr::visit (HIR::ComparisonExpr &expr)
   auto lhs = TypeCheckExpr::Resolve (expr.get_lhs ());
   auto rhs = TypeCheckExpr::Resolve (expr.get_rhs ());
 
+  auto borrwed_rhs
```
CohenArthur, line 347:
```suggestion
  auto borrowed_rhs
```

nit, typo

CohenArthur (approved):
LGTM! thanks Philip!

CohenArthur:
@philberty what part should be `derive(PartialEq)` in the testcase?

philberty:
We need it on the enum because enum comparisons are not correct because we need that derive to do it using that enum discriminant intrinsic

PR #3356, opened 2025-01-07
gccrs: fix ICE in borrows to invalid expressions

philberty:
We need to check if the borrowed value is valid before creating the reference type. Otherwise this will lead to an ICE.

Fixes Rust-GCC#3140

gcc/rust/ChangeLog:

	* typecheck/rust-hir-type-check-expr.cc (TypeCheckExpr::visit): check for error
	* typecheck/rust-tyty-call.cc (TypeCheckCallExpr::visit): likewise and remove debug error

gcc/testsuite/ChangeLog:

	* rust/compile/issue-3046.rs: remove old error message
	* rust/compile/issue-3140.rs: New test.


CohenArthur (approved):
LGTM!

PR #3357, opened 2025-01-07
gccrs: cleanup our enum type layout to be closer to rustc

philberty:
This changes our enum type layout so for example:

```rust
  enum Foo {
      A,
      B,
      C(char),
      D { x: i32, y: i32 },
  }
```

Used to get layed out like this in gccrs:

```c
  union {
    struct A { int RUST$ENUM$DISR; };
    struct B { int RUST$ENUM$DISR; };
    struct C { int RUST$ENUM$DISR; char __0; };
    struct D { int RUST$ENUM$DISR; i64 x; i64 y; };
  }
```

This has some issues notably with the constexpr because this is just a giant union it means its not simple to constify what enum variant we are looking at because the discriminant is a mess.

This now gets layed out as:

```c
  struct {
     int RUST$ENUM$DISR;
     union {
         struct A { };
         struct B { };
         struct C { char __0; };
         struct D { i64 x; i64 y; };
     } payload;
  }
```

This layout is much cleaner and allows for our constexpr to work properly.

gcc/rust/ChangeLog:

	* backend/rust-compile-expr.cc (CompileExpr::visit): new layout
	* backend/rust-compile-pattern.cc (CompilePatternCheckExpr::visit): likewise
	(CompilePatternBindings::visit): likewise
	* backend/rust-compile-resolve-path.cc: likewise
	* backend/rust-compile-type.cc (TyTyResolveCompile::visit): implement new layout
	* rust-gcc.cc (constructor_expression): get rid of useless assert

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -538,24 +538,32 @@ CompileExpr::visit (HIR::StructExprStructFields &struct_expr)
 	}
     }
 
-  // the constructor depends on whether this is actually an enum or not if
-  // its an enum we need to setup the discriminator
-  std::vector<tree> ctor_arguments;
-  if (adt->is_enum ())
+  if (!adt->is_enum ())
     {
-      HIR::Expr &discrim_expr = variant->get_discriminant ();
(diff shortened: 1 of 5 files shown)


philberty:
our gdb output looks much better with this too.

PR #3359, opened 2025-01-08
nr2.0: Resolve `Self` inside impl blocks

powerboat9:


CohenArthur (approved):
thanks, that's a nice trick! LGTM!

PR #3360, opened 2025-01-09
gccrs: match arms are a LUB

philberty:
Unify rules are not the same as coercion rules. The coercion of ! is allowed to any type but not for a unify site which is different.

gcc/rust/ChangeLog:

	* backend/rust-compile-expr.cc (CompileExpr::visit): implement coercion
	* typecheck/rust-hir-type-check-expr.cc (TypeCheckExpr::visit): this is a least upper bound coercion
	* typecheck/rust-unify.cc (UnifyRules::go): remove unify ! coercion


CohenArthur (approved):
I feel like I had to add the bit in `rust-unify.cc` for some other testcase, but if nothing is failing then I'm completely fine with it. LGTM!

PR #3361, opened 2025-01-10
Revert "refactor all attribute functions in rust-collect-lang-items.cc"

CohenArthur:
Reverts Rust-GCC/gccrs#3338

@saeitoshi-10 the PR's content is not bad but we need to fix the commit message. Can you reopen a PR?


CohenArthur:
We'll have to remember to skip those two commits when upstreaming

PR #3363, opened 2025-01-10
Remove dead code related to external functions

powerboat9:


philberty (approved):
LGTM

P-E-P (approved):
Looks good. Although at first I couldn't remember why we removed `NamedFunctionParam`, answer lies in 513b0154ab1f00e92d2072a5858f585510599b9b. We introduced a `Param` abstract class as well as a `VariadicParam` class, extern function parameters could thus be handled in the same way as usual arguments.

PR #3366, opened 2025-01-15
Refactor lang item paths

CohenArthur:
This PR contains the base for #3343 and refactors the way lang item paths are handled, either as path in expressions or as type paths. The changes are required for #3343 to work properly.

PR #3367, opened 2025-01-15
Improve `derive(Clone)` for generic types

CohenArthur:
When deriving `Clone` for a type with generic, we need to bound the impl's generic to be `: Clone`. e.g. the derive of `struct Foo<T>` should always be `impl<T: Clone> Clone for Foo<T>` even if `T` isn't necessarily bound by `Clone` in the struct's definition. Requires #3343 and thus #3366 so only review the last 3 commits


CohenArthur:
Ah, we also need to do that for `derive(Copy)` I think

PR #3370, opened 2025-01-16
Add basic handling for auto-traits

CohenArthur:
This will get us closer to compiling core by adding auto trait bounds unconditionally. This is extremely wrong and we will need to fix it when properly implementing opting out of auto-traits.

A look at the error we were getting with the testcase before that PR:

```rust
auto-traits.rs:12:15: error: bounds not satisfied for i32 ‘Send’ is not satisfied [E0277]
    6 | fn take_send(_: &dyn Send) {}
      |                      ~~~~
......
   12 |     take_send(&a);
      |               ^
auto-traits.rs:12:15: error: mismatched types, expected ‘& dyn [Send<Self>]’ but got ‘& i32’ [E0308]
    6 | fn take_send(_: &dyn Send) {}
      |              ~
......
   12 |     take_send(&a);
      |               ^
auto-traits.rs:13:15: error: bounds not satisfied for i32 ‘Sync’ is not satisfied [E0277]
    7 | fn take_sync(_: &dyn Sync) {}
      |                      ~~~~
......
   13 |     take_sync(&a);
      |               ^
auto-traits.rs:13:15: error: mismatched types, expected ‘& dyn [Sync<Self>]’ but got ‘& i32’ [E0308]
    7 | fn take_sync(_: &dyn Sync) {}
      |              ~
......
   13 |     take_sync(&a);
      |               ^
```


gcc/testsuite/rust/compile/auto_traits2.rs
```diff
@@ -15,12 +15,11 @@ fn foo(a: &(dyn A + Send + Sync)) {
...
 impl A for S {
-    fn a_method(&self) {}
+    fn a_method(&self) {} // { dg-warning "unused name" }
 }
 
 fn main() {
     let s = S;
 
-    foo(&s); // { dg-error "bounds not satisfied" }
-             // { dg-error "mismatched type" "" { target *-*-* } .-1 }
```
P-E-P, line 25: I'm unsure about what has changed here, care to give more details?

CohenArthur: so now that we can actually apply auto traits, this code will work because `S` is automatically `Send` and `Sync`

P-E-P (approved):
Looks great!

philberty (approved):
LGTM

PR #3373, opened 2025-01-20
Nr2 closure captures

P-E-P:
Remove usage of nr1 resolver with closure captures.

### DIFF
--- a/gcc/rust/resolve/rust-late-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-late-name-resolver-2.0.cc
@@ -390,8 +390,9 @@ Late::visit (AST::GenericArg &arg)
   DefaultResolver::visit (arg);
 }
 
-void
-Late::visit (AST::ClosureExprInner &closure)
+template <class Closure>
+static void
+add_captures (Closure &closure, NameResolutionContext &ctx)
 {
   auto vals = ctx.values.peek ().get_values ();
   for (auto &val : vals)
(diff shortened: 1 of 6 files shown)


gcc/rust/resolve/rust-late-name-resolver-2.0.cc
```diff
@@ -345,5 +345,18 @@ Late::visit (AST::GenericArg &arg)
...
+  auto vals = ctx.values.peek ().get_values ();
+  for (auto &val : vals)
+    {
+      ctx.mappings.add_capture (closure.get_node_id (),
+				val.second.get_node_id ());
+    }
+
+  DefaultResolver::visit (closure);
+}
+
```
powerboat9, line 360: Would also have to handle `ClosureExprInnerTyped` -- maybe with some `visit (static_cast <AST::ClosureExpr &> (closure))` shenanigans to avoid duplicating code?

powerboat9 (commented):
Otherwise, looks great

P-E-P:
I had to delete the commit with the now passing nr2 tests because they contain a multi segment TypePath.

gcc/rust/resolve/rust-late-name-resolver-2.0.cc
P-E-P: I used a simple templated function instead, I don't want anyone to be more confused, we've got 3 degrees of inheritance with this visitor, I already can't expect people to get it right the first time 100% of the time so I'd prefer if we could avoid a static upcast :smile:  Thanks for noticing!

PR #3375, opened 2025-01-21
lang-items: Add Result and Try lang items

CohenArthur:
This is a necessary step for starting to handle the desugar of `<expr>?`. In Rust 1.49, the desugar looks like this:

```rust
    /// match Try::into_result(<expr>) {
    ///     Ok(val) => #[allow(unreachable_code)] val,
    ///     Err(err) => #[allow(unreachable_code)]
    ///                 return Try::from_error(From::from(err)),
    /// }
```

which uses 4 out of the 5 lang items added in this PR. the remaining lang item is used for try blocks, an unstable feature which I'm not sure we need to support at first for `std`


philberty (approved):
LGTM

PR #3376, opened 2025-01-21
gccrs: add support for ref literal patterns

philberty:
Fixes Rust-GCC#3174

### DIFF
--- a/gcc/testsuite/rust/compile/issue-3174.rs
+++ b/gcc/testsuite/rust/compile/issue-3174.rs
@@ -1,0 +1,28 @@
+extern "C" {
+    fn printf(s: *const i8, ...);
+}
+
+enum Option {
+    Some(i32),
+    None,
+}
+
+impl Option {
+    fn add(&mut self) {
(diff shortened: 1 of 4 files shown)


gcc/testsuite/rust/compile/issue-3174.rs
CohenArthur, line 1: should we make this an execution test as well?

CohenArthur:
Also @philberty I think your commit message should be `Fixes gccrs#3174` instead to not confuse people on the GCC side

gcc/testsuite/rust/compile/issue-3174.rs
philberty: right enough could do will update it

philberty:
Wierd must be the way github renders the link to the issue but it is prefixed:

image

CohenArthur:
> Wierd must be the way github renders the link to the issue but it is prefixed:
> 
> image

oh you're right, damn! this is so weird. sorry!

PR #3377, opened 2025-01-21
Check for type paths nr2.0 can't handle yet

powerboat9:
Some of our tests only work with name resolution 2.0 because the latter misinterprets type paths. This change should cause the compiler to error out if it would otherwise misinterpret a type path. A fix for type path resolution isn't included in this comment, since doing so would make it harder to track the meaningfulness of test regressions.


philberty (approved):
seems like a good idea for now to isolate test cases

P-E-P (approved):

PR #3378, opened 2025-01-22
Prepare lang-item `{AST, HIR}::PathInExpressions`

CohenArthur:
This is an important base for desugaring expression lang items, e.g. when calling something like `IntoIterator::into_iter(i)`. This changes our AST and HIR visitors to make it possible to use lang items in more places which will be necessary.


gcc/rust/hir/rust-ast-lower.cc
```diff
@@ -494,6 +477,18 @@ ASTLoweringExprWithBlock::visit (AST::MatchExpr &expr)
 void
 ASTLowerPathInExpression::visit (AST::PathInExpression &expr)
 {
+  auto crate_num = mappings.get_current_crate ();
+  Analysis::NodeMapping mapping (crate_num, expr.get_node_id (),
+				 mappings.get_next_hir_id (crate_num),
+				 UNKNOWN_LOCAL_DEFID);
+
+  if (expr.get_path_kind () == AST::Path::Kind::LangItem)
```
CohenArthur, line 485: Here too :grimacing:

gcc/rust/hir/tree/rust-hir-path.cc
```diff
@@ -133,6 +133,9 @@ PathExprSegment::operator= (PathExprSegment const &other)
 void
 PathPattern::iterate_path_segments (std::function<bool (PathExprSegment &)> cb)
 {
+  // FIXME: Is this necessary?
```
CohenArthur, line 136: Yes, remove the comment

gcc/rust/hir/tree/rust-hir.cc
```diff
@@ -1210,6 +1211,9 @@ ClosureExpr::as_string () const
 std::string
 PathPattern::as_string () const
 {
+  if (kind == PathPattern::Kind::LangItem)
+    return LangItem::ToString (*lang_item);
```
CohenArthur, line 1215: Use `LangItem::PrettyString` instead

gcc/rust/resolve/rust-ast-resolve-path.cc
```diff
@@ -49,6 +50,10 @@ ResolvePath::go (AST::SimplePath &expr)
 NodeId
 ResolvePath::resolve_path (AST::PathInExpression &expr)
 {
+  if (expr.get_path_kind () == AST::Path::Kind::LangItem)
```
CohenArthur, line 53: Use `.is_lang_item()`

gcc/rust/resolve/rust-early-name-resolver.cc
```diff
@@ -156,9 +156,10 @@ EarlyNameResolver::visit (AST::ConstGenericParam &)
 void
 EarlyNameResolver::visit (AST::PathInExpression &path)
 {
-  for (auto &segment : path.get_segments ())
-    if (segment.has_generic_args ())
-      resolve_generic_args (segment.get_generic_args ());
+  if (path.get_path_kind () != AST::Path::Kind::LangItem)
```
CohenArthur, line 159: Here too

gcc/rust/typecheck/rust-hir-type-check-path.cc
```diff
@@ -179,20 +183,72 @@ void
...
+	  HIR::EnumItem *enum_item = enum_item_lookup.second;
+	  resolved_node_id = enum_item->get_mappings ().get_nodeid ();
+
+	  // insert the id of the variant we are resolved to
+	  context->insert_variant_definition (expr.get_mappings ().get_hirid (),
+					      variant_id);
+
+	  query_type (variant_id, &infered);
+	  infered = SubstMapper::InferSubst (infered, expr.get_locus ());
+	}
```
CohenArthur, line 218: This is super dodgy, @philberty please check it good lol

gcc/rust/util/rust-lang-item.cc
```diff
@@ -248,4 +248,10 @@ LangItem::NegationOperatorToLangItem (NegationOperator op)
   rust_unreachable ();
 }
 
+bool
+LangItem::IsEnumVariant (LangItem::Kind type)
+{
+  return type == Kind::OPTION_NONE || type == Kind::OPTION_SOME;
```
CohenArthur, line 254: Here we need to add `Result::Ok` and `Result::Err`

gcc/rust/typecheck/rust-hir-type-check-path.cc
liamnaddell: If I could ask, how could you end up with a lang-item path that DOESN'T need substitution inferencing? I'm curious because I think the only way you could generate an option type that doesn't need inferencing would be to generate a lang-item path like Option::<i32>::Some(3), but that wouldn't be possible since lang-item paths wouldn't have PathExprSegments which can be generic arguments

liamnaddell: I think it's probably better to keep it I would imagine, since if it ever becomes an issue, the comment clarifies that ALWAYS doing substitution inference is not a design choice

liamnaddell:
LGTM

gcc/rust/backend/rust-compile-resolve-path.cc
```diff
@@ -35,15 +35,29 @@ namespace Compile {
 void
 ResolvePathRef::visit (HIR::QualifiedPathInExpression &expr)
 {
-  resolved = resolve (expr.get_final_segment ().get_segment (),
-		      expr.get_mappings (), expr.get_locus (), true);
+  auto final_segment = HIR::PathIdentSegment::create_error ();
+  if (expr.is_lang_item ())
+    final_segment
+      = HIR::PathIdentSegment (LangItem::ToString (expr.get_lang_item ()));
```
liamnaddell, line 41: Could this end up getting confused if there are multiple `Some` declarations floating around? 

I.e. 

fn Some () {
 let _ = option_env!("PWD"); // Yields lang-item path to Some("/tmp")
}

gcc/rust/typecheck/rust-hir-type-check-path.cc
philberty: there might be a change needed for the generics but would need to try it out a bit myself to try our some tests on it

philberty (approved):
LGTM lets get this merged soon

gcc/rust/backend/rust-compile-resolve-path.cc
CohenArthur: Maybe - I'm adding a different implementation right now which is more precise and makes use of the lang item's known NodeId

liamnaddell: I think the version you have up now works, at least from tracing the code by eye.

gcc/rust/backend/rust-compile-resolve-path.cc
```diff
@@ -32,32 +32,39 @@
 namespace Rust {
 namespace Compile {
 
-void
-ResolvePathRef::visit (HIR::QualifiedPathInExpression &expr)
+template <typename T>
```
philberty, line 35: I always forget you can do this!!! thanks for reminding me of this trick :D

gcc/rust/backend/rust-compile-resolve-path.cc
```diff
@@ -32,32 +32,39 @@
...
-  auto final_segment = HIR::PathIdentSegment::create_error ();
   if (expr.is_lang_item ())
-    final_segment
-      = HIR::PathIdentSegment (LangItem::ToString (expr.get_lang_item ()));
+    {
+      auto lang_item
+	= Analysis::Mappings::get ().get_lang_item_node (expr.get_lang_item ());
+
+      // FIXME: Is that correct? :/
+      auto final_segment
```
philberty, line 45: yes this should work the only reason i passed the node around was because it was handy to get the location and node id so in theory it can fall back to the query system interface.

Not part of your pr but yeah this has given me alot of though of loads of good-first-pr's to cleanup the interfaces here
its super confusing for new people.

gcc/rust/backend/rust-compile-resolve-path.cc
```diff
@@ -32,32 +32,39 @@
...
+	= HIR::PathIdentSegment (LangItem::ToString (expr.get_lang_item ()));
+
+      return resolve_with_node_id (final_segment, expr.get_mappings (),
+				   expr.get_locus (), true, lang_item);
+    }
   else
```
philberty, line 51: I think the only change i might do here is get rid of the else and just return that

gcc/rust/backend/rust-compile-resolve-path.cc
```diff
@@ -32,32 +32,39 @@
...
-  auto final_segment = HIR::PathIdentSegment::create_error ();
-  if (expr.is_lang_item ())
-    final_segment
-      = HIR::PathIdentSegment (LangItem::ToString (expr.get_lang_item ()));
-  else
-    final_segment = expr.get_final_segment ().get_segment ();
-
-  resolved
-    = resolve (final_segment, expr.get_mappings (), expr.get_locus (), true);
+  resolved = resolve_path_like (expr);
```
philberty, line 67: i love this so much

gcc/rust/backend/rust-compile-resolve-path.cc
```diff
@@ -106,42 +113,17 @@ ResolvePathRef::attempt_constructor_expression_lookup (
 }
 
 tree
-ResolvePathRef::resolve (const HIR::PathIdentSegment &final_segment,
-			 const Analysis::NodeMapping &mappings,
-			 location_t expr_locus, bool is_qualified_path)
+ResolvePathRef::resolve_with_node_id (
+  const HIR::PathIdentSegment &final_segment,
+  const Analysis::NodeMapping &mappings, location_t expr_locus,
+  bool is_qualified_path, NodeId resolved_node_id)
```
philberty, line 119: I have a feeling the is_qualified_path is no longer used here now which is neat

philberty (approved):
LGTM nice work the artification ident segment will work fine here

gcc/rust/backend/rust-compile-resolve-path.cc
CohenArthur: ah I kept it cause we need it for `query_compile`

gcc/rust/backend/rust-compile-resolve-path.cc
CohenArthur: thanks! fixed in the latest version

PR #3379, opened 2025-01-22
backend: Allow anything as a match scrutinee

CohenArthur:
This commit is required for this type of pattern, which isn't really pattern matching:

```rust
match something {
    new_binding => { ... }
}
```

This is required for desugaring for-loops and error propagation expressions.


philberty (approved):
LGTM with the cmp.rs stuff i am doing i think its going to be time next month to rework the code-gen of math expressions anyway to do if statements potentially instead of switches like we've talked about before

PR #3380, opened 2025-01-23
gccrs: catch missing guard for optional result

philberty:
When we lookup here it returns an optional which can lead to a crash because it needs a guard if it has a value.

gcc/rust/ChangeLog:

	* backend/rust-compile-resolve-path.cc (HIRCompileBase::query_compile): add guard

gcc/testsuite/ChangeLog:

	* rust/compile/nr2/exclude: these tests now work it seems


CohenArthur (approved):
LGTM!

PR #3383, opened 2025-01-23
gccrs: add new -frust-overflow-checks to control overflow checks

philberty:
This will be crucial for more complex gimple debugging to make it easier to follow the code vs the original rust code.

gcc/rust/ChangeLog:

	* backend/rust-compile-expr.cc (CompileExpr::visit): disable overflow checks
	* lang.opt: new flag

### DIFF
--- a/gcc/rust/lang.opt
+++ b/gcc/rust/lang.opt
@@ -225,4 +225,8 @@ Enum(frust_panic) String(unwind) Value(0)
 EnumValue
 Enum(frust_panic) String(abort) Value(1)
 
+frust-overflow-checks
+Rust Var(flag_overflow_checks) Init(1)
+Enable the overflow checks in code generation
+
 ; This comment is to ensure we retain the blank line above.
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -159,27 +159,28 @@ CompileExpr::visit (HIR::ArithmeticOrLogicalExpr &expr)
       return;
     }
 
-  if (ctx->in_fn () && !ctx->const_context_p ())
-    {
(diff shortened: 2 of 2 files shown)


powerboat9:
Should probably be `-frust-overflow-checks` with a default setting of true, so gcc gives us `-fno-rust-overflow-checks` for free

CohenArthur:
> Should probably be `-frust-overflow-checks` with a default setting of true, so gcc gives us `-fno-rust-overflow-checks` for free

If we can do that I'd like it too

philberty:
sounds good i will change it over

gcc/rust/lang.opt
```diff
@@ -225,4 +225,8 @@ Enum(frust_panic) String(unwind) Value(0)
 EnumValue
 Enum(frust_panic) String(abort) Value(1)
 
+frust-overflow-checks
+Rust Var(flag_overflow_checks) Init(1)
+Enable the overflow checks in code generation
+
```
CohenArthur, line 231: Perfect, and I tried it out with `-fno-rust-overflow-checks` and it works

CohenArthur (approved):
thank you! LGTM!

gcc/rust/lang.opt
```diff
@@ -225,4 +225,8 @@ Enum(frust_panic) String(unwind) Value(0)
 EnumValue
 Enum(frust_panic) String(abort) Value(1)
 
+frust-overflow-checks
```
thesamesam, line 228: If it's default-enabled, shouldn't `-fno-rust-overflow-checks` be the documented variant instead?

powerboat9: https://gcc.gnu.org/onlinedocs/gccint/Option-file-format.html

It looks like putting a tab character in the help text would be used for that

PR #3385, opened 2025-01-23
ast: Add new Expr::Kinds

CohenArthur:
Collapses all of the OperatorExprs into Expr instead of first having to check for OperatorExpr and then check for each OperatorExpr::Kind.

### DIFF
--- a/gcc/rust/ast/rust-expr.h
+++ b/gcc/rust/ast/rust-expr.h
@@ -407,6 +407,8 @@ public:
   bool get_is_double_borrow () const { return double_borrow; }
   bool is_raw_borrow () const { return raw_borrow; }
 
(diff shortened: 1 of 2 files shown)


gcc/rust/ast/rust-expr.h
```diff
@@ -2139,6 +2168,8 @@ class CallExpr : public ExprWithoutBlock
     return *function;
   }
 
+  std::unique_ptr<Expr> &get_function_expr_ptr () { return function; }
```
P-E-P, line 2171: Unless you need to reseat that pointer a returning an `Expr &` alone should be sufficient.

CohenArthur: I need to reseat it for example for the `?` operator desugar. at least with the current design

P-E-P (approved):

PR #3386, opened 2025-01-23
ast: builder: Add Return expression builder

CohenArthur:
Add new methods for building `return <x>;` expressions/statements in our AST.

### DIFF
--- a/gcc/rust/ast/rust-ast-builder.cc
+++ b/gcc/rust/ast/rust-ast-builder.cc
@@ -214,6 +214,13 @@ Builder::block (std::vector<std::unique_ptr<Stmt>> &&stmts,
 					       LoopLabel::error (), loc, loc));
 }
 
+std::unique_ptr<Expr>
+Builder::return_expr (std::unique_ptr<Expr> &&to_return)
+{
+  return std::unique_ptr<Expr> (
+    new ReturnExpr (std::move (to_return), {}, loc));
+}
+
 std::unique_ptr<Stmt>
 Builder::let (std::unique_ptr<Pattern> pattern, std::unique_ptr<Type> type,
 	      std::unique_ptr<Expr> init) const
--- a/gcc/rust/ast/rust-ast-builder.h
+++ b/gcc/rust/ast/rust-ast-builder.h
@@ -85,6 +85,10 @@ public:
 			       std::unique_ptr<Expr> &&tail_expr
 			       = nullptr) const;
 
+  /* Create an early return expression with an optional expression */
+  std::unique_ptr<Expr> return_expr (std::unique_ptr<Expr> &&to_return
+				     = nullptr);
+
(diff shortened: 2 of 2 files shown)


gcc/rust/ast/rust-ast-builder.cc
```diff
@@ -214,6 +214,13 @@ Builder::block (std::vector<std::unique_ptr<Stmt>> &&stmts,
 					       LoopLabel::error (), loc, loc));
 }
 
+std::unique_ptr<Expr>
+Builder::return_expr (std::unique_ptr<Expr> &&to_return)
+{
+  return std::unique_ptr<Expr> (
+    new ReturnExpr (std::move (to_return), {}, loc));
```
P-E-P, line 221: Couldn't we use `make_unique here` ? I believe we can even get rid of our implementation since baseline cpp version has been bumped to cpp14.

gcc/rust/ast/rust-ast-builder.h
```diff
@@ -85,6 +85,10 @@ class Builder
 			       std::unique_ptr<Expr> &&tail_expr
 			       = nullptr) const;
 
+  /* Create an early return expression with an optional expression */
+  std::unique_ptr<Expr> return_expr (std::unique_ptr<Expr> &&to_return
+				     = nullptr);
```
P-E-P, line 90: `std::unique_ptr<Expr> &&` I wish it was an optional instead, this would draw a clear line and we would be able to easily catch rogue null pointers.

CohenArthur: I think we'd need to change the AST for this then, because the `ReturnExpr` class only keeps a unique_ptr. So the builder's code would look like this:

```c++

std::unique_ptr<Expr>
Builder::return_expr (tl::optional<std::unique_ptr<Expr>> &&to_return)
{
  if (to_return)
    return std::unique_ptr<Expr> (
      new ReturnExpr (std::move (*to_return), {}, loc));

  return std::unique_ptr<Expr> (new ReturnExpr (nullptr, {}, loc));
}
```

which is quite heavy

gcc/rust/ast/rust-ast-builder.cc
CohenArthur: tried using it and it complains about not finding a matching constructor:

```c++
../../gcc/rust/ast/rust-ast-builder.cc:220:39: error: no matching function for call to ‘make_unique<Rust::AST::ReturnExpr>(std::remove_reference<std::unique_ptr<Rust::AST::Expr>&>::type, <brace-enclosed initializer list>, location_t&)’
  220 |   return std::make_unique<ReturnExpr> (std::move (to_return), {}, loc);
      |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

but maybe I'm missing something

PR #3391, opened 2025-01-28
Resolve `Self` prefix `PathInExpression` with nr2

P-E-P:
Fixing resolution of some path with `Self` prefix.


powerboat9:
It seems like path resolution should, if it can't resolve a segment, attempt to resolve that segment as if it was the last segment.

powerboat9:
svlBSJ

PR #3392, opened 2025-01-29
Desugar for-loops in the AST

CohenArthur:
I'm not a big fan of mutating the AST to do this, however this enables us to desugar for-loops easily. The alternative is desugaring during the lowering phase, which has the issue of not inserting any information about the generated nodes into our name resolver. This leads to failures during typechecking which is annoying.

The best thing would be to figure out a way to insert resolved nodes when we do the Desugar, but it's a bit tricky.

### DIFF
--- a/gcc/rust/ast/rust-desugar-for-loops.cc
+++ b/gcc/rust/ast/rust-desugar-for-loops.cc
@@ -1,0 +1,204 @@
+// Copyright (C) 2025 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#include "rust-desugar-for-loops.h"
+#include "rust-ast-visitor.h"
+#include "rust-ast.h"
+#include "rust-hir-map.h"
+#include "rust-path.h"
+#include "rust-pattern.h"
+#include "rust-stmt.h"
+#include "rust-expr.h"
+#include "rust-ast-builder.h"
+
+namespace Rust {
+namespace AST {
+
+DesugarForLoops::DesugarForLoops () {}
+
+void
+DesugarForLoops::go (AST::Crate &crate)
+{
+  DefaultASTVisitor::visit (crate);
+}
+
+static void
+replace_for_loop (std::unique_ptr<Expr> &for_loop,
+		  std::unique_ptr<Expr> &&expanded)
+{
+  for_loop = std::move (expanded);
+}
+
+MatchArm
+DesugarForLoops::DesugarCtx::make_match_arm (std::unique_ptr<Pattern> &&path)
+{
+  auto patterns = std::vector<std::unique_ptr<Pattern>> ();
+  patterns.emplace_back (std::move (path));
+
+  return MatchArm (std::move (patterns), loc);
+}
+
+MatchCase
+DesugarForLoops::DesugarCtx::make_break_arm ()
+{
+  auto arm = make_match_arm (std::unique_ptr<Pattern> (new PathInExpression (
+    builder.path_in_expression (LangItem::Kind::OPTION_NONE))));
+
+  auto break_expr = std::unique_ptr<Expr> (
+    new BreakExpr (Lifetime::error (), nullptr, {}, loc));
+
+  return MatchCase (std::move (arm), std::move (break_expr));
+}
+
+MatchCase
+DesugarForLoops::DesugarCtx::make_continue_arm ()
+{
+  auto val = builder.identifier_pattern (DesugarCtx::continue_pattern_id);
+
+  auto patterns = std::vector<std::unique_ptr<Pattern>> ();
+  patterns.emplace_back (std::move (val));
+
+  auto pattern_item = std::unique_ptr<TupleStructItems> (
+    new TupleStructItemsNoRange (std::move (patterns)));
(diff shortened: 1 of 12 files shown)


philberty:
Although we wanted to do this desugar in the HIR, i see this as our first pass so that we show that our internals are 'working' and i bet down the line we will figure it out.

I bet there is a way that during HIR lowering we actually decouple the name resolution entirely by linking name resolution information in a different way to not relay on the ast_node_id but use something like link to the HIR node directly or something else. It would be a fairly big refactor which is basically not worth the effort right now but would be a good google summer of code project we could scope out.

philberty:
As always great work sir

gcc/testsuite/rust/execute/torture/for-loop2.rs
```diff
@@ -0,0 +1,544 @@
...
+    #[lang = "next"]
+    fn next(&mut self) -> Option<Self::Item>;
+}
+
+impl<A: Step> Iterator for Range<A> {
+    type Item = A;
+
+    fn next(&mut self) -> Option<A> {
+        if self.start < self.end {
+            // We check for overflow here, even though it can't actually
```
philberty, line 501: i'd love to know more about this bit in interators in rustc why this was introduced i wonder what the test case was

philberty (approved):
Lets gooooo

gcc/rust/ast/rust-desugar-for-loops.cc
```diff
@@ -0,0 +1,198 @@
+// Copyright (C) 2024 Free Software Foundation, Inc.
```
P-E-P, line 1: Should probably be 2025

gcc/rust/ast/rust-desugar-for-loops.cc
```diff
@@ -0,0 +1,198 @@
...
+DesugarForLoops::DesugarForLoops () {}
+
+void
+DesugarForLoops::go (AST::Crate &crate)
+{
+  DefaultASTVisitor::visit (crate);
+}
+
+void
+replace_for_loop (std::unique_ptr<Expr> &for_loop,
```
P-E-P, line 41: could be made static / put into an anonymous namespace. I don't think we want to leak this function out of the TU anyway.

gcc/rust/ast/rust-desugar-for-loops.cc
```diff
@@ -0,0 +1,198 @@
...
+  auto break_expr = std::unique_ptr<Expr> (
+    new BreakExpr (Lifetime::error (), nullptr, {}, loc));
+
+  return MatchCase (std::move (arm), std::move (break_expr));
+}
+
+MatchCase
+DesugarForLoops::DesugarCtx::make_continue_arm ()
+{
+  auto val = builder.identifier_pattern ("val");
```
P-E-P, line 71: Should we use some underscore before `val` ?

P-E-P, line 71: Also I would probably use some constexpr string or char* with some descriptive name. That way:
- grepping val would lead to an explanation
- renaming all those values wouldn't be too difficult, one location only
- a tool could rename all the variable easily

gcc/rust/ast/rust-desugar-for-loops.h
```diff
@@ -0,0 +1,103 @@
...
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#ifndef RUST_DESUGAR_FOR_LOOPS
```
P-E-P, line 19: Most of our header guards use the suffix `_H`. Should we enforce it ?

gcc/rust/ast/rust-desugar-for-loops.h
```diff
@@ -0,0 +1,103 @@
+// Copyright (C) 2024 Free Software Foundation, Inc.
```
P-E-P, line 1: Once again the copyright year seems incorrect.

P-E-P (approved):
Looks great!

gcc/rust/ast/rust-desugar-for-loops.cc
CohenArthur: ah, right - I think we should use an even weirder character like a `.` or a `#` to really prevent it from leaking out

P-E-P: Tbh I don't really like the fact that we could name it, it should really be some kind of long unameable identifier that changes depending on the locus and a few other factor or something along this. Not sure how we should do this properly.

CohenArthur: well we can't name it if theidentifier starts with `#` because it won't parse. so a user cannot name it. the other option is to add extra hygiene checks (e.g. having a special-cased location for generated code like rustc does) or generating those identifiers in the HIR, after name resolution has gone through

CohenArthur: we could also use `$__next` or `.__next` or any other kind of invalid identifier which won't parse in Rust - that's enough to prevent the user from coliding/using these identifiers

gcc/rust/ast/rust-desugar-for-loops.h
```diff
@@ -0,0 +1,108 @@
...
+
+    MatchArm make_match_arm (std::unique_ptr<Pattern> &&pattern);
+    MatchCase make_break_arm ();
+    MatchCase make_continue_arm ();
+    std::unique_ptr<Stmt> statementify (std::unique_ptr<Expr> &&expr);
+
+    constexpr static const char *continue_pattern_id = "#val";
+    constexpr static const char *next_value_id = "#__next";
+    constexpr static const char *iter_id = "#iter";
+    constexpr static const char *result_id = "#result";
```
P-E-P, line 96: Should we use uppercase letter instead ?

CohenArthur: I think in the end it doesn't make a difference, I can make the change if you want

P-E-P: Yeah, it would not even be used anywhere else, let's keep it.

P-E-P (approved):

PR #3393, opened 2025-01-29
inline-asm: Fix some warnings

CohenArthur:


CohenArthur:
@badumbatish I'm going to make some changes to inline assembly, if you have time to review some PRs I'd appreciate your input :) I'll ping you

badumbatish:
> @badumbatish I'm going to make some changes to inline assembly, if you have time to review some PRs I'd appreciate your input :) I'll ping you

Happy to help :) i'll try to put in some reviews soon!

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -39,13 +39,18 @@ std::map<AST::InlineAsmOption, std::string> InlineAsmOptionMap{
 std::set<std::string> potentially_nonpromoted_keywords
   = {"in", "out", "lateout", "inout", "inlateout", "const", "sym", "label"};
 
+// Helper function strips the beginning and ending double quotes from a
+// string.
 std::string
 strip_double_quotes (const std::string &str)
 {
```
badumbatish, line 46: it's probably just a nitpick on myself but back then it didnt occur to me that somebody might use this on sth like str = "\'dsda\'" or "abcd" despite the name being `strip_double_quotes()`.

I think an assertion on the first and last character being double quotes would be nice :)

badumbatish (approved):

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -62,6 +62,8 @@ const BiMap<std::string, BuiltinMacro> MacroBuiltin::builtins = {{
   {"concat_idents", BuiltinMacro::ConcatIdents},
   {"module_path", BuiltinMacro::ModulePath},
   {"asm", BuiltinMacro::Asm},
+  // FIXME: Is that okay
+  {"llvm_asm", BuiltinMacro::Asm},
```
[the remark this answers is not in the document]
CohenArthur: So it has been removed in recent Rust versions, but is still used in Rust 1.49 - so we'll need to support some form of it. I think the only thing we really need to support is `llvm_asm!` in `core`, and I think we can use regular inline assembly for this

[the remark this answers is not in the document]
powerboat9: It looks like the only `llvm_asm!` usages in the rustc libraries are for `black_box`, which seems fairly simple, and a few `stdarch` things.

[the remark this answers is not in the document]
badumbatish: i see!

gcc/rust/expand/rust-macro-builtins-asm.cc
CohenArthur: Fixed in latest version

badumbatish:
LGTM :)

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -39,16 +39,28 @@ std::map<AST::InlineAsmOption, std::string> InlineAsmOptionMap{
...
 
+  rust_assert (!str.empty ());
+
+  rust_assert (str.front () == '\"');
+  rust_assert (str.back () == '\"');
+
+  // we have to special case empty strings which just contain a set of quotes
+  // so, if the string is "\"\"", just return ""
+  if (result.size () == 2)
+    return "";
```
powerboat9, line 57: Why is this special cased, anyways?

powerboat9: Might be a good idea to use number 4 from https://en.cppreference.com/w/cpp/string/basic_string/basic_string

CohenArthur: because of the assertion later on, which I did not want to tweak. since this is a static function for that one specific file I was hoping to keep it as simple and as tailored as possible - it doesn't leak anywhere else and doesn't need to be too complicated

PR #3394, opened 2025-01-29
WIP typecheck issues with Try

philberty:
When we pass generic types as the Self on traits they need to be handled via a special case on fntypes, ADT's and placeholders as the type params won't necessarily match up because of the implicit Self.

Fixes Rust-GCC#3382

gcc/rust/ChangeLog:

	* typecheck/rust-substitution-mapper.cc (SubstMapperInternal::visit): special case

gcc/testsuite/ChangeLog:

	* rust/compile/issue-3382.rs: New test.


philberty:
closing this down to open up sperate PR's for the fixes

PR #3404, opened 2025-01-31
Resolve lang-item PathInExpressions in nr2.0 late resolver

CohenArthur:
Needs #3094 so only review the last commit

@liamnaddell @powerboat9 how does that look to you?

### DIFF
--- a/gcc/rust/resolve/rust-late-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-late-name-resolver-2.0.cc
@@ -254,8 +254,14 @@ Late::visit (AST::PathInExpression &expr)
   // TODO: How do we have a nice error with `can't capture dynamic environment
   // in a function item` error here?
   // do we emit it in `get<Namespace::Labels>`?
+
   if (expr.is_lang_item ())
-    return;
+    {
+      ctx.map_usage (Usage (expr.get_node_id ()),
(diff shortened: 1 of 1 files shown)


liamnaddell (approved):
LGTM

gcc/rust/resolve/rust-late-name-resolver-2.0.cc
```diff
@@ -254,8 +254,14 @@ Late::visit (AST::PathInExpression &expr)
...
-    return;
+    {
+      ctx.map_usage (Usage (expr.get_node_id ()),
+		     Definition (Analysis::Mappings::get ().get_lang_item_node (
+		       expr.get_lang_item ())));
+      return;
+    }
 
   auto resolved
     = ctx.values.resolve_path (expr.get_segments ()).or_else ([&] () {
```
liamnaddell, line 267: When I was looking at this code, I got lost trying to figure out what resolve_path does, and how forever stacks work, and missed that the important part was calling map_usage.

CohenArthur: no worries it's all good :) dealing with the macro expansion is already complex enough for one PR haha

powerboat9 (approved):
LGTM

PR #3414, opened 2025-02-03
gccrs: Fix compilation of trait-items which map to impl items

philberty:
When we have paths such as Try::from_error the Try references the Trait and then from_error references the from_error trait item. So this resolves directly to a trait implementation which has the type:

  fn <Self> (v: placeholder) -> Self

Remember that a path such as: Try::from_error gets handled by doing a
clever substitution: <T? as Try>::from_error

The main piece here is that we resolve this function type and for monomoprhization we know this is a trait call but we know that all trait's have an implicit Self type param which in this case is Result<i32, i32> so when it comes to knowing which impl block this is we got rid of the old bad insert/lookup receiver hack and use the specified type to know which impl block we are talking about to generate the function.

The hard part here is inside rust-compil-item.cc, where we have the 'concete' type which is the trait item fntype of:

  fn <Result<i32, i32>> (v : i32) -> Result<i32,i32>

This does not really match the signiture of the impl item for Result which is:

  fn <T, E> (v: i32) -> Result<T, E>

So what we need to do is actually infer this by inputing inference variables on the impl fntype then unify'ing the trait object to this to compute the types of this to monomorphize this.

Fixes Rust-GCC#3381

gcc/rust/ChangeLog:

	* backend/rust-compile-expr.cc (CompileExpr::visit): remove receiver interface
	* backend/rust-compile-item.cc (CompileItem::visit): monomorphize trait to impl item
	* backend/rust-compile-resolve-path.cc (HIRCompileBase::query_compile): use trait item Self
	* typecheck/rust-hir-type-check-expr.cc (TypeCheckExpr::visit): remove receiver interface
	(TypeCheckExpr::resolve_fn_trait_call): likewise
	* typecheck/rust-hir-type-check-path.cc (TypeCheckExpr::visit): likewise
	(TypeCheckExpr::resolve_segments): likewise
	* typecheck/rust-hir-type-check-type.cc (TypeCheckType::visit): likewise
	* typecheck/rust-hir-type-check.h: likewise
	* typecheck/rust-typecheck-context.cc (TypeCheckContext::insert_receiver): remove
	(TypeCheckContext::lookup_receiver): remove

gcc/testsuite/ChangeLog:

	* rust/execute/torture/issue-3381.rs: New test.

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -1394,8 +1394,8 @@ CompileExpr::visit (HIR::MethodCallExpr &expr)
   TyTy::FnType *fntype = static_cast<TyTy::FnType *> (lookup_fntype);
 
   TyTy::BaseType *receiver = nullptr;
-  ok = ctx->get_tyctx ()->lookup_receiver (expr.get_mappings ().get_hirid (),
-					   &receiver);
+  ok = ctx->get_tyctx ()->lookup_type (
+    expr.get_receiver ().get_mappings ().get_hirid (), &receiver);
   rust_assert (ok);
 
   bool is_dyn_dispatch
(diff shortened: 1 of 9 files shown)


philberty:
this will need rebased before merge

PR #3415, opened 2025-02-04
derive(PartialEq): Add partial implementation (hehe)

### LINKED ISSUE #2992: Eq

CohenArthur:
Fixes #2992 

This adds a base for deriving `PartialEq` and `Eq` but misses some more complex stuff like struct enum items, as we need to add more to our nameres and codegen to handle these rebinding patterns.

~~We also need to rework this part of the expansion pass as `derive(Eq)` and `derive(PartialEq)` should both create *two* impls: one of their trait, and one of the associated `Structural*` trait~~ this is now done in this PR.

This is ready for reviewing and merging IMO as a first basic implementation of `derive(PartialEq)` even if it does not support all types of Rust enums

### DIFF
--- a/gcc/rust/expand/rust-derive-eq.cc
+++ b/gcc/rust/expand/rust-derive-eq.cc
@@ -27,6 +27,16 @@
 namespace Rust {
 namespace AST {
 
+DeriveEq::DeriveEq (location_t loc) : DeriveVisitor (loc) {}
+
+std::vector<std::unique_ptr<AST::Item>>
+DeriveEq::go (Item &item)
+{
+  item.accept_vis (*this);
+
+  return std::move (expanded);
+}
+
 std::unique_ptr<AssociatedItem>
 DeriveEq::assert_receiver_is_total_eq_fn (
   std::vector<std::unique_ptr<Type>> &&types)
@@ -98,33 +108,29 @@ DeriveEq::assert_type_is_eq (std::unique_ptr<Type> &&type)
   return builder.let (builder.wildcard (), std::move (full_path));
 }
 
-std::unique_ptr<Item>
-DeriveEq::eq_impl (
+std::vector<std::unique_ptr<Item>>
+DeriveEq::eq_impls (
   std::unique_ptr<AssociatedItem> &&fn, std::string name,
   const std::vector<std::unique_ptr<GenericParam>> &type_generics)
 {
   auto eq = builder.type_path ({"core", "cmp", "Eq"}, true);
+  auto steq = builder.type_path (LangItem::Kind::STRUCTURAL_TEQ);
 
   auto trait_items = vec (std::move (fn));
 
-  auto generics
+  auto eq_generics
     = setup_impl_generics (name, type_generics, builder.trait_bound (eq));
+  auto steq_generics = setup_impl_generics (name, type_generics);
 
-  return builder.trait_impl (eq, std::move (generics.self_type),
-			     std::move (trait_items),
-			     std::move (generics.impl));
-}
+  auto eq_impl = builder.trait_impl (eq, std::move (eq_generics.self_type),
+				     std::move (trait_items),
+				     std::move (eq_generics.impl));
+  auto steq_impl
+    = builder.trait_impl (steq, std::move (steq_generics.self_type),
+			  std::move (trait_items),
+			  std::move (steq_generics.impl));
 
-DeriveEq::DeriveEq (location_t loc) : DeriveVisitor (loc), expanded (nullptr) {}
-
-std::unique_ptr<AST::Item>
-DeriveEq::go (Item &item)
-{
-  item.accept_vis (*this);
-
-  rust_assert (expanded);
-
-  return std::move (expanded);
+  return vec (std::move (eq_impl), std::move (steq_impl));
 }
 
 void
@@ -135,9 +141,9 @@ DeriveEq::visit_tuple (TupleStruct &item)
   for (auto &field : item.get_fields ())
     types.emplace_back (field.get_field_type ().clone_type ());
 
-  expanded
-    = eq_impl (assert_receiver_is_total_eq_fn (std::move (types)),
-	       item.get_identifier ().as_string (), item.get_generic_params ());
+  expanded = eq_impls (assert_receiver_is_total_eq_fn (std::move (types)),
+		       item.get_identifier ().as_string (),
+		       item.get_generic_params ());
 }
 
 void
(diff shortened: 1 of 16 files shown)


gcc/rust/expand/rust-derive-eq.cc
```diff
@@ -0,0 +1,213 @@
...
+
+  return builder.let (builder.wildcard (), std::move (full_path));
+}
+
+std::vector<std::unique_ptr<Item>>
+DeriveEq::eq_impls (
+  std::unique_ptr<AssociatedItem> &&fn, std::string name,
+  const std::vector<std::unique_ptr<GenericParam>> &type_generics)
+{
+  auto eq = builder.type_path ({"core", "cmp", "Eq"}, true);
```
P-E-P, line 116: Should we make this path (or individual segments) available globally like weak keywords ?

gcc/rust/expand/rust-derive-eq.cc
```diff
@@ -0,0 +1,213 @@
...
+  for (auto &variant : item.get_variants ())
+    {
+      switch (variant->get_enum_item_kind ())
+	{
+	case EnumItem::Kind::Identifier:
+	case EnumItem::Kind::Discriminant:
+	  // nothing to do as they contain no inner types
+	  continue;
+	  case EnumItem::Kind::Tuple: {
+	    auto tuple = static_cast<EnumItemTuple &> (*variant);
```
P-E-P, line 176: We want to avoid the copy here `auto &tuple = static_cast<EnumItemTuple &> (*variant);`.

gcc/rust/expand/rust-derive-eq.cc
```diff
@@ -0,0 +1,213 @@
...
+
+	    for (auto &field : tuple.get_tuple_fields ())
+	      types.emplace_back (field.get_field_type ().clone_type ());
+
+	    break;
+	  }
+	  case EnumItem::Kind::Struct: {
+	    auto tuple = static_cast<EnumItemStruct &> (*variant);
```
P-E-P, line 184: Likewise.

gcc/rust/expand/rust-derive-partial-eq.cc
```diff
@@ -0,0 +1,313 @@
+// Copyright (C) 2020-2024 Free Software Foundation, Inc.
```
P-E-P, line 1: Date range is incorrect, should be `2025` instead of `2020-2024`.

gcc/rust/expand/rust-derive-partial-eq.cc
```diff
@@ -0,0 +1,313 @@
...
+}
+
+std::unique_ptr<AssociatedItem>
+DerivePartialEq::eq_fn (std::unique_ptr<Expr> &&cmp_expression,
+			std::string type_name)
+{
+  auto block = builder.block (tl::nullopt, std::move (cmp_expression));
+
+  auto self_type
+    = std::unique_ptr<TypeNoBounds> (new TypePath (builder.type_path ("Self")));
```
P-E-P, line 73: We may use `rust-keyword-values.h`, unsure if it would make it more readable though.

P-E-P (changes requested):

gcc/rust/expand/rust-derive-eq.cc
```diff
@@ -0,0 +1,213 @@
...
+
+  return builder.function ("assert_receiver_is_total_eq",
+			   vec (std::move (self)), {}, std::move (block));
+}
+
+std::unique_ptr<Stmt>
+DeriveEq::assert_param_is_eq ()
+{
+  auto eq_bound = std::unique_ptr<TypeParamBound> (
+    new TraitBound (builder.type_path ({"core", "cmp", "Eq"}, true), loc));
```
powerboat9, line 65: Should this be a lang item path?

gcc/testsuite/rust/compile/nr2/exclude
```diff
@@ -119,9 +119,11 @@ traits12.rs
 try-trait.rs
 derive-debug1.rs
 issue-3382.rs
-derive-default1.rs
 issue-3402-1.rs
 for-loop1.rs
 for-loop2.rs
 issue-3403.rs
+derive-default1.rs
```
powerboat9, line 126: unnecessary move

gcc/rust/expand/rust-derive-eq.cc
CohenArthur: no, `Eq` isn't a lang item in 1.49:

```rust
#[doc(alias = "==")]
#[doc(alias = "!=")]
#[stable(feature = "rust1", since = "1.0.0")]
pub trait Eq: PartialEq<Self> {
    // this method is used solely by #[deriving] to assert
    // that every component of a type implements #[deriving]
    // itself, the current deriving infrastructure means doing this
    // assertion without using a method on this trait is nearly
    // impossible.
    //
    // This should never be implemented by hand.
    #[doc(hidden)]
    #[inline]
    #[stable(feature = "rust1", since = "1.0.0")]
    fn assert_receiver_is_total_eq(&self) {}
}
```

gcc/rust/expand/rust-derive-eq.cc
CohenArthur: I don't think that's important here as this is only used when generating code, which isn't done too often. we rarely repeat paths, and making individual segments available doesn't really help a lot I feel

gcc/rust/expand/rust-derive-partial-eq.cc
CohenArthur: I don't think it would really help haha but if you want I can change it

gcc/rust/expand/rust-derive-eq.cc
powerboat9: Dang

powerboat9: Dang

gcc/rust/expand/rust-derive-eq.cc
powerboat9: We should also probably create two type paths here, one for creating `eq_impl` and one for creating `eq_generics` -- otherwise, two AST nodes sharing the same node id throws the 2.0 name resolver for a loop

powerboat9: Thinking it might be best to just ignore duplicate `map_usage` calls, as long as one usage is only ever mapped to one definition, in which case my previous suggestion can be ignored

CohenArthur: I think it's probably cleaner to have different node IDs for different nodes even if it makes the derive code very slightly messier

PR #3416, opened 2025-02-04
nr2.0: Resolve paths which start with Self

powerboat9:
Overlaps with #3391 (apologies to @P-E-P, but this should be a more general resolution), depends on #3220


P-E-P:
> Overlaps with #3391 (apologies to @P-E-P, but this should be a more general resolution), depends on #3220

I'm kinda glad you came up with another solution because mine felt wrong and I was even thinking about starting it from scratch again.

P-E-P (approved):

powerboat9:
@CohenArthur

CohenArthur (approved):

CohenArthur:
thanks! LGTM :)

PR #3422, opened 2025-02-07
Get rid of duplicate interface #3396

### LINKED ISSUE #3396: Get rid of duplicate interface
We need to remove this helper:

https://github.com/Rust-GCC/gccrs/blob/3c21102b7378b9a4d61c36689063f6104e751ff6/gcc/rust/typecheck/rust-typecheck-context.cc#L94

And instead we should be just calling:

https://github.com/Rust-GCC/gccrs/blob/3c21102b7378b9a4d61c36689063f6104e751ff6/gcc/rust/typecheck/rust-typecheck-context.cc#L101

and pass the hir id as part of the caller to make it more explicit that you need to think about what id you are using.

GS-GOAT:
- Updated single-parameter calls to pass an explicit HirId alongside the type.

- This refactor ensures that all calls to insert_implicit_type consistently use the two-parameter version, improving clarity and maintainability of type mapping tracking.

Fixes #3396

### DIFF
--- a/gcc/rust/typecheck/rust-autoderef.cc
+++ b/gcc/rust/typecheck/rust-autoderef.cc
@@ -113,7 +113,7 @@ Adjuster::try_unsize_type (TyTy::BaseType *ty)
   auto slice
     = new TyTy::SliceType (mappings.get_next_hir_id (), ty->get_ident ().locus,
 			   TyTy::TyVar (slice_elem->get_ref ()));
-  context->insert_implicit_type (slice);
+  context->insert_implicit_type (slice->get_ref (), slice);
 
   return Adjustment (Adjustment::AdjustmentType::UNSIZE, ty, slice);
 }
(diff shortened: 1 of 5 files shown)


powerboat9:
Does `git clang-format` work? It should auto format the code, so it'll pass the formatting checks. You may have to install clang-format, and copy [contrib/clang-format](https://github.com/Rust-GCC/gccrs/blob/master/contrib/clang-format) to `.clang-format` in the project root.

As for the changelog checks, you can run [this line](#L32) to enable `git gcc-commit-mklog`, which should make formatting commit messages easier. You can borrow other aliases from the same file (I'd also recommend `gcc-verify`) or instead run the entire file as a shell script.

Your commits will also have to be signed off on, as described [here](https://gcc.gnu.org/dco.html). You can do that automatically by passing `-s` to `git commit` or `git gcc-commit-mklog`.

powerboat9:
It's a good PR, you don't have to close it

PR #3430, opened 2025-02-12
NR2: Use parent scope instead of enum scope to resolve enum content

P-E-P:
```rust
struct E2 /* (1) */;

enum Test {
    E1,
    E2 /* (2) */ (E2 /* (3) */),
}
```

E2 (3) is resolved as E2 (2) instead of E2(1) because enum's scope has been pushed. In `Late::visit(TypePath &type)` we should map the usage to E2(1) definition.

For now I'm trying to find an elegant solution (resolve the `TypePath` from it's scope's parent) but this should only work for `Enum` (Reuse `ContextualVisitor` ?).

I'm not a fan of big forever stack modification, and obviously modifying it for that kind of edge case feels off.

- We could introduce a new `resolve_from_parent` instead of that extra boolean
- If we keep that extra argument I should probably change it to an explicit enum
- Should we introduce a function to resolve from anywhere instead of the node's parent ? Would it be used anywhere ?
- Should we look for parent module instead of parent scope ?

Should fix nr2 test `exhaustiveness2.rs`

### DIFF
--- a/gcc/rust/resolve/rust-forever-stack.h
+++ b/gcc/rust/resolve/rust-forever-stack.h
@@ -650,7 +650,8 @@ public:
    * @return a valid option with the Definition if the identifier is present in
    * the current map, an empty one otherwise.
    */
-  tl::optional<Rib::Definition> get (const Identifier &name);
+  tl::optional<Rib::Definition> get (const Identifier &name,
+				     bool from_parent = false);
 
   /**
    * Resolve a path to its definition in the current `ForeverStack`
@@ -663,7 +664,8 @@ public:
   template <typename S>
   tl::optional<Rib::Definition> resolve_path (
     const std::vector<S> &segments,
-    std::function<void (const S &, NodeId)> insert_segment_resolution);
+    std::function<void (const S &, NodeId)> insert_segment_resolution,
+    bool from_parent = false);
 
   // FIXME: Documentation
   tl::optional<Resolver::CanonicalPath> to_canonical_path (NodeId id) const;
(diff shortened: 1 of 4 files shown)


powerboat9:
It doesn't look like it'll be [quite that simple](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8321a6e21645c1881cca9d7569e0f550). I think skipping enum ribs when attempting to resolve the first segment of a path would do it.

P-E-P:
Ah exactly what I was looking for, a counter example.

CohenArthur:
I don't think there's any reason for `E2` to resolve to the enum variant in the first place, it should only be possible if the variant has been imported by a use statement unless I'm forgetting something. But I don't think you can ever reach `E2` without typing `Test::E2` so we are doing something wrong indeed. So I wonder if we couldn't fix this by not adding new definitions for enum variants or something of the sort, but it might just be easier to skip enum ribs like powerboat mentioned

powerboat9:
Can I try a fix?

P-E-P:
> Can I try a fix?

I've been on vacation for the past week hence the delayed answer/no activity on this PR, actually I'd like to tackle it myself.

powerboat9:
> > Can I try a fix?
> 
> I've been on vacation for the past week hence the delayed answer/no activity on this PR, actually I'd like to tackle it myself.

Oh, yeah, that's fine

P-E-P:
Closed in favor of  #3465

PR #3433, opened 2025-02-13
Add testing on powerpc to the github CI

powerboat9:
This should help solve #2184


powerboat9:
A file last changed before I was born is failing to compile because it uses implicit function declarations

thesamesam:
My own take (as someone who cares very much about GCC and Rust portability in the long-term) is that it's not worth spending time on this yet.

powerboat9:
> erm, darwin ppc (original report) is a native target without rustc support, and you're adding a cross target, still without rustc support, but it's even worse as we don't support cross compilation yet.
> 
> Do we really want to add this?

I was thinking this would be good for making sure we don't break big endian back ends, but if rustc doesn't support it, it would be impossible at this point to compile the rust libraries we use. Would probably be good to leave this unmerged for now, although the CI logs on this PR might help us troubleshoot anyways.

dkm:
> > erm, darwin ppc (original report) is a native target without rustc support, and you're adding a cross target, still without rustc support, but it's even worse as we don't support cross compilation yet.
> > Do we really want to add this?
> 
> I was thinking this would be good for making sure we don't break big endian back ends, but if rustc doesn't support it, it would be impossible at this point to compile the rust libraries we use. Would probably be good to leave this unmerged for now, although the CI logs on this PR might help us troubleshoot anyways.

I'm not really sure what you're going to test, as I believe it won't even pass the configure step for gcc.

But if you're willing to add cross compilation, an easy way is to use crosstool-ng (we use it in compiler explorer to build the various cross gcc). But it won't solve the fact that we only support native compiler with existing rustc support :shrug: ç

powerboat9:
> I'm not really sure what you're going to test, as I believe it won't even pass the configure step for gcc.
The configure script succeeds -- I don't think the compile would succeed even without the newlib issue, but its worth a shot.
> But if you're willing to add cross compilation, an easy way is to use crosstool-ng (we use it in compiler explorer to build the various cross gcc). But it won't solve the fact that we only support native compiler with existing rustc support 🤷 ç
I'll look into it, might solve the newlib issue.

The big issue we'd have with compiling for a target `rustc` doesn't support would be the rust runtime libraries we build with cargo, right?

barracuda156:
> erm, darwin ppc (original report) is a native target without rustc support

(Strictly speaking, it is the other way round: Rust is without support for Darwin ppc.)

But any way of testing on a Big-endian PowerPC will be helpful. There is nothing to CI on Darwin at the moment, of course, since Rust lacks support for the ABI.

gccrs from earlier gcc did build and mostly pass tests on Darwin, but rust just doesn’t exist.
That probably won’t change until gcc codegen, without llvm dependency, will be fully supported.

Once gcc backend is supported, then it might work via mrustc to bootstrap rust.

barracuda156:
Re gcc and rust on PowerPC, see also: https://github.com/rust-lang/rfcs/issues/1312

powerboat9:
I think the plan is to do something like https://rust-gcc.github.io/2024/09/20/reusing-rustc-components.html#how-do-we-use-them -- compile with rust components disabled, then recompile with them enabled.

Come to think of it, we could probably even do some shenanigans like

1. Linking essential object files into `crab1-basic`
2. Producing the rust object files using `crab1-basic`
3. Linking all object files into `crab1` for the final output

which would only work when building a native compiler, of course

powerboat9:
Our priority list for selecting a rust compiler would look something like

1. `gccrs`
2. `crab1-basic` (only when building the first stage of a native compiler)
3. `rustc`

The stated conditions placed on option 2 would be partially redundant, as after stage 1 of a native compiler build we would have `gccrs` available and option 1 supersedes option 2

barracuda156:
@iains Iain, tagging you re `gccrs`.

powerboat9:
Keep in mind, everything past the post I linked is just me brainstorming. It hasn't been decided yet or anything

powerboat9:
> A [file](https://sourceware.org/git?p=newlib-cygwin.git;a=history;f=libgloss/rs6000/sim-print.c;h=c0b9adced5eef6e3d87617c9376a35c0847f0475;hb=5e5e51f1dc56a99eb4648c28e00d73b6ea44a8b0) last modified before I was born is failing to compile because it intentionally uses implicit function declarations

It looks like this was fixed in https://sourceware.org/git/?p=newlib-cygwin.git;a=commit;h=467b85c58f815b0695c32ff9cb7b8da07d1dfb24 -- if we get compiling working without rustc again before the ~end-of-year release of newlib 4.6, we may have to manually apply it.

Issue #3315, opened 2024-12-21
Modules with same names as built in types causes ICE


matthewjasper:
### Code

```rust
mod i32 {}
```

### Meta

Version: gccrs (Compiler-Explorer-Build-gcc-b5c354d038f800695a8d730c56c4a4f744134adb-binutils-2.42) 14.0.1 20240309 (experimental)

### Error output

```
crab1: internal compiler error: in setup_builtin_types, at rust/resolve/rust-late-name-resolver-2.0.cc:97
0x27eb5ec internal_error(char const*, ...)
	???:0
0xaf6cc9 fancy_abort(char const*, int, char const*)
	???:0
0xdbb522 Rust::Resolver2_0::Late::go(Rust::AST::Crate&)
	???:0
0xc4bef0 Rust::Session::compile_crate(char const*)
	???:0
Please submit a full bug report, with preprocessed source (by using -freport-bug).
Please include the complete backtrace with any bug report.
See <https://gcc.gnu.org/bugs/> for instructions.
Compiler returned: 1
```

liamnaddell:
From reading here: https://doc.rust-lang.org/reference/names/namespaces.html

builtins and modules both belong in the Types namespace, however, that can't really be correct, since `rustc` allows you to name a module `i32`. I guess the idea is that builtins are resolved separately from other types? Otherwise, how would you prevent NR conflicts here?

liamnaddell:
As a note, `rustc` does NOT let you write the following:

```rust
liam@newgame ~/projects/gccrs/gccrs-build $ cat main.rs
pub mod i32 {
    pub fn a() {}
}

struct i32 {}

fn main() {
    i32::a();
}
liam@newgame ~/projects/gccrs/gccrs-build $ rustc main.rs
error[E0428]: the name `i32` is defined multiple times
 --> main.rs:5:1
  |
1 | pub mod i32 {
  | ----------- previous definition of the module `i32` here
...
5 | struct i32 {}
  | ^^^^^^^^^^ `i32` redefined here
  |
  = note: `i32` must be defined only once in the type namespace of this module
```

liamnaddell:
I guess what we are expected to do is remove builtin types from the resolution context when another `Item` overshadows it

liamnaddell:
Modules are clearly handled separately from other types like structs. I think modules are somehow part of their own namespace?

powerboat9:
[...]

PR #3437, opened 2025-02-14
Fix modules with same name as builtins causing ICE (#3315)

liamnaddell:
gcc/rust/ChangeLog:

	* resolve/rust-forever-stack.h: Add a dedicated prelude node for the Language prelude
	* resolve/rust-forever-stack.hxx: Add support code for the prelude node
	* resolve/rust-late-name-resolver-2.0.cc (Late::visit): Move language prelude builtins to the prelude context
	* resolve/rust-name-resolution-context.cc: Add code for handling the prelude corner case
	* resolve/rust-rib.h: Add a special Prelude rib type

gcc/testsuite/ChangeLog:

	* rust/compile/issue-3315-1.rs: Add test for module with same name as builtin
	* rust/compile/issue-3315-2.rs: Test with utilization of i32 type
	* rust/compile/nr2/exclude: issue-3315-2.rs Does not work with NR2.0

in the git commit message.

- \[ ] Run `clang-format`

order to get feedback.
---


gcc/rust/resolve/rust-late-name-resolver-2.0.cc
```diff
@@ -91,16 +91,22 @@ Late::setup_builtin_types ()
...
+  // There's a special Rib for putting prelude items, since prelude items need
+  // to satisfy certain special rules.
+  ctx.scoped (Rib::Kind::Prelude, 0, [this, &ty_ctx] (void) -> void {
+    for (const auto &builtin : builtins)
+      {
+	// we should be able to use `insert_at_root` or `insert` here, since
+	// we're at the root :) hopefully! Globbed declaration allows
+	// user-defined items in the `types` namespace
+	//  to shadow builtins. Probably unwise, but allowed by rustc!
+	auto ok = ctx.types.insert_globbed (builtin.name, builtin.node_id);
```
powerboat9, line 103: Since it's in a separate rib, `insert`, `insert_shadowable`, or `insert_globbed` would all work fine here

gcc/rust/resolve/rust-name-resolution-context.cc
```diff
@@ -107,6 +107,11 @@ NameResolutionContext::scoped (Rib::Kind rib_kind, NodeId id,
 			       std::function<void (void)> lambda,
 			       tl::optional<Identifier> path)
 {
+  // Prelude doesn't have an access path
+  // TODO: Assert current cursor is at the root
```
powerboat9, line 111: This should already hit an assertion, so you could remove the comment or add another assertion

gcc/rust/resolve/rust-name-resolution-context.cc
```diff
@@ -107,6 +107,11 @@ NameResolutionContext::scoped (Rib::Kind rib_kind, NodeId id,
...
+  if (rib_kind == Rib::Kind::Prelude)
+    rust_assert (!path);
```
powerboat9, line 113: This could be moved to `push_inner` or copied there, but an assertion like this should probably be in `push_inner`

powerboat9:
> @powerboat9 I can update to move the language prelude to rust-forever-stack.hxx if you still want the change

It's alright, it looks good

liamnaddell:
@powerboat9 Thank you for the reviews, and sorry about the TODO: comments -- I forgot about those. 

I think I fixed all of the issues you pointed out.

powerboat9 (approved):

philberty (commented):
Looks like a regression for:

Running /home/runner/work/gccrs/gccrs/gcc/testsuite/rust/borrowck/borrowck.exp ...
Running /home/runner/work/gccrs/gccrs/gcc/testsuite/rust/compile/compile.exp ...
Running /home/runner/work/gccrs/gccrs/gcc/testsuite/rust/compile/macros/builtin/builtin_macro.exp ...
Running /home/runner/work/gccrs/gccrs/gcc/testsuite/rust/compile/macros/mbe/mbe_macro.exp ...
Running /home/runner/work/gccrs/gccrs/gcc/testsuite/rust/compile/macros/proc/proc_macro.exp ...
Running /home/runner/work/gccrs/gccrs/gcc/testsuite/rust/compile/nr2/compile.exp ...
FAIL: /home/runner/work/gccrs/gccrs/gcc/testsuite/rust/compile/macros/mbe/macro54.rs: nr2 failure: FAIL: rust/compile/macros/mbe/macro54.rs -frust-name-resolution-2.0 (test for excess errors)

liamnaddell:
> Running /home/runner/work/gccrs/gccrs/gcc/testsuite/rust/compile/macros/mbe/mbe_macro.exp ...
> Running /home/runner/work/gccrs/gccrs/gcc/testsuite/rust/compile/macros/proc/proc_macro.exp ...
> Running /home/runner/work/gccrs/gccrs/gcc/testsuite/rust/compile/nr2/compile.exp ...
> FAIL: /home/runner/work/gccrs/gccrs/gcc/testsuite/rust/compile/macros/mbe/macro54.rs: nr2 failure: FAIL: rust/compile/macros/mbe/macro54.rs -frust-name-resolution-2.0 (test for excess errors)

For some reason after rebasing these fail now. I'll take a look at these

liamnaddell:
I think the strategy I chose to implement this doesn't work, since the test involves resolving something like u32::VALUE.

I'll try to re-implement

liamnaddell:
@powerboat9 @philberty I added some code to get_segments to handle the prelude rib as a special case. I think my PR should work now

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -63,6 +70,17 @@ template <Namespace N>
 void
 ForeverStack<N>::push_inner (Rib rib, Link link)
 {
+  if (rib.kind == Rib::Kind::Prelude)
+    {
+      // If you push_inner into the prelude from outside the root, you will pop
+      // back into the
+      //  root, which could screw up a traversal.
```
powerboat9, line 77: Should probably adjust the formatting on this comment

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -487,6 +519,7 @@ ForeverStack<N>::resolve_segments (
        *
```
powerboat9, line 519: Should probably add an entry to the numbered list in the above comment, mentioning changing `searched_prelude` from `false` to `true`

gcc/rust/resolve/rust-forever-stack.h
```diff
@@ -750,7 +754,15 @@ template <Namespace N> class ForeverStack
...
   void update_cursor (Node &new_cursor);
 
+  /* The forever stack's actual nodes */
   Node root;
+  /*
+   * A special prelude node used currently for resolving language builtins
+   * It has the root node as a parent, and acts as a "special case" for name
+   * resolution
+   */
+  Node prelude;
```
powerboat9, line 764: I probably should have noticed and commented on this earlier, but this should really be called something like `language_prelude`, given that we'll have to deal with multiple preludes

gcc/rust/resolve/rust-name-resolution-context.cc
```diff
@@ -107,6 +107,7 @@ NameResolutionContext::scoped (Rib::Kind rib_kind, NodeId id,
 			       std::function<void (void)> lambda,
 			       tl::optional<Identifier> path)
 {
+  // NOTE: You should be at the root node when calling this function.
```
powerboat9, line 110: You should be at the root node, if pushing a `Prelude` rib kind

powerboat9 (commented):
Otherwise, looks great

powerboat9:
It looks like rustc has some [`backtracking`](https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_resolve/src/lib.rs#L1612) capability. We may have to implement something like that later, for full parity, but that's probably outside the scope of this PR

gcc/rust/resolve/rust-forever-stack.h
liamnaddell: I considered that as well, I figured that we would also dump the std::prelude and core::prelude in here as well

gcc/rust/resolve/rust-name-resolution-context.cc
liamnaddell: :facepalm:

liamnaddell: I'll update to clarify that we only need to be at the root for the *prelude* rib pushes

gcc/rust/resolve/rust-forever-stack.h
powerboat9: I figured we'd keep some sort of reference to the standard library prelude -- since that should already be a module with its own `Node`, we shouldn't need to try to copy entries into another `Node`

gcc/rust/resolve/rust-forever-stack.hxx
liamnaddell: fixed

gcc/rust/resolve/rust-forever-stack.hxx
liamnaddell: I added a comment for this, let me know if it needs modification.

powerboat9 (approved):

gcc/rust/resolve/rust-forever-stack.h
liamnaddell: I'll add a comment to clarify this

dkm:
Small note, your changelog is not correct. It should not only give the file name, but also the "item" it is modifying. Usually, a function, but can be a type, or something else.

```
   * file.c (function_name): Adjusted call to bla
```

liamnaddell:
> Small note, your changelog is not correct. It should not only give the file name, but also the "item" it is modifying. Usually, a function, but can be a type, or something else.
> 
> ```
>    * file.c (function_name): Adjusted call to bla
> ```



> Small note, your changelog is not correct. It should not only give the file name, but also the "item" it is modifying. Usually, a function, but can be a type, or something else.

@dkm Is the log better now?

P-E-P (approved):
I'm unsure this is the correct approach, but I can't find any satisfying way to solve this problem for now so let's keep it as is. Code-wise I've got nothing to say, LGTM.

PR #3438, opened 2025-02-14
Desugar error propagation

CohenArthur:
Draft for now as there are still issues late in the pipeline when using the desugared code. I think it has to do with lang items.

### DIFF
--- a/gcc/testsuite/rust/compile/try-expr1.rs
+++ b/gcc/testsuite/rust/compile/try-expr1.rs
@@ -1,0 +1,82 @@
+#[lang = "sized"]
+trait Sized {}
+
+enum Result {
+    #[lang = "Ok"]
+    Ok(i32),
+    #[lang = "Err"]
+    Err(i32)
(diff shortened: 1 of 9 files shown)


gcc/testsuite/rust/compile/try-expr1.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// Wrap an OK value to construct the composite result. For example,
+    /// `Result::Ok(x)` and `Result::from_ok(x)` are equivalent.
+    #[lang = "from_ok"]
+    #[unstable(feature = "try_trait", issue = "42327")]
+    fn from_ok(v: i32) -> Self;
+}
+
+impl Try for Result {
+    // type Ok = i32;
+    // type Error = i32;
```
CohenArthur, line 61: hmmm the test should probably use the actual trait especially if we are stopping after lowering

P-E-P (approved):

PR #3443, opened 2025-02-18
nr2.0: Fix issue with external crates

powerboat9:
When visiting an external crate declaration, handle failed crate name lookups. This can happen when `Session::load_extern_crate` fails to load a crate during the `CfgStrip` phase.


philberty (approved):
LGTM

PR #3457, opened 2025-02-25
expansion: Correctly expand `$crate` metavar

### LINKED ISSUE #2833: Support `println!()` macro

CohenArthur:
This is better than what we currently do but it is missing a test for checking that we can properly call macros from external crates. I might add it later but it's a bit of a pain. Fixes #2833


P-E-P (approved):
LGTM

PR #3458, opened 2025-02-26
Fix Issue 3351: Shared discriminator having the same value<

### LINKED ISSUE #3351: Shared discriminator does not cause compiler error
The following code should generate a compiler error stating how the same discriminant value is assigned multiple times.
```rust
enum Foo {
    Bar = 1,
    Qaz = 1,
}

fn main() {}
```
But it seems to generate no errors on gccrs.
Godbolt: https://godbolt.org/z/f1aY19v4q

badumbatish:
Fixes #3351

Overriding on the compile item visitor, I am not sure if this is the right way to go?

### DIFF
--- a/gcc/rust/backend/rust-compile-item.cc
+++ b/gcc/rust/backend/rust-compile-item.cc
@@ -270,6 +270,16 @@ CompileItem::visit (HIR::Function &function)
 }
 
 void
+CompileItem::visit (HIR::Enum &e)
+{
+  TyTy::BaseType *resolved_type = nullptr;
+  bool ok = ctx->get_tyctx ()->lookup_type (e.get_mappings ().get_hirid (),
+					    &resolved_type);
+  rust_assert (ok);
+  tree type = TyTyResolveCompile::compile (ctx, resolved_type);
+  reference = type;
+}
+void
 CompileItem::visit (HIR::ImplBlock &impl_block)
 {
   TyTy::BaseType *self_lookup = nullptr;
--- a/gcc/rust/backend/rust-compile-type.cc
+++ b/gcc/rust/backend/rust-compile-type.cc
@@ -20,10 +20,10 @@
 #include "rust-compile-expr.h"
 #include "rust-constexpr.h"
 #include "rust-gcc.h"
+#include "rust-diagnostics.h"
 
 #include "tree.h"
 #include "stor-layout.h"
-
 namespace Rust {
 namespace Compile {
 
@@ -248,6 +248,65 @@ TyTyResolveCompile::visit (const TyTy::FnPtr &type)
 					   type.get_ident ().locus);
 }
 
+bool
+check_variant_record_collision (Context *ctx, const TyTy::ADTType &type,
+				std::vector<tree> &variant_records)
+{
+  // bdbt: we're checking if shared discriminants crash with each other or
+  // not. lets make a map from uhwi to hir id. A clash of uhwi in a variant
+  // record to which said record can be converted uhwi is indicative of
+  // issue 3351 of gccrs
+
+  std::map<HOST_WIDE_INT, std::vector<size_t>> shwi_to_index;
+  for (size_t i = 0; i < variant_records.size (); i++)
+    {
+      TyTy::VariantDef *variant = type.get_variants ().at (i);
+      if (variant->has_discriminant ())
+	{
+	  ctx->push_const_context ();
+	  tree discriminant_expr
+	    = CompileExpr::Compile (variant->get_discriminant (), ctx);
+	  ctx->push_const_context ();
+	  tree folded_expr = fold_expr (discriminant_expr);
+	  if (folded_expr == error_mark_node)
+	    {
+	      // if we have discriminant but we fail to fold it, return false
+	      return false;
+	    }
+	  HOST_WIDE_INT discriminant_integer = tree_to_shwi (folded_expr);
+	  shwi_to_index[discriminant_integer].push_back (i);
+	}
+    }
+
+  bool has_failed = false;
+  for (const auto &map_item : shwi_to_index)
+    {
+      auto discriminant_integer = map_item.first;
+      const auto &index_vector = map_item.second;
+      // collision doesn't happen, move to next item
+      if (index_vector.size () <= 1)
+	continue;
+
+      has_failed = true;
+      rich_location r (line_table, type.get_locus ());
+      std::string assigned_here_msg
+	= expand_message (HOST_WIDE_INT_PRINT_DEC " assigned here",
+			  discriminant_integer);
+      std::string assigned_more_once_msg
+	= expand_message ("discriminant value " HOST_WIDE_INT_PRINT_DEC
+			  " assigned more than once",
+			  discriminant_integer);
+      for (auto index : index_vector)
+	{
+	  TyTy::VariantDef *variant = type.get_variants ().at (index);
+	  r.add_fixit_replace (variant->get_discriminant ().get_locus (),
+			       assigned_here_msg.c_str ());
+	}
+      rust_error_at (r, ErrorCode::E0081, "%s",
+		     assigned_more_once_msg.c_str ());
+    }
+  return !has_failed;
+}
 void
 TyTyResolveCompile::visit (const TyTy::ADTType &type)
 {
@@ -255,7 +314,6 @@ TyTyResolveCompile::visit (const TyTy::ADTType &type)
   if (!type.is_enum ())
     {
       rust_assert (type.number_of_variants () == 1);
-
       TyTy::VariantDef &variant = *type.get_variants ().at (0);
       std::vector<Backend::typed_identifier> fields;
       for (size_t i = 0; i < variant.num_fields (); i++)
@@ -358,9 +416,15 @@ TyTyResolveCompile::visit (const TyTy::ADTType &type)
 	  // add them to the list
 	  variant_records.push_back (named_variant_record);
 	}
-
-      // now we need to make the actual union, but first we need to make
-      // named_type TYPE_DECL's out of the variants
+      //  TODO: bdbt set up defid and a map (or set?) to check if we have
+      // checked for collision already.
+      if (!check_variant_record_collision (ctx, type, variant_records))
+	{
+	  translated = error_mark_node;
+	  return;
+	}
+      // the actual union, but first we need to make named_type TYPE_DECL's out
+      // of the variants
 
       size_t i = 0;
       std::vector<Backend::typed_identifier> enum_fields;
--- a/gcc/rust/rust-diagnostics.h
+++ b/gcc/rust/rust-diagnostics.h
@@ -80,6 +80,9 @@ enum class ErrorCode : unsigned int
 #undef ERROR
 
 // clang-format off
+std::string
+expand_message (const char *fmt, ...) RUST_ATTRIBUTE_GCC_DIAG (1, 2);
+
 extern void
 rust_internal_error_at (const location_t, const char *fmt, ...)
   RUST_ATTRIBUTE_GCC_DIAG (2, 3)
--- a/gcc/rust/backend/rust-compile-item.h
+++ b/gcc/rust/backend/rust-compile-item.h
@@ -44,7 +44,7 @@ public:
   void visit (HIR::ImplBlock &impl_block) override;
   void visit (HIR::ExternBlock &extern_block) override;
   void visit (HIR::Module &module) override;
-
+  void visit (HIR::Enum &) override;
   // Empty visit for unused Stmt HIR nodes.
   void visit (HIR::TupleStruct &) override {}
   void visit (HIR::EnumItem &) override {}
@@ -58,7 +58,6 @@ public:
   void visit (HIR::UseDeclaration &) override {}
   void visit (HIR::TypeAlias &) override {}
   void visit (HIR::StructStruct &) override {}
-  void visit (HIR::Enum &) override {}
   void visit (HIR::Union &) override {}
   void visit (HIR::Trait &) override {}
   void visit (HIR::EmptyStmt &) override {}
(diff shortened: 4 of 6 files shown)


badumbatish:
redundant header include, i have noticed, will fix after review

gcc/rust/backend/rust-compile-type.cc
```diff
@@ -17,12 +17,14 @@
...
 
 #include "rust-compile-type.h"
+#include "rust-compile-context.h"
 #include "rust-compile-expr.h"
 #include "rust-constexpr.h"
 #include "rust-gcc.h"
 
 #include "tree.h"
 #include "stor-layout.h"
+#include <cstdio>
```
P-E-P, line 27: Use `rust-system.h` instead

gcc/rust/backend/rust-compile-type.cc
```diff
@@ -239,14 +241,54 @@ TyTyResolveCompile::visit (const TyTy::FnPtr &type)
...
+	{
+	  tree discriminant_expr
+	    = CompileExpr::Compile (variant->get_discriminant (), ctx);
+	  HOST_WIDE_INT discriminant_integer
+	    = tree_to_shwi (fold_expr (discriminant_expr));
+	  shwi_to_index[discriminant_integer].push_back (i);
+	}
+    }
+
+  for (auto &[discriminant_integer, index_vector] : shwi_to_index)
```
philberty, line 268: iirc this is a c++17 or c++20 thing not sure we can use this style of for loop over a map? @CohenArthur or @P-E-P ?

gcc/rust/backend/rust-compile-type.cc
```diff
@@ -239,14 +241,54 @@ TyTyResolveCompile::visit (const TyTy::FnPtr &type)
...
+      if (index_vector.size () >= 2)
+	{
+	  rust_error_at (type.get_locus (), ErrorCode::E0081,
+			 "discriminant value `%d` assigned more than once",
+			 discriminant_integer);
+	  for (auto index : index_vector)
+	    {
+	      TyTy::VariantDef *variant = type.get_variants ().at (index);
+	      rust_error_at (variant->get_discriminant ().get_locus (),
+			     "`%d` assigned here", discriminant_integer);
```
philberty, line 280: instead of using quotes you need to use "%<%d%> assigned here"

Actually you can improve this by using a rich location such as:

```c++
      rich_location r (line_table, type.get_locus ());
      for (const auto& index : index_vector)
      {
         TyTy::VariantDef *variant = type.get_variants ().at (index);
         r.add_fixit_replace (variant->get_discriminant ().get_locus (), "%<%d%> here", discriminant_integer);
      }
      rust_error_at (r, ErrorCode::E0018, "discriminant value %<%d%> assigned more than once", discriminant_integer);
```

gcc/rust/backend/rust-compile-type.cc
```diff
@@ -239,14 +241,54 @@ TyTyResolveCompile::visit (const TyTy::FnPtr &type)
...
+				std::vector<tree> &variant_records)
+{
+  // bdbt: we're checking if shared discriminants crash with each other or
+  // not. lets make a map from uhwi to hir id. A clash of uhwi in a variant
+  // record to which said record can be converted uhwi is indicative of
+  // issue 3351 of gccrs
+
+  size_t i = 0;
+  std::map<HOST_WIDE_INT, std::vector<size_t>> shwi_to_index;
+  for (; i < variant_records.size (); i++)
```
philberty, line 255: i prefer to make sure we write the for loop as for (size_t i = 0; ... )

gcc/rust/backend/rust-compile-type.cc
```diff
@@ -239,14 +241,54 @@ TyTyResolveCompile::visit (const TyTy::FnPtr &type)
...
+    {
+      TyTy::VariantDef *variant = type.get_variants ().at (i);
+      if (variant->has_discriminant ())
+	{
+	  tree discriminant_expr
+	    = CompileExpr::Compile (variant->get_discriminant (), ctx);
+	  HOST_WIDE_INT discriminant_integer
+	    = tree_to_shwi (fold_expr (discriminant_expr));
```
philberty, line 263: You will need to add a check here from fold_expr to check for error_mark_node. Or you need to check how tree_to_shwi handles an error_mark_node

gcc/rust/backend/rust-compile-item.cc
```diff
@@ -265,6 +265,17 @@ CompileItem::visit (HIR::Function &function)
     ctx->pop_const_context ();
 }
 
+void
+CompileItem::visit (HIR::Enum &e)
```
philberty, line 269: do you need this? This will fail to compile for generic enums so we just let the query system handle this stuff.

what happens when you remove this?

philberty (changes requested):
Fantastic work i've left you comments for style. The error message using %< etc will fix the warnings that are introduced.

I don't think you should need the changes to rust-compile-item.cc but left me know if removing it causes an issue for you

I think in theory we might need a guard in your code where check_variant_record_collision needs to check if this has been checked before because you will get duplicate warnings here for example:

```
#[lang="sized"]
trait Sized {}

enum Greeting<T> {
    Hi(T),
    Bye
}

fn test() {
    let a = Greeting::<i32>::Hi(123);
    let b = Greeting::<f32>::Bye;
}
```

So I think lets get this first part merged. Then you will need to start tracking the HIR::Enum def-id on the TyTy::ADTType and you can keep a map in the backend if this DefID has already been checked to solve the above problem.

gcc/rust/backend/rust-compile-type.cc
```diff
@@ -239,14 +241,54 @@ TyTyResolveCompile::visit (const TyTy::FnPtr &type)
...
+	  shwi_to_index[discriminant_integer].push_back (i);
+	}
+    }
+
+  for (auto &[discriminant_integer, index_vector] : shwi_to_index)
+    {
+      // collision happens
+      if (index_vector.size () >= 2)
```
philberty, line 271: i would probably write this as:

```
bool collision = iindex_vector.size () > 1;
if (!collision)
   continue;
```

It will flatten the code make it bit neater.

gcc/rust/backend/rust-compile-type.cc
philberty, line 27: we cant include cstdio it should be part of rust-system.h but I don't think your using any of the stuff from that header so you can remove this

gcc/rust/backend/rust-compile-type.cc
P-E-P: yeah structured binding are cpp17

gcc/rust/backend/rust-compile-type.cc
P-E-P: Actually IIRC when you don't know the actual value and intend to quote a variable you can use `%qd` or `%qs` which is shorter.

gcc/rust/backend/rust-compile-type.cc
badumbatish: ill change this down to cpp14

gcc/rust/backend/rust-compile-type.cc
badumbatish: resolved with both your recommendation

gcc/rust/backend/rust-compile-type.cc
badumbatish: fixed, this is from copying code from the caller function

gcc/rust/backend/rust-compile-type.cc
badumbatish: i'll ICE this out

gcc/rust/backend/rust-compile-item.cc
badumbatish: it's a bit unclear to me on how to use the query system.... when i remove this basically there would be no checks into the enum

gcc/rust/backend/rust-compile-type.cc
```diff
@@ -239,14 +239,64 @@ TyTyResolveCompile::visit (const TyTy::FnPtr &type)
...
+      std::string assigned_more_once_msg
+	= "discriminant value `" + std::to_string (discriminant_integer) + "`"
+	  + " assigned more than once";
+      for (auto index : index_vector)
+	{
+	  TyTy::VariantDef *variant = type.get_variants ().at (index);
+	  r.add_fixit_replace (variant->get_discriminant ().get_locus (),
+			       assigned_here_msg.c_str ());
+	}
+      rust_error_at (r, ErrorCode::E0081, assigned_more_once_msg.c_str ());
```
philberty, line 290: you will need to make this

rust_error_at (r, ErrorCode::E0081, "%s", assigned_more_once_msg.c_str ())

That will fix the last warning for you.

badumbatish:
> LGTM nice work

thank you for your guidance!

gcc/rust/backend/rust-compile-type.cc
```diff
@@ -239,14 +239,65 @@ TyTyResolveCompile::visit (const TyTy::FnPtr &type)
...
+    {
+      auto discriminant_integer = map_item.first;
+      const auto &index_vector = map_item.second;
+      // collision doesn't happen, move to next item
+      if (index_vector.size () <= 1)
+	continue;
+
+      rich_location r (line_table, type.get_locus ());
+      std::string assigned_here_msg
+	= "`" + std::to_string (discriminant_integer) + "`" + " assigned here";
```
P-E-P, line 280: There are more backquote here, GCC probably does't complain about those because they are not in the format string. I'm unsure if we should change them.

P-E-P (approved):

gcc/rust/backend/rust-compile-type.cc
badumbatish: I can change it to %qlld via format string i think? The reason i did it this way is because different platform has different width for HOST WIDTH INT and im not sure if the final answer would be %qlld or %qld or if there is any complication if HWI is only %d and we use %qlld

P-E-P: Yeah, I'm kinda torn on the subject. Raw back quotes don't play well with different locales, this is not only  a matter of style.

powerboat9: Would `HOST_WIDE_INT_PRINT_DEC` work?

badumbatish: i think it could work for rust_error_at, 

for `add_fixit_replace()` they dont take formatting arguments, is there a way to create a standalone formatting string in the codebase without  explicitly using snprintf?

powerboat9: I think it's probably fine as-is, unless someone can come up with something else more concrete

powerboat9: Maybe use normal quotation marks (`"`) instead of ``` ` ``` and include a `TODO` comment, so we're more likely to notice and fix this in the future

powerboat9: Actually, wait, we have `rust_open_quote` and `rust_close_quote` -- that should do the trick

gcc/rust/backend/rust-compile-type.cc
```diff
@@ -239,14 +239,65 @@ TyTyResolveCompile::visit (const TyTy::FnPtr &type)
...
+      TyTy::VariantDef *variant = type.get_variants ().at (i);
+      if (variant->has_discriminant ())
+	{
+	  tree discriminant_expr
+	    = CompileExpr::Compile (variant->get_discriminant (), ctx);
+	  tree folded_expr = fold_expr (discriminant_expr);
+	  if (folded_expr == error_mark_node)
+	    {
+	      // if we have discriminant but we fail to fold it, ICE
+	      rust_unreachable ();
```
philberty, line 263: This will be fine for now but will need to remove this. Ideally we need to have this function return boolean ok = true else false. So in this case it should just return false. An error will have been emitted already and in the caller it can just fail with returning error_mark_node early too as it cannot compile the enum then.

philberty (approved):

gcc/rust/backend/rust-compile-type.cc
badumbatish: > Oh, wait, check out `expand_message` in `rust-diagnostics.cc` -- that should be what you're looking for

the function is only in rust-diagnostics.cc, im not sure if i move it to the header to use it solely for rust-compile-type.cc?

gcc/rust/backend/rust-compile-type.cc
badumbatish: fixed with latest push

gcc/rust/backend/rust-compile-type.cc
powerboat9: You could make it not-`static` and add a declaration to `rust-diagnostics.h`

philberty:
Looks like you just need to fix that clang format lint

badumbatish:
> Looks like you just need to fix that clang format lint

sorry, will fix this after my spring break

gcc/rust/rust-diagnostics.h
```diff
@@ -31,7 +31,7 @@
 #if __GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 1)
 #define RUST_ATTRIBUTE_GCC_DIAG(m, n)                                          \
   __attribute__ ((__format__ (__gcc_tdiag__, m, n)))                           \
-    __attribute__ ((__nonnull__ (m)))
+__attribute__ ((__nonnull__ (m)))
```
powerboat9, line 34: This reformatting is a bit odd -- might be worth not including if `clang-format` lets you get away with it

philberty:
just need this formatting fixed and we can merge

badumbatish:
apologies,  im back

powerboat9:
> apologies, im back

No problem

badumbatish:
close this due to staleness and lack of capacity, please feel free to reimplement these changes in your own PRs.

PR #3465, opened 2025-03-04
Fix nr2 exhaustiveness

P-E-P:


gcc/rust/resolve/rust-rib.cc
```diff
@@ -22,7 +22,8 @@
 namespace Rust {
 namespace Resolver2_0 {
 
-Rib::Definition::Definition (NodeId id, Mode mode)
+Rib::Definition::Definition (NodeId id, Mode mode, bool enum_variant)
```
powerboat9, line 25: Definitions shouldn't need an `enum_variant` property, since all items in an enum-kind rib (and only those items) should be enum variants

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -265,34 +273,37 @@ ForeverStack<N>::update_cursor (Node &new_cursor)
...
-      // if there was no candidate, we keep iterating
-      KeepGoing::Yes);
-  });
+  reverse_iter (
+    [&resolved_definition, &name, skip_enum_variant] (Node &current) {
+      auto candidate = current.rib.get (name.as_string ());
+
+      return candidate.map_or (
+	[&resolved_definition, skip_enum_variant] (Rib::Definition found) {
+	  if (skip_enum_variant && found.is_variant ())
```
powerboat9, line 287: This can be moved out of the `map_or`, and check the kind of the current node's rib instead

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -265,34 +273,37 @@ ForeverStack<N>::update_cursor (Node &new_cursor)
 
 template <Namespace N>
 tl::optional<Rib::Definition>
-ForeverStack<N>::get (const Identifier &name)
+ForeverStack<N>::get (const Identifier &name, bool skip_enum_variant)
```
powerboat9, line 276: Because `get` only checks the current scope and its parents, we should always skip enum variants

powerboat9:
A similar tweak as done to `get` could be made to `resolve_segments`, and perhaps should be to prevent weird edge cases, but realistically I don't think it could ever cause any issues.

gcc/rust/resolve/rust-rib.cc
P-E-P: I think we do need them because of generic types:
```
enum Test<G> {
// G should be available to either make a new variant of any items inside the variant definition
}
```

gcc/rust/resolve/rust-forever-stack.hxx
P-E-P: This is  related to your previous comment, right ? Correct me if I'm wrong but with my understanding of  generics we can't do that.

gcc/rust/resolve/rust-forever-stack.hxx
P-E-P: Yeah that seems right. Will fix.

gcc/rust/resolve/rust-rib.cc
powerboat9: Dang, forgot about that. Maybe two nodes/ribs, like this?

```text
Node (Kind::Enum)
| - Node (Kind::EnumInner)
    | - G
| - Other
| - Another
```

gcc/rust/resolve/rust-forever-stack.hxx
P-E-P: Fixed, thank you!

gcc/rust/resolve/rust-rib.cc
P-E-P: I'm not thrilled about this because it seems to change quite fundamentally the behavior of the foreverstack.

CohenArthur (approved):
as I said to PE privately, while this may not be a perfect solution it gets us further and makes us move forward so I'm happy merging it. we can think about how to improve it by looking at what rustc does in this situation and trying to figure out an even better way to do this, but the priority is about reducing the gap between nr1.0 and nr2.0 which this PR achieves. so LGTM

gcc/rust/resolve/rust-rib.cc
powerboat9: I should probably adjust the diagram a bit for clarity. The node structure would be like this

```
Node (enum item definitions are in scope)
| - Node (generics are in scope) <--- current node
```

Would also block paths to generics, like `Test::G`, from working

P-E-P: Unless I'm mistaken this diagram does not change how I interpret your proposal:
Enum variants have their own scope and a new anonymous rib dedicated to generics should be pushed within the enum among the variants (Correct me if I'm wrong).

I think there are two problems with this:
- We kinda have to "jump" over a scope in enums, this means we need to change the behavior of many things for what I consider an edge case. Once again correct me if I'm wrong but that behavior can only be found with enum variants ?
- Generics do not appear naturally at the same level as variants, this means that variant childs declaration won't be separated by their parent but should be put in the generic scope to gain access to those generics ?

```rust
enum Test<T : Default + Into<u32>> {
    A(T),
    B = {
         let b = 2;
         T::default::into() + b
    },
}
```

As I'm trying to write an example diagram I found that this does not work, I must have misunderstood something. Would you write the corresponding diagram to this situation for me ?

Issue #3395, opened 2025-01-29
Refactor this to use the implicit insert


philberty:
We are calling xxxx

Harishankar14:
Hey @philberty,can i work on this issue? // 
Refactor to use insert_implicit_type instead of insert_type for cleaner handling.
This change will simplify the code by passing the hirid directly without using NodeMapping.

mostafa630:
Hi @philberty , I'd like to work on this issue. Could you please assign it to me?

mostafa630:
@philberty Thanks for the opportunity! I just made a pull request [PR #3489], and all tests passed successfully.

PR #3467, opened 2025-03-05
Refactor implicit insert   issue #3395

mostafa630:
This solution addresses issue #3395 by refactoring the code to use "implicit_insert_type" instead of "insert_type" to make it cleaner.


P-E-P:
You should probably squash your commits and reword them to include the changelog. Take a look at this [section of the wiki page](https://github.com/Rust-GCC/gccrs/wiki/First-contributions#commits) for further informations.

philberty (changes requested):
Can you remove the insert_type method entirely from the context object as well?

mostafa630:
ok , I will do that

> Can you remove the insert_type method entirely from the context object as well?

P-E-P:
> ok , I will do that
> 
> > Can you remove the insert_type method entirely from the context object as well?

You do not need to close the PR each time, you can rebase and rewrite your commits instead.

mostafa630:
thank you ,  I started doing that in the latest pull request ,  [PR #3489] (https://github.com/Rust-GCC/gccrs/pull/3489)
i only force the push at each time.

> You do not need to close the PR each time, you can rebase and rewrite your commits instead.

PR #3469, opened 2025-03-05
Typecheck `else` expression of `let-else` to ensure it diverges

CohenArthur:
The `else` expression of a `let-else` must always diverge, which we enforce during typecheck. Needs #3468

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-stmt.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-stmt.cc
@@ -17,12 +17,14 @@
 // <http://www.gnu.org/licenses/>.
 
 #include "rust-hir-type-check-stmt.h"
+#include "rich-location.h"
 #include "rust-hir-type-check-type.h"
 #include "rust-hir-type-check-expr.h"
 #include "rust-hir-type-check-implitem.h"
 #include "rust-hir-type-check-item.h"
 #include "rust-hir-type-check-pattern.h"
 #include "rust-type-util.h"
+#include "rust-tyty.h"
 
 namespace Rust {
 namespace Resolver {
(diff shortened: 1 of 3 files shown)


gcc/rust/typecheck/rust-hir-type-check-stmt.cc
```diff
@@ -97,6 +100,22 @@ TypeCheckStmt::visit (HIR::LetStmt &stmt)
...
+
+      if (else_expr_ty->get_kind () != TyTy::TypeKind::NEVER)
+	{
+	  rust_error_at (else_expr.get_locus (),
+			 "%<else%> clause of let-else does not diverge");
+	  return;
+	}
+
+      // FIXME: Is that enough? Do we need to do something like
+      // `append_reference` here as well?
```
CohenArthur, line 116: @philberty I'm not sure if there's anything else we need to do here. I was trying to use something like `coercion_site(...)` but obviously it doesn't work because any type is coercible to `!`, which is not the behavior we actually want here

powerboat9: `!` should be coercible to any other type, but not the other way around

philberty: Ah ok yeah this is enough because all you need to do is check for ! in the else case so yes this is correct! Dont need to do unify or coercion site because it doesnt need to do anything else.

Dont think we support this in codegen

philberty (approved):
LGTM this is enough

powerboat9:
@CohenArthur still working on this? Looks ~good, just needs a bit of touching up

powerboat9:
(disabled automerge, so it doesn't go off unexpectedly)

gcc/rust/typecheck/rust-hir-type-check-stmt.cc
powerboat9: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=83c424a3b36b5c35edcf6d8995c97ee2

It looks like an `if` and `else` branch with the same types should also work.

PR #3490, opened 2025-03-06
nr2.0: Set the node id of the root node

powerboat9:


gcc/rust/resolve/rust-forever-stack.h
```diff
@@ -554,6 +554,10 @@ template <Namespace N> class ForeverStack
   {
     rust_assert (root.is_root ());
     rust_assert (root.is_leaf ());
+
+    // TODO: is this how we should be getting the crate node id?
```
P-E-P, line 558: Comment line 550 suggest we should change the constructor instead. Relying on the mappings current crate _should_ work, but if we need to create another foreverstack in a late compilation stage to somehow modify it, the "current crate" might be innacurate.

Might be worth removing the comment once settled ?

PR #3493, opened 2025-03-07
Fix nr2 as string ice

P-E-P:


P-E-P:
It did not make sense to store the crate as we already had it in the foreverstack. Closing in favor of #3495

PR #3494, opened 2025-03-07
testsuite: Add more testcases for `cfg()` in core

CohenArthur:
This adds a couple more testcases extracted from `cfg` directives in `core` 1.49 to ensure we can properly compile them and don't introduce regressions later on.

gcc/testsuite/ChangeLog:

	* rust/compile/cfg-core1.rs: New test.
	* rust/compile/cfg-core2.rs: New test.

### DIFF
--- a/gcc/testsuite/rust/compile/cfg-core1.rs
+++ b/gcc/testsuite/rust/compile/cfg-core1.rs
@@ -1,0 +1,12 @@
+// { dg-additional-options "-frust-cfg=A -frust-cfg=B" }
+
+#[cfg_attr(A, cfg(B))]
+struct Foo0;
+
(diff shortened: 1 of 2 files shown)


gcc/testsuite/rust/compile/cfg-core1.rs
```diff
@@ -0,0 +1,12 @@
+// { dg-additional-options "-frust-cfg=A -frust-cfg=B" }
+
+#[cfg_attr(A, cfg(B))]
+struct Foo0;
+
+#[cfg_attr(A, cfg(C))]
```
powerboat9, line 6: Are `B` and `C` supposed to be swapped?

PR #3499, opened 2025-03-13
Fix nr2 use declaration import

### LINKED ISSUE #3487: NR2: Too many super keywords
- issue-1786.rs
- issue-2330.rs
- iterators1.rs
- issue-3033.rs
- for-loop1.rs
- for-loop2.rs

`cursor()` is invalid, pointing to the crate -> invalid because we're building the mappings from the crate's scope.

Two possibles fixes : 
- build the mappings in place from the actual scope
- store the scope and resolve the path from the scope

P-E-P:
Fixes #3487


CohenArthur (approved):
LGTM at first glance! I'm not sure I understand if there is still a fixed-point going on, but the fact that two new tests are passing is reassuring. good job!

Issue #2376, opened 2023-07-04
Error: Type or Const Parameters on Foreign Items


MahadMuhammad:
### I tried this code from [E0044](https://doc.rust-lang.org/error_codes/E0044.html):
[...]

PR #3503, opened 2025-03-15
Fix Error: Type or Const Parameters on Foreign Items

sakupan102:
addresses https://github.com/Rust-GCC/gccrs/issues/2376

- \[ ] Run `clang-format`

order to get feedback.
---


philberty (approved):
LGTM

PR #3517, opened 2025-03-20
Nr2: Add missing function to silence error

P-E-P:


powerboat9:
Although it doesn't really matter, as this test is never supposed to get to type checking, the `printf` string should probably be cast to a `* const i8` (by first casting to a `*const str`) and end with an `\0`

P-E-P:
> Although it doesn't really matter, as this test is never supposed to get to type checking, the `printf` string should probably be cast to a `* const i8` (by first casting to a `*const str`) and end with an `\0`

Yeah you're right.

powerboat9:
Also, the two commits should probably be squashed together, so that the testsuite passes on every commit added to `master`

CohenArthur (approved):
Agree with @powerboat9 regarding the commits, but other than that this looks good to me!

PR #3521, opened 2025-03-21
nr2.0: Fix test `const_generics_3.rs`

powerboat9:


powerboat9:
I suppose, but I don't see a reason not to have this as a full compile test

P-E-P (approved):

PR #3527, opened 2025-03-24
nr2.0: Fix test `macros/mbe/macro43.rs`

powerboat9:
The removal of `NonZero` (unsure where it would have been defined) brings the test more in line with [libcore](https://github.com/rust-lang/rust/blob/1.49.0/library/core/src/num/nonzero.rs)


powerboat9:
https://github.com/Rust-GCC/gccrs/pull/1513 @CohenArthur it's been almost 3 years, so it's a long shot, but any thoughts?

P-E-P (approved):
Looks good.

PR #3529, opened 2025-03-24
Fix unresolved imports

### LINKED ISSUE #3486: NR2: Unresolved import
- issue-850.rs
- issue-855.rs
- iterators1.rs
- issue-3033.rs
- for-loop2.rs

P-E-P:
- [x] issue-850.rs
- [x] issue-855.rs
- [x] iterators1.rs
- [x] issue-3033.rs
- [x] for-loop2.rs

Fixes #3486

### DIFF
--- a/gcc/rust/resolve/rust-early-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-early-name-resolver-2.0.cc
@@ -417,10 +417,19 @@ Early::finalize_rebind_import (const Early::ImportPair &mapping)
       declared_name = rebind.get_identifier ().as_string ();
       locus = rebind.get_identifier ().get_locus ();
       break;
-    case AST::UseTreeRebind::NewBindType::NONE:
-      declared_name = path.get_final_segment ().as_string ();
-      locus = path.get_final_segment ().get_locus ();
-      break;
+      case AST::UseTreeRebind::NewBindType::NONE: {
+	const auto &segments = path.get_segments ();
+	// We don't want to insert `self` with `use module::self`
+	if (path.get_final_segment ().is_lower_self_seg ())
+	  {
+	    rust_assert (segments.size () > 1);
+	    declared_name = segments[segments.size () - 2].as_string ();
+	  }
+	else
+	  declared_name = path.get_final_segment ().as_string ();
+	locus = path.get_final_segment ().get_locus ();
+	break;
+      }
     case AST::UseTreeRebind::NewBindType::WILDCARD:
       rust_unreachable ();
       break;
--- a/gcc/rust/resolve/rust-forever-stack.hxx
+++ b/gcc/rust/resolve/rust-forever-stack.hxx
@@ -594,6 +594,26 @@ ForeverStack<N>::resolve_segments (
   return *current_node;
 }
 
+template <>
+inline tl::optional<Rib::Definition>
+ForeverStack<Namespace::Types>::resolve_final_segment (Node &final_node,
+						       std::string &seg_name,
+						       bool is_lower_self)
+{
+  if (is_lower_self)
+    return Rib::Definition::NonShadowable (final_node.id);
+  else
+    return final_node.rib.get (seg_name);
+}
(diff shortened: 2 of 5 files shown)


gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -643,11 +643,16 @@ ForeverStack<N>::resolve_path (
       if (final_node.rib.kind == Rib::Kind::TraitOrImpl)
 	return tl::nullopt;
 
-      std::string seg_name
-	= unwrap_type_segment (segments.back ()).as_string ();
+      auto seg = unwrap_type_segment (segments.back ());
```
powerboat9, line 646: This should probably be `auto &seg`

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -643,11 +643,16 @@ ForeverStack<N>::resolve_path (
...
+      std::string seg_name = seg.as_string ();
 
       // assuming this can't be a lang item segment
-      tl::optional<Rib::Definition> res = final_node.rib.get (seg_name);
+      tl::optional<Rib::Definition> res;
+      if (seg.is_lower_self_seg ())
+	res = final_node.parent.value ().rib.get (
+	  unwrap_type_segment (segments[segments.size () - 2]).as_string ());
```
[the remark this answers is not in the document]
powerboat9: https://doc.rust-lang.org/reference/items/use-declarations.html#self-imports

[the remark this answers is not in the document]
P-E-P: I'm far from done with this PR, I'll take a look at your comments tomorrow.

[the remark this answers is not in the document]
P-E-P: I've introduced the example from the specification to make sure we won't miss it next time.

gcc/rust/resolve/rust-early-name-resolver-2.0.cc
```diff
@@ -417,10 +417,16 @@ Early::finalize_rebind_import (const Early::ImportPair &mapping)
...
       locus = rebind.get_identifier ().get_locus ();
       break;
-    case AST::UseTreeRebind::NewBindType::NONE:
-      declared_name = path.get_final_segment ().as_string ();
-      locus = path.get_final_segment ().get_locus ();
-      break;
+      case AST::UseTreeRebind::NewBindType::NONE: {
+	const auto &segments = path.get_segments ();
+	// We don't want to insert `self` with `use module::self`
+	if (path.get_final_segment ().is_lower_self_seg ())
```
powerboat9, line 423: Would probably want to `rust_assert` there are enough segments in `path` that these accesses don't go out of bounds

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -594,6 +594,29 @@ ForeverStack<N>::resolve_segments (
...
+
+template <>
+template <typename S>
+tl::optional<Rib::Definition>
+ForeverStack<Namespace::Types>::resolve_final_segment (
+  Node &final_node, std::string &seg_name, const std::vector<S> &segments)
+{
+  auto &seg = unwrap_type_segment (segments.back ());
+  if (seg.is_lower_self_seg ())
+    return final_node.parent.value ().rib.get (
```
powerboat9, line 614: Should be able to just grab the node id from `final_node` and convert that into a `Rib::Definition` -- that should also remove the need for the `segments` parameter

powerboat9 (changes requested):
Otherwise, looks good

gcc/rust/resolve/rust-early-name-resolver-2.0.cc
P-E-P: Technically at this point the previous steps would error out if there was a self alone, bet we're never too safe.

powerboat9 (approved):

PR #3532, opened 2025-03-25
nr2.0: Handle global paths

powerboat9:
I think there may be a few more spots where we disregard opening scope resolution operators, but unless anyone objects I think I'll leave it for another PR (would be nice to pass paths directly to `NameResolutionContext::resolve_path`, rather than the properties of a given path).

### DIFF
--- a/gcc/rust/resolve/rust-forever-stack.h
+++ b/gcc/rust/resolve/rust-forever-stack.h
@@ -549,6 +549,7 @@ public:
   ForeverStack ()
     : root (Node (Rib (Rib::Kind::Normal), UNKNOWN_NODEID)),
       lang_prelude (Node (Rib (Rib::Kind::Prelude), UNKNOWN_NODEID, root)),
+      extern_prelude (Node (Rib (Rib::Kind::Prelude), UNKNOWN_NODEID)),
       cursor_reference (root)
   {
     rust_assert (root.is_root ());
@@ -671,7 +672,7 @@ public:
    */
   template <typename S>
   tl::optional<Rib::Definition> resolve_path (
-    const std::vector<S> &segments,
+    const std::vector<S> &segments, bool has_opening_scope_resolution,
     std::function<void (const S &, NodeId)> insert_segment_resolution);
 
   // FIXME: Documentation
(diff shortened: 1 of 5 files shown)


powerboat9:
Now depends on #3532

powerboat9:
`extern_prelude` would have to be populated, but as I'm not quite sure how that should work, it seems like that can be left to its own PR

gcc/rust/resolve/rust-forever-stack.h
```diff
@@ -671,7 +672,7 @@ template <Namespace N> class ForeverStack
    */
   template <typename S>
   tl::optional<Rib::Definition> resolve_path (
-    const std::vector<S> &segments,
+    const std::vector<S> &segments, bool has_opening_scope_resolution,
```
P-E-P, line 675: Could be worth using it's own enum for clarity at call site. But great as is, your call.

gcc/rust/resolve/rust-name-resolution-context.h
```diff
@@ -232,33 +233,63 @@ class NameResolutionContext
...
     return tl::nullopt;
   }
 
+  template <typename... Args>
+  tl::optional<Rib::Definition> resolve_path (const AST::SimplePath &path,
+					      Args... ns_args)
+  {
+    return resolve_path (path.get_segments (),
+			 path.has_opening_scope_resolution (), ns_args...);
+  }
```
P-E-P, line 275: Couldn't we refactor all those functions under one template ?

P-E-P (approved):
Great PR overall, I've made a few suggestions but feel free to merge without any change if you think they don't fit.

gcc/rust/resolve/rust-name-resolution-context.h
powerboat9: Would require merging `has_opening_scope_resolution`, `opening_scope_resolution`, and `has_opening_scope_resolution_op` with some renames. Good idea, but probably best left to another PR

powerboat9:
I don't have merge perms lol

gcc/rust/resolve/rust-forever-stack.h
powerboat9: It should be alright -- there aren't intended to be many direct callers

PR #3574, opened 2025-03-26
Use Rust 1.49 in CI

### LINKED ISSUE #3569: Use Rust 1.49 in CI

CohenArthur:
Fixes #3569

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -42,7 +42,7 @@ jobs:
                   g++-multilib \
                   dejagnu;
           # install Rust directly using rustup
-          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain=1.72.0;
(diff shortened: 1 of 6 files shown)


tschwinge:
Does this switch all or just a few pipelines to 1.49?  I guess we'll want to mix of "old" (minimum; 1.49), "common" (whatever different distributions ship), and "latest"?

CohenArthur:
@tschwinge I think we should stick with 1.49 in the goal of compiling it with gccrs as early as possible. I'm also happy doing it with multiple toolchains to ensure we don't rely on behavior that's removed in later Rust versions, maybe in a later PR? Also maybe something we could do daily like the bootstrap build?

CohenArthur:
(so at the moment all pipelines are switched to 1.49)

powerboat9:
Yeah, I'll need to work more on #4074.

powerboat9:
@CohenArthur still relevant?

PR #3576, opened 2025-03-26
gccrs: Fix ICE when using super mid way though path

philberty:
Fixes Rust-GCC#3568

gcc/rust/ChangeLog:

	* resolve/rust-ast-resolve-path.cc (ResolvePath::resolve_path): check for super mid path

gcc/testsuite/ChangeLog:

	* rust/compile/nr2/exclude: nr2 puts out a different error multiple times
	* rust/compile/issue-3568.rs: New test.

### DIFF
--- a/gcc/rust/resolve/rust-ast-resolve-path.cc
+++ b/gcc/rust/resolve/rust-ast-resolve-path.cc
@@ -370,6 +370,12 @@ ResolvePath::resolve_path (AST::SimplePath &expr)
 	}
       else if (segment.is_super_path_seg ())
 	{
+	  if (!is_first_segment)
+	    {
+	      rust_error_at (segment.get_locus (),
+			     "%<super%> can only be used in start position");
+	      return UNKNOWN_NODEID;
+	    }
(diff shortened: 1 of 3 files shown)


philberty:
nr2 handles this but outputs:

```
test.rs:6:14: error: leading path segment ‘super’ can only be used at the beginning of a path [E0433]
    6 | pub use foo::super::foo::S as T;                                                                        
      |              ^~~~~                                                                                                                                                                                                       
crab1: note: [ARTHUR]: resolving segment part: foo
crab1: note: [ARTHUR]: resolving segment part: super
test.rs:6:14: error: leading path segment ‘super’ can only be used at the beginning of a path [E0433]           
crab1: note: [ARTHUR]: resolving segment part: foo      
crab1: note: [ARTHUR]: resolving segment part: super 
test.rs:6:14: error: leading path segment ‘super’ can only be used at the beginning of a path [E0433]           
test.rs:6:26: error: unresolved import ‘foo::super::foo::S’ [E0433]                                             
    6 | pub use foo::super::foo::S as T;
      |                          ^
```

PR #3604, opened 2025-03-28
Add ending newline to `rust-macro-builtins-log-debug.cc`

powerboat9:
I noticed this while experimenting with a possible merge-from-upstream strategy


philberty (approved):
LGTM

PR #3677, opened 2025-04-01
install.texi: Mention Rust requirement for building gccrs

CohenArthur:
Addresses PR#117869

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117869

gcc/ChangeLog:

	* doc/install.texi: Add requirements for building gccrs.

### DIFF
--- a/gcc/doc/install.texi
+++ b/gcc/doc/install.texi
@@ -312,6 +312,12 @@ documentation including the target @code{SYSTEM} definition module.
 If Python3 is unavailable Modula-2 documentation will include a target
 independent version of the SYSTEM modules.
 
+@item @anchor{gccrs-prerequisite}gccrs
+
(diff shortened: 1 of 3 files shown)


gcc/doc/install.texi
```diff
@@ -312,6 +312,12 @@ documentation including the target @code{SYSTEM} definition module.
 If Python3 is unavailable Modula-2 documentation will include a target
 independent version of the SYSTEM modules.
 
+@item @anchor{gccrs-prerequisite}gccrs
+
+The official Rust compiler is required for building various parts of
```
P-E-P, line 317: We should mention that cargo is required alongside rustc.

P-E-P (changes requested):

gcc/doc/install.texi
CohenArthur: Thanks!

PR #3678, opened 2025-04-01
Add base for handling `#![feature(min_specialization)]`

CohenArthur:
This paves the way for properly stubbing out `min_specialization`, but does not properly handle these functions in later passes like method resolution or typechecking.


philberty (approved):
LGTM

PR #3679, opened 2025-04-01
Nr2 fix missing label

P-E-P:


powerboat9:
Looks to be an issue with warnings introduced by a gcc update in our CI?

P-E-P:
> Looks to be an issue with warnings introduced by a gcc update in our CI?

No don't worry, I also have those warning on my machine, I'll fix them soon, they are mostly coming from the expected header (`value()`) and I haven't taken a close look for now.

P-E-P:
I'd like to highlight a few bits for the review:
- This commit 4b07b95ac2ca17810977b8d9ecf475e888606c34
- Those two parts [here](https://github.com/Rust-GCC/gccrs/pull/3679/files#diff-a7f07dc44edb99c7da36b189800ff045b7a1dbbde3cce065e2250130a8e489b7R427-R441) and [here](https://github.com/Rust-GCC/gccrs/pull/3679/files#diff-a7f07dc44edb99c7da36b189800ff045b7a1dbbde3cce065e2250130a8e489b7R451-R465)

powerboat9:
@CohenArthur I'd assume the `#ifdef _MSC_VER` and matching `#endif` were supposed to be removed?

gcc/testsuite/rust/compile/invalid_label_name.rs
```diff
@@ -1,20 +1,24 @@
...
         break 'extern;
         // { dg-error "invalid label name .'extern." "" { target *-*-* } .-1 }
+        // { dg-error "use of undeclared label .'extern." "" { target *-*-* } .-2 }
     }
 
     'break: loop {
         // { dg-error "invalid label name .'break." "" { target *-*-* } .-1 }
         break 'for;
         // { dg-error "invalid label name .'for." "" { target *-*-* } .-1 }
+        // { dg-error "use of undeclared label .'for." "" { target *-*-* } .-2 }
```
powerboat9, line 14: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=7468829dd6bbda6bc3776cad99636eb1

It looks like invalid label errors are actually supposed to produce an error during parsing

P-E-P: Currently we're handling those during the AST validation step, I didn't change that part, I merely updated the test to match rustc's output (both errors are emitted). We should open an issue.

P-E-P: I tried with rustc 1.49 and no error was thrown, this behavior must have changed in between: https://godbolt.org/z/81roreTK9

powerboat9:
It does look like a lot of these commits could be fused together, or maybe split off into other smaller PRs

gcc/rust/util/expected.h
```diff
@@ -213,6 +213,8 @@ template <typename E>
 #else
   (void)e;
 #ifdef _MSC_VER
+  __assume(0);
+#else
   gcc_unreachable();
 #endif
```
powerboat9, line 219: You could try removing the `#ifdef _MSC_VER` and `#endif` instead -- seems like `gcc_unreachable` would be portable, and leaving those lines in was probably a mistake in the commit you're reverting

CohenArthur: yes, I think that makes sense. I would assume the gcc macros to be portable, and if it breaks we'll fix it in time.

P-E-P: I removed the MSVC branch and kept `gcc_unreachable()`

PR #3692, opened 2025-04-03
fix: add check for null feature gate to display error properly

### LINKED ISSUE #3616: ICE: segfault with empty feature gate
````rust
#![feature]

````

original:
````rust
#![feature]

use foo::bar::*;

fn foo<G,T>(x: impl Trt) { }

fn index1() {
    vec(); //~ ERROR cannot provide explicit generic arguments
}

````

Version information:

4019d70517af4d6d95650eb2ee3a793c57cdbe2b

Command:
`crab1 -Warray-bounds -mtune=generic -march=x86-64 -O0 -L/lib/x86_64-linux-gnu -L/lib/../lib64 -L/usr/lib/x86_64-linux-gnu -L/usr/lib/../lib64 -frust-incomplete-and-experimental-compiler-do-not-use`

<p>

```
0x1702eaf crash_signal
	../../gcc/toplev.cc:319
0x106e0f0 Rust::FeatureGate::visit(Rust::AST::Crate&)
	../../gcc/rust/checks/errors/rust-feature-gate.cc:44
0xc5c225 Rust::Session::compile_crate(char const*)
	../../gcc/rust/rust-session-manager.cc:641

```

</p>
</details>

### PR DESCRIPTION
**Pull request by matty-kuhn**
- [x] Run `clang-format`

---

Fixes #3616 

As noted in the issue, having an empty feature, `#![feature]`, causes an ICE segfault. This should actually be treated as an error by the compiler, and it now properly checks for missing input, and emits an error

By making a contribution to this project, I certify that:

    The contribution was created in whole or in part by me and I have the right to submit it under the open source license indicated in the file; or
    The contribution is based upon previous work that, to the best of my knowledge, is covered under an appropriate open source license and I have the right under that license to submit that work with modifications, whether created in whole or in part by me, under the same open source license (unless I am permitted to submit under a different license), as indicated in the file; or
    The contribution was provided directly to me by some other person who certified (a), (b) or (c) and I have not modified it.
    I understand and agree that this project and the contribution are public and that a record of the contribution (including all personal information I submit with it, including my sign-off) is maintained indefinitely and may be redistributed consistent with this project or the open source license(s) involved.

    Signed-off-by: Matty Kuhn <matty.kuhn.1@gmail.com>

gcc/rust/ChangeLog:

```
gcc/rust/ast/rust-ash.h: (AST::Attribute): add empty_input function
gcc/rust/checks/errors/rust-feature-gate.cc: (FeatureGate::visit): check if a feature is empty before checking its input
```

gcc/testsuite/ChangeLog:

```
rust/compile/feature.rs: add an invalid empty feature to produce an error
```

### CODE REVIEW

File: gcc/rust/checks/errors/rust-feature-gate.cc [hunk 1]
```diff
@@ -22,6 +22,7 @@
 #include "rust-ast-visitor.h"
 #include "rust-feature.h"
 #include "rust-ast-full.h"
+#include <cstdio>
```
P-E-P, line 25: We can't use raw system headers in gcc, you need to replace it with `rust-system.h`

gcc/rust/checks/errors/rust-feature-gate.cc
```diff
@@ -40,6 +41,13 @@ FeatureGate::visit (AST::Crate &crate)
     {
       if (attr.get_path ().as_string () == "feature")
 	{
+	  // check for empty feature, such as `#![feature], this is an error
+	  if (attr.empty_input ())
+	    {
+	      rust_error_at (attr.get_locus (), ErrorCode::E0556,
+			     "malformed feature attribute input");
```
P-E-P, line 48: We should probably quote "feature" like so: `"malformed %<feature%> attribute input"`

P-E-P (changes requested):
Some small nits can be corrected but great PR overall!

gcc/rust/checks/errors/rust-feature-gate.cc
matty-kuhn: good catch, thanks!

gcc/rust/checks/errors/rust-feature-gate.cc
matty-kuhn: thank you so much, i knew something like that was needed but for the life of me couldn't find it

gcc/testsuite/rust/compile/feature.rs
```diff
@@ -2,5 +2,7 @@
 #![feature(AA)] //{ dg-error "unknown feature .AA." }
 #![feature(iamcrabby)] // { dg-error "unknown feature .iamcrabby." }
 #![feature(nonexistent_gccrs_feature)] // { dg-error "unknown feature .nonexistent_gccrs_feature." }
+// ErrorCode - E0556
+#![feature] // { dg-error "malformed 'feature' attribute input" }
```
powerboat9, line 6: Should single quotes be used here, or should `.` be used to match any character? I'm not quite sure how the quotes work with locale, or if dejagnu sets a locale

matty-kuhn: that one i'm not sure, the single quotes passed `make check-rust` but probably wouldn't be the end of the world to change them

P-E-P: @powerboat9 is correct, a dot should be used here since the quotation changes depending on the locale (that's the same reason why we're forced to use `%qs` and `%<`/`%>`.

matty-kuhn: fixed!

gcc/testsuite/rust/compile/feature.rs
```diff
@@ -2,5 +2,7 @@
 #![feature(AA)] //{ dg-error "unknown feature .AA." }
 #![feature(iamcrabby)] // { dg-error "unknown feature .iamcrabby." }
 #![feature(nonexistent_gccrs_feature)] // { dg-error "unknown feature .nonexistent_gccrs_feature." }
+// ErrorCode - E0556
```
powerboat9, line 5: Are you trying to test for the error code here? I'm not sure how to do that myself, but I'm pretty sure this won't check the error code.

matty-kuhn: no that was just to indicate that this is a different rust error code than the one mentioned above

powerboat9 (approved):

P-E-P (approved):

PR #3702, opened 2025-04-07
nr2.0: Fix multiple bindings

### LINKED ISSUE #3478: NR2: Multiple definition
- `issue-2812.rs`

P-E-P:
Fixes #3478

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -2102,6 +2102,19 @@ template <> struct less<Rust::Identifier>
     return lhs.as_string () < rhs.as_string ();
   }
 };
+
+template <> struct hash<Rust::Identifier>
+{
+  std::size_t operator() (const Rust::Identifier &k) const
+  {
+    using std::hash;
+    using std::size_t;
+    using std::string;
+
+    return hash<string> () (k.as_string ()) ^ (hash<int> () (k.get_locus ()));
+  }
+};
+
 } // namespace std
 
 #endif
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -57,6 +57,11 @@ public:
 
   bool empty () const { return ident.empty (); }
 
+  bool operator== (const Identifier &other) const
+  {
+    return ident == other.ident;
+  }
+
 private:
   std::string ident;
   location_t loc;
--- a/gcc/rust/resolve/rust-late-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-late-name-resolver-2.0.cc
@@ -133,11 +133,11 @@ Late::visit (AST::ForLoopExpr &expr)
 {
   visit_outer_attrs (expr);
 
-  ctx.bindings.new_binding (BindingSource::For);
+  ctx.bindings.enter (BindingSource::For);
 
   visit (expr.get_pattern ());
 
-  ctx.bindings.clear ();
+  ctx.bindings.exit ();
 
   visit (expr.get_iterator_expr ());
   visit (expr.get_loop_label ());
@@ -149,12 +149,12 @@ Late::visit (AST::IfLetExpr &expr)
 {
   visit_outer_attrs (expr);
 
-  ctx.bindings.new_binding (BindingSource::Let);
+  ctx.bindings.enter (BindingSource::Let);
 
   for (auto &pattern : expr.get_patterns ())
     visit (pattern);
 
-  ctx.bindings.clear ();
+  ctx.bindings.exit ();
 
   visit (expr.get_value_expr ());
   visit (expr.get_if_block ());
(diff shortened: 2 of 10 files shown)


gcc/testsuite/rust/compile/nr2/exclude
```diff
@@ -1,13 +1,10 @@
 canonical_paths1.rs
 cfg1.rs
 generics9.rs
-issue-2043.rs
```
P-E-P, line 4: This shouldn't have been fixed by the changes, I'll take a closer look soon.

gcc/rust/resolve/rust-late-name-resolver-2.0.cc
```diff
@@ -167,9 +220,80 @@ Late::visit (AST::IdentifierPattern &identifier)
...
+    for (auto &generic : function.get_generic_params ())
+      visit (generic);
+
+    // We only care about params
+    ctx.bindings.new_binding (BindingSource::Param);
+
+    for (auto &param : function.get_function_params ())
+      visit (param);
+
+    ctx.bindings.clear ();
```
P-E-P, line 284: We could introduce some functions in the AST visitor specially dedicated to visiting the parameters, this way we could override those functions instead of the whole visit.

gcc/rust/resolve/rust-name-resolution-context.h
```diff
@@ -156,6 +156,108 @@ class Definition
...
+
+  /**
+   * Remove current layer's last binding level and merge it with it's parent.
+   */
+  void merge ();
+
+  /**
+   * Push a new binding level in the current binding layer.
+   */
+  void push (Binding::Kind kind);
```
P-E-P, line 238: Not a fan of those forwarding functions, we could expose a `current()` function instead that would return the current layer. Thoughts ?

gcc/rust/ast/rust-ast.h
```diff
@@ -57,6 +57,11 @@ class Identifier
 
   bool empty () const { return ident.empty (); }
 
+  bool operator== (const Identifier &other) const
+  {
+    return ident == other.ident && loc == other.loc;
```
powerboat9, line 62: Two identifiers aren't equal if they have different locations?

gcc/rust/resolve/rust-name-resolution-context.h
```diff
@@ -156,6 +156,108 @@ class Definition
...
+  void insert_ident (Identifier ident);
+
+  void merge ();
+
+  BindingSource get_source () const;
+};
+
+class BindingContext
+{
+  std::vector<BindingLayer> bindings;
```
powerboat9, line 217: This is meant to handle cases if, say, we go from visiting a pattern -> block expression -> let statement -> new binding layer?

powerboat9: I'm not sure we could ever hit a situation like that -- maybe if `rustc` ever gets support for more complex range/literal patterns?

powerboat9: Might be best to replace `BindingContext` with a `tl::optional<BindingLayer>`

P-E-P: In fact we do hit a situation like that in one test (issue-3566-2.rs), because we need to hold the binding product between multiple function arguments we need to be able to stack binding layers in case another binding is introduced in a type const context.

CohenArthur, line 217: You can use a `StackedContext` here I think? would help add some safety around the `.pop()` and `.back()` operations

gcc/rust/resolve/rust-name-resolution-context.h
```diff
@@ -156,6 +156,108 @@ class Definition
...
+public:
+  /**
+   * Identifies if the identifier is and-binded in the current layer's bindings.
+   */
+  bool and_binded (Identifier ident);
+
+  /**
+   * Identifies if the identifier is or-binded in the current layer's bindings.
+   */
+  bool or_binded (Identifier ident);
```
CohenArthur, line 228: I would rename these `is_and_bound()` and `is_or_bound()` since they return a boolean. I was a bit confused when reading the code using these functions above

P-E-P:
@CohenArthur @philberty shall I drop e169a88a67fbee712dcdf6b34dfbb5f048d865bb ? I'm unsure about it.

On one hand it allows us to avoid mistakes and reduce maintenance cost (we don't have to duplicate this visit function code, only one small part). But on the other hand it introduces a new kind of interface to the default visitor that is not as straightforward as the current one.

powerboat9:
> @CohenArthur @philberty shall I drop [e169a88](https://github.com/Rust-GCC/gccrs/commit/e169a88a67fbee712dcdf6b34dfbb5f048d865bb) ? I'm unsure about it.
> 
> On one hand it allows us to avoid mistakes and reduce maintenance cost (we don't have to duplicate this visit function code, only one small part). But on the other hand it introduces a new kind of interface to the default visitor that is not as straightforward as the current one.
> 
> We could also change is to something like `visit(std::vector<Param>)` instead, that would be more "in line" with the rest of the visitor, but the "intermediate" aspect would be less obvious and that would prevent us from using this kind of intermediate visitor step for subsets of fields that are not in the same container (unless we use multiple arguments ?).

It's a bit odd, but it seems like the best solution

PR #3732, opened 2025-04-10
Fix narrowing conversion warnings

powerboat9:
Fixes https://gcc.gnu.org/bugzilla/show_bug.cgi?id=119641

### DIFF
--- a/gcc/rust/checks/errors/borrowck/rust-bir-builder-internal.h
+++ b/gcc/rust/checks/errors/borrowck/rust-bir-builder-internal.h
@@ -357,8 +357,10 @@ protected: // Helpers to add BIR statements
 protected: // CFG helpers
   BasicBlockId new_bb ()
   {
+    size_t id = ctx.basic_blocks.size ();
+    rust_assert (id < INVALID_BB.value);
     ctx.basic_blocks.emplace_back ();
-    return {ctx.basic_blocks.size () - 1};
+    return {static_cast<BasicBlockId::value_type> (id)};
   }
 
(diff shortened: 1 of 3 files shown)


powerboat9:
Open to suggestions on how to make this a bit less ugly, but we should probably fix this bug before GCC 15 releases

CohenArthur (changes requested):
I don't think this is the best way to go about this. In some places the ID stored was `size - 1`, which is different here. I think we can also find a cleaner way to do this, like maybe storing `size_t` instead of `uint32_t` in all of those `*Id` structs? It used to be an issue when they were type aliases but that's no longer the case

powerboat9:
The `emplace_back` calls increment the value returned by `size`. As for changing the value type to `size_t`, that does sound ideal, but I don't know why we're using `uint32_t` in the first place, some data structure size optimization?

powerboat9:
@jdupak

powerboat9:
> I don't think this is the best way to go about this. In some places the ID stored was `size - 1`, which is different here. I think we can also find a cleaner way to do this, like maybe storing `size_t` instead of `uint32_t` in all of those `*Id` structs? It used to be an issue when they were type aliases but that's no longer the case

What was an issue with type aliases? Some 32-bit compatibility issue, where `uint32_t` may or may not have been equivalent to `size_t`? I think I remember something like that

powerboat9:
I think I've found a better way to do this, by modifying `IndexVec`

CohenArthur (approved):
ooooh I really like that, well done!! I think this would be really good

PR #3738, opened 2025-04-15
Handle const blocks

CohenArthur:
This adds some handling towards `const { ... }` expressions. I haven't added specific handling for const folding them, but they are properly handled by the const checker


gcc/rust/hir/rust-ast-lower-expr.cc
```diff
@@ -126,6 +127,43 @@ ASTLoweringExpr::visit (AST::BlockExpr &expr)
...
+				   std::unique_ptr<Expr> (inner_expr),
+				   expr.get_locus ());
+}
+
+void
+ASTLoweringExpr::visit (AST::ConstBlock &expr)
+{
+  auto inner_expr = ASTLoweringExpr::translate (expr.get_const_expr ());
+
+  // we know this will always be an `AnonConst`, or we have an issue. Let's
```
philberty, line 151: should this not be in the visitor above?

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -410,7 +411,8 @@ CompileExpr::visit (HIR::BlockExpr &expr)
   if (!ctx->get_tyctx ()->lookup_type (expr.get_mappings ().get_hirid (),
 				       &block_tyty))
     {
-      rust_error_at (expr.get_locus (), "failed to lookup type of BlockExpr");
+      rust_error_at (expr.get_locus (),
+		     "failed to lookup type of BlockExpr 414");
```
P-E-P, line 415: What does the 414 means here ? Is it supposed to be
```cpp
rust_error_at (expr.get_locus (), ErrorCode::E0414,
		     "failed to lookup type of BlockExpr");
```
instead ?

gcc/rust/hir/tree/rust-hir.cc
```diff
@@ -1047,6 +1047,33 @@ BlockExpr::as_string () const
   return str;
 }
 
+std::string
+AnonConst::as_string () const
+{
+  std::string istr = indent_spaces (enter);
```
P-E-P, line 1053: Could be replaced with a string_stream

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -7234,6 +7234,30 @@ Parser<ManagedTokenSource>::parse_block_expr (
...
+						    location_t locus)
+{
+  auto block = parse_block_expr ();
+
+  if (!block)
+    {
+      add_error (Error (locus, "failed to parse inner block in const block"));
+      skip_after_end_block ();
+
+      return nullptr;
```
P-E-P, line 7251: We should probably return an unexpected error and let the caller chose instead.

P-E-P (approved):

gcc/rust/backend/rust-compile-expr.cc
CohenArthur: oops no that was just a random string for grepping, as this error message is duplicated in the compiler. I should have removed it

CohenArthur:
stupid uninitialized booleans getting me again. I have half a mind to add a `safe_bool` class to the project lol, or turn all of them into enums

powerboat9:
> stupid uninitialized booleans getting me again. I have half a mind to add a `safe_bool` class to the project lol, or turn all of them into enums

It's surprising that GCC isn't (?) warning about them

gcc/rust/hir/tree/rust-hir.cc
CohenArthur: there are no string streams in any other implementations, I think it'd be better to fix them all at once

CohenArthur:
> > stupid uninitialized booleans getting me again. I have half a mind to add a `safe_bool` class to the project lol, or turn all of them into enums
> 
> It's surprising that GCC isn't (?) warning about them

yeah it bothers me a great deal...

PR #3741, opened 2025-04-16
Add basic struct pattern rebinding

CohenArthur:
This PR adds support for basic struct ident patterns, specifically in the case of rebinding struct fields to new names:

```rust
match instance {
    Foo { field_name: new_name } => { do_something(new_name) }
}
```

This is required for finish support for `derive(PartialEq)` and `derive(PartialOrd)`

### DIFF
--- a/gcc/rust/backend/rust-compile-pattern.cc
+++ b/gcc/rust/backend/rust-compile-pattern.cc
@@ -22,6 +22,8 @@
 #include "rust-constexpr.h"
 #include "rust-compile-type.h"
 #include "print-tree.h"
+#include "rust-hir-pattern.h"
+#include "rust-system.h"
 
 namespace Rust {
 namespace Compile {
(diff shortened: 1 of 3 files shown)


philberty (approved):
LGTM the sooner we get the match expression rewritten to become conditionals the the better at this point i think this is peek switch statement now

powerboat9:
> LGTM the sooner we get the match expression rewritten to become conditionals the the better at this point i think this is peek switch statement now

It's been a while, but I thought I already did that

gcc/rust/backend/rust-compile-pattern.h
```diff
@@ -78,6 +80,13 @@ class CompilePatternBindings : public HIRCompileBase,
     pattern.accept_vis (compiler);
   }
 
+  void handle_struct_pattern_ident (HIR::StructPatternField &pat,
+				    TyTy::ADTType *adt,
+				    TyTy::VariantDef *variant,
+				    int variant_index);
+  void handle_struct_pattern_ident_pat (HIR::StructPatternField &pat);
+  void handle_struct_pattern_tuple_pat (HIR::StructPatternField &pat);
+
```
powerboat9, line 95: These should probably be `private`

gcc/rust/backend/rust-compile-pattern.cc
```diff
@@ -259,34 +259,16 @@ CompilePatternCheckExpr::visit (HIR::StructPattern &pattern)
...
-
-	    // we may be offsetting by + 1 here since the first field in the
-	    // record is always the discriminator
-	    offs += adt->is_enum ();
-	    tree field_expr
-	      = Backend::struct_field_expression (match_scrutinee_expr, offs,
-						  ident.get_locus ());
+	case HIR::StructPatternField::ItemType::IDENT_PAT:
+	  // we only support rebinding patterns at the moment, which always
+	  // match - so do nothing
```
powerboat9, line 264: What's wrong with the old checks?

gcc/rust/backend/rust-compile-pattern.h
CohenArthur: good catch thanks

gcc/rust/backend/rust-compile-pattern.cc
CohenArthur: they caused an invalid call to `struct_field_expression` which I didn't want to explore. since when compiling struct pattern matching with `StructPatternFieldIdentPat` (so `match x { Foo { field: new_name } => {} }` we would always reach a `rust_unreachable()`, I think this code was actually never ever reached or executed and was just wrong

powerboat9: Yeah, I think this was meant to handle struct-like enum items, but was probably broken by https://github.com/Rust-GCC/gccrs/commit/806166db9db83bee816d169a13ab9992de9d66a0. Is there any chance you could fix it, rather than remove it? I could explain any parts of the code you have questions about, or fix it myself and give you the patch.

CohenArthur: but maybe I'm wrong? happy to change it and try to fix it instead if you think this was correct

powerboat9: I think the code is close enough to right that getting it to work should be pretty simple -- check out where `match_scrutinee_expr` is being modified in the if statement above, you may have to remove that for `struct_field_expression` to work.

CohenArthur: sounds good, I'll take a look then :) didn't realize it had been modified recently, my bad! I thought I was just hitting an uncharted part of the codebase lol

P-E-P (approved):

gcc/rust/backend/rust-compile-pattern.cc
CohenArthur: okay so because we can only handle rebinding instead of all kinds of struct sub patterns idents, the code should stay empty for now - because it's like compiling an Identifier pattern, it's just always true.

however, it could benefit from a nice comment explaining what we currently do (nothing) and what we need to do for more complex struct patterns, e.g. non rebinds like this:

```rust
match x {
    Foo { field: 15 } => { ... }
}
```

where we actually have to check that `field` matches `15`, or something like

```rust
match x {
    Foo { field: CONST_VALUE } => { ... }
}
```

which is a little bit harder to handle. but that should be done in a later pass, and possibly as part of this year's GSoC

CohenArthur: oh the comment is already pretty good actually. but yeah I think this needs to be fixed in a later PR

PR #3742, opened 2025-04-16
Finish `derive(PartialEq)`

### LINKED ISSUE #2989: PartialEq

CohenArthur:
Fixes #2989 

Needs to be rebased on #3741 once it is merged


gcc/testsuite/rust/execute/torture/derive-partialeq2.rs
```diff
@@ -0,0 +1,66 @@
...
+
+extern "C" {
+    fn puts(s: *const i8);
+}
+
+fn print(b: bool) {
+    if b {
+        unsafe { puts("true" as *const str as *const i8) }
+    } else {
+        unsafe { puts("false" as *const str as *const i8) }
```
P-E-P, line 46:
```suggestion
        unsafe { puts("true\0" as *const str as *const i8) }
    } else {
        unsafe { puts("false\0" as *const str as *const i8) }
```

PR #3744, opened 2025-04-16
gccrs: Fix ICE when checking shift's which are behind array refs

philberty:
I copied a bad form of this check from the c front-end this updates it to ensure the rhs is an INTEGER_CST and the lhs needs checked in the first place.

Fixes Rust-GCC#3664

gcc/rust/ChangeLog:

	* rust-gcc.cc (arithmetic_or_logical_expression): Ensure this is an integer

gcc/testsuite/ChangeLog:

	* rust/compile/issue-3664.rs: New test.


powerboat9:
Why `TYPE_OVERFLOW_WRAPS`?

philberty:
I copied it from c front-end let me fine the link

PR #3745, opened 2025-04-16
gccrs: Add check for placeholder (infer) type in return position

philberty:
It is not allowed to have a declared inference variable in the return position of a function as this may never get infered you need good points of truth.

Ideally if we get a student for GSoC 25 we will get the Default Hir Visitor so that we can grab the HIR::InferredType locus instead of using the ref location lookups.

Fixes Rust-GCC#402

gcc/rust/ChangeLog:

	* typecheck/rust-hir-type-check-item.cc (TypeCheckItem::visit): add diagnostic
	* typecheck/rust-tyty.cc (BaseType::contains_infer): new helper to grab first infer var
	* typecheck/rust-tyty.h: prototype

gcc/testsuite/ChangeLog:

	* rust/compile/issue-402.rs: New test.


philberty:
The error looks pretty good to me:

image

PR #3751, opened 2025-04-17
gccrs: prealloc the initilizer vector

philberty:
There are two cases when initilizing an array, this is the const context which means we need to build the array ctor, which means using lots of memory, its super inefficient because we are using a big wrapper over the GCC internals here but preallocating the vectors here causes a:

  terminate called after throwing an instance of 'std::bad_alloc'

So this is a handy error condition to rely on for this senario.

Fixes Rust-GCC#3713
Fixes Rust-GCC#3727

gcc/rust/ChangeLog:

	* backend/rust-compile-expr.cc (CompileExpr::array_copied_expr): prealloc the vector

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -1965,8 +1965,12 @@ CompileExpr::array_copied_expr (location_t expr_locus,
   if (ctx->const_context_p ())
     {
       size_t idx = 0;
+
       std::vector<unsigned long> indexes;
       std::vector<tree> constructor;
+
+      indexes.reserve (len);
+      constructor.reserve (len);
       for (unsigned HOST_WIDE_INT i = 0; i < len; i++)
 	{
 	  constructor.push_back (translated_expr);


powerboat9:
GCC doesn't support, ex, `[12; 1024*1024*1024]` requiring a reasonable amount of memory in GIMPLE?

powerboat9:
Maybe we could declare a custom tree + gimplification hooks? Or add a new tree to the gcc middle end for lazy-initialized array construction?

powerboat9:
```text
/* Used to represent the brace-enclosed initializers for a structure or an
   array.  It contains a sequence of component values made out of a VEC of
   constructor_elt.

   For RECORD_TYPE, UNION_TYPE, or QUAL_UNION_TYPE:
   The field INDEX of each constructor_elt is a FIELD_DECL.
   
   For ARRAY_TYPE:
   The field INDEX of each constructor_elt is the corresponding index.
   If the index is a RANGE_EXPR, it is a short-hand for many nodes,
   one for each index in the range.  (If the corresponding field VALUE
   has side-effects, they are evaluated once for each element.  Wrap the
   value in a SAVE_EXPR if you want to evaluate side effects only once.)

   Components that aren't present are cleared as per the C semantics,
   unless the CONSTRUCTOR_NO_CLEARING flag is set, in which case their
   value becomes undefined.  */
DEFTREECODE (CONSTRUCTOR, "constructor", tcc_exceptional, 0)
```

Looks like `CONSTRUCTOR` trees actually already support this

philberty:
I think letting the alloc fail is about the best we can do or we impose some kind of limits. When you create the constructor its basically a big linked list so i think doing this vector reserve upfront is about the best way to detect this case atm.

I'll leave this one up so others can have a think about it.

philberty:
I tagged in @dafaust because he implemented the other case where we already handle these giant arrays let me find the test case see 5559bdc86635

But the crashes are for const evaluation, maybe now that we have ported over the c++ constexpr handler maybe we can just fallback to this old code i must give that a try and we can just let the constant folder do the work?

powerboat9:
You can use `RANGE_EXPR` as an index -- that might be outside the scope of this PR though

philberty:
Actually using RANGE_EXPR to make the ctor for copied array's might be a neat solution to try out

PR #3759, opened 2025-04-18
nr2.0: Improve error handling

powerboat9:
This should allow errors produced directly or indirectly by `ForeverStack::resolve_path` to be intelligently emitted or discarded. This can be used to avoid duplicate error messages, and in the near future for better language prelude and `pub(in <path>)` handling.

### DIFF
--- a/gcc/rust/resolve/rust-early-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-early-name-resolver-2.0.cc
@@ -141,6 +141,10 @@ Early::build_import_mapping (
       // be moved into the newly created import mappings
       auto path = import.to_resolve;
 
+      // used to skip the "unresolved import" error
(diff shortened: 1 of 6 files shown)


gcc/rust/resolve/rust-early-name-resolver-2.0.cc
```diff
@@ -154,7 +158,7 @@ Early::build_import_mapping (
 	  break;
 	}
 
-      if (!found)
+      if (!found && old_error_count == macro_resolve_errors.size ())
```
philberty, line 161: is this to detect no new errors have been emitted?

powerboat9: Yep

CohenArthur (approved):
LGTM! well done!

Issue #3740, opened 2025-04-16
nr2.0: Tests unexpectedly passing upstream


powerboat9:
I saw this today on GCC trunk:
```
+               === rust tests ===
+
+Running rust:/gcc/testsuite/rust/compile/nr2/compile.exp ...
+XPASS: canonical_paths1.rs on nr2
+XPASS: cfg1.rs on nr2
+XPASS: derive-debug1.rs on nr2
+XPASS: derive-default1.rs on nr2
+XPASS: multiple_bindings1.rs on nr2
+XPASS: multiple_bindings2.rs on nr2
+XPASS: privacy5.rs on nr2
+XPASS: privacy8.rs on nr2
+XPASS: pub_restricted_1.rs on nr2
+XPASS: pub_restricted_2.rs on nr2
+XPASS: pub_restricted_3.rs on nr2
+XPASS: torture/alt_patterns1.rs on nr2
```

But it seems far rarer than before (was many more tests last time, and I hit it very frequently).

_Originally posted by @thesamesam in https://github.com/Rust-GCC/gccrs/issues/3674#issuecomment-2809337094_

powerboat9:
@thesamesam it happens only occasionally?

thesamesam:
Yeah, only occasionally. I've only managed to hit it once since your earlier fix (and I build at least once daily).

thesamesam:
cc @tschwinge as he can reproduce it more easily, it seems

thesamesam:
Looking at https://inbox.sourceware.org/gcc-rust/ (e.g. https://builder.sourceware.org/buildbot/#/builders/16/builds/3471/steps/6/logs/stdio), it looks like the buildbot is hitting this sometimes now.

PR #3760, opened 2025-04-19
Disable parallel testing for 'rust/compile/nr2/compile.exp' [PR119508]

tschwinge:
..., using the standard idiom.  This '*.exp' file doesn't adhere to the parallel testing protocol as defined in 'gcc/testsuite/lib/gcc-defs.exp'.

This also restores proper behavior for '*.exp' files executing after (!) this one, which erroneously caused hundreds or even thousands of individual test cases get duplicated vs. skipped, randomly, depending on the '-jN' level.

	PR testsuite/119508
	gcc/testsuite/
	* rust/compile/nr2/compile.exp: Disable parallel testing.

(cherry picked from commit 79d2c3089f480738613b7d338d86d8be710f8158)

Resolves Rust-GCC/gccrs#3740.

### DIFF
--- a/gcc/testsuite/rust/compile/nr2/compile.exp
+++ b/gcc/testsuite/rust/compile/nr2/compile.exp
@@ -19,6 +19,15 @@
 # Load support procs.
 load_lib rust-dg.exp
 
+# These tests don't run runtest_file_p consistently if it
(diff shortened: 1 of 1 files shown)


powerboat9:
Oh, that explains what was going on. Thanks for fixing that

PR #3775, opened 2025-05-03
nr2.0: Adjust enum item visitors

powerboat9:
This should make the visitation of `EnumItem` instances more straightforward, as well as add some missing calls to base class visitors


philberty (approved):
LGTM

PR #3776, opened 2025-05-05
nr2.0: Separate out canonical path handling

powerboat9:
~~Depends on a few other PRs~~

### DIFF
--- a/gcc/rust/resolve/rust-default-resolver.cc
+++ b/gcc/rust/resolve/rust-default-resolver.cc
@@ -25,6 +25,21 @@ namespace Rust {
 namespace Resolver2_0 {
 
 void
+DefaultResolver::visit (AST::Crate &crate)
+{
+  auto inner_fn = [this, &crate] () { AST::DefaultASTVisitor::visit (crate); };
+
+  auto &mappings = Analysis::Mappings::get ();
+
+  auto crate_num = mappings.lookup_crate_num (crate.get_node_id ());
+  rust_assert (crate_num.has_value ());
+  auto crate_name = mappings.get_crate_name (*crate_num);
+  rust_assert (crate_name.has_value ());
+
+  ctx.canonical_ctx.scope_crate (crate.get_node_id (), *crate_name, inner_fn);
+}
+
+void
 DefaultResolver::visit (AST::BlockExpr &expr)
 {
   // extracting the lambda from the `scoped` call otherwise the code looks like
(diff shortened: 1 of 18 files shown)


powerboat9:
Now only depends on #3768

gcc/rust/resolve/rust-name-resolution-context.cc
```diff
@@ -82,8 +82,69 @@ BindingLayer::get_source () const
...
+      auto res = ctx.lookup (lookup_id).and_then (
+	[&ctx] (NodeId id) { return ctx.canonical_ctx.get_record_opt (id); });
+
+      if (!res)
+	{
+	  // HACK: use a dummy value
+	  // this should bring us roughly to parity with nr1.0
+	  // since nr1.0 doesn't seem to handle canonical paths for generics
+	  //   quite right anyways
+	  return Resolver::CanonicalPath::new_seg (UNKNOWN_NODEID, "XXX");
```
philberty, line 120: yeah generics are a right pain to handle sometims the paths always default to their non monomorphized version sometimes not but yeah. In order to do it all "correctly" there is going to need to be some mappings between the segment and the conical path segments.. We get away with it for now because it doesnt matter, plus rust will never get a stable ABI and the hash at the end of the managled symbol.

My current feel is that if we can do what we can with canonical paths for now purely in name resolution we get like 80% the way there and its ok.

gcc/rust/util/rust-canonical-path.h
```diff
@@ -57,10 +57,11 @@ class CanonicalPath
     return *this;
   }
 
-  static CanonicalPath new_seg (NodeId id, const std::string &path)
+  static CanonicalPath new_seg (NodeId id, std::string path)
```
philberty, line 60: Not sure i like getting rid of the const reference here. I personally like to avoid move on std::string but if this is the style in nr2 then go for it.

philberty (approved):
LGTM

gcc/rust/util/rust-canonical-path.h
powerboat9: I was having some issue with binding to a constant reference -- in hindsight, I'm not sure why

gcc/rust/resolve/rust-default-resolver.cc
```diff
@@ -98,23 +132,27 @@ DefaultResolver::visit (AST::InherentImpl &impl)
...
     visit (impl.get_type ());
 
-    ctx.scoped (Rib::Kind::TraitOrImpl, impl.get_node_id (), inner_fn_inner);
+    ctx.scoped (Rib::Kind::TraitOrImpl, impl.get_node_id (), inner_fn_1);
   };
 
-  ctx.scoped (Rib::Kind::Generics, impl.get_node_id (), inner_fn_outer);
+  auto inner_fn_3 = [this, &impl, &inner_fn_2] () {
+    ctx.canonical_ctx.scope_impl (impl, std::move (inner_fn_2));
+  };
```
CohenArthur, line 153: why do we use `.scope_impl(()` instead of `.scope()` here?

powerboat9: Since the canonical path of an inherent impl block is dependent on its impl type

PR #3778, opened 2025-05-06
gccrs: desugar APIT impl traits

philberty:
Argument position impl traits are simply syntatic sugar for generics. This adds a new desugar pass to do this. So for example:

    fn foo(a: impl Value, b: impl Value) -> i32

Is desugared into:

    fn foo<T: Value, U: Value> (a: T, b: U) -> i32

So it just works like any normal generic function. There are more complex cases such as:

    fn foo(_value: impl Bar<Baz = impl Foo>) -> i32

Which has a generic argument binding which needs to be turned into a where constraint:

    fn foo<T, U>(_value: T) -> i32
        where
            T: Bar<Baz = U>,
            U: Foo,

Fixes Rust-GCC#2015
Fixes Rust-GCC#1487
Fixes Rust-GCC#3454
Fixes Rust-GCC#1482

gcc/rust/ChangeLog:

	* Make-lang.in: new desugar file
	* ast/rust-ast.cc (ImplTraitTypeOneBound::as_string): its a unique_ptr now
	(FormatArgs::set_outer_attrs): reformat
	* ast/rust-path.h: remove has_generic_args assertion (can be empty because of desugar)
	* ast/rust-type.h (class ImplTraitTypeOneBound): add copy ctor and use unique_ptr
	* hir/rust-ast-lower-type.cc (ASTLoweringType::visit): update to use unique_ptr
	* parse/rust-parse-impl.h (Parser::parse_type): reuse the existing unique_ptr instead
	(Parser::parse_type_no_bounds): likewise
	(Parser::parse_pattern): likewise
	* resolve/rust-ast-resolve-type.cc (ResolveType::visit): its a unique_ptr now
	* rust-session-manager.cc (Session::compile_crate): call desugar
	* ast/rust-desugar-apit.cc: New file.
	* ast/rust-desugar-apit.h: New file.

gcc/testsuite/ChangeLog:

	* rust/compile/issue-2015.rs: fully supported now
	* rust/compile/nr2/exclude: nr2 cant handle some of these
	* rust/compile/issue-1487.rs: New test.
	* rust/compile/issue-3454.rs: New test.
	* rust/execute/torture/impl_desugar-2.rs: New test.
	* rust/execute/torture/impl_desugar.rs: New test.
	* rust/execute/torture/impl_trait1.rs: New test.
	* rust/execute/torture/impl_trait2.rs: New test.
	* rust/execute/torture/impl_trait3.rs: New test.
	* rust/execute/torture/issue-1482.rs: New test.


gcc/rust/ast/rust-ast.cc
```diff
@@ -5047,7 +5047,8 @@ FormatArgs::get_outer_attrs ()
   rust_unreachable ();
 }
 
-void FormatArgs::set_outer_attrs (std::vector<Attribute>)
+void
+FormatArgs::set_outer_attrs (std::vector<Attribute>)
```
powerboat9, line 5051: Looks like you should ditch this formatting change -- it breaks the CI

powerboat9:
🫠

gcc/rust/ast/rust-desugar-apit.cc
```diff
@@ -0,0 +1,570 @@
...
+
+public:
+  static std::pair<AST::Type *, std::vector<std::unique_ptr<GenericParam>>>
+  Desugar (AST::Type &type)
+  {
+    DesugarApitType visitor;
+    type.accept_vis (visitor);
+    return std::make_pair (visitor.translated,
+			   std::move (visitor.implicit_generic_params));
+  }
```
powerboat9, line 38: Is this supposed to return `nullptr` when a change isn't made, or a pointer to `type`? It looks like some code assumes the first, while other code assumes the second.

philberty: i will tweak this because in theory it should never return nullptr now the way it works good spot

gcc/rust/ast/rust-desugar-apit.cc
```diff
@@ -0,0 +1,570 @@
...
+
+    // Replace the element type with our new one, preserving the node
+    type.get_elem_type_ptr () = std::move (new_elem_type);
+
+    // Collect all the implicit generic parameters we found
+    for (auto &implicit_generic : implicit_generics)
+      implicit_generic_params.push_back (std::move (implicit_generic));
+
+    // Keep using the original node
+    translated = &type;
```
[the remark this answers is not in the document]
philberty: yeah maybe a generic might be handy will do

gcc/rust/ast/rust-desugar-apit.cc
philberty: this now defaults properly so no nullptr needed

Issue #3746, opened 2025-04-17
Missing `InlineAsm` node dump


P-E-P:
We miss the `InlineAsm` node dump in `rust-ast-collector.cc` which may be required if we need to expand an attribute proc macro on a function containing inline assembly.

powerboat9:
Sure, go ahead

PR #3779, opened 2025-05-07
ast: collect InlineAsm node dump

0xn4utilus:
Visit `InlineAsm` in ast-collector. Move `AST::InlineAsmOption` to `AST::InlineAsm::Option`.

Fixes https://github.com/Rust-GCC/gccrs/issues/3746.


philberty (changes requested):
This should all just be 1 commit it looks all related to a single change in order to improve ast dump of inline asm

PR #3781, opened 2025-05-08
Fix ICE segfault on empty static loops

### LINKED ISSUE #3618: ICE: segfault assigning loop to static
````rust
static _X: () = loop {};

````

original:
````rust
static _X: () = loop {};

fn main() {}

````

Version information:

4019d70517af4d6d95650eb2ee3a793c57cdbe2b

Command:
`crab1 -Warray-bounds -mtune=generic -march=x86-64 -O0 -L/lib/x86_64-linux-gnu -L/lib/../lib64 -L/usr/lib/x86_64-linux-gnu -L/usr/lib/../lib64 -frust-incomplete-and-experimental-compiler-do-not-use`

<p>

```
0x1702eaf crash_signal
	../../gcc/toplev.cc:319
0xffec20 contains_struct_check(tree_node const*, tree_node_structure_enum, char const*, int, char const*)
	../../gcc/tree.h:4045
0xffec20 error_operand_p(tree_node const*)
	../../gcc/tree.h:4637
0xffec20 eval_constant_expression
	../../gcc/rust/backend/rust-constexpr.cc:1896
0xffefd2 eval_constant_expression
	../../gcc/rust/backend/rust-constexpr.cc:2196
0xfff7c8 eval_loop_expr
	../../gcc/rust/backend/rust-constexpr.cc:4894
0xfff7c8 eval_constant_expression
	../../gcc/rust/backend/rust-constexpr.cc:2319
0x1000086 eval_statement_list
	../../gcc/rust/backend/rust-constexpr.cc:4658
0x1000086 eval_constant_expression
	../../gcc/rust/backend/rust-constexpr.cc:2192
0xffefd2 eval_constant_expression
	../../gcc/rust/backend/rust-constexpr.cc:2196
0xffdd4b eval_call_expression
	../../gcc/rust/backend/rust-constexpr.cc:3583
0xffef98 eval_constant_expression
	../../gcc/rust/backend/rust-constexpr.cc:2110
0x1001b90 Rust::Compile::fold_expr(tree_node*)
	../../gcc/rust/backend/rust-constexpr.cc:581
0x100e4d2 Rust::Compile::HIRCompileBase::compile_constant_item(unsigned int, Rust::TyTy::BaseType*, Rust::TyTy::BaseType*, Rust::Resolver::CanonicalPath const&, Rust::HIR::Expr&, unsigned int, unsigned int)
	../../gcc/rust/backend/rust-compile-base.cc:869
0xfd9227 Rust::Compile::CompileItem::visit(Rust::HIR::StaticItem&)
	../../gcc/rust/backend/rust-compile-item.cc:73
0xc64a14 Rust::Compile::CompileItem::compile(Rust::HIR::Item*, Rust::Compile::Context*, Rust::TyTy::BaseType*, unsigned int)
	../../gcc/rust/backend/rust-compile-item.h:37
0xc64a14 Rust::Compile::CompileCrate::go()
	../../gcc/rust/backend/rust-compile.cc:48
0xc64afb Rust::Compile::CompileCrate::Compile(Rust::HIR::Crate&, Rust::Compile::Context*)
	../../gcc/rust/backend/rust-compile.cc:41
0xc5c50b Rust::Session::compile_crate(char const*)
	../../gcc/rust/rust-session-manager.cc:733

```

</p>
</details>

### PR DESCRIPTION
**Pull request by tommyscholly**
This PR addresses #3618. With this change, the expected error of reaching the iteration limit is reported:

```
test.rs:1:17: error: ‘constexpr’ loop iteration count exceeds limit of 262144 (use ‘-fconstexpr-loop-limit=’ to increase the limit)
    1 | static _X: () = loop {};
      |                 ^~~~
```

My handling of `t` being a `NULL_TREE` may not be appropriate in `eval_constant_expression`, seeking guidance there.

Fixes #3618

### DIFF
--- a/gcc/rust/backend/rust-constexpr.cc
+++ b/gcc/rust/backend/rust-constexpr.cc
@@ -1901,6 +1901,9 @@ eval_constant_expression (const constexpr_ctx *ctx, tree t, bool lval,
 
   location_t loc = EXPR_LOCATION (t);
 
+  if (t == NULL_TREE)
+    return NULL_TREE;
+
   if (CONSTANT_CLASS_P (t))
     {
       if (TREE_OVERFLOW (t))
--- a/gcc/testsuite/rust/compile/issue-3618.rs
+++ b/gcc/testsuite/rust/compile/issue-3618.rs
@@ -1,0 +1,1 @@
+static _X: () = loop {}; // { dg-error "loop iteration count exceeds limit" }

### CODE REVIEW

**Review by philberty — CHANGES_REQUESTED** [review]
LGTM but i would slightly tweak this to be:

```c++
if (t == NULL_TREE)
  return NULL_TREE;
  
if (CONST_CLASS_P (t))
   ... <continue on with rest of code>
```

just means we have an explicit guard against null at the start. Just for reference this ia port from the gcc/cp/constexpr.cc implementation which fits well as it operates on GCC tree's

Also for your commit message please prefix with gccrs: so something like:

"gccrs: fix ICE on empty constexpr loops

Empty loops have no body which means this is a NULL_TREE during const
evaluation which needs a check.

Fixes Rust-GCC #  issue number

<changelog entry>
"

tommyscholly:
Implemented those changes!

philberty (approved):
LGTM

philberty (changes requested):
Actually sorry you need to add a test case too

philberty:
it should be enough to add a test case of :

```
static _X: () = loop {};
```

into a file in gcc/testsuite/rust/compile/issue-3618.rs

philberty:
Nice work but i just realised yes this need another change sorry. This null_tree check should still stay but you need another check the error should be:

```
error[E0658]: `loop` is not allowed in a `static`
 --> <source>:1:17
  |
1 | static _X: () = loop {};
  |                 ^^^^^^^
  |
  = note: see issue #52000 <https://github.com/rust-lang/rust/issues/52000> for more information

```

So what you need to do is add a catch here: https://github.com/Rust-GCC/gccrs/blob/fc6b54365731c2ebfa1974058c48772e03203fdd/gcc/rust/backend/rust-compile-expr.cc#L671

```
if (ctx->const_context_p ())
 {
     rich_location r;
     rust_error_at (r, ErrorCode::0658, "%<loop%> is not allowed in const context");
     return;
  }
```

tommyscholly:
Oh that's my bad. I was comparing with a 1.86 `rustc`. On version 1.49 it still gives a "exceeded interpreter step limit (see `#[const_eval_limit]`)" error. Should I not be basing functionality on 1.49?

philberty:
Oh good point whats really funny is if you test 1.45 you get loop is not allowed but in 1.49 you get the exceeded limit error.

uughh.. I think we should go both of these fixes in one commit. So we emit that error and if we need to we can remove it.

philberty:
Hmm maybe we should just open up an issue relating the what behaviour we want since your current PR is more like the latest rustc I bet they let loops return values using break etc now so yeah.. lets go with your patch then.

philberty (approved):
LGTM

tommyscholly:
Yeah, even on 1.49, this compiles fine actually:

```rs
static _X: () = loop {
    break;
};
```

PR #3786, opened 2025-05-09
Adapt attribute lang hook and do some cleanup

dkm:
Since r14-6076, the LANG_HOOKS_COMMON_ATTRIBUTE_TABLE should not be used and LANG_HOOKS_ATTRIBUTE_TABLE replaces it.

Add the "cold" attribute to fix PR120018 (and the matching "hot" attribute).

Replace some gcc_assert() by explicit warnings (same as D frontend).

Add some clang-format off/on comment around code that's not correctly
formatted.

### DIFF
--- a/gcc/rust/rust-attribs.cc
+++ b/gcc/rust/rust-attribs.cc
@@ -38,41 +38,43 @@ along with GCC; see the file COPYING3.  If not see
  * future.
  */
 
-extern const attribute_spec grs_langhook_common_attribute_table[];
+extern const struct scoped_attribute_specs grs_langhook_gnu_attribute_table;
+extern const struct scoped_attribute_specs grs_langhook_common_attribute_table;
+
+/* clang-format off */
+/* Disable clang-format because it insists in having the return type on a
+   single line (that's for definitions) */
 
 /* Internal attribute handlers for built-in functions.  */
-static tree
-handle_noreturn_attribute (tree *, tree, tree, int, bool *);
-static tree
-handle_leaf_attribute (tree *, tree, tree, int, bool *);
-static tree
-handle_const_attribute (tree *, tree, tree, int, bool *);
-static tree
-handle_malloc_attribute (tree *, tree, tree, int, bool *);
-static tree
-handle_pure_attribute (tree *, tree, tree, int, bool *);
-static tree
-handle_novops_attribute (tree *, tree, tree, int, bool *);
-static tree
-handle_nonnull_attribute (tree *, tree, tree, int, bool *);
-static tree
-handle_nothrow_attribute (tree *, tree, tree, int, bool *);
-static tree
-handle_type_generic_attribute (tree *, tree, tree, int, bool *);
-static tree
-handle_transaction_pure_attribute (tree *, tree, tree, int, bool *);
-static tree
-handle_returns_twice_attribute (tree *, tree, tree, int, bool *);
-static tree
-handle_fnspec_attribute (tree *, tree, tree, int, bool *);
-static tree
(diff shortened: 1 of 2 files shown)


dkm:
clang-format is not doing TRT here... The declarations should not have their return type on a single lines, only definitions should... Makes it easier to grep  `grep ^func_name`. If declaration are also indented like this, it doesn't work :'( (I'll fix the other issues)

philberty (approved):
Nice one! LGTM

gcc/rust/rust-attribs.cc
```diff
@@ -217,9 +242,14 @@ handle_malloc_attribute (tree *node, tree, tree, int, bool *)
...
 {
-  gcc_assert (TREE_CODE (*node) == FUNCTION_DECL);
+  if (TREE_CODE (*node) != FUNCTION_DECL)
+    {
+      warning (OPT_Wattributes, "%qE attribute ignored", name);
+      *no_add_attrs = true;
+    }
+
   DECL_PURE_P (*node) = 1;
   return NULL_TREE;
```
tschwinge, line 254: `DECL_PURE_P` does `FUNCTION_DECL_CHECK`, therefore is valid for `FUNCTION_DECL` only, but you're now permitting reaching it for other DECLs, too.  Either needs an early `return NULL`, or otherwise restructuring of the code.

Also, what about adding test cases for the respective cases?  :-)

gcc/rust/rust-attribs.cc
```diff
@@ -228,9 +258,14 @@ handle_pure_attribute (tree *node, tree, tree, int, bool *)
...
 {
-  gcc_assert (TREE_CODE (*node) == FUNCTION_DECL);
+  if (TREE_CODE (*node) != FUNCTION_DECL)
+    {
+      warning (OPT_Wattributes, "%qE attribute ignored", name);
+      *no_add_attrs = true;
+    }
+
   DECL_IS_NOVOPS (*node) = 1;
   return NULL_TREE;
```
tschwinge, line 270: Likewise for `DECL_IS_NOVOPS`.

gcc/rust/rust-attribs.cc
```diff
@@ -301,9 +336,14 @@ handle_nonnull_attribute (tree *node, tree, tree args, int, bool *)
...
 {
-  gcc_assert (TREE_CODE (*node) == FUNCTION_DECL);
+  if (TREE_CODE (*node) != FUNCTION_DECL)
+    {
+      warning (OPT_Wattributes, "%qE attribute ignored", name);
+      *no_add_attrs = true;
+    }
+
   TREE_NOTHROW (*node) = 1;
   return NULL_TREE;
```
tschwinge, line 348: Similarly, `TREE_NOTHROW` also isn't well-defined for all DECLs.

gcc/rust/rust-attribs.cc
```diff
@@ -339,9 +379,14 @@ handle_transaction_pure_attribute (tree *node, tree, tree, int, bool *)
...
+				bool *no_add_attrs)
 {
-  gcc_assert (TREE_CODE (*node) == FUNCTION_DECL);
+  if (TREE_CODE (*node) != FUNCTION_DECL)
+    {
+      warning (OPT_Wattributes, "%qE attribute ignored", name);
+      *no_add_attrs = true;
+    }
 
   DECL_IS_RETURNS_TWICE (*node) = 1;
```
tschwinge, line 391: `DECL_IS_RETURNS_TWICE` again does `FUNCTION_DECL_CHECK`.

tschwinge (commented):
Nice -- and a few items for follow-on work.  ;-)

PR #3787, opened 2025-05-09
Fix some duplicate node id issues

powerboat9:


philberty:
Hmm this smells like the a derive/desguar is cloning a node-id they should have unique ones. We've had these issues before early on.

philberty:
It would be better if we can drill down in on where these dups are comming from rather than add a pass to fix them

powerboat9:
It looks like the duplication is mostly coming from `clone_type` calls, and it seems like other callers depend on `clone_type` preserving node id

philberty:
Hmm yeah need to see why thats being called that rings a bell from a good while back... do you mind if i take a look into this too

philberty:
Oh wait i think i see what you mean a bit more...

philberty:
i found the fixes for this: https://github.com/Rust-GCC/gccrs/pull/3792

We actually have a few stray clones going on that need fixed. I think its better to finish filling out the builder than making a new pass to add new getters/setters on the AST.

P-E-P (changes requested):
I agree with @philberty, you're treating the symptoms rather than the disease. Those node ids shouldn't have been duplicated in the first place.

philberty:
closing this in favor of the other fix https://github.com/Rust-GCC/gccrs/pull/3792

PR #3795, opened 2025-05-13
Remove rvalue reference binding

powerboat9:
This should be correct, according to a careful reading of [cppreference.com](https://en.cppreference.com/w/cpp/language/return#Automatic_move_from_local_variables_and_parameters)


powerboat9:
Apparently there are some pitfalls with implicit move not occuring, but this PR should be in the clear

https://www.youtube.com/watch?v=hA1WNtNyNbo
https://en.cppreference.com/w/cpp/memory/unique_ptr/unique_ptr

philberty (approved):
LGTM

PR #3796, opened 2025-05-18
Adjust CI to ignore non-rust warnings

powerboat9:
This should make it much easier for us to maintain the expected warnings files


P-E-P:
I kinda dislike the idea of batch disabling warning but I fail to find any counter argument. It wouldn't trigger any warning on some rarely modified common files right ?

P-E-P:
> > It wouldn't trigger any warning on some rarely modified common files right ?
> 
> What do you mean?

Grepping rust will trigger only warning on rust-XXX.{cc,h} files or within the rust/libgrust directories, we rarely modify any files outside of those but this wouldn't be checked by the CI script right ?

powerboat9:
I wonder if having our CI use GCC 15 would help much

dkm:
We don't really care about warning outside of rust anyway. And from my recent experience of rebasing over upstream frequently, I can see how this change is needed as I've had to fixup the expected warning several times for stuff we haven't modified.

And you're correct, we should probably use latest GCC as the goal is to fix any issue the compiler may find, so the latest GCC should be more accurate than anything older...

P-E-P:
Well fine you convinced me, @powerboat9 could you bump the compiler version to gcc 15 ?

dkm:
What is missing for getting this one in? If possible, we should merge it before the next bump (I was willing to bump today or tomorrow)

.github/workflows/ccpp32alpine.yml
```diff
@@ -92,7 +93,6 @@ jobs:
     - name: Check for new warnings
       run: |
            cd gccrs-build
-           < log grep 'warning: ' | sort > log_warnings
```
dkm, line 95: Why is this one removed?

powerboat9: Looks to be accidental

powerboat9: Actually, nevermind -- looks like we've been generating `log_warnings` from `log` twice

powerboat9:
This PR should be done as soon as I can finish bugfixing it -- I had pivoted to trying to upgrade our CI to GCC 15, but it looks like that'll be a bit more involved

powerboat9:
TIL `grep` exits with `0` if it finds at least one match, `2` if it has an error, and `1` if it doesn't find any matches

PR #3797, opened 2025-05-19
ast: Check before visiting a while-let's label

CohenArthur:
gcc/rust/ChangeLog:

	* ast/rust-ast-visitor.cc (DefaultASTVisitor::visit): Check that the WhileLet has a label
	before visiting it.

gcc/testsuite/ChangeLog:

	* rust/compile/while_let_without_label.rs: New test.

### DIFF
--- a/gcc/rust/ast/rust-ast-visitor.cc
+++ b/gcc/rust/ast/rust-ast-visitor.cc
@@ -594,7 +594,10 @@ DefaultASTVisitor::visit (AST::WhileLetLoopExpr &expr)
   visit_outer_attrs (expr);
   for (auto &pattern : expr.get_patterns ())
     visit (pattern);
-  visit (expr.get_loop_label ());
+
+  if (expr.has_loop_label ())
+    visit (expr.get_loop_label ());
+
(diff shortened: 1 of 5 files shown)


powerboat9:
Looks like we're missing binding context handling for the `while let` patterns -- speaking of, are they supposed to have multiple patterns, or should they just store a pattern which could be an `AltPattern`?

https://doc.rust-lang.org/reference/expressions/loop-expr.html#predicate-pattern-loops

CohenArthur:
> Looks like we're missing binding context handling for the `while let` patterns -- speaking of, are they supposed to have multiple patterns, or should they just store a pattern which could be an `AltPattern`?
> 
> https://doc.rust-lang.org/reference/expressions/loop-expr.html#predicate-pattern-loops

I added a commit to handle while-let loops in NR2.0 in this PR, since we don't have the exclude file anymore - but I can split it in two PRs if you'd like @powerboat9.

regarding the vector of patterns vs alt-pattern - I'm not sure. this is how it's done in the base loop expression, but I think it could be changed to just one optional pattern (as loop loops don't have them)

PR #3798, opened 2025-05-19
nr2.0: Catch `Self` in impl block self types

powerboat9:


gcc/rust/resolve/rust-late-name-resolver-2.0.cc
```diff
@@ -478,6 +478,15 @@ Late::visit (AST::PathInExpression &expr)
 		 Definition (resolved->get_node_id ()));
 }
 
+void
+Late::visit_impl_type (AST::Type &type)
+{
+  bool old_block = block_big_self;
+  block_big_self = true;
+  visit (type);
+  block_big_self = old_block;
```
P-E-P, line 487: shoudnt we set the boolean back to false ?

powerboat9: I was thinking there might be an edge case where this would be reentrant, but in hindsight I'm not sure. Best to leave it for another PR, if it ever comes up.

gcc/rust/resolve/rust-late-name-resolver-2.0.cc
```diff
@@ -478,6 +478,16 @@ Late::visit (AST::PathInExpression &expr)
 		 Definition (resolved->get_node_id ()));
 }
 
+void
+Late::visit_impl_type (AST::Type &type)
+{
+  // TODO: does this have to handle reentrancy?
```
CohenArthur, line 484: I think it does right? or am I misunderstanding?

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=1e69ad4a9684e42ea15cafb1e9c34f74

powerboat9: Reentrency would in this case be something like `impl (... impl ... {} ...) {}` -- putting an impl item inside the description of the self type of another impl.

CohenArthur: ah, I think that's fine then

CohenArthur: the assertion you added saves us anyway in that case so I think we're good :)

CohenArthur (approved):
LGTM! thanks!

PR #3801, opened 2025-05-20
nr2.0: Run a final `TopLevel` pass after desugaring

powerboat9:


gcc/rust/rust-session-manager.cc
```diff
@@ -622,6 +622,14 @@ Session::compile_crate (const char *filename)
   AST::DesugarQuestionMark ().go (parsed_crate);
   AST::DesugarApit ().go (parsed_crate);
 
+  // HACK: run a final toplevel pass
+  // since desugaring may have added definitions
+  if (!saw_errors () && flag_name_resolution_2_0)
```
P-E-P, line 627: Shouldn't we also run early nr again since some of those definitions could technically be matched to some invocations ?

powerboat9: I was thinking we'd be done with early resolution before the desugaring and would only need to run `TopLevel` for the benefit of `Late` resolution.

CohenArthur:
I don't know if this is the best way to go about this, it's definitely hacky. maybe we could do all the desugars as part of the fixed point and mark the visitor as dirty when we find something to desugar

powerboat9:
@CohenArthur @P-E-P

CohenArthur:
> It looks like doing exactly that would screw up recursion_limit tracking

how do you mean? desugaring shouldn't count for the expansion limit, as we control the expansion (as opposed to a user-defined macro).

powerboat9:
> > It looks like doing exactly that would screw up recursion_limit tracking
> 
> how do you mean? desugaring shouldn't count for the expansion limit, as we control the expansion (as opposed to a user-defined macro).

The fixed point loop would either have to skip desugaring if we hit the expansion limit or run an extra iteration past the expansion limit.

CohenArthur:
I think this is not really an issue, if we hit the expansion limit we'll have an error anyway and will stop the pipeline. so the desugaring does not matter. I think the resulting code would be more correct

powerboat9:
> I think this is not really an issue, if we hit the expansion limit we'll have an error anyway and will stop the pipeline. so the desugaring does not matter. I think the resulting code would be more correct

I'm having trouble writing code that both includes desugaring in the main loop and that I can verify doesn't introduce an off-by-one error to `cfg.recursion_limit`. I've pushed what I have so far -- it *looks* correct, but I'm not sure its better than what I had before.

powerboat9:
@P-E-P this changeset?

PR #3804, opened 2025-05-23
parser: fix ICE std::out_of_range with path attrs to nonexisting path

vishruth-thimmaiah:
Stops an ICE from occuring when path attribute is empty

Fixes Rust-GCC#3607

gcc/rust/Changelog:
    * parse/rust-parse.cc (Rust::extract_module_path): Handle a case where the path is empty / has only whitespaces.

gcc/testsuite/Changelog:
    * rust/compile/torture/extern_mod2.rs: New tests to verify an error is thrown if the path is empty.

Signed-off by: vishruth-thimmaiah <vishruththimmaiah@gmail.com>

- \[x] Run `clang-format`

### DIFF
--- a/gcc/rust/parse/rust-parse.cc
+++ b/gcc/rust/parse/rust-parse.cc
@@ -42,8 +42,7 @@ extract_module_path (const AST::AttrVec &inner_attrs,
     {
       rust_error_at (
 	path_attr.get_locus (),
-	// Split the format string so that -Wformat-diag does not complain...
-	"path attributes must contain a filename: '%s'", "#![path = \"file\"]");
+	"path attributes must contain a filename: %<#[path = \"file\"]%>");
       return name;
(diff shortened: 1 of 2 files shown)


gcc/rust/parse/rust-parse.cc
```diff
@@ -42,8 +42,7 @@ extract_module_path (const AST::AttrVec &inner_attrs,
     {
       rust_error_at (
 	path_attr.get_locus (),
-	// Split the format string so that -Wformat-diag does not complain...
-	"path attributes must contain a filename: '%s'", "#![path = \"file\"]");
+	"path attributes must contain a filename: %<'#[path = \"file\"]'%>");
```
powerboat9, line 45: `%<` and `%>` print quotations, so you shouldn't need `'`

PR #3812, opened 2025-06-07
fix: ICE when parsing unterminated raw byte strings

vishruth-thimmaiah:
Fixes an ICE when a raw byte string is not terminated

Fixes Rust-GCC#3731

gcc/rust/ChangeLog:

	* lex/rust-lex.cc (Lexer::parse_raw_byte_string): Fix infinite looping when a raw byte string is not terminated.

gcc/testsuite/ChangeLog:

	* rust/compile/torture/unended-raw-byte-string.rs: New test to ensure correct error message for unended raw byte string.

- \[x] Run `clang-format`

### DIFF
--- a/gcc/rust/lex/rust-lex.cc
+++ b/gcc/rust/lex/rust-lex.cc
@@ -1894,17 +1894,17 @@ Lexer::parse_raw_byte_string (location_t loc)
 	      break;
 	    }
 	}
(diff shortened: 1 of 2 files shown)


CohenArthur (approved):
looks great, thank you for your work!! well done!

PR #3818, opened 2025-06-10
Add `derive(Ord)` and `derive(PartialOrd)`

CohenArthur:


philberty:
I'm working on your bracnh your code looks good i found 1 issue and fixed in a seperate PR already and i have found another but might take me a bit longer i think this is good to merge would be easier if this was merged for me

PR #3821, opened 2025-06-10
nr2.0: Adjust alternate identifier handling

powerboat9:


gcc/rust/resolve/rust-late-name-resolver-2.0.cc
```diff
@@ -236,8 +236,9 @@ visit_identifier_as_pattern (NameResolutionContext &ctx,
 
   if (ctx.bindings.peek ().is_or_bound (ident))
     {
-      // FIXME: map usage instead
-      std::ignore = ctx.values.insert_shadowable (ident, node_id);
+      auto res = ctx.values.get (ident);
+      rust_assert (res.has_value () && !res->is_ambiguous ());
+      ctx.map_usage (Usage (node_id), Definition (res->get_node_id ()));
```
CohenArthur, line 241: don't we need to sometimes insert a shadowable name when resolving identifier patterns? e.g. 

```rust
struct Foo(i32);

match foo {
    Foo(a) => { use(a) },
}
```

or am I misunderstanding?

powerboat9:
```text
let Foo(a | a) = x;
        ^   ^
        |   |_ 'a' is or-bound, map to last definition
        |_ 'a' is not or-bound, insert definition (see 'else' block)
```

CohenArthur: thank youuuu perfect haha

PR #3822, opened 2025-06-13
gccrs: Implement compilation of IdentifierPattern's subpattern bindings

Polygonalr:
Includes renaming of `get_pattern_to_bind` to `get_subpattern` which I missed out in my previous PR.

I'm aware that there are still other visit functions that lack support for `subpattern` - some changes for those are currently under my draft branch in my fork which I'll eventually get around to fully implementing and testing. For now, I think I want to move on and focus on implementing minimal support for other unsupported patterns first.

- \[x] Run `clang-format`

### DIFF
--- a/gcc/rust/backend/rust-compile-pattern.cc
+++ b/gcc/rust/backend/rust-compile-pattern.cc
@@ -666,6 +666,12 @@ CompilePatternBindings::visit (HIR::ReferencePattern &pattern)
 void
 CompilePatternBindings::visit (HIR::IdentifierPattern &pattern)
 {
+  if (pattern.has_subpattern ())
+    {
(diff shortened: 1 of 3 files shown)


gcc/rust/resolve/rust-late-name-resolver-2.0.cc
```diff
@@ -250,6 +250,11 @@ visit_identifier_as_pattern (NameResolutionContext &ctx,
 void
 Late::visit (AST::IdentifierPattern &identifier)
 {
+  if (identifier.has_subpattern ())
```
P-E-P, line 253: Do we really need a condition check over this ? Calling `DefaultResolver::visit(identifier)` directly would help future changes and let the condition in the default visitor.

PR #3826, opened 2025-06-15
nr2.0: Adjust resolution of modules

powerboat9:
This prioritizes resolution in the language prelude over resolution as a module.

### DIFF
--- a/gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
@@ -101,8 +101,6 @@ TopLevel::go (AST::Crate &crate)
 void
 TopLevel::visit (AST::Module &module)
 {
-  insert_or_error_out (module.get_name (), module, Namespace::Types);
-
   // Parse the module's items if they haven't been expanded and the file
   // should be parsed (i.e isn't hidden behind an untrue or impossible cfg
   // directive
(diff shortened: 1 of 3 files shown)


gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
```diff
@@ -101,8 +101,6 @@ TopLevel::go (AST::Crate &crate)
 void
 TopLevel::visit (AST::Module &module)
 {
-  insert_or_error_out (module.get_name (), module, Namespace::Types);
```
P-E-P, line 104: Can't wrap my mind around this line, why is it required to removed the module name from type namespace ?

powerboat9: To change the resolution order to `non-modules -> lang items -> modules`

P-E-P: I feel we're doing something wrong somewhere, rustc resolves modules first during top level and from what I heard struggles with their implementation. I'll need some time to think about that.

powerboat9: I'm seeing
```
            // In `a(::assoc_item)*` `a` cannot be a module. If `a` does resolve to a module we
            // don't report an error right away, but try to fallback to a primitive type.
            // So, we are still able to successfully resolve something like
            //
            // use std::u8; // bring module u8 in scope
            // fn f() -> u8 { // OK, resolves to primitive u8, not to std::u8
            //     u8::max_value() // OK, resolves to associated function <u8>::max_value,
            //                     // not to non-existent std::u8::max_value
            // }
            //
            // Such behavior is required for backward compatibility.
            // The same fallback is used when `a` resolves to nothing.
```
in the rustc source code (`rustc_resolve/src/late.rs`, 1.49)

PR #3828, opened 2025-06-16
nr2.0: Update IdentifierPattern's subpattern name resolution

Polygonalr:
Addresses [this comment](https://github.com/Rust-GCC/gccrs/pull/3822#discussion_r2149932447), thanks to @P-E-P for explaining the rationale for this change over meeting :)

- \[x] Run `clang-format`

### DIFF
--- a/gcc/rust/resolve/rust-late-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-late-name-resolver-2.0.cc
@@ -251,10 +251,7 @@ visit_identifier_as_pattern (NameResolutionContext &ctx,
 void
 Late::visit (AST::IdentifierPattern &identifier)
 {
-  if (identifier.has_subpattern ())
-    {
-      DefaultResolver::visit (identifier.get_subpattern ());
-    }
+  DefaultResolver::visit (identifier);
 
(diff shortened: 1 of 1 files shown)


powerboat9:
@P-E-P it looks like `IdentifierPattern::get_subpattern` doesn't return an optional and has the line

```rust
rust_assert (has_subpattern ());
```

P-E-P:
> @P-E-P it looks like `IdentifierPattern::get_subpattern` doesn't return an optional and has the line
> 
> ```rust
> rust_assert (has_subpattern ());
> ```

Right now the focus is the semantic aspect within the class ("yes that value can be missing and that null pointer IS a bug"), not at the class boundaries. This means you can keep the assert on the getter, if you really want to go the extra mile you can even rename it to `get_subpattern_unchecked` (it has been done in the codebase before, grep for `_unchecked` within the rust directory). If we have some time later we'll introduce proper optional APIs for the getters (and grepping "unchecked" will help) but for now the true goal is communicating the intent of the class to easily tell appart buggy null pointers from none values.

powerboat9:
Oh, wait, nvm -- I should have read the patch more closely. Looks good

PR #3829, opened 2025-06-16
nr2.0: Add more checks for alternate patterns

powerboat9:


gcc/rust/resolve/rust-late-name-resolver-2.0.cc
```diff
@@ -258,7 +258,9 @@ Late::visit (AST::IdentifierPattern &identifier)
 
   visit_identifier_as_pattern (ctx, identifier.get_ident (),
 			       identifier.get_locus (),
-			       identifier.get_node_id ());
+			       identifier.get_node_id (),
```
P-E-P, line 258: At this point we'd better pass the identifier down the function instead of it's fields separately.

powerboat9: That would require some modifications to `AST::StructPatternFieldIdent` -- which I suppose I could do

P-E-P: I've taken a closer look and I'm not sure it's worth the changes.

PR #3830, opened 2025-06-17
Adjust external crate lowering and type checking

powerboat9:
Name resolution 2.0 doesn't like crates being typechecked before the initialization of `ImmutableNameResolutionContext`

### DIFF
--- a/gcc/rust/hir/rust-ast-lower-item.cc
+++ b/gcc/rust/hir/rust-ast-lower-item.cc
@@ -732,6 +732,25 @@ ASTLoweringItem::visit (AST::MacroRulesDefinition &def)
   lower_macro_definition (def);
 }
 
+void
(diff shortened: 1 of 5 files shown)


powerboat9:
I thought about initializing `ImmutableNameResolutionContext` sooner, but I didn't think it would make sense to do so before the name resolution context is expected to become immutable.

philberty:
I think i need an example of the issue you were seeing that you needed this try to add a little more info into your commit messages its super important later on.

powerboat9:
> I think i need an example of the issue you were seeing that you needed this try to add a little more info into your commit messages its super important later on.

Fixed?

philberty (approved):
Ah ok i see what you mean now this looks kind of neat actually. LGTM

PR #3831, opened 2025-06-17
Fix execute testsuite

powerboat9:


philberty:
What is this change?

powerboat9:
Our non-torture execute tests weren't actually being run

philberty (approved):
LGTM

powerboat9:
@Polygonalr I had to move a test to `xfail`, since it looks like we can't parse `IdentifierPattern`s with subpatterns yet.

PR #3839, opened 2025-06-20
nr2.0: Adjust resolution of external crates

powerboat9:


philberty (commented):
I think this looks mostly fine but i think your patch needs to have a line talking about why your making this change. Its usually grand to omit this when more tests have been added or something but its not quite as clear here why this patch was made.

PR #3840, opened 2025-06-20
nr2.0: Enable by default

powerboat9:
This depends on a few other PRs, but when merged should switch us over to nr2.0 by default


powerboat9:
@P-E-P @CohenArthur @philberty This should be ready to merge

philberty:
This is kind of exciting and very scary all at the same time I am happy with it but this needs to come from @CohenArthur to merge it.

Once its merged we really need a MR to remove the old one having all that old code hanging around will be really annoying.

powerboat9:
Thank you

P-E-P (approved):

PR #3841, opened 2025-06-20
nr2.0: Fix resolution of constant items

powerboat9:
#3834
> Nr2 does handle this and puts out an extra good diagnostic

🙃

### DIFF
--- a/gcc/rust/resolve/rust-default-resolver.cc
+++ b/gcc/rust/resolve/rust-default-resolver.cc
@@ -361,19 +361,15 @@ DefaultResolver::visit (AST::MatchExpr &expr)
 void
 DefaultResolver::visit (AST::ConstantItem &item)
 {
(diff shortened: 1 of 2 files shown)


gcc/rust/resolve/rust-default-resolver.cc
```diff
@@ -361,19 +361,15 @@ DefaultResolver::visit (AST::MatchExpr &expr)
 void
 DefaultResolver::visit (AST::ConstantItem &item)
 {
-  if (item.has_expr ())
```
philberty, line 364: dont you still need a has_expr check?

powerboat9: That's the thing -- `DefaultASTVisitor` handles that

P-E-P (approved):

PR #3849, opened 2025-06-21
gccrs: Add unify rules for fnptr and closures

philberty:
Its valid to unify a closure to an fnptr as we are working on the fn traits. There are still other issues but this is part of the patch set.

gcc/rust/ChangeLog:

	* typecheck/rust-unify.cc (UnifyRules::expect_fnptr): add unify rules


powerboat9:
Only if it has no captures, I think

powerboat9:
Have you updated your `.clang-format` file?

PR #3850, opened 2025-06-22
gccrs: Fix TupleStructPattern compilation throwing error

Polygonalr:
Compilation of non-enum `TupleStructPattern` will result in an error, this PR fixes it.

- \[x] Run `clang-format`


philberty (changes requested):
code LGTM but can you add an extra line in your commit message describing the patch here. One line is enough.

philberty (approved):
LGTM nice work

PR #3308, opened 2024-12-17
[WIP] Explicitly install gettext for macos CI

powerboat9:
This might fix the macos CI

PR #3851, opened 2025-06-22
Reenable macos CI

powerboat9:
Originally disabled in #3312

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -368,64 +368,66 @@ jobs:
               exit 0; \
             fi
 
-  # build-and-check-clang-macos:
+  build-and-check-clang-macos:
 
-  #   env:
-  #     # Force CC/CXX to be explicitly clang to make it clear which compiler is used
-  #     CC: clang
-  #     CXX: clang++
+    env:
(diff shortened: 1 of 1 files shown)


P-E-P:
Do we know what changed in between ?

philberty:
Oh interesting so the only failure is:

FAIL: rust/compile/macros/builtin/recurse2.rs scan-assembler "abheyho"

iirc that testcase relies on scan assembler for that maybe its not totally great on macos, can you just add a dg-skip on this for macos and get this turned back on

powerboat9:
> Do we know what changed in between ?

I'd guess this was fixed upstream -- we were getting https://github.com/Rust-GCC/gccrs/pull/3308#issuecomment-2551034499

powerboat9:
> Oh interesting so the only failure is:
> 
> FAIL: rust/compile/macros/builtin/recurse2.rs scan-assembler "abheyho"
> 
> iirc that testcase relies on scan assembler for that maybe its not totally great on macos, can you just add a dg-skip on this for macos and get this turned back on

I'll see if I can fix the test before I disable it -- I think I'm on the right track

powerboat9:
Fixed, macos seems to output the assembly directive `.ascii` (not automatically zero terminated) rather than `.string` (automatically zero terminated)

PR #3856, opened 2025-06-24
Update C++ version check in rust-lang.cc

powerboat9:
I don't know if this version check is redundant, but it probably shouldn't be *this* redundant


philberty:
Will leave it up to @dkm not sure what the c++ version is allowed in gcc at the moment

P-E-P:
> not sure what the c++ version is allowed in gcc at the moment

C++14 is now the baseline requirement in GCC.

PR #3857, opened 2025-06-24
gccrs: Add getter for non const lifetime object

sakupan102:
Since `SelfParam` has a lifetime but there was no way to obtain it as a non-const object, I added a method to do so.

```
gcc/rust/ChangeLog:

	* hir/tree/rust-hir-item.h: Add getter for non const lifetime object
```

- \[x] Run `clang-format`

order to get feedback.
---


dkm:
The changelog should point at the item being modified, not only the file. Something like:
```
	* hir/tree/rust-hir-item.h  (SelfParam::get_lifetime): Add getter for non const lifetime object.
```

philberty (approved):
LGTM

PR #3860, opened 2025-06-26
Bump test.

dkm:


powerboat9:
Could we just do a regular merge? It looks like it'd go through with no conflicts

dkm:
> Could we just do a regular merge? It looks like it'd go through with no conflicts

Sounds like we've already discussed all this when we decided how to proceed. At least here is public and other interested parties could chime in (e.g. @thesamesam ).

I would like to use the process as discussed, at least for a few rounds. We can always adjust if something is painful, not friendly, anything else.

The good property of doing this is that the main branch is always mostly ready for the upstreaming. The main branch is always: a few gccrs commits, an empty merge commit with the `--strategy=ours` which first parent is a sequence of gccrs commits rebased over an upstream gcc base (recent). The git history becomes a bit like the history of the gccrs commit rebased: first parent of the merge is the current series, the second parent is the previous branch (from there, you again have the exact same pattern).

The daily routine is rather easy and doesn't involve to be a git expert to do. See for example my WIP for gerris... 

https://github.com/Rust-GCC/gerris/blob/dkm/upstream-command/src/rebaseupstream.rs => this is used to rebase the github branch onto a fresh upstream (it rebases onto a new base, and it moves all the commits that were after the commit before the new one, so that you have: merge<-gccrs sequence<-fresh upstream gcc).

https://github.com/Rust-GCC/gerris/blob/11b5ee7372da00a40a493a88a53a3530cd81d38b/src/upstream.rs#L249 => this is used to prepare the branch before upstreaming. It is not 100% done, but it gives a good start.

I fear that having regular merge would make things harder to read, and the upstreaming would need a different process. Again, I'm not saying it's not a good idea, but I think I like the process as proposed initially and would like to give it a try.

powerboat9:
Alright, we can table the discussion for now. I figured I'd bring it up again since #3761 has already been merged.

dkm:
> Alright, we can table the discussion for now. I figured I'd bring it up again since #3761 has already been merged.

The goal, at least as I understand it, is to do similar merge on a weekly basis, with corresponding upstreaming...

dkm:
Not sure why there's an extra warning here, but I have a fix for it that I'll submit later (will be afk until monday)

powerboat9:
Oddly enough, it's not an extra warning, it's a missing warning

dkm:
> Oddly enough, it's not an extra warning, it's a missing warning

We probably need to read what are the warnings and get rid of them if needed.
I do have the warning locally on master and I have a fix... Not sure why it's missing I'll have a better look later.

dkm:
@CohenArthur I think, if the CI is :heavy_check_mark: , the branch can be pushed in master. This will sync with GCC and allow easier upstreaming

dkm:
> @CohenArthur I think, if the CI is ✔️ , the branch can be pushed in master. This will sync with GCC and allow easier upstreaming

of course the changelog check is not happy because of old commits, but none of them is meant to be upstreamed (github infra things, or moving readme around)

dkm:
@CohenArthur this is ready to be pushed in master branch (if all CI jobs are :heavy_check_mark: , except the git log checks)

powerboat9:
`git diff` with the result of a merge between https://github.com/Rust-GCC/gccrs/commit/3fbed69502770851c8a5bfece99e9ee8c6d9e4b0 and https://github.com/Rust-GCC/gccrs/commit/5dc8c4d20d62462ebce137c6e5146235ab9e9e2e is empty, looks good to merge

dkm:
CI is happy, everything looks good :heavy_check_mark:

PR #3862, opened 2025-06-28
Create `LocalVariable`

powerboat9:
This should make it easier for us to move away from leaking pointers to `Bvariable` everywhere. Since `LocalVariable` has a single field of type tree, it should be the same size as a pointer to `Bvariable`, making the switch to `LocalVariable` wherever possible strictly an improvement.

### DIFF
--- a/gcc/rust/rust-gcc.cc
+++ b/gcc/rust/rust-gcc.cc
@@ -83,6 +83,23 @@ Bvariable::error_variable ()
   return new Bvariable (error_mark_node);
 }
 
+// Get the tree of a variable for use as an expression
+tree
+LocalVariable::get_tree (location_t location) const
+{
+  if (error_operand_p (t))
+    return error_mark_node;
+
+  TREE_USED (t) = 1;
+  return t;
+}
+
+LocalVariable
+LocalVariable::error_variable ()
+{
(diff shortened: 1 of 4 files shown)


powerboat9:
I have a patch which converts `Bvariable *` to `Bvariable`/`tl::optional<Bvariable>` ~everywhere, but it touches a lot of files and breaks something, so I figure I'll start small.

gcc/rust/rust-gcc.cc
```diff
@@ -2057,7 +2074,7 @@ temporary_variable (tree fndecl, tree bind_tree, tree type_tree, tree init_tree,
       || error_operand_p (fndecl))
     {
       *pstatement = error_mark_node;
-      return Bvariable::error_variable ();
+      return LocalVariable::error_variable ();
```
P-E-P, line 2077: Not a fan of error state hidden within a single class. Couldn't we return an `Expected<LocalVariable>` ?

powerboat9: Probably, though I'd like to leave that for another PR

powerboat9:
I'd think it would make sense to merge this as-is, as it seems like a good stepping stone towards removing `Bvariable` and addresses some memory leaks.

PR #3863, opened 2025-06-29
gccrs: Implement compilation support for TuplePatternItems::RANGED

Polygonalr:
This is part of implementing support for RestPatterns, RestPattern support for slices will be next.

Includes a fix to my little oopsie made in my type-checking code, and test updates to ensure that the mistake will not repeat.


gcc/testsuite/rust/execute/torture/match-restpattern-tuple.rs
```diff
@@ -0,0 +1,13 @@
+fn main() -> i32 {
```
philberty, line 6: can you copy some of the other test cases to use a dg-output comment at the top and use a call to puts or printf to show that its outputing what you expect?

Oh neat i see you are using the ret variable to do this to make sure we return 0 i think it would be good to also add in a puts call here too

gcc/rust/backend/rust-compile-pattern.cc
```diff
@@ -430,8 +430,50 @@ CompilePatternCheckExpr::visit (HIR::TuplePattern &pattern)
...
+	for (auto &pat : items.get_lower_patterns ())
+	  {
+	    tree field_expr
+	      = Backend::struct_field_expression (match_scrutinee_expr,
+						  tuple_field_index++,
+						  pat->get_locus ());
+
+	    tree check_expr_sub
+	      = CompilePatternCheckExpr::Compile (*pat, field_expr, ctx);
+	    check_expr = Backend::arithmetic_or_logical_expression (
```
philberty, line 454: can you share what the gimple output is for some of these tests? Possibly show an example gimple in the commit message too is a good idea.

philberty (changes requested):
Looks great but just minor comment on the test case and we need to see what the gimple is being generated here would help the review.

dkm:
@Polygonalr FYI, no need to add extra indentation in the changelog:
```
<TAB>* file.cc (item): bla bla
<TAB>bla bla bla.
```
not
```
<TAB>* file.cc (item): bla bla
<TAB><TAB>bla bla bla.
```

PR #3871, opened 2025-07-04
Load modules during `CfgStrip` phase

powerboat9:
`TopLevel` would ignore just-loaded modules but `Early` and `ExpandVisitor` wouldn't. The latter would produce errors when it hit attributes which should have been indirectly `CfgStrip`'d away.

### DIFF
--- a/gcc/rust/expand/rust-cfg-strip.cc
+++ b/gcc/rust/expand/rust-cfg-strip.cc
@@ -1765,16 +1765,17 @@ CfgStrip::visit (AST::Module &module)
       return;
     }
 
-  // A loaded module might have inner attributes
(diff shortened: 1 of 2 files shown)


CohenArthur (changes requested):
I think this causes issues with modules being loaded when they should be cfg-stripped, e.g

```rust
#[cfg(target_arch = "not_my_stuff")]
mod foo;
```

when running the compiler on `core` with this change, we error out much earlier than without, and on a module specifically for the `x86` arch.

CohenArthur:
nevermind my comment - the order just changed and I'm hitting the same error now that I'm progressing on name resolution. so I think this change is probably fine

PR #3884, opened 2025-07-08
attributes: Add #[test] and #[simd_test]

CohenArthur:
gcc/rust/ChangeLog:

	* util/rust-attribute-values.h: Add declarations for them.
	* util/rust-attributes.cc: Add definitions.


powerboat9:
I *think* at least `simd_test` is only used because of a bug #3871 fixes -- it should be disabled by a `#[cfg]`

powerboat9:
I think `simd_test` is a proc macro

PR #3890, opened 2025-07-08
Use `MacroInvocLexer` in `AttributeParser`

powerboat9:
This should make it easier for us to handle attribute meta items of the form `<SimplePath> '=' <Expression>` where the expression isn't a literal. Some low hanging fruit remains here, but I think I should keep this patch small as I had some trouble debugging it as-is (see: `Rust::Token::as_string` vs `Rust::Token::get_str` vs `Rust::AST::Token::as_string`).


philberty (approved):
LGTM

CohenArthur:
aaaah that's amazing, thank you

CohenArthur:
Does it fix #3897 ?

PR #3889, opened 2025-07-08
gccrs: Do proper const folding during typechecking for array capacities

philberty:
This patch adds proper folding to the const expression for array capacity we already have the const folding mechanics and the query system needed to handle cases where the capacity is a function call in a const context. This leverages and pulls the gcc tree capacity into the TyTy::ArrayType so it can be used for more typechecking and eventually doing more const generics work.

Addresses Rust-GCC#3885

gcc/rust/ChangeLog:

	* backend/rust-compile-base.cc (HIRCompileBase::query_compile_const_expr): new wrapper
	* backend/rust-compile-base.h: add prototype
	* backend/rust-compile-context.cc (Context::get): singleton helper
	* backend/rust-compile-context.h: likewise
	* backend/rust-compile-type.cc (TyTyResolveCompile::visit): handle infer's that can default
	* rust-session-manager.cc (Session::compile_crate): create the gcc context earlier for tychk
	* typecheck/rust-hir-type-check-base.cc (TypeCheckBase::resolve_literal): const fold it
	* typecheck/rust-hir-type-check-expr.cc (TypeCheckExpr::visit): likewise
	* typecheck/rust-hir-type-check-type.cc (TypeCheckType::visit): likewise
	* typecheck/rust-tyty.cc (BaseType::monomorphized_clone): fix constructor call
	(ArrayType::as_string): print capacity
	(ArrayType::clone): fix constructor call
	* typecheck/rust-tyty.h: track capacity
	* typecheck/rust-unify.cc (UnifyRules::expect_array): check the capacities

gcc/testsuite/ChangeLog:

	* rust/compile/all-cast.rs: shows array capacity now
	* rust/compile/arrays2.rs: likewise
	* rust/compile/const3.rs: fix error message
	* rust/compile/const_generics_3.rs: disable until typecheck we get proper errors now!
	* rust/compile/usize1.rs: proper capacity error message

PR #3913, opened 2025-07-11
gccrs: Add size checking to SlicePattern

Polygonalr:
Previously not possible without const folding of array capacities. Since that issue has been resolved with #3889 (Thanks @philberty!), it is now possible to check whether the capacity of the array parent matches the size of a `SlicePattern` during compile time.

This is part of implementing support for SlicePattern, compilation of the pattern will be next.


philberty (approved):
LGTM

PR #3979, opened 2025-07-15
Improve parsing of simple paths

powerboat9:
I noticed this earlier, while trying to reduce code duplication between `AttributeParser` and `Parser`


philberty (approved):
LGTM

PR #3980, opened 2025-07-16
gccrs: Reject loop in const/static context

### LINKED ISSUE #3783: Rustc behaviour with loops in const/static context
Nice work but i just realised yes this need another change sorry. This null_tree check should still stay but you need another check the error should be:

```
error[E0658]: `loop` is not allowed in a `static`
 --> <source>:1:17
  |
1 | static _X: () = loop {};
  |                 ^^^^^^^
  |
  = note: see issue #52000 <https://github.com/rust-lang/rust/issues/52000> for more information

```

So what you need to do is add a catch here: https://github.com/Rust-GCC/gccrs/blob/fc6b54365731c2ebfa1974058c48772e03203fdd/gcc/rust/backend/rust-compile-expr.cc#L671

```
if (ctx->const_context_p ())
 {
     rich_location r;
     rust_error_at (r, ErrorCode::0658, "%<loop%> is not allowed in const context");
     return;
  }
```

_Originally posted by @philberty in https://github.com/Rust-GCC/gccrs/issues/3781#issuecomment-2862764681_

Lishin1215:
Fixes #3783 .

gcc/rust/ChangeLog:
    
            * backend/rust-compile-expr.cc (CompileExpr::visit): add a catch for const/static
    
gcc/testsuite/ChangeLog:
    
            * rust/compile/loop_constant_context.rs: New test.


Lishin1215:
In this patch I use `ctx->const_context_p()` to block `loop` in `const`/`static`,
and using `!DECL_DECLARED_CONSTEXPR_P(fnctx.fndecl)` to allow `loop` inside `const fn`. (like [this example](https://github.com/Rust-GCC/gccrs/blob/5dc8c4d20d62462ebce137c6e5146235ab9e9e2e/gcc/testsuite/rust/compile/const9.rs) )

However, this may still reject the valid case mentioned here [here](https://github.com/Rust-GCC/gccrs/pull/3781#issuecomment-2863542416)
```
static _X: () = loop {
    break;
};
```
Thanks!

philberty:
Good catch i raised that test case as an issue. I think lets go for your Patch and merge it for now. That test case we can handle because we could add in a check for the result of the loop if its a ! never type or () i bet we just ignore it

PR #3981, opened 2025-07-17
gccrs: Implement compilation for SlicePattern matching against ArrayType scrutinee

Polygonalr:
Only `ArrayType` scrutinee is supported, not too sure how to implement it for `SliceType` scrutinee, as `HIR::SliceType` does not seem to have enough information for me to implement the appropriate checks. Would appreciate some guidance for it!


powerboat9:
Slice types are unsized arrays -- so all references to them (ex: `&[u8]`) are wide pointers (see: https://doc.rust-lang.org/nomicon/exotic-sizes.html)

philberty:
@Polygonalr i think we have all the bits we need let me start a thread on zulip

philberty (approved):
LGTM good work

philberty:
I've given you notes about slices in zulip

PR #3445, opened 2025-02-20
Fix expansion of macros inside modules

powerboat9:
This should fix #3444

PR #3990, opened 2025-07-19
Use X macros to simplify `ASTVisitor` inheritance

powerboat9:
`ASTVisitor` itself is left as-is for now, as a sanity check against the contents of `rust-ast-visitable.h`. This should slightly reduce the amount of repetition involved with inheriting from `ASTVisitor`, which might help us deal with the footgun that is `ExpandVisitor` later.

Any ideas for generally solving `ExpandVisitor` problems à la #3445? I suppose having `ExpandVisitor` inherit from `ASTVisitor` instead of `DefaultASTVisitor` might help, but it doesn't seem ideal. Maybe transforming `DefaultASTVisitor` into some kind of complex template so it can be specialized to operate as an in-place visitor?


powerboat9:
I came up with something awesome (in the archaic sense)

Edit: this was referring to some template metaprogramming which didn't pan out

CohenArthur (approved):
at the same time I really like it and really hate it lmao. I think this is good for when we add new nodes, it will save a lot of time

PR #3995, opened 2025-07-21
Use X macros to generate `DefaultASTVisitor`

powerboat9:
This is a bit messy, but basing default AST visitation on a descriptive DSL is probably one of our best options for regularizing AST visitors. `clang-format` is disabled in `ast/rust-ast-visitable-inner.h`, since I'm not sure how to get it to handle the DSL well.


powerboat9:
It looks like https://clang.llvm.org/docs/ClangFormatStyleOptions.html#macros would be useful, whenever we bump clang-format to version 17

PR #3997, opened 2025-07-22
Handle easy try-block desugar

CohenArthur:
This PR adds a desugar dispatcher for expressions, called expression yeast (since yeast consumes all the sugar...) and a desugar for try-block in the easy case, which is the most common in `core`. Basically, it can handle try-blocks with a tail expression and no statements - as that requires block labels to be handled properly


CohenArthur:
Actually, this is enough for all of `core` - the crate only uses the form `try { <expr> }`

PR #3998, opened 2025-07-22
gccrs: Implement compilation for SlicePattern matching against SliceType scrutinee

Polygonalr:
Figured out how to use the `debug` function in `rust.gcc.cc`.

Looking for feedback on my implementation for `slice_data_index_expression`, I am unsure whether it's the best way to do so. (should I make it such that the 1st parameter is the actual slice tree and not the data field?)

Supporting rest patterns `[..]` in slice patterns will be my next task.

### DIFF
--- a/gcc/rust/backend/rust-compile-pattern.cc
+++ b/gcc/rust/backend/rust-compile-pattern.cc
@@ -529,7 +529,8 @@ CompilePatternCheckExpr::visit (HIR::SlicePattern &pattern)
   // pattern must either be ArrayType or SliceType, should be already confirmed
   // by type checking
   rust_assert (lookup->get_kind () == TyTy::TypeKind::ARRAY
-	       || lookup->get_kind () == TyTy::TypeKind::SLICE);
+	       || lookup->get_kind () == TyTy::TypeKind::SLICE
+	       || lookup->get_kind () == TyTy::REF);
 
   size_t array_element_index = 0;
   switch (lookup->get_kind ())
@@ -895,7 +896,8 @@ CompilePatternBindings::visit (HIR::SlicePattern &pattern)
   rust_assert (ok);
 
   rust_assert (lookup->get_kind () == TyTy::TypeKind::ARRAY
-	       || lookup->get_kind () == TyTy::TypeKind::SLICE);
+	       || lookup->get_kind () == TyTy::TypeKind::SLICE
+	       || lookup->get_kind () == TyTy::REF);
 
   size_t array_element_index = 0;
   switch (lookup->get_kind ())
--- a/gcc/rust/backend/rust-compile-pattern.cc
+++ b/gcc/rust/backend/rust-compile-pattern.cc
@@ -555,7 +555,38 @@ CompilePatternCheckExpr::visit (HIR::SlicePattern &pattern)
     case TyTy::TypeKind::SLICE:
       rust_sorry_at (
 	pattern.get_locus (),
-	"SlicePattern matching against slices are not yet supported");
+	"SlicePattern matching against non-ref slices are not yet supported");
+      break;
+    case TyTy::TypeKind::REF:
+      {
+	rust_assert (RS_DST_FLAG_P (TREE_TYPE (match_scrutinee_expr)));
+	tree size_field
+	  = Backend::struct_field_expression (match_scrutinee_expr, 1,
+					      pattern.get_locus ());
+
+	// First compare the size
+	check_expr = Backend::comparison_expression (
+	  ComparisonOperator::EQUAL, size_field,
+	  build_int_cst (size_type_node, pattern.get_items ().size ()),
+	  pattern.get_locus ());
+
+	// Then compare each element in the slice pattern
+	for (auto &pattern_member : pattern.get_items ())
+	  {
+	    tree slice_index_tree
+	      = Backend::size_constant_expression (array_element_index++);
+	    tree element_expr
+	      = Backend::slice_index_expression (match_scrutinee_expr,
+						 slice_index_tree,
+						 pattern.get_locus ());
+	    tree check_expr_sub
+	      = CompilePatternCheckExpr::Compile (*pattern_member, element_expr,
+						  ctx);
+	    check_expr = Backend::arithmetic_or_logical_expression (
+	      ArithmeticOrLogicalOperator::BITWISE_AND, check_expr,
+	      check_expr_sub, pattern.get_locus ());
+	  }
+      }
       break;
     default:
       rust_unreachable ();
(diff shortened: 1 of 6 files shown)


gcc/rust/rust-gcc.cc
```diff
@@ -25,35 +25,35 @@
...
 #include "rust-backend.h"
-#include "rust-object-export.h"
 #include "rust-gcc.h"
+#include "rust-linemap.h"
+#include "rust-location.h"
+#include "rust-object-export.h"
 
-#include "backend/rust-tree.h"
 #include "backend/rust-builtins.h"
+#include "backend/rust-tree.h"
```
powerboat9: Is this necessary?

gcc/rust/typecheck/rust-hir-type-check-pattern.cc
```diff
@@ -675,8 +675,22 @@ TypeCheckPattern::visit (HIR::SlicePattern &pattern)
...
-	auto &array_ty_ty = static_cast<TyTy::SliceType &> (*parent);
-	parent_element_ty = array_ty_ty.get_element_type ();
+	auto &slice_ty_ty = static_cast<TyTy::SliceType &> (*parent);
+	parent_element_ty = slice_ty_ty.get_element_type ();
+	break;
+      }
+    case TyTy::REF:
+      {
+	auto &ref_ty_ty = static_cast<TyTy::ReferenceType &> (*parent);
+	TyTy::BaseType *deref = ref_ty_ty.get_base ();
```
philberty: you can use:

```c++
TyTy::SliceType* slice = nullptr;
if (!ref_ty_ty->is_dyn_slice_type (&slice))
{
  rust_error_at (...);
  return;
}

parent_element_Ty = slice->get_element_type ();
```

gcc/rust/rust-gcc.cc
philberty: yeah these include changes look sus

gcc/rust/rust-gcc.cc
```diff
@@ -1504,6 +1504,25 @@ array_index_expression (tree array_tree, tree index_tree, location_t location)
...
 }
 
+// Return an expression representing SLICE[INDEX]
+
+tree
+slice_data_index_expression (tree slice_data_ref_tree, tree index_tree,
+			     location_t location)
+{
+  tree element_type = TREE_TYPE (slice_data_ref_tree);
+  tree element_size = TYPE_SIZE_UNIT (element_type);
```
philberty: I think a better way to do this is:

```c++
tree
slice_index_expression (tree slice, tree index, location_t locus)
{
  if (error_operand_p (slice) || error_operand_p (index))
    return error_mark_node;

  // a slice is created in TyTyResolvecompile::create_slice_type_record
  //
  // for example:
  //   &[i32] is turned directly into a struct { i32* data, usize len };
  //   [i32] is also turned into struct { i32* data, usize len }

  rust_assert (RS_DST_FLAG_P (TREE_TYPE (slice)));
  tree data_pointer = struct_field_expression (slice, 0, locus);
  tree data_pointer_type = TREE_TYPE (data_pointer);
  rust_assert (POINTER_TYPE_P (TREE_TYPE (data_pointer)));

  tree data_offset_expr
    = Rust::pointer_offset_expression (data_pointer, index, locus);

  return build1_loc (locus, INDIRECT_REF, TREE_TYPE (data_pointer_type),
		     data_offset_expr);
}
```

gcc/rust/backend/rust-compile-pattern.cc
```diff
@@ -915,8 +951,26 @@ CompilePatternBindings::visit (HIR::SlicePattern &pattern)
     case TyTy::TypeKind::SLICE:
       rust_sorry_at (
 	pattern.get_locus (),
-	"SlicePattern matching against slices are not yet supported");
+	"SlicePattern matching against non-ref slices are not yet supported");
       break;
+    case TyTy::TypeKind::REF:
+      {
+	tree data_field
```
philberty: i dont think you need these now if you just use 

slice_index_expression (match_scruitinee, expr, index, locus)

gcc/rust/backend/rust-compile-pattern.cc
```diff
@@ -915,8 +951,26 @@ CompilePatternBindings::visit (HIR::SlicePattern &pattern)
     case TyTy::TypeKind::SLICE:
```
philberty, line 948: Not sure if you actually need this case here but maybe just leave this rust_sorry_at for now and we will see

gcc/rust/backend/rust-compile-pattern.cc
```diff
@@ -554,7 +555,41 @@ CompilePatternCheckExpr::visit (HIR::SlicePattern &pattern)
...
+					      pattern.get_locus ());
+
+	// First compare the size
+	check_expr = Backend::comparison_expression (
+	  ComparisonOperator::EQUAL, size_field,
+	  build_int_cst (size_type_node, pattern.get_items ().size ()),
+	  pattern.get_locus ());
+
+	// Then compare each element in the slice pattern
+	tree derefed = indirect_expression (data_field, pattern.get_locus ());
```
philberty: I think we dont need the indirect_expression if you use the slice_index_expression function with the changes i suggested.

philberty (changes requested):
Really good job navigating all of this. We need this slice_index_expression in other places so thanks for doing it.

I made an example slice_index_expr in the comments there that should work but if not let me know.

I think it simplifies the logic you need to do in these handlers.

philberty:
DST's are confusing because in terms for Rust code they look like references but under the hood they are a struct to GCC.

gcc/rust/rust-gcc.cc
Polygonalr: Sorry, it was late at night when I pushed this PR and I didn't check through the commit properly. Think my clang-format config messed up and the erroneous changes were committed by accident, will fix it in another push.

powerboat9: It's alright, I had two bugs in https://github.com/Rust-GCC/gccrs/pull/3995/files#diff-91bba4c0349ca7467edd6345c7b685a92ec12b5fd720c8e2531bdafb1e055fbaR315-R325 because it was where I stopped working and went to bed.

gcc/rust/typecheck/rust-hir-type-check-pattern.cc
Polygonalr: ~~The actual code would turn out a little weird, considering `is_dyn_slice_type` takes in `const TyTy::SliceType **slice` as parameter. I don't think doing `const_cast` like below is tidy.~~

```cpp
TyTy::SliceType *slice = nullptr;
if (!ref_ty_ty.is_dyn_slice_type (
      const_cast<const TyTy::SliceType **> (&slice)))
  {
    rust_error_at (pattern.get_locus (), "expected %s, found slice",
		   parent->as_string ().c_str ());
    return;
  }
```

Had a brain fart, I'll just declare the slice pointer as `const`.

dkm:
Side note: don't add extra indentation in the changelog part.. keep a single tab :)

PR #4001, opened 2025-07-23
ci: Bump from `macos-13` to `macos-15`

powerboat9:
https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down


philberty (changes requested):
ideally we should use a tag if he has one. Or we should fork his repo and tag it ourselves to avoid intermittent failures.

Issue #3915, opened 2025-07-12
Bad handling of path expression during monomoprhization


philberty:
### Summary

This bug is hidden most of the time but we get away with it in simple test cases so technically this test case "works" but only by accident.

### Reproducer

I tried this code:

```rust
#[lang = "sized"]
trait Sized {}

trait Trait {
    fn do_thing();
}

struct MyType;

impl Trait for MyType {
    fn do_thing() {}
}

struct Wrapper<T: Trait> {
    value: T,
}

impl<T: Trait> Wrapper<T> {
    fn call_it() {
        T::do_thing(); // <---------- this only works by accident... 'T' isnt getting monomorphized to MyType here
    }
}

fn main() {
    let _ = Wrapper::<MyType> { value: MyType };
    Wrapper::<MyType>::call_it();
}

```

### Does the code make use of any (1.49) nightly feature ?

- [ ] Nightly

### Actual behavior

The current behavior is, it doesnt monomoprhize but there is only 1 impl anyway and it defaults

### Expected behavior

I expected to see the type get monomorphized in the logs properly

### GCC Version

110783acbdcac12e821d4d0a7b4bbec0f13c22e1

Issue #1247, opened 2022-05-12
libcore SIP hasher


philberty:
This is an interesting testcase thanks to @bjorn3

Related Bugs: 

- [x]  #1048 
- [x] #1125 
- [x] #1126 
- [x] #1127 
- [x] #1237
- [x] #1006
- [x] #1128 
- [x] #1129 
- [x] #1130 
- [x] #1131 
- [x] #1132 
- [x] #1224 
- [x] #1233 
- [x] #1234 
- [x] #1235 
- [x] #1226
- [x] #3916 
- [x] #3915 

```rust
//! An implementation of SipHash.

//#[lang = "sized"]
//trait Sized {}

mod intrinsics {
    extern "rust-intrinsic" {
        pub fn wrapping_add<T>(a: T, b: T) -> T;
        pub fn rotate_left<T>(a: T, b: T) -> T;
        pub fn offset<T>(ptr: *const T, count: isize) -> *const T;
    }
}

#[lang = "add"]
trait Add<Rhs = Self> {
    type Output;

    fn add(self, rhs: Rhs) -> Self::Output;
}

macro_rules! add_impl {
    ($($t:ty)*) => ($(
        impl Add for $t {
            type Output = $t;

            #[inline]
            #[rustc_inherit_overflow_checks]
            fn add(self, other: $t) -> $t { self + other }
        }

        //forward_ref_binop! { impl const Add, add for $t, $t }
    )*)
}

add_impl! { usize u8 u16 u32 u64 u128 isize i8 i16 i32 i64 i128 f32 f64 }

impl<T> *const T {
    pub unsafe fn add(self, count: usize) -> Self {
        // SAFETY: the caller must uphold the safety contract for `offset`.
        unsafe { self.offset(count as isize) }
    }

    pub unsafe fn offset(self, count: isize) -> *const T {
        // SAFETY: the caller must uphold the safety contract for `offset`.
        unsafe { intrinsics::offset(self, count) }
    }
}

macro_rules! impl_uint {
    ($($ty:ident = $lang:literal),*) => {
        $(
            //#[lang = $lang]
            impl $ty {
                pub fn wrapping_add(self, rhs: Self) -> Self {
                    intrinsics::wrapping_add(self, rhs)
                }

                pub fn rotate_left(self, n: u32) -> Self {
                    intrinsics::rotate_left(self, n as Self)
                }

                pub fn to_le(self) -> Self {
                    #[cfg(target_endian = "little")]
                    {
[...]

PR #4002, opened 2025-07-23
ci: Compile with checking enabled

powerboat9:


powerboat9:
This doesn't help resolve https://builder.sourceware.org/buildbot/#/builders/27 like I was hoping, but it should still be nice to have.

tschwinge:
The fails are for a 32-bit x86 host system.  We have `.github/workflows/ccpp32alpine.yml`, so what is different here, hmm...

tschwinge:
Haha!  "GCC Rust build and test (Alpine 32-bit) / build-alpine-32bit-and-check-alpine-32bit" produced:

```
UNSUPPORTED: rust/execute/torture/sip-hasher.rs   -O0 
UNSUPPORTED: rust/execute/torture/sip-hasher.rs   -O1 
UNSUPPORTED: rust/execute/torture/sip-hasher.rs   -O2 
UNSUPPORTED: rust/execute/torture/sip-hasher.rs   -O3 -g 
UNSUPPORTED: rust/execute/torture/sip-hasher.rs   -Os 
UNSUPPORTED: rust/execute/torture/sip-hasher.rs   -O2 -flto -fno-use-linker-plugin -flto-partition=none 
UNSUPPORTED: rust/execute/torture/sip-hasher.rs   -O2 -flto -fuse-linker-plugin -fno-fat-lto-objects 
```

..., because of `// { dg-skip-if "" { *-*-* } { "-m32" } { "" } }`, and this testing here is running with explicit `-m32` (nothing wrong with that).  That's why we don't see the `internal compiler error: in expand_shift_1, at expmed.cc:2713` here.

That test case was added in #3915, #1247 commit 9526906f13955348a5f214bad92f4d42ce9ffa5a "gccrs: fix bad monomophization of generic paths".  @philberty: What's the rationale behind the `-m32` skip?  If anything, this should skip for effective-target `ilp32` (?) or only allow for `lp64` (?), but that all sounds dubious, papering over an actual problem -- like this ICE (I suppose?), so yet better, instead of `dg-skip-if`, this should (conditionally) mark up the ICE with `dg-ice`.

philberty:
@tschwinge this is my excitement where this goal test case now works but its uncovered a bug with how we do the intrinsic u32 rotate which is causing this ICE on m32

I think the issue as far as i can tell is we need to change the shift count to be masked by the size of the type but it was an related change to that patch. I'll open an issue

Issue #4003, opened 2025-07-23
Fix sip-hasher on 32bit


philberty:
There is an issue with shifts and rorates on 32 bit systems we need to figure out

powerboat9:
https://github.com/Rust-GCC/gccrs/pull/4002#issuecomment-3106738285

PR #4007, opened 2025-07-24
gccrs: 128bit integer support is limited on 32 bit platforms

philberty:
This is a wider gcc issue we will need to address at some point but this test runs fine under 32 bit more without the 128bit lang item for rotate here.

gcc/testsuite/ChangeLog:

	* rust/execute/torture/sip-hasher.rs: remove 128bit lang item

### DIFF
--- a/gcc/testsuite/rust/execute/torture/sip-hasher.rs
+++ b/gcc/testsuite/rust/execute/torture/sip-hasher.rs
@@ -1,4 +1,3 @@
-// { dg-skip-if "" { *-*-* } { "-m32" } { "" } }
 // { dg-options "-w" }
 // { dg-output "Hash: 0x63d53fd2170bbb8c\r*\n" }
 #![feature(intrinsics)]
@@ -36,7 +35,7 @@ macro_rules! add_impl {
     )*)
 }
 
-add_impl! { usize u8 u16 u32 u64 u128 isize i8 i16 i32 i64 i128 f32 f64 }
+add_impl! { usize u8 u16 u32 u64 isize i8 i16 i32 i64 f32 f64 }
 
 impl<T> *const T {
     pub unsafe fn add(self, count: usize) -> Self {
@@ -83,7 +82,6 @@ impl_uint!(
     u16 = "u16",
     u32 = "u32",
     u64 = "u64",
-    u128 = "u128",
     usize = "usize"
 );
 


gcc/testsuite/rust/execute/torture/sip-hasher.rs
```diff
@@ -1,4 +1,3 @@
-// { dg-skip-if "" { *-*-* } { "-m32" } { "" } }
```
thesamesam, line 1: Maybe as @tschwinge suggested it'd be better to dg-ice for 32-bit?

thesamesam: Should link to https://github.com/Rust-GCC/gccrs/issues/4003 in the commit message anyway

philberty:
yeah i was testing quite a bit here and normal things like the arithmetic operators like +, - etc wok on 128 bit integers but rotate doesnt the C front-end explicitly says 128bit doesnt work on 32 bit.

From testing it seems as though rotates and shifts dont work.

The RTL passes do try to expand the rotate into two parts like two u64's as far as i can tell in gdb but this a bit beyond my ability right now its hitting stuff in the code relating to the target option support so i think its a wider thing to ask GCC folks about at this point.

philberty:
@tschwinge @dkm @CohenArthur what do you think i should do? I think i can tweak this test case to pass for this alpine 32 bit CI but not sure who to poke in GCC about this maybe i should just ask on the gcc mailing list i guess?

tschwinge:
As far as I know, GCC supports 128-bit integers only for certain (64-bit?) targets, but not for 32-bit targets.  As far as I remember, the new GCC/Cobol support is also running into that (because they need 128-bit integers) -- so maybe that'll get resolved sooner or later.

I see C, C++ test cases conditionalize `__int128` usage on `#ifdef __SIZEOF_INT128__`, or effective-target `int128`, for example.  The latter works for GCC/Rust, too, so I suppose we could do something like `dg-additional-options -frust-cfg=int128 { target int128 }`, and conditionalize `u128`, `i128` on `#[cfg(int128)]` (untested)?

tschwinge:
A follow-on question then is whether Rust mandates `u128`, `i128` support unconditionally, so we'll have to make that work also for 32-bit configurations?  (Like Cobol, apparently.)

tschwinge:
By the way, I see `u128`, `i128` appear in some of our test cases, and commented out in others.  All that could -- at least for now -- use effective-target `int128`, I suppose?

powerboat9:
https://blog.rust-lang.org/2024/03/30/i128-layout-update/
https://github.com/rust-lang/lang-team/issues/255
https://godbolt.org/z/EYMbdx63K

Looks like rustc didn't have a stable ABI for `i128` in `1.49`. However, I'm fairly certain `rustc` supports `i128` on ~all targets.

powerboat9:
https://github.com/rust-lang/rustc_codegen_gcc/issues/71
https://github.com/rust-lang/rustc_codegen_gcc/pull/103

dkm:
> @tschwinge @dkm @CohenArthur what do you think i should do? I think i can tweak this test case to pass for this alpine 32 bit CI but not sure who to poke in GCC about this maybe i should just ask on the gcc mailing list i guess?

if you don't feel like sending a mail, IRC is also an option.

PR #1632, opened 2022-11-07
Fix mac-os regression in apply generic arguments to method calls

philberty:
When applying generic arguments to method calls such as:

```receiver.method_name<i32, bool>()```

This ended up wrongly using the default constructor with an empty generic arguments which seems
like a possible bug in the new version of clang. This explicitly sets up all relevant copy constructors
for HIR::PathExprSegment and removes the defaults.

PR #4008, opened 2025-07-25
Reenable test on *-*-darwin*

powerboat9:
This seems to have been disabled to work around a bug in clang, back in 2022. See #1632


dkm (approved):

philberty:
I would maybe just skip that test for macos if you can it would be nice to turn this back on.

philberty:
I think we've stopped the mac ci job for now

PR #4009, opened 2025-07-25
[WIP] Enable 128 bit integers on 32 bit targets

powerboat9:


gcc/rust/rust-gcc.cc
```diff
@@ -95,9 +96,104 @@ namespace Backend {
...
+       long_long_unsigned_type_node, long_long_integer_type_node};
+
+  for (size_t i = 0; i < 4; i++)
+    {
+      // check for a power of 2 or 0
+      if (normal_int_sizes[i] & (normal_int_sizes[i] - 1))
+	continue;
+
+      // assume size is not 0
+      int idx = __builtin_ctz (normal_int_sizes[i]) - 3;
```
pinskia: Don't use __builtin_ctz directly. 

Do:
```
int idx = exact_log2 (normal_int_sizes[i]);
if (idx == -1)
  continue;
idx -= 3;
```

gcc/rust/rust-backend.h
```diff
@@ -50,6 +50,9 @@ using Rust::GGC::Ident;
 
 } // namespace GGC
 
+extern tree rust_int_trees[10];
```
pinskia: This almost definitely needs GTY(()) here so the trees in rust_int_trees don't get GC'ed.

gcc/rust/backend/rust-compile-type.cc
```diff
@@ -504,32 +504,23 @@ TyTyResolveCompile::visit (const TyTy::IntType &type)
   switch (type.get_int_kind ())
     {
     case TyTy::IntType::I8:
-      translated = Backend::named_type ("i8", Backend::integer_type (false, 8),
-					BUILTINS_LOCATION);
+      translated = Backend::rust_int_trees[5];
```
pinskia: I would have an enum for the special # 5 here.

gcc/rust/backend/rust-compile-intrinsic.cc
```diff
@@ -557,8 +558,16 @@ rotate_handler (Context *ctx, TyTy::FnType *fntype, tree_code op)
   // BUILTIN rotate FN BODY BEGIN
   tree x = Backend::var_expression (x_param, UNDEF_LOCATION);
   tree y = Backend::var_expression (y_param, UNDEF_LOCATION);
+
+  tree y_mask
+    = build_int_cst (unsigned_type_node, TYPE_SIZE (TREE_TYPE (x)) - 1);
```
pinskia: s/TYPE_SIZE (TREE_TYPE (x)) - 1/element_precision (TREE_TYPE (x)) - 1/

gcc/rust/backend/rust-compile-type.cc
powerboat9: I've introduced macros for roughly the same effect

PR #4010, opened 2025-07-28
Add HIR::OffsetOf node

CohenArthur:
- **hir: Add OffsetOf node**


gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -375,6 +375,12 @@ CompileExpr::visit (HIR::LlvmInlineAsm &expr)
   ctx->add_statement (asm_codegen.tree_codegen_asm (expr));
 }
 
+void
+CompileExpr::visit (HIR::OffsetOf &expr)
+{
+  rust_unreachable ();
```
philberty, line 381: I think there are maybe two ways to do this but i would probably do it like this:

1. lookup the tyty type on the expr.get_type () HirId;

2. Then call ``tree offset_type = TyTyResolveCompile::compile (ctx, tyty_of_the_type);```

3. Then you will need to assert its a RECORD_TYPE iirc

4. You will need to iterate the record to make sure this field exists we probably should do that check on the ADTType in the type check pass upfront

5. call type_field_offset from rust-gcc.cc to get the offset and turn it into a tree using build_int_cst (size_node_type, value)

will probably do this for you

philberty (approved):
All LGTM to me i gave you a suggestion on how to fill out the code-gen pass for it but that can be a separate pr if you want

philberty:
just noticed gcc provides both:

```
tree
byte_position (const_tree field)
{
  return byte_from_pos (DECL_FIELD_OFFSET (field),
			DECL_FIELD_BIT_OFFSET (field));
}

/* Likewise, but return as an integer.  It must be representable in
   that way (since it could be a signed value, we don't have the
   option of returning -1 like int_size_in_byte can.  */

HOST_WIDE_INT
int_byte_position (const_tree field)
{
  return tree_to_shwi (byte_position (field));
}
```

You could call the byte_position on the field index instead so you dont have to convert back to a tree again

You'll see what i mean when you read rust-gcc.cc:type_field_offset

PR #4011, opened 2025-07-29
nr1.0: Remove `rust/typecheck` support

powerboat9:
Follows up on #3865


philberty (approved):
LGTM its good to see it going away and just one path of code

PR #4013, opened 2025-07-29
Fix semicolon after some namespace

P-E-P:
Remove namespace comment after classes and structs to avoid warnings.


philberty (approved):
LGTM

PR #3914, opened 2025-07-12
[WIP] Compile `libcore`

powerboat9:
I'm making this PR just so it's easier to track what I've done locally

PR #4016, opened 2025-07-30
Parse expression instead of literal in attributes

P-E-P:


P-E-P:
@powerboat9 I've just seen that this PR clashes with #3914 , #3914 contains some other changes I wished to work on next (rustc_allow_const_fn_unstable) is it ready to merge ? Could you bring up some PRs with some of those changes ? Should we merge this PR ?

powerboat9:
I'll create some PRs -- I think this PR in particular is ~equivalent to a commit I have ready, although I need to do some rebasing + conflict resolution to find that

powerboat9:
I've got #4018, along with a follow-up commit ready that's ~equivalent to this PR. I think I'll probably just use the latter commit to make suggestions for this PR

gcc/rust/ast/rust-ast-collector.cc
```diff
@@ -839,13 +839,13 @@ TokenCollector::visit (MetaItemLitExpr &item)
...
 void
-TokenCollector::visit (MetaItemPathLit &item)
+TokenCollector::visit (MetaItemPathExpr &item)
 {
-  auto path = item.get_path ();
-  auto lit = item.get_literal ();
+  auto &path = item.get_path ();
+  auto &expr = item.get_expr ();
   visit (path);
   push (Rust::Token::make (COLON, item.get_locus ()));
```
powerboat9: Should this be `EQUAL` instead of `COLON`?

gcc/rust/ast/rust-ast.cc
```diff
@@ -4144,10 +4136,10 @@ MetaNameValueStr::check_cfg_predicate (const Session &session) const
 }
 
 bool
-MetaItemPathLit::check_cfg_predicate (const Session &session) const
+MetaItemPathExpr::check_cfg_predicate (const Session &session) const
 {
   return session.options.target_data.has_key_value_pair (path.as_string (),
-							 lit.as_string ());
+							 expr->as_string ());
 }
```
powerboat9: Should probably add a `rust_assert (expr->is_literal ())`

gcc/rust/ast/rust-ast.cc
```diff
@@ -4234,13 +4226,6 @@ MetaListNameValueStr::to_attribute () const
 		    std::move (new_seq_container));
 }
 
-Attribute
-MetaItemPathLit::to_attribute () const
-{
-  return Attribute (path, std::unique_ptr<AttrInputLiteral> (
-			    new AttrInputLiteral (lit)));
-}
-
```
powerboat9: Is this function really unnecessary?

gcc/rust/ast/rust-expr.h
```diff
@@ -245,36 +245,45 @@ class MetaItemLitExpr : public MetaItemInner
...
+    : path (std::move (path)), expr (std::move (expr))
+  {}
+
+  MetaItemPathExpr (const MetaItemPathExpr &other)
+    : path (other.path), expr (other.expr->clone_expr ())
   {}
 
+  MetaItemPathExpr operator= (const MetaItemPathExpr &other)
+  {
+    path = other.path;
```
powerboat9: Should probably stick a `MetaItem::operator= (other);` in here

gcc/rust/ast/rust-expr.h
```diff
@@ -245,36 +245,45 @@ class MetaItemLitExpr : public MetaItemInner
...
   }
 
   MetaItem::ItemKind get_item_kind () const override
   {
-    return MetaItem::ItemKind::PathLit;
+    return MetaItem::ItemKind::PathExpr;
   }
 
-  // There are two Locations in MetaItemPathLit (path and lit_expr),
+  // There are two Locations in MetaItemPathExpr (path and lit_expr),
```
powerboat9: `s/lit_expr/expr/`

powerboat9 (changes requested):

gcc/rust/ast/rust-ast.cc
P-E-P: I'm not sure about that, it could be a path expression as well right ?

powerboat9: True. You'd probably want to verify that though -- `as_string` varies a lot in terms of behavior

P-E-P: Yeah, I wish we had some specified interface/ way to ensure it outputs what we really want. Right now it is kinda a mess, sometimes it outputs the node as the token collector would and sometimes it is more like a debug output.

I don't even like the fact we mix `to_string` and `as_string` even though we don't care if the function gives a reference or a new value.

gcc/rust/ast/rust-ast.cc
P-E-P: I wasn't really sure what to do with it since we cannot guarantee AttrInput hold literals anymore. I could have modified AttrInputLiteral to AttrInputExpr but this would have potentially snowballed into a lot of other changes. Since it is not used anywhere I decided against keeping it.

powerboat9: In my patch, I just stuck a `rust_assert (expr->is_literal ())` in there. It's probably better to do that than remove (and later add back? idk) this function just yet. Also, it looks like it's used to override `MetaItemInner::to_attribute`

gcc/rust/ast/rust-expr.h
```diff
@@ -249,36 +249,46 @@ class MetaItemLitExpr : public MetaItemInner
...
 
 public:
-  MetaItemPathLit (SimplePath path, LiteralExpr lit_expr)
-    : path (std::move (path)), lit (std::move (lit_expr))
+  MetaItemPathExpr (SimplePath path, std::unique_ptr<Expr> expr)
+    : path (std::move (path)), expr (std::move (expr))
+  {}
+
+  MetaItemPathExpr (const MetaItemPathExpr &other)
+    : path (other.path), expr (other.expr->clone_expr ())
```
powerboat9: Should include `MetaItem (other)` in the member initialization list

gcc/rust/ast/rust-expr.h
```diff
@@ -249,36 +249,46 @@ class MetaItemLitExpr : public MetaItemInner
...
   {}
 
+  MetaItemPathExpr operator= (const MetaItemPathExpr &other)
+  {
+    MetaItem::operator= (other);
+    path = other.path;
+    expr = other.expr->clone_expr ();
+    return *this;
+  }
+
```
powerboat9, line 278: Should probably include `MetaItemPathExpr (MetaItemPathExpr &&) = default;` and `MetaItemPathExpr &operator= (MetaItemPathExpr &&) = default;`

powerboat9 (changes requested):
Otherwise, looks good. I wouldn't be noticing these if I wasn't making line-by-line comparisons with https://github.com/powerboat9/gccrs/commit/a690152603e354f3a0e9492af5fcec81cba69a6f

gcc/rust/ast/rust-expr.h
P-E-P: @powerboat9 IIRC default constructors are produced by default and do not require to be explicitly set.

powerboat9: Not if you explicitly define copy constructor/copy assignment
https://en.cppreference.com/w/cpp/language/move_assignment.html#Implicitly-declared_move_assignment_operator

PR #4020, opened 2025-07-30
Add a few missing delete to allocated temporaries

P-E-P:


powerboat9:
I have #3862 which seems to overlap

PR #4028, opened 2025-08-01
Add `derived` flag to item

sakupan102:
To distinguish between items that are automatically generated by #[derive()] and those that aren't, I added a derived flag to Item.

- \[x] Run `clang-format`

order to get feedback.
---


powerboat9:
Why distinguish? Producing better errors for derive-generated code?

P-E-P:
> This LGTM but lets see what @P-E-P and @CohenArthur say.
> 
> Another option to avoid changing the HIR could be to make a std::map<NodeId, bool AST::Item&
> 
> in the rust-mappings.h and during hir lowering you can look that up to determine if an ITEM was derived.
> 
> But yeah see what the guys think.

A set instead of a `map<NodeId, bool>` would probably be better but I agree, this kind of informations should probably be kept outside of the AST.

philberty:
yeah sounds like if you need this flag then follow @P-E-P advice

CohenArthur:
`rustc` adds the `#[automatically_derived]` attribute to the impls it creates, which is something we can consider as well. I'm not sure if the attribute is used for other things

sakupan102:
I will close this PR and create a new issue because the implementation is entirely different.

PR #4029, opened 2025-08-01
Explicitely create empty vector

P-E-P:


P-E-P:
https://github.com/rust-lang/rust/blob/1.49.0/library/core/src/marker.rs#L202

Issue #3867, opened 2025-07-04
Handle `#[link_name]` attribute


CohenArthur:

philberty:
This should be fairly easy to support we just need to set the ASM name node on the fndecl so in:

rust-compile-extern.h:   void visit (HIR::ExternalFunctionItem &function) override

You just need to set the asm name from this attribute if it exists at the moment its just doing:

"std::string asm_name = function.get_item_name ().as_string ();"

powerboat9:
I have a patch for this in my dev branch -- I'll create a PR

moabo3li:
Is this should be closed?

PR #4031, opened 2025-08-01
Handle `link_name` attribute

powerboat9:
Unless I'm mistaken, we don't need to put `no_mangle` in `DECL_ATTRIBUTES (fndecl)`, right? Since we do all of the mangling, and gcc doesn't seem to have a `no_mangle` attribute?

### DIFF
--- a/gcc/rust/backend/rust-compile-base.cc
+++ b/gcc/rust/backend/rust-compile-base.cc
@@ -33,6 +33,7 @@
 #include "rust-compile-implitem.h"
 #include "rust-attribute-values.h"
 #include "rust-immutable-name-resolution-context.h"
+#include "rust-ggc.h"
 
 #include "fold-const.h"
 #include "stringpool.h"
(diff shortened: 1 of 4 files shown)


powerboat9:
Fixes #3867

philberty (changes requested):
I think this is partly there but there is special handling for the function declarations in: ust-compile-extern.h: void visit (HIR::ExternalFunctionItem &function) override

philberty:
Also you should put the Fixes into the commit message too since this is almost it finished

gcc/rust/backend/rust-compile-base.cc
```diff
@@ -46,11 +47,59 @@
...
+	  AST::LiteralExpr *literal;
+
+	  if (!attr.has_attr_input ())
+	    goto err_link_name;
+
+	  input = &attr.get_attr_input ();
+
+	  if (input->get_attr_input_type ()
+	      != AST::AttrInput::AttrInputType::LITERAL)
+	    goto err_link_name;
```
P-E-P: Not a fan of this goto. You could have split the code in multiple functions and I don't want newcomers to grep the codebase, find this goto and sprinkle some everywhere nor be confused about it.

philberty: good catch yeah avoid goto's please

powerboat9:
Now depends on #4044

powerboat9:
It looks like `link_name` isn't supposed to affect non-extern items, and extern items aren't mangled regardless of ABI

gcc/testsuite/rust/execute/torture/link-name.rs
```diff
@@ -0,0 +1,14 @@
...
+
+extern "C" {
+    #[link_name = "printf"]
+    fn druckt(fmt: *const i8, ...);
+}
+
+fn main() -> i32 {
+    let a = "gcc\0";
+
+    unsafe { druckt("%s\n\0" as *const str as *const i8, a as *const str as *const i8); }
```
P-E-P, line 13: A scan-assembler test over here may be helpful, it would emphasize the work of this PR although it would be limited to some architectures.

P-E-P (approved):
Small nit otherwise LGTM

gcc/testsuite/rust/execute/torture/link-name.rs
powerboat9: I've added some directives to scan the RTL

PR #4037, opened 2025-08-02
Avoid including `rust-parse-impl.h` in `rust-parse.h`

powerboat9:
This should reduce incremental compile times when modifying `rust-parse-impl.h` and not `rust-parse.h`


philberty (approved):
LGTM

PR #4038, opened 2025-08-03
Fix object copying issue causing pointer inconsistency

sakupan102:
# Problem Summary
This PR fixes an issue where an object was unintentionally copied instead of moved, causing inconsistencies in object state and pointer references. Specifically, the copy constructor was invoked, resulting in a duplicated object where internal pointers or identifiers became misaligned. The original object was left in an outdated state, and any pointers referring to it could point to the wrong memory.

- \[x] Run `clang-format`


philberty (approved):
LGTM

PR #4039, opened 2025-08-03
gccrs: Refactor HIR::PatternItem class and its derivatives

Polygonalr:
Includes some renaming to generalise the previous `TupleItems` class (now `PatternItems`) so that it can be used for `SlicePattern` as well. Renamed the enums to `NO_REST`/`HAS_REST` to differentiate from `RangePattern` which uses the same `..` syntax.

This is a pretty meaty patch as there are many resulting cascading renames, but it is a pre-requisite for an [even meatier upcoming patch to make `HIR` changes for rest pattern support in `HIR::SlicePattern`.](https://github.com/Polygonalr/gccrs/pull/2/files)


Polygonalr:
I chose `HAS_REST` and `NO_REST` since the Rust reference refers the `..` as rest patterns. I did consider something like `PARTIAL` and `COMPLETE`, but it doesn't feel like it conveys the information as directly as `HAS_REST` and `NO_REST`. However, I'm open to other opinions and will be happy to change it if that is better for other maintainers.

philberty:
yeah i cant think of a better name right now so i think thats a minor issue we can sort out later if we think of it

Polygonalr:
Rebased to master and did the renames in `gcc/rust/checks/errors/rust-readonly-check2.cc` as well.

PR #4040, opened 2025-08-03
Handle attributes in expression macros

powerboat9:


powerboat9:
I know this fixes an existing issue, but I can't find it

PR #4047, opened 2025-08-08
Add derived set to collect derived nodes.

sakupan102:

order to get feedback.
---

### DIFF
--- a/gcc/rust/util/rust-hir-map.cc
+++ b/gcc/rust/util/rust-hir-map.cc
@@ -1358,5 +1358,17 @@ Mappings::lookup_captures (NodeId closure)
     return cap->second;
 }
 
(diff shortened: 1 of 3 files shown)


gcc/rust/util/rust-hir-map.cc
```diff
@@ -1358,5 +1358,17 @@ Mappings::lookup_captures (NodeId closure)
     return cap->second;
 }
 
+void
+Mappings::add_derived_nodes (NodeId node_id)
```
powerboat9: Should probably be `add_derived_node` (not plural)

powerboat9:
Otherwise, looks good

P-E-P (approved):

Issue #3444, opened 2025-02-20
Macro expansions replace parent modules


powerboat9:
### Summary

It seems like `ExpandVisitor` will, rather than expand macro invocations within a module, expand modules to the result of a contained macro invocation.

### Reproducer

I tried this code:

```rust
mod foo {
    macro_rules! bar {
        () => ()
    }

    bar! ();

    pub struct S;
}

pub fn buzz(_: foo::S) {
}
```

### Does the code make use of any (1.49) nightly feature ?

- [ ] Nightly

### Godbolt link

[...]

PR #4052, opened 2025-08-09
Add checks to `ExpandVisitor`

powerboat9:
This should help detect issues like #3444


CohenArthur (approved):
this looks good to me, thanks! note that I think it would be worth reworking this visitor to have a more generic "ExternalVisitor" framework which can replace pointers like we do here and in the desugars

PR #4055, opened 2025-08-10
Implement unused variable checker on HIR

sakupan102:
This PR closes https://github.com/Rust-GCC/gccrs/issues/4086, https://github.com/Rust-GCC/gccrs/issues/4085

- \[x] Run `clang-format`


gcc/rust/checks/lints/unused-var/rust-unused-var-context.cc
```diff
@@ -0,0 +1,69 @@
...
+namespace Analysis {
+
+UnusedVarContext *
+UnusedVarContext::get ()
+{
+  static UnusedVarContext *instance;
+  if (instance == nullptr)
+    instance = new UnusedVarContext ();
+  return instance;
+}
```
CohenArthur: I think this is fine but we should also check if we can do without a global variable

gcc/rust/checks/lints/unused-var/rust-unused-var-collector.cc
```diff
@@ -0,0 +1,130 @@
...
+UnusedVarCollector::go (HIR::Crate &crate)
+{
+  for (auto &item : crate.get_items ())
+    item->accept_vis (*this);
+}
+void
+UnusedVarCollector::visit (HIR::LetStmt &stmt)
+{
+  HIR::Pattern &pattern = stmt.get_pattern ();
+  collect_variable (pattern);
```
CohenArthur: can't we visit instead here? mabye I'm missing something but I think all patterns should create variables to check for, so we could just do `visit(pattern)` and then create visitors for all our kinds of patterns instead of manually dispatching like in `collect_variable`

CohenArthur (commented):
I think this is a really good base and the code looks good to me! I think it is important to add testcases for features like these, so it would be good if you could add a flag to enable the lints you've written and then add testcases to the testsuite

gcc/rust/checks/lints/unused-var/rust-unused-var-collector.h
```diff
@@ -0,0 +1,67 @@
...
+
+  template <typename T> void mark_path_used (T &path_expr)
+  {
+    NodeId ast_node_id = path_expr.get_mappings ().get_nodeid ();
+    NodeId def_id;
+
+    if (auto id = nr_context.lookup (ast_node_id))
+      def_id = *id;
+    else
+      return;
```
CohenArthur: so if we fail to lookup the variable we silently return and do nothing - I think it would be worth crashing if we encounter this instead, as it should never happen
```suggestion
    NodeId  def_id = nr_context.lookup (ast_node_id).value();
```

gcc/rust/checks/lints/unused-var/rust-unused-var-collector.h
```diff
@@ -0,0 +1,67 @@
...
+
+    auto hir_id = mappings.lookup_node_to_hir (def_id);
+    if (!hir_id)
+      return;
```
CohenArthur: likewise

gcc/testsuite/rust/unused-var/unused_var.exp
CohenArthur, line 1: I think we don't need this file, instead...

gcc/testsuite/rust/unused-var/template_function_0.rs
CohenArthur, line 1: ... You can move this file to the regular `gcc/testsuite/rust/compile` folder, and add the following line to the top of the file:

```rust
// { dg-additional-options "-frust-unused-check-2.0" }
```

gcc/testsuite/rust/unused-var/static_item_0.rs
CohenArthur, line 1: Likewise here :)

CohenArthur (approved):
Thank you @sakupan102, this is really good work!!! Just a couple nitpicks and we should merge this. Well done :)

sakupan102:
@CohenArthur 
I made the suggested changes based on your review. If everything looks good, I'll squash the commits.

philberty:
There are merge conflicts here @sakupan102

PR #4058, opened 2025-08-10
Allow early resolution to use the language prelude

powerboat9:


gcc/rust/resolve/rust-resolve-builtins.cc
```diff
@@ -0,0 +1,107 @@
...
+
+  TyTy::BaseType *create_type (HirId hir_id) const { return type_gen (hir_id); }
+
+private:
+  const char *name;
+  TyTy::BaseType *(*type_gen) (HirId);
+};
+
+#define GEN_DEF(name, expr)                                                    \
+  TypeBuilder ((name), [] (HirId hir_id) -> TyTy::BaseType * { return (expr); })
```
CohenArthur:
```suggestion
  TypeBuilder ((name), [] (HirId hir_id) -> TyTy::BaseType * { return new (expr); })
```

?

feel free to ignore

powerboat9: I could fold that into the macro. It looks like C++14 doesn't have `constexpr` conversion of lambdas to function pointers though, so I'll have to find a workaround.

powerboat9:
I've used X macros to replace the lambdas

powerboat9:
@CohenArthur thoughts on merging?

CohenArthur (approved):
That looks good to me, thanks for working on this Owen!

PR #4060, opened 2025-08-11
Vendor `libcore 1.49`

powerboat9:
This vendors `libcore` under `libgrust`. In the future we can use this for building `libcore` as a runtime library, and in the near-future we can use it to test progressively later phases of the compiler on `libcore` as part of the testsuite.


powerboat9:
Note that this PR leaves usage of `libcore` to future PRs

CohenArthur (approved):
LGTM I like the idea. I assume this is completely unmodified?

powerboat9:
Yep. `LICENSE-*` files are copied from the `rustc` repo root

P-E-P (approved):
Mmh a bit sad that the Cargo.toml in libcore does not document the library's version. I like @philberty 's idea to track the source of the library, not sure the hash is required but the version should be. And it should be somewhere close so we won't forget to update it. Otherwise LGTM.

powerboat9:
> Mmh a bit sad that the Cargo.toml in libcore does not document the library's version. I like @philberty 's idea to track the source of the library, not sure the hash is required but the version should be. And it should be somewhere close so we won't forget to update it. Otherwise LGTM.

I've created a file `libgrust/version-info`

philberty (approved):
LGTM

PR #4061, opened 2025-08-11
Fix Macro builtins asm clobber infinite loop

P-E-P:


gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -1124,8 +1125,12 @@ parse_llvm_clobbers (LlvmAsmContext &ctx)
 	{
 	  ctx.llvm_asm.get_clobbers ().push_back (
 	    {strip_double_quotes (token->as_string ()), token->get_locus ()});
+
+	  parser.skip_token (STRING_LITERAL);
 	}
-      parser.skip_token (STRING_LITERAL);
+
+      if (parser.peek_current_token ()->get_id () != ctx.last_token_id)
```
powerboat9: What about `COLON`?

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -1177,25 +1182,28 @@ parse_llvm_asm (location_t invoc_locus, AST::MacroInvocData &invoc,
...
       // If the macro invocation has a semicolon (`asm!("...");`), then we
       // need to make it a statement. This way, it will be expanded
       // properly.
       if (semicolon == AST::InvocKind::Semicoloned)
-	single_vec.emplace_back (AST::SingleASTNode (
-	  std::make_unique<AST::ExprStmt> (std::move (node), invoc_locus,
-					   semicolon
-					     == AST::InvocKind::Semicoloned)));
+	{
+	  bool has_semicolon = semicolon == AST::InvocKind::Semicoloned;
```
powerboat9: Isn't `has_semicolon` a duplicate of the `if` condition?

P-E-P: It is, I only made that change to make the line easier to read without checking the context.

gcc/rust/expand/rust-macro-builtins-asm.cc
P-E-P: coming in a fixup later, that PR was meant a draft

gcc/rust/expand/rust-macro-builtins-asm.cc
powerboat9: ah

P-E-P: Fixed, thanks for noticing.

gcc/rust/expand/rust-macro-builtins-asm.cc
```diff
@@ -1124,8 +1125,11 @@ parse_llvm_clobbers (LlvmAsmContext &ctx)
 	{
 	  ctx.llvm_asm.get_clobbers ().push_back (
 	    {strip_double_quotes (token->as_string ()), token->get_locus ()});
+
+	  parser.skip_token (STRING_LITERAL);
 	}
-      parser.skip_token (STRING_LITERAL);
+
+      parser.skip_token (COMMA);
```
powerboat9: What if there isn't a trailing `COMMA`?

powerboat9: I just realized `maybe_skip_token` exists -- much more convenient than the `if` statement I have locally

P-E-P: ~~I rediscovered yesterday that skip_token does not abort when the token is wrong, it simply returns false but a call to maybe_skip_token may make the intention clearer.~~ It doesn't and effectively prints an error. Will change to maybe_skip_token.

PR #4063, opened 2025-08-11
Improve handling of non-final path segments

powerboat9:


gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -531,7 +531,7 @@ ForeverStack<N>::resolve_segments (
 				       || seg.is_lower_self_seg ()))
 	return tl::nullopt;
 
-      tl::optional<typename ForeverStack<N>::Node &> child = tl::nullopt;
+      Node *child = nullptr;
```
P-E-P: I'm trying to get rid of most raw pointers in the repository and here you are introducing one more :smile: I'd prefer a reference. Hidden behind an optional if it may not be available, but you clearly changed that, what's your rationale behind that ?

powerboat9: I had ~https://www.foonathan.net/2018/07/optional-reference/ in mind, but looking over our `optional.h` it seems like an optional reference wouldn't cause problems here. I think I'll just make it a `tl::optional<std::reference_wrapper<Node>>`

P-E-P (approved):

Issue #3896, opened 2025-07-10
Allow eager macro invocations in `format_args!()`


CohenArthur:
```rust
format_args!(concat!("...", "..."));
```

PR #4064, opened 2025-08-12
Allow `format_args!` with a macro as first argument

powerboat9:
Fixes https://github.com/Rust-GCC/gccrs/issues/3896. `FormatArgsPieces` isn't great, but I don't see a better alternative for now.


powerboat9:
Now depends on ~#4068~ #4074

CohenArthur (commented):
have you looked at reusing the code for expanding eager macro invocations for other builtin macros? I haven't reviewed your changes yet but it seems like a lot of trouble. I was hoping we could just reuse the existing infrastructure and create the `AST::FormatArgs` node only when there are no eager macro invocations left

PR #4065, opened 2025-08-12
Remove `#[simd_test]` support

powerboat9:
This attribute is a procedural macro defined by the crate `simd-test-macro`, not a built-in macro.


CohenArthur (approved):
thanks! good catch!

PR #4066, opened 2025-08-12
Collect lang items after expansion

powerboat9:
This allows macros to create lang items.


philberty (approved):
LGTM and sounds good since lots of lang items are defined after expansion

PR #4067, opened 2025-08-12
`CfgStrip` AST nodes marked with `#[test]`

powerboat9:


P-E-P (approved):
I was under the impression that `#[test]` did not remove the function because we usually put them in a `cfg(test)` module, turns out I was wrong. LGTM

PR #4068, opened 2025-08-12
Fix copy construction and assignment of `Pieces`

powerboat9:
Pointers stored in `Pieces::pieces_vector` refer to objects with lifetimes tied to `Pieces::handle`. This commit also removes the move constructor for Pieces, as it's equivalent to the copy constructor.


powerboat9:
I've removed the move constructor instead of creating an accurate implementation, as I'd want to be very careful about maintaining invariants on a moved-from `ffi::FormatArgsHandle`

powerboat9:
Superseded by #4074

PR #4069, opened 2025-08-12
Replace old read-only checker with new one.

sakupan102:
- \[x] Run `clang-format`


philberty (approved):
LGTM

gcc/rust/checks/errors/rust-readonly-check.cc
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2021-2025 Free Software Foundation, Inc.
+// Copyright (C) 2025 Free Software Foundation, Inc.
```
powerboat9, line 1: @CohenArthur we'll probably want to fix up this copyright header before upstreaming

Issue #4071, opened 2025-08-12
Unable to parse identifier pattern's subpattern binds if it is a closure parameter


Polygonalr:
Parser error caused by doing subpattern binds in a closure param.

### Code

[LINK TO GODBOLT](https://godbolt.org/z/E74rf13qq)

```rust
pub fn foo() {
    let c = |bar @ (baz,)| {
        match baz {
            _ => {}
        }
    };

    c((1,));
}
```

While this throws an stability error in `rustc` 1.49 (which we are targeting, this feature stabilizes in `rustc` 1.56), `gccrs` shouldn't be throwing parser errors for this valid Rust code.

### Git commit hash

`5b7a0d7a3aab545799046a3f397ff239b6bc0628`

### Error output

```
<source>:2:28: error: unexpected token '{' in pattern
    2 |     let c = |bar @ (baz,)| {
      |                            ^
<source>:2:28: error: expecting '|' but '{' found
<source>:2:28: error: failed to parse expression in let statement
<source>:2:5: error: failed to parse statement or expression in block expression
    2 |     let c = |bar @ (baz,)| {
      |     ^~~
<source>:8:5: error: failed to parse item in crate
```

PR #4072, opened 2025-08-12
Finish up subpattern codegen support for identifier patterns

Polygonalr:
Implements `let` statement binding support for identifier patterns' subpatterns, together with closure param bindings as well (albeit untested because of #4071)


philberty (approved):
LGTM

PR #4079, opened 2025-08-14
Add `AST::PointerVisitor`

CohenArthur:
This adds a new visitor for visiting unique_ptr members and possibly reseating them, for example when desugaring a `while-let` loop to a regular `loop` expression.

I *think* we could use this for `ExpandVisitor` but it's not as easy since we also do cfg stripping in there, which may remove elements from vectors.

Also, I'm not sure if the name is right - I was debating between `PointerVisitor`, `ReseatVisitor` and `ExternalVisitor`. other ideas welcome

### DIFF
--- a/gcc/rust/ast/rust-expression-yeast.cc
+++ b/gcc/rust/ast/rust-expression-yeast.cc
@@ -21,10 +21,8 @@
 #include "rust-desugar-question-mark.h"
 #include "rust-desugar-try-block.h"
 #include "rust-desugar-for-loops.h"
-#include "rust-ast-full.h"
 #include "rust-desugar-while-let.h"
 #include "rust-expr.h"
-#include "rust-stmt.h"
 
 namespace Rust {
 namespace AST {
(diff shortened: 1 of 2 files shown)


powerboat9:
If we want to use this for `ExpandVisitor` (which I think would be a good idea) it'd have to have support for replacing nodes with `0 <= N` other nodes, instead of just doing 1->1 mappings

powerboat9:
Maybe add something like `virtual void reseat_multi (std::vector<std::unique_ptr<AST::Expr>> &vec) {}`?

CohenArthur:
Note that we'd also need some way to expand inner attributes which might be a little more difficult. I'd like to use this visitor for ExpandVisitor as well but it's proving a little difficult

CohenArthur:
actually I don't think we need anything with inner attributes, but we need to think about this since expanding multiple children may also delete nodes. for now I have something working where ExpandVisitor inherits from this so let's just keep it simple and see how we can improve it later on

PR #4080, opened 2025-08-14
Add desugaring to fixed point

CohenArthur:
Marking this as a draft since it might not be what we actually want, but we do get a lot further when expanding macros in `core`. For example, things like `write!(fmt, ...)?;` now work properly. Note that this may also be because the ExpandVisitor is lacking in many places... so just looking for comments on this for now


powerboat9:
I don't *think* desugaring needs to be fixed point -- unless the resolver is supposed to rely on things being desugared to complete resolution? It seems like the macro issues are all problems local to `ExpandVisitor`.

PR #4089, opened 2025-08-22
Eagerly expand `format_args` macro

powerboat9:
This is a simpler version of #4064. I think I'll probably revisit this latter, to try to improve eager expansion -- it seems like we do a lot of unnecessary `tokens -> AST -> tokens -> AST` conversions.


CohenArthur:
Should we close #4064?

PR #4099, opened 2025-08-25
Improve ExpandVisitor's completeness

CohenArthur:
- **ast: Add PointerVisitor**
- **desugar: Use PointerVisitor for ExpressionYeast**
- **expand: Inherit from PointerVisitor**

This makes ExpandVisitor a lot more complete by making it inherit from PointerVisitor. This solves all of the unexpected macro ID errors we get when compiling core.

### DIFF
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -84,6 +84,7 @@ GRS_OBJS = \
     rust/rust-ast-dump.o \
     rust/rust-ast-collector.o \
     rust/rust-ast-visitor.o \
+    rust/rust-ast-pointer-visitor.o \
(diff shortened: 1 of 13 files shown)


powerboat9:
You should also make sure to remove `ExpandVisitor::visit (BorrowExpr &)`

powerboat9:
It looks like this should actually get us past the expansion phase for `libcore`

CohenArthur:
Yeah I think it does :smile: I'm quite excited

CohenArthur:
I'll spend a little more time on it to clean up functions which are functionally the same as their version in PointerVisitor, to reduce the visitor's size even more

powerboat9:
@CohenArthur

PR #4102, opened 2025-08-26
Implement compilation support for rest patterns in tuple struct patterns

Polygonalr:
Can't believe I forgot about it until now :skull:

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-pattern.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-pattern.cc
@@ -204,22 +204,66 @@ TypeCheckPattern::visit (HIR::TupleStructPattern &pattern)
       {
 	HIR::TupleStructItemsHasRest &items_has_rest
 	  = static_cast<HIR::TupleStructItemsHasRest &> (items);
-	size_t pattern_min_cap = items_has_rest.get_lower_patterns ().size ()
-				 + items_has_rest.get_upper_patterns ().size ();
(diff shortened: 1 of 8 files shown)


gcc/rust/typecheck/rust-hir-type-check-pattern.cc
```diff
@@ -202,31 +202,73 @@ TypeCheckPattern::visit (HIR::TupleStructPattern &pattern)
...
       {
-	// TODO
-	rust_unreachable ();
+	HIR::TupleStructItemsHasRest &items_has_rest
+	  = static_cast<HIR::TupleStructItemsHasRest &> (items);
+	size_t pattern_min_cap = items_has_rest.get_lower_patterns ().size ()
+				 + items_has_rest.get_upper_patterns ().size ();
+	if (variant->num_fields () < pattern_min_cap)
+	  {
+	    rust_error_at (pattern.get_locus (), ErrorCode::E0023,
```
philberty: can you change this to be a rich_location gives options for more locus info

philberty: Also can you add a test case for this error code?

Polygonalr: That was a nice exercise to familiarize myself with `rich_location`, I updated the `NO_REST` case as well.

Polygonalr:
@philberty Should be good to go now after rebasing to master to solve the Alpine Linux test failure.

PR #4104, opened 2025-08-27
Coercion: prioritise right hand type on error

P-E-P:


P-E-P:
Already fixed with #4118

PR #4112, opened 2025-08-28
Improve resolution of `ForLoopExpr`

powerboat9:


CohenArthur (approved):
LGTM but what does this fix? any issues we should add as testcases?

powerboat9:
Trying to find a testcase -- it looks like, actually, for-loops shouldn't be making it to late resolution anyways

philberty:
if thats the case i would close the PR

PR #4117, opened 2025-08-29
nr2.0: Add extra check when resolving identifier patterns

### LINKED ISSUE #4116: Single segment variants are name resolved to identifiers
```rust
enum Foo {
    Relaxed,
    SeqCst,
    StressedOut,
}

use Foo::{Relaxed, SeqCst, StressedOut};

fn main() {
    let a = (Relaxed, Relaxed);

    match a {
        (Relaxed, Relaxed) => {}
        _ => {}
    }
}
```

we get an error since an identifier is bound multiple times, but this is incorrect.

CohenArthur:
gcc/rust/ChangeLog:

	* resolve/rust-late-name-resolver-2.0.cc (Late::visit): Check if an identifier pattern
	can be resolved in the types namespace before visiting it as an identifier proper.

gcc/testsuite/ChangeLog:

	* rust/compile/issue-4116.rs: New test.

Fixes #4116

### DIFF
--- a/gcc/rust/resolve/rust-late-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-late-name-resolver-2.0.cc
@@ -206,6 +206,17 @@ Late::visit (AST::IdentifierPattern &identifier)
 {
   DefaultResolver::visit (identifier);
 
+  // We need to check if the Identifier resolves to a variant or empty struct
+  auto path = AST::SimplePath (identifier.get_ident ());
+
+  if (auto resolved = ctx.resolve_path (path, Namespace::Types))
+    {
+      ctx.map_usage (Usage (identifier.get_node_id ()),
+		     Definition (resolved->definition.get_node_id ()),
+		     resolved->ns);
+      return;
+    }
+
   visit_identifier_as_pattern (ctx, identifier.get_ident (),
 			       identifier.get_locus (),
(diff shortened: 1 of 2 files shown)


powerboat9:
https://doc.rust-lang.org/reference/patterns.html#path-patterns

It looks like path patterns should only resolve to a few types of items, which seems like it'd fix the testsuite issues

gcc/rust/resolve/rust-late-name-resolver-2.0.cc
```diff
@@ -206,6 +206,16 @@ Late::visit (AST::IdentifierPattern &identifier)
 {
   DefaultResolver::visit (identifier);
 
+  // We need to check if the Identifier resolves to a variant or empty struct
+  auto path = AST::SimplePath (identifier.get_ident ());
+
+  if (auto resolved = ctx.resolve_path (path, Namespace::Types))
```
powerboat9, line 212: This should include the values namespace

powerboat9: Actually, never mind for now, https://godbolt.org/z/P89qdThE1 is producing some interesting results

powerboat9:
@CohenArthur it looks like this is blocking late resolution of libcore

CohenArthur:
Rebasing on top of #4551

PR #4105, opened 2025-08-27
Sync with upstream (2025-08-27): 5de587c1f7afa2f7a619441a7b8887853662a655


dkm:
This is a sync with upstream GCC:
 - upstream GCC revision: 5de587c1f7afa2f7a619441a7b8887853662a655

 - gccrs github: 456dbb4391b3215099795f999240a13b642077fa

-- [gerris](https://github.com/Rust-GCC/gerris) 🦀

PR #4121, opened 2025-09-01
ci: Bump commit to merge in arm64 ci

powerboat9:
This *should* fix the arm CI in #4105


dkm:
Any chance you could add some comment explaining what is needed?

dkm:
So, IIUC, for every upstream bump, we should also bump this revision to keep it as close as possible to the upstream rev?
Looks a bit brittle, but if it works no real reason to complain. Thanks for the additional comment.

dkm:
Let's see if it fixes the current macos failure.

PR #4122, opened 2025-09-01
[WIP] Fix alpine CI

powerboat9:


philberty:
Here thats mental looks like crab1 runs from your logs with no backtrace like! it must be the binutils or something crashing.

You'll need to run the compiler driver with -v to get more info on it

philberty:
you should be able to run this:

./gcc/gccrs -v -B./gcc/ -g -O0 -Wall -frust-debug test.rs -o test

to see where its failing

philberty:
or wait does this revert fix it? https://github.com/Rust-GCC/gccrs/pull/4122/commits/4f048c2fb6d6665eec7c1b8f2f9f9d1e6b85bee4

Issue #4004, opened 2025-07-23
Implement `TokenCollector::visit(AST::FormatArgs&)`


CohenArthur:
The implementation is currently missing but should be pretty easy to add

Magnus-Mage:
Anyone working on this? Otherwise i can open a PR

CastilloDel:
Hey! Just triaging, this has already been solved by https://github.com/Rust-GCC/gccrs/pull/4135, right?

PR #4135, opened 2025-09-08
Implement `TokenCollector::visit(AST::FormatArgs&)` 

Magnus-Mage:
Replace rust_sorry_at() stub with basic implementation that outputs format_args!() macro syntax.
Related to Issue - #4004 

This is WIP so let me know if this is ok for now or need changes.

### DIFF
--- a/gcc/rust/ast/rust-ast-collector.cc
+++ b/gcc/rust/ast/rust-ast-collector.cc
@@ -2997,8 +2997,216 @@ TokenCollector::visit (BareFunctionType &type)
 void
 TokenCollector::visit (AST::FormatArgs &fmt)
 {
-  rust_sorry_at (fmt.get_locus (), "%s:%u: unimplemented FormatArgs visitor",
-		 __FILE__, __LINE__);
+  push (Rust::Token::make_identifier (fmt.get_locus (), "format_args"));
+  push (Rust::Token::make (EXCLAM, fmt.get_locus ()));
+  push (Rust::Token::make (LEFT_PAREN, fmt.get_locus ()));
+
+  std::string reconstructed_template = "\"";
+  const auto &template_pieces = fmt.get_template ();
+
+  for (const auto &piece : template_pieces.get_pieces ())
+    {
+      if (piece.tag == Fmt::ffi::Piece::Tag::String)
+	{
+	  std::string literal = piece.string._0.to_string ();
+	  for (char c : literal)
+	    {
+	      if (c == '"' || c == '\\')
+		{
+		  reconstructed_template += '\\';
+		}
+	      else if (c == '\n')
+		{
+		  reconstructed_template += "\\n";
+		  continue;
+		}
+	      else if (c == '\r')
+		{
+		  reconstructed_template += "\\r";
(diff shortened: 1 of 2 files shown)


gcc/rust/ast/rust-ast-collector.cc
```diff
@@ -3010,8 +3010,16 @@ TokenCollector::visit (BareFunctionType &type)
 void
 TokenCollector::visit (AST::FormatArgs &fmt)
 {
-  rust_sorry_at (fmt.get_locus (), "%s:%u: unimplemented FormatArgs visitor",
-		 __FILE__, __LINE__);
+  // Ouput the macro name
```
powerboat9: Typo -- "Ouput"

Magnus-Mage: I updated it, its fixed now

gcc/rust/ast/rust-ast-collector.cc
```diff
@@ -3010,8 +3010,16 @@ TokenCollector::visit (BareFunctionType &type)
...
 {
-  rust_sorry_at (fmt.get_locus (), "%s:%u: unimplemented FormatArgs visitor",
-		 __FILE__, __LINE__);
+  // Output the macro name
+  push (Rust::Token::make_identifier (UNDEF_LOCATION, "format_args"));
+  push (Rust::Token::make (EXCLAM, UNDEF_LOCATION));
+  push (Rust::Token::make (LEFT_PAREN, UNDEF_LOCATION));
+
+  // Output placeholder format string
+  push (Rust::Token::make (DOUBLE_QUOTE, UNDEF_LOCATION));
```
P-E-P: I don't know if this is still WIP but I would have expected the content of the format string to be pushed here.

Magnus-Mage: @P-E-P  can you check again and confirm if this much is correct for now?

gcc/rust/expand/rust-macro-builtins-format-args.cc
```diff
@@ -58,7 +58,7 @@ format_args_parse_expr (location_t invoc_locus, AST::MacroInvocData &invoc,
 
   // TODO(Arthur): Clean this up - if we haven't parsed a string literal but a
   // macro invocation, what do we do here? return a tl::unexpected?
-  rust_assert (format_expr->is_literal ());
+  // rust_assert (format_expr->is_literal ());
```
CohenArthur: I don't think this needs to be part of this PR

Magnus-Mage:
Thank you for the suggestions.

I have made the necessary changes:

- Removed unnecessary `const_casts`, using const references where possible
- Reconstructed templates using `Fmt::ffi::Piece`, handling literals and placeholders
- Added support for format specifiers position, alignment, flags, width, precision, and type
- Argument handling for normal, named, and captured arguments

Also, tested this manually results [here](https://gist.github.com/877dd226bf40a6497bf9ee469d6a7c67.git)

Do let me know if the workaround is ok or need changes.

P-E-P (approved):

Issue #2886, opened 2024-02-28
Unused parentheses in parameters causes internal compiler error


0xn4utilus:
The unused parentheses in parameters causes internal compiler error however it is perfectly fine in the body.
rustc provides a nice warning and compiles the code. 

### Code

https://godbolt.org/z/xM4Mnvohr

```rust
// error
fn increment(x: i32) -> i32{
    ((x+1))
}

fn increment1((x): (i32)) -> (i32){
    ((x+1))
}

fn increment2(x: i32) -> (i32){
    ((x+1))
}

// correct
fn increment3(x: i32) -> i32{
    ((x+1))
}

```

<details><summary>Backtrace</summary>
  <p>
  
  ```
crab1: internal compiler error: in resolve_path_to_trait, at rust/typecheck/rust-hir-trait-resolve.cc:128
0x236776c internal_error(char const*, ...)
	???:0
0x9fc178 fancy_abort(char const*, int, char const*)
	???:0
0xcfb1b7 Rust::Resolver::TraitResolver::Resolve(Rust::HIR::TypePath&)
	???:0
0xcf4c17 Rust::Resolver::TypeCheckBase::get_predicate_from_bound(Rust::HIR::TypePath&, Rust::HIR::Type*, Rust::BoundPolarity)
	???:0
0xd0eda4 Rust::Resolver::TypeCheckType::visit(Rust::HIR::TraitObjectType&)
	???:0
0xd0b675 Rust::Resolver::TypeCheckType::Resolve(Rust::HIR::Type*)
	???:0
0xd07352 Rust::Resolver::TypeCheckItem::visit(Rust::HIR::Function&)
	???:0
0xd00b90 Rust::Resolver::TypeCheckItem::Resolve(Rust::HIR::Item&)
	???:0
0xcc16bb Rust::Resolver::TypeResolution::Resolve(Rust::HIR::Crate&)
	???:0
0xb37894 Rust::Session::compile_crate(char const*)
	???:0
Please submit a full bug report, with preprocessed source (by using -freport-bug).
Please include the complete backtrace with any bug report.
See <https://gcc.gnu.org/bugs/> for instructions.
Compiler returned: 1
  ```
  
  </p>
</details>

philberty:
This is funny its actually parsing this function wrongly:

```
fn increment1((x): (i32)) -> (i32){
    ((x+1))
}
```

The parameter is getting parsed as:

crab1: note: Preparing to parse files.                                                                          
[...]

PR #4137, opened 2025-09-08
tests: added extra parens compile test

ghost:
Adds a test to verify extra parentheses are discarded, and the source is compiled as expected.

Resolves: #2886, #4148

in the git commit message.

- \[x] Run `clang-format`

order to get feedback.
---

### DIFF
--- a/gcc/testsuite/rust/compile/issue-4148.rs
+++ b/gcc/testsuite/rust/compile/issue-4148.rs
@@ -1,0 +1,26 @@
+// { dg-excess-errors "warnings" }
+
+// TODO: all `xfail` conditions should be changed to `target` once the ICE in #4148 is resolved
+
+pub fn ret_parens(x: i32) -> i32 {
+    // { dg-warning "unnecessary parentheses around block return value" "#4148" { xfail *-*-* } .+1 }
+    ((x+1))
+}
+
+// { dg-warning "unnecessary parentheses around type" "#4148" { xfail *-*-* } .+1 }
+// { dg-warning "unnecessary parentheses around pattern" "#4148" { xfail *-*-* } .+1 }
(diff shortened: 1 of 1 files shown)


CohenArthur:
Thanks for the contribution @0xllx0! Your test is correct but unfortunately gccrs parses it as a trait and does the wrong thing, and errors out :grimacing: could you create an issue for this? You can also work on fixing it within this PR if you want

ghost:
> gccrs parses it as a trait and does the wrong thing, and errors out 😬 could you create an issue for this? You can also work on fixing it within this PR if you want

How do I get the test to fail then? That's what I was expecting, but marking it as a "should fail" test didn't feel right, as the test is expected to eventually pass, but should fail now.

I opened a tracking issue for the test mismatch: #4148 

I'd be happy to work on the fix in this PR.

ghost:
> Your test is correct but unfortunately gccrs parses it as a trait and does the wrong thing, and errors out 😬 could you create an issue for this? You can also work on fixing it within this PR if you want

I think I figured out the problem, just needed to add `dg-warning` directive comments. The `parens` test now fails as expected, and includes checks for the expected warning messages (can be adjusted based on the actual warnings output by `gccrs` after a parsing fix).

powerboat9:
If `gccrs` is behaving incorrectly, I'd think it'd be proper to add a test verifying `gccrs` behaves ~correctly after/while making sure that `gccrs` behaves correctly.

ghost:
> If `gccrs` is behaving incorrectly, I'd think it'd be proper to add a test verifying `gccrs` behaves ~correctly after/while making sure that `gccrs` behaves correctly.

Alright, so while the ICE exists, mark this test as `xfail`, and change the test to check for expected warnings once the fix is in?

PR #4149, opened 2025-09-12
name-resolution: Add std/core prelude resolution

CohenArthur:
- **nr: Add prelude field to NRCtx, and fill it upon encountering a prelude.**
- **forever-stack: Add extra path resolution from a known NodeId.**
- **nr: Do prelude resolution for Identifiers**

This PR adds name resolution within the prelude marked with ``#[prelude_import]``. This is important for compiling core, and for compiling crates in general - as the compiler is supposed to inject the ``#[prelude_import]`` use-declaration if it is not user defined.

The current implementation attempts to resolve within the prelude if regular resolution has failed.

The prelude is collected and stored in the name-resolution context.

### DIFF
--- a/gcc/rust/resolve/rust-name-resolution-context.h
+++ b/gcc/rust/resolve/rust-name-resolution-context.h
@@ -592,25 +592,27 @@ public:
     // If it fails, switch to std prelude resolution if it exists
     if (prelude && !resolved)
       {
+	std::vector<Error> ignore_errors = {};
+
 	// TODO: Factor this with the above
 	switch (ns)
 	  {
 	  case Namespace::Values:
 	    return values.resolve_path (segments, mode,
 					insert_segment_resolution,
-					collect_errors, *prelude);
+					ignore_errors, *prelude);
 	  case Namespace::Types:
 	    return types.resolve_path (segments, mode,
-				       insert_segment_resolution,
-				       collect_errors, *prelude);
+				       insert_segment_resolution, ignore_errors,
+				       *prelude);
 	  case Namespace::Macros:
 	    return macros.resolve_path (segments, mode,
 					insert_segment_resolution,
-					collect_errors, *prelude);
+					ignore_errors, *prelude);
 	  case Namespace::Labels:
 	    return labels.resolve_path (segments, mode,
 					insert_segment_resolution,
-					collect_errors, *prelude);
+					ignore_errors, *prelude);
 	  default:
 	    rust_unreachable ();
 	  }


gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -647,17 +658,50 @@ ForeverStack<N>::resolve_final_segment (Node &final_node, std::string &seg_name,
...
+tl::optional<Rib::Definition>
+ForeverStack<N>::resolve_path (
+  const std::vector<S> &segments, ResolutionMode mode,
+  std::function<void (const S &, NodeId)> insert_segment_resolution,
+  std::vector<Error> &collect_errors,
+  std::reference_wrapper<Node> starting_point)
 {
   rust_assert (!segments.empty ());
 
-  std::reference_wrapper<Node> starting_point = cursor ();
```
powerboat9, line 660: You should probably just add a new variant to `ResolutionMode`, instead of adding a user-provided `starting_point` in addition to `mode`

gcc/rust/resolve/rust-name-resolution-context.h
```diff
@@ -560,23 +560,63 @@ class NameResolutionContext
...
+					insert_segment_resolution,
+					collect_errors, *prelude);
+	  case Namespace::Types:
+	    return types.resolve_path (segments, mode,
+				       insert_segment_resolution,
+				       collect_errors, *prelude);
+	  case Namespace::Macros:
+	    return macros.resolve_path (segments, mode,
+					insert_segment_resolution,
+					collect_errors, *prelude);
```
powerboat9: You should probably use a temporary vector instead of `collect_errors` for the prelude resolution -- if prelude resolution does emit errors and fail, only the errors produced during the normal resolution attempt should be collected.

gcc/rust/resolve/rust-forever-stack.hxx
powerboat9: `ForeverStack` could probably fetch the node id associated with the language prelude from `NameResolutionContext` -- might be kind of tricky to get that working, though. You could also just have each `ForeverStack` keep a `std::reference_wrapper` to the currently set prelude `Node`.

CohenArthur: I'm not sure which I prefer between having the forever stack user (so, the name resolver) providing these starting points like the std prelude or the language prelude, and the foreverstack keeping track of these preludes. to me it makes more sense, conceptually, to keep preludes away from the ForeverStack, but I'm happy to be convinced otherwise.

I did it that way because it was possible but maybe it isn't for the lang prelude?

powerboat9: As long as it works, I suppose this is alright. Resolution is kind of a mess, and it's due for some cleanup anyways

Issue #4145, opened 2025-09-09
ICE `in process_file_path, at rust/ast/rust-ast.cc:3386` , path name attr == mod name


matthiaskrgr:
auto-reduced (treereduce-rust):
````rust
//@compile-flags: -frust-incomplete-and-experimental-compiler-do-not-use
struct S {
    field: [u8; {
        #[path = "outer/inner.rs"]
        mod inner;
        // OK
        0
    }],
}

````

original:
````rust
//@ proc-macro: issue-118809.rs

#[macro_use]
extern crate issue_118809;

#[derive(Empty)]
struct S {
    field: [u8; {
        #[path = "outer/inner.rs"]
        mod inner; //~ ERROR non-inline modules in proc macro input are unstable
        mod inner_inline {} // OK
        0
    }]
}

fn main() {}

````

Version information:

c742973c23448e321fba4d9ff98a9092dc0df7ba

Possibly related line of code:
https://github.com/Rust-GCC/gccrs/blob/c742973c23448e321fba4d9ff98a9092dc0df7ba/gcc/rust/ast/rust-ast.cc#L3380-L3392

Command:
`/home/matthias/vcs/github/gccrs/gccrs-build/gcc/crab1 -frust-incomplete-and-experimental-compiler-do-not-use`

<details><summary><strong>Program output</strong></summary>
<p>

```
/tmp/icemaker_global_tempdir.4jbydGKfmYZn/rustc_testrunner_tmpdir_reporting.jMr7LWNZjpn8/mvce.rs:4:9: error: cannot open module file ./outer/inner.rs: No such file or directory
    4 |         mod inner;
      |         ^~~
crab1: internal compiler error: in process_file_path, at rust/ast/rust-ast.cc:3386
0x300c7d8 internal_error(char const*, ...)
	../../gcc/diagnostic-global-context.cc:534
0xc23693 fancy_abort(char const*, int, char const*)
	../../gcc/diagnostics/context.cc:1640
0x9048e5 Rust::AST::Module::process_file_path()
	../../gcc/rust/ast/rust-ast.cc:3386
0xdea490 Rust::AST::Module::load_items()
	../../gcc/rust/ast/rust-ast.cc:3472
0xeaf627 Rust::CfgStrip::visit(Rust::AST::Module&)
	../../gcc/rust/expand/rust-cfg-strip.cc:1777
0xeaf627 Rust::CfgStrip::visit(Rust::AST::Module&)
[...]

PR #4151, opened 2025-09-12
fix: early return for module file error in AST parsing

ghost:

in the git commit message.

- \[x] Run `clang-format`

order to get feedback.
---

Converts an assert into an early return during AST parsing.

Resolves: https://github.com/Rust-GCC/gccrs/issues/4145

### DIFF
--- a/gcc/testsuite/rust/compile/issue-4145.rs
+++ b/gcc/testsuite/rust/compile/issue-4145.rs
@@ -1,0 +1,13 @@
+// { dg-excess-errors "warnings" }
+
+struct S {
+    field: [u8; {
+        #[path = "outer/inner.rs"]
+        // { dg-warning "error handling module file for .inner." "#4145" { xfail *-*-* } .+1 }
+        mod inner;
+        // OK
+        0
+    }],
+}
+
+fn main() {}


gcc/testsuite/rust/compile/issue-4145.rs
```diff
@@ -0,0 +1,8 @@
+struct S {
```
ghost, line 3: Not sure on the exact decorators needed here, or if this test should go into the `xfail` directory. The test is expected to fail compilation, just with an error message instead of an ICE.

ChangeLog
```diff
@@ -1,3 +1,8 @@
+2025-09-12  Elle Rhumsaa  <elle@weathered-steel.dev>
```
powerboat9: You shouldn't be updating the ChangeLog file -- there's a script which automatically does it on the GCC master branch

gcc/testsuite/rust/compile/issue-4145.rs
powerboat9: `grep` for `dg-error` in the testsuite directory -- that should give you some nice examples for expecting error messages

gcc/rust/ast/rust-ast.cc
```diff
@@ -3383,7 +3383,13 @@ void
 Module::process_file_path ()
 {
   rust_assert (kind == Module::ModuleKind::UNLOADED);
-  rust_assert (module_file.empty ());
+
+  if (module_file.empty ())
+    {
+      rust_error_at (locus, "error handling module file for %s",
```
CohenArthur:
```suggestion
      rust_error_at (locus, "error handling module file for %qs",
```

This will output a quoted string which looks better and allows the user to better see which path is wrong

gcc/testsuite/rust/compile/issue-4145.rs
```diff
@@ -0,0 +1,13 @@
+// { dg-excess-errors "warnings" }
+
+struct S {
+    field: [u8; {
+        #[path = "outer/inner.rs"]
+        // { dg-error "error handling module file for inner" "#4145" { xfail *-*-* } .-1 }
```
CohenArthur:
```suggestion
        // { dg-error "error handling module file for .inner." "#4145" { xfail *-*-* } .-1 }
```

So if you apply the change I suggested above you also need to do this, as the quotes are different depending on your system - e.g. it can be output as `` `inner` `` or `‘inner’`

ChangeLog
P-E-P: @0xllx0 You still need the changelog section within your commit message, the changelog file is generated using the content of your commit message.

P-E-P: A script can generate the changelog structure for you. You can install it for this repository using the script `$TOP_LEVEL/contrib/gcc-git-customization.sh`, it will ask you a bunch of things (you can leave all default values) and then it will install a new git command.

You may then use `git gcc-commit-mklog -s` instead of `git commit -s` when making a commit.

ghost: Thanks, I was looking for documentation on which script to run, and how to do it. Are these steps included in any documentation? If not, I'd be happy to write something up.

P-E-P: It is briefly mentioned [here](https://github.com/Rust-GCC/gccrs/wiki/First-contributions#changelog).

philberty:
You need to fixup the regressions here those tests probably emit a new error now and they need updated accordingly

ghost:
> You need to fixup the regressions here those tests probably emit a new error now and they need updated accordingly

I rebased on the latest master (`32ce955`), and didn't get a failing test for the `gcc/testsuite/rust/compile/issue-4145.rs` test. Is there something specific where you are noticing regressions? Can you please provide reproduction steps so I can work on resolving the issue?

gcc/testsuite/rust/compile/issue-4145.rs
```diff
@@ -0,0 +1,13 @@
+// { dg-excess-errors "warnings" }
+
+struct S {
+    field: [u8; {
+        #[path = "outer/inner.rs"]
+        // { dg-error "error handling module file for .inner." "#4145" { xfail *-*-* } .-1 }
```
powerboat9: I think `.-1` should be `.1`, if you want to catch an error on the next line instead of the previous line

ghost: Thanks, I changed it to `.+1`, and changed `xfail` to `target` since the compiler successfully returns an error (opposed to an ICE). Please let me know if that logic makes sense to you.

P-E-P:
> I rebased on the latest master (`32ce955`), and didn't get a failing test for the `gcc/testsuite/rust/compile/issue-4145.rs` test. Is there something specific where you are noticing regressions? Can you please provide reproduction steps so I can work on resolving the issue?

I can't remember the state of that PR when you first submitted it. This message seems to be out of date (there are multiple tests that are not passing anymore). Do you struggle with it ? Do you need some help ?

You probably need more constraints over your condition. Try executing the test using GDB to understand why it emits an error. Note that the tests may also have a slight chance of being wrong, even though I doubt it.

ghost:
> Do you struggle with it ? Do you need some help ?

I don't know enough at this point about the internals of `gccrs` to determine exactly why the early return is breaking the other tests. The test for `issue-4145.rs` passes, but obviously the other tests break.

Not sure what the best way forward is.

> You probably need more constraints over your condition.

Possibly, but I currently lack the appropriate context for what the test condition should be that captures the motivating edge-case, and doesn't break the rest of the test suite. It will take me some time to get more familiar with the surrounding code.

> Try executing the test using GDB to understand why it emits an error. Note that the tests may also have a slight chance of being wrong, even though I doubt it.

Thanks, that will probably help trace the relevant code paths.

ghost:
> You probably need more constraints over your condition.

Thanks for mentioning this, because it made me review the condition for the early return. Somehow, I inverted the boolean condition at some point. It's fixed now.

Issue #4155, opened 2025-09-13
segfault `clone_pattern` / `Rust::AST::AltPattern::AltPattern`


matthiaskrgr:
auto-reduced (treereduce-rust):
````rust
//@compile-flags: -frust-incomplete-and-experimental-compiler-do-not-use
struct Bug {
    inner: [(); match Vec::new {
        
        f @  |n() => 1
    }],
}
````

original:
````rust
struct Bug {
    inner: [(); match Vec::new {
        n => n().len(),
        f @  |n() => 1
    }],
}
````

Version information:

4db0cf74f64086e2bb2eff029c597239df8442e1

Command:
`/home/matthias/vcs/github/gccrs/gccrs-build/gcc/crab1 -frust-incomplete-and-experimental-compiler-do-not-use`

<details><summary><strong>Program output</strong></summary>
<p>

```
crab1: internal compiler error: Segmentation fault
0x301e249 internal_error(char const*, ...)
	../../gcc/diagnostic-global-context.cc:786
0x199e5df crash_signal
	../../gcc/toplev.cc:323
0xe1f04b Rust::AST::Pattern::clone_pattern() const
	../../gcc/rust/ast/rust-ast.h:1427
0xe1f04b Rust::AST::AltPattern::AltPattern(Rust::AST::AltPattern const&)
	../../gcc/rust/ast/rust-pattern.h:1787
0xe1f04b Rust::AST::AltPattern::clone_pattern_impl() const
	../../gcc/rust/ast/rust-pattern.h:1828
0xe11dbc Rust::AST::Pattern::clone_pattern() const
	../../gcc/rust/ast/rust-ast.h:1427
0xe11dbc Rust::AST::MatchArm::MatchArm(Rust::AST::MatchArm const&)
	../../gcc/rust/ast/rust-expr.h:4898
0xe11dbc Rust::AST::MatchCase::MatchCase(Rust::AST::MatchCase const&)
	../../gcc/rust/ast/rust-expr.h:4986
0xe11dbc void std::_Construct<Rust::AST::MatchCase, Rust::AST::MatchCase const&>(Rust::AST::MatchCase*, Rust::AST::MatchCase const&)
	/usr/include/c++/15.2.1/bits/stl_construct.h:133
[...]

PR #4169, opened 2025-09-14
Protect from errors in alternate pattern parsing

powerboat9:
Fixes https://github.com/Rust-GCC/gccrs/issues/4155


CohenArthur (approved):
Good find!

Issue #4171, opened 2025-09-16
`FFIOpt<T>::get_opt()` method has incorrect return type implementation


Magnus-Mage:
## Summary

This was a unrelated bug i came across while working on #4135 , not sure if it should work like this but implemented a standalone example to mimic it.

The `FFIOpt<T>::get_opt()` method in `gcc/rust/ast/rust-fmt.h` has a fundamental type mismatch that prevents compilation when trying to access optional format specifier fields in FormatArgs AST nodes.

## Problem
The method signature promises to return `tl::optional<std::reference_wrapper<const T>>` but the implementation attempts to return a raw pointer `const T*`. There is no implicit conversion between these types, causing compilation failures.

## Location
**File:** `gcc/rust/ast/rust-fmt.h`  
**Line:** 141  
**Method:** `FFIOpt<T>::get_opt() const`

<details>
  <summary>Current (Broken) Implementation </summary>

```cpp
template<typename T>
tl::optional<std::reference_wrapper<const T>> get_opt() const
{
  return (const T *) inner.data;  // ← ERROR: Cannot convert pointer to reference_wrapper
}
```
</details>

<details>
  <summary>Compilation Error </summary>

```
error: could not convert '(const T*)...' from 'const T*' to 'tl::optional<std::reference_wrapper<const T> >'
```

</details>

## Root Cause

The error occurs because:

1. **Expected return type**: `tl::optional<std::reference_wrapper<const T>>`
2. **Actual return attempt**: Raw pointer `const T*`
3. **Missing conversion**: `std::reference_wrapper<T>` requires a reference (`T&`), not a pointer (`T*`)

<details>
  <summary>Minimal Reproduction Case </summary>

```cpp
#include <utility>
#include <functional>

namespace tl {
  template<typename T>
  class optional {
    bool valid = false;
    alignas(T) char storage[sizeof(T)];
    
  public:
    optional() : valid(false) {}
    
    template<typename U>
    optional(U&& val) : valid(true) {
      new(storage) T(std::forward<U>(val));
    }
    
    bool has_value() const { return valid; }
  };
}

struct MockType { int x; };

template<typename T>
class FFIOpt {
public:
  // Reproduces the exact broken implementation from gccrs
  tl::optional<std::reference_wrapper<const T>> get_opt() const {
    return (const T*) nullptr; // ERROR: Cannot convert pointer to reference_wrapper
  }
};

int main() {
  FFIOpt<MockType> opt;
[...]

PR #4174, opened 2025-09-16
Improve `FFIOpt`

powerboat9:
Also fixes https://github.com/Rust-GCC/gccrs/issues/4171.


powerboat9:
@Magnus-Mage

powerboat9:
I wasn't sure about relying on the behavior described in https://github.com/rust-lang/rfcs/blob/master/text/2195-really-tagged-unions.md when I wrote https://github.com/Rust-GCC/gccrs/pull/4074, since I wasn't sure whether it was standard*-compliant, but according to the RFC it is. At least I think -- it's been a while

gcc/rust/ast/rust-fmt.h
```diff
@@ -83,38 +83,41 @@ template <typename T> class FFIVec
...
+template <typename T,
+	  typename =
+	    typename std::enable_if<std::is_standard_layout<T>::value>::type>
+class FFIOpt
 {
-  struct alignas (T) Inner
-  {
-    char data[sizeof (T)];
-  } inner;
-  bool is_some;
```
Magnus-Mage, line 92: I am not the best user of unions but should keep bool `is_some` for proper tracking of union members otherwise will cause UB and change the rest accordingly?

powerboat9: https://en.cppreference.com/w/cpp/language/union.html says that
> If two union members are [standard-layout](https://en.cppreference.com/w/cpp/named_req/StandardLayoutType.html) types, it's well-defined to examine their common subsequence on any compiler.

so what I've implemented should work

Magnus-Mage: let me check quickly

Magnus-Mage: @powerboat9 you are right, i overlooked that. Thanks for the fix.

Magnus-Mage (approved):

CohenArthur (approved):
LGTM, thank you!

PR #4176, opened 2025-09-17
Test `libcore`

powerboat9:
Depends on #4175 and probably #4099. This patch only tries to compile `libcore` up to the AST validation phase, for now.


powerboat9:
Looks like this depends on #4031 now, and I'll have to have the compiler ignore warnings

powerboat9:
This should be ready to merge now

PR #4184, opened 2025-09-22
ci: new workflow for sending emails after merging a PR

dkm:
This new workflow is used to send emails to the gcc-patches@ mailing list after every merged PR.

For each merged PR:
- remove commits touching files not to-be-upstreamed (e.g. github workflow, README.md, ...)
- if the number of commit is 1, send a single email and add a header explaining what this is about, a link to the PR, ...
- if the number of commits is > 1, then create a cover letter and put the header there.

Authors of commits will be put in Cc: of emails.

Configure it by setting the following secrets in github:
- PATCH_TO: the "To:" field for the emails
- PATCH_CC: optional "Cc:"
- SMTP_FROM, SMTP_PASSWORD, SMTP_PORT, SMTP_SERVER, STMP_USERNAME: self explanatory

### DIFF
--- a/.github/workflows/send-emails.yml
+++ b/.github/workflows/send-emails.yml
@@ -1,0 +1,208 @@
+name: Send emails for merged PR
+
+on:
+  pull_request:
+    types:
+      - closed
+    branches:
+      - "master"
+
+jobs:
+  send_patches:
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repo with full history
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 100
+
+      - name: Install deps (git send-email and jq)
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y jq
+
+          ## GHA images have a special git package that conflicts with
+          ## git-send-email So this is a workaround to force the installation of
+          ## the conflicting package.
+          sudo apt-get install -y libauthen-sasl-perl \
+            libemail-valid-perl \
+            libio-socket-ssl-perl \
+            libmailtools-perl perl \
+            libnet-smtp-ssl-perl
+
+          apt-get download git-email
+          sudo dpkg -i --force-all git-email*deb
+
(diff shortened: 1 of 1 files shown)


dkm:
(I should have already correctly setup the test values for configuring the workflow => the last thing to try it is to merge this PR)

philberty (approved):
LGTM looking forward to seeing our patches in gcc-patches regularly now

dkm:
As a first step, I'll be receiving all the emails. Let's wait a bit before really spa^Wsending emails to gcc ml :sweat_smile:

philberty:
I thinjk the only issue is if we have a PR for like a sync from upstream but i think we should just push that directly to master would get around that.

philberty:
Or we have a filter for only emit emails for stuff that touches gcc/rust or gcc/testsuite/rust or libgrust stuff

dkm:
Yes, sync are pushed directly so should not trigger the workflow :crossed_fingers:

philberty:
But lets just see whats this is like in my opinion.. comment spam :P

dkm:
> Or we have a filter for only emit emails for stuff that touches gcc/rust or gcc/testsuite/rust or libgrust stuff

Hmmm, this can be done. I'm already dropping commit touching files that should not be sent to gcc. We could also check we only send for stuff touching actual rust stuff

.github/workflows/send-emails.yml
```diff
@@ -0,0 +1,208 @@
...
+          fetch-depth: 100
+
+      - name: Install deps (git send-email and jq)
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y jq
+
+          ## GHA images have a special git package that conflicts with
+          ## git-send-email So this is a workaround to force the installation of
+          ## the conflicting package.
```
powerboat9, line 28: This must have been fun to figure out

dkm: Was the first thing that blew up (and I wasn't expecting a fight with a very common package...) "apt install git-send-email" fails miserably...

PR #4196, opened 2025-10-01
ci: fix step condition

dkm:


dkm:
I'll squash all these commits during next upstream sync.

PR #4198, opened 2025-10-02
Test `libcore` compilation until `"nameresolution"`

powerboat9:
This should depend on #4082, or something equivalent


P-E-P:
Looks like this is already on master https://github.com/Rust-GCC/gccrs/blob/master/gcc/testsuite/rust/core/core.exp#L33

PR #4211, opened 2025-10-14
gccrs: add error check if derive has wrong item

lucasly-ba:
Derive may only be applied to structs, enums and unions.

Fixes Rust-GCC#3971

gcc/rust/ChangeLog:

	* expand/rust-expand-visitor.cc(ExpandVisitor::expand_inner_items): Added check and error.

gcc/testsuite/ChangeLog:

	* rust/compile/issue-3971.rs: New test.

Signed-off-by: Lucas Ly Ba <lucas.ly-ba@outlook.com>

### DIFF
--- a/gcc/rust/expand/rust-derive.cc
+++ b/gcc/rust/expand/rust-derive.cc
@@ -25,6 +25,7 @@
 #include "rust-derive-ord.h"
 #include "rust-derive-partial-eq.h"
 #include "rust-derive-hash.h"
+#include "rust-system.h"
 
 namespace Rust {
 namespace AST {
@@ -38,6 +39,16 @@ DeriveVisitor::derive (Item &item, const Attribute &attr,
 		       BuiltinMacro to_derive)
 {
   auto loc = attr.get_locus ();
+
+  using Kind = AST::Item::Kind;
+  auto item_kind = item.get_item_kind ();
+  if (item_kind != Kind::Enum && item_kind != Kind::Struct
+      && item_kind != Kind::Union)
+    {
+      rust_error_at (loc,
+		     "derive may only be applied to structs, enums and unions");
+      return {};
+    }
 
   switch (to_derive)
     {
--- a/gcc/testsuite/rust/compile/issue-3971.rs
+++ b/gcc/testsuite/rust/compile/issue-3971.rs
@@ -1,0 +1,11 @@
+#[lang = "copy"]
+trait Copy {}
+
+// since the macro expansion fails, the current nameres fixpoint error is emitted - just accept it for now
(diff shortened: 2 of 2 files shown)


gcc/rust/expand/rust-expand-visitor.cc
```diff
@@ -180,6 +181,15 @@ ExpandVisitor::expand_inner_items (
...
 		{
+		  if (item.get_item_kind () != AST::Item::Kind::Enum
+		      && item.get_item_kind () != AST::Item::Kind::Struct
+		      && item.get_item_kind () != AST::Item::Kind::Union)
+		    {
+		      rust_error_at (current.get_locus (),
+				     "derive may only be applied to structs, "
+				     "enums and unions");
+		      return;
+		    }
```
CohenArthur: I would put that check in `DeriveVisitor::derive` instead: https://github.com/Rust-GCC/gccrs/blob/b92a68441e51d370f2161704c27fcdd0ae79e454/gcc/rust/expand/rust-derive.cc#L37

The `ExpandVisitor` is already a massive visitor, so I would rather we move specific checks to more specific visitors.

The code itself is great - maybe just create a temporary variable so it's not as long:

```c++
auto item_kind = item.get_item_kind();

if (item_kind != AST::Item::Kind::Struct && item_kind != ...)
    rust_error_at(...);
```

gcc/testsuite/rust/compile/issue-3971.rs
```diff
@@ -0,0 +1,7 @@
+#[derive()]
```
CohenArthur:
```suggestion
#[lang = "copy"]
trait Copy {}

#[derive(Copy)]
```

Add this so the derive code is correct and would actually derive something if it wasn't a function

gcc/rust/expand/rust-expand-visitor.cc
```diff
@@ -18,6 +18,7 @@
 
 #include "rust-expand-visitor.h"
 #include "rust-ast-fragment.h"
+#include "rust-diagnostics.h"
```
P-E-P: There's no modification within rust-expand-visitor.cc, is this include really required ?

gcc/rust/expand/rust-derive.cc
```diff
@@ -39,6 +40,16 @@ DeriveVisitor::derive (Item &item, const Attribute &attr,
 {
   auto loc = attr.get_locus ();
 
+  using kind = AST::Item::Kind;
```
P-E-P: I would use an uppercase `Kind` alias here to avoid confusion.

philberty (approved):
LGTM

PR #4224, opened 2025-10-16
ci: stop using pull_request event, use push

dkm:
We can't access github's secrets from a pull_request event (for security reasons).

Use the push event, and try not to send emails when we should not.


dkm:
Let's see how that works.

PR #4237, opened 2025-10-17
gccrs: make invalid inner attributes show error

lucasly-ba:
Fixes Rust-GCC#4219

gcc/rust/ChangeLog:

	* ast/rust-ast.cc (Attribute::is_derive): Change is_derive method with its valid path.
	* util/rust-attribute-values.h: Delete redudant derive attribute.
	* util/rust-attributes.cc (AttributeChecker::check_inner_attribute): Give can_be_inner a value for each atribute. (AttributeChecker::check_inner_attributes): Implement method for errors check.
	* util/rust-attributes.h (struct BuiltinAttrDefinition): Add can_be_inner object in struct.

gcc/testsuite/ChangeLog:

	* rust/compile/issue-4212.rs:

### DIFF
--- a/gcc/rust/util/rust-attributes.cc
+++ b/gcc/rust/util/rust-attributes.cc
@@ -90,8 +90,6 @@ static const BuiltinAttrDefinition __definitions[]
      {Attrs::PROC_MACRO, EXPANSION},
      {Attrs::PROC_MACRO_DERIVE, EXPANSION},
      {Attrs::PROC_MACRO_ATTRIBUTE, EXPANSION},
-
-     {Attrs::DERIVE, EXPANSION},
      // FIXME: This is not implemented yet, see
      // https://github.com/Rust-GCC/gccrs/issues/1475
      {Attrs::TARGET_FEATURE, CODE_GENERATION},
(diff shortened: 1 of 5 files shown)


gcc/rust/util/rust-attributes.cc
```diff
@@ -67,69 +67,60 @@ using Attrs = Values::Attributes;
...
-     {Attrs::RUSTC_MACRO_TRANSPARENCY, EXPANSION},
-     {Attrs::PATH, EXPANSION},
-     {Attrs::MACRO_USE, NAME_RESOLUTION},
-     {Attrs::MACRO_EXPORT, NAME_RESOLUTION},
-     {Attrs::PROC_MACRO, EXPANSION},
-     {Attrs::PROC_MACRO_DERIVE, EXPANSION},
-     {Attrs::PROC_MACRO_ATTRIBUTE, EXPANSION},
-
-     {Attrs::DERIVE, EXPANSION},
+  = {{Attrs::INLINE, CODE_GENERATION, false},
```
CohenArthur: I think it would be better to add a second map here rather than add a boolean attribute to every BuiltinAttrDefinition

CohenArthur (approved):
I think just add another map instead of reusing the existing one and we will merge this :) good job!

philberty (approved):
LGTM

PR #4239, opened 2025-10-18
Update tuple pattern type-check to attempt to continue resolving items after size error

Polygonalr:
Specifically for case with rest patterns `..`.


philberty (approved):
LGTM

PR #4240, opened 2025-10-19
gccrs: Fix ICE in no input file

### LINKED ISSUE #3523: ICE: segfault when no file is passed
### Meta

local build @ b4895dae9b30413d77024a445af37b3df12d8dfa

### Error output
I tried to run my locally built crab1 but didn't pass a soure file
I got the command form https://github.com/Rust-GCC/gccrs?tab=readme-ov-file#running-gcc-rust
`./gcc/crab1  -frust-debug -frust-dump-ast-pretty -Warray-bounds -dumpbase test.rs -mtune=generic -march=x86-64 -O0 -version -fdump-tree-gimple -o test.s -L/lib/x86_64-linux-gnu -L/lib/../lib64 -L/usr/lib/x86_64-linux-gnu -L/usr/lib/../lib64 -frust-incomplete-and-experimental-compiler-do-not-use -freport-bug`
to my surprise this would already cause a segfault 😅 

<details><summary>Backtrace</summary>
  <p>
  
  ```GNU Rust (GCC) version 14.0.1 20240309 (experimental) (x86_64-pc-linux-gnu)
	compiled by GNU C version 14.2.1 20250207, GMP version 6.3.0, MPFR version 4.2.1, MPC version 1.3.1, isl version isl-0.27-GMP

GGC heuristics: --param ggc-min-expand=30 --param ggc-min-heapsize=4096
options passed: -frust-debug -frust-dump-ast-pretty -mtune=generic -march=x86-64 -O0 -frust-incomplete-and-experimental-compiler-do-not-use -freport-bug
0x14ffe4f crash_signal
	../../gcc/toplev.cc:319
0x28b7d36 lrealpath
	../../libiberty/lrealpath.c:215
0x28b696a canonical_filename_eq
	../../libiberty/filename_cmp.c:216
0xad732d init_asm_output
	../../gcc/toplev.cc:707
0xad732d lang_dependent_init
	../../gcc/toplev.cc:1847
0xad732d do_compile
	../../gcc/toplev.cc:2139
Please submit a full bug report, with preprocessed source.
  
  ```
  
  </p>
</details>

shreyas-omkar:
Fixes #3523

### Problem Description

Invoking the `crab1` front-end without an input source file currently results in an Internal Compiler Error (ICE) due to a segmentation fault. This was caused by two different code paths attempting to process a `NULL` filename pointer.

### Solution

This patch fixes both crash scenarios by making the Rust front-end gracefully default to reading from standard input (`stdin`), aligning its behavior with other GCC front-ends.

1.  **`Session::handle_input_files`** (in `gcc/rust/rust-session-manager.cc`) is updated to handle the `num_files == 0` case by setting the filename to `"-"` (stdin).
2.  **`grs_langhook_post_options`** (in `gcc/rust/rust-lang.cc`) is updated to check for cases where flags are present but no positional file is given, also setting the filename pointer to `"-"` to prevent the original crash in `init_asm_output`.

### Testing

The fix was verified by testing the two scenarios that previously caused a crash. Both now correctly default to `stdin` and exit gracefully without a segfault.

**1. Running `./gcc/crab1` with no arguments:**
(Shows the experimental compiler warning and exits safely)
image

**2. Running `./gcc/crab1` with flags but no file:**
(Correctly defaults to stdin and waits for input)
image
image

Regression testing by compiling a file normally (e.g., `gccrs test.rs`) was also performed and passed, confirming the fix does not break existing functionality. A full `make check-rust` also passes locally.

### DIFF
--- a/gcc/rust/rust-lang.cc
+++ b/gcc/rust/rust-lang.cc
@@ -285,6 +285,10 @@ grs_langhook_post_options (const char **pfilename ATTRIBUTE_UNUSED)
 {
   // can be used to override other options if required
 
+  // check for input file
+  if (!*pfilename && num_in_fnames == 0)
+    *pfilename = "-";
+
   // satisfies an assert in init_excess_precision in toplev.cc
   if (flag_excess_precision /*_cmdline*/ == EXCESS_PRECISION_DEFAULT)
     flag_excess_precision /*_cmdline*/ = EXCESS_PRECISION_STANDARD;
--- a/gcc/rust/rust-session-manager.cc
+++ b/gcc/rust/rust-session-manager.cc
@@ -464,6 +464,12 @@ Session::handle_excluded_node (std::string arg)
 void
 Session::handle_input_files (int num_files, const char **files)
 {
+  static const char *stdin_file[] = {"-"};
+  if (num_files == 0)
+    {
+      files = stdin_file;
+      num_files = 1;
+    }
   if (num_files != 1)
     rust_fatal_error (UNDEF_LOCATION,
 		      "only one file may be specified on the command line");


shreyas-omkar:
@dkm  Please review the changes.

gcc/rust/rust-lang.cc
```diff
@@ -110,6 +110,9 @@ grs_langhook_init (void)
 
   using_eh_for_cleanups ();
 
+  if (num_in_fnames == 0)
+    main_input_filename = "-";
```
CohenArthur: Why is this needed? Is it how the other frontends handle that case?

gcc/rust/rust-lang.cc
```diff
@@ -286,6 +289,10 @@ grs_langhook_post_options (const char **pfilename ATTRIBUTE_UNUSED)
 {
   // can be used to override other options if required
 
+  // check for input file
+  if (!*pfilename && num_in_fnames == 0)
+    *pfilename = "-";
```
CohenArthur, line 290: Likewise here

CohenArthur (commented):
The changes look okay to me but I'm not sure I understand all of them. I don't see why we have to modify `grs_langhook_post_options` and `grs_langhook_init` - could you explain how you arrived to these changes? Did you take inspiration from other frontends? Why are those particular changes?

The changes to `handle_input_files` seem completely fine

gcc/rust/rust-lang.cc
shreyas-omkar: In `grs_langhook_init`, I'm covering the case when the global variable `num_in_fnames` is zero before `main_input_filename` is used.

This is like a safety check if the `grs_langhook_post_options` didnt catch the error.

gcc/rust/rust-lang.cc
shreyas-omkar: In `grs_langhook_post_options`, I'm covering the case when no filename was passed in CLI and the filename pointer is null.

shreyas-omkar:
The changes to `grs_langhook_post_options` and `grs_langhook_init` ensure the Rust frontend handles the “no input file” case the same way as other GCC frontends (e.g., D and Go).

Both hooks are updated because GCC’s initialization sequence references these variables at different stages  post_options handles the option-parsing phase, while init covers later setup before session creation.

If my approach isnt right, I am down for inputs.

gcc/rust/rust-session-manager.cc
```diff
@@ -399,6 +399,12 @@ Session::enable_dump (std::string arg)
 void
 Session::handle_input_files (int num_files, const char **files)
 {
+  if (num_files == 0)
+    {
+      static const char *stdin_file[] = {"-"};
```
philberty: does this work? Doesnt this just cause an fopen on "-"? Shouldnt there be a catch in our lexer for "-" to become stdin or something?

philberty (changes requested):
Not sure i understand this PR have you tried it? like use the stdin stream ?

```echo 'fn main() {}' | gccrs -```

does that actually work?

gcc/rust/rust-session-manager.cc
shreyas-omkar: Okay, so if I am not wrong, `compile_crate(file)` calls the `compile_crate()` function which then creates a `RAIIFile` object from `rust-lex.h`. The constructor of RAIIFile explicitly handles the - case by checking `if (strcmp(filename, "-") == 0)` and redirecting it to stdin instead of calling fopen(). So the stdin redirection is already properly implemented.  So I used this particular part of code.

philberty (changes requested):
Are you sure you dont just need to do handle_input_files if num_in_fnames == 0?

philberty:
can @dkm  and @tschwinge  review this, i am not sure how front ends are meant to respect pipe and no input files what do you think

dkm:
> can @dkm and @tschwinge review this, i am not sure how front ends are meant to respect pipe and no input files what do you think

Yes, I should be able to review this one. Sorry for the delay!

shreyas-omkar:
@dkm 
Screenshot from 2025-11-11 00-25-48

shreyas-omkar:
@dkm Any Changes required??

shreyas-omkar:
Hi @dkm, any changes in the PR?

shreyas-omkar:
Hi @dkm, any changes in the PR?

CohenArthur (approved):
LGTM :)

powerboat9:
@philberty @dkm

PR #4270, opened 2025-11-06
gccrs:fix ICE with continue/break/return in while condition

### LINKED ISSUE #3977: ICE in fold_convert_loc, at fold-const.cc:2662
### Code

```rust
fn main() {
    loop {
        while continue {}
    }
}

```

### Version information

```
5dc8c4d
```

### Command

```
./gcc/crab1 ../test.rs -frust-debug -frust-dump-ast-pretty -Warray-bounds -dumpbase ../test.rs -mtune=generic -march=x86-64 -O0 -version -fdump-tree-gimple -o test.s -L/lib/x86_64-linux-gnu -L/lib/../lib64 -L/usr/lib/x86_64-linux-gnu -L/usr/lib/../lib64 -frust-incomplete-and-experimental-compiler-do-not-use
```

### Program output
```
./test.rs: In function ‘test::main’:
./test.rs:3:9: internal compiler error: in fold_convert_loc, at fold-const.cc:2662
    3 |         while continue {}
      |         ^~~~~
0x29abeb6 internal_error(char const*, ...)
        ../../gccrs/gcc/diagnostic-global-context.cc:517
0xb48036 fancy_abort(char const*, int, char const*)
        ../../gccrs/gcc/diagnostic.cc:1803
0x9618c0 fold_convert_loc(unsigned long, tree_node*, tree_node*)
        ../../gccrs/gcc/fold-const.cc:2662
0x128880a gimple_boolify(tree_node*)
        ../../gccrs/gcc/gimplify.cc:5340
0x12ba21b gimplify_cond_expr
        ../../gccrs/gcc/gimplify.cc:5548
0x129dd43 gimplify_expr(tree_node**, gimple**, gimple**, bool (*)(tree_node*), int)
        ../../gccrs/gcc/gimplify.cc:19743
0x129f8aa gimplify_stmt(tree_node**, gimple**)
        ../../gccrs/gcc/gimplify.cc:8461
0x129c68b gimplify_statement_list
        ../../gccrs/gcc/gimplify.cc:2288
0x129c68b gimplify_expr(tree_node**, gimple**, gimple**, bool (*)(tree_node*), int)
        ../../gccrs/gcc/gimplify.cc:20240
0x129f8aa gimplify_stmt(tree_node**, gimple**)
        ../../gccrs/gcc/gimplify.cc:8461
0x12a0951 gimplify_bind_expr
        ../../gccrs/gcc/gimplify.cc:1683
0x129cb60 gimplify_expr(tree_node**, gimple**, gimple**, bool (*)(tree_node*), int)
        ../../gccrs/gcc/gimplify.cc:19990
0x129f8aa gimplify_stmt(tree_node**, gimple**)
        ../../gccrs/gcc/gimplify.cc:8461
0x129cc8a gimplify_and_add(tree_node*, gimple**)
        ../../gccrs/gcc/gimplify.cc:550
0x129cc8a gimplify_loop_expr
        ../../gccrs/gcc/gimplify.cc:2262
0x129cc8a gimplify_expr(tree_node**, gimple**, gimple**, bool (*)(tree_node*), int)
        ../../gccrs/gcc/gimplify.cc:19994
0x129f8aa gimplify_stmt(tree_node**, gimple**)
        ../../gccrs/gcc/gimplify.cc:8461
0x129c68b gimplify_statement_list
        ../../gccrs/gcc/gimplify.cc:2288
0x129c68b gimplify_expr(tree_node**, gimple**, gimple**, bool (*)(tree_node*), int)
        ../../gccrs/gcc/gimplify.cc:20240
0x129f8aa gimplify_stmt(tree_node**, gimple**)
        ../../gccrs/gcc/gimplify.cc:8461
```
  
  </p>
</details>

Harishankar14:
Fixes ICE when continue/break/return/loop are used as while conditions.

Fixes #3977

The predicate expression must be evaluated before type checking to ensure side effects occur even when the predicate has `Never` type. This prevents skipping function calls, panics, or other side effects in diverging predicates.

### Proof of fix using `-fdump-tree-gimple`

As requested, here is the GIMPLE dump for the test case `while loop {} {}`. It confirms that the infinite loop side-effect is preserved correctly:

```c
__attribute__((cdecl))
struct () test::main ()
{
  struct () D.107;

  <D.101>:
  <D.103>:
  {
    <D.104>:
    {
      <D.102>:
      goto <D.102>; // Side-effect correctly preserved
      if (0 != 0) goto <D.105>; else goto <D.106>;
      <D.106>:
      {

      }
    }
    goto <D.104>;
    <D.105>:
  }
  goto <D.103>;
  return D.107;
}
```

- \[ ] Run `clang-format`

order to get feedback.
---

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -803,7 +803,18 @@ CompileExpr::visit (HIR::WhileLoopExpr &expr)
   ctx->add_statement (loop_begin_label_decl);
   ctx->push_loop_begin_label (loop_begin_label);
 
-  tree condition = CompileExpr::Compile (expr.get_predicate_expr (), ctx);
+  HIR::Expr &predicate = expr.get_predicate_expr ();
+  TyTy::BaseType *predicate_type = nullptr;
+  bool ok
+    = ctx->get_tyctx ()->lookup_type (predicate.get_mappings ().get_hirid (),
+				      &predicate_type);
+  rust_assert (ok && predicate_type != nullptr);
+  tree condition = CompileExpr::Compile (predicate, ctx);
+  if (predicate_type->get_kind () == TyTy::TypeKind::NEVER)
+    {
+      ctx->add_statement (condition);
+      condition = boolean_true_node;
+    }
   tree exit_condition = fold_build1_loc (expr.get_locus (), TRUTH_NOT_EXPR,
 					 boolean_type_node, condition);
   tree exit_expr = Backend::exit_expression (exit_condition, expr.get_locus ());
--- a/gcc/rust/typecheck/rust-hir-type-check-expr.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-expr.cc
@@ -1541,18 +1541,25 @@ void
 TypeCheckExpr::visit (HIR::WhileLoopExpr &expr)
 {
   context->push_new_while_loop_context (expr.get_mappings ().get_hirid ());
-
-  TypeCheckExpr::Resolve (expr.get_predicate_expr ());
+  TyTy::BaseType *predicate_type
+    = TypeCheckExpr::Resolve (expr.get_predicate_expr ());
+  if (predicate_type->get_kind () != TyTy::TypeKind::BOOL
+      && predicate_type->get_kind () != TyTy::TypeKind::NEVER)
+    {
+      rust_error_at (expr.get_predicate_expr ().get_locus (),
+		     "expected boolean expression in %<while%> condition");
+      context->pop_loop_context ();
+      return;
+    }
   TyTy::BaseType *block_expr = TypeCheckExpr::Resolve (expr.get_loop_block ());
-
   if (!block_expr->is_unit ())
     {
       rust_error_at (expr.get_loop_block ().get_locus (),
 		     "expected %<()%> got %s",
 		     block_expr->as_string ().c_str ());
+      context->pop_loop_context ();
       return;
     }
-
   context->pop_loop_context ();
   infered = TyTy::TupleType::get_unit_type ();
 }
@@ -1602,7 +1609,6 @@ TypeCheckExpr::visit (HIR::ContinueExpr &expr)
 		     "%<continue%> outside of a loop");
       return;
     }
-
   infered = new TyTy::NeverType (expr.get_mappings ().get_hirid ());
 }
 
--- a/gcc/testsuite/rust/compile/issue-3977.rs
+++ b/gcc/testsuite/rust/compile/issue-3977.rs
@@ -1,0 +1,65 @@
+// Test for issue #3977 - ICE with continue/break/return in while condition
+
+fn diverge() -> ! {
+    loop {}
+}
+
+fn test_continue() {
+    loop {
+        while continue {}
+    }
+}
+
+fn test_break() {      
+    loop {
+        while break {}
+    }
+}
+
+fn test_return() {
+    loop {
+        while return {}
+    }
+}
+
+fn test_labeled_break() {
+    'outer: loop {
+        loop {
+            while break 'outer {}
+        }
+    }
+}
+
+fn test_labeled_continue() {
+    'outer: loop {
+        loop {
+            while continue 'outer {}
+        }
+    }
+}
+
+fn test_complex_if_else() {
+    loop {
+        while if true { continue } else { break } {}
+    }
+}
+
+fn foo() {
+    while diverge() {
+        break
+    }
+    let _x = 5;
+}
+
+fn main() {
+    // Just reference them so they're "used"
+    if false {
+        test_continue();
+        test_break();
+        test_return();
+        test_labeled_break();
+        test_labeled_continue();
+        test_complex_if_else();
+        foo();
+    }
+}


powerboat9:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=f4db2628116a78ca94e09a3cf1bcfb21

Having a predicate of type `!` should work -- it gets coerced into a boolean. See https://doc.rust-lang.org/reference/types/never.html, though don't get confused by the last paragraph

Harishankar14:
@powerboat9 Thanks for the feedback! I dont really know how did i miss this.. I've updated the fix to properly handle never-type coercion. Instead of rejecting continue as a value, the compiler now:

Detects when the while predicate has type ! (never)
Coerces it to bool by treating it as true (infinite loop)
Issues a warning about unreachable code

Harishankar14:
@powerboat9, @philberty

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -802,8 +802,24 @@ CompileExpr::visit (HIR::WhileLoopExpr &expr)
...
+  bool ok = ctx->get_tyctx()->lookup_type(predicate.get_mappings().get_hirid(), &predicate_type);
+  bool is_diverging = false;
+  if (ok && predicate_type !=nullptr){
+    is_diverging = (predicate_type->get_kind() == TyTy::TypeKind::NEVER 
+                    || predicate_type->get_kind() == TyTy::TypeKind::ERROR);
+  }
+  tree condition;
+  if (is_diverging){
+    condition = boolean_true_node;
+  }else{
```
philberty: the code formatting looks wrong  you need to use clang format

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -802,8 +802,24 @@ CompileExpr::visit (HIR::WhileLoopExpr &expr)
...
   ctx->add_statement (loop_begin_label_decl);
   ctx->push_loop_begin_label (loop_begin_label);
-
-  tree condition = CompileExpr::Compile (expr.get_predicate_expr (), ctx);
+  
+  HIR::Expr &predicate = expr.get_predicate_expr ();
+  TyTy::BaseType *predicate_type = nullptr;
+  bool ok = ctx->get_tyctx()->lookup_type(predicate.get_mappings().get_hirid(), &predicate_type);
+  bool is_diverging = false;
+  if (ok && predicate_type !=nullptr){
```
philberty: i would do rust_assert (ok and != nullptr) this would be an ICE if it happened

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -802,8 +802,24 @@ CompileExpr::visit (HIR::WhileLoopExpr &expr)
...
+    is_diverging = (predicate_type->get_kind() == TyTy::TypeKind::NEVER
```
philberty: not sure why your checking for error surely we just use error_mark_node if that happens?

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -802,8 +802,24 @@ CompileExpr::visit (HIR::WhileLoopExpr &expr)
...
+  HIR::Expr &predicate = expr.get_predicate_expr ();
+  TyTy::BaseType *predicate_type = nullptr;
+  bool ok = ctx->get_tyctx()->lookup_type(predicate.get_mappings().get_hirid(), &predicate_type);
+  bool is_diverging = false;
+  if (ok && predicate_type !=nullptr){
+    is_diverging = (predicate_type->get_kind() == TyTy::TypeKind::NEVER 
+                    || predicate_type->get_kind() == TyTy::TypeKind::ERROR);
+  }
+  tree condition;
+  if (is_diverging){
```
philberty: i think this becomes:

if (predicate_type->is<TyTy::NeverType> ()) 

I think reads better

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -802,8 +802,24 @@ CompileExpr::visit (HIR::WhileLoopExpr &expr)
...
+    condition = boolean_true_node;
+  }else{
+    condition = CompileExpr::Compile(predicate,ctx);
+    if(condition == error_mark_node || TREE_TYPE(condition) == NULL_TREE){
```
philberty: hmm checking for TREE_TYPE null seems odd.. surely its just if error_mark_node? if not you should add a comment

gcc/rust/typecheck/rust-hir-type-check-expr.cc
```diff
@@ -1541,18 +1541,39 @@ void
...
   TyTy::BaseType *block_expr = TypeCheckExpr::Resolve (expr.get_loop_block ());
-
   if (!block_expr->is_unit ())
     {
       rust_error_at (expr.get_loop_block ().get_locus (),
-		     "expected %<()%> got %s",
-		     block_expr->as_string ().c_str ());
+         "expected %<()%> got %s",
+         block_expr->as_string ().c_str ());
+      infered = new TyTy::ErrorType (expr.get_mappings ().get_hirid ());
```
philberty: i think the type check expression defaults infered to be an error type anyways I don't think you need this

gcc/rust/typecheck/rust-hir-type-check-expr.cc
```diff
@@ -1541,18 +1541,39 @@ void
...
+      context->pop_loop_context ();
+      return;
+    }
+  if (block_expr->get_kind () == TyTy::TypeKind::ERROR)
+    {
+      infered = new TyTy::ErrorType (expr.get_mappings ().get_hirid ());
+      context->pop_loop_context ();
```
philberty, line 1560: same here the Resolve call checks for nullptr and makes an error type if you read the top of this file

gcc/rust/typecheck/rust-hir-type-check-expr.cc
```diff
@@ -1541,18 +1541,39 @@ void
 TypeCheckExpr::visit (HIR::WhileLoopExpr &expr)
 {
   context->push_new_while_loop_context (expr.get_mappings ().get_hirid ());
-
-  TypeCheckExpr::Resolve (expr.get_predicate_expr ());
+  TyTy::BaseType *predicate_type
+    = TypeCheckExpr::Resolve (expr.get_predicate_expr ());
+  if (predicate_type->get_kind () == TyTy::TypeKind::ERROR)
+    {
+      infered = new TyTy::ErrorType (expr.get_mappings ().get_hirid ());
```
philberty: yeah just pop and return dont need to explicitly set error type here

philberty (changes requested):
i think this can be simplified a fair bit but good job.

gcc/rust/backend/rust-compile-expr.cc
Harishankar14: Fixed.

gcc/rust/backend/rust-compile-expr.cc
Harishankar14: Fixed.

gcc/rust/backend/rust-compile-expr.cc
Harishankar14: fixed

gcc/rust/backend/rust-compile-expr.cc
Harishankar14: I was overcautious.. i guess.But i agree, with you. It was unwanted.
Fixed it.

gcc/rust/typecheck/rust-hir-type-check-expr.cc
Harishankar14: Fixed.

gcc/rust/typecheck/rust-hir-type-check-expr.cc
Harishankar14: Fixed

gcc/rust/typecheck/rust-hir-type-check-expr.cc
Harishankar14: fixed.

Harishankar14:
@philberty  LMK, if any changes required. Applied all the ones which you pointed out. :)

Harishankar14:
> https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=f4db2628116a78ca94e09a3cf1bcfb21
> 
> Having a predicate of type `!` should work -- it gets coerced into a boolean. See https://doc.rust-lang.org/reference/types/never.html, though don't get confused by the last paragraph

@powerboat9 This should do, lmk your thoughts.

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -803,7 +803,25 @@ CompileExpr::visit (HIR::WhileLoopExpr &expr)
...
+  HIR::Expr &predicate = expr.get_predicate_expr ();
+  TyTy::BaseType *predicate_type = nullptr;
+  bool ok
+    = ctx->get_tyctx ()->lookup_type (predicate.get_mappings ().get_hirid (),
+				      &predicate_type);
+  rust_assert (ok && predicate_type != nullptr);
+  tree condition;
+  if (predicate_type->get_kind () == TyTy::TypeKind::NEVER)
+    {
+      condition = boolean_true_node;
```
powerboat9, line 816: The predicate expression should still be evaluated, as otherwise code that executes but never returns is skipped and continued past.

Harishankar14: Hmm, interesting, i guess even though it returns nothing, might have side effects (like function calls) that should still execute, even though it never returns.I guess that is what you meant right @powerboat9

powerboat9: Yep. For example,

```rust
fn foo() {
    while panic!("stop immediately") {
        break
    }
    never_ever_call_this ();
}
```

Harishankar14: @powerboat9 understood, will fix it and let you know !!

Harishankar14: @powerboat9  I have done the changes, tested it out with few testcases and examples, works !! Sorry for the multiple request notif..Some bug within my side from github

powerboat9: It's alright. It looks like the code is still broken though -- if the condition doesn't actually make it into the final tree output, it won't make it into the final compiled program.

powerboat9 (changes requested):

gcc/rust/backend/rust-compile-expr.cc
Harishankar14: @powerboat9  Ah !! Sorry i pushed the wrong code. 
https://github.com/Rust-GCC/gccrs/pull/4270#discussion_r2531332395
My solution was l
```
if (condition == error_mark_node)
{
    condition = boolean_true_node;
}
else if (predicate_type->get_kind() == TyTy::TypeKind::NEVER)
{
    // The predicate was compiled and added to the tree.
    // For NEVER type, we need to ensure it's in the output, so explicitly add it
    ctx->add_statement(condition);
    condition = boolean_true_node;
}
```


Haven't committed or updated yet.. But yeah go through this and lmk !!

Harishankar14: `fn side_effect() ->i32{
           42
}`

`fn never_returns(_x: i32) -> ! {
    loop {}
}`

`fn test_nested() {
    while never_returns(side_effect()) {
        break;
    }
    // side_effect() must be compiled and evaluated before never_returns()
}`

`fn main() {
    if false {
        test_nested();
    }
}`

Test Output.
image

gcc/rust/typecheck/rust-hir-type-check-expr.cc
```diff
@@ -1609,9 +1608,9 @@ TypeCheckExpr::visit (HIR::ContinueExpr &expr)
     {
       rust_error_at (expr.get_locus (), ErrorCode::E0268,
 		     "%<continue%> outside of a loop");
+      infered = new TyTy::ErrorType (expr.get_mappings ().get_hirid ());
```
powerboat9: Is this necessary?

Harishankar14: @powerboat9  that could be ignored tbh! Do you want me to remove it ??

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -803,7 +803,23 @@ CompileExpr::visit (HIR::WhileLoopExpr &expr)
...
   ctx->push_loop_begin_label (loop_begin_label);
 
-  tree condition = CompileExpr::Compile (expr.get_predicate_expr (), ctx);
+  HIR::Expr &predicate = expr.get_predicate_expr ();
+  TyTy::BaseType *predicate_type = nullptr;
+  bool ok
+    = ctx->get_tyctx ()->lookup_type (predicate.get_mappings ().get_hirid (),
+				      &predicate_type);
+  rust_assert (ok && predicate_type != nullptr);
+  tree condition;
```
philberty: you dont need to declare the variable seperate just do tree condition = CompileExpr::Compile (predicate, ctx)

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -803,7 +803,23 @@ CompileExpr::visit (HIR::WhileLoopExpr &expr)
...
+  condition = CompileExpr::Compile (predicate, ctx);
+  if (condition == error_mark_node)
```
philberty: you can just ignore the case of condition == error mark node... because it will show up in -fdump-tree-gimple, not sure how it could error something else should spit out and error message so i would just do:

```
tree condition = CompileExpr::Compile (predicate, ctx);
if (predicate_type->get_kind() == TyTy::TypeKind::NEVER)
{
    // The predicate was compiled and added to the tree.
    // For NEVER type, we need to ensure it's in the output, so explicitly add it
    ctx->add_statement(condition);
    condition = boolean_true_node;
}
```

Harishankar14: @philberty  Resolved.

gcc/rust/backend/rust-compile-expr.cc
Harishankar14: @philberty Haha yeah, that was ugly, fixed.

Harishankar14:
> nearly there

Yeah completed as per your suggestions,waiting for your review @philberty

philberty:
you need to fix the commit format CI is failing for you.

Harishankar14:
> you need to fix the commit format CI is failing for you.

Yep, done  @philberty

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -803,7 +803,18 @@ CompileExpr::visit (HIR::WhileLoopExpr &expr)
...
+  TyTy::BaseType *predicate_type = nullptr;
+  bool ok
+    = ctx->get_tyctx ()->lookup_type (predicate.get_mappings ().get_hirid (),
+				      &predicate_type);
+  rust_assert (ok && predicate_type != nullptr);
+  tree condition = CompileExpr::Compile (predicate, ctx);
+  if (predicate_type->get_kind () == TyTy::TypeKind::NEVER)
+    {
+      ctx->add_statement (condition);
+      condition = boolean_true_node;
```
philberty, line 816: i dont think this is right to just override to boolean_true_node it means the never expression will never get evaluated

powerboat9: I think `ctx->add_statement` takes care of that

philberty (dismissed):
OK I think this is ready but before we merge can you show an example of -fdump-tree-gimple in the git commit message of what we are doing here for this senario? But also put it in the github description

Harishankar14:
> OK I think this is ready but before we merge can you show an example of -fdump-tree-gimple in the git commit message of what we are doing here for this senario? But also put it in the github description

@philberty  modified it, should be okay ?

Harishankar14:
@P-E-P

PR #4271, opened 2025-11-07
gccrs: fix segfault on empty doc attribute

lucasly-ba:
Fixes Rust-GCC#4266
gcc/rust/ChangeLog:

	* hir/rust-ast-lower-base.cc (ASTLoweringBase::handle_doc_item_attribute): Make error.

gcc/testsuite/ChangeLog:

	* rust/compile/issue-4226.rs: New test.

### DIFF
--- a/gcc/rust/hir/rust-ast-lower-base.cc
+++ b/gcc/rust/hir/rust-ast-lower-base.cc
@@ -820,9 +820,17 @@ void
 ASTLoweringBase::handle_doc_item_attribute (const ItemWrapper &,
 					    const AST::Attribute &attr)
 {
-  auto simple_doc_comment = attr.has_attr_input ()
-			    && attr.get_attr_input ().get_attr_input_type ()
-				 == AST::AttrInput::AttrInputType::LITERAL;
+  if (!attr.has_attr_input ())
+    {
+      rust_error_at (attr.get_locus (),
+		     "attribute must be of the form %qs or %qs",
+		     "#[doc(hidden|inline|...)]", "#[doc = string]");
+      return;
+    }
+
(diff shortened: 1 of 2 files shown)


lucasly-ba:
We also need to fix #[doc = 1] should not work since 1 is not a string

gcc/testsuite/rust/compile/issue-4226.rs
```diff
@@ -0,0 +1,3 @@
+#[doc]
+// { dg-error "attribute must be of the form .#[doc(hidden|inline|...)]. or .#[doc = string]." "" { target *-*-* } .-1 }
```
philberty: you need to excape those characters or use .

gcc/rust/hir/rust-ast-lower-base.cc
```diff
@@ -820,9 +820,18 @@ void
 ASTLoweringBase::handle_doc_item_attribute (const ItemWrapper &,
 					    const AST::Attribute &attr)
 {
-  auto simple_doc_comment = attr.has_attr_input ()
-			    && attr.get_attr_input ().get_attr_input_type ()
-				 == AST::AttrInput::AttrInputType::LITERAL;
+  if (!attr.has_attr_input ())
+    {
+      // FIXME: attr_input must be a string
```
philberty: is this FIXME not handled by emitting an error? what do you mean?

lucasly-ba: I forgot to move this FIXME, but we don't check if ``` #[doc("attribute")] ``` is a string or not. We should emit an error if we have ```#[doc(123)]```

lucasly-ba: Nevermind, i must have been confused with another attribute

PR #4280, opened 2025-11-11
ci: drop macos build

dkm:
It's really getting in the way. It needs special care to be in sync with Iain's own rebased branch.

Until it is easier to use, remove it from our CI.


dkm:
> RIP

Yes, sorry about that. Hopefully we'll be able to revive it later...

PR #4286, opened 2025-11-14
ci: fix reference check

dkm:
Adjust grep command.

### DIFF
--- a/.github/workflows/commit-format.yml
+++ b/.github/workflows/commit-format.yml
@@ -125,16 +125,23 @@ jobs:
           retval=0;
           rev_list="origin/${{ github.event.pull_request.base.ref }}..${{ github.event.pull_request.head.sha }}"
           for commit in $(git rev-list --reverse "$rev_list"); do
-            if [ "$(git log --format=%B -n 1 \"$commit\" | grep '#[0-9]*' | grep -v -i 'Rust-GCC/gccrs#[0-9]*' | wc -l)" -ne 0 ]; then
-              echo "$commit: KO"
+
+            git log --format=%B -n 1 "$commit" > /tmp/commit_log
+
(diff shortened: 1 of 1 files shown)


.github/workflows/commit-format.yml
```diff
@@ -122,10 +122,11 @@ jobs:
 
       - name: Check for issue number reference in commit messages
         run: |
+          set -o pipefail
           retval=0;
           rev_list="origin/${{ github.event.pull_request.base.ref }}..${{ github.event.pull_request.head.sha }}"
           for commit in $(git rev-list --reverse "$rev_list"); do
-            if [ "$(git log --format=%B -n 1 \"$commit\" | grep '#[0-9]*' | grep -v -i 'Rust-GCC/gccrs#[0-9]*' | wc -l)" -ne 0 ]; then
+            if git log --format=%B -n 1 "$commit" | grep '#[0-9]*' | grep -vq -i 'Rust-GCC/gccrs#[0-9]*' ; then
```
powerboat9: This should probably use a negative look-behind, instead of two greps -- as is, this could probably produce false negatives if multiple issues were referenced on the same line.

CohenArthur:
The changes look good to me, I'm not super familiar with the best approach here - but your commit has a small typo :) referernce -> reference

.github/workflows/commit-format.yml
dkm: hmmm, correct... I thought it would be good enough, but as you mention it, I need to fix that :sweat_smile: I'll test some things here, but until it's "ok" I'll leave it as draft.

dkm: Haven't used negative look-behind, but it should catch most of the bad cases...

PR #4294, opened 2025-11-18
Refactor dead code lint 

lucasly-ba:
This patch is simple, it only moves the check of unused static items and unused const items into the dead-code scan visitor.

### DIFF
--- a/gcc/rust/checks/lints/unused/rust-unused-checker.cc
+++ b/gcc/rust/checks/lints/unused/rust-unused-checker.cc
@@ -55,14 +55,6 @@ void
 UnusedChecker::visit (HIR::ConstantItem &item)
 {
   std::string var_name = item.get_identifier ().as_string ();
-  auto id = item.get_mappings ().get_hirid ();
-  if (!unused_context.is_variable_used (id) && var_name[0] != '_')
-    rust_warning_at (item.get_locus (), OPT_Wunused_variable,
-		     "unused variable %qs",
-		     item.get_identifier ().as_string ().c_str ());
-
-  // The unused_visibilities lint: a visibility qualifier on a `const _` item
-  // has no effect.
   if (var_name == "_" && item.get_visibility ().is_public ())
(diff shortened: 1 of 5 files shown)


gcc/rust/checks/lints/unused/rust-unused-checker.cc
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2025-2026 Free Software Foundation, Inc.
```
P-E-P, line 1: You should keep the original changes

lucasly-ba:
@P-E-P @CohenArthur  This should be good

lucasly-ba:
@philberty  this should be good too

lucasly-ba:
@P-E-P

P-E-P (approved):

P-E-P:
@lucasly-ba I believe your commit changelog description is not correct, you're supposed to describe every field, AFAIK you cannot stack a description for multiple functions. You may use the `Likewise` mention though.

powerboat9:
> @lucasly-ba I believe your commit changelog description is not correct, you're supposed to describe every field, AFAIK you cannot stack a description for multiple functions. You may use the `Likewise` mention though.

I think we can stack descriptions, see https://www.gnu.org/prep/standards/html_node/Style-of-Change-Logs.html, it's just that we (?) haven't been doing it.

P-E-P:
> I think we can stack descriptions, see https://www.gnu.org/prep/standards/html_node/Style-of-Change-Logs.html, it's just that we (?) haven't been doing it.

@powerboat9 I'm coming back to that PR because I stumbled upon a comment from @dkm on a similar PR advising for those stacked description. I thought this would break the changelog generator script, but it does not. Let's merge this!

P-E-P:
@lucasly-ba Looks like the CI turned red on `rust/compile/static-mut-refs_0.rs`

P-E-P (approved):

PR #4297, opened 2025-11-20
Remove match arm pattern vector to single pattern

lucasly-ba:
This patch refactors all uses of vector patterns since a vector of patterns would be considered as an alt pattern, a vector is considered useless.


CohenArthur (commented):
looks great!

PR #4299, opened 2025-11-20
ci: refactor ccpp workflow

dkm:
Use dtolnay/rust-toolchain action to install rustc toolchain.
Use a composite action for the various build/check on ubuntu.

More cleanup possible (the 32/64bits builds could be simplified and only
build 1 compiler and split the checks)


dkm:
Thanks @P-E-P for the quick review! :rocket:

Issue #4246, opened 2025-10-21
#[lang] not feature gated


bjorn3:
### Summary

See title

### Reproducer

I tried this code:

```rust
#[lang = "sized"]
trait Sized {}
```

### Does the code make use of any (1.49) nightly feature ?

- [ ] Nightly

### Actual behavior

It compiles

### Expected behavior

It should require `#![feature(lang_items)]`

### GCC Version

gccrs (Compiler-Explorer-Build-gcc--binutils-2.44) 15.2.0

tamaroning:
i will take on this

PR #4304, opened 2025-11-23
#[lang] not feature gated

tamaroning:
Fix https://github.com/Rust-GCC/gccrs/issues/4246


P-E-P (approved):
I wish you provided a test with a failure error when using a lang item without the lang feature enabled. Otherwise LGTM!

PR #4305, opened 2025-11-24
WIP: Handle identifier after for inside impl

lucbf:

Adds 

- \[x] Run `clang-format`

Note: Currently several test cases are failing.
This still doesn't fix Rust-GCC#3608


lucbf:
This is already handled as `// parse type
      std::unique_ptr<AST::Type> type = parse_type ();`

PR #4311, opened 2025-11-26
Handle IDENTIFIER at parse_bare_function_type(

lucbf:
- \[x] Run `clang-format`

This is useful when compiling:
```
impl Foo for
fn my_function() {
//do something
}
```


lucbf:
Mark advised some changes in zulip chat

PR #4316, opened 2025-11-28
Add check for token at parse_bare_function_type

lucbf:
previously the compiler emmited:

> expecting ‘(’ but ‘identifier’ found

now it emmits:

> unexpected token 'identifier' - expected bare function

code example:
```
impl Foo for
fn my_function() {
//do something
}
```

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -9643,7 +9643,7 @@ Parser<ManagedTokenSource>::parse_bare_function_type (
     return nullptr;
 
   auto t = lexer.peek_token ();
-  if (t->get_id () == IDENTIFIER)
+  if (t->get_id () != LEFT_PAREN)
     {
       Error error (t->get_locus (),
 		   "unexpected token %qs - expected bare function",
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -9653,9 +9653,6 @@ Parser<ManagedTokenSource>::parse_bare_function_type (
       return nullptr;
     }
 
-  if (!skip_token (LEFT_PAREN))
-    return nullptr;
-
   // parse function params, if they exist
   std::vector<AST::MaybeNamedParam> params;
   bool is_variadic = false;
--- a/gcc/rust/parse/rust-parse-impl.h
+++ b/gcc/rust/parse/rust-parse-impl.h
@@ -9642,6 +9642,17 @@ Parser<ManagedTokenSource>::parse_bare_function_type (
   if (!skip_token (FN_KW))
     return nullptr;
 
+  auto t = lexer.peek_token ();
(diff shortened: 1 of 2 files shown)


lucbf:
Not sure if I'm doing things right, but I had to delete my previous branch to rebase to the current gccrs

P-E-P:
> Not sure if I'm doing things right, but I had to delete my previous branch to rebase to the current gccrs

No this doesn't feel right. You don't have to delete anything to rebase your work on top of master. If you maintain a master branch on your fork then you can simply fetch it from the main repository and rebase it, either with your local git or online using the "sync fork" button.

lucbf:
I pushed the button and suddenly all the compiler checks failed, but I guess I could just delete my local branch

lucbf:
Alpine build couldn't run properly because It couldn't download gettext-0.22.tar.gz

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -9670,6 +9670,17 @@ Parser<ManagedTokenSource>::parse_bare_function_type (
   if (!skip_token (FN_KW))
     return nullptr;
 
+  auto t = lexer.peek_token ();
+  if (t->get_id () == IDENTIFIER)
```
powerboat9: You should probably replace this condition with `!= LEFT_PAREN`

powerboat9: unless this is just intended to cover the identifier case

lucbf: It definitely wouldn't be right to check for just one case

gcc/rust/parse/rust-parse-impl.h
```diff
@@ -9670,6 +9670,17 @@ Parser<ManagedTokenSource>::parse_bare_function_type (
...
+      Error error (t->get_locus (),
+		   "unexpected token %qs - expected bare function",
+		   t->get_token_description ());
+      add_error (std::move (error));
+
+      return nullptr;
+    }
+
   if (!skip_token (LEFT_PAREN))
     return nullptr;
```
P-E-P: This condition can be removed, we've already checked for a `LEFT_PAREN` and returned a nullptr, we only need to skip it.

powerboat9:
Looks like there are some merge conflicts with #4322, which split `rust-parse-impl.h` into separate files. You'll probably have to manually move your changes into the file that `parse_bare_function_type` was moved into.

P-E-P:
> Looks like there are some merge conflicts with #4322, which split `rust-parse-impl.h` into separate files. You'll probably have to manually move your changes into the file that `parse_bare_function_type` was moved into.

There shouldn't be too many conflicts because I specifically left  type parsing in the main file to avoid conflicts with this PR. Hopefully this won't be too hard @lucbf , if you need any help, please shout!

powerboat9:
@lucbf let me know if you want this reopened

PR #4322, opened 2025-12-03
Split the parser implementation into multiple files

P-E-P:
That 14 kloc file isn't practical.


P-E-P:
@philberty @CohenArthur @powerboat9 I'm not asking for a deep review since this is mostly code moving around. I'd like to get your feelings around this PR. And more specifically the last two commits 96739e328e0d45bf0842cc69c736dfea1f8a37da and cc799826ad835fa42b93d1bce2e15b8f0cd6d569

Do we want to split the parser ? Do we want to keep it as is ?

Do we want to rename the template implementation file to make sure it won't be included like any other header files (there is no header guard) ?

CohenArthur:
I don't have strong feelings one way or another. does this help with things like clangd/other LSPs to work on the parser? if yes then I'm more than happy to split it

P-E-P:
> I don't have strong feelings one way or another. does this help with things like clangd/other LSPs to work on the parser? if yes then I'm more than happy to split it

If I remember correctly the LSP was indeed giving up at the end of the file but I think this has been resolved with a recent clangd release.

powerboat9 (approved):
LGTM

PR #4344, opened 2025-12-31
core: Add -frust-compile-core flag

CohenArthur:
Add a specific flag for compiling core which changes compiler generated paths. This is important for derives, and especially for core, where we want to implement crate::marker::Copy rather than core::marker::Copy.

gcc/rust/ChangeLog:

	* lang.opt: Add -frust-compile-core option.
	* ast/rust-ast-builder.h: Add get_path_start method which depends on -frust-compile-core
	* ast/rust-ast-builder.cc (Builder::discriminant_value): Use it.
	* expand/rust-derive-debug.cc (DeriveDebug::stub_debug_fn): Likewise.
	* expand/rust-derive-default.cc (DeriveDefault::default_call): Likewise.
	* expand/rust-derive-eq.cc (DeriveEq::assert_param_is_eq): Likewise.
	* expand/rust-derive-hash.cc (DeriveHash::hash_call): Likewise. (DeriveHash::hash_fn): Likewise.
	* expand/rust-derive-ord.cc (DeriveOrd::cmp_call): Likewise. (DeriveOrd::cmp_fn): Likewise. (DeriveOrd::make_equal): Likewise. (DeriveOrd::recursive_match): Likewise.
	* expand/rust-expand-format-args.cc (format_arg): Likewise. (expand_format_args): Likewise.

### DIFF
--- a/gcc/rust/lang.opt
+++ b/gcc/rust/lang.opt
@@ -237,4 +237,8 @@ frust-unused-check-2.0
 Rust Var(flag_unused_check_2_0)
 Use the new unused variable check implementation.
 
+frust-compile-core
+Rust Var(flag_compile_core)
+Tell the compiler that we are currently compiling the core library
+
 ; This comment is to ensure we retain the blank line above.
--- a/gcc/rust/ast/rust-ast-builder.cc
+++ b/gcc/rust/ast/rust-ast-builder.cc
@@ -541,7 +541,8 @@ std::unique_ptr<Stmt>
 Builder::discriminant_value (std::string binding_name, std::string instance)
 {
   auto intrinsic = ptrify (
-    path_in_expression ({"core", "intrinsics", "discriminant_value"}, true));
+    path_in_expression ({get_path_start (), "intrinsics", "discriminant_value"},
+			true));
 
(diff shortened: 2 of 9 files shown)


gcc/rust/lang.opt
```diff
@@ -237,4 +237,8 @@ frust-unused-check-2.0
 Rust Var(flag_unused_check_2_0)
 Use the new unused variable check implementation.
 
+frust-compile-core
```
P-E-P, line 240: We could maybe rely on `#[no_core]` instead.

powerboat9: It might be worth it to keep the `::core::...` vs `crate::...` decision separate from `#![no_core]`, until we can figure out exactly how rustc is doing it. From what I can tell, rustc offloads the problem to `$crate` resolution somehow.

powerboat9:
@CohenArthur are you still working on this?

CohenArthur:
I think @P-E-P is right that this should be done based on `#[no_core]`, so this needs to be reworked (and possibly the testsuite as well). I'll keep it in the back of my mind but it's not high priority right now

PR #4345, opened 2025-12-31
nr: Add better error to unresolved attribute macro

CohenArthur:
gcc/rust/ChangeLog:

	* resolve/rust-early-name-resolver-2.0.cc (Early::visit_attributes): Explicit the name of the attribute macro that hasn't been found.


powerboat9:
> Explicit the name of the attribute macro that hasn't been found
Lol. Looks good otherwise though

PR #4347, opened 2026-01-01
gccrs: Fix rogue macro error during lowering on expansion failure

### LINKED ISSUE #4213: rogue macro detected during lowering
Macro that shouldn't be there detected during lowering.

```rust
macro_rules! inner {
    () => {
        $crate::
    }
}

pub fn main() {
    inner!();
}
```
Program output
```
a.rs:8:5: fatal error: rogue macro detected during lowering
    8 |     inner!();
      |     ^~~~~
compilation terminated.
```

Harishankar14:
When a macro expansion fails (e.g. due to a parsing error like invalid syntax in the macro body), the expander previously returned an error fragment but did not update the AST. This left the original macro invocation in place, which subsequently caused an ICE (rogue macro detected) during the lowering phase.

This patch updates `expand_invoc` to replace the macro invocation with an empty fragment if expansion fails, ensuring the compiler can proceed (or exit gracefully) without crashing.

Fixes #4213

gcc/rust/ChangeLog:

	* expand/rust-macro-expand.cc (MacroExpander::expand_invoc): Handle error fragments by replacing them with empty fragments.

gcc/testsuite/ChangeLog:

	* rust/compile/issue-4213.rs: New test.

in the git commit message.

- \[ ] Run `clang-format`

order to get feedback.
---

### DIFF
--- a/gcc/rust/expand/rust-macro-expand.cc
+++ b/gcc/rust/expand/rust-macro-expand.cc
@@ -327,7 +327,14 @@ MacroExpander::expand_invoc (AST::MacroInvocation &invoc,
   else
     fragment
       = expand_decl_macro (invoc.get_locus (), invoc_data, *rdef, semicolon);
-
+  // fix: if the expansion is failing, we must replace the marco with an empty
+  // error or node
+  // makes sure that it doesn't panic on Rouge macro (it -> Lowering Phase)
+  // added the parsing errors in gcc/testsuite/rust/compile/issue-4213.rs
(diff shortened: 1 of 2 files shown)


powerboat9:
Looks pretty good, but the parsing errors should get emitted -- see https://godbolt.org/z/P3vae8En6

Harishankar14:
@powerboat9  thanks for the mention, will provide a fix for it, completely forgot about it  (was completing this in Airport lol)

PR #3799, opened 2025-05-20
AST: Add `reconstruct_type()` method

CohenArthur:
- **ast: reconstruct: Add base for reconstructing and asserting different IDs**
- **ast: Add reconstruct_type() method**
- **ast: builder: Remove ASTTypeBuilder**

  This is the first step of adding the `reconstruct_*()` methods as discussed on Zulip. The goal is to have something similar to `clone_*()`, but each time reconstructing the node and assigning it a different `NodeID` - and modifying `clone_*()` where necessary to ensure the node ID stays the same. Subsequently, this also deletes the `ASTTypeBuilder` visitor as it is replaced by simple calls to `.reconstruct_type()` instead.

PR #4350, opened 2026-01-03
Add wrapper class for automatic node cloning

powerboat9:
This patch adds a wrapper class called `Cloneable`, which can be used to automatically perform polymorphic copying. This should greatly reduce the need for error-prone user defined copy constructors and assignment operators.

As a demonstration/first step, this patch also uses Cloneable to simplify the pattern nodes described in `ast/rust-pattern.h`.

### DIFF
--- a/gcc/rust/ast/rust-ast.h
+++ b/gcc/rust/ast/rust-ast.h
@@ -26,6 +26,7 @@
 #include "rust-location.h"
 #include "rust-diagnostics.h"
 #include "rust-keyword-values.h"
+#include "rust-cloneable.h"
 
 namespace Rust {
 // TODO: remove typedefs and make actual types for these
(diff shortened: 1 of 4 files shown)


powerboat9:
(this is partially yak shaving related to #4323, partially a follow up on #1914)

CohenArthur:
I haven't really looked at the PR yet @powerboat9, does it tie in with something like https://github.com/Rust-GCC/gccrs/pull/3799?

powerboat9:
Not really, though it would be compatible. However, on a side note, #4323 does tie in/replace the reconstruct mechanism. This PR should provide a way to remove virtually all user defined copy constructors/assignment operators.

powerboat9:
@CohenArthur thoughts?

powerboat9:
@CohenArthur

PR #4353, opened 2026-01-03
util/attributes: error on malformed #[link_name] input

Pasta-coder:
Emit a diagnostic when #[link_name] is used without arguments, matching rustc behavior. This prevents silent acceptance of empty attributes and provides a helpful diagnostic that shows the expected form.

Fixes Rust-GCC#4228

gcc/rust/ChangeLog:

	* util/rust-attributes.cc: Emit diagnostic.

gcc/testsuite/ChangeLog:

	* rust/compile/link_name-malformed.rs: New test.


dkm:
nitpick: the changelog is not correct, it only points at the file, not the modified function...

Issue #3875, opened 2025-07-05
ICE in go, at rust/expand/rust-derive-clone.cc:119


sososopy:
### Code

```rust
#[derive(Clone)]
pub trait Map{}
```

### Version information:

```
a1a56c6
```
### Command:
```
./gcc/crab1 ../test.rs -frust-debug -frust-dump-ast-pretty -Warray-bounds -dumpbase ../test.rs -mtune=generic -march=x86-64 -O0 -version -fdump-tree-gimple -o test.s -L/lib/x86_64-linux-gnu -L/lib/../lib64 -L/usr/lib/x86_64-linux-gnu -L/usr/lib/../lib64 -frust-incomplete-and-experimental-compiler-do-not-use
```

### Program output

```
GNU Rust (GCC) version 16.0.0 20250604 (experimental) (x86_64-pc-linux-gnu)
        compiled by GNU C version 11.4.0, GMP version 6.2.1, MPFR version 4.1.0, MPC version 1.2.1, isl version none
GGC heuristics: --param ggc-min-expand=30 --param ggc-min-heapsize=4096
options passed: -frust-debug -frust-dump-ast-pretty -mtune=generic -march=x86-64 -O0 -frust-incomplete-and-experimental-compiler-do-not-use
crab1: note: Preparing to parse files. 
crab1: note: Attempting to parse file: ../test.rs
crab1: note: finished parsing new delim token tree - peeked token is now ']' while t is ')'
crab1: note: inferred crate name: test
crab1: note: SUCCESSFULLY PARSED CRATE 
crab1: note: ran register_plugins (with no body)
crab1: note: SUCCESSFULLY REGISTERED PLUGINS 
crab1: note: started injection
crab1: note: finished injection
crab1: note: SUCCESSFULLY FINISHED INJECTION 
crab1: note: started expansion
crab1: note: inserting link: Link(5 [Map]): existed? no
crab1: note: inserting link: Link(5 [Map]): existed? no
crab1: note: inserting link: Link(5 [Map]): existed? no
crab1: note: inserting link: Link(6 [<anon>]): existed? no
crab1: note: inserting link: Link(6 [<anon>]): existed? no
crab1: note: inserting link: Link(6 [<anon>]): existed? no
crab1: note: popping link
crab1: note: popping link
crab1: note: new cursor: k: Self, v: (S)[] (NS)[6] (G)[]
crab1: note: popping link
crab1: note: inserting link: Link(5 [<anon>]): existed? no
[...]

PR #4359, opened 2026-01-06
expand: Add error check if derive has wrong item

Pasta-coder:
The DeriveVisitor previously dispatched calls to specific handlers (like DeriveClone) without validating the input item type. This caused internal compiler errors (ICEs) when users attempted to derive traits on invalid items like functions or traits, as the handlers assumed the input was a struct, enum, or union.

This patch adds a validation predicate  to the DeriveVisitor to reject invalid items gracefully with a diagnostic error before dispatching.

Fixes Rust-GCC#3875

gcc/rust/ChangeLog:

	* expand/rust-derive.h: Add validate_item prototype.
	* expand/rust-derive.cc (DeriveVisitor::derive): Use visitor validation.
	(DeriveVisitor::validate_item): New function to check item kind.

gcc/testsuite/ChangeLog:

	* rust/compile/derive-on-invalid-item.rs: New test.

- \[x] Run `clang-format`

### DIFF
--- a/gcc/testsuite/rust/compile/issue-3875.rs
+++ b/gcc/testsuite/rust/compile/issue-3875.rs
@@ -1,0 +1,28 @@
+#![crate_type = "lib"]
+#![feature(no_core)]
+#![feature(lang_items)]
+#![no_core]
+// { dg-options "-w" }
+
+#[lang = "sized"]
+pub trait Sized {}
+
+#[lang = "clone"]
+pub trait Clone: Sized {
+    fn clone(&self) -> Self;
+}
+
+#[lang = "copy"]
+pub trait Copy: Clone {}
+
+#[derive(Clone)] // { dg-error "derive may only be applied to structs, enums and unions" }
+fn foo() {}
+
+#[derive(Clone)] // { dg-error "derive may only be applied to structs, enums and unions" }
+trait Bar {}
+
+// Valid test: Derive Copy and Clone on a unit struct
+#[derive(Copy, Clone)]
+struct TupleStruct;
+
+// { dg-prune-output "could not resolve" }


dkm:
Hello, thanks for your contribution!

Small nitpick on the Changelog part...

```
* expand/rust-derive.h: Add validate_item prototype.
```
Does not target a specific "item". You could be more precise using:

```
* expand/rust-derive.h (DeriveVisitor::validate_item): New.
```

Pasta-coder:
@dkm hello,  

commit message regarding changelog changed .

thanks

Pasta-coder:
I realized there might be a potential false positive with `Tuple Structs` (e.g., struct `Foo(i32);`).

If `AST::Item::Kind` differentiates between named `Structs` and `TupleStructs`, my current whitelist might incorrectly reject valid tuple structs.

I am going to verify this locally by adding a test case for a tuple struct. If it fails, I will update `validate_item` to explicitly include `Kind::TupleStruct` and push the fix along with the test.

dkm:
> @dkm hello,
> 
> commit message regarding changelog changed .
> 
> thanks

Thanks!

Pasta-coder:
@dkm hello,

I reverted the extra Kind check since `TupleStruct` falls under `Kind::Struct`. I kept the test case to verify that tuple structs are correctly accepted. 

Open for review .

thankyou

Pasta-coder:
@P-E-P please review .

thankyou .

Pasta-coder:
@CohenArthur please review .

thankyou .

Pasta-coder:
introducing helper function  does have a side effect . 

why i did what i did ? the ICE was solved in #4211  but issue #3875  was still open with comments suggesting a helper function , also test was missing `TupleStructs` check . so , i opened a pr .

1) should i remove the helper function or return `bool` from helper and emit error from caller ? (do you prefer such helpers?)
2) or should this pr be solely for adding a new test ?

thankyou

P-E-P:
> introducing helper function does have a side effect .
> 
> why i did what i did ? the ICE was solved in #4211 but issue #3875 was still open with comments suggesting a helper function , also test was missing `TupleStructs` check . so , i opened a pr .
> 
>     1. should i remove the helper function or return `bool` from helper and emit error from caller ? (do you prefer such helpers?)
> 
>     2. or should this pr be solely for adding a new test ?
> 
> 
> thankyou

I think we should definitely keep the test, but I wouldn't keep the rest of the changes.

Pasta-coder:
@P-E-P now this pr only adds the test .

thankyou.

gcc/testsuite/rust/compile/issue-3875.rs
```diff
@@ -0,0 +1,14 @@
+#![feature(no_core)]
+#![no_core]
+// { dg-options "-w" }
+
+#[derive(Clone)] // { dg-error "derive may only be applied to structs, enums and unions" }
+fn foo() {}      // { dg-excess-errors "" }
```
P-E-P: I wonder, maybe you could remove those `dg-excess-errors` if you declare `Clone`. Have you tried putting those traits ?
```suggestion
#![feature(lang_items)]
fn foo() {}
#[lang = "sized"]
pub trait Sized {}
#[lang = "clone"]
pub trait Clone: Sized {}
```

Excess errors should usually be avoided as they'll catch any errors, and this means that if the compiler's behavior change in the future they won't catch an error change and things may break silently.

Pasta-coder:
@P-E-P this can also be merged .

review once , thanks .

edit - didn't see you self requested a review .

Pasta-coder:
@P-E-P this pr is getting stale , any chance of merging it ?

P-E-P:
> @P-E-P this pr is getting stale , any chance of merging it ?

Sorry I was in vacation and did not see your message until now. I'll take a look.

PR #4369, opened 2026-01-08
testsuite: Adds test from issue 1245

### LINKED ISSUE #1245: LiteralPatterns lose their minus prefix
(Note: this came up while testing #1244, below case will not compile without support for integer primitives in match, but it is a general problem.)

I tried this code:

```rust
extern "C" {
    fn printf(s: *const i8, ...);
}

fn foo (x: i32) {
    match x {
        -2 => {
            let a = "minus two!\n\0";
            let b = a as *const str;
            let c = b as *const i8;
            printf (c);
        },
        _ => {
            let a = "else\n\0";
            let b = a as *const str;
            let c = b as *const i8;
            printf (c);
        }
    }
}

fn main () {
    foo (-2);
    foo (2);
}
```

```
minus two!
else
```

Instead, this happened: 
```
else
minus two!
```

Looking at the pattern in the AST, we have no information recording that the literal actually has a `-` prefix
```
Breakpoint 6, Rust::HIR::ASTLoweringPattern::visit (this=0x7fffffffcb00, pattern=...) at ../../gccrs/gcc/rust/hir/rust-ast-lower-pattern.cc:201
(gdb) p pattern
$4 = (Rust::AST::LiteralPattern &) @0x32b0be0: {<Rust::AST::Pattern> = {
    _vptr.Pattern = 0x2531a50 <vtable for Rust::AST::LiteralPattern+16>}, lit = {value_as_string = {
      static npos = 18446744073709551615, 
      _M_dataplus = {<std::allocator<char>> = {<__gnu_cxx::new_allocator<char>> = {<No data fields>}, <No data fields>}, _M_p = 0x32b0bf8 "2"}, _M_string_length = 1, {
        _M_local_buf = "2\000\377\377\377\177\000\000@\377\332\000\000\000\000", 
        _M_allocated_capacity = 140737488289842}}, type = Rust::AST::Literal::INT, 
    type_hint = Rust::CORETYPE_STR}, locus = {gcc_loc_ = 33120}, node_id = 22}
```
So after lowering etc. when compiling the pattern for the match, we build a case which actually matches `2` not `-2`. 

### Meta

- What version of Rust GCC were you using, git sha if possible. master with #1244

Villosse:
Fixes #1245

As the issue as already been solved, i added the test from the issue.

The issue can be closed.

### DIFF
--- a/gcc/testsuite/rust/execute/issue-1245.rs
+++ b/gcc/testsuite/rust/execute/issue-1245.rs
@@ -1,0 +1,31 @@
+// { dg-output "minus two!\r*\nelse\r*\n" }
+extern "C" {
+    fn printf(s: *const i8, ...);
+}
+
+fn foo(x: i32) {
+    match x {
+        -2 => {
+            let a = "minus two!\n\0";
+            let b = a as *const str;
+            let c = b as *const i8;
+            unsafe {
+                printf(c);
+            }
+        }
+        _ => {
+            let a = "else\n\0";
+            let b = a as *const str;
+            let c = b as *const i8;
+            unsafe {
+                printf(c);
+            }
+        }
+    }
+}
+
+fn main() -> i32 {
+    foo(-2);
+    foo(2);
+    0
+}


gcc/testsuite/rust/execute/issue-1245.rs
```diff
@@ -0,0 +1,29 @@
...
+}
+
+fn foo(x: i32) {
+    match x {
+        -2 => {
+            let a = "minus two!\n\0";
+            let b = a as *const str;
+            let c = b as *const i8;
+            unsafe {
+                printf(c);
```
P-E-P, line 13: We probably need to put this within the execute directory and match the output to ensure the minus is indeed kept correctly.

P-E-P (changes requested):

gcc/testsuite/rust/execute/issue-1245.rs
Villosse: Done !
Waiting for pipeline to pass

Should I keep the first commit or not ?

powerboat9: You should probably squash them together

P-E-P (approved):

Issue #3526, opened 2025-03-24
Issue with column location


powerboat9:
[...]

PR #4372, opened 2026-01-10
FAKE - testsuite: Add regression test for issue #3526

Pasta-coder:
This patch adds a regression test to verify the location of the error diagnostic for foreign items with generic parameters (E0044).

Issue #3526 reported that the diagnostic was incorrectly pointing to the function identifier instead of the generic parameter list. This test ensures that the error is reported at the correct location.

Fixes Rust-GCC#3526

gcc/testsuite/ChangeLog:

	* rust/compile/issue-3526.rs: New test.


Pasta-coder:
was used to reproduce a bug , closed due to bad commit history .

PR #4374, opened 2026-01-10
typecheck: Fix location and add help for E0044 foreign generics

Pasta-coder:
Related to Rust-GCC#3526


Pasta-coder:
while looking at the refactored code i came across that the primary reason the error was misleading was that the TypeCheck pass was explicitly reading the location of the `ExternalFunctionItem` (the `fn` keyword) instead of the `GenericParam` (the `<T>`).


the lexer issue still exists .

this doesn't not resolve issue #3526 .

gcc/rust/typecheck/rust-hir-type-check-base.cc
```diff
@@ -555,6 +555,10 @@ TypeCheckBase::resolve_generic_params (
 	      {
 		rust_error_at (generic_param->get_locus (), ErrorCode::E0044,
 			       "foreign items may not have const parameters");
+
```
philberty, line 572: i think you should be using a rich_location type error for something like this

Pasta-coder:
@philberty what do you think about these changes ?

thankyou .

Pasta-coder:
@P-E-P any reviews on this ?

thankyou

gcc/testsuite/rust/compile/issue-3526.rs
P-E-P: This line is not required, it is automatically injected by the testsuite.

P-E-P (changes requested):

gcc/testsuite/rust/compile/issue-3526.rs
Pasta-coder: okay

Pasta-coder:
cc @P-E-P

Issue #4161, opened 2025-09-13
tree check: expected class ‘type’, have ‘exceptional’ (error_mark) in visit, at rust/backend/rust-compile-pattern.cc:1255


matthiaskrgr:
auto-reduced (treereduce-rust):
````rust
//@compile-flags: -frust-incomplete-and-experimental-compiler-do-not-use
type X = ();

fn main() {
    let () = ::X = ();
}

````

original:
````rust
type X = ();

fn main() {
    let ()=::X = ();
}
````

Version information:

4db0cf74f64086e2bb2eff029c597239df8442e1

Command:
`/home/matthias/vcs/github/gccrs/gccrs-build/gcc/crab1 -frust-incomplete-and-experimental-compiler-do-not-use`

<details><summary><strong>Program output</strong></summary>
<p>

```
crab1: internal compiler error: tree check: expected class ‘type’, have ‘exceptional’ (error_mark) in visit, at rust/backend/rust-compile-pattern.cc:1255
0x301e249 internal_error(char const*, ...)
	../../gcc/diagnostic-global-context.cc:786
0xa8cfdd tree_class_check_failed(tree_node const*, tree_code_class, char const*, int, char const*)
	../../gcc/tree.cc:9234
0x90ea74 tree_class_check(tree_node*, tree_code_class, char const*, int, char const*)
	../../gcc/tree.h:3888
0x90ea74 Rust::Compile::CompilePatternLet::visit(Rust::HIR::TuplePattern&)
	../../gcc/rust/backend/rust-compile-pattern.cc:1255
0x112add6 Rust::Compile::CompilePatternLet::Compile(Rust::HIR::Pattern*, tree_node*, Rust::TyTy::BaseType*, unsigned long, Rust::Compile::Context*)
	../../gcc/rust/backend/rust-compile-pattern.h:123
0x112add6 Rust::Compile::CompileStmt::visit(Rust::HIR::LetStmt&)
	../../gcc/rust/backend/rust-compile-stmt.cc:91
0x112ae8c Rust::Compile::CompileStmt::Compile(Rust::HIR::Stmt*, Rust::Compile::Context*)
	../../gcc/rust/backend/rust-compile-stmt.cc:36
0x115c84c Rust::Compile::HIRCompileBase::compile_function_body(tree_node*, Rust::HIR::BlockExpr&, Rust::TyTy::BaseType*)
	../../gcc/rust/backend/rust-compile-base.cc:607
[...]

PR #4376, opened 2026-01-13
gccrs: fix(type-check): Resolve 'exceptional' error_mark in visit pattern compilation

Harishankar14:
This patch fixes two Internal Compiler Errors (ICEs) encountered during compilation.

First, it resolves a crash in `CompilePatternLet::visit` where the backend panicked upon encountering an `error_mark` node within a tuple pattern. A check was added to handle erroneous types gracefully.

Second, it fixes a crash in the `MarkLive` lint pass. Previously, visiting a `TypeAlias` could trigger `rust_unreachable()` if the AST-to-HIR mapping lookup failed. This assertion has been removed to allow the compiler to continue execution when the mapping is missing.

Fixes Rust-GCC/gccrs#4161

gcc/rust/ChangeLog:

	* backend/rust-compile-pattern.cc (CompilePatternLet::visit): Check for error_mark_node before processing tuple patterns.
	* checks/lints/rust-lint-marklive.cc (MarkLive::visit): Remove rust_unreachable() assertion to prevent ICE on missing mappings.

gcc/testsuite/ChangeLog:

	* rust/compile/issue-4161.rs: New test.

in the git commit message.

- \[ ] Run `clang-format`

order to get feedback.
---

### DIFF
--- a/gcc/rust/backend/rust-compile-stmt.cc
+++ b/gcc/rust/backend/rust-compile-stmt.cc
@@ -74,6 +74,17 @@ CompileStmt::visit (HIR::LetStmt &stmt)
   tree init = CompileExpr::Compile (stmt.get_init_expr (), ctx);
   // FIXME use error_mark_node, check that CompileExpr returns error_mark_node
   // on failure and make this an assertion
+  if (init == error_mark_node)
+    {
+      if (stmt.get_pattern ().get_pattern_type ()
+	  == HIR::Pattern::PatternType::TUPLE)
+	{
+	  if (!seen_error ())
+	    rust_error_at (stmt.get_init_expr ().get_locus (),
+			   "expected value, found invalid expression");
+	  return;
+	}
+    }
   if (init == nullptr)
     return;
 
--- a/gcc/testsuite/rust/compile/issue-4161.rs
+++ b/gcc/testsuite/rust/compile/issue-4161.rs
@@ -1,0 +1,6 @@
+
+type X = ();
+
+fn main() {
+    let () = ::X = (); // { dg-error "expected value, found invalid expression" }
+}


gcc/testsuite/rust/compile/issue-4161.rs
```diff
@@ -0,0 +1,5 @@
+type X = ();
+
+fn main() {
+    let ()=::X = ();
```
P-E-P: We cannot use that test like this because it outputs an error with rustc https://godbolt.org/z/K9zc1fKo4

```
expected value, found type alias `X`
```

We either need to handle that error properly and match it with `dg-error` or we need to find a different testcase that highlight the fixed ICE without throwing an error.

gcc/rust/backend/rust-compile-pattern.cc
```diff
@@ -1320,6 +1320,10 @@ CompilePatternLet::visit (HIR::WildcardPattern &pattern)
 void
 CompilePatternLet::visit (HIR::TuplePattern &pattern)
 {
+  if (init_expr == error_mark_node)
```
P-E-P: I'm unsure this is the correct place to handle that. We already emitted that error_mark_node, it would make sense to handle the problem earlier (emit an error message and bail out) rather than sidestepping it later down the line.

Harishankar14:
> I'm not 100% confident because I'm not too familiar with that part of the codebase but it feels like you're trying here to fix the symptoms rather than the source of the problem.

yes @P-E-P  I have modified as per the changes requested, works fine now.

gcc/rust/backend/rust-compile-stmt.cc
```diff
@@ -74,6 +74,17 @@ CompileStmt::visit (HIR::LetStmt &stmt)
   tree init = CompileExpr::Compile (stmt.get_init_expr (), ctx);
   // FIXME use error_mark_node, check that CompileExpr returns error_mark_node
   // on failure and make this an assertion
+  if (init == error_mark_node)
+    {
+      if (stmt.get_pattern ().get_pattern_type ()
+	  == HIR::Pattern::PatternType::TUPLE)
+	{
+	  if (!seen_error ())
```
philberty, line 82: this isnt the right place to check for this i've loioked into this before there is a more general case we should be trying to catch this its during path code gen not stmt.

Also in your git commit subject you mention type-check.. this is not the type check pass.

Harishankar14:
@philberty Interesting !! Thanks for the pointer on the previous iteration! I’ve reverted the changes in `rust-compile-stmt.cc` and moved the logic to the `PathInExpression` handling as you suggested, the flow was bit confusing to understand.

It turned into a bit of a chain reaction to get this fully stable, but here is what I ended up doing:

1. **The Core Fix:** I modified `ResolvePathRef::resolve_with_node_id` (in `rust-compile-resolve-path.cc`) to handle cases where the type lookup fails (returning `false`). Instead of asserting/crashing there, I added a check to print a backend error and return `error_mark_node` so the compiler can recover.
 
2. **Fixing Downstream ICEs:** Once I let the error propagate, it triggered crashes further down the pipeline that I also had to patch:
* **Pattern Compilation:** `rust-compile-pattern.cc` was crashing because `TYPE_MAIN_VARIANT` tried to access the `error_mark_node`. I added a guard clause in `CompilePatternLet::visit` to return early if the init expression is invalid.
* **Linter/Dead Code:** The `MarkLive` pass was asserting when it couldn't look up the HIR ID for the broken Type Alias. I updated `rust-lint-marklive.cc` to simply skip marking the item as live if the lookup fails, rather than hitting `rust_unreachable()`.

3.Now the issue, which am facing is that it doesn't really throw up an error tbh, it let's it reside silently, which is indeed a huge issue.

I will mark this as draft for now, and will let you know.

Harishankar14:
@P-E-P  , @philberty  i am closing this PR, will make a fresh one with the updated changes.

PR #4395, opened 2026-01-16
derive: Implement reconstruct_path helper for cmp

Pasta-coder:
This patch introduces a local helper  to clone
PathInExpression objects. This allows removing duplicate calls to , fulfilling the TODOs in EnumMatchBuilder.
This ensures the self and other paths are identical without reconstructing them from strings twice.

gcc/rust/ChangeLog:

	* expand/rust-derive-cmp-common.cc (reconstruct_path): New helper.
	(EnumMatchBuilder::tuple): Use reconstruct_path.
	(EnumMatchBuilder::strukt): Use reconstruct_path.

### DIFF
--- a/gcc/rust/ast/rust-path.h
+++ b/gcc/rust/ast/rust-path.h
@@ -84,6 +84,15 @@ public:
     // and also identifier is empty, but cheaper computation
   }
 
+  GenericArgsBinding reconstruct () const
+  {
+    std::unique_ptr<Type> new_type = nullptr;
+    if (type)
+      new_type = type->reconstruct ();
+
+    return GenericArgsBinding (identifier, std::move (new_type), locus);
+  }
+
   // Creates an error state generic args binding.
   static GenericArgsBinding create_error ()
   {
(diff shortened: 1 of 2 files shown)


CohenArthur (commented):
So I'm not necessarily against these changes, but the goal with this comment here was to do something like https://github.com/Rust-GCC/gccrs/pull/3799 but for other base classes. I think we can merge this but I'd rather see the entire thing being implemented

Pasta-coder:
thank you for the review .

will modify accordingly .

Pasta-coder:
@CohenArthur Thanks for the review ! 

I've refactored the implementation to follow the established `reconstruct()` pattern across the entire Path hierarchy, similar to what was done for types in #3799 .

thankyou .

Pasta-coder:
The changes implement `reconstruct()` methods for all Path-related classes in `rust-path.h`, eliminating the need for manual field-by-field reconstruction in derive macros. This ensures consistent NodeID generation and removes the redundant path construction .

also tagging @powerboat9 to get reviews .

thankyou .

Pasta-coder:
@P-E-P please review .

thankyou . 

Note: Since `Path` inherits from `Pattern`, I will be updating these implementations in my PR (#4431) to override the new `Pattern::reconstruct_impl` virtual method. This PR establishes the logic, and #4431 will adapt it to the full architecture.

PR #4396, opened 2026-01-16
derive: Factor out Eq trait path generation

Pasta-coder:
This patch introduces a local helper  in DeriveEq. This removes the duplication of the hardcoded path {core, cmp, Eq} used in both  and .

gcc/rust/ChangeLog:

	* expand/rust-derive-eq.cc (get_eq_trait_path): New helper.
	(DeriveEq::assert_param_is_eq): Use helper.
	(DeriveEq::eq_impls): Use helper.

### DIFF
--- a/gcc/rust/expand/rust-derive-eq.cc
+++ b/gcc/rust/expand/rust-derive-eq.cc
@@ -27,6 +27,12 @@
 namespace Rust {
 namespace AST {
 
+static TypePath
+get_eq_trait_path (Builder &builder)
+{
+  return builder.type_path ({"core", "cmp", "Eq"}, true);
+}
+
 DeriveEq::DeriveEq (location_t loc) : DeriveVisitor (loc) {}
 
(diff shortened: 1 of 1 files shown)


gcc/rust/expand/rust-derive-eq.cc
```diff
@@ -112,10 +118,8 @@ DeriveEq::eq_impls (
   std::unique_ptr<AssociatedItem> &&fn, std::string name,
   const std::vector<std::unique_ptr<GenericParam>> &type_generics)
 {
-  // We create two copies of the type-path to avoid duplicate NodeIds
```
CohenArthur, line 115: Please keep the comment here

Pasta-coder: done .

thankyou

Pasta-coder:
@CohenArthur , thank you for the review . 

will also keep this in mind for future prs .

PR #4397, opened 2026-01-16
backend: Factor out asm operand chaining

Pasta-coder:
This patch introduces a static helper  in .
This removes code duplication when building tree lists for assembly inputs and outputs. It also removes commented-out debug code to clean up the file.

gcc/rust/ChangeLog:

	* backend/rust-compile-asm.cc (chain_asm_operand): New helper.
	(CompileAsm::asm_construct_outputs): Use helper and remove dead code.
	(CompileAsm::asm_construct_inputs): Use helper and remove dead code.

### DIFF
--- a/gcc/rust/backend/rust-compile-asm.cc
+++ b/gcc/rust/backend/rust-compile-asm.cc
@@ -5,6 +5,14 @@
 namespace Rust {
 namespace Compile {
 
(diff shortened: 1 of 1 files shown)


CohenArthur (approved):
Thanks!

PR #4404, opened 2026-01-21
nr: Add modules to types NS

### LINKED ISSUE #4403: Modules need to be inserted in the type namespace
```rust
mod test {
    mod Toto {

    }

    struct Toto;
}
```

https://godbolt.org/z/r55x4jrGh

Thanks to @P-E-P for the testcase and godbolt link

This is important as the prelude import in core fails since it only contains a mod declaration for `v1`:

```rust
// core/src/lib.rs
use prelude::v1::*;

// core/src/prelude/mod.rs
pub mod v1;

// core/src/prelude/v1.rs
pub use crate::option::Option::{None, Some};
```

### LINKED ISSUE #4563: Invalid multiple definitions errors for module and function with the same name
```rust
#![feature(no_core)]
#![no_core]

mod foo;

use foo::foo;

fn main() {
    foo();
}
```

with a foo module file only containing the following function:

```rust
pub fn foo() {}
```

This causes the following issue:

```rust
test.rs:6:5: error: ‘foo’ defined multiple times [E0428]
    4 | mod foo;
      | ~~~
    5 |
    6 | use foo::foo;
      |     ^~~
```

rustc doesn't emit any issues. This is a problem since modules should be inserted in the types NS, while functions should only be inserted in the Values NS.

Note that this is likely caused by the current implementation of #4404 and should be fixed before it gets merged

CohenArthur:
gcc/rust/ChangeLog:

	* util/rust-hir-map.cc (Mappings::insert_module_id): New function. (Mappings::is_module): Likewise.
	* util/rust-hir-map.h: Store a set of AST modules, declare functions for adding and retrieving them.
	* resolve/rust-toplevel-name-resolver-2.0.cc (TopLevel::visit): Insert modules in the type namespace and store them in mappings.
	* resolve/rust-late-name-resolver-2.0.cc (resolve_type_path_like): Error out when an expected type resolves to a module.

gcc/testsuite/ChangeLog:

	* rust/compile/mod_in_types_ns.rs: New test.
	* rust/compile/mod_in_types_ns2.rs: New test.

Fixes #4403
Fixes #4563

### DIFF
--- a/gcc/rust/resolve/rust-name-resolution-context.cc
+++ b/gcc/rust/resolve/rust-name-resolution-context.cc
@@ -256,10 +256,10 @@ NameResolutionContext::insert_globbed (Identifier name, NodeId id, Namespace ns)
 void
 NameResolutionContext::map_usage (Usage usage, Definition definition)
 {
-  auto inserted = resolved_nodes.emplace (usage, definition).second;
+  resolved_nodes[usage] = definition;
 
-  // is that valid?
-  rust_assert (inserted);
+  // TODO: Are reinsertions okay?
+  // rust_assert (inserted);
 }
 
 tl::optional<NodeId>
--- a/gcc/rust/resolve/rust-name-resolution-context.hxx
+++ b/gcc/rust/resolve/rust-name-resolution-context.hxx
@@ -160,28 +160,6 @@ NameResolutionContext::resolve_path (
 	      // TODO: does NonShadowable matter?
 	      return Rib::Definition::NonShadowable (id);
 	    }
-	  else
-	    {
-	      // HACK: check for a module after we check the language prelude
-	      for (auto &kv :
-		   stack.find_closest_module (starting_point.get ()).children)
-		{
-		  auto &link = kv.first;
-
-		  if (link.path.map_or (
-			[&seg] (Identifier path) {
-			  auto &path_str = path.as_string ();
-			  return path_str == seg.name;
-			},
-			false))
-		    {
-		      // FIXME: Is the NS to insert_segment_resolution valid?
-		      insert_segment_resolution (Usage (seg.node_id),
-						 Definition (kv.second.id), N);
-		      return Rib::Definition::NonShadowable (kv.second.id);
-		    }
-		}
-	    }
 	}
 
       // FIXME: Is the NS to insert_segment_resolution valid?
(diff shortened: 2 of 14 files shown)


P-E-P (approved):

gcc/rust/resolve/rust-name-resolution-context.cc
```diff
@@ -252,10 +252,10 @@ NameResolutionContext::insert_globbed (Identifier name, NodeId id, Namespace ns)
 void
 NameResolutionContext::map_usage (Usage usage, Definition definition)
 {
-  auto inserted = resolved_nodes.emplace (usage, definition).second;
+  resolved_nodes[usage] = definition;
 
-  // is that valid?
-  rust_assert (inserted);
+  // TODO: Are reinsertions okay?
+  // rust_assert (inserted);
```
powerboat9: Problem is, a lot of reinsertions aren't reinsertions so much as "we have multiple nodes sharing the same node id" (though we should really have another pass checking for that). Failing that, resolutions theoretically shouldn't be changing over time, even though the specific use case you're going for makes sense. Could you add a `map_usage_unsafe` method instead?

powerboat9:
I think I'm going to try to add backtracking to the resolver, as I don't think the type prelude will work quite right otherwise.

CohenArthur:
> I think I'm going to try to add backtracking to the resolver, as I don't think the type prelude will work quite right otherwise.

what do you mean by backtracking?

powerboat9:
> > I think I'm going to try to add backtracking to the resolver, as I don't think the type prelude will work quite right otherwise.
> 
> what do you mean by backtracking?

Backtracking in the sense that resolution doesn't commit to certain segment resolutions until it succeeds, and can retry with a different namespace/starting location (type prelude vs not type prelude) with fewer issues.

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
```diff
@@ -106,12 +106,16 @@ TopLevel::go (AST::Crate &crate)
...
   if (Analysis::Mappings::get ().lookup_glob_container (module.get_node_id ())
       == tl::nullopt)
     Analysis::Mappings::get ().insert_glob_container (module.get_node_id (),
 						      &module);
+
+  insert_or_error_out (module.get_name (), module, Namespace::Types);
+
+  Analysis::Mappings::get ().insert_module_id (module.get_node_id ());
+
+  DefaultResolver::visit (module);
```
powerboat9, line 119: Why reorder this? Wouldn't multiple `TopLevel` passes cause any bug to return?

powerboat9: On second thought, this is neater

gcc/rust/resolve/rust-name-resolution-context.cc
CohenArthur: This is also an issue for #4551 - I'd rather not think about it now if I'm honest since I've been working on this for months at this point. I would rather we merge this, merge #4551 and then add an issue for improving `map_usage` and making it more correct

CohenArthur:
Probably this should get merged *after* #4551 because #4551 is so big

gcc/rust/resolve/rust-name-resolution-context.hxx
```diff
@@ -157,27 +157,28 @@ NameResolutionContext::resolve_path (
...
-			},
-			false))
-		    {
-		      insert_segment_resolution (Usage (seg.node_id),
-						 Definition (kv.second.id));
-		      return Rib::Definition::NonShadowable (kv.second.id);
-		    }
-		}
-	    }
+	  //  else
```
P-E-P: Should we clean this up instead of using comments ?

P-E-P (approved):

Issue #4402, opened 2026-01-21
Cannot resolve module path properly in simple setting


CohenArthur:
With the following `bar.rs` file:

```rust
use foo::Bar;

pub mod foo;

fn main() {
    let a = Bar;
}
```

and in the same directory a `foo.rs` file:

```rust
pub struct Bar;
```

we have the following error:

```rust
bar.rs:3:5: error: no candidate found for module foo
    3 | pub mod foo;
      |     ^~~
bar.rs:3:5: error: no candidate found for module foo
bar.rs:1:10: error: unresolved import ‘foo::Bar’ [E0433]
    1 | use foo::Bar;
      |          ^~~
bar.rs:6:13: error: cannot find value ‘Bar’ in this scope [E0425]
    6 |     let a = Bar;
      |             ^~~
```

`rustc bar.rs` does not produce any error.

Harishankar14:
@CohenArthur Let me know if it is good, or any modifications to be done, I feel its fine.

PR #4406, opened 2026-01-22
ast: Fix module path resolution for sibling modules in root

Harishankar14:
Currently, the compiler fails to resolve module paths when a file in the root directory (e.g. 'bar.rs') attempts to load a sibling module (e.g. 'foo.rs') if a 'bar/' directory does not exist. The compiler incorrectly assumes that if 'bar.rs' is not 'mod.rs', it must look exclusively in a subdirectory.

This patch adds a fallback mechanism in 'Module::process_file_path'. If the subdirectory search fails, it strips the implicit subdirectory and attempts to resolve the module in the parent directory, consistent with Rust 2018 path rules.

Fixes Rust-GCC/gccrs#4402

gcc/rust/ChangeLog:

	* ast/rust-ast.cc (Module::process_file_path): Add fallback search for sibling modules when subdirectory search fails.

gcc/testsuite/ChangeLog:

	* rust/compile/issue-4402.rs: New test.
	* rust/compile/issue_4402_foo.rs: New test.

in the git commit message.

- \[ ] Run `clang-format`

order to get feedback.
---

### DIFF
--- a/gcc/testsuite/rust/compile/issue_4402_foo.rs
+++ b/gcc/testsuite/rust/compile/issue_4402_foo.rs
@@ -1,0 +1,1 @@
+pub struct Bar;
--- a/gcc/rust/ast/rust-ast.cc
+++ b/gcc/rust/ast/rust-ast.cc
@@ -3418,9 +3418,13 @@ Module::process_file_path ()
   auto path_string = filename_from_path_attribute (get_outer_attrs ());
 
   std::string including_subdir;
+  bool subdir_was_added = false;
   if (path_string.empty () && module_scope.empty ()
       && get_file_subdir (including_fname, including_subdir))
-    current_directory_name += including_subdir + file_separator;
+    {
+      current_directory_name += including_subdir + file_separator;
+      subdir_was_added = true;
+    }
 
   // Handle inline module declarations adding path components.
   for (auto const &name : module_scope)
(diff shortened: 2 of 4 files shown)


Harishankar14:
@CohenArthur

Harishankar14:
@P-E-P ANY ISSUES , the bot removed the PR from merge queue, i guess have to make some modification accordingly with respect to your new PR that was added ?

P-E-P:
> @P-E-P ANY ISSUES , the bot removed the PR from merge queue, i guess have to make some modification accordingly with respect to your new PR that was added ?

Yes I forgot about it again; you need to add `#[no_core]` to the crate. If you could also rebase your branch on master and delete that merge commit that would be great.

Harishankar14:
@P-E-P Should be GTG?

gcc/testsuite/rust/compile/issue_4402_foo.rs
```diff
@@ -0,0 +1,3 @@
+#![feature(no_core)]
```
P-E-P: You don't need `#![feature(no_core)]` and `#![no_core]` here, it is only required once per crate, since you already declared it in `issue_4402.rs`, it shouldn't be used over here.

Harishankar14: Ah, yeah true, missed removing,lol !!

P-E-P:
You could use something that looks like the following:

In `compile.exp`
```
set ignore_list [ "issue_4402_foo.rs" ]
dg-runtest [lsearch -inline -all -not -exact [lsort [glob -nocomplain $srcdir/$subdir/*.rs]] [ lmap file $ignore_list "$src/$subdir/$file"]] "" ""
```

I haven't tested it so you'll have to experiment a bit but this could do the trick

Harishankar14:
@P-E-P  I think it worked !! No failures when i tested it out locally

PR #4407, opened 2026-01-23
Implicitely inject `extern crate core`

P-E-P:
Implicitely inject `core` to all crates without the `#![no_core]` attribute.


P-E-P:
I'm still unsure about how I'll deal with the testsuite. Either I'll add the `#![no_core]` attribute to every file or I'll have to inject the `#![no_core]` attribute at the directory level using a compiler argument.

PR #4416, opened 2026-01-30
ci: fix send email workflow in case of large commit log

dkm:
We've had at least one case of mails not being sent because the commit log was so big that the $GH_EVENT var couldn't be expanded by the shell. Using envsubst avoids having the shell do the expansion.


P-E-P (approved):
NGL, the quotation marks got me at first. LGTM!

Issue #4414, opened 2026-01-28
segfault Rust::Parser<Rust::Lexer>::parse_arithmetic_or_logical_expr(


matthiaskrgr:
auto-reduced (treereduce-rust):
````rust
//@compile-flags: -frust-incomplete-and-experimental-compiler-do-not-use


impl<T: Speak> Speak for Option<T> {
    fn say(&self, s:&str) -> String {
        match something!hello - none {
            
            
        }
    }
}




````

original:
````rust
//@ run-pass

trait Speak : Sized {
    fn say(&self, s:&str) -> String;
    fn hi(&self) -> String { hello(self) }
}

fn hello<S:Speak>(s:&S) -> String{
    s.say("hello")
}

impl Speak for assert_eq!(Some(3).hi(), "something!hello: 3".to_string()) {
    fn say(&self, s:&str) -> String {
        format!("{}: {}", s, *self)
    }
}

impl<T: Speak> Speak for Option<T> {
    fn say(&self, s:&str) -> String {
        match something!hello - none {
            None => format!("{} - none", s),
            Some(ref x) => { format!("something!{}", x.say(s)) }
        }
    }
}


fn hello<S:Speak>(s:&S) -> String{
    s.say("hello")
}

````

Version information:

145e66efdf603476ef6b0d66523936ff55187162

Command:
`/home/matthias/vcs/github/gccrs/gccrs-build/gcc/crab1 -frust-incomplete-and-experimental-compiler-do-not-use`

<details><summary><strong>Program output</strong></summary>
<p>

```
crab1: internal compiler error: Segmentation fault
0x30d80b9 internal_error(char const*, ...)
	../../gcc/diagnostic-global-context.cc:787
0x1a0b21f crash_signal
	../../gcc/toplev.cc:325
0xdc8727 Rust::Parser<Rust::Lexer>::parse_arithmetic_or_logical_expr(std::shared_ptr<Rust::Token const>, std::unique_ptr<Rust::AST::Expr, std::default_delete<Rust::AST::Expr> >, std::vector<Rust::AST::Attribute, std::allocator<Rust::AST::Attribute> >, ArithmeticOrLogicalOperator, Rust::ParseRestrictions)
	../../gcc/rust/parse/rust-parse-impl-expr.hxx:2930
[...]

PR #4417, opened 2026-01-31
parser: Fix segfault in parse_arithmetic_or_logical_expr 

Harishankar14:
The compiler previously crashed with a segmentation fault when parsing an arithmetic expression where the left-hand side was invalid (e.g., after a failed macro expansion). This occurred because the parser attempted to access the location of the null 'left' pointer. This patch adds a check to ensure the left operand is valid before proceeding.

Fixes Rust-GCC/gccrs#4414

gcc/rust/ChangeLog:

	* parse/rust-parse-impl-expr.hxx (parse_arithmetic_or_logical_expr): Check for null left operand to prevent segmentation fault.

gcc/testsuite/ChangeLog:

	* rust/compile/issue-4414.rs: New test.

in the git commit message.

- \[ ] Run `clang-format`

order to get feedback.
---

### DIFF
--- a/gcc/rust/parse/rust-parse-impl-expr.hxx
+++ b/gcc/rust/parse/rust-parse-impl-expr.hxx
@@ -1794,7 +1794,9 @@ Parser<ManagedTokenSource>::parse_expr (int right_binding_power,
     = null_denotation ({}, null_denotation_restrictions);
   if (!expr)
     return tl::unexpected<Parse::Error::Expr> (Parse::Error::Expr::CHILD_ERROR);
-
+  if (expr.value () == nullptr)
+    return tl::unexpected<Parse::Error::Expr> (Parse::Error::Expr::CHILD_ERROR);
+  
   return left_denotations (std::move (expr), right_binding_power,
 			   std::move (outer_attrs), restrictions);
 }
--- a/gcc/testsuite/rust/compile/issue-4414.rs
+++ b/gcc/testsuite/rust/compile/issue-4414.rs
@@ -1,0 +1,30 @@
+#![feature(no_core)]
+#![no_core]
+
+trait Speak : Sized {
+    fn say(&self, s:&str) -> String;
+    fn hi(&self) -> String { hello(self) }
+}
+
+fn hello<S:Speak>(s:&S) -> String{
+    s.say("hello")
+}
+
+impl Speak for assert_eq!(Some(3).hi(), "something!hello: 3".to_string()) { 
+    fn say(&self, s:&str) -> String {
+        format!("{}: {}", s, *self)
+    }
+}
+
+impl<T: Speak> Speak for Option<T> {
+    fn say(&self, s:&str) -> String {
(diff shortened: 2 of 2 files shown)


Harishankar14:
@P-E-P

gcc/rust/parse/rust-parse-impl-expr.hxx
powerboat9: You should probably remove the curly brackets on the new if-statement, to match the older if-statement

Harishankar14:
@powerboat9  Thanks !! Sorry for displaying my poor inadapatability, lol !! 
@P-E-P  Should be gtg !!

gcc/rust/parse/rust-parse-impl-expr.hxx
```diff
@@ -2920,6 +2920,8 @@ Parser<ManagedTokenSource>::parse_arithmetic_or_logical_expr (
   AST::ArithmeticOrLogicalExpr::ExprType expr_type,
   ParseRestrictions restrictions)
 {
+  if (!left)
```
P-E-P: We should probably check this before passing it down the arguments of a function ? Unless explicitly required I wouldn't put that check over here. We got that value from somewhere else higher in the call stack and it could slip in other function calls.

Harishankar14: @P-E-P  That makes sense. I agree that checking inside parse_arithmetic_or_logical_expr was too late in the chain and just patching the symptom.I've moved the check up to left_denotation instead. Since that function acts as the dispatcher for all binary operators within the file catching a null left expression there prevents the invalid state from propagating down to any handler, not just this one. Thanks for the suggestion

Harishankar14:
@P-E-P  Thanks for that

Harishankar14:
> I believe this is still not high enough in the call chain.

Will Look into it !!

Harishankar14:
@P-E-P  Updated !!

gcc/rust/parse/rust-parse-impl-expr.hxx
```diff
@@ -1794,7 +1794,9 @@ Parser<ManagedTokenSource>::parse_expr (int right_binding_power,
     = null_denotation ({}, null_denotation_restrictions);
   if (!expr)
     return tl::unexpected<Parse::Error::Expr> (Parse::Error::Expr::CHILD_ERROR);
-
+  if (expr.value () == nullptr)
```
P-E-P, line 1797: Ok, so this is unexpected. I've been slowly converting errors from nullptr to tl::expected and I missed one somewhere in the null_denotation call tree. We shouldn't have to check for a nullptr if it is wrapped in a tl::unexpected.

Harishankar14: @P-E-P Could we merge this PR as-is to fix the crash immediately? I'm happy to open a separate, dedicated PR to hunt down the origin of that null pointer in null_denotation.

Since I've already cleaned up my local branches, a fresh PR would be much easier for me to manage and would keep the history cleaner for the bug hunt. What do you think?

PR #4418, opened 2026-02-01
Make `AST::Crate` final

powerboat9:
This should prevent the warnings which seem to be breaking bootstrapping.


P-E-P:
We could have made the dtor virtual but I don't see any usecase to a derived `AST::Crate` anyway. @powerboat9 Did you bootstrap locally ?

powerboat9:
https://github.com/Rust-GCC/gccrs/actions/runs/21614357704/job/62289809249

PR #4420, opened 2026-02-01
parser: Fix ICE in AnonConst when expression parsing fails

### LINKED ISSUE #4412: ICE `in AnonConst, at rust/ast/rust-expr.h`
````rust
struct Bug([u8; panic!"\t"}]);



````

original:
````rust
struct Bug([u8; panic!"\t"}]);

fn main() {}

````

Version information:

145e66efdf603476ef6b0d66523936ff55187162

Possibly related line of code:
https://github.com/Rust-GCC/gccrs/blob/145e66efdf603476ef6b0d66523936ff55187162/gcc/rust/ast/rust-expr.h#L2887-L2899

Command:
`/home/matthias/vcs/github/gccrs/gccrs-build/gcc/crab1 -frust-incomplete-and-experimental-compiler-do-not-use`

<p>

```
crab1: internal compiler error: in AnonConst, at rust/ast/rust-expr.h:2893
0x30d80b9 internal_error(char const*, ...)
	../../gcc/diagnostic-global-context.cc:787
0xc3ae5d fancy_abort(char const*, int, char const*)
	../../gcc/diagnostics/context.cc:1805
0xdc3962 Rust::AST::AnonConst::AnonConst(std::unique_ptr<Rust::AST::Expr, std::default_delete<Rust::AST::Expr> >&&, unsigned long)
	../../gcc/rust/ast/rust-expr.h:2893
0xdc3962 Rust::Parser<Rust::Lexer>::parse_anon_const()
	../../gcc/rust/parse/rust-parse-impl-expr.hxx:113
0xdda166 Rust::Parser<Rust::Lexer>::parse_slice_or_array_type()
	../../gcc/rust/parse/rust-parse-impl.hxx:6455
0xdd8ef9 Rust::Parser<Rust::Lexer>::parse_type(bool)
	../../gcc/rust/parse/rust-parse-impl.hxx:5624
0xdde15b Rust::Parser<Rust::Lexer>::parse_tuple_field()
	../../gcc/rust/parse/rust-parse-impl.hxx:3285
0xdde561 Rust::Parser<Rust::Lexer>::parse_tuple_fields()
	../../gcc/rust/parse/rust-parse-impl.hxx:3219
0xdf7fa6 Rust::Parser<Rust::Lexer>::parse_struct(Rust::AST::Visibility, std::vector<Rust::AST::Attribute, std::allocator<Rust::AST::Attribute> >)
	../../gcc/rust/parse/rust-parse-impl.hxx:3007
0xe03672 Rust::Parser<Rust::Lexer>::parse_vis_item(std::vector<Rust::AST::Attribute, std::allocator<Rust::AST::Attribute> >)
	../../gcc/rust/parse/rust-parse-impl.hxx:538
0xe0433a Rust::Parser<Rust::Lexer>::parse_item(bool)
	../../gcc/rust/parse/rust-parse-impl.hxx:402
0xe0580c Rust::Parser<Rust::Lexer>::parse_items()
	../../gcc/rust/parse/rust-parse-impl.hxx:278
0xe05a36 Rust::Parser<Rust::Lexer>::parse_crate()
	../../gcc/rust/parse/rust-parse-impl.hxx:305
0xec8ff8 Rust::Session::compile_crate(char const*)
	../../gcc/rust/rust-session-manager.cc:601

```

</p>
</details>

### PR DESCRIPTION
**Pull request by Pasta-coder**
The parse_anon_const function calls parse_expr to parse the expression inside an anonymous constant (e.g., array length [T; N]). Previously, it only checked if the Result object indicated success, but did not check if the contained expression pointer was null.

In cases like malformed macro invocations, parse_expr can return a success Result containing a nullptr. Passing this nullptr to the AnonConst constructor triggered an assertion failure (ICE).

This patch adds a check to ensure the parsed expression is not null before constructing the AnonConst object.

Fixes #4412

gcc/rust/ChangeLog:

	* parse/rust-parse-impl-expr.hxx (parse_anon_const): Check for null expression.

gcc/testsuite/ChangeLog:

	* rust/compile/issue-4412.rs: New test.

### DIFF
--- a/gcc/rust/parse/rust-parse-impl-expr.hxx
+++ b/gcc/rust/parse/rust-parse-impl-expr.hxx
@@ -1957,9 +1957,15 @@ Parser<ManagedTokenSource>::null_denotation_path (
   switch (t->get_id ())
     {
     case EXCLAM:
-      // macro
-      return parse_macro_invocation_partial (std::move (path),
-					     std::move (outer_attrs));
+      {
+	// macro
+	auto macro = parse_macro_invocation_partial (std::move (path),
+						     std::move (outer_attrs));
+	if (macro == nullptr)
+	  return tl::unexpected<Parse::Error::Expr> (
+	    Parse::Error::Expr::CHILD_ERROR);
+	return std::unique_ptr<AST::Expr> (std::move (macro));
+      }
     case LEFT_CURLY:
       {
 	bool not_a_block = lexer.peek_token (1)->get_id () == IDENTIFIER
--- a/gcc/testsuite/rust/compile/issue-4412.rs
+++ b/gcc/testsuite/rust/compile/issue-4412.rs
@@ -1,0 +1,9 @@
+// { dg-options "-frust-incomplete-and-experimental-compiler-do-not-use" }
+#![feature(no_core)]
+#![no_core]
+struct Bug([u8; panic!"\t"]);
+// { dg-error "unexpected token" "" { target *-*-* } .-1 }
+// { dg-error "failed to parse" "" { target *-*-* } .-2 }
+// { dg-error "could not parse" "" { target *-*-* } .-3 }
+// { dg-error "expecting" "" { target *-*-* } .-4 }
+fn main() {}

### CODE REVIEW

**Comment by Pasta-coder** [comment]
@P-E-P please review .

thankyou .

**Comment by powerboat9** [comment]
The root problem here seems to be that `parse_expr` is returning a null pointer. It should really be returning either a successfully parsed expression or an error.

**Comment by Pasta-coder** [comment]
I've updated the PR to address the feedback. I moved the fix to `parse_expr` so it now correctly returns an error instead of a nullptr when parsing fails, preventing the issue from propagating to `AnonConst`.

Ready for re-review .

thankyou .

File: gcc/rust/parse/rust-parse-impl-expr.hxx [hunk 1]
```diff
@@ -1795,6 +1795,12 @@ Parser<ManagedTokenSource>::parse_expr (int right_binding_power,
   if (!expr)
     return tl::unexpected<Parse::Error::Expr> (Parse::Error::Expr::CHILD_ERROR);
 
+  if (expr.value () == nullptr)
```
P-E-P: We've got two layers of error here, a `nullptr` embeded within an `unexpected`. We shouldn't have to check for null here, instead a null shouldn't have been emitted.

gcc/testsuite/rust/compile/issue-4412.rs
```diff
@@ -0,0 +1,10 @@
+// { dg-options "-frust-incomplete-and-experimental-compiler-do-not-use" }
+#![feature(no_core)]
+#![no_core]
+
+struct Bug([u8; panic!"\t"}]);
+// { dg-error "failed to parse expression|unexpected token" "" { target *-*-* } .-1 }
+
+fn main() {}
+
+// { dg-excess-errors "Parser cascade due to malformed macro and mismatched braces" }
```
P-E-P: Please do not use excess errors.

P-E-P (changes requested):
We need to add a null check and report an error when that `tl::expected` gets created. Not all part of the parser have been reworked with `tl::expected` in mind, it may be when the parsing fail or higher up in the calls when the transition goes from raw unique null to expected.

codblack789:
Dud you use an LLM or something 🤔. Kind of looks like it

Pasta-coder:
the ICE is resolved, and the previous excess error emission has also been addressed .

ready for re-review .

PR #4423, opened 2026-02-06
refactor: cleanup macro invoc

v1bh475u:
Addresses Rust-GCC/gccrs#2919

- \[x] Run `clang-format`

### DIFF
--- a/gcc/rust/ast/rust-macro.h
+++ b/gcc/rust/ast/rust-macro.h
@@ -709,8 +709,10 @@ private:
     MacroInvocData invoc_data, std::vector<Attribute> outer_attrs,
     location_t locus, bool is_semi_coloned,
     std::vector<std::unique_ptr<MacroInvocation>> &&pending_eager_invocs)
-    : TraitItem (locus), outer_attrs (std::move (outer_attrs)), locus (locus),
-      node_id (Analysis::Mappings::get ().get_next_node_id ()),
+    : TraitItem (locus),
+      ExternalItem (Analysis::Mappings::get ().get_next_node_id ()),
+      outer_attrs (std::move (outer_attrs)), locus (locus),
+      node_id (ExternalItem::get_node_id ()),
       invoc_data (std::move (invoc_data)), is_semi_coloned (is_semi_coloned),
       kind (kind), builtin_kind (builtin_kind),
       pending_eager_invocs (std::move (pending_eager_invocs))


v1bh475u:
@P-E-P  @CohenArthur , I have opened this draft PR to get some suggestions. I have made sure that we only use `MacroInvocation`'s member `node_id` everywhere instead of using `ExprWithoutBlock` at some places and member `node_id`. However, due to many inherited classes, there are a lot of duplicate `node_id` fields, most of which are not even initialized. Since these duplicate fields (`node_id` and `locus`) are under `protected` scope and accessed via getter methods which are virtual, I don't think we need to worry about having some issue. But I do feel like the inheritance is kinda messy as it is creating memory for a lot of fields which are never used. Any suggestions for some better abstraction?

P-E-P:
> @P-E-P @CohenArthur , I have opened this draft PR to get some suggestions. I have made sure that we only use `MacroInvocation`'s member `node_id` everywhere instead of using `ExprWithoutBlock` at some places and member `node_id`. However, due to many inherited classes, there are a lot of duplicate `node_id` fields, most of which are not even initialized. Since these duplicate fields (`node_id` and `locus`) are under `protected` scope and accessed via getter methods which are virtual, I don't think we need to worry about having some issue. But I do feel like the inheritance is kinda messy as it is creating memory for a lot of fields which are never used. Any suggestions for some better abstraction?

So I've taken a look at your PR and it looks great. Year inheritance is messy, we would need to de duplicate some fields by separating some bits from common ancestors, or even better, use composition instead. But this requires a lot of work on moving parts.

v1bh475u:
> > @P-E-P @CohenArthur , I have opened this draft PR to get some suggestions. I have made sure that we only use `MacroInvocation`'s member `node_id` everywhere instead of using `ExprWithoutBlock` at some places and member `node_id`. However, due to many inherited classes, there are a lot of duplicate `node_id` fields, most of which are not even initialized. Since these duplicate fields (`node_id` and `locus`) are under `protected` scope and accessed via getter methods which are virtual, I don't think we need to worry about having some issue. But I do feel like the inheritance is kinda messy as it is creating memory for a lot of fields which are never used. Any suggestions for some better abstraction?
> 
> So I've taken a look at your PR and it looks great. Year inheritance is messy, we would need to de duplicate some fields by separating some bits from common ancestors, or even better, use composition instead. But this requires a lot of work on moving parts.

I agree as this would require complete paradigm change from current existing codebase. Thus, this would require changes that would create ripples across the entire codebase. Are we aiming to do this right now or should we focus on features and add a todo?

P-E-P:
> I agree as this would require complete paradigm change from current existing codebase. Thus, this would require changes that would create ripples across the entire codebase. Are we aiming to do this right now or should we focus on features and add a todo?

We could merge your PR but completely changing the AST to another paradigm is not possible right now, I would not even create an issue for it, the AST is still very unstable and modified daily by multiple contributors. I think we should improve what can be improved bit by bit. For example I've started a few months ago a transition to remove null unique_ptr to tl::optional<unique_ptr> as it better convey the optional side of things. But working on the whole inheritance thing is completely out of scope for now, it works and we would not gain much from this transition, we better work on improving the compiler's output rather than our internal representation.

v1bh475u:
Currently, `MacroInvocation` inherits from base as well its subclasses. Should we remove those inheritances and add the required methods/data members directly in the class?

P-E-P:
> Currently, `MacroInvocation` inherits from base as well its subclasses. Should we remove those inheritances and add the required methods/data members directly in the class?

No, I've haven't looked too closely but there's a chance a lot of code relies on MacroInvocation being a children of those classes (mainly BlockExpr -> Expr IIRC), removing those inheritance links will definitely break some parts of the compiler.

CohenArthur:
Yes, MacroInvocation *needs* to be a child of all of those classes so that macro expansion is performed properly in all the places that macro invocations can exist. For example this is valid Rust:

```rust
fn foo(a: bar!()) -> bar!() {
    let bar: bar!() = baz!();
    
    baz!()
}
```

and having MacroInvocations inherit from all these base classes makes it easy to write our expanding visitors using virtual method dispatching

v1bh475u:
> Yes, MacroInvocation _needs_ to be a child of all of those classes so that macro expansion is performed properly in all the places that macro invocations can exist. For example this is valid Rust:
> 
> ```rust
> fn foo(a: bar!()) -> bar!() {
>     let bar: bar!() = baz!();
>     
>     baz!()
> }
> ```
> 
> and having MacroInvocations inherit from all these base classes makes it easy to write our expanding visitors using virtual method dispatching

I agree but using inheritance creates pretty rigid hierarchy. So, I think after we are done with getting the compiler working, we should maybe invest some time to change.

PR #4425, opened 2026-02-06
Improve feature collection and permit macro attribute input

P-E-P:


gcc/rust/checks/errors/feature/contrib/fetch
```diff
@@ -18,7 +18,7 @@
 # along with GCC; see the file COPYING3.  If not see
 # <http://www.gnu.org/licenses/>.
 
-RUST_VERSION="1.49.0"
+RUST_VERSION="1.50.0"
```
powerboat9: 1.49 libraries need features new to the 1.50 compiler?

powerboat9 (approved):

gcc/rust/checks/errors/feature/contrib/fetch
P-E-P: I'm experimenting with rust for linux, initially I thought this would be fine because we only had a superset of features but since the regeneration removes a feature I'll make a separate header with whatever features I need for linux.

P-E-P: I've made a different header specifically for rfl. It will contains whatever features will be required for rfl without changing `rust-feature-defs.h`.

gcc/rust/checks/errors/feature/contrib/fetch
```diff
@@ -27,4 +27,4 @@ RUST_VERSION="1.49.0"
 URL_PREFIX='https://raw.githubusercontent.com/rust-lang/rust/refs/tags'
 URL_TEMPLATE="$URL_PREFIX/$RUST_VERSION/compiler/rustc_feature/src"
 
-wget -O $1 "$URL_TEMPLATE"/{accepted,active,removed}.rs
+wget -O $1 "$URL_TEMPLATE/accepted.rs" "$URL_TEMPLATE/active.rs" "$URL_TEMPLATE/removed.rs"
```
powerboat9, line 30: Why change this? Readability?

P-E-P: No, the brace expansion did not work on my system when I first launched the script. I did not investigate much, I simply replaced it.

dkm: Because the shebang has `!/bin/sh` and brace expansion is a bash feature. I suspect the original code was using a system where /bin/sh is bash.

PR #4428, opened 2026-02-07
Improve path resolution

powerboat9:
This makes path resolution less dependent on templates for path handling.

### DIFF
--- a/gcc/rust/resolve/rust-forever-stack.h
+++ b/gcc/rust/resolve/rust-forever-stack.h
@@ -25,6 +25,8 @@
 #include "rust-path.h"
 #include "optional.h"
 #include "expected.h"
+#include "rust-name-resolution.h"
+#include "rust-unwrap-segment.h"
 
 namespace Rust {
 namespace Resolver2_0 {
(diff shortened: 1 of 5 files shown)


powerboat9:
Path resolution is still kind of a mess, so I'll probably put in a few more PRs later

gcc/rust/resolve/rust-forever-stack.h
```diff
@@ -550,6 +552,75 @@ enum class ResolutionMode
...
+  const std::vector<Segment> &get_segments () const { return segments; }
+
+  NodeId get_node_id () const { return node_id; }
+
+  std::string as_string () const
+  {
+    std::string ret;
+    if (lang_prefix)
+      // TODO
+      ret = "L::";
```
CohenArthur: I think let's do it now before getting it merged, or at least use something like `#[lang]::` to make it clear it's a lang item and not just a random `L` path segment

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -794,40 +794,46 @@ ForeverStack<N>::resolve_path (
...
+  auto iterator = segments.begin ();
+  if (can_descend)
+    {
+      if (auto res
+	  = find_starting_point (segments, starting_point,
+				 insert_segment_resolution, collect_errors))
+	iterator = *res;
+      else
+	return tl::nullopt;
+    }
```
CohenArthur, line 824: what's the reasoning behind this? why do we do things differently now?

gcc/rust/resolve/rust-name-resolution-context.h
```diff
@@ -619,37 +588,136 @@ class NameResolutionContext
...
+
+      return tl::nullopt;
+    }
+
+    void reset ()
+    {
+      NameResolutionContext *ctx = this->ctx;
+      this->~ResolutionBuilder ();
+      new (this) ResolutionBuilder (*ctx);
+    }
```
CohenArthur: wouldn't it be better to just recreate a `ResolutionBuilder` instead of resetting an existing one?

CohenArthur: and `reset` is not used anywhere at the moment so yeah I would rather delete this method

gcc/rust/resolve/rust-name-resolution-context.h
```diff
@@ -619,37 +588,136 @@ class NameResolutionContext
...
+  resolve_path (const std::vector<S> &path_segments, ResolutionMode mode,
 		tl::optional<std::vector<Error> &> collect_errors,
 		Namespace ns_first, Args... ns_args)
   {
-    std::initializer_list<Namespace> namespaces = {ns_first, ns_args...};
+    ResolutionBuilder builder (*this);
+    builder.set_path (path_segments, UNKNOWN_NODEID, mode);
+    builder.add_namespaces (ns_first, ns_args...);
+    if (collect_errors.has_value ())
+      builder.set_collect_errors (collect_errors.value ());
```
CohenArthur: considering that the `collect_errors` can be empty wouldn't it be better to store a `tl::optional<std::vector<Error>&>` instead of a `std::vector<Error> *` in `ResolutionBuilder` and directly pass the optional reference to the builder? this way, we don't need to check for a value or anything, just assign it before resolving and be done with it.

gcc/rust/resolve/rust-name-resolution-context.h
```diff
@@ -619,37 +588,136 @@ class NameResolutionContext
...
+
+  private:
+    ResolutionPath path;
+    ResolutionMode mode;
+    bool has_path_set;
+
+    static constexpr int NS_VALUE = 1;
+    static constexpr int NS_TYPE = 2;
+    static constexpr int NS_LABEL = 4;
+    static constexpr int NS_MACRO = 8;
```
CohenArthur: what are these for?

gcc/rust/resolve/rust-forever-stack.hxx
powerboat9: It's meant to handle paths that had started with a lang item lookup, since `ResolutionPath` doesn't treat lang prefixes as a segment.

gcc/rust/resolve/rust-name-resolution-context.h
powerboat9: We could, but optional references are a bit cursed (weird assignment behavior, not supported by C++17 std)

gcc/rust/resolve/rust-name-resolution-context.h
powerboat9: Dead code, I'll remove it

gcc/rust/resolve/rust-name-resolution-context.h
CohenArthur: yes, but TartanLlama's optional handles them properly, better than the STL's `optional`. which is also why I never want to switch to `std::optional` lol

https://thephd.dev/to-bind-and-loose-a-reference-optional

PR #4430, opened 2026-02-10
Remove `ProcMacroInvocLexer`

powerboat9:
`ProcMacroInvocLexer` and `MacroInvocLexer` were only separate because one stored instances of `const_TokenPtr` directly, while the other stored instances of `AST::Token` (a thin wrapper around `const_TokenPtr`). This patch removes `ProcMacroInvocLexer` and makes `MacroInvocLexer` directly store `const_TokenPtr` instances.


P-E-P (approved):
I was a bit worried this would break some kind of compat with libproc_macro_internal initially but looks all good. LGTM!

PR #4435, opened 2026-02-12
ast: Fix missing location info in derive macros

Pasta-coder:
Fixes Rust-GCC#4433

gcc/rust/ChangeLog:

	* ast/rust-ast.cc (Attribute::get_traits_to_derive): Propagate location info from MetaWord to SimplePath.

gcc/testsuite/ChangeLog:

	* rust/compile/issue-4433.rs: Update test to expect valid line numbers.

### DIFF
--- a/gcc/rust/ast/rust-ast.cc
+++ b/gcc/rust/ast/rust-ast.cc
@@ -302,8 +302,9 @@ Attribute::get_traits_to_derive ()
 			auto word = static_cast<AST::MetaWord *> (meta_item);
 			// Convert current word to path
 			current = std::make_unique<AST::MetaItemPath> (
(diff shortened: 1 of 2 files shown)


gcc/rust/ast/rust-macro.h
```diff
@@ -955,6 +955,11 @@ class MetaWord : public MetaItem
 
   void accept_vis (ASTVisitor &vis) override;
 
+  SimplePath to_path_item () const override
```
P-E-P: Is this function used anywhere ?

Pasta-coder: no , removed .

PR #4437, opened 2026-02-14
expand: Fix ICE on unimplemented RustcEncodable/Decodable derives

Pasta-coder:
Fixes Rust-GCC#3951

### DIFF
--- a/gcc/rust/expand/rust-derive.cc
+++ b/gcc/rust/expand/rust-derive.cc
@@ -74,6 +74,13 @@ DeriveVisitor::derive (Item &item, const Attribute &attr,
       return vec (DeriveOrd (DeriveOrd::Ordering::Total, loc).go (item));
     case BuiltinMacro::PartialOrd:
       return vec (DeriveOrd (DeriveOrd::Ordering::Partial, loc).go (item));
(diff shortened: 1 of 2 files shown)


gcc/rust/expand/rust-derive.cc
```diff
@@ -74,6 +74,13 @@ DeriveVisitor::derive (Item &item, const Attribute &attr,
       return vec (DeriveOrd (DeriveOrd::Ordering::Total, loc).go (item));
     case BuiltinMacro::PartialOrd:
       return vec (DeriveOrd (DeriveOrd::Ordering::Partial, loc).go (item));
+    case BuiltinMacro::RustcEncodable:
+    case BuiltinMacro::RustcDecodable:
+      rust_sorry_at (loc, "derive(%s) is not yet implemented",
+		     to_derive == BuiltinMacro::RustcEncodable
+		       ? "RustcEncodable"
+		       : "RustcDecodable");
```
CohenArthur, line 82: I think this implementation is fine for now, but we will need to think about how to actually handle these derives later on. As far as I can tell, they are only used in the `backtrace` crate for Rust 1.49

Issue #4409, opened 2026-01-25
Duplicate metavariables allowed in unused macro definitions


powerboat9:
### Summary

Duplicate metavariables are allowed in macro definitions, as long as those macro definitions are unused. Also, we don't warn when a macro definition is unused. See #4405

### Reproducer

I tried this code:

```rust
macro_rules! foo {
    ($a:ident, $a:ident) => {()}
}

pub fn foo() {
    //foo!(a, b)
}
```

### Does the code make use of any (1.49) nightly feature ?

- [ ] Nightly

### Godbolt link

https://godbolt.org/z/5G3ad4bTo

### Actual behavior

No error

### Expected behavior

```
error: duplicate matcher binding
 --> <source>:2:16
  |
2 |     ($a:ident, $a:ident) => {()}
  |      --------  ^^^^^^^^ duplicate binding
  |      |
  |      previous binding

warning: unused macro definition
 --> <source>:1:1
  |
1 | / macro_rules! foo {
2 | |     ($a:ident, $a:ident) => {()}
3 | | }
  | |_^
  |
  = note: `#[warn(unused_macros)]` on by default
```

### GCC Version

d47331f13c3f16262ffa4fbe6f919b83f1056d7f

hriztam:
I could take this issue @powerboat9 @evanphx , can you assign this to me

PR #4444, opened 2026-02-21
fixed the duplicate metavariables issue

hriztam:
Addresses #4409.

This change moves duplicate matcher binding validation for `macro_rules!` from invocation-time matching into macro-definition validation during early name resolution. This makes duplicate metavariable bindings fail even when the macro is never invoked, matching rustc’s behavior for `duplicate matcher binding`.

It also adds an unused macro-definition warning pass in expansion by scanning AST macro definitions and checking whether each definition is referenced by a resolved invocation (excluding builtins and `#[macro_export]` cases), producing `unused macro definition`.

The expander-side duplicate-binding guard logic was removed since diagnostics are now emitted at definition validation time.

Tests:
- Updated `gcc/testsuite/rust/compile/macros/mbe/macro-duplicate-binding.rs` to cover the uninvoked reproducer and assert both:
  - `duplicate matcher binding` error
  - `unused macro definition` warning

Also I tried to format the code using clang-format but that resulted in a much larger pr with a lot of files formatted at once, so I reverted it for the time being. Let me know if that is something which is needed

Signed-off-by: Hritam Shrivastava <hritamstark05@gmail.com>


P-E-P (changes requested):
Some files are artifacts generated from the configure script and should not be pushed.

hriztam:
ok let me update the PR

CohenArthur:
@hriztam changes to the .gitignore that are due to your personal setup shouldn't be pushed :) also the testsuite is failing in the CI, please make sure it passes before we can merge this. Finally before merging you'll need to squash your commits that add and remove the generated files. thanks!

PR #4450, opened 2026-02-25
Prevent error emission with multiple wildcards

P-E-P:
Every wildcard should be considered different and no error should be emitted when multiple wildcards are used within the same rib.


powerboat9:
I'm not sure wildcards should be making it to the `insert_or_error_out` call, especially since I think it only applies to `const` items (maybe not even `static` items?)

P-E-P:
> I'm not sure wildcards should be making it to the `insert_or_error_out` call, especially since I think it only applies to `const` items (maybe not even `static` items?)

I didn't know static items could not be used with wildcard, thank you for noticing.

https://godbolt.org/z/bnEMn7dqz

I think I'll even rework the parser a bit to have a similar error message.

powerboat9:
Looks like structs/enums also can't be '_'

Issue #4411, opened 2026-01-28
ICE resolve_path


matthiaskrgr:
auto-reduced (treereduce-rust):
````rust
//@compile-flags: -frust-incomplete-and-experimental-compiler-do-not-use
mod framing {

    pub use ::*;
}

````

original:
````rust
// https://github.com/rust-lang/rust/pull/113099

mod framing {
    mod public_message {
        use super::super::*;

        #[derive(Debug)]
        pub struct ConfirmedTranscriptHashInput;
    }

    mod public_message_in {
        use super::*;

        #[derive(Debug)]
        pub struct ConfirmedTranscriptHashInput;
    }

    pub use self::public_message::*;
    pub use self::public_message_in::*;
}

use crate::framing::ConfirmedTranscriptHashInput;
//~^ ERROR `ConfirmedTranscriptHashInput` is ambiguous
//~| WARNING this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!

fn main() { }

````

Version information:

145e66efdf603476ef6b0d66523936ff55187162

Possibly related line of code:
https://github.com/Rust-GCC/gccrs/blob/145e66efdf603476ef6b0d66523936ff55187162/gcc/rust/resolve/rust-forever-stack.hxx#L710-L722

Command:
`/home/matthias/vcs/github/gccrs/gccrs-build/gcc/crab1 -frust-incomplete-and-experimental-compiler-do-not-use`

<details><summary><strong>Program output</strong></summary>
<p>

```
crab1: error: failed to locate crate ‘core’
crab1: error: unknown crate ‘core’
crab1: internal compiler error: in resolve_path, at rust/resolve/rust-forever-stack.hxx:716
0x30d80b9 internal_error(char const*, ...)
	../../gcc/diagnostic-global-context.cc:787
0xc3ae5d fancy_abort(char const*, int, char const*)
	../../gcc/diagnostics/context.cc:1805
[...]

PR #4451, opened 2026-02-27
resolve: Fix ICE on ambiguous glob re-exports

Harishankar14:
This patch adds an ambiguity check in 'finalize_rebind_import'. If a Definition is ambiguous, it emits a proper error diagnostic instead of crashing, consistent with rustc's behavior(verified)

Fixes Rust-GCC/gccrs#4411

gcc/rust/ChangeLog:

	* resolve/rust-early-name-resolver-2.0.cc (Early::finalize_rebind_import): Add ambiguity check before calling get_node_id() on glob import definitions.

gcc/testsuite/ChangeLog:

	* rust/compile/issue-4411.rs: New test.

in the git commit message.

- \[ ] Run `clang-format`

order to get feedback.
---

### DIFF
--- a/gcc/rust/resolve/rust-early-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-early-name-resolver-2.0.cc
@@ -123,6 +123,17 @@ Early::resolve_rebind_import (NodeId use_dec_id,
   // if we've found at least one definition, then we're good
   if (definitions.empty ())
     return false;
+  for (const auto &def : definitions)
+    {
+      if (def.first.is_ambiguous ())
+	{
+	  rich_location rich_locus (line_table,
+				    rebind_import.to_resolve.get_locus ());
+	  rust_error_at (rich_locus, ErrorCode::E0659, "%qs is ambiguous",
+			 rebind_import.to_resolve.as_string ().c_str ());
+	  return true;
+	}
+    }
 
   auto &imports = import_mappings.new_or_access (use_dec_id);
 
--- a/gcc/testsuite/rust/compile/issue-4411.rs
+++ b/gcc/testsuite/rust/compile/issue-4411.rs
@@ -1,0 +1,16 @@
+#![feature(no_core)]
+#![no_core]
+mod framing {
+    mod public_message {
+        pub struct ConfirmedTranscriptHashInput;
+    }
+    mod public_message_in {
+        pub struct ConfirmedTranscriptHashInput;
+    }
+    pub use self::public_message::*;
+    pub use self::public_message_in::*;
+}
+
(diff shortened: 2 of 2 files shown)


Harishankar14:
@P-E-P  @CohenArthur @philberty

P-E-P:
I think this directly collides with some work from @CohenArthur that refactors the whole re-export mechanisms, I'll let him decide whether we should accept those changes or wait for the use declaration rework.

Harishankar14:
@P-E-P  Alright.

Harishankar14:
@CohenArthur  Thank you. I have updated the changes !!

gcc/testsuite/rust/compile/issue-4411.rs
```diff
@@ -0,0 +1,17 @@
+// { dg-options "-frust-incomplete-and-experimental-compiler-do-not-use" }
```
P-E-P: This option should not be required for the testsuite

gcc/rust/resolve/rust-early-name-resolver-2.0.cc
```diff
@@ -461,8 +472,10 @@ Early::finalize_rebind_import (const Early::ImportPair &mapping)
     }
 
   for (auto &&definition : data.definitions ())
-    toplevel.insert_or_error_out (
-      declared_name, locus, definition.first.get_node_id (), definition.second /* TODO: This isn't clear - it would be better if it was called .ns or something */);
+    {
```
P-E-P: This hunk look identical, even though it is not enforced by the clang format one line if/else/for/while shall not have any braces. I suggest you discard this hunk from your commit.

gcc/testsuite/rust/compile/issue-4411.rs
Harishankar14: sorry, that was dodgy from myside

Harishankar14:
@P-E-P  Thank you for addressing the small details that i had  missed. I have fixed it !!

gcc/rust/resolve/rust-early-name-resolver-2.0.cc
```diff
@@ -461,6 +472,7 @@ Early::finalize_rebind_import (const Early::ImportPair &mapping)
     }
 
   for (auto &&definition : data.definitions ())
+
```
P-E-P: This blank line isn't required and is confusing

Harishankar14:
@P-E-P  should be okay now ?

gcc/rust/resolve/rust-early-name-resolver-2.0.cc
Harishankar14: yep, me being silly again :)

PR #4452, opened 2026-02-27
intrinsic: Add bswap intrinsic

nsvke:
This patch implements the bswap intrinsic for integer types. It maps to the respective GCC built-ins for 16, 32, 64 and 128-bit integers.

Key changes:
- Implemented `bswap_handler` in `rust-compile-intrinsic.cc` to map the intrinsic to GCC's built-ins.
- Registered the `bswap` intrinsic in the `generic_intrinsics`.
- Added functional tests in `gcc/testsuite/rust/execute/` and type-checking tests in `gcc/testsuite/rust/compile/`.

The implementation has been verified against the current master branch.

gcc/rust/ChangeLog:

	* backend/rust-compile-intrinsic.cc (bswap_handler): New function.

gcc/testsuite/ChangeLog:

	* rust/compile/bswap.rs: New test.
	* rust/execute/bswap.rs: New test.

Addresses #658

### DIFF
--- a/gcc/testsuite/rust/compile/bswap.rs
+++ b/gcc/testsuite/rust/compile/bswap.rs
@@ -1,0 +1,17 @@
+// { dg-do compile }
+#![feature(intrinsics, lang_items, no_core)]
+#![no_core]
+
+#[lang = "sized"]
+pub trait Sized {}
+
+#[lang = "copy"]
+pub trait Copy {}
+
+extern "rust-intrinsic" {
+    pub fn bswap<T>(x: T) -> T; // { dg-error "bswap intrinsics can only be used with basic integer types .got 'bool'." }
+}
+
+fn main() {
+    let _ = bswap(true);
+}
--- a/gcc/testsuite/rust/execute/bswap.rs
+++ b/gcc/testsuite/rust/execute/bswap.rs
@@ -1,0 +1,263 @@
+// { dg-do run }
+// { dg-options "-O2" }
+
+/*
+ * Note on test design:
+ * This test uses a union-based type punning approach rather than standard
+ * integer literals (e.g., 0x..._u128) or bitwise shifts for 64-bit and 128-bit types.
+ * We do this for a few practical reasons:
+ *  gccrs doesn't support 128-bit integer literals yet.
+ *  Doing manual bitwise shifts on 32-bit targets currently triggers ICEs in the frontend.
+ *  Casting via raw pointers gets optimized away by GCC under -O2 due to strict aliasing rules.
+ * Using a #[repr(C)] union gives us a safe workaround.
+ *  As a bonus, by asserting against the physical memory layout ([u8; N]) instead
+ * of the logical integer value, this test is inherently endian-agnostic.
+ */
+
+#![feature(intrinsics, lang_items, no_core)]
+#![no_core]
+
+#[lang = "sized"]
+pub trait Sized {}
+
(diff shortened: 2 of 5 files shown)


nsvke:
​Identified the issue causing the ICE on ```32-bit``` targets. It appears ```__builtin_bswap128``` is not available on all target architectures. I plan to implement a fallback mechanism to ensure cross-platform compatibility. I will update the PR as soon as I'm back at my desk. Cheers!

gcc/testsuite/rust/execute/bswap.rs
```diff
@@ -0,0 +1,269 @@
...
+    val: u128,
+}
+#[repr(C)]
+union I128Pun {
+    bytes: [u8; 16],
+    val: i128,
+}
+
+fn test_bswap_16_and_32() {
+    unsafe {
```
P-E-P: unless this has changed after 1.49 both bswap and abort are safe intrinsics

gcc/testsuite/rust/compile/bswap.rs
```diff
@@ -0,0 +1,17 @@
...
+#![no_core]
+
+#[lang = "sized"]
+pub trait Sized {}
+
+#[lang = "copy"]
+pub trait Copy {}
+
+extern "rust-intrinsic" {
+    pub fn bswap<T>(x: T) -> T; // { dg-error "bswap intrinsics can only be used with basic integer types .got 'bool'." }
```
P-E-P, line 12: This error is weirdly placed, it should be emitted on call location instead of intrinsic definition.

nsvke: I agree that pointing to the call site provides much better UX. I initially followed the existing pattern (e.g., [testsuite/rust/compile/torture/transmute-size-check-1.rs](https://github.com/Rust-GCC/gccrs/blob/master/gcc/testsuite/rust/compile/torture/transmute-size-check-1.rs), where the error is also emitted at the intrinsic definition.

The core issue is that bswap_handler operates during the backend code generation and only receives Context and TyTy::FnType. This strictly limits access to the specific CallExpr's instantiation location.

Is there a preferred way to access the call site locus within the current backend structure? Or should this specific bounds check be moved earlier to the typecheck phase to properly capture the call location?

gcc/testsuite/rust/execute/bswap.rs
nsvke: You're right. I've fixed this locally and will include it in my next push once we figure out the locus issue.

gcc/testsuite/rust/compile/bswap.rs
P-E-P: Yeah this should probably be handled somewhere else earlier, probably within the typechecker. Since every other handler also treats it like this I'm willing to merge this PR without changing the location but we should probably make it an assert then and emit the correct error message earlier in another PR.

nsvke: Just to make sure I understand correctly: do you mean I should replace the error emission with an assert inside bswap_handler right now, and temporarily mark the compile/bswap.rs test as xfail (or dg-ice) for this PR?

P-E-P: I forgot the test would fail, I don't think it is worth making it an XFAIL given all other intrinsics have this behavior. Please keep it with the incorrect location for now, I will create an issue for all intrinsics.

PR #4454, opened 2026-02-28
gccrs: Fix 128-bit non-decimal integer literal saturation

### LINKED ISSUE #4453: Lexer: Non-decimal integer literals (hex, bin, oct) are saturated at 64-bit limits
### Summary

Currently, the Rust frontend's lexer handles non-decimal integer literals (hexadecimal, binary, and octal) by converting them into decimal strings using std::strtol. Since std::strtol is limited by the size of a 64-bit long (on 64-bit systems), any literal exceeding this range (such as large u128 values) is silently saturated to LONG_MAX (2^63−1) or ULONG_MAX. This leads to incorrect constant values being passed to the backend.

### Reproducer

I tried this code:

```rust
#![feature(no_core)]
#![no_core]
pub fn main() {
        let hex_val: u128 = 0x10000000000000000_u128;
        let bin_val: u128 =
            0b10000000000000000000000000000000000000000000000000000000000000000_u128;
        let oct_val: u128 = 0o2000000000000000000000_u128;
}
```

- [ ] Nightly

### Godbolt link

Reproduced on local build of gccrs

#### For the reproducible code above

```
[DEBUG-OLD] RADIXSTRINGHEX: 0x10000000000000000
[DEBUG-OLD] DECIMALSTRINGHEX: 9223372036854775807
```
```
[DEBUG-OLD] RADIXSTRINGBIN: 010000000000000000000000000000000000000000000000000000000000000000
[DEBUG-OLD] DECIMALSTRINGBIN: 9223372036854775807
```
```
[DEBUG-OLD] RADIXSTRINGOCT: 02000000000000000000000
[DEBUG-OLD] DECIMALSTRINGOCT: 9223372036854775807
```

#### For the reproducible code above

```
[DEBUG-NEW] RADIXSTRINGHEX: 010000000000000000
[DEBUG-NEW] DECIMALSTRINGHEX: 18446744073709551616
```
```
[DEBUG-NEW] RADIXSTRINGBIN: 010000000000000000000000000000000000000000000000000000000000000000
[DEBUG-NEW] DECIMALSTRINGBIN: 18446744073709551616
```
```
[DEBUG-NEW] RADIXSTRINGOCT: 02000000000000000000000
[DEBUG-NEW] DECIMALSTRINGOCT: 18446744073709551616
```

### GCC Version

commit hash: 6b891ee3e6b879e402b037c1322a8350c72b7798

nsvke:
### Description
This PR addresses the saturation bug where large non-decimal integer literals (such as 128-bit hex, bin, and octal values) were truncated to 64-bit limits (`LONG_MAX`) during the lexing phase.

**Key Changes:**
* Replaced the usage of `std::strtol` in `Lexer::parse_non_decimal_int_literal` with GNU MP (GMP) for arbitrary-precision base conversion.
* Corrected the hex dispatcher in `Lexer::parse_non_decimal_int_literals` to pass the raw string without prepending the `"0x"` prefix, ensuring compatibility with `mpz_set_str`.
* Added a regression test (`non_decimal_128_saturation.rs`) to verify that $2^{64}$ limits are properly exceeded without truncation.

gcc/rust/ChangeLog:

	* lex/rust-lex.cc (Lexer::parse_non_decimal_int_literal): Use GMP for base conversion to support 128-bit literals.
	(Lexer::parse_non_decimal_int_literals): Fix hex prefix inconsistency by passing pure string.

gcc/testsuite/ChangeLog:

	* rust/execute/non_decimal_128_saturation.rs: New test.

Closes #4453

### DIFF
--- a/gcc/rust/lex/rust-lex.cc
+++ b/gcc/rust/lex/rust-lex.cc
@@ -2239,8 +2239,10 @@ Lexer::parse_raw_string (location_t loc, int initial_hash_count)
 template <typename IsDigitFunc>
 TokenPtr
 Lexer::parse_non_decimal_int_literal (location_t loc, IsDigitFunc is_digit_func,
-				      std::string existent_str, int base)
+				      int base)
 {
+  std::string raw_str;
+
   int length = 1;
 
   skip_input ();
@@ -2265,15 +2267,19 @@ Lexer::parse_non_decimal_int_literal (location_t loc, IsDigitFunc is_digit_func,
       length++;
 
       // add raw numbers
-      existent_str += current_char;
+      raw_str += current_char;
       skip_input ();
       current_char = peek_input ();
     }
 
   // convert value to decimal representation
-  long dec_num = std::strtol (existent_str.c_str (), nullptr, base);
-
-  existent_str = std::to_string (dec_num);
+  mpz_t dec_num;
+  mpz_init (dec_num);
+  mpz_set_str (dec_num, raw_str.c_str (), base);
+  char *s = mpz_get_str (NULL, 10, dec_num);
+  std::string dec_str = s;
+  free (s);
+  mpz_clear (dec_num);
 
   // parse in type suffix if it exists
   auto type_suffix_pair = parse_in_type_suffix ();
@@ -2297,35 +2303,29 @@ Lexer::parse_non_decimal_int_literal (location_t loc, IsDigitFunc is_digit_func,
 
   loc += length - 1;
 
-  return Token::make_int (loc, std::move (existent_str), type_hint);
+  return Token::make_int (loc, std::move (dec_str), type_hint);
 }
 
 // Parses a hex, binary or octal int literal.
 TokenPtr
 Lexer::parse_non_decimal_int_literals (location_t loc)
 {
-  std::string str;
-  str.reserve (16); // some sensible default
-  str += current_char;
-
   current_char = peek_input ();
 
   if (current_char == 'x')
     {
       // hex (integer only)
-      return parse_non_decimal_int_literal (loc, is_x_digit, str + "x", 16);
+      return parse_non_decimal_int_literal (loc, is_x_digit, 16);
     }
   else if (current_char == 'o')
     {
       // octal (integer only)
-      return parse_non_decimal_int_literal (loc, is_octal_digit,
-					    std::move (str), 8);
+      return parse_non_decimal_int_literal (loc, is_octal_digit, 8);
     }
   else if (current_char == 'b')
     {
       // binary (integer only)
-      return parse_non_decimal_int_literal (loc, is_bin_digit, std::move (str),
-					    2);
+      return parse_non_decimal_int_literal (loc, is_bin_digit, 2);
     }
   else
     {
--- a/gcc/testsuite/rust/execute/non_decimal_128_saturation.rs
+++ b/gcc/testsuite/rust/execute/non_decimal_128_saturation.rs
@@ -1,0 +1,24 @@
+// { dg-do run }
+// { dg-require-effective-target lp64 }
+// { dg-options "-O0" }
+#![feature(no_core)]
+#![no_core]
+
+fn main() -> i32 {
+    let hex_val: u128 = 0x1_0000_0000_0000_0000_u128;
+    if (hex_val >> 64) as u8 != 1 {
+        return 1;
+    }
+
+    let bin_val: u128 =
+        0b1_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000_u128;
+    if (bin_val >> 64) as u8 != 1 {
+        return 1;
+    }
+
+    let oct_val: u128 = 0o2_000_000_000_000_000_000_000_u128;
+    if (oct_val >> 64) as u8 != 1 {
+        return 1;
+    }
+    0
+}
--- a/gcc/rust/lex/rust-lex.h
+++ b/gcc/rust/lex/rust-lex.h
@@ -154,8 +154,7 @@ private:
 
   template <typename IsDigitFunc>
   TokenPtr parse_non_decimal_int_literal (location_t loc,
-					  IsDigitFunc is_digit_func,
-					  std::string existent_str, int base);
+					  IsDigitFunc is_digit_func, int base);
 
 public:
   // Construct lexer with input file and filename provided


gcc/rust/lex/rust-lex.cc
```diff
@@ -2313,7 +2317,8 @@ Lexer::parse_non_decimal_int_literals (location_t loc)
   if (current_char == 'x')
     {
       // hex (integer only)
-      return parse_non_decimal_int_literal (loc, is_x_digit, str + "x", 16);
```
powerboat9, line 2316: I'm not sure what this was doing, since it's *appending* an `x`

gcc/testsuite/rust/execute/non_decimal_128_saturation.rs
```diff
@@ -0,0 +1,29 @@
...
+        let bin_val: u128 =
+            0b10000000000000000000000000000000000000000000000000000000000000000_u128;
+        if *((&bin_val as *const u128 as usize + 8) as *const u8) != 1 {
+            abort();
+        }
+
+        let oct_val: u128 = 0o2000000000000000000000_u128;
+        if *((&oct_val as *const u128 as usize + 8) as *const u8) != 1 {
+            abort();
+        }
```
powerboat9: This could break with little vs big endian. You should probably do something like `(oct_val >> 64) as u8 == 1` instead.

nsvke:
Thanks for the feedback! I'll update the test shortly.

gcc/testsuite/rust/execute/non_decimal_128_saturation.rs
nsvke: You are absolutely right. I initially used this workaround because I've previously encountered Internal Compiler Errors when performing large shifts on 128-bit values in 32-bit environments. I'll update the test and ensure it's as robust as possible for all targets.

powerboat9: Yeah, you might hit that. If you do, you can disable the test on 32 bit targets

gcc/testsuite/rust/execute/non_decimal_128_saturation.rs
```diff
@@ -0,0 +1,30 @@
...
+
+extern "C" {
+    fn abort();
+}
+
+pub fn main() -> i32 {
+    unsafe {
+        let hex_val: u128 = 0x10000000000000000_u128;
+        if (hex_val >> 64) as u8 != 1 {
+            abort();
```
P-E-P: I would change this to `return 1` and check for the status using dejaGNU to avoid relying on the extern abi and keep the test somewhat minimal.

nsvke: Thanks for the suggestion! I removed the extern abi and abort calls. The test now simply returns 1 on failure and 0 on success to leverage dejaGNU's exit status checking. It's indeed much cleaner and minimal this way.

gcc/rust/lex/rust-lex.cc
powerboat9: I don't think `existent_str` needs to be a parameter -- seems like it's just adding a `0x` or `0` prefix

nsvke: You are absolutely right. The original code passed 0x or 0 via existent_str based on a switch-case. However, since my mpz implementation doesn't need the 0x prefix and we already have the base parameter, passing it through the signature is now redundant.

 I will remove existent_str from the parameters, clean up the function signature, and rename the local variable to something more appropriate. I'll push the fix as soon as I get back to my computer.

gcc/rust/lex/rust-lex.cc
```diff
@@ -2265,15 +2267,19 @@ Lexer::parse_non_decimal_int_literal (location_t loc, IsDigitFunc is_digit_func,
...
   // convert value to decimal representation
-  long dec_num = std::strtol (existent_str.c_str (), nullptr, base);
-
-  existent_str = std::to_string (dec_num);
+  mpz_t dec_num;
+  mpz_init (dec_num);
+  mpz_set_str (dec_num, raw_str.c_str (), base);
+  char *s = mpz_get_str (NULL, 10, dec_num);
+  std::string dec_str = s;
+  free (s);
```
moabo3li, line 2281: GMP may use a custom memory allocator, so `free()` isn't guaranteed to match.
take a look https://gmplib.org/manual/Converting-Integers

```suggestion
  void (*freefunc)(void *, size_t);                                                                                 
  mp_get_memory_functions (NULL, NULL, &freefunc);                                                                  
  freefunc (s, strlen (s) + 1); 
```
https://gmplib.org/manual/Custom-Allocation

gcc/rust/lex/rust-lex.cc
```diff
@@ -2265,15 +2267,19 @@ Lexer::parse_non_decimal_int_literal (location_t loc, IsDigitFunc is_digit_func,
...
       current_char = peek_input ();
     }
 
   // convert value to decimal representation
-  long dec_num = std::strtol (existent_str.c_str (), nullptr, base);
-
-  existent_str = std::to_string (dec_num);
+  mpz_t dec_num;
+  mpz_init (dec_num);
+  mpz_set_str (dec_num, raw_str.c_str (), base);
```
moabo3li, line 2278: `mpz_set_str` return value unchecked return -1 on invalid input.
Digits are already validated by `is_digit_func` 
assert checking this value would be dfensive.

```suggestion
  int ret = mpz_set_str (dec_num, raw_str.c_str (), base);
  gcc_assert (ret == 0);
```

gcc/rust/lex/rust-lex.cc
nsvke: This document covers general usage. Looking at the GMP usages across the GCC source, they all call `free()` directly

gcc/rust/lex/rust-lex.cc
nsvke: Yes, it should be fine since it's validated by is_digit_func. That said, it might still be a good idea, I'll add it.

nsvke:
> It would be great to test the other integer types (`u/i8`, `u/i16`, `u/i32`, `u/i64`, and `i128`) using their maximum and minimum values.

The test currently checks the smallest value that exceeds the minimum for each type, i.e., it verifies whether an overflow occurs. As long as this works, I don't think testing min or max values would broaden the test coverage.

nsvke:
> If the input is `0x_u128`, it should fail with the error `no valid digits found for number`, similar to `rustc`.

We are aware of this. There is an issue for it (#4490) and you can check the related discussions on Zulip. This was not addressed in this PR to avoid scope creep.

moabo3li:
> > It would be great to test the other integer types (`u/i8`, `u/i16`, `u/i32`, `u/i64`, and `i128`) using their maximum and minimum values.
> 
> The test currently checks the smallest value that exceeds the minimum for each type, i.e., it verifies whether an overflow occurs. As long as this works, I don't think testing min or max values would broaden the test coverage.

I was thinking that since we are going to change the way we parse literals, this will not only affect 128-bit types.
We already encountered some edge cases, like `Alpine 32-bit CI` failing at the maximum value, and there are no tests covering this.
(If this is outside the scope of this PR, that’s totally fine.)

PR #4455, opened 2026-03-01
Disambiguate identifier and path patterns

powerboat9:


powerboat9:
@CohenArthur it *seems* like this is breaking because the identifier is getting resolved to the use statement and not the enum item -- thoughts?

powerboat9:
Turns out that was actually an issue with match arm scoping

powerboat9:
Now depends on #4480

powerboat9:
Looks like this still isn't working with glob use declarations

powerboat9:
Now depends on #4499 (also, glob use declaration handling was indirectly fixed in #4480)

powerboat9:
Should be ready now

gcc/rust/resolve/rust-identifier-path.h
```diff
@@ -0,0 +1,50 @@
...
+class IdentifierPathPass : public AST::PointerVisitor
+{
+public:
+  IdentifierPathPass (NameResolutionContext &ctx,
+		      std::set<NodeId> ident_path_to_convert);
+
+  static void go (AST::Crate &crate, NameResolutionContext &ctx,
+		  std::set<NodeId> ident_path_to_convert);
+
+  using AST::PointerVisitor::reseat;
```
P-E-P, line 38: Is this re-export required ? IIRC reseat is not called outside of the visitor.

gcc/testsuite/rust/execute/ident_pat_vs_path_2.rs
```diff
@@ -0,0 +1,18 @@
...
+    A,
+    B,
+    C
+}
+
+fn main() -> i32 {
+    use E::*;
+
+    match A {
+        C => 1,
```
P-E-P, line 15: I believe we should also have a testcase were the identifier is a path pattern.

P-E-P (approved):

gcc/testsuite/rust/execute/ident_pat_vs_path_2.rs
powerboat9: What do you mean? I think `gcc/testsuite/rust/execute/ident_pat_vs_path_1.rs` covers that.

gcc/rust/resolve/rust-identifier-path.h
powerboat9: It's probably not required, but it should prevent weird behavior if anyone ever tried to call `reseat`, and it's effectively a single line of boilerplate.

gcc/testsuite/rust/execute/ident_pat_vs_path_2.rs
P-E-P: Yes, I focused too hard on that C and did not see B

Issue #3372, opened 2025-01-20
Add `black_box` intrinsic


CohenArthur:
Existing GCC implementation by @antoyo: https://github.com/rust-lang/rustc_codegen_gcc/blob/8037b6139fea50894978509744f00484150e6816/src/intrinsic/mod.rs#L366

PR #2360, opened 2023-07-02
gccrs: minor HIR interface cleanup

dkm:
Minor changes in HIR interface: add missing getters, make some classes
inherit from FullVisitable.

gcc/rust/ChangeLog:

	* hir/tree/rust-hir-expr.h (class ArrayElems): Inherit from FullVisitable.
	(class StructExprField): Likewise.
	* hir/tree/rust-hir-item.h (class WhereClauseItem): Likewise.
	(class UseTree): Likewise.
	(UseTreeRebind::get_path, UseTreeRebind::get_identifier)
	(UseTreeRebind::get_bind_type): New.
	(UseDeclaration::get_use_tree): New.
	(struct TraitFunctionDecl): Change struct to ...
	(class TraitFunctionDecl): ... class.
	(TraitFunctionDecl::get_where_clause): New.
	(StructField::get_outer_attrs): New.
	(struct TupleField): Change struct to ...
	(class TupleField): ... class.
	(TupleField::get_visibility, TupleField::get_outer_attrs): New.
	* hir/tree/rust-hir-pattern.h (class TupleStructItems): Inherit
	from FullVisitable.
	* hir/tree/rust-hir-pattern.h (class TuplePatternItems): Likewise

Signed-off-by: Marc Poulhiès <dkm@kataplop.net>

PR #4456, opened 2026-03-02
intrinsic: Add black_box intrinsic

Pasta-coder:
This patch implements the `black_box` intrinsic, which is used to prevent the compiler from optimizing away a piece of code. It maps to an empty volatile inline assembly block with a memory clobber and a general input constraint, matching the behavior expected by the standard library.

Addresses #3372

gcc/rust/ChangeLog:

	* backend/rust-compile-intrinsic.cc (black_box_handler): New handler.

gcc/testsuite/ChangeLog:

	* rust/compile/intrinsic-black-box.rs: New test.
	* rust/execute/intrinsic-black-box.rs: New test.

### DIFF
--- a/gcc/rust/backend/rust-compile-intrinsic.cc
+++ b/gcc/rust/backend/rust-compile-intrinsic.cc
@@ -81,6 +81,7 @@ static tree uninit_handler (Context *ctx, TyTy::FnType *fntype);
 static tree move_val_init_handler (Context *ctx, TyTy::FnType *fntype);
 static tree assume_handler (Context *ctx, TyTy::FnType *fntype);
 static tree discriminant_value_handler (Context *ctx, TyTy::FnType *fntype);
+static tree black_box_handler (Context *ctx, TyTy::FnType *fntype);
 static tree variant_count_handler (Context *ctx, TyTy::FnType *fntype);
 
 enum class Prefetch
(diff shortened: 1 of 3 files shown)


P-E-P:
I don't think this was already available with rust 1.49 ?

Pasta-coder:
While `black_box` was stabilized in `std::hint` in 1.66, it has existed as a core compiler-internal intrinsic since 2019 (see [RFC #2360](https://github.com/rust-lang/rfcs/pull/2360) and tracking issue [#64102](https://github.com/rust-lang/rust/issues/64102)).

This implementation follows the same "infrastructure parity" logic as the existing `assume` and `likely` handlers in `rust-compile-intrinsic.cc`. Although their stable library wrappers (`like assert_unchecked` in 1.81) appeared in much later Rust versions, the underlying intrinsics were 1.49-era requirements for the test crate and core library optimization barriers.

P-E-P:
I'm not sure I'm following you here, black box was indeed available in 1.49 but what I meant really was "did it exist as an intrinsic yet ?".
As you can see here black box was then implemented using inline assembly https://github.com/rust-lang/rust/blob/1.49.0/library/core/src/hint.rs#L159

Pasta-coder:
yeah ,  you are right . i just gazed over that file . disregard my comment .

PR #4457, opened 2026-03-03
[WIP] Drop trait demo

Islam-Imad:
The Datum system is the central point for tracking values as they flow through expression compilation. Instead of letting raw tree nodes flow through expression we will have a rich Datum that holds other important info like drop and value kind.

The Drop Place is used to hold all needed information to drop a variable and track its state at runtime if this variable is initialized or uninitialized or partially initialized.

cc @philberty

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ cscope.out
 .local.vimrc
 .lvimrc
 .vscode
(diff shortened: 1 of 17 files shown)


Islam-Imad:
RIP

PR #4460, opened 2026-03-05
Fix match arm scope

powerboat9:
This fixes a bug where bindings introduced by match arms are given the same scope as their match expression.

### DIFF
--- a/gcc/testsuite/rust/compile/match-scope.rs
+++ b/gcc/testsuite/rust/compile/match-scope.rs
@@ -1,0 +1,8 @@
+#![no_core]
+
+pub fn main() -> i32 {
+    match 12 {
+        x => {}
+    }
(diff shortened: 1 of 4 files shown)


gcc/testsuite/rust/compile/match-scope.rs
```diff
@@ -0,0 +1,8 @@
+#![no_core]
```
P-E-P, line 1: Did the test pass without enabling the no_core feature ?

powerboat9: Yes

P-E-P: Indeed we accept `#![no_core]` without the feature enabled. I'll create an issue.

PR #2872, opened 2024-02-24
Add `default` as function qualifier.

braw-lee:
Fixes #2819

PR #4463, opened 2026-03-05
enforce function qualifier order

### LINKED ISSUE #2819: Enforce rust keyword order
Rust keyword have should be used in a given order
1. `pub`
2. `default`
3. `const`
4. `async`
5. `unsafe`
6. `extern`

Rust provides multiple messages and notes to help with that:

```
5 | extern const pub fn myfun() {
  | -------^^^^^
  | |      |
  | |      expected `fn`
  | help: `const` must come before `extern`: `const extern`
  |
  = note: keyword order for functions declaration is `pub`, `default`, `const`, `async`, `unsafe`, `extern`
```

- We do have some `FunctionQualifier` structure to store the state of a function but not all those informations are kept in it.
- We parse in different ways those qualifiers and provide multiple error message.

We should probably unify all those under one same structure (`FunctionQualifier`) and create a function to parse a `FunctionQualifier` that could be used everytime we expect a function.

This function could emit an error as well as a note (`rust_note`) when the syntax is wrong.

Chose an appropriate API for this function. Should it return an `Optional<FunctionQualifier>` ? Always return a `FunctionQualifier` with private/safe/sync attributes ? Probably the latter.

phi-gamma:
Fixes #2819 "enforce rust keyword order", building on top of PR #2872
which got closed due to inactivity.

I made the function qualifier parsing in ``Parser<>::parse_function_qualifiers``
a little more involved to provide a little more constructive feedback
for the user. It will now try and read as many qualifier tokens as
possible (or until it hits a duplicate) so as to respond with them in
the correct order.

Example output:

    /home/phg/src/rust/gccrs/gcc/testsuite/rust/compile/func-qualifier-order-3.rs:11:5: error: invalid order of function qualifiers; found ‘default async unsafe extern const’, expected ‘default const async unsafe extern’
       11 |     default async unsafe extern "C" const fn all_the_qualifiers() {}
          |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Special case for duplicate qualifiers:

    /home/phg/src/rust/gccrs/gcc/testsuite/rust/compile/func-qualifier-order-2.rs:5:14: error: encountered duplicate function qualifier ‘async’
        5 | async unsafe async fn duplicate_qualifier() {}
          |              ^~~~~

Note that I preserved the two commits by @braw-lee in #2872 for their
work to get credited correctly; the commits needed some adapting to
the current state of gccrs.

---

- \[+] GCC development requires copyright assignment or the Developer's Certificate of Origin sign-off, see https://gcc.gnu.org/contribute.html or https://gcc.gnu.org/dco.html
- \[+] Read contributing guidelines
- \[+] `make check-rust` passes locally
- \[+] Run `clang-format`
- \[+] Added any relevant test cases to `gcc/testsuite/rust/`


gcc/rust/parse/rust-parse-impl.hxx
```diff
@@ -1634,70 +1639,170 @@ Parser<ManagedTokenSource>::parse_function (AST::Visibility vis,
...
+  size_t last_priority = 0;
+  for (auto token_id : found_order)
+    {
+      const size_t priority = token_priority (token_id);
+      if (priority <= last_priority)
 	{
-	  lexer.skip_token ();
-	  abi = next_tok->get_str ();
+	  auto qualifiers_to_str = [] (const std::vector<TokenId> &token_ids) {
+	    std::string acc;
```
P-E-P: You should probably use a sstream over here.

gcc/rust/parse/rust-parse-impl.hxx
```diff
@@ -1634,70 +1639,170 @@ Parser<ManagedTokenSource>::parse_function (AST::Visibility vis,
...
+
+	    for (auto id : token_ids)
+	      {
+		if (!acc.empty ())
+		  acc += ' ';
+
+		if (id == IDENTIFIER)
+		  acc += "default";
```
P-E-P: Please use `Rust::WeakKeywords::DEFAULT` instead.

gcc/rust/parse/rust-parse-impl.hxx
```diff
@@ -1557,7 +1559,10 @@ Parser<ManagedTokenSource>::parse_function (AST::Visibility vis,
 {
   location_t locus = lexer.peek_token ()->get_locus ();
   // Get qualifiers for function if they exist
-  AST::FunctionQualifiers qualifiers = parse_function_qualifiers ();
+  std::unique_ptr<AST::FunctionQualifiers> qualifiers
+    = parse_function_qualifiers ();
+  if (qualifiers == nullptr)
```
P-E-P: We're trying to not rely on null pointer checking in the parser, please use `tl::expected<std::unique_ptr<AST::FunctionQualifiers>` instead.

gcc/rust/parse/rust-parse-impl.hxx
```diff
@@ -1634,70 +1639,170 @@ Parser<ManagedTokenSource>::parse_function (AST::Visibility vis,
...
+		       "encountered duplicate function qualifier %qs",
+		       lexer.peek_token ()->get_token_description ());
+	  add_error (std::move (error));
+
+	  return nullptr;
+	}
+
+      switch (token_id)
+	{
+	default:
```
P-E-P: We usually put the default case at the end of the switch, this location might surprise some people.

gcc/rust/parse/rust-parse-impl.hxx
```diff
@@ -1634,70 +1639,170 @@ Parser<ManagedTokenSource>::parse_function (AST::Visibility vis,
...
+	  goto done;
```
P-E-P: You should probably rely on a separate function.

gcc/rust/parse/rust-parse-impl.hxx
```diff
@@ -1634,70 +1639,170 @@ Parser<ManagedTokenSource>::parse_function (AST::Visibility vis,
...
-      has_extern = true;
-
-      // detect optional abi name
-      const_TokenPtr next_tok = lexer.peek_token ();
-      if (next_tok->get_id () == STRING_LITERAL)
+  // Check in order of default, const, async, unsafe, extern
+  auto token_priority = [] (const TokenId id) {
+    switch (id)
+      {
+      default: // unreachable, but will cause an error
```
P-E-P: You could use `gcc_unreachable()`

gcc/rust/parse/rust-parse-impl.hxx
```diff
@@ -1634,70 +1639,170 @@ Parser<ManagedTokenSource>::parse_function (AST::Visibility vis,
...
+
+  size_t last_priority = 0;
+  for (auto token_id : found_order)
+    {
+      const size_t priority = token_priority (token_id);
+      if (priority <= last_priority)
 	{
-	  lexer.skip_token ();
-	  abi = next_tok->get_str ();
+	  auto qualifiers_to_str = [] (const std::vector<TokenId> &token_ids) {
```
P-E-P: This whole found vs expected tokens list building code should be in a separate function.

phi-gamma: Please have a look at the refactor commit that I put on top. It splits up that monster function into a number of smaller ones so parsing, validation, and extracting the qualifier flags are separate steps.

It's a little more invasive hence the extra commit. I'll squash that guy if it passes QA.

gcc/rust/parse/rust-parse-impl.hxx
phi-gamma: Done.

gcc/rust/parse/rust-parse-impl.hxx
phi-gamma: I think I did initially but it came out uppercased ... in any case with the ``sstream`` it works fine now. Thanks.

gcc/rust/parse/rust-parse-impl.hxx
phi-gamma: ``rust_unreachable()``? Done.

P-E-P (approved):

P-E-P:
@phi-gamma Your last commit misses a sign off, I'll merge this PR once this get resolved.

phi-gamma:
2026-03-30 (Monday), ***@***.*** (P-E-P):

> P-E-P left a comment (Rust-GCC/gccrs#4463)
> 
> @phi-gamma Your last commit misses a sign off, I'll merge this PR once this get resolved.

Alright, there you go.

It was missing the SoB because I thought I’d squash it after review.

phi-gamma:
Thanks!

PR #4466, opened 2026-03-06
Add no_std to builtin attributes

P-E-P:
no_std is a builtin attribute required for rust-for-linux but was missing from the builtin attribute list and instead considered an unknown attribute macro invocation.


powerboat9:
We'd need something like #4462 as well -- probably as a separate PR

PR #4487, opened 2026-03-12
Use pretty printer instead ofcall to open_memstream

### LINKED ISSUE #4483: generate_tree_str uses non-standard open_memstream
### Summary

Forwarded from https://gcc.gnu.org/bugzilla/show_bug.cgi?id=124233 .

The code should really be (this is from me and not the original reporter):
```
static std::string
generate_tree_str (tree value)
{
  pretty_printer pp;
  dump_generic_node (&pp, value, 0, TDF_NONE, true);
  std::string result = pp_formatted_text (&pp);

  if (!result.empty () && result.back () == '\n')
    result.pop_back ();

  return result;
}
```

This is a regression so it would be good to include this. Plus this is faster since there is no going through libc. And it also much much simplier than before.

### Reproducer

I tried this code:

```rust
<code>
```

- [ ] Nightly

### Godbolt link

_No response_

### Actual behavior

The current behavior is...

### Expected behavior

I expected to see...

### GCC Version

r16-7965-g3fcb690f3ca5b7

P-E-P:
Remove open_memstream call. Code from @pinskia 

Fixes #4483

### DIFF
--- a/gcc/rust/typecheck/rust-tyty.cc
+++ b/gcc/rust/typecheck/rust-tyty.cc
@@ -3517,18 +3517,9 @@ ParamType::is_implicit_self_trait () const
 static std::string
 generate_tree_str (tree value)
 {
-  char *buf = nullptr;
-  size_t size = 0;
-
-  FILE *stream = open_memstream (&buf, &size);
-  if (!stream)
-    return "<error>";
-
-  print_generic_stmt (stream, value, TDF_NONE);
-  fclose (stream);
-
-  std::string result = (buf ? std::string (buf, size) : "<error>");
-  free (buf);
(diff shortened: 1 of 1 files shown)


powerboat9:
How should authorship/sign-off be handled?

pinskia:
> How should authorship/sign-off be handled?

Signed-off-by: Andrew Pinski <andrew.pinski@oss.qualcomm.com>

P-E-P:
> How should authorship/sign-off be handled?

Initially I put andrew as the author of the commit, I wasn't sure if andrew has any fsf attribution hence the missing sign off. With andrew's recent confirmation I added a sign-off.

PR #4489, opened 2026-03-13
Deduplicate some functions within PointerVisitor

P-E-P:
Some functions within the PointerVisitor were obvious duplicate from the inherited DefaultASTVisitor.


gcc/rust/ast/rust-ast-pointer-visitor.cc
```diff
@@ -154,47 +154,6 @@ PointerVisitor::visit (AST::QualifiedPathType &path)
...
-void
-PointerVisitor::visit (AST::AttrInputLiteral &attr_input)
-{
-  visit (attr_input.get_literal ());
-}
-
-void
-PointerVisitor::visit (AST::AttrInputMacro &attr_input)
-{
-  visit (attr_input.get_macro ());
```
powerboat9, line 182: I'm not sure we should remove this -- seems like we'll eventually want `AttrInputMacro` to be an ~`AttrInputExpr`

gcc/rust/ast/rust-ast-pointer-visitor.cc
```diff
@@ -154,47 +154,6 @@ PointerVisitor::visit (AST::QualifiedPathType &path)
...
-}
-
-void
-PointerVisitor::visit (AST::MetaItemLitExpr &meta_item)
-{
-  visit (meta_item.get_literal ());
```
powerboat9, line 188: same here

gcc/rust/ast/rust-ast-pointer-visitor.cc
```diff
@@ -1221,28 +880,6 @@ PointerVisitor::visit (AST::StructPatternFieldIdentPat &field)
...
-PointerVisitor::visit (AST::StructPatternFieldIdent &field)
-{
-  visit_outer_attrs (field);
-}
-
-void
-PointerVisitor::visit (AST::StructPatternElements &spe)
-{
-  for (auto &field : spe.get_struct_pattern_fields ())
-    visit (field);
```
powerboat9, line 1234: this should probably be `reseat`

gcc/rust/ast/rust-ast-pointer-visitor.cc
P-E-P: I'm working on this in another PR right now. I'll remove it for now and add it back when we properly handle AttrInputExpr. Right now those macros do not work anyway.

gcc/rust/ast/rust-ast-pointer-visitor.cc
P-E-P: Should be handled within StructPattern.*Pat

gcc/rust/ast/rust-ast-pointer-visitor.cc
P-E-P: @powerboat9 I'm unsure about this one. Why should I add it back ? `MetaItemLitExpr` contains a `LiteralExpr`, the default visitor will visit the `LiteralExpr` that'll get reseated. So we shouldn't require anything over here right ?

powerboat9: Yeah, in that case, nevermind

PR #4498, opened 2026-03-22
Fix glob import all ICE issue

### LINKED ISSUE #4475: internal compiler error: in resolve_path, at rust/resolve/rust-forever-stack.hxx
````rust
mod openssl {
    pub use ::*;
}

````

original:
````rust
#![feature(no_core)]
#![no_core]
//@ check-pass
// https://github.com/rust-lang/rust/pull/112743#issuecomment-1601986883

#![warn(ambiguous_glob_imports)]

macro_rules! m {
    () => {
      pub fn id() {}
    };
}

mod openssl {
    pub use ::*;
    //~^ WARNING ambiguous glob re-exports
    pub use self::main::*;

    mod evp {
      main!();
    }

    mod handwritten {
      m!();
    }
}

pub use self::evp::*;

fn main() {
    {
    pub use self::evp::*;
    //~^ WARNING ambiguous glob re-exports
    pub use self::handwritten::*;

    mod evp {
      m!();
    }

    mod handwritten {
      m!();
    }
};
    //~^ WARNING `id` is ambiguous
    // https://github.com/rust-lang/rust/pull/112743#issuecomment-1601986883
}

````

Version information:

1a9102843099b18f9dbb6b5e979436f6c4566f1c

Possibly related line of code:
https://github.com/Rust-GCC/gccrs/blob/1a9102843099b18f9dbb6b5e979436f6c4566f1c/gcc/rust/resolve/rust-forever-stack.hxx#L710-L722

Command:
`/home/matthias/vcs/github/gccrs/gccrs-build/gcc/crab1 -frust-incomplete-and-experimental-compiler-do-not-use`

<p>

```
crab1: error: failed to locate crate ‘core’
crab1: error: failed to locate crate ‘std’
crab1: error: unknown crate ‘core’
crab1: error: unknown crate ‘std’
crab1: internal compiler error: in resolve_path, at rust/resolve/rust-forever-stack.hxx:716
0x3097ef9 internal_error(char const*, ...)
	../../gcc/diagnostic-global-context.cc:787
0xc3aa61 fancy_abort(char const*, int, char const*)
	../../gcc/diagnostics/context.cc:1813
0xfaf4e3 tl::optional<Rust::Resolver2_0::Rib::Definition> Rust::Resolver2_0::ForeverStack<(Rust::Resolver2_0::Namespace)1>::resolve_path<Rust::AST::SimplePathSegment>(std::vector<Rust::AST::SimplePathSegment, std::allocator<Rust::AST::SimplePathSegment> > const&, Rust::Resolver2_0::ResolutionMode, std::function<void (Rust::AST::SimplePathSegment const&, unsigned int)>, std::vector<Rust::Error, std::allocator<Rust::Error> >&, std::reference_wrapper<Rust::Resolver2_0::ForeverStack<(Rust::Resolver2_0::Namespace)1>::Node>)
	../../gcc/rust/resolve/rust-forever-stack.hxx:716
0xfb1e84 tl::optional<Rust::Resolver2_0::Rib::Definition> Rust::Resolver2_0::ForeverStack<(Rust::Resolver2_0::Namespace)1>::resolve_path<Rust::AST::SimplePathSegment>(std::vector<Rust::AST::SimplePathSegment, std::allocator<Rust::AST::SimplePathSegment> > const&, Rust::Resolver2_0::ResolutionMode, std::function<void (Rust::AST::SimplePathSegment const&, unsigned int)>, std::vector<Rust::Error, std::allocator<Rust::Error> >&)
	../../gcc/rust/resolve/rust-forever-stack.hxx:704
0xfb1e84 tl::optional<Rust::Resolver2_0::Rib::Definition> Rust::Resolver2_0::NameResolutionContext::resolve_path<Rust::AST::SimplePathSegment>(std::vector<Rust::AST::SimplePathSegment, std::allocator<Rust::AST::SimplePathSegment> > const&, Rust::Resolver2_0::ResolutionMode, std::vector<Rust::Error, std::allocator<Rust::Error> >&, Rust::Resolver2_0::Namespace)
	../../gcc/rust/resolve/rust-name-resolution-context.h:575
0xfa76f5 tl::optional<Rust::Resolver2_0::Rib::Definition> Rust::Resolver2_0::NameResolutionContext::resolve_path<Rust::AST::SimplePathSegment>(std::vector<Rust::AST::SimplePathSegment, std::allocator<Rust::AST::SimplePathSegment> > const&, Rust::Resolver2_0::ResolutionMode, tl::optional<std::vector<Rust::Error, std::allocator<Rust::Error> >&>, Rust::Resolver2_0::Namespace)
	../../gcc/rust/resolve/rust-name-resolution-context.h:633
0xfa76f5 tl::optional<Rust::Resolver2_0::Rib::Definition> Rust::Resolver2_0::NameResolutionContext::resolve_path<Rust::AST::SimplePathSegment>(std::vector<Rust::AST::SimplePathSegment, std::allocator<Rust::AST::SimplePathSegment> > const&, bool, tl::optional<std::vector<Rust::Error, std::allocator<Rust::Error> >&>, Rust::Resolver2_0::Namespace)
	../../gcc/rust/resolve/rust-name-resolution-context.h:671
0xfa76f5 tl::optional<Rust::Resolver2_0::Rib::Definition> Rust::Resolver2_0::NameResolutionContext::resolve_path<Rust::AST::SimplePathSegment>(std::vector<Rust::AST::SimplePathSegment, std::allocator<Rust::AST::SimplePathSegment> > const&, bool, Rust::Resolver2_0::Namespace)
	../../gcc/rust/resolve/rust-name-resolution-context.h:681
0xfa76f5 tl::optional<Rust::Resolver2_0::Rib::Definition> Rust::Resolver2_0::NameResolutionContext::resolve_path<Rust::Resolver2_0::Namespace>(Rust::AST::SimplePath const&, Rust::Resolver2_0::Namespace&&)
	../../gcc/rust/resolve/rust-name-resolution-context.h:699
0xfa76f5 Rust::Resolver2_0::Early::resolve_glob_import(unsigned int, Rust::Resolver2_0::TopLevel::ImportKind&&)
	../../gcc/rust/resolve/rust-early-name-resolver-2.0.cc:78
0xfa99fb Rust::Resolver2_0::Early::build_import_mapping(std::pair<unsigned int, std::vector<Rust::Resolver2_0::TopLevel::ImportKind, std::allocator<Rust::Resolver2_0::TopLevel::ImportKind> > >&&)
	../../gcc/rust/resolve/rust-early-name-resolver-2.0.cc:167
0xfa9ec5 Rust::Resolver2_0::Early::visit(Rust::AST::UseDeclaration&)
	../../gcc/rust/resolve/rust-early-name-resolver-2.0.cc:499
0xe300bf void Rust::AST::DefaultASTVisitor::visit<Rust::AST::Item>(std::unique_ptr<Rust::AST::Item, std::default_delete<Rust::AST::Item> >&)
	../../gcc/rust/ast/rust-ast-visitor.h:423
0xe300bf Rust::AST::DefaultASTVisitor::visit(Rust::AST::Module&)
	../../gcc/rust/ast/rust-ast-visitor.cc:849
0xf93189 operator()
	../../gcc/rust/resolve/rust-default-resolver.cc:61
0xf93189 scope_inner<Rust::Resolver2_0::CanonicalPathCtx::scope<Rust::Resolver2_0::DefaultResolver::visit(Rust::AST::Module&)::<lambda()> >(Rust::NodeId, std::string, Rust::Resolver2_0::DefaultResolver::visit(Rust::AST::Module&)::<lambda()>&&)::<lambda()>, Rust::Resolver2_0::DefaultResolver::visit(Rust::AST::Module&)::<lambda()> >
	../../gcc/rust/resolve/rust-name-resolution-context.h:470
0xf93189 scope<Rust::Resolver2_0::DefaultResolver::visit(Rust::AST::Module&)::<lambda()> >
	../../gcc/rust/resolve/rust-name-resolution-context.h:413
0xf93189 scope<Rust::Resolver2_0::DefaultResolver::visit(Rust::AST::Module&)::<lambda()> >
	../../gcc/rust/resolve/rust-name-resolution-context.h:406
0xf93189 operator()
	../../gcc/rust/resolve/rust-default-resolver.cc:64
0xf93189 __invoke_impl<void, Rust::Resolver2_0::DefaultResolver::visit(Rust::AST::Module&)::<lambda()>&>
	/usr/include/c++/15.2.1/bits/invoke.h:63
/home/matthias/vcs/github/gccrs/gccrs-build/gcc/crab1 /tmp/icemaker_global_tempdir.6L6WNx0nPA0K/rustc_testrunner_tmpdir_reporting.SrsQgFOHOlh9/mvce.rs -o/tmp/icemaker_global_tempdir.6L6WNx0nPA0K/rustc_testrunner_tmpdir_reporting.SrsQgFOHOlh9/binaryname -frust-incomplete-and-experimental-compiler-do-not-use

```

</p>
</details>

### PR DESCRIPTION
**Pull request by AhmedSaid3617**
Fixes #4475 
This attempts to fix the ICE that happens when you import modules like this:
```
pub use ::*;
``` 
I have noticed that rustc handles this with an error message tailored to this specific case, see [godbolt](https://godbolt.org/z/nj7z7nPTK). I tried to do the same.
This is my first PR, I'm trying to get feedback on my approach. Also, I know I need to fix the format, commit message and add test cases. Please tell me if I missed something.

### DIFF
--- a/gcc/rust/ast/rust-item.h
+++ b/gcc/rust/ast/rust-item.h
@@ -1074,7 +1074,7 @@ public:
 
   void accept_vis (ASTVisitor &vis) override;
 
-  PathType get_glob_type () { return glob_type; }
+  PathType get_glob_type () const { return glob_type; }
 
   Kind get_kind () const override { return Glob; }
 
--- a/gcc/rust/resolve/rust-forever-stack.hxx
+++ b/gcc/rust/resolve/rust-forever-stack.hxx
@@ -729,21 +729,24 @@ ForeverStack<N>::resolve_path (
     }
   else
     {
-      rust_assert (!path.get_segments ().empty ());
+      switch (mode)
+	{
+	case ResolutionMode::Normal:
+	  break; // default
+	case ResolutionMode::FromRoot:
+	  starting_point = root;
+	  break;
+	case ResolutionMode::FromExtern:
+	  starting_point = extern_prelude;
+	  break;
+	default:
+	  rust_unreachable ();
+	}
     }
 
-  switch (mode)
+  if (path.get_segments ().empty ())
     {
-    case ResolutionMode::Normal:
-      break; // default
-    case ResolutionMode::FromRoot:
-      starting_point = root;
-      break;
-    case ResolutionMode::FromExtern:
-      starting_point = extern_prelude;
-      break;
-    default:
-      rust_unreachable ();
+      return Rib::Definition::NonShadowable (starting_point.get ().id);
     }
 
   auto &segments = path.get_segments ();
--- a/gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
@@ -502,7 +502,9 @@ flatten_glob (const AST::UseTreeGlob &glob, std::vector<AST::SimplePath> &paths,
   if (glob.has_path ())
     paths.emplace_back (glob.get_path ());
   else
-    paths.emplace_back (AST::SimplePath ({}, false, glob.get_locus ()));
+    paths.emplace_back (AST::SimplePath (
+      {}, glob.get_glob_type () == AST::UseTreeGlob::PathType::GLOBAL,
+      glob.get_locus ()));
 }
 
 static bool
--- a/gcc/testsuite/rust/compile/glob-import-all.rs
+++ b/gcc/testsuite/rust/compile/glob-import-all.rs
@@ -1,0 +1,9 @@
+// { dg-additional-options "-frust-edition=2015" }
+
+#![feature(no_core)]
+#![no_core]
+
+mod a {
+    #[allow(unused)]
+    use ::*;
+}

### CODE REVIEW

**Comment by powerboat9** [comment]
Yeah, it looks like rustc manages to resolve the path (with edition-specific behavior even, see https://godbolt.org/z/ox76Woqob). A better solution would probably be to remove the assert in `resolve_path`, then stick an early return (with a value based on `starting_point`)
```rust
  switch (mode)
    {
    case ResolutionMode::Normal:
      break; // default
    case ResolutionMode::FromRoot:
      starting_point = root;
      break;
    case ResolutionMode::FromExtern:
      starting_point = extern_prelude;
      break;
    default:
      rust_unreachable ();
    }

    <HERE>

    auto &segments = path.get_segments ();
```

to resolve the ICE, and we can tweak #4475 from an ICE to describe whatever is still broken with `use ::*` handling

powerboat9:
> This is my first PR

Welcome

powerboat9:
Oh, you should probably also move the switch statement into the else statement, where the assertion currently is. It not being in that else statement is probably a bug.

AhmedSaid3617:
Couple of questions
1- You're saying I should only handle the ICE in this PR and get gccrs to not crash, then leave outputting the error to another PR? If so then what should this function return in case of a failed resolution like this?
2- I checked this switch case. I was expecting the mode to evaluate to ```ResolutionMode::FromRoot``` or ```ResolutionMode::FromExtern```, but it never does. If I understand correctly, this is the resolution mode we're dealing with here ```use ::*;```
I looked through the code and it seems this is what this code is trying to achieve:

gcc/rust/resolve/rust-name-resolution-context.h:591
```rust
  class ResolutionBuilder
  {
  public:
    ResolutionBuilder (NameResolutionContext &ctx) : ctx (&ctx) {}

    template <typename S>
    void set_path (const std::vector<S> &path_segments, NodeId node_id,
		   bool has_opening_scope)
    {
      path = ResolutionPath (path_segments, node_id);
      mode = ResolutionMode::Normal;
      if (has_opening_scope)
	{
	  if (get_rust_edition () == Edition::E2015)
	    mode = ResolutionMode::FromRoot;
	  else
	    mode = ResolutionMode::FromExtern;
	}
      has_path_set = true;
    }
 ```
 but either it doesn't work, or I don't understand what `has_opening_scope` means, I'm a bit of a novice at rust :) .
 
 If resolution mode is in fact broken, then I can't depend on starting_point to determine if this error has happened. In which case, should I try and fix it?

**Comment by powerboat9** [comment]
I think the problem is that `::*` isn't being interpreted as having an opening scope operator (ex: `::a::b::c` has it, `a::b::c` doesn't). You'd have to make sure that `::*` produces a path (see: `glob.to_resolve`) with `has_opening_scope` set to true.

`rustc` seems to treat `::*` as a valid glob import -- it's just that glob imports can't point to the list of all crates or import the contents of a module into itself, which are only tangentially related issues.

**Comment by powerboat9** [comment]
see: https://godbolt.org/z/jraohWToT, which compiles on edition 2015 but not later editions

**Comment by powerboat9** [comment]
You might have to take a look at the parsing code

**Comment by AhmedSaid3617** [comment]
> I think the problem is that ::* isn't being interpreted as having an opening scope operator

Yep, I can confirm that now. I'll take a look at parsing and try to fix it.
Regarding the outcome of this PR, should I try to display errors like rustc does, with different error based on rust edition/startpoint?

**Comment by powerboat9** [comment]
I'd think that getting https://godbolt.org/z/jraohWToT to pass when the edition is set to 2015 would be good enough for this PR, both to generally break changes up into smaller PRs and since this is your first PR.

**Comment by AhmedSaid3617** [comment]
1 - I found a bug in the parser code and fixed it, to detect the opening scope resolution, but the top level resolver ruins my fix when calling glob_flatten. 
2 - I'm not sure exactly the best way to send an error message from resolve_path, and I can't find a locus to point the error to because the path has no segments.
Either I fix these problems or handle this case in the parser.
Should I modify top level resolver to fix the first problem?
I included the details for my questions in comments, which I will clean up later.

File: gcc/rust/ast/rust-item.h
Comment (powerboat9) [inline gcc/rust/ast/rust-item.h:1049]: Actually, it looks like you can use `GLOBAL` to detect a `::*` use tree. See `Parser<ManagedTokenSource>::parse_use_tree`

File: gcc/rust/resolve/rust-forever-stack.hxx [hunk 1]
```diff
@@ -729,21 +729,30 @@ ForeverStack<N>::resolve_path (
...
+      // I can't use resolution mode to output a specific error, because of the
+      // problem in flatten_glob so any path with zero segments is considered a
+      // global import from root.
+      if (path.get_segments ().empty ())
+	{
+	  // I can't find a locus to pass to this, since there are no segments.
+	  rust_error_at ((location_t) 0, ErrorCode::E0433,
+			 "cannot import a module onto itself");
+	  return tl::nullopt;
+	}
```
powerboat9: You don't need this

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
```diff
@@ -502,6 +502,9 @@ flatten_glob (const AST::UseTreeGlob &glob, std::vector<AST::SimplePath> &paths,
   if (glob.has_path ())
     paths.emplace_back (glob.get_path ());
   else
+    // This resets the has_opening_scope to false, so we can't differentiate between use * and use ::*.
+    // I can't glob.get_path().has_opening_scope_resolution() because get_path() assers has_path().
+    // How should i fix this?
     paths.emplace_back (AST::SimplePath ({}, false, glob.get_locus ()));
```
powerboat9: You can use `glob.get_glob_type () == AST::UseTreeGlob::PathType::GLOBAL`

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -743,15 +743,23 @@ ForeverStack<N>::resolve_path (
...
 	{
-	  // I can't find a locus to pass to this, since there are no segments.
-	  rust_error_at ((location_t) 0, ErrorCode::E0433,
-			 "cannot import a module onto itself");
-	  return tl::nullopt;
+	  if (starting_point.get().id == extern_prelude.id)
+	    {
+              
+	      // I can't find a locus to pass to this, since there are no segments.
+	      rust_error_at ((location_t) 0, ErrorCode::E0433,
```
powerboat9: Use `UNKNOWN_LOCATION`

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -743,15 +743,23 @@ ForeverStack<N>::resolve_path (
...
+			     "cannot glob-import all possible crates");
+	      return tl::nullopt;
+	    }
+          else // includes 'use *;'
+            {
+              // I can't find a locus to pass to this, since there are no segments.
+	      rust_error_at ((location_t) 0, ErrorCode::E0433,
```
powerboat9: Same

powerboat9:
Yeah, if you just use `UNKNOWN_LOCATION` and add https://godbolt.org/z/jraohWToT as a test (with edition 2015, no errors emitted), it'll be good.

AhmedSaid3617:
I also added a check in Early::build_import_mapping to stop it from checking the error that I already handled and segfaulting.

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -747,16 +747,13 @@ ForeverStack<N>::resolve_path (
...
-	      rust_error_at ((location_t) 0, ErrorCode::E0433,
+	      rust_error_at (UNKNOWN_LOCATION, ErrorCode::E0433,
 			     "cannot glob-import all possible crates");
 	      return tl::nullopt;
 	    }
           else // includes 'use *;'
             {
-              // I can't find a locus to pass to this, since there are no segments.
-	      rust_error_at ((location_t) 0, ErrorCode::E0433,
+	      rust_error_at (UNKNOWN_LOCATION, ErrorCode::E0433,
```
powerboat9: Oh, you should probably append to `collect_errors`

gcc/rust/resolve/rust-early-name-resolver-2.0.cc
```diff
@@ -174,7 +174,7 @@ Early::build_import_mapping (
 	  break;
 	}
 
-      if (!found && old_error_count == macro_resolve_errors.size ())
+      if (!found && old_error_count == macro_resolve_errors.size () && !path.get_segments().empty())
```
powerboat9: ...which would make this change unnecessary

AhmedSaid3617: Tried appending to collect_errors, the errors are emitted by ```ResolutionBuilder:resolve()``` gcc/rust/resolve/rust-name-resolution-context.h:674.
```Early::build_import_mapping``` still segfaults.

File: gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc (hunk 2 above)
Comment (powerboat9) [reply gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc]: Needs to use `glob.get_glob_type ()`, to get the type of the glob use tree (global), not the type of the use tree (glob)

File: gcc/rust/resolve/rust-forever-stack.hxx [hunk 7]
```diff
@@ -729,21 +729,33 @@ ForeverStack<N>::resolve_path (
...
+	  if (starting_point.get ().id == extern_prelude.id)
+	    collect_errors.emplace_back (
+	      UNKNOWN_LOCATION, ErrorCode::E0433,
+	      "cannot glob-import all possible crates");
+
+	  else // includes 'use *;'
+	    collect_errors.emplace_back (UNKNOWN_LOCATION, ErrorCode::E0433,
+					 "cannot import a module onto itself");
+
+	  return tl::nullopt;
```
powerboat9: This is subtly wrong -- can you replace this with
```rust
return Rib::Definition::NonShadowable (starting_point.get ().id);
```
and leave the error messages for a separate PR?

powerboat9:
Could you add https://godbolt.org/z/jraohWToT as a test?

powerboat9 (changes requested):

powerboat9:
Oh, and remember to write out a changelog in your commit message, although you don't have to do that until your PR is otherwise ready to merge (adjusting it is kind of a pain)

AhmedSaid3617:
I still have to check for non-empty segments vector at ```Early::build_import_mapping```, or else it segfaults even if I append to ```collect_errors```.
I'll add a test case in a separate commit if everything looks good here.
Should I create a separate issue for handling the errors?

powerboat9:
> I still have to check for non-empty segments vector at `Early::build_import_mapping`, or else it segfaults even if I append to `collect_errors`.

Huh. https://godbolt.org/z/jraohWToT shouldn't produce any errors, so it really shouldn't segfault while trying to print one.

> Should I create a separate issue for handling the errors?

Yeah

gcc/rust/parse/rust-parse-impl.hxx
```diff
@@ -1357,9 +1355,12 @@ Parser<ManagedTokenSource>::parse_use_tree ()
...
 	  if (is_global)
-	    return std::unique_ptr<AST::UseTreeGlob> (
-	      new AST::UseTreeGlob (AST::UseTreeGlob::GLOBAL,
-				    AST::SimplePath::create_empty (), locus));
+	    return std::unique_ptr<AST::UseTreeGlob> (new AST::UseTreeGlob (
+	      AST::UseTreeGlob::GLOBAL,
+	      AST::SimplePath (std::vector<AST::SimplePathSegment> (), true,
+			       0UL),
+	      locus)); // global import with opening scope resolution and empty
+		       // path
```
powerboat9: This shouldn't be necessary anymore, because of the `get_glob_type` call

gcc/rust/resolve/rust-forever-stack.hxx
```diff
@@ -729,21 +729,24 @@ ForeverStack<N>::resolve_path (
...
-      break;
-    case ResolutionMode::FromExtern:
-      starting_point = extern_prelude;
-      break;
-    default:
-      rust_unreachable ();
+      if (path.get_segments ().empty ())
+	{
+	  return Rib::Definition::NonShadowable (starting_point.get ().id);
+	}
```
powerboat9: You should probably move this to just below the else statement

AhmedSaid3617:
I added a test case in rust/compile that doesn't cause any errors, once we add error handling I'll modify the test so it checks for the correct errors.
I'll push once the first commit is all clear, to make things easier git-wise.

gcc/rust/resolve/rust-early-name-resolver-2.0.cc
```diff
@@ -178,7 +174,8 @@ Early::build_import_mapping (
 	  break;
 	}
 
-      if (!found && old_error_count == macro_resolve_errors.size ())
+      if (!found && old_error_count == macro_resolve_errors.size ()
+	  && !path.get_segments ().empty ())
```
powerboat9: Is this still required? Trying to resolve `::` shouldn't ever _fail_, just possibly resolve to something that can't currently be glob imported (the current module or the extern crate prelude)

AhmedSaid3617: Since we're not throwing errors it's not required. I'll remove it from this PR, and I'll look into it when I create the new issue.
I think the problem is that the next line tries to get locus for the segment that failed to resolve, and in our case there are no segments.

AhmedSaid3617:
BTW, I think this part in ```gcc/rust/ast/rust-item.h:1040``` has bad naming? Global here doesn't mean *, it means ::, on the other hand UseTreeGlob means *, which confused me at first, and may be the root cause why this case was missed.
Should I open an issue to refactor this? Or is it too much of a hassle?
```rust
// Use tree with a glob (wildcard) operator
class UseTreeGlob : public UseTree
{
public:
  enum PathType
  {
    NO_PATH,
    GLOBAL,
    PATH_PREFIXED
  };
```

powerboat9:
> BTW, I think this part in `gcc/rust/ast/rust-item.h:1040` has bad naming? Global here doesn't mean *, it means ::, on the other hand UseTreeGlob means *, which confused me at first, and may be the root cause why this case was missed. Should I open an issue to refactor this? Or is it too much of a hassle?
> 
> ```rust
> // Use tree with a glob (wildcard) operator
> class UseTreeGlob : public UseTree
> {
> public:
>   enum PathType
>   {
>     NO_PATH,
>     GLOBAL,
>     PATH_PREFIXED
>   };
> ```

Yeah, the thing that made it click for me was `Parser<ManagedTokenSource>::parse_use_tree` (`parse/rust-parse-impl.hxx:1310`). You're welcome to make a PR to add some more comments -- it is a bit hard to figure out. Renaming all the enumerator usages probably would be too much of a hassle, comparatively.

powerboat9 (approved):

PR #4499, opened 2026-03-23
Improve unit struct matching

powerboat9:
Unit structs weren't *actually* working in match statements -- we don't currently handle identifier pattern vs path pattern disambiguation.


CohenArthur (approved):
thanks!

PR #4508, opened 2026-03-30
Replace usages of `chainon`

powerboat9:
Building a list using `chainon` is technically `O(n^2)`, so this PR replaces it with a simple linked list abstraction. The abstraction isn't *required*, but it is convenient and should be ~zero cost.

### DIFF
--- a/gcc/rust/backend/rust-compile-asm.cc
+++ b/gcc/rust/backend/rust-compile-asm.cc
@@ -1,16 +1,16 @@
 #include "rust-compile-asm.h"
 #include "rust-compile-expr.h"
 #include "rust-system.h"
+#include "rust-ggc.h"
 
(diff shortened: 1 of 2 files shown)


P-E-P (approved):
There is few chainon in GCCRS and I doubt their chain will grow big. Those are mostly around inline assembly arguments, and even I doubt we really need to chase performance over there, there is much lower hanging fruits. I'm not even sure we should look at performance right now, the compiler may change completely and those speed improvements would be made completely useless in no time.

I've still taken a look at the code and everything looks sane, I would be inclined to accept it if you think it's useful.

powerboat9:
Yeah, I just figured it was an ~easy theoretical performance gain, so I may as well do it

PR #4467, opened 2026-03-07
Intrinsic: Add `ctlz` and `ctlz_nonzero` intrinsics

moabo3li:
This commit introduces the `ctlz` and `ctlz_nonzero` intrinsics for counting leading zeros in integer types.
The implementation includes handlers for both intrinsics, which validate input types and utilize GCC built-ins for the actual computation.

Addresses: #658

gcc/rust/ChangeLog:

	* backend/rust-compile-intrinsic.cc: Register ctlz and ctlz_nonzero intrinsics.
	* backend/rust-intrinsic-handlers.cc (ctlz_handler): Implement ctlz_handler.
	(ctlz_nonzero_handler): Implement ctlz_nonzero_handler.
	* backend/rust-intrinsic-handlers.h (ctlz_handler): Add decl for ctlz_handler.
	(ctlz_nonzero_handler): Add decl for ctlz_nonzero_handler.
	* util/rust-intrinsic-values.h: map ctlz_nonzero to its name.

gcc/testsuite/ChangeLog:

	* rust/compile/ctlz.rs: New test.
	* rust/compile/ctlz_nonzero.rs: New test.
	* rust/execute/torture/ctlz.rs: New test.
	* rust/execute/torture/ctlz_nonzero.rs: New test.

PR #4511, opened 2026-04-01
Add cttz and cttz_nonzero support

moabo3li:
Introduces the `cttz` and `cttz_nonzero` intrinsics for counting leading zeros in integer types.

The implementation includes handlers for both intrinsics, which validate input types and utilize

Addresses: Rust-GCC#658


moabo3li:
@powerboat9 cttz logic is similar to #4467

PR #4513, opened 2026-04-03
Add assert macro handler

P-E-P:
Assert macro handler was missing, insert a basic handler that desugars to a condition and a call to panic.

### DIFF
--- a/gcc/rust/expand/rust-macro-builtins-log-debug.cc
+++ b/gcc/rust/expand/rust-macro-builtins-log-debug.cc
@@ -19,6 +19,9 @@
 #include "rust-ast-fragment.h"
 #include "rust-macro-builtins.h"
 #include "rust-macro-builtins-helpers.h"
+#include "rust-ast-builder.h"
+#include "optional.h"
+#include "rust-ast-collector.h"
 
 namespace Rust {
 tl::optional<AST::Fragment>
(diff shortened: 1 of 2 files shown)


gcc/rust/expand/rust-macro-builtins-log-debug.cc
```diff
@@ -27,6 +30,79 @@ MacroBuiltin::assert_handler (location_t invoc_locus,
...
+    }
+
+  auto pending_invocations = check_for_eager_invocations (expanded_expr);
+  if (!pending_invocations.empty ())
+    return make_eager_builtin_invocation (BuiltinMacro::Assert, invoc_locus,
+					  invoc.get_delim_tok_tree (),
+					  std::move (pending_invocations));
+
+  if (expr_to_assert->get_expr_kind () != AST::Expr::Kind::MacroInvocation)
+    {
```
CohenArthur: can we even reach that `if` if we have checked for pending eager invocations above and returned?

CohenArthur (approved):
LGTM :)

gcc/rust/expand/rust-macro-builtins-log-debug.cc
P-E-P: Fixed.

Issue #4476, opened 2026-03-08
internal compiler error: in throw_exception, at rust/util/expected.h


matthiaskrgr:
auto-reduced (treereduce-rust):
````rust
//@compile-flags: -frust-incomplete-and-experimental-compiler-do-not-use

fn main() {
     Foo { #[cfg(feature = -1)] .. } = S {};
    
}

````

original:
````rust
#![feature(no_core)]
#![no_core]
// #81282: Attributes are not allowed on struct field rest patterns (the ..).

struct S {}

fn main() {
    if 0 {} else if 0 {#![attr]} Foo { #[cfg(feature = -1)] .. } = S {};
    //~^ ERROR expected identifier, found `..`
}

````

Version information:

1a9102843099b18f9dbb6b5e979436f6c4566f1c

Possibly related line of code:
https://github.com/Rust-GCC/gccrs/blob/1a9102843099b18f9dbb6b5e979436f6c4566f1c/gcc/rust/util/expected.h#L209-L221

Command:
`/home/matthias/vcs/github/gccrs/gccrs-build/gcc/crab1 -frust-incomplete-and-experimental-compiler-do-not-use`

<details><summary><strong>Program output</strong></summary>
<p>

```
crab1: internal compiler error: in throw_exception, at rust/util/expected.h:215
0x3097ef9 internal_error(char const*, ...)
	../../gcc/diagnostic-global-context.cc:787
0xc3aa61 fancy_abort(char const*, int, char const*)
	../../gcc/diagnostics/context.cc:1813
0xd66899 void tl::detail::throw_exception<tl::bad_expected_access<Rust::Parse::Error::StructExprField> >(tl::bad_expected_access<Rust::Parse::Error::StructExprField>&&)
	../../gcc/rust/util/expected.h:215
0xd66899 std::unique_ptr<Rust::AST::StructExprField, std::default_delete<Rust::AST::StructExprField> >& tl::expected<std::unique_ptr<Rust::AST::StructExprField, std::default_delete<Rust::AST::StructExprField> >, Rust::Parse::Error::StructExprField>::value<std::unique_ptr<Rust::AST::StructExprField, std::default_delete<Rust::AST::StructExprField> >, (void*)0>() &
	../../gcc/rust/util/expected.h:1989
[...]

PR #4514, opened 2026-04-03
rust: avoid ICE on attributed struct base parsing

hriztam:
Address #4476

rust: avoid ICE on attributed struct base parsing

Parsing a struct expression like `Foo { #[cfg(feature = -1)] .. }`
could ICE because `parse_struct_expr_struct_partial()` unconditionally
called `value()` on an errored `StructExprField` result after
`parse_struct_expr_field()` had already identified `..` as the struct
base.

Handle that case explicitly, emit a normal parser diagnostic when
attributes are attached to the struct-base `..`, and add a regression
test for the reduced reproducer.

Signed-off-by: Hritam Shrivastava <hritamstark05@gmail.com>

### DIFF
--- a/gcc/testsuite/rust/compile/issue-4476.rs
+++ b/gcc/testsuite/rust/compile/issue-4476.rs
@@ -1,0 +1,10 @@
+#![feature(no_core)]
+#![no_core]
+
+struct Foo {}
+struct S {}
+
+fn main() {
+    Foo { #[cfg(feature = -1)] .. } = S {};
+    // { dg-error "attributes are not allowed before .* in a struct expression" "" { target *-*-* } .-1 }
+}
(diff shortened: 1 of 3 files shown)


gcc/testsuite/rust/compile/issue-4476.rs
```diff
@@ -0,0 +1,12 @@
+// { dg-options "-frust-incomplete-and-experimental-compiler-do-not-use" }
```
P-E-P: This is not required, it is already injected thanks to `compile.exp`

P-E-P (approved):

gcc/testsuite/rust/compile/issue-4476.rs
hriztam: I updated the PR

philberty (approved):
LGTM

PR #4515, opened 2026-04-04
[dropck] Emit Drop Error Codes

Islam-Imad:
emit : 
- [[0740]](https://doc.rust-lang.org/error_codes/E0740.html) union fields must not need drop.
- [[0184]](https://doc.rust-lang.org/error_codes/E0184.html) no type cannot have copy & drop trait.
- [[0040]](https://doc.rust-lang.org/error_codes/E0040.html) prevent direct drop call.

### DIFF
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -194,6 +194,7 @@ GRS_OBJS = \
     rust/rust-hir-generic-param.o \
     rust/rust-type-util.o \
     rust/rust-coercion.o \
+    rust/rust-drop-check.o \
     rust/rust-casts.o \
     rust/rust-unify.o \
     rust/rust-hir-type-check-base.o \
--- a/gcc/rust/Make-lang.in
+++ b/gcc/rust/Make-lang.in
@@ -195,6 +195,7 @@ GRS_OBJS = \
     rust/rust-hir-generic-param.o \
     rust/rust-type-util.o \
     rust/rust-coercion.o \
+    rust/rust-drop-check.o \
     rust/rust-casts.o \
     rust/rust-unify.o \
     rust/rust-hir-type-check-base.o \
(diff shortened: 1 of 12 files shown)


dkm:
Hi!

while having a look at the changelog, I think it may need some adjustment. For example:
```
	* typecheck/rust-hir-type-check.h: peek_trait_impl_map
	* typecheck/rust-typecheck-context.cc
	(TypeCheckContext::insert_associated_impl_mapping): peek_trait_impl_map
```
doesn't really help in knowing what happened in these files. The names are off, I think.
You're adding `peek_associated_impl_mapping_for_self` in the rust-hir-type-check.h file, and I'm not sure why you mention `insert_associated_impl_mapping`.

Can you respin the changelog part? I looks like you probably did it during an earlier phase of your dev.
Thanks!

dkm (changes requested):
Changelog needs love ❤️

Islam-Imad:
didn't forget you. a little bit busy nowadays
this work needs non-trivial refactor :)
i will update it and ping you again

CohenArthur:
@Islam-Imad I will put the PR in Draft mode for now, and you can mark it as ready once you are ready for a review again :)

Islam-Imad:
> @Islam-Imad I will put the PR in Draft mode for now, and you can mark it as ready once you are ready for a review again :)

the PR is ready for review. 
i have rebased and updated it.

CohenArthur:
One nit @Islam-Imad - in your Changelog you wrote "Make-lang.in: add rust-drop-check.h" when it should be "rust-drop-check.o" or "rust-drop-check object" or "rust-drop-check.cc". It's not the header file that we want to add to the list of compilation object but the .cc file or its resulting object rather.

CohenArthur (commented):
I think the code is good - can you check that this is still updated and in-line with the work that was done recently on the Drop infrastructure? I'd love to get these changes in :) thanks @Islam-Imad !!!

Islam-Imad:
> I think the code is good - can you check that this is still updated and in-line with the work that was done recently on the Drop infrastructure? I'd love to get these changes in :) thanks @Islam-Imad !!!

yes i will update it :)

PR #4516, opened 2026-04-06
Defer number literal evaluation to parser

nsvke:
This PR defers number literal evaluation (underscore stripping, base conversion) and suffix validation from the Lexer to the Parser phase.

Previously, the Lexer eagerly stripped underscores and validated suffixes. This violated source fidelity for procedural macros, causing them to receive mutated strings or fail prematurely on invalid suffixes.

gcc/rust/ChangeLog:

	* ast/rust-ast-collector.cc (TokenCollector::visit): Update Token::make_int and Token::make_float calls to include suffix_start and LITERALBASE_DECIMAL.
	* expand/rust-macro-builtins-location.cc (MacroBuiltin::column_handler): Pass string length and base to Token::make_int. (MacroBuiltin::line_handler): Likewise.
	* lex/rust-lex.cc (Lexer::parse_in_type_suffix): Rename to parse_in_suffix and return string instead of PrimitiveCoreType. (Lexer::parse_in_suffix): Remove underscore stripping to preserve source fidelity for macros. (Lexer::parse_in_exponent_part): Preserve '+' and '-' characters in the raw string. (Lexer::parse_in_decimal): Remove underscore stripping. (Lexer::parse_non_decimal_int_literal): Track suffix start index and pass literal base. (Lexer::parse_non_decimal_int_literals): Use IntegerLiteralBase enum values instead of raw integers. (Lexer::parse_decimal_int_or_float): Track suffix string length and pass base parameters to token creation.
	* lex/rust-lex.h: Update method signatures for suffix parsing.
	* lex/rust-token.h (enum IntegerLiteralBase): New enum to represent numeric bases.
	* parse/rust-parse-impl-expr.hxx: use LiteralResolve functions to evaluate raw token strings.
	* parse/rust-parse-impl-pattern.hxx: Use evaluated literal strings for INT and FLOAT tokens.
	* parse/rust-parse.cc (resolve_literal_suffix): Move suffix validation logic from lexer to parser. (evaluate_integer_literal): New function to strip underscores and convert to decimal via GMP. (evaluate_float_literal): New function to strip underscores from floats.
	* parse/rust-parse.h (evaluate_integer_literal): Declare in LiteralResolve namespace. (evaluate_float_literal): Likewise. (resolve_literal_suffix): Likewise.
	* util/rust-token-converter.cc (from_literal): Safely reconstruct raw text and suffix to dynamically determine base and suffix_start for ProcMacros.

gcc/testsuite/ChangeLog:

	* rust/compile/deferred-suffix-validation.rs: New test.
	* rust/compile/evaluate-integer-or-float.rs: New test.
	* rust/compile/tuple-index.rs: New test.

### DIFF
--- a/gcc/rust/lex/rust-token.h
+++ b/gcc/rust/lex/rust-token.h
@@ -218,6 +218,17 @@ enum TokenId
 #undef RS_TOKEN
 };
 
+// Holds the base information for integer-literal tokens. For other tokens, its
+// value is 0.
+enum class IntegerLiteralBase : uint8_t
+{
+  None = 0,
+  Binary,
+  Octal,
+  Decimal,
+  Hex,
+};
+
 // dodgy "TokenPtr" declaration with Token forward declaration
 class Token;
 // A smart pointer (shared_ptr) to Token.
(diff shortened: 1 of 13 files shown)


powerboat9 (approved):

gcc/rust/lex/rust-token.h
```diff
@@ -218,6 +218,17 @@ enum TokenId
 #undef RS_TOKEN
 };
 
+// Holds the base information for integer-literal tokens. For other tokens, its
+// value is 0.
+enum IntegerLiteralBase : uint8_t
```
CohenArthur: this is more of a nitpick so feel free to ignore. is there any reason for making the enum representation an `uint8_t`? also, could we maybe turn it into an `enum class` to have stronger namespacing? this could make it easier to use in my opinion but that's personal:

```c++
enum class IntegerLiteralBase : uint8_t {
    None = 0,
    Binary,
    Octal,
    Decimal,
    Hex,
}
```

CohenArthur (approved):
I think this is a really good fix and the implementation looks good! well done @nsvke!

gcc/rust/lex/rust-token.h
nsvke: You're right, using an `enum class` is better. I'll update it soon. 

I defined it as `uint8_t` so it can fit into the existing padding of the `Token` class without increasing its overall size.

gcc/testsuite/rust/compile/tuple-index.rs
```diff
@@ -0,0 +1,13 @@
+// { dg-do compile }
```
P-E-P: Are those DejaGNU instructions really required ? IIRC do-compile is implied by compile.exp ?

nsvke: My mistake. I've removed it since it's redundant.

P-E-P (approved):

PR #4525, opened 2026-04-15
Resolve non-final path segments in type namespace

powerboat9:
This is very hack-y, and the `ForeverStack`/`NameResolutionContext` boundary is due for some major refactoring, but should be correct. Credit to @CohenArthur for mentioning the idea in DMs
> so I did a quick and dirty implementation where ForeverStack<N>::resolve_path now takes an extra parameter, a ForeverStack<Namespace::Types> &types_stack, and does resolve_segments within the types NS before resolving the last segment.

### DIFF
--- a/gcc/rust/resolve/rust-forever-stack.h
+++ b/gcc/rust/resolve/rust-forever-stack.h
@@ -622,6 +622,8 @@ private:
 
 template <Namespace N> class ForeverStack
 {
+  template <Namespace N2> friend class ForeverStack;
+
 public:
   ForeverStack ()
     : root (Node (Rib (Rib::Kind::Normal), UNKNOWN_NODEID)),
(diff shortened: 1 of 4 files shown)


CohenArthur:
Hey @powerboat9 sorry for the confusion, I actually have a branch with these changes which I think is less hacky. We ended up working on the same thing in parallel. I didn't think you would work on this as well, my apologies.

I'll submit my branch this weekend and we can see which approach we like best. Apologies :(

powerboat9:
No worries, I figured you had probably been working on something similar, though I wasn't sure how far along you had gotten with it, or if you had abandoned it for being too hack-y.

P-E-P:
@powerboat9 I'm a bit out of the loop. Should we close this ? I think we recently merged some very similar changes.

CohenArthur:
@powerboat9 I think we should close this in favor of #4551, let me know what you think

powerboat9:
Yep, superseded by #4551

PR #4526, opened 2026-04-17
Fix ICE in get_function_expr when cfg'd return type inside macro

Harishankar14:
the problem is cfg-strip emits an error for unstrippable expressions but doesn't mark the parent for strip, leaving a broken subtree for later passes to ICE on.

gcc/rust/ChangeLog:

	* expand/rust-cfg-strip.cc (CfgStrip::visit): mark CallExpr for strip when function expression fails stripping. (CfgStrip::visit): mark ArrayIndexExpr for strip when array or index expression fails stripping.

gcc/testsuite/ChangeLog:

	* rust/compile/issue-4167.rs: New test.

in the git commit message.

- \[ ] Run `clang-format`

order to get feedback.
---

### DIFF
--- a/gcc/rust/expand/rust-cfg-strip.cc
+++ b/gcc/rust/expand/rust-cfg-strip.cc
@@ -828,9 +828,13 @@ CfgStrip::visit (AST::ArrayIndexExpr &expr)
 
   const auto &array_expr = expr.get_array_expr ();
   if (array_expr.is_marked_for_strip ())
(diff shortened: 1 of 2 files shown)


Harishankar14:
@powerboat9 , @P-E-P

P-E-P (approved):
Looks good, I wonder if we could not extend this to any node that does not have outer attributes.

Thank you!

PR #4527, opened 2026-04-18
Do path resolution in types NS for all segments except the last one

CohenArthur:
Per multiple comments in the rustc frontend this is how path resolution should actually be done. In a path like `foo::bar::baz`, the segments `foo` and `bar` must be resolved in the types NS, regardless of the namespace that we are trying to resolve the entire path in (e.g. even in a path function call like `foo::bar::baz()` where we expect to find the function in the values namespace). This also links with a further change where imports and modules must be inserted in the types namespace in order to be resolved to properly


gcc/rust/resolve/rust-forever-stack.h
```diff
@@ -857,8 +832,6 @@ template <Namespace N> class ForeverStack
   void reverse_iter (const Node &start,
 		     std::function<KeepGoing (const Node &)> lambda) const;
 
-  Node &cursor ();
-  const Node &cursor () const;
```
powerboat9: Why move `cursor ()` away from `update_cursor ()`?

gcc/rust/resolve/rust-name-resolution-context.hxx
```diff
@@ -0,0 +1,269 @@
...
+
+/**
+ * Split the actual path resolution logic in its own file because it's a lot,
+ * and it could get even worse for certain edge cases.
+ */
+
+namespace Rust {
+namespace Resolver2_0 {
+
+template <Namespace N>
```
powerboat9, line 30: I think we should make `ForeverStack` a non-template class before trying to merge it into `NameResolutionContext`

CohenArthur: I think that it makes more sense for ForeverStacks to still be separated by namespaces and I like the templated solution, but maybe I'm not seeing the issues it has?

gcc/rust/resolve/rust-forever-stack.h
CohenArthur: This was probably in an attempt to make it public before saying fuck it and putting everything as public, considering the ForeverStack is just a helper struct/class that gets used by the NRCtx and doesn't really "need" encapsulation. I'll move it back

gcc/rust/resolve/rust-name-resolution-context.hxx
CohenArthur: Also, in later commits I move more of the path resolution methods like `resolve_segments` and `resolve_final_segment` which benefit from being able to be specialized for the Types NS and so on

P-E-P (approved):

PR #4535, opened 2026-04-21
Implement Block Label.

### LINKED ISSUE #4534: [backend] cannot compile labeled block.
in #2689 labeled block are parsed and resolved but still we cannot compile it.

Islam-Imad:
implement block label.
- add a label after the end of the block.
- when we hit a break with a block label we assign a value to block temp var and goto the end of the block.

closes #4534

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -18,6 +18,7 @@
 
 #include "rust-compile-expr.h"
 #include "rust-backend.h"
+#include "rust-compile-context.h"
 #include "rust-compile-type.h"
 #include "rust-compile-struct-field-expr.h"
 #include "rust-compile-pattern.h"
@@ -32,6 +33,7 @@
 #include "realmpfr.h"
 #include "convert.h"
 #include "print-tree.h"
+#include "rust-hir-bound.h"
 #include "rust-hir-expr.h"
 #include "rust-system.h"
 #include "rust-tree.h"
@@ -436,12 +438,6 @@ CompileExpr::visit (HIR::IfExprConseqElse &expr)
 void
 CompileExpr::visit (HIR::BlockExpr &expr)
 {
-  if (expr.has_label ())
-    {
-      rust_error_at (expr.get_locus (), "labeled blocks are not supported");
-      return;
-    }
-
   TyTy::BaseType *block_tyty = nullptr;
   if (!ctx->get_tyctx ()->lookup_type (expr.get_mappings ().get_hirid (),
 				       &block_tyty))
@@ -454,18 +450,31 @@ CompileExpr::visit (HIR::BlockExpr &expr)
   fncontext fnctx = ctx->peek_fn ();
   tree enclosing_scope = ctx->peek_enclosing_scope ();
   tree block_type = TyTyResolveCompile::compile (ctx, block_tyty);
+  tree block_label = NULL_TREE;
 
   bool is_address_taken = false;
   tree ret_var_stmt = nullptr;
   tmp = Backend::temporary_variable (fnctx.fndecl, enclosing_scope, block_type,
 				     NULL, is_address_taken, expr.get_locus (),
 				     &ret_var_stmt);
+
   ctx->add_statement (ret_var_stmt);
+
+  if (expr.has_label ())
+    {
+      ctx->insert_var_decl (
+	expr.get_label ().get_lifetime ().get_mappings ().get_hirid (), tmp);
+      block_label = construct_block_label (expr);
+    }
 
   auto block_stmt = CompileBlock::compile (expr, ctx, tmp);
   rust_assert (TREE_CODE (block_stmt) == BIND_EXPR);
   ctx->add_statement (block_stmt);
 
+  if (block_label != NULL_TREE)
+    {
+      ctx->add_statement (block_label);
+    }
   translated = Backend::var_expression (tmp, expr.get_locus ());
 }
 
@@ -853,11 +862,25 @@ CompileExpr::visit (HIR::WhileLoopExpr &expr)
 void
 CompileExpr::visit (HIR::BreakExpr &expr)
 {
+  if (expr.has_break_expr () && expr.has_label ())
+    {
+      HIR::Lifetime &label = expr.get_label ();
+      auto tvar = lookup_temp_var (label.get_mappings ().get_nodeid ());
+      tree value = CompileExpr::Compile (expr.get_expr (), ctx);
+      tree assign
+	= Backend::assignment_statement (tvar->get_tree (label.get_locus ()),
+					 value, label.get_locus ());
+      tree block_label = lookup_label (label.get_mappings ().get_nodeid ());
+      tree go_to = Backend::goto_statement (block_label, label.get_locus ());
+      ctx->add_statement (assign);
+      ctx->add_statement (go_to);
+      return;
+    }
   if (expr.has_break_expr ())
     {
       tree compiled_expr = CompileExpr::Compile (expr.get_expr (), ctx);
-
       translated = error_mark_node;
+
       if (!ctx->have_loop_context ())
 	return;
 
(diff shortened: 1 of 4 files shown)


dkm:
Hello!

Thanks for this new contribution :)
We're going to try and have more descriptive commit log from now on, when that makes sense. It would be nice in this case to add extra text in the commit log so that readers can have a better idea of the problem you're handling, how you do that, optional issue reference, etc.

Thanks again!

Islam-Imad:
> Hello!
> 
> Thanks for this new contribution :) We're going to try and have more descriptive commit log from now on, when that makes sense. It would be nice in this case to add extra text in the commit log so that readers can have a better idea of the problem you're handling, how you do that, optional issue reference, etc.
> 
> Thanks again!

I have update the commit message and cleanup the remaining stuff :)

thanks Marc

Islam-Imad:
ping ?

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -32,11 +33,14 @@
...
 #include "convert.h"
 #include "print-tree.h"
+#include "rust-hir-bound.h"
 #include "rust-hir-expr.h"
 #include "rust-system.h"
 #include "rust-tree.h"
 #include "rust-tyty.h"
 #include "tree-core.h"
+#include "tree.h"
+#include <cassert>
```
CohenArthur: This include shouldn't be necessary

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -437,10 +441,13 @@ CompileExpr::visit (HIR::IfExprConseqElse &expr)
 void
 CompileExpr::visit (HIR::BlockExpr &expr)
 {
+  tree block_label = NULL_TREE;
   if (expr.has_label ())
     {
-      rust_error_at (expr.get_locus (), "labeled blocks are not supported");
-      return;
+      // rust_error_at (expr.get_locus (), "labeled blocks are not supported");
+      // return;
```
CohenArthur: dead code?

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -437,10 +441,13 @@ CompileExpr::visit (HIR::IfExprConseqElse &expr)
...
+      block_label = construct_block_label (expr);
+      // construct_temp_var (expr);
```
CohenArthur: here too?

gcc/testsuite/rust/execute/cf-labeled-block.rs
```diff
@@ -0,0 +1,36 @@
...
+            break 'block_b 333;
+        }
+        444
+    };
+    dump_number(b);
+    // ISSUE HERE :)
+    // let c = 'block_c: {
+    //     break 'block_c 555;
+    // };
+    // dump_number(c);
```
CohenArthur: what is the issue here?

CohenArthur (commented):
thanks for the work @Islam-Imad :) a couple comments but it looks good to me. I'll leave @philberty to review it as this touches the backend

gcc/testsuite/rust/execute/cf-labeled-block.rs
Islam-Imad: https://github.com/Rust-GCC/gccrs/issues/4538

gcc/rust/backend/rust-compile-expr.cc
Islam-Imad: removed.

gcc/testsuite/rust/execute/cf-labeled-block.rs
P-E-P: @Islam-Imad It would be worth mentioning it within the code. Please uses `Rust-GCC/gccrs#4538`

gcc/rust/backend/rust-compile-context.h
```diff
@@ -417,6 +432,7 @@ class Context
   std::map<HirId, tree> compiled_fn_map;
   std::map<HirId, tree> compiled_consts;
   std::map<HirId, tree> compiled_labels;
+  std::map<HirId, Bvariable *> compiled_temp_vars;
```
P-E-P: It just occurred to me that we're using map here but we should probably use unordered_map instead. I don't think we care about the HirdId order. It does not need to be fixed within that PR. We should probably open an issue instead.

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -853,11 +862,25 @@ CompileExpr::visit (HIR::WhileLoopExpr &expr)
 void
 CompileExpr::visit (HIR::BreakExpr &expr)
 {
+  if (expr.has_break_expr () && expr.has_label ())
+    {
+      HIR::Lifetime label = expr.get_label ();
```
P-E-P:
```suggestion
      HIR::Lifetime& label = expr.get_label ();
```

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -2758,5 +2780,58 @@ CompileExpr::generate_possible_fn_trait_call (HIR::CallExpr &expr,
   return true;
 }
 
+tree
+CompileExpr::construct_block_label (HIR::BlockExpr &expr)
+{
+  if (expr.has_label ())
+    {
+      fncontext fnctx = ctx->peek_fn ();
+      HIR::LoopLabel label = expr.get_label ();
```
P-E-P: Likewise

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -2758,5 +2780,58 @@ CompileExpr::generate_possible_fn_trait_call (HIR::CallExpr &expr,
...
+}
+
+Bvariable *
+CompileExpr::lookup_temp_var (NodeId to_be_resolved)
+{
+  HirId ref = resolve_NodeId (to_be_resolved);
+  Bvariable *ltemp = nullptr;
+  if (!ctx->lookup_temp_var (ref, &ltemp))
+    {
+      assert (false && "hey we failed to lookup temp var here bro");
```
P-E-P: Even if those messages are targeted at the developer I'd prefer if we kept things tidy and clean.
```suggestion
      assert (false && "failed to lookup temp var");
```

gcc/rust/backend/rust-compile-context.h
P-E-P: Couldn't we use `compiled_var_decls` ? Or do you want to explicitly keep those labels in a separate container ?  We need to either rename this or merge both maps.

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -2758,5 +2780,58 @@ CompileExpr::generate_possible_fn_trait_call (HIR::CallExpr &expr,
...
+    }
+  return NULL_TREE;
+}
+
+tree
+CompileExpr::lookup_label (NodeId to_be_resolved)
+{
+  HirId ref = resolve_NodeId (to_be_resolved);
+  tree label = NULL_TREE;
+  assert (ctx->lookup_label_decl (ref, &label) && "failed to lookup a label");
```
P-E-P: Please use rust_assert instead, this will insert the file location and help debugging.
```suggestion
  rust_assert (ctx->lookup_label_decl (ref, &label) && "failed to lookup a label");
```

P-E-P (changes requested):
This looks quite good given the amount of changes and their nature. I've put a few comments here and there but this is a great start!

gcc/rust/backend/rust-compile-context.h
Islam-Imad: yes i agree with you. 
The access time will be O(1) instead of O(logn) which is away more faster.
i will open an issue for it and i think it should be `good first`.

Islam-Imad: i didn't notice `compiled_var_decls` :)
the simple solution is to use it again instead of creating another map.

Islam-Imad:
ping ?

P-E-P (approved):
Looks great!

PR #4541, opened 2026-04-29
lex: Emit E0768 for empty non-decimal literals

### LINKED ISSUE #4490: Lexer silently accepts non-decimal integer literals without digits
### Summary

While refactoring the non-decimal integer literal parsing logic in the lexer, I noticed that gccrs does not enforce the presence of valid digits after a base prefix (0x, 0b, 0o). It silently accepts the empty prefix as a valid literal and evaluates it to 0.

### Reproducer

I tried this code:

```rust
fn main() {
    let _a: u32 = 0x;
    let _b: i32 = 0b;
    let _c: usize = 0o;
}
```

### Actual behavior

gccrs compiles the above code without emitting any errors. The lexer produces a valid token with the value 0 because the internal string parsing loop naturally terminates, and the empty string is evaluated as 0 by the underlying conversion logic.

### Expected behavior

```
error[E0768]: no valid digits found for number
 --> src/main.rs:2:19
  |
2 |     let _a: u32 = 0x;
  |                   ^^
```
```
error[E0768]: no valid digits found for number
 --> src/main.rs:3:19
  |
3 |     let _b: i32 = 0b;
  |                   ^^
```
```
error[E0768]: no valid digits found for number
 --> src/main.rs:4:21
  |
4 |     let _c: usize = 0o;
  |                     ^^
```
```
For more information about this error, try `rustc --explain E0768`.
error: could not compile `t` (bin "t") due to 3 previous errors
```

### GCC Version

commit-hash: 0c1834c239292244d3d9ed5b59d49f125fffe72d

nsvke:
Previously, the lexer evaluated empty non-decimal literals (like 0x, 0b, 0o) as 0. Now, it emits error E0768 when there are no valid digits.

gcc/rust/ChangeLog:

	* lex/rust-lex.cc (Lexer::parse_non_decimal_int_literal): Emit E0768.

gcc/testsuite/ChangeLog:

	* rust/compile/empty-non-decimal.rs: New test.

Closes #4490

### DIFF
--- a/gcc/rust/lex/rust-lex.cc
+++ b/gcc/rust/lex/rust-lex.cc
@@ -2150,15 +2150,22 @@ Lexer::parse_non_decimal_int_literal (location_t loc, IsDigitFunc is_digit_func,
   raw_str += current_char; // x, o, b
   skip_input ();
 
-  int length = 1;
+  int length = 2;
+  bool has_valid_digit = false;
 
   current_char = peek_input ();
 
-  length++;
-
   // loop through to add entire number to string
-  while (is_digit_func (current_char.value) || current_char == '_')
+  while (true)
     {
+      if (is_digit_func (current_char.value))
+	{
+	  has_valid_digit = true;
+	}
+      else if (current_char != '_')
+	{
+	  break;
+	}
       length++;
 
       raw_str += current_char;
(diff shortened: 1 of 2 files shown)


P-E-P (approved):
Looks good, thank you!

PR #4543, opened 2026-05-02
Return *this instead of this in `bbitmap.h`

SaiCharanMarrivada:
-  Fix the code to return reference to current `bbitmap<N>` object and correctly dereference `this` in `gcc/bbitmap.h` file.


powerboat9:
This looks good to me, but since this git repo is for the rust front end specifically, you'll want to email this in to gcc-patches instead.

PR #4545, opened 2026-05-06
Nix flake for Reproducible GCC-Rust development environment

nsvke:
This PR introduces a reproducible Nix development environment
for gccrs. It automatically handles all toolchain, glibc, and dependency
configurations.

Provides completely separate environments for 64-bit and 32-bit targets,
automatically preventing linker conflicts.

Injects useful global scripts (gccrs-setup, gccrs-build, gccrs-test, gx,
gccrs-commit-mklog, gccrs-fix-changelog, gccrs-mklog, gccrs-style,
gccrs-verify and gccrs-help) to automate the build process and ensure
GNU commit standards are met effortlessly.

contrib/ChangeLog:

	* nix/bash/gccrs-commit-mklog.sh: New file.
	* nix/bash/gccrs-fix-changelog.sh: New file.
	* nix/bash/gccrs-mklog.sh: New file.
	* nix/bash/gccrs-style.sh: New file.
	* nix/bash/gccrs-verify.sh: New file.
	* nix/flake.lock: New file.
	* nix/flake.nix: New file.
	* nix/packages/gccrs-contrib-tools.nix: New file.
	* nix/packages/gccrs-env-tools.nix: New file.

### DIFF
--- a/contrib/nix/bash/gccrs-commit-mklog.sh
+++ b/contrib/nix/bash/gccrs-commit-mklog.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+set -e
+
+ROOT_DIR=$(git rev-parse --show-toplevel 2>/dev/null || true)
+if [ -z "$ROOT_DIR" ]; then
+    echo "Warning: You must be inside a git repository to run this command." >&2
+    exit 1
+fi
+
+if git diff --cached --quiet; then
+    echo "Warning: No staged changes found. Run 'git add' first." >&2
+    exit 1
+fi
+
+TMP_MSG_FILE=$(mktemp -t gccrs-changelog-XXXXXX.txt)
+trap 'rm -f "$TMP_MSG_FILE"' EXIT
+
+if ! gmklog > "$TMP_MSG_FILE" && ! gccrs-mklog > "$TMP_MSG_FILE"; then
+    echo "Warning: Failed to generate ChangeLog template." >&2
+    exit 1
+fi
+
+if [ ! -s "$TMP_MSG_FILE" ]; then
+    echo "Warning: ChangeLog template is empty. Aborting commit." >&2
+    exit 1
+fi
+
+git commit -t "$TMP_MSG_FILE" "$@"
--- a/contrib/nix/bash/gccrs-fix-changelog.sh
+++ b/contrib/nix/bash/gccrs-fix-changelog.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+ROOT_DIR=$(git rev-parse --show-toplevel 2>/dev/null)
+
+if [ -z "$ROOT_DIR" ]; then
+    echo "Warning: You must be inside a git repository to run this command." >&2
+    exit 1
+fi
+
+SCRIPT_PATH="$ROOT_DIR/contrib/git-fix-changelog.py"
+
+if [ ! -f "$SCRIPT_PATH" ]; then
+    echo "Warning: '$SCRIPT_PATH' not found! Are you sure this is the gccrs repository?" >&2
+    exit 1
+fi
+
+python3 "$SCRIPT_PATH" "$@"
--- a/contrib/nix/bash/gccrs-mklog.sh
+++ b/contrib/nix/bash/gccrs-mklog.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+ROOT_DIR=$(git rev-parse --show-toplevel 2>/dev/null)
+
+if [ -z "$ROOT_DIR" ]; then
+    echo "Warning: You must be inside a git repository to run this command." >&2
+    exit 1
+fi
+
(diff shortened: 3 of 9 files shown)


gcc/rust/nix/flake.nix
```diff
@@ -0,0 +1,249 @@
+{
+  description = "Reproducible Nix environment for GCC Rust (gccrs) development";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
```
P-E-P: Could we pin this dependency to a specific commit ?

gcc/rust/nix/flake.nix
```diff
@@ -0,0 +1,249 @@
+{
```
P-E-P: I think the location of this file is incorrect. We should probably put this in `$GCC_TOP/contrib/`. I can't remember how we usually tag files that should not be sent upstream though.

@dkm probably has a lot of advice and opinions on that matter (and I'd like to hear them)

EDIT: Didn't read that part about the side repository for the flake at first. This would probably help a lot and reduce potential friction when upstreaming.

gcc/rust/nix/flake.nix
```diff
@@ -0,0 +1,249 @@
...
+      # gcc mklog script wrapper
+      gccrs-mklog = pkgs.writeShellApplication {
+        name = "gccrs-mklog";
+        runtimeInputs = [
+          pkgs.git
+          pythonEnv
+        ];
+        text = ''
+          if [ ! -f "contrib/mklog.py" ]; then
+              echo "Warning: 'contrib/mklog.py' not found! Use in 'gccrs' root folder."
```
P-E-P: Correct me if I'm wrong but this shouldn't happen if we retrieve the script from the store, you could use builtin.FilterSource to make it lighter.

gcc/rust/nix/flake.nix
nsvke: I was initially relying on the `flake.lock` file to handle the version pinning, but we can definitely pin the exact commit hash directly in `flake.nix` if preferred. I can update this in the next commit.

gcc/rust/nix/flake.nix
nsvke: You are absolutely right. However, if we proceed with the side-repository approach, the flake won't be able to access this script during the evaluation phase. The source code is only available after the environment is initialized (either via the `gccrs-setup` command or a manual clone). 

Depending on which repository structure we ultimately choose to move forward with, I can definitely rethink and refactor this for the most optimal solution.

gcc/rust/nix/flake.nix
dkm: Correct, that's really not something we should send upstream, at least not in this form.

It would be nice to have it in `contrib/` if it was meant to be shared, but as it's really a gccrs specific thing.

We could have it in `gcc/rust/<something>` to avoid any mistake. We already have a non-code `doc/` subdir.

Or, we could still have it in `contrib/`, BUT:
-  we also need to adjust the hardcoded list in https://github.com/Rust-GCC/gerris/blob/dkm/upstream-command/src/upstream.rs#L314 .
- as we don't directly push anything outside the rust directories, anything touching this file would NOT be sent upstream. Meaning that you'll need to avoid having commits mixing code changes and flake change, as these commits won't be sent upstream.

TLDR: the easier would be to keep it in `rust/` (sent upstream), but the cleaner would be to keep it in `contrib/` and be careful.

dkm:
I would try to add more info in the commit log. You have the info as the PR description is very well detailed.
Keep in mind that the git repository will probably be live for a long time. It's also really nice to have the basic info in the git log without having to browse github (e.g. think offline work, or if github is down, or if github dies).

gcc/rust/nix/flake.nix
P-E-P: Completely forgot about lock file, forget what I said.

gcc/rust/nix/flake.nix
powerboat9: `/contrib/gcc-git-customization.sh` shows a method for using `rev-parse` to handle execution outside the root folder.

nsvke:
I have completed a major refactoring of the Nix environment in my reference repository. To preserve the existing inline comment history, I am not pushing these commits to this draft PR.

gcc/rust/nix/flake.nix
CohenArthur: I think it's fine to keep it in contrib and add it to the hardcoded gerris list. If commits touch both the flake and general gccrs code, then there is an issue with the commit and it should be split up anyways. @P-E-P and @nsvke if you think it would be useful to have it upstream then I'm also okay with putting it in the `rust/` folder

nsvke: I agree that it should be in the `contrib/` folder. It cannot compile the full `gcc` anyway, as the environment is specifically designed for `gccrs`. If @P-E-P also approves this location, I can refactor it to work in `contrib/`.

PR #4555, opened 2026-05-25
ci: fix the issue reference scrapper when there's no ref

dkm:
grep returned an error when there was no reference, causing the workflow to stop.


CohenArthur (approved):
LGTM!

PR #4558, opened 2026-05-26
gccrs: Fix ICE when parsing empty path expression

Polygonalr:
This PR is basically a revival of #3286, that PR can be closed when this gets merged.

### DIFF
--- a/gcc/rust/parse/rust-parse-impl.hxx
+++ b/gcc/rust/parse/rust-parse-impl.hxx
@@ -7284,6 +7284,15 @@ Parser<ManagedTokenSource>::parse_stmt_or_expr ()
     case DOLLAR_SIGN:
       {
 	AST::PathInExpression path = parse_path_in_expression ();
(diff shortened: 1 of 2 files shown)


gcc/testsuite/rust/compile/paamayim_nekudotayim.rs
```diff
@@ -0,0 +1,8 @@
+#![feature(no_core)]
+#![no_core]
+
+// http://phpsadness.com/sad/1
```
dkm: I think you need to explain a bit more that "phpsadness" link, and the test file name...

Polygonalr: I was bringing the original joke from #3286's testcase over 🌚, but alright I've changed it to something more explicit.

CohenArthur (approved):
Looks good! Well done. I think your commit message looks good too :)

PR #4564, opened 2026-05-29
gccrs: Add Drop support for block-local variable

Lishin1215:
This PR consists of 4 commits:

For now, it handles cases like:

```rust
fn main() -> i32 {
    {
        let _x = Droppable;
    }

    0
}
```
It also emits a temporary rust_sorry_at for unsupported Drop ref/subpatterns.

----

**commit 1**
 gccrs: add block drop candidate tracking infrastructure
    
    gcc/rust/ChangeLog:
    
          * backend/rust-compile-context.h (struct DropCandidate): New struct
            for tracking block-local drop candidates.
---

**commit 2**
gccrs: add block-exit Drop calls
    
    gcc/rust/ChangeLog:
    
            * backend/rust-compile-block.cc (compile_drop_call): New helper to
            build a drop call.
            (CompileBlock::visit): Add drop calls for candidates.
            * backend/rust-compile-pattern.cc (type_has_drop_impl): New helper to
            check whether a type implements Drop.
            (CompilePatternLet::visit): Save initialized simple drop candidates.
    
    gcc/testsuite/ChangeLog:
    
            * rust/execute/drop-block-scope.rs: New test.

---

**commit 3**
gccrs: report unsupported Drop ref patterns
    
    gcc/rust/ChangeLog:
    
            * backend/rust-compile-pattern.cc (CompilePatternLet::visit): Check ref
            pattern base types and emit a sorry for unsupported Drop ref/subpatterns.
    
    gcc/testsuite/ChangeLog:
    
            * rust/execute/drop-block-scope.rs: Extend test coverage for simple
            immutable and mutable bindings.
            * rust/compile/drop-ref-pattern.rs: New test.

---

**commit 4**
gccrs: move Drop helpers to separate files

gcc/rust/ChangeLog:

            * Make-lang.in: Add rust-compile-drop.o.
            * backend/rust-compile-block.cc (compile_drop_call): Move to CompileDrop.
            (CompileBlock::visit): Use CompileDrop to emit scope drop calls.
            * backend/rust-compile-context.h (struct DropCandidate): Move to rust-compile-drop.h.
            * backend/rust-compile-pattern.cc (type_has_drop_impl): Move to CompileDrop.
            (CompilePatternLet::visit): Use CompileDrop to check Drop impls.
            * backend/rust-compile-drop.cc: New file.
            * backend/rust-compile-drop.h: New file.
        
---
Tested with:

```bash
        make all-gcc
        make check-rust RUNTESTFLAGS="--all compile.exp=drop-ref-pattern.rs"
        make check-rust RUNTESTFLAGS="--all compile.exp=drop-lang-item.rs"
        make check-rust RUNTESTFLAGS="--all execute.exp=drop-block-scope.rs"

### DIFF
--- a/gcc/rust/backend/rust-compile-block.cc
+++ b/gcc/rust/backend/rust-compile-block.cc
@@ -17,52 +17,13 @@
 // <http://www.gnu.org/licenses/>.
 
 #include "rust-compile-block.h"
-#include "rust-compile-stmt.h"
+#include "rust-compile-drop.h"
 #include "rust-compile-expr.h"
-#include "rust-compile-implitem.h"
+#include "rust-compile-stmt.h"
 #include "rust-hir-expr.h"
-#include "rust-hir-path-probe.h"
-#include "rust-lang-item.h"
 
 namespace Rust {
 namespace Compile {
-
-static tree
-compile_drop_call (Context *ctx, Bvariable *var, TyTy::BaseType *ty,
-		   location_t locus)
-{
-  auto drop_lang = ctx->get_mappings ().lookup_lang_item (LangItem::Kind::DROP);
-  if (!drop_lang.has_value ())
-    return NULL_TREE;
-
-  Resolver::TraitReference *drop_ref = nullptr;
-  bool ok
-    = ctx->get_tyctx ()->lookup_trait_reference (drop_lang.value (), &drop_ref);
-  if (!ok)
-    return NULL_TREE;
-
-  HIR::PathIdentSegment segment ("drop");
-  auto candidates
-    = Resolver::PathProbeImplTrait::Probe (ty->get_root (), segment, drop_ref);
-
-  for (auto &candidate : candidates)
-    {
-      if (!candidate.is_impl_candidate ()
-	  || candidate.ty->get_kind () != TyTy::TypeKind::FNDEF)
-	continue;
-
-      auto *fn_type = static_cast<TyTy::FnType *> (candidate.ty);
-      tree fn_addr
-	= CompileInherentImplItem::Compile (candidate.item.impl.impl_item, ctx,
-					    fn_type, locus);
-
-      tree var_expr = Backend::var_expression (var, locus);
-      tree var_addr = HIRCompileBase::address_expression (var_expr, locus);
-
-      return Backend::call_expression (fn_addr, {var_addr}, nullptr, locus);
-    }
-  return NULL_TREE;
-}
 
 CompileBlock::CompileBlock (Context *ctx, Bvariable *result)
   : HIRCompileBase (ctx), translated (nullptr), result (result)
@@ -124,24 +85,7 @@ CompileBlock::visit (HIR::BlockExpr &expr)
 					 expr.get_locus ());
       ctx->add_statement (assignment);
     }
-
-  auto &drop_candidates = ctx->peek_block_drop_candidates ();
-
-  for (auto it = drop_candidates.rbegin (); it != drop_candidates.rend (); ++it)
-    {
-      TyTy::BaseType *ty = nullptr;
-      Bvariable *var = nullptr;
-
-      bool ok = ctx->get_tyctx ()->lookup_type (it->hirid, &ty);
-      rust_assert (ok);
-
-      ok = ctx->lookup_var_decl (it->hirid, &var);
-      rust_assert (ok);
-
-      tree drop_call = compile_drop_call (ctx, var, ty, it->locus);
-      if (drop_call != NULL_TREE)
-	ctx->add_statement (convert_to_void (drop_call, ICV_STATEMENT));
-    }
+  CompileDrop::emit_current_scope_drop_calls (ctx);
 
   ctx->pop_block ();
   translated = new_block;
--- a/gcc/rust/backend/rust-compile-block.cc
+++ b/gcc/rust/backend/rust-compile-block.cc
@@ -19,10 +19,50 @@
 #include "rust-compile-block.h"
 #include "rust-compile-stmt.h"
 #include "rust-compile-expr.h"
+#include "rust-compile-implitem.h"
 #include "rust-hir-expr.h"
+#include "rust-hir-path-probe.h"
+#include "rust-lang-item.h"
 
 namespace Rust {
 namespace Compile {
+
+static tree
+compile_drop_call (Context *ctx, Bvariable *var, TyTy::BaseType *ty,
+		   location_t locus)
+{
+  auto drop_lang = ctx->get_mappings ().lookup_lang_item (LangItem::Kind::DROP);
+  if (!drop_lang.has_value ())
+    return NULL_TREE;
+
+  Resolver::TraitReference *drop_ref = nullptr;
+  bool ok
+    = ctx->get_tyctx ()->lookup_trait_reference (drop_lang.value (), &drop_ref);
+  if (!ok)
+    return NULL_TREE;
+
+  HIR::PathIdentSegment segment ("drop");
+  auto candidates
+    = Resolver::PathProbeImplTrait::Probe (ty->get_root (), segment, drop_ref);
+
+  for (auto &candidate : candidates)
+    {
+      if (!candidate.is_impl_candidate ()
+	  || candidate.ty->get_kind () != TyTy::TypeKind::FNDEF)
+	continue;
+
+      auto *fn_type = static_cast<TyTy::FnType *> (candidate.ty);
+      tree fn_addr
+	= CompileInherentImplItem::Compile (candidate.item.impl.impl_item, ctx,
+					    fn_type, locus);
+
+      tree var_expr = Backend::var_expression (var, locus);
+      tree var_addr = HIRCompileBase::address_expression (var_expr, locus);
+
+      return Backend::call_expression (fn_addr, {var_addr}, nullptr, locus);
+    }
+  return NULL_TREE;
+}
 
 CompileBlock::CompileBlock (Context *ctx, Bvariable *result)
   : HIRCompileBase (ctx), translated (nullptr), result (result)
@@ -83,6 +123,24 @@ CompileBlock::visit (HIR::BlockExpr &expr)
 	= Backend::assignment_statement (result_reference, compiled_expr,
 					 expr.get_locus ());
       ctx->add_statement (assignment);
+    }
+
+  auto &drop_candidates = ctx->peek_block_drop_candidates ();
+
+  for (auto it = drop_candidates.rbegin (); it != drop_candidates.rend (); ++it)
+    {
+      TyTy::BaseType *ty = nullptr;
+      Bvariable *var = nullptr;
+
+      bool ok = ctx->get_tyctx ()->lookup_type (it->hirid, &ty);
+      rust_assert (ok);
+
+      ok = ctx->lookup_var_decl (it->hirid, &var);
+      rust_assert (ok);
+
+      tree drop_call = compile_drop_call (ctx, var, ty, it->locus);
+      if (drop_call != NULL_TREE)
+	ctx->add_statement (convert_to_void (drop_call, ICV_STATEMENT));
     }
 
   ctx->pop_block ();
--- a/gcc/rust/backend/rust-compile-context.h
+++ b/gcc/rust/backend/rust-compile-context.h
@@ -50,6 +50,12 @@ struct CustomDeriveInfo
   std::vector<std::string> attributes;
 };
 
+struct DropCandidate
+{
+  HirId hirid;
+  location_t locus;
+};
+
 class Context
 {
 public:
@@ -101,6 +107,7 @@ public:
   {
     scope_stack.push_back (scope);
     statements.push_back ({});
+    block_drop_candidates.emplace_back ();
   }
 
   tree pop_block ()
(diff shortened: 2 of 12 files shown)

### CODE REVIEW

File: gcc/rust/backend/rust-compile-pattern.cc [hunk 1]
```diff
@@ -1343,6 +1368,11 @@ CompilePatternLet::visit (HIR::IdentifierPattern &pattern)
       auto s = Backend::init_statement (fnctx.fndecl, var, init_expr);
       ctx->add_statement (s);
     }
+
```
Lishin1215, line 1350: For now I only handle the simple case like:
- let x = Droppable;
- let mut x = Droppable;
- let _x = Droppable;

I skipped more complex patterns for this first version, like ref and subpatterns.
I wanted to start with the smallest case first and then extend it later.
Does this scope look okay for a first draft?

gcc/rust/backend/rust-compile-context.h
```diff
@@ -131,6 +141,18 @@ class Context
 
   void add_statement (tree stmt) { statements.back ().push_back (stmt); }
 
+  std::vector<DropCandidate> &peek_block_drop_candidates ()
+  {
+    rust_assert (!block_drop_candidates.empty ());
+    return block_drop_candidates.back ();
+  }
+
+  void note_simple_drop_candidate (HirId hirid, location_t locus)
```
Lishin1215, line 145: I am storing the drop candidates in Context for now.
The idea is that when we encounter an eligible variable in `let`, I store it in the array.
Please let me know if this is an ideal place.

gcc/rust/backend/rust-compile-block.cc
```diff
@@ -19,11 +19,51 @@
...
+#include "rust-compile-implitem.h"
 #include "rust-hir-expr.h"
+#include "rust-hir-path-probe.h"
+#include "rust-lang-item.h"
 
 namespace Rust {
 namespace Compile {
 
+static tree
+compile_drop_call (Context *ctx, Bvariable *var, TyTy::BaseType *ty,
```
Lishin1215: This helper finds the `Drop trait`, looks for the drop method, and builds the function call.
I put it in `rust-compile-block.cc` for now because it is only used when leaving a block.
I think this may need to move to a separate file later, since other scenarios may also need to use it.

gcc/rust/backend/rust-compile-pattern.cc
```diff
@@ -27,11 +27,36 @@
...
+#include "rust-hir-type-bounds.h"
+#include "rust-hir-trait-reference.h"
+#include "rust-lang-item.h"
 #include "tree.h"
 
 namespace Rust {
 namespace Compile {
 
+static bool
+type_has_drop_impl (Context *ctx, TyTy::BaseType *ty)
```
Lishin1215: I added this helper to avoid getting variables whose type does not implement `Drop`.

gcc/rust/backend/rust-compile-pattern.cc
Polygonalr: Then you might want to add a placeholder error message for now:

```c++
else if (type_has_drop_impl (ctx, ty))
    rust_sorry_at (pattern.get_locus (), "drop trait not supported for subpatterns & ref patterns");
```

gcc/testsuite/rust/execute/drop-block-scope.rs
```diff
@@ -0,0 +1,34 @@
...
+
+extern "C" {
+    fn printf(s: *const i8, ...);
+}
+
+#[lang = "sized"]
+pub trait Sized {}
+
+#[lang = "drop"]
+pub trait Drop {
```
Polygonalr, line 15: In the future, I think this should be implemented as part of the compiler. You may want to take a look at how other traits (e.g. `Copy`, `Eq`) are implemented under `gcc/rust/expand/rust-derive.cc`.

Polygonalr (commented):
Arthur or Pierre-Emmanuel will do the proper review, but minor stuff from me.

gcc/rust/backend/rust-compile-pattern.cc
Lishin1215: Thank you for pointing this out. I’ll add a `rust_sorry_at` for Drop types with ref bindings and subpatterns.

gcc/testsuite/rust/execute/drop-block-scope.rs
Lishin1215: Thank you, I’ll take a look at `rust-derive.cc`. For this draft, I kept the no_core test self-contained, but I agree the Drop support should be handled more directly in the compiler in the future.

gcc/rust/backend/rust-compile-block.cc
P-E-P: > This helper finds the Drop trait, looks for the drop method, and builds the function call.

If you felt you had to write this in a github comment writing it as a regular code comment may make sense

gcc/rust/backend/rust-compile-context.h
```diff
@@ -131,6 +141,18 @@ class Context
...
+  {
+    rust_assert (!block_drop_candidates.empty ());
+    block_drop_candidates.back ().push_back ({hirid, locus});
```
P-E-P: Couldn't you emplace_back here since you're building your DropCandidate?

P-E-P (approved):

gcc/rust/backend/rust-compile-block.cc
Lishin1215: Thanks, that makes sense. I'll move that explanation into a comment above `compile_drop_call`.

gcc/rust/backend/rust-compile-context.h
Lishin1215: Yes, good point. I'll add a small DropCandidate constructor and use `emplace_back` here.

gcc/rust/backend/rust-compile-context.h
```diff
@@ -50,6 +50,12 @@ struct CustomDeriveInfo
   std::vector<std::string> attributes;
 };
 
+struct DropCandidate
+{
+  HirId hirid;
+  location_t locus;
+};
+
```
CohenArthur: I would like to see separate files for these since they are exclusively drop-related. I think it's a great idea to create a separate struct, but it will eventually become more complicated and more methods/functions will be added, so it makes sense to create its own file now

gcc/rust/backend/rust-compile-block.cc
CohenArthur: Likewise this is a great function but I would put it in a class in a separate file. It would also help with caching down the line if that's something we eventually want to do

gcc/rust/backend/rust-compile-context.h
CohenArthur: I think it's fine for now, but we should think about what future APIs will look like and start building towards that in a later PR

gcc/rust/backend/rust-compile-pattern.cc
CohenArthur: Likewise I think this function would be great in a file dedicated to Drop functions

CohenArthur (commented):
I think this looks really good as a first approach and the code is clear. Well done

gcc/rust/backend/rust-compile-context.h
Lishin1215: I'll keep the current approach for now and think about what future APIs should look like in a follow-up PR.
Thank you!

gcc/rust/backend/rust-compile-pattern.cc
Lishin1215: Done, I moved it to `CompileDrop` in `rust-compile-drop.h` and `rust-compile-drop.cc`.
Thank you!

gcc/rust/backend/rust-compile-context.h
Lishin1215: Done, I moved `DropCandidate` to `rust-compile-drop.h` together with the other Drop-related helpers. 
Thank you!

gcc/rust/backend/rust-compile-block.cc
Lishin1215: Done, I also moved it into the `CompileDrop` in `rust-compile-drop.h` and `rust-compile-drop.cc`, which should make it easier to extend later. 
Thanks!

gcc/rust/backend/rust-compile-drop.h
```diff
@@ -0,0 +1,42 @@
+#ifndef RUST_COMPILE_DROP
```
P-E-P: This is not always enforced within GCC but to avoid collision we usually suffix the header guard variable with _H

gcc/rust/backend/rust-compile-drop.cc
```diff
@@ -0,0 +1,97 @@
+#include "rust-compile-drop.h"
```
P-E-P, line 19: You need to add the GNU copyright comments. If you copy and paste it watch out the years. This file has been created just now, the copyright mention should reflect that.

gcc/rust/backend/rust-compile-drop.h
P-E-P: Likewise about copyright header.

gcc/rust/backend/rust-compile-drop.cc
CohenArthur: Check a recent file like gcc/rust/util/rust-cloneable.h and you can copy paste the Copyright comment from there

gcc/rust/backend/rust-compile-drop.h
Lishin1215: Thanks! I fixed it. 
Thank you for catching this.

gcc/rust/backend/rust-compile-drop.cc
Lishin1215: Thanks! I added the copyright header and changed the year to 2026.

gcc/rust/backend/rust-compile-drop.h
```diff
@@ -1,29 +1,28 @@
...
 
-class Context;
+#ifndef RUST_COMPILE_DROP_H
+#define RUST_COMPILE_DROP_H
 
-struct DropCandidate
-{
-  DropCandidate (HirId hirid, location_t locus) : hirid (hirid), locus (locus)
-  {}
+#include "rust-compile-context.h"
```
Lishin1215, line 22: I reworked this to avoid the forward declarations. `rust-compile-drop.h` now includes the required context header directly.

`DropCandidate` now lives in `rust-compile-drop-candidate.h`, so it remains in a drop-related file, while `rust-compile-context.h` only includes that small state type instead of the full `rust-compile-drop.h` header.

Does this look good to you?

P-E-P: Yes it does!

P-E-P (approved):
LGTM

gcc/rust/backend/rust-compile-drop.cc
```diff
@@ -0,0 +1,115 @@
...
+      tree fn_addr
+	= CompileInherentImplItem::Compile (candidate.item.impl.impl_item, ctx,
+					    fn_type, locus);
+
+      tree var_expr = Backend::var_expression (var, locus);
+      tree var_addr = HIRCompileBase::address_expression (var_expr, locus);
+
+      return Backend::call_expression (fn_addr, {var_addr}, nullptr, locus);
+    }
+  return NULL_TREE;
```
CohenArthur: Here we should only ever have one candidate which corresponds to the actual `Drop` lang item trait and its `drop` function, otherwise we are doing something really evil and the compiler should crash.


```suggestion

  rust_assert (candidates.size () == 1);

  auto &candidate = *candidates.begin ();

  rust_assert (candidate.is_impl_candidate ());
  rust_assert (candidate.ty->get_kind () == TyTy::TypeKind::FNDEF);

  auto *fn_type = static_cast<TyTy::FnType *> (candidate.ty);
  tree fn_addr
    = CompileInherentImplItem::Compile (candidate.item.impl.impl_item, ctx,
					fn_type, locus);

  tree var_expr = Backend::var_expression (var, locus);
  tree var_addr = HIRCompileBase::address_expression (var_expr, locus);

  return Backend::call_expression (fn_addr, {var_addr}, nullptr, locus);

```

gcc/rust/backend/rust-compile-drop.h
```diff
@@ -0,0 +1,41 @@
...
+
+class CompileDrop
+{
+public:
+  static bool type_has_drop_impl (Context *ctx, TyTy::BaseType *ty);
+
+  static tree compile_drop_call (Context *ctx, Bvariable *var,
+				 TyTy::BaseType *ty, location_t locus);
+
+  static void emit_current_scope_drop_calls (Context *ctx);
```
CohenArthur, line 35: Since we are always passing the `Context *ctx` argument in each of these functions, I think it would be worth in a later PR to keep it as a class member and reuse it. But don't do it now as this is fine for now 👍

CohenArthur (approved):
One small comment but the implementation looks really good to me, well done Janet

gcc/rust/backend/rust-compile-drop.cc
Lishin1215: Thank you, I understand now. Since this should resolve to exactly one `Drop candidate`, I changed it to use assertions so invalid compiler state is caught instead of being ignored.

gcc/rust/backend/rust-compile-drop.h
Lishin1215: Thanks, that makes sense. I’ll keep the current approach for now and consider this in a follow-up cleanup PR.

CohenArthur (approved):
looks good to me! well done!!

PR #4565, opened 2026-05-31
gccrs: Implement C-style string literals

Polygonalr:
This is not part of Rust 1.49, but it's nice to backport it for now, since Rust for Linux uses C-style string literals a bit.

Current behaviour is similar to normal string literals, except they are type-checked as slice types instead of array, and the compiled fat pointer has +1 to size to include the null terminator.

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -717,8 +717,23 @@ CompileExpr::visit (HIR::FieldAccessExpr &expr)
 				       nullptr, &field_index);
       rust_assert (ok);
 
-      tree indirect = indirect_expression (receiver_ref, expr.get_locus ());
-      receiver_ref = indirect;
+      // TODO this check is only used for CStr, test again when we support
+      // compilation of #[repr(transparent)] structs
+      if (RS_DST_FLAG_P (TREE_TYPE (receiver_ref)))
+	{
+	  const TyTy::StructFieldType *field
+	    = variant->get_field_at_index (field_index);
+	  tree field_type
+	    = TyTyResolveCompile::compile (ctx, field->get_field_type ());
+	  translated = fold_build1_loc (expr.get_locus (), VIEW_CONVERT_EXPR,
+					field_type, receiver_ref);
+	  return;
+	}
+      else
+	{
+	  tree indirect = indirect_expression (receiver_ref, expr.get_locus ());
+	  receiver_ref = indirect;
+	}
     }
 
   translated = Backend::struct_field_expression (receiver_ref, field_index,
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -1089,6 +1089,10 @@ CompileExpr::visit (HIR::LiteralExpr &expr)
     case HIR::Literal::BYTE_STRING:
       translated = compile_byte_string_literal (expr, tyty);
       return;
+
+    case HIR::Literal::C_STRING:
+      translated = compile_c_string_literal (expr, tyty);
+      return;
     }
 }
 
@@ -1900,6 +1904,31 @@ CompileExpr::compile_byte_string_literal (const HIR::LiteralExpr &expr,
 					     expr.get_locus ());
 
   return address_expression (constructed, expr.get_locus ());
+}
+
+tree
+CompileExpr::compile_c_string_literal (const HIR::LiteralExpr &expr,
+				       const TyTy::BaseType *tyty)
+{
+  // Copied from compile_string_literal
+  tree fat_pointer = TyTyResolveCompile::compile (ctx, tyty);
+
+  rust_assert (expr.get_lit_type () == HIR::Literal::C_STRING);
+  const auto literal_value = expr.get_literal ();
+
+  auto base = Backend::string_constant_expression (literal_value.as_string ());
+  tree data = address_expression (base, expr.get_locus ());
+
+  TyTy::BaseType *usize = nullptr;
+  bool ok = ctx->get_tyctx ()->lookup_builtin ("usize", &usize);
+  rust_assert (ok);
+  tree type = TyTyResolveCompile::compile (ctx, usize);
+
+  // +1 for null terminator, unlike Rust string literals.
+  tree size = build_int_cstu (type, literal_value.as_string ().size () + 1);
+
+  return Backend::constructor_expression (fat_pointer, false, {data, size}, -1,
+					  expr.get_locus ());
 }
 
 tree
--- a/gcc/rust/backend/rust-compile-type.cc
+++ b/gcc/rust/backend/rust-compile-type.cc
@@ -650,6 +650,7 @@ TyTyResolveCompile::visit (const TyTy::ReferenceType &type)
   const TyTy::SliceType *slice = nullptr;
   const TyTy::StrType *str = nullptr;
   const TyTy::DynamicObjectType *dyn = nullptr;
+  const TyTy::ADTType *adt = nullptr;
   if (type.is_dyn_slice_type (&slice))
     {
       tree type_record = create_slice_type_record (*slice);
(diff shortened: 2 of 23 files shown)


gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -1900,6 +1904,31 @@ CompileExpr::compile_byte_string_literal (const HIR::LiteralExpr &expr,
...
+  bool ok = ctx->get_tyctx ()->lookup_builtin ("usize", &usize);
+  rust_assert (ok);
+  tree type = TyTyResolveCompile::compile (ctx, usize);
+
+  // +1 for null terminator, unlike Rust string literals.
+  tree size = build_int_cstu (type, literal_value.as_string ().size () + 1);
+
+  return Backend::constructor_expression (fat_pointer, false, {data, size}, -1,
+					  expr.get_locus ());
+}
```
CohenArthur, line 1947: I think we have to be careful here since C string literals are expected to produce references of `CStr` if I'm understanding things right

From https://doc.rust-lang.org/nightly/edition-guide/rust-2021/c-string-literals.html :

 > Literals of the form c"foo" or cr"foo" represent a string of type [&core::ffi::CStr](https://doc.rust-lang.org/nightly/core/ffi/struct.CStr.html).

So while the layout of a `CStr` and a `[c_char]` is the same...

```rust
#[derive(Hash)]
#[stable(feature = "rust1", since = "1.0.0")]
// FIXME:
// `fn from` in `impl From<&CStr> for Box<CStr>` current implementation relies
// on `CStr` being layout-compatible with `[u8]`.
// When attribute privacy is implemented, `CStr` should be annotated as `#[repr(transparent)]`.
// Anyway, `CStr` representation and layout are considered implementation detail, are
// not documented and must not be relied upon.
pub struct CStr {
    // FIXME: this should not be represented with a DST slice but rather with
    //        just a raw `c_char` along with some form of marker to make
    //        this an unsized type. Essentially `sizeof(&CStr)` should be the
    //        same as `sizeof(&c_char)` but `CStr` should be an unsized type.
    inner: [c_char],
}
```

I think we should still figure out how to convert to the expected type, otherwise we're going to have failures when compiling the kernel

gcc/rust/lex/rust-lex.cc
```diff
@@ -1063,6 +1063,10 @@ Lexer::build_token ()
 	    return parse_raw_byte_string (loc);
 	}
 
+      // C-style strings
+      else if (current_char == 'c' && peek_input () == '"')
+	return parse_c_string (loc);
```
CohenArthur, line 1068: Note that in a later PR you can also add C-style raw strings the same way, but I don't know if the kernel makes use of those

gcc/rust/lex/rust-lex.cc
```diff
@@ -1749,6 +1753,82 @@ Lexer::parse_byte_string (location_t loc)
...
+	}
+
+      current_column++;
+      if (current_char.value == '\n')
+	{
+	  current_line++;
+	  current_column = 1;
+	  // tell line_table that new line starts
+	  start_line (current_line, max_column_hint);
+	}
```
CohenArthur, line 1803: If this is the same code as other string parsing logic let's think about factoring it into a function

gcc/rust/parse/rust-parse-impl-expr.hxx
```diff
@@ -2101,6 +2123,20 @@ Parser<ManagedTokenSource>::null_denotation_not_path (
       return std::unique_ptr<AST::LiteralExpr> (
 	new AST::LiteralExpr (tok->get_str (), AST::Literal::RAW_STRING,
 			      tok->get_type_hint (), {}, tok->get_locus ()));
+    case C_STRING_LITERAL:
+      if (flag_c_style_string_literals)
+	return std::unique_ptr<AST::LiteralExpr> (
+	  new AST::LiteralExpr (tok->get_str (), AST::Literal::C_STRING,
+				tok->get_type_hint (), {}, tok->get_locus ()));
```
CohenArthur: Formatting nitpick since we have an else block with brackets
```suggestion
      if (flag_c_style_string_literals)
    {
	return std::unique_ptr<AST::LiteralExpr> (
	  new AST::LiteralExpr (tok->get_str (), AST::Literal::C_STRING,
				tok->get_type_hint (), {}, tok->get_locus ()));
    }
```

CohenArthur (commented):
I think the code is really good, but I'd like to have a think about that resulting `&CStr` type vs `[u8]` type before we merge it. Can you please check that as soon as C-style string literals were added, `CStr` became a lang item, in which case we should be able to refer to it within the typechecker and use that type, and if so, we have to figure out a way to get to the `CStr` type with a 1.49 libcore which does not have the `CStr` as a lang item yet

dkm:
FWIW, the changelog is not correct. You need to follow the ` * file.ext (item): What changed` pattern.
Please do not simply point at the file :) See for example:

```
gcc/rust/ChangeLog:
	* lex/rust-token.h: Add C_STRING_LITERAL and RAW_C_STRING_LITERAL (unused
	for now).
	* lex/rust-lex.h: Define new parse_c_string function.
	* lex/rust-lex.cc: Implement lexing for C-style string literals.
	* hir/tree/rust-hir-literal.h: Define new C_STRING LitType.
```

Can you do some cleanup there? Thanks!

gcc/rust/backend/rust-compile-expr.cc
Polygonalr: Added `CStr` language item in my latest commit.

Polygonalr:
> FWIW, the changelog is not correct. You need to follow the ` * file.ext (item): What changed` pattern. Please do not simply point at the file :) See for example:
> 
> ```
> gcc/rust/ChangeLog:
> 	* lex/rust-token.h: Add C_STRING_LITERAL and RAW_C_STRING_LITERAL (unused
> 	for now).
> 	* lex/rust-lex.h: Define new parse_c_string function.
> 	* lex/rust-lex.cc: Implement lexing for C-style string literals.
> 	* hir/tree/rust-hir-literal.h: Define new C_STRING LitType.
> ```
> 
> Can you do some cleanup there? Thanks!

@dkm Just did some commit message clean-up, apologies for not being specific enough in my commit messages :sweat:, I'll do my best not to repeat this in the future.

gcc/rust/util/rust-lang-item.cc
```diff
@@ -92,6 +92,7 @@ const BiMap<std::string, LangItem::Kind> Rust::LangItem::lang_items = {{
   {"slice_u8", Kind::SLICE_U8},
   {"slice", Kind::SLICE},
   {"str", Kind::STR},
+  {"CStr", Kind::CSTR},
```
CohenArthur, line 97: This needs an extra comment (and possibly to be put at the end of the `BiMap`) with an explanation saying it's only used when compiling the Linux kernel with a patched core, and isn't present in Rust 1.49

gcc/rust/util/rust-lang-item.h
```diff
@@ -125,6 +125,7 @@ class LangItem
     SLICE_U8,
     SLICE,
     STR,
+    CSTR,
```
CohenArthur, line 130: Likewise

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -1902,6 +1904,31 @@ CompileExpr::compile_byte_string_literal (const HIR::LiteralExpr &expr,
...
+  auto base = Backend::string_constant_expression (literal_value.as_string ());
+  tree data = address_expression (base, expr.get_locus ());
+
+  TyTy::BaseType *usize = nullptr;
+  bool ok = ctx->get_tyctx ()->lookup_builtin ("usize", &usize);
+  rust_assert (ok);
+  tree type = TyTyResolveCompile::compile (ctx, usize);
+
+  // +1 for null terminator, unlike Rust string literals.
+  tree size = build_int_cstu (type, literal_value.as_string ().size () + 1);
```
CohenArthur, line 1943: This is correct, but don't we also need to add an extra zero byte at the end?

CohenArthur: Is it taken care of by the backend already?

CohenArthur (commented):
Good work @Polygonalr :) well done

gcc/rust/backend/rust-compile-expr.cc
Polygonalr: [C++11 strings already come with a null terminator](https://stackoverflow.com/questions/6077189/is-stdstring-always-null-terminated-in-c11), so I don't think we need to handle it.

gcc/rust/lex/rust-lex.cc
Polygonalr: There are subtle differences (different error messages, using `parse_escape` instead of `parse_utf8_escape`), so this was not abstracted out to a separate function.

gcc/rust/backend/rust-compile-type.cc
```diff
@@ -684,6 +685,22 @@ TyTyResolveCompile::visit (const TyTy::ReferenceType &type)
 
       return;
     }
+  // Check for CStr, create a specific record for it
+  else if (type.is_dyn_cstr_type (&adt))
+    {
+      // &CStr is a fat pointer: { *const u8, usize }
```
CohenArthur:
```rust
#[repr(transparent)]
CStr {
   inner: [u8]
}
```

CohenArthur (approved):
LGTM, sorry for the nitpick :)

CohenArthur (commented):
Please add an execute test to ensure that the last character in the CStr is indeed a zero byte. Something like this:

```rust
fn main() -> i32 {
   let a = c"something";
   
   last_byte(a) as i32
}
```

philberty (approved):
LGTM

PR #4569, opened 2026-06-02
Limit globbing visitor

powerboat9:
The globbing visitor would previously descend into some kinds of rust item, instead of registering them as glob imports or ignoring them. This would cause items nested in those items to be glob imported, despite them not being top level items inside the glob container.

Kudos to @CohenArthur for finding this bug.


CohenArthur (approved):
well done on the fix! thanks!

CohenArthur:
Tested on my branch with `core` and it does fix the issue I had found, thanks @powerboat9

PR #4573, opened 2026-06-03
ci: prevent @noreply.github.com emails

### LINKED ISSUE #4568: Forbid no-reply github email addresses in commits and commits messages
Since the goal is for every commit to be upstreamed, it is expected that the author can be contacted via email to be asked about their contribution

dkm:
github allows contribution without disclosing author's email. It uses a "@noreply.github.com" in the commit log, which is not OK for upstream contribution.

Closes Rust-GCC/gccrs#4568

### DIFF
--- a/.github/workflows/commit-format.yml
+++ b/.github/workflows/commit-format.yml
@@ -109,6 +109,38 @@ jobs:
 
           exit $retval;
 
+  check-valid-email:
+    runs-on: ubuntu-latest
+    name: check-valid-email
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+
+      - name: Check author's email looks correct
+        run: |
+          retval=0;
+
+          rev_list="origin/${{ github.event.pull_request.base.ref }}..${{ github.event.pull_request.head.sha }}"
+          for commit in $(git rev-list --reverse "$rev_list"); do
+
+            if M=$(git show -s "$commit" | grep "noreply.github.com"); then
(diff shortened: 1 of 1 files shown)


.github/workflows/commit-format.yml
```diff
@@ -109,6 +109,38 @@ jobs:
...
+          fetch-depth: 0
+
+      - name: Check author's email looks correct
+        run: |
+          retval=0;
+
+          rev_list="origin/${{ github.event.pull_request.base.ref }}..${{ github.event.pull_request.head.sha }}"
+          for commit in $(git rev-list --reverse "$rev_list"); do
+
+            if M=$(git show -s --format='%ae' "$commit" | grep "noreply.github.com"); then
```
CohenArthur: I think it would be fine to just grep for `noreply.github.com` in the full commit, without necessarily just looking at the author's email format. This way, if people have a wrong Signed-off-by line, or Co-author someone with a noreply github email, it also gets caught
```suggestion
            if M=$(git show -s "$commit" | grep "noreply.github.com"); then
```

CohenArthur (approved):
thank you for taking care of this!

.github/workflows/commit-format.yml
dkm: Makes sense, I'll adjust

dkm: It works :sweat_smile: 
```
8a7757a3bf014160405d6324c6df8ee542075a98: KO => an invalid email was found:     ci: prevent @noreply.github.com emails
    "@noreply.github.com" in the commit log, which is not OK for upstream
```

PR #4579, opened 2026-06-05
Fix use after move in lexer

P-E-P:
The str variable was left in an unspecified but valid state after the prior move on the same line. This lead to a value of zero on some platforms and the prefix was considered empty.


CohenArthur (approved):
All hail our crab overlords who would have caught this before it got even committed

PR #4580, opened 2026-06-06
gccrs: Search lang_prelude when reaching module boundary during name resolution

Polygonalr:
Name resolution does not resolve language preludes properly within a module. This PR fixes that.

### DIFF
--- a/gcc/rust/resolve/rust-name-resolution-context.hxx
+++ b/gcc/rust/resolve/rust-name-resolution-context.hxx
@@ -29,6 +29,26 @@ namespace Rust {
 namespace Resolver2_0 {
 
 template <Namespace N>
+bool
+NameResolutionContext::should_search_prelude (
+  const typename ForeverStack<N>::Node *current_node,
+  const typename ForeverStack<N>::SegIterator &iterator,
+  const std::vector<ResolutionPath::Segment> &segments)
+{
+  // Check whether the current_node is a root node
+  if (current_node->is_root ())
+    return true;
+
+  // Check whether we're at the start of a module (we can't travel elsewhere
+  // from the start of a module)
+  if (is_start (iterator, segments)
+      && current_node->rib.kind == Rib::Kind::Module)
+    return true;
+
+  return false;
+}
+
+template <Namespace N>
 tl::optional<Rib::Definition>
 NameResolutionContext::resolve_path (
   ForeverStack<N> &stack, const ResolutionPath &path, ResolutionMode mode,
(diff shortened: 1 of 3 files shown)


powerboat9 (approved):
I'm pretty sure there's some better way we should be doing this, but I'm not quite sure what, so LGTM

gcc/rust/resolve/rust-name-resolution-context.hxx
```diff
@@ -383,7 +383,10 @@ NameResolutionContext::resolve_segments (
 		}
 	    }
 
-	  if (current_node->is_root () && !searched_prelude)
+	  if ((current_node->is_root ()
+	       || (is_start (iterator, segments)
+		   && current_node->rib.kind == Rib::Kind::Module))
+	      && !searched_prelude)
```
CohenArthur: I would suggest doing something like this which would be cleaner and easier to modify later:

```suggestion
	  if (should_search_prelude())
```

CohenArthur: You'll need to do something like `should_search_prelude(current_node, searched_prelude)` if we want to keep it as a simple `static void` function helper and not a full-on class function

gcc/testsuite/rust/compile/name_resolution28.rs
```diff
@@ -0,0 +1,13 @@
+#![feature(no_core)]
+#![no_core]
+
+pub mod lateresolve {
+    #![feature(no_core, lang_items)]
+    #![no_core]
```
CohenArthur: let's remove these

gcc/rust/resolve/rust-name-resolution-context.hxx
Polygonalr: Oh there's some `template` jank involved around here, see whether my current abstraction works out :v

gcc/rust/resolve/rust-name-resolution-context.hxx
```diff
@@ -383,7 +403,9 @@ NameResolutionContext::resolve_segments (
 		}
 	    }
 
-	  if (current_node->is_root () && !searched_prelude)
+	  if (!searched_prelude
+	      && this->template should_search_prelude<N> (current_node,
+							  iterator, segments))
```
CohenArthur:
```suggestion
	      && should_search_prelude<N> (current_node,
							  iterator, segments))
```

CohenArthur (approved):
Once the style comment is fixed I think this is good to go 👍 well done

gcc/rust/resolve/rust-name-resolution-context.hxx
Polygonalr: Could have sworn that there's an error if I compile my code without this, but whatever that might have caused it could have been resolved when I was cleaning up my code.

CohenArthur (approved):
LGTM

PR #4581, opened 2026-06-07
Handle assert with custom message

### LINKED ISSUE #4578: Implement second form of `assert!()` for `core`
In `core/src/str/mod.rs:94`:

```rust
    // 2. begin <= end
    assert!(
        begin <= end,
        "begin <= end ({} <= {}) when slicing `{}`{}",
        begin,
        end,
        s_trunc,
        ellipsis
    );

```

powerboat9:
Fixes #4578

### DIFF
--- a/gcc/rust/expand/rust-macro-builtins-log-debug.cc
+++ b/gcc/rust/expand/rust-macro-builtins-log-debug.cc
@@ -35,39 +35,45 @@ MacroBuiltin::assert_handler (location_t invoc_locus,
   Parser<MacroInvocLexer> parser (lex);
 
   auto last_token_id = macro_end_token (tt, parser);
-  bool has_error = false;
 
-  auto expanded_expr = try_expand_many_expr (parser, last_token_id,
-					     invoc.get_expander (), has_error);
-  if (expanded_expr.size () < 1)
+  // TODO: less args?
+  auto expr_to_assert = parser.parse_expr ();
+  if (!expr_to_assert.has_value ())
     {
       rust_error_at (invoc_locus,
 		     "macro requires a boolean expression as an argument");
       return AST::Fragment::create_error ();
     }
-  auto expr_to_assert = std::move (expanded_expr[0]);
-  expanded_expr.erase (expanded_expr.begin ());
 
-  if (expanded_expr.size () > 1)
-    {
-      rust_sorry_at (
-	invoc_locus,
-	"The second form of assert with a message is not supported yet");
-
-      return AST::Fragment::create_error ();
-    }
-
-  auto pending_invocations = check_for_eager_invocations (expanded_expr);
-  if (!pending_invocations.empty ())
-    return make_eager_builtin_invocation (BuiltinMacro::Assert, invoc_locus,
-					  invoc.get_delim_tok_tree (),
-					  std::move (pending_invocations));
+  // don't need to expand macros -- panic! will handle it
 
   AST::Builder b (invoc_locus);
 
   std::vector<std::unique_ptr<AST::TokenTree>> panic_tree;
   const_TokenPtr open = Token::make (TokenId::LEFT_PAREN, invoc_locus);
   panic_tree.push_back (std::make_unique<AST::Token> (std::move (open)));
+
+  if (parser.maybe_skip_token (COMMA))
+    {
+      bool needs_stringify = true;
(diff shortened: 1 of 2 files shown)


gcc/rust/expand/rust-macro-builtins-log-debug.cc
```diff
@@ -35,40 +35,46 @@ MacroBuiltin::assert_handler (location_t invoc_locus,
...
+	  auto token_tree = parser.parse_token_tree ();
+	  if (!token_tree.has_value ())
+	    {
+	      // error already emitted (?)
+	      return AST::Fragment::create_error ();
+	    }
+	  panic_tree.push_back (std::move (token_tree.value ()));
+	}
+      if (needs_stringify)
+	{
```
CohenArthur, line 72: Where does the `need_stringify` come from? Is it from somewhere in the Rust reference?

CohenArthur (commented):
Looks good!

CohenArthur:
Can you add a testcase for something of this form?

```rust

    // 2. begin <= end
    assert!(
        begin <= end,
        "begin <= end ({} <= {}) when slicing `{}`{}",
        begin,
        end,
        s_trunc,
        ellipsis
    );
```

this is one of the hindrances I ran into with `core`

gcc/rust/expand/rust-macro-builtins-log-debug.cc
powerboat9: https://doc.rust-lang.org/std/macro.assert.html

```rust
// the panic message for these assertions is the stringified value of the
// expression given.
assert!(true);

fn some_computation() -> bool {
    // Some expensive computation here
    true
}

assert!(some_computation());
```

powerboat9:
> Can you add a testcase for something of this form?
> 
> ```rust
>     // 2. begin <= end
>     assert!(
>         begin <= end,
>         "begin <= end ({} <= {}) when slicing `{}`{}",
>         begin,
>         end,
>         s_trunc,
>         ellipsis
>     );
> ```
> 
> this is one of the hindrances I ran into with `core`

I could, but the part of a test I added should already cover this (and does seem to prevent this bit of core from erroring). Note that the formatting itself isn't tested, just the compilation down to a `panic!` invocation (I don't think we support panicking yet).

gcc/rust/expand/rust-macro-builtins-log-debug.cc
CohenArthur: gotcha, thanks!

gcc/rust/expand/rust-macro-builtins-log-debug.cc
```diff
@@ -35,40 +35,46 @@ MacroBuiltin::assert_handler (location_t invoc_locus,
...
+	  // TODO: insert stringify invocation
+	  (void) 0;
```
CohenArthur, line 74: Maybe let's put a warning here? Not a full-on sorry at, but a warning saying this is a stub implementation of `assert`

CohenArthur: Also please open an issue for this, I think it shouldn't be too hard to implement even by a newcomer contributor

gcc/rust/expand/rust-macro-builtins-log-debug.cc
powerboat9: np

gcc/rust/expand/rust-macro-builtins-log-debug.cc
powerboat9: I'm thinking we'd run into issues making sure that `stringify!` actually calls the built-in `stringify!`, in all cases. Also, I'll open an issue, but I'm not sure I should insert a sorry warning -- it's a preexisting issue that only effects the message an `assert!` emits upon failure. If you do still want me to put it in, I will.

CohenArthur: I think it would be nice to have a `rust_warning_at` but not a `rust_sorry` as yeah it's not super important and very specific to `assert!` as you said

CohenArthur (approved):
I think we should merge this now as I need it for core and see how we can improve it in the future

powerboat9:
Opened #4635

PR #4582, opened 2026-06-08
intrinsic: Add min_align_of

nsvke:
This patch implements the 'min_align_of' compiler intrinsic. It resolves the minimum alignment of a given compile-time sized type and returns it as a size_type_node integer constant expression.

gcc/rust/ChangeLog:

	* backend/rust-compile-intrinsic.cc (generic_intrinsics): Add min_align_of to map.
	* backend/rust-intrinsic-handlers.cc (min_align_of_handler): New function.
	* backend/rust-intrinsic-handlers.h (min_align_of_handler): New declaration.

gcc/testsuite/ChangeLog:

	* rust/compile/min_align_of.rs: New test.

### DIFF
--- a/gcc/rust/backend/rust-intrinsic-handlers.cc
+++ b/gcc/rust/backend/rust-intrinsic-handlers.cc
@@ -1612,6 +1612,45 @@ sizeof_handler (Context *ctx, TyTy::FnType *fntype)
   return fndecl;
 }
 
+/**
+ * pub fn min_align_of<T>() -> usize;
+ */
+tree
+min_align_of_handler (Context *ctx, TyTy::FnType *fntype)
+{
+  // min_align_of has _zero_ parameters its parameter is the generic one
+  rust_assert (fntype->get_params ().size () == 0);
+
+  tree lookup = NULL_TREE;
+  if (check_for_cached_intrinsic (ctx, fntype, &lookup))
+    return lookup;
+
+  auto fndecl = compile_intrinsic_function (ctx, fntype);
+
+  // get the template parameter type tree fn min_align_of<T>();
+  rust_assert (fntype->get_num_substitutions () == 1);
+  auto &param_mapping = fntype->get_substs ().at (0);
+  const auto param_tyty = param_mapping.get_param_ty ();
+  auto resolved_tyty = param_tyty->resolve ();
+  tree template_parameter_type
+    = TyTyResolveCompile::compile (ctx, resolved_tyty);
+
+  enter_intrinsic_block (ctx, fndecl);
+
+  // BUILTIN min_align_of FN BODY BEGIN
+  tree align_expr
+    = build_int_cst (size_type_node, TYPE_ALIGN_UNIT (template_parameter_type));
+
+  auto return_statement
+    = Backend::return_statement (fndecl, align_expr, UNDEF_LOCATION);
+  ctx->add_statement (return_statement);
+  // BUILTIN min_align_of FN BODY END
+
+  finalize_intrinsic_block (ctx, fndecl);
+
+  return fndecl;
+}
+
 tree
 offset (Context *ctx, TyTy::FnType *fntype)
 {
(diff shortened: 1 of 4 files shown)


gcc/testsuite/rust/compile/min_align_of.rs
```diff
@@ -0,0 +1,26 @@
...
+pub fn get_u16_align() -> usize {
+    unsafe { min_align_of::<u16>() }
+}
+
+pub fn get_i32_align() -> usize {
+    unsafe { min_align_of::<i32>() }
+}
+
+// { dg-final { scan-tree-dump {D\.[0-9]+ = 2;} "gimple" } }
+// { dg-final { scan-tree-dump {D\.[0-9]+ = 4;} "gimple" } }
```
CohenArthur: What you can do here instead is make it an execution test, add the two alignments and then make the main function return the substraction of the sum of the alignments and the expected sum:

```rust
fn main() -> i32 {
    let align_u16 = get_u16_align();
    let align_u32 = get_u32_align();

    align_u16 + align_u32 - 6
}
```

CohenArthur (approved):
Looks good to me, @philberty if you have the time can you take a look? Note that we have https://forge.sourceware.org/gcc/gcc-TEST/pulls/161 going on at the same time so you might need to be careful with merge conflicts

gcc/testsuite/rust/compile/min_align_of.rs
nsvke: Thanks for the good tip!

gcc/rust/backend/rust-intrinsic-handlers.cc
```diff
@@ -1612,6 +1612,45 @@ sizeof_handler (Context *ctx, TyTy::FnType *fntype)
...
 }
 
+/**
+ * pub fn min_align_of<T>() -> usize;
+ */
+tree
+min_align_of_handler (Context *ctx, TyTy::FnType *fntype)
+{
+  // min_align_of has _zero_ parameters its parameter is the generic one
+  rust_assert (fntype->get_params ().size () == 0);
```
P-E-P, line 1622: We should probably have a step earlier that throws an error as well as a test to highlight that behavior.

gcc/testsuite/rust/execute/min_align_of.rs
```diff
@@ -0,0 +1,25 @@
...
+
+extern "rust-intrinsic" {
+    pub fn min_align_of<T>() -> usize;
+}
+
+fn main() -> i32 {
+    let align_u16 = get_u16_align() as i32;
+    let align_u32 = get_i32_align() as i32;
+
+    align_u16 + align_u32 - 6
```
P-E-P: Not sure what to think about this. We probably want something more explicit ?

gcc/testsuite/rust/execute/min_align_of.rs
```diff
@@ -0,0 +1,25 @@
+// { dg-additional-options "-fdump-tree-gimple" }
```
P-E-P: This test relies on ABI specific details, we do not want to run it on all targets.

P-E-P (approved):

gcc/testsuite/rust/execute/min_align_of.rs
CohenArthur: Is it ABI specific if the size is specified in the types? Aren't `u32` always supposed to be 32 bits long?

P-E-P: Their size yes but not their alignment ?

nsvke: I will update it to run on only `lp64`

gcc/rust/backend/rust-intrinsic-handlers.cc
nsvke: I followed the existing pattern from size_of. I'm not yet familiar with where intrinsic signatures are verified during type checking, but I will investigate it.

P-E-P (approved):
Looks great!

PR #4587, opened 2026-06-17
Enforce intrinsic signatures during the type checking

nsvke:
This patch adds a validation mechanism to the typechecker. This mechanism
    verifies intrinsic declarations against predefined signature rules. If
    a signature is incorrect, the compiler emits the proper diagnostic and
    prevents the malformed intrinsic from reaching the backend.
    
    gcc/rust/ChangeLog:
    
            * Make-lang.in: Add rust-hir-type-check-intrinsic.o.
            * typecheck/rust-hir-type-check-implitem.cc
            (TypeCheckTopLevelExternItem::visit): Check intrinsic signature.
            * typecheck/rust-hir-type-check-intrinsic.cc: New file.
            * typecheck/rust-hir-type-check-intrinsic.h: New file.
    
    gcc/testsuite/ChangeLog:
    
            * rust/compile/ctlz.rs: Fix incorrect intrinsic signature.
            * rust/compile/ctlz_nonzero.rs: Likewise.
            * rust/compile/cttz.rs: Likewise.
            * rust/compile/cttz_nonzero.rs: Likewise.
            * rust/compile/issue-4245.rs: Likewise.
            * rust/compile/torture/intrinsics-3.rs: Likewise.
            * rust/execute/torture/ctlz.rs: Likewise.
            * rust/execute/torture/ctlz_i16.rs: Likewise.
            * rust/execute/torture/ctlz_i32.rs: Likewise.
            * rust/execute/torture/ctlz_i64.rs: Likewise.
            * rust/execute/torture/ctlz_i8.rs: Likewise.
            * rust/execute/torture/ctlz_nonzero.rs: Likewise.
            * rust/execute/torture/ctlz_nonzero_i16.rs: Likewise.
            * rust/execute/torture/ctlz_nonzero_i32.rs: Likewise.
            * rust/execute/torture/ctlz_nonzero_i64.rs: Likewise.
            * rust/execute/torture/ctlz_nonzero_i8.rs: Likewise.
            * rust/execute/torture/ctlz_nonzero_u16.rs: Likewise.
            * rust/execute/torture/ctlz_nonzero_u32.rs: Likewise.
            * rust/execute/torture/ctlz_nonzero_u64.rs: Likewise.
            * rust/execute/torture/ctlz_nonzero_u8.rs: Likewise.
            * rust/execute/torture/ctlz_u16.rs: Likewise.
            * rust/execute/torture/ctlz_u32.rs: Likewise.
            * rust/execute/torture/ctlz_u64.rs: Likewise.
            * rust/execute/torture/ctlz_u8.rs: Likewise.
            * rust/execute/torture/cttz.rs: Likewise.
            * rust/execute/torture/cttz_nonzero.rs: Likewise.
            * rust/compile/intrinsic-checker.rs: New test.

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-intrinsic.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-intrinsic.cc
@@ -1,0 +1,788 @@
+// Copyright (C) 2026 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#include "rust-hir-type-check-intrinsic.h"
+#include "rust-diagnostics.h"
+#include "rust-intrinsic-values.h"
+#include "rust-system.h"
+#include "rust-tyty.h"
+
+namespace Rust {
+namespace Resolver {
+
+using IValue = Values::Intrinsics;
+using IRT = IntrinsicRuleType;
+
+const std::unordered_map<std::string, IntrinsicRules>
+  IntrinsicChecker::intrinsic_rules = {
+    // fn unreachable() -> !;
+    {IValue::UNREACHABLE, {0, {}, IRT::Never}},
+
+    // pub fn sqrtf32(x: f32) -> f32;
+    {IValue::SQRTF32, {0, {IRT::F32}, IRT::F32}},
+    // pub fn sqrtf64(x: f64) -> f64;
+    {IValue::SQRTF64, {0, {IRT::F64}, IRT::F64}},
+    // pub fn sinf32(x: f32) -> f32;
+    {IValue::SINF32, {0, {IRT::F32}, IRT::F32}},
+    // pub fn sinf64(x: f64) -> f64;
+    {IValue::SINF64, {0, {IRT::F64}, IRT::F64}},
+    // pub fn cosf32(x: f32) -> f32;
+    {IValue::COSF32, {0, {IRT::F32}, IRT::F32}},
+    // pub fn cosf64(x: f64) -> f64;
+    {IValue::COSF64, {0, {IRT::F64}, IRT::F64}},
+    // pub fn expf32(x: f32) -> f32;
+    {IValue::EXPF32, {0, {IRT::F32}, IRT::F32}},
+    // pub fn expf64(x: f64) -> f64;
+    {IValue::EXPF64, {0, {IRT::F64}, IRT::F64}},
+    // pub fn exp2f32(x: f32) -> f32;
+    {IValue::EXP2F32, {0, {IRT::F32}, IRT::F32}},
+    // pub fn exp2f64(x: f64) -> f64;
+    {IValue::EXP2F64, {0, {IRT::F64}, IRT::F64}},
+    // pub fn logf32(x: f32) -> f32;
+    {IValue::LOGF32, {0, {IRT::F32}, IRT::F32}},
+    // pub fn logf64(x: f64) -> f64;
+    {IValue::LOGF64, {0, {IRT::F64}, IRT::F64}},
+    // pub fn log10f32(x: f32) -> f32;
+    {IValue::LOG10F32, {0, {IRT::F32}, IRT::F32}},
+    // pub fn log10f64(x: f64) -> f64;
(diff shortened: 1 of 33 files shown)


gcc/testsuite/rust/compile/ctlz.rs
```diff
@@ -9,7 +9,7 @@ pub trait Sized {}
 pub trait Copy {}
 
 extern "rust-intrinsic" {
-    pub fn ctlz<T>(x: T) -> u32;
+    pub fn ctlz<T>(x: T) -> T;
```
P-E-P, line 12: Good catch, this was changed since 1.49.

gcc/rust/typecheck/rust-hir-type-check-intrinsic.h
```diff
@@ -0,0 +1,83 @@
...
+  MutPtrU8, // *mut u8
+  RefStaticStr, // &'static str
+  AssocTypePlaceholder,
+  // for discriminant_value<T>(v: &T) -> <T as DiscriminantKind>::Discriminant
+
+  TupleFirstGenericAndBool, // (T, bool)
+
+  Fn_MutPtrU8,	       // fn(*mut u8)
+  Fn_MutPtrU8MutPtrU8, // fn(*mut u8, *mut u8)
+};
```
P-E-P, line 57: Interesting approach, I wonder if it wouldn't be easier to "build" the expected signature type like we do with builtin macros within the compiler and then make sure both function match ?

Anyway this is just a random idea thrown out of thin air, I'm not asking you to reimplement that thing differently.

gcc/rust/typecheck/rust-hir-type-check-intrinsic.cc
```diff
@@ -0,0 +1,788 @@
...
+    {IValue::RUSTC_PEEK, {1, {IRT::FirstGeneric}, IRT::FirstGeneric}},
+    // pub fn type_name<T: ?Sized>() -> &'static str;
+    {IValue::TYPE_NAME, {1, {}, IRT::RefStaticStr}},
+    // pub fn forget<T: ?Sized>(_: T);
+    {IValue::FORGET, {1, {IRT::FirstGeneric}, IRT::Unit}},
+    // pub fn black_box<T>(mut dummy: T) -> T
+    {IValue::BLACK_BOX, {1, {IRT::FirstGeneric}, IRT::FirstGeneric}},
+};
+
+bool
```
P-E-P: Nit: You could introduce a new enum, I'm found of explicit values that document themselves. It would avoid the situation were someone misread the signature to "it must return true when it finds an error" at the call site.

You're already making the code a bit more explicit with that
```c++
check(/* [...] */) == false
```
at the call site, why not going even further ?

gcc/rust/typecheck/rust-hir-type-check-intrinsic.cc
```diff
@@ -0,0 +1,788 @@
...
+      result += " -> ";
+      result += to_string (rule->return_type);
+    }
+  return result;
+}
+
+std::string
+IntrinsicChecker::found_intrinsic_as_string (const TyTy::FnType *fntype)
+{
+  std::string result = "extern \"rust-intrinsic\" fn(";
```
P-E-P: Use a stringstream instead.

gcc/rust/typecheck/rust-hir-type-check-intrinsic.cc
```diff
@@ -0,0 +1,788 @@
...
+  rust_unreachable ();
+}
+
+std::string
+IntrinsicChecker::expected_intrinsic_as_string (const TyTy::FnType *fntype)
+{
+  auto it = intrinsic_rules.find (fntype->get_identifier ());
+  rust_assert (it != intrinsic_rules.end ());
+  auto rule = &it->second;
+  std::string result = "extern \"rust-intrinsic\" fn(";
```
P-E-P: Likewise with the stringstream.

P-E-P (approved):
I'm not 100% sold on this approach as I feel it duplicates a lot of work (you had to implement all those "to_string" functions and we get another representation for types).

However, work has been done and is here, ready to be used. This is a low priority issue and it works as is, it doesn't make sense to work on another way which could very well be unsatisfactory. I'm willing to merge this once the few minor comments have been addressed.

gcc/rust/typecheck/rust-hir-type-check-intrinsic.h
nsvke: I actually thought about doing it that way initially, but decided to go with this approach as it was the fastest solution for a low-priority issue. I agree it would definitely be a much more natural fit, though. Would you like me to open an issue to document this technical debt?

nsvke:
I've addressed the minor comments. To be honest, I'm not 100% sold on it either 😅. But I agree it's best to not sink more time into this low-priority issue.

gcc/rust/typecheck/rust-hir-type-check-intrinsic.h
P-E-P: @nsvke  Yes please open an issue!

P-E-P (approved):

philberty:
yeah i think there are a few issues here i get what its trying to do though. But since i rebased ontop of this its breaking my GATS work i'm gonna have to make it a bit more permissive.. but also the to_string implementations are not needed here

PR #4588, opened 2026-06-17
Extern crate loading improvements

P-E-P:
Extern crates such as core were not loaded correctly. Among the issues:
- modules were not handled
- reexported items were not handled

Changes the metadata dump from an HIR dump which does not contain any UseDeclaration to an AST dump. This lead to some problems with `crate::XXXX` path which were no handled from the external crate's POV. To avoid that we now name resolve the external crates on their own alone and merge them later within the current crate.

We now dump the AST instead of the HIR for metadata for the following reasons:
- AST dump accuracy is garanteed by proc macros
- We do not have to lower use declaration to hir

### DIFF
--- a/gcc/rust/ast/rust-ast-builder.cc
+++ b/gcc/rust/ast/rust-ast-builder.cc
@@ -182,10 +182,10 @@ Builder::type_path_segment (std::string seg) const
 }
 
 std::unique_ptr<TypePathSegment>
-Builder::type_path_segment (LangItem::Kind lang_item) const
+Builder::type_path_segment (LangItem::Kind lang_item, std::string &name) const
 {
   return std::unique_ptr<TypePathSegment> (
-    new TypePathSegment (lang_item, loc));
+    new TypePathSegment (lang_item, PathIdentSegment (name, loc), loc));
 }
 
 std::unique_ptr<TypePathSegment>
@@ -196,11 +196,12 @@ Builder::type_path_segment_generic (std::string seg, GenericArgs args) const
 }
 
 std::unique_ptr<TypePathSegment>
-Builder::type_path_segment_generic (LangItem::Kind lang_item,
+Builder::type_path_segment_generic (LangItem::Kind lang_item, std::string &name,
 				    GenericArgs args) const
 {
   return std::unique_ptr<TypePathSegment> (
-    new TypePathSegmentGeneric (lang_item, args, loc));
+    new TypePathSegmentGeneric (lang_item, PathIdentSegment (name, loc), args,
+				loc));
 }
 
 std::unique_ptr<Type>
@@ -228,11 +229,11 @@ Builder::single_generic_type_path (std::string type, GenericArgs args) const
 }
 
 std::unique_ptr<Type>
-Builder::single_generic_type_path (LangItem::Kind lang_item,
+Builder::single_generic_type_path (LangItem::Kind lang_item, std::string &name,
 				   GenericArgs args) const
 {
   auto segments = std::vector<std::unique_ptr<TypePathSegment>> ();
-  segments.emplace_back (type_path_segment_generic (lang_item, args));
+  segments.emplace_back (type_path_segment_generic (lang_item, name, args));
 
   return std::unique_ptr<Type> (new TypePath (std::move (segments), loc));
 }
@@ -272,9 +273,9 @@ Builder::type_path (std::string type) const
 }
 
 TypePath
-Builder::type_path (LangItem::Kind lang_item) const
+Builder::type_path (LangItem::Kind lang_item, std::string &name) const
 {
-  return type_path (type_path_segment (lang_item));
+  return type_path (type_path_segment (lang_item, name));
 }
 
 std::unique_ptr<Type>
--- a/gcc/rust/ast/rust-ast-builder.cc
+++ b/gcc/rust/ast/rust-ast-builder.cc
@@ -182,8 +182,10 @@ Builder::type_path_segment (std::string seg) const
 }
 
 std::unique_ptr<TypePathSegment>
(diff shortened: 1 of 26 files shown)


gcc/rust/expand/rust-extern-crate-loader.cc
```diff
@@ -0,0 +1,50 @@
...
+    {
+      Session &session = Session::get_instance ();
+      auto crate
+	= session.load_extern_crate (extern_crate.get_referenced_crate (),
+				     extern_crate.get_locus ());
+      if (crate)
+	{
+	  loaded_crates.push_back (std::move (crate.value ()));
+	}
+    }
```
CohenArthur: could we do something like

```c++
if (extern_crate.references_self())
    return;

/* rest of the logic */
```

gcc/rust/resolve/rust-default-resolver.cc
```diff
@@ -442,7 +442,11 @@ void
 DefaultResolver::visit_extern_crate (AST::ExternCrate &extern_crate,
 				     AST::Crate &crate, CrateNum num)
 {
-  visit (crate);
+  // We do *NOT* visit the crate because loaded crates are resolved
+  // independently.
+#if 0
+    visit (crate);
+#endif
```
CohenArthur: I would just remove this here, I think your comment is good and clear

gcc/rust/resolve/rust-name-resolution-context.cc
```diff
@@ -384,6 +384,36 @@ NameResolutionContext::scoped (Rib::Kind rib_kind, Namespace ns,
...
+{
+  auto merge_fstack = [&] (auto &stack, auto &other_stack) {
+    auto node = stack.dfs_node (stack.root, at);
+    if (node)
+      {
+	auto &extern_crate_node = node.value ();
+	for (auto kv : other_stack.root.children)
+	  {
+	    auto link = kv.first;
+	    auto child = kv.second;
```
CohenArthur, line 401: Since we are in C++14 now, can we do `for (auto [link, child]: other_stack.root.children)`?

gcc/rust/resolve/rust-name-resolution-context.cc
```diff
@@ -384,6 +384,36 @@ NameResolutionContext::scoped (Rib::Kind rib_kind, Namespace ns,
     }
 }
 
+void
+NameResolutionContext::merge (NameResolutionContext &other, NodeId at)
+{
+  auto merge_fstack = [&] (auto &stack, auto &other_stack) {
+    auto node = stack.dfs_node (stack.root, at);
+    if (node)
```
CohenArthur, line 395: Should we do an assertion here? I don't know if we want to silently ignore definitions if the node doesn't exist in the target ForeverStack

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
```diff
@@ -192,7 +192,11 @@ TopLevel::visit_extern_crate (AST::ExternCrate &extern_crate, AST::Crate &crate,
 	mappings.insert_bang_proc_macro_def (macro);
     }
 
-  visit (crate);
+    // We do *NOT* visit the crate because loaded crates are resolved
+    // independently.
+#if 0
+    visit (crate);
+#endif
```
CohenArthur: Likewise I think this should be removed as your comment is clear enough to me

gcc/rust/rust-session-manager.h
```diff
@@ -433,7 +433,46 @@ struct Session
...
+
+  struct AlreadyLoadedError
+  {
+  public:
+    enum Kind
+    {
+      ALREADY_LOADED,
+      FAILED_TO_LOCATE,
+      COLLISION,
+    } kind;
```
CohenArthur: Make this an enum class so there's no collision with other frontends or anything of the sort - `COLLISION` in particular could be a conflicting identifier

gcc/rust/ast/rust-ast-builder.cc
```diff
@@ -272,9 +273,9 @@ Builder::type_path (std::string type) const
 }
 
 TypePath
-Builder::type_path (LangItem::Kind lang_item) const
+Builder::type_path (LangItem::Kind lang_item, std::string &name) const
```
CohenArthur: Since the builder is here for convenience functions and to make things like the `Derive`s nicer, I would keep the old interface

```c++
Builder::type_path (LangItem::Kind lang_item)
```

and make the builder figure out the path from the given lang item kind.

You could also keep a second `BiMap` in the `LangItem` class of "public names" for each lang item kind. I think could be useful for future things

CohenArthur: (and likewise for the other Builder methods)

CohenArthur (approved):
Great work :) well done

gcc/rust/resolve/rust-name-resolution-context.cc
P-E-P: Structured bindings were introduced in cpp17.

gcc/rust/resolve/rust-name-resolution-context.cc
P-E-P: There is a few edge cases were we do want to ignore it silently (proc macros)

gcc/rust/ast/rust-ast-builder.cc
P-E-P: I do not like that Bimap but I reckon this is temporary. Retrieving & recording the actual type segment would mean moving the lang item recognition routine within the expansion loop, this would make this bit a lot more complex than what it currently is all for something that will get rid off soon.

CohenArthur (approved):
Looks perfect, well done

PR #4590, opened 2026-06-18
nr: Properly insert macros in the current rib

CohenArthur:
Previously, only decl macros (macros 2.0) were being inserted in the current rib, as it was believed that macros only needed to exist within the textual scope. This is however wrong, and macros need to be inserted in both the current rib and the textual scope.

gcc/rust/ChangeLog:

	* resolve/rust-toplevel-name-resolver-2.0.cc (TopLevel::visit): Insert macros in the current rib always.

gcc/testsuite/ChangeLog:

	* rust/compile/name_resolution29.rs: New test.

### DIFF
--- a/gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
@@ -208,10 +208,17 @@ is_macro_export (AST::MacroRulesDefinition &def)
 void
 TopLevel::visit (AST::MacroRulesDefinition &macro)
 {
-  // we do not insert macros in the current rib as that needs to be done in the
-  // textual scope of the Early pass. we only insert them in the root of the
-  // crate if they are marked with #[macro_export]. The execption to this is
-  // macros 2.0, which get resolved and inserted like regular items.
+  // Macros are inserted in both the current rib and the textual scope within
+  // the Early pass. We also insert them at the root of the crate if they are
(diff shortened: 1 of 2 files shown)


powerboat9 (approved):

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
```diff
@@ -229,8 +236,7 @@ TopLevel::visit (AST::MacroRulesDefinition &macro)
 	}
     }
 
-  if (macro.get_kind () == AST::MacroRulesDefinition::MacroKind::DeclMacro)
-    insert_or_error_out (macro.get_rule_name (), macro, Namespace::Macros);
+  insert_or_error_out (macro.get_rule_name (), macro, Namespace::Macros);
```
powerboat9, line 239: Actually, I think this should be gated behind `is_macro_export`

https://doc.rust-lang.org/reference/macros-by-example.html#r-macro.decl.scope.macro_export.path-based

powerboat9 (changes requested):

gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
CohenArthur: yeah I think I did something wrong here with the way my `core` was setup. I think I was using a much more recent `stdarch` submodule than it should have been

PR #4591, opened 2026-06-21
gccrs: Support function-scope drops on normal function exit

Lishin1215:
This PR adds function-scope drop handling for normal function exits only.
Early exits such as `return`, `break`, and `continue` are left for follow-up work.

It covers two missing cases:

1. Unit tail expressions
```rust
fn foo() {}

fn f() {
    let _x = Droppable;
    foo()
}
```

2. Function parameters
```rust
fn f(x: Droppable) {}
```

### DIFF
--- a/gcc/rust/backend/rust-compile-base.cc
+++ b/gcc/rust/backend/rust-compile-base.cc
@@ -21,6 +21,7 @@
 #include "rust-compile-stmt.h"
 #include "rust-compile-expr.h"
 #include "rust-compile-drop.h"
+#include "rust-compile-drop-builder.h"
 #include "rust-compile-fnparam.h"
 #include "rust-compile-var-decl.h"
 #include "rust-compile-type.h"
@@ -719,7 +720,7 @@ HIRCompileBase::compile_function_body (tree fndecl,
 	  // just add the stmt expression
 	  ctx->add_statement (return_value);
 
-	  CompileDrop::emit_current_scope_drop_calls (ctx);
+	  CompileDrop (ctx).emit_current_scope_drop_calls ();
 
 	  // now just return unit expression
 	  tree unit_expr = unit_expression (locus);
@@ -825,6 +826,7 @@ HIRCompileBase::compile_function (
   // setup the params
   TyTy::BaseType *tyret = fntype->get_return_type ();
   std::vector<Bvariable *> param_vars;
+  std::vector<DropCandidate> param_drop_candidates;
   if (self_param)
     {
       rust_assert (fntype->is_method ());
@@ -860,6 +862,11 @@ HIRCompileBase::compile_function (
       const HIR::Pattern &param_pattern = referenced_param.get_param_name ();
       ctx->insert_var_decl (param_pattern.get_mappings ().get_hirid (),
 			    compiled_param_var);
+
+      if (CompileDrop (ctx).type_has_drop_impl (param_tyty))
+	param_drop_candidates.emplace_back (
+	  param_pattern.get_mappings ().get_hirid (),
+	  param_pattern.get_locus ());
     }
 
   if (!Backend::function_set_parameters (fndecl, param_vars))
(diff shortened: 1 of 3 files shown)


gcc/rust/backend/rust-compile-base.cc
```diff
@@ -860,6 +861,11 @@ HIRCompileBase::compile_function (
       const HIR::Pattern &param_pattern = referenced_param.get_param_name ();
       ctx->insert_var_decl (param_pattern.get_mappings ().get_hirid (),
 			    compiled_param_var);
+
+      if (CompileDrop::type_has_drop_impl (ctx, param_tyty))
+	param_drop_candidates.emplace_back (
+	  param_pattern.get_mappings ().get_hirid (),
+	  param_pattern.get_locus ());
     }
```
Lishin1215, line 870: I collect parameter drop candidates while compiling parameters, then register them after the function body block is created because drop candidates are stored on the current block scope.
This also keeps locals dropping before parameters on normal function exit.

Does this approach make sense to you? 
Thank you!

philberty:
I'm gonna have to catchup on your work sorry your probably more ahead than i think

gcc/rust/backend/rust-compile-base.cc
CohenArthur: yes it does! can you check that this is how rustc does it (I assume you checked already) and put a link to the reference or the source code here? feel free to directly link to rustc source code in a comment

CohenArthur (approved):
Good work :)

gcc/rust/backend/rust-compile-base.cc
Lishin1215: I checked rustc 1.49.0.
I believe the core result/order is similar, but the structure is a little different.

---
The similar part is,
parameter drops are registered before body locals are lowered, so the final drop order becomes 
body locals first, then parameters.

  - In rustc,
  
  1. `construct_fn` creates an `arg_scope` for function parameters
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/mod.rs#L608-L628
  
  2. adds the parameter drops to that scope using `schedule_drop`:
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/mod.rs#L882-L888
  
  3. Body/local scopes are created while lowering the block, and local drops are scheduled into 
  those local scopes:
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/mod.rs#L943-L944
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/block.rs#L28-L43
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/block.rs#L104-L132
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/matches/mod.rs#L357-L375
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/matches/mod.rs#L561-L565

  4. When a scope is popped, rustc emits drops for the current top scope before popping it, 
  so inner body/local scopes are dropped before the outer `arg_scope`
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/block.rs#L200-L204
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/scope.rs#L539-L542
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/scope.rs#L556-L568
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/scope.rs#L656-L672

  Both parameter and local drops go through `schedule_drop`, which stores the drop 
in the matching scope's drop list:
https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/scope.rs#L840-L852

---
The difference part is,

1. Registration timing

    In gccrs, drop candidates are stored on the current backend block scope, so 
    parameter drop candidates are collected temporarily and registered after the function block is created. 
    
    In rustc, `arg_scope` already exists before `args_and_body`, so parameter drops can be 
    scheduled directly into that scope.

2. Scope structure

    In gccrs, function parameters and body locals are currently stored in the same
    drop candidate list for the current block. 
    
    In rustc, parameters and body locals are in different nested scopes: 
    parameters are in the outer `arg_scope`, while body locals are in inner body/local scopes.

CohenArthur: okay, then this is something we need to do as well. We need to get the drop order as close to rustc's as possible even if it means making massive changes. This doesn't need to be done in this PR, but this is something I would like you to work on for your GSoC. Can you open an issue for this with all the links you've put here? This is great documentation and will help tremendously

Lishin1215: No problem, I just opened an issue for this:
https://github.com/Rust-GCC/gccrs/issues/4712
And I will work on it!

PR #4594, opened 2026-06-22
Support GATS

### LINKED ISSUE #4293: Support GAT's
```rust
trait Foo {
    type Bar<T>;
    
    fn foo<T>(self, value: T) -> Self::Bar<T>;
    fn baz<T>(self, value: T) -> Self::Bar<T>;
}

impl Foo for i32 {
    type Bar<T> = T;
    
    fn foo<T>(self, value: T) -> Self::Bar<T> {
        value
    }
    
    fn baz<T>(self, value: T) -> Self::Bar<T> {
        value
    }
}

fn main() {
   let a = 15; 
   
   let b = a.foo::<i8>(14i8);
   let c = a.baz::<u32>(32u32);
   
   dbg!(a, b, c);
}
```

From zulip @CohenArthur  https://gcc-rust.zulipchat.com/#narrow/channel/281658-compiler-development/topic/GAT's/near/556945336

philberty:
This patch adds in GATs using proper projection types instead of mutating associated
 type placeholders during resolution which rapidly becomes very messy.

The simple case is trait impl handling, entering a trait impl block we first
walk the associated type items and build an ImplTraitFrame for that impl. This
frame records the trait, the impl self type and the mapping from trait associated
items to the impl associated type values. While typechecking items in that impl
we can then look through the active impl trait context to resolve projections
such as `<Self as Trait>::Item`.
    
The more complex cases need lazy evaluation. A projection can appear outside the
impl body, inside a where-clause binding, behind another projection, or after
substitution has rebound Self and the trait arguments. For those cases we keep
associated types as ProjectionType values carrying the trait ref, associated
item, Self type and substitutions, then normalize them when the type is actually
used. This avoids global associated type mutation and lets generic associated
type arguments compose through substitution, unification and backend lowering.

Fixes Rust-GCC#4293


gcc/rust/ast/rust-ast-collector.cc
```diff
@@ -639,7 +639,10 @@ TokenCollector::visit (TypePathSegmentGeneric &segment)
   //    `<` `>`
   //    | `<` ( GenericArg `,` )* GenericArg `,`? `>`
   describe_node (std::string ("TypePathSegmentGeneric"), [this, &segment] () {
-    auto ident_segment = segment.get_ident_segment ();
+    auto ident_segment
+      = segment.is_lang_item ()
+	  ? PathIdentSegment ("LANG_ITEM", segment.get_locus ())
```
P-E-P, line 644: I've changed this within #4588, we should not emit "lang item" but the real segment name. You do not need to change anything, I'll do it later when #4588 will get merged.

gcc/rust/typecheck/rust-hir-impl-trait-context.h
```diff
@@ -0,0 +1,51 @@
+// Copyright (C) 2021-2026 Free Software Foundation, Inc.
```
P-E-P, line 1: Copyright  year should be 2026

PR #4595, opened 2026-06-23
Move a test into `compile/macros/mbe`

powerboat9:
This test should go in the folder for "macros by example", not the top level rust compile test folder. Follows up on #4040


philberty (approved):
LGTM

PR #4599, opened 2026-06-24
backend: Refactor dynamic object fat pointers and vtable generation

nsvke:
Previously, the compiler incorrectly embedded the vtable as an array directly inside the fat pointer. Furthermore, it lacked essential trait object metadata such as `size`, `align` and `drop_in_place`.

Now, the fat pointer is strictly fixed to 2 words. The vtable is generated as a separate global static struct and it's structure correctly includes `drop_in_place`, `size` and `align` fields, followed by trait methods. Also implemented a caching mechanism in the compilation context to prevent duplicate vtable generation and linker conflicts for the same Type-Trait combinations.

gcc/rust/ChangeLog:

	* backend/rust-compile-context.h (class Context): Add insert_vtable and lookup_vtable methods.
	* backend/rust-compile-expr.cc (CompileExpr::get_fn_addr_from_dyn): Change access method array access to struct field access.
	* backend/rust-compile-type.cc (TyTyResolveCompile::create_dyn_obj_record): Create vtable type as a record instead of an array.
	* backend/rust-compile.cc (HIRCompileBase::coerce_to_dyn_object): Change vtable structure.

gcc/testsuite/ChangeLog:

	* rust/execute/torture/dyn_object.rs: New test.

### DIFF
--- a/gcc/rust/backend/rust-compile.cc
+++ b/gcc/rust/backend/rust-compile.cc
@@ -17,6 +17,7 @@
 // <http://www.gnu.org/licenses/>.
 
 #include "rust-compile.h"
+#include "fold-const.h"
 #include "rust-compile-item.h"
 #include "rust-compile-implitem.h"
 #include "rust-hir-type-bounds.h"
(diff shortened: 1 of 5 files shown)


gcc/rust/backend/rust-compile.cc
```diff
@@ -193,39 +194,81 @@ HIRCompileBase::coerce_to_dyn_object (tree compiled_ref, TyTy::BaseType *actual,
...
   auto probed_bounds_for_receiver = Resolver::TypeBoundsProbe::Probe (actual);
   tree address_of_compiled_ref = null_pointer_node;
   if (!actual->is_unit ())
     address_of_compiled_ref = address_expression (compiled_ref, locus);
 
-  std::vector<tree> vtable_ctor_elems;
-  std::vector<unsigned long> vtable_ctor_idx;
-  unsigned long i = 0;
-  for (auto &bound : ty->get_object_items ())
+  std::string vtable_name
```
nsvke: I am not sure about this mangling method. We might need a better one.

gcc/rust/backend/rust-compile.cc
```diff
@@ -193,39 +194,81 @@ HIRCompileBase::coerce_to_dyn_object (tree compiled_ref, TyTy::BaseType *actual,
...
+					   false, vtable_ctor_elems, -1, locus);
+
+      // https://github.com/rust-lang/rust/blob/e1884a8e3c3e813aada8254edfa120e85bf5ffca/compiler/rustc_mir/src/interpret/traits.rs#L17
+      Bvariable *vtable_bvar
+	= Backend::global_variable (vtable_name, vtable_name,
+				    TREE_TYPE (vtable_ctor), false, true, true,
+				    locus);
+      Backend::global_variable_set_init (vtable_bvar, vtable_ctor);
+      ctx->push_var (vtable_bvar);
+      ctx->insert_vtable (vtable_name, vtable_bvar);
```
nsvke: I used a string as the key here. I know it should be lighter, 
but `ty->get_ty_ref()` changes every time.

gcc/rust/backend/rust-compile.cc
```diff
@@ -193,39 +194,81 @@ HIRCompileBase::coerce_to_dyn_object (tree compiled_ref, TyTy::BaseType *actual,
...
+
+      vtable_decl = Backend::var_expression (vtable_bvar, locus);
 
-  tree vtable_ctor
-    = Backend::array_constructor_expression (TREE_TYPE (vtable_field),
-					     vtable_ctor_idx, vtable_ctor_elems,
-					     locus);
+      DECL_ARTIFICIAL (vtable_decl) = 1;
+      TREE_READONLY (vtable_decl) = 1;
+      DECL_IGNORED_P (vtable_decl) = 1;
```
nsvke, line 267: I based this on the [C++ vtable approach.](https://github.com/Rust-GCC/gccrs/blob/568ff0de54e269986ac3caa47ce409ca6bce95af/gcc/cp/class.cc#L827).

gcc/rust/backend/rust-compile.cc
philberty: There is the type hash canon we followed some of that from the cfront end family so in TyTyResolveCompile we try to always normalize the same types back to a single tree.. you should try to check the hash on the 

tree dynamic_object = TyTyResolveCompile::compile (ctx, &r);

should in theory do it

gcc/rust/backend/rust-compile.cc
philberty: yeah ty ref isnt designed for that its to do with generics and its a bit messy now but were kind of stuck with it

nsvke: Using a `std::pair` of `actual->get_ty_ref()` and `TREE_HASH(dynamic_object)` as the cache key seems to be working good.

gcc/rust/backend/rust-compile.cc
nsvke: I updated the vtable mangling to something like `__R_vtable_{actual->get_ty_ref()}_{TREE_HASH(dynamic_object)}` and it successfully prevents linker collisions. However, I noticed a separate frontend issue: when structs with the exact same name in different modules are coerced via the same trait, their method addresses get mixed up during trait resolution. I've opened Issue #4600  to track this.

gcc/rust/backend/rust-compile.cc
philberty: can you try and just use TREE_HASH dynamic_object? does that work

philberty: or even worth double checking the pointer value for the dynmaic object tree because our resolve type should be doing canonicalization to the same tree

philberty:
I think we should merge this but lets just see if we can avoid using ty id's in the hash here first

gcc/rust/backend/rust-compile.cc
nsvke: but it causes cache conflicts when two different concrete types coerce to the same trait:

StructX, dyn [TraitC<Self>]
`actual->get_ty_ref()` (**41**), `TYPE_HASH(dynamic_object)` (**861**), `TREE_HASH(dynamic_object)` (**64800**), addr (**0x7b5f0eb0fd20**)

StructX, dyn [TraitA<Self>]
`actual->get_ty_ref()` (**41**), `TYPE_HASH(dynamic_object)` (**874**), `TREE_HASH(dynamic_object)` (**71144**), addr (**0x7b5f0eb115e8**)

StructX, dyn [TraitA<Self>]
`actual->get_ty_ref()` (**41**), `TYPE_HASH(dynamic_object)` (**874**), `TREE_HASH(dynamic_object)` (**71144**), addr (**0x7b5f0eb115e8**)

StructY, dyn [TraitA<Self>]
`actual->get_ty_ref()` (**45**), `TYPE_HASH(dynamic_object)` (**874**), `TREE_HASH(dynamic_object)` (**71144**), addr (**0x7b5f0eb115e8**)

StructX, dyn [TraitB<Self>]
`actual->get_ty_ref()` (**41**), `TYPE_HASH(dynamic_object)` (**909**), `TREE_HASH(dynamic_object)` (**89376**), addr (**0x7b5f0eb15d20**)

philberty: well looks like TREE_HASH works there on its own because the unique set is:

dyn TraitC 64800

dyn TraitA 71144

dyn trait B 89376

really dont rely on ty ref

nsvke: But StructX -> dyn TraitA gives 71144 and StructY -> dyn TraitA also gives 71144. This shouldn't happen because the vtable contains type-specific information. Therefore, do you think the TYPE_HASH of compiled_ref's TREE_TYPE is a reliable value instead of ty ref? According to my tests, it seems to work good. Does using a pair in the format of (concrete type hash, dyn trait hash) make sense to you?

philberty: yeah that will work nice

philberty (approved):
LGTM

PR #4602, opened 2026-06-24
gccrs: Evaluate non-unit tail expressions before drops

Lishin1215:
I initially opened this branch to add coverage for non-unit tail expression testcase.

The new test exposed that a non-unit tail call could be evaluated after local drops, so this PR also includes the fix to store the tail expression result before emitting drops.

For example:
```rust
fn foo() -> i32 {
    0
}

fn f() -> i32 {
    let _x = Droppable;
    foo()
}
```
The expected order is to evaluate `foo()` first, then drop `_x`, then return the stored result.

---
gcc/rust/ChangeLog:

	* backend/rust-compile-base.cc (HIRCompileBase::compile_function_body): Store non-unit tail expression results before emitting current scope drops.

gcc/testsuite/ChangeLog:

	* rust/execute/drop-function-scope-non-unit-tail.rs: New test.

### DIFF
--- a/gcc/rust/backend/rust-compile-base.cc
+++ b/gcc/rust/backend/rust-compile-base.cc
@@ -708,10 +708,30 @@ HIRCompileBase::compile_function_body (tree fndecl,
 	  return_value = coercion_site (id, return_value, actual, expected,
 					lvalue_locus, rvalue_locus);
 
+	  /* Save the non-unit tail expression result before emitting scope
+	    drops, so a tail call like foo() is evaluated before locals are
+	    dropped.  Conceptually, this changes lowering from:
+
+	      drop (_x);
+	      return foo ();
+
+	    to:
+
+	      ret_slot = foo ();
+	      drop (_x);
+	      return ret_slot; */
+	  fncontext fnctx = ctx->peek_fn ();
+	  tree result_reference
+	    = Backend::var_expression (fnctx.ret_addr, lvalue_locus);
+	  tree assignment = Backend::assignment_statement (result_reference,
+							   return_value, locus);
+	  ctx->add_statement (assignment);
+
 	  CompileDrop (ctx).emit_current_scope_drop_calls ();
 
+	  result_reference = Backend::var_expression (fnctx.ret_addr, locus);
 	  tree return_stmt
-	    = Backend::return_statement (fndecl, return_value, locus);
+	    = Backend::return_statement (fndecl, result_reference, locus);
 	  ctx->add_statement (return_stmt);
 	}
       else
--- a/gcc/testsuite/rust/execute/drop-function-scope-non-unit-tail.rs
+++ b/gcc/testsuite/rust/execute/drop-function-scope-non-unit-tail.rs
@@ -1,0 +1,46 @@
+// { dg-output "f\r*\nd\r*\n" }
+// { dg-additional-options "-w" }
+#![feature(no_core)]
+#![feature(lang_items)]
+#![no_core]
+
+extern "C" {
+    fn printf(s: *const i8, ...);
+}
+
+#[lang = "sized"]
+pub trait Sized {}
+
+#[lang = "drop"]
+pub trait Drop {
+    fn drop(&mut self);
+}
+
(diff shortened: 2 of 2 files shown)


gcc/rust/backend/rust-compile-base.cc
```diff
@@ -708,10 +708,18 @@ HIRCompileBase::compile_function_body (tree fndecl,
 	  return_value = coercion_site (id, return_value, actual, expected,
 					lvalue_locus, rvalue_locus);
 
+	  fncontext fnctx = ctx->peek_fn ();
+	  tree result_reference
```
Lishin1215, line 724: I store the non-unit tail expression result before emitting scope drops so that 
a tail call like `foo()` is evaluated before locals are dropped. 
The function then returns this saved value instead of returning the original tail expression directly.

This changes the lowering from

`drop(_x); `
`return foo();`

to

`ret_slot = foo(); `
`drop(_x); `
`return ret_slot;`

Does this approach look reasonable?

philberty: Yeah i think for the initial drop support thats a sensible start.. but i think when it comes to emitting the drop calls i think its gonna get a little more complex

because i am imagining something like:

```
let a = Foo() // which is dropable
let b = a; // b might get caught as a dropable twice maybe
```

But i think there is also a rule in rust where things can be moved when passing by value which will change whats dropable at that point.. so i think i need to refresh my memory here 

but i think it all starts from this initial tracking

philberty: I think you will also need to look how the go front-end implements defer because it uses a try finally block in gimple so really it gives that final block that we know things should be droped and we dont miss things

Lishin1215: Hi, thank you for the feedback and for confirming that the current approach is reasonable.

For the move scenarios (the example you gave), I had originally planned to handle them at a later stage of my GSoC work (week 8). I am now first working on function-scope drops and explicit-return/continue/break drops.

For the Go frontend `defer` implementation, I understand the idea that the current approach of emitting drops directly should not be the long-term solution.

I’ll take a closer look at how the try/finally lowering works over the weekend, and I plan to work on this after finishing the explicit return case. Do you think this makes sense?

Thanks again!

CohenArthur: I think it is fine to implement things piece by piece and add more as you progress through your GSoC. The current approach seems good to me, and I agree that you will need to be careful of such edge cases eventually and you'll need to cover them before the end of GSoC - but for now this is already really good and better than what we had :)

gcc/rust/backend/rust-compile-base.cc
```diff
@@ -708,10 +708,18 @@ HIRCompileBase::compile_function_body (tree fndecl,
...
+	    = Backend::var_expression (fnctx.ret_addr, lvalue_locus);
+	  tree assignment = Backend::assignment_statement (result_reference,
+							   return_value, locus);
```
CohenArthur, line 727: Can you add a comment here explaining why you are doing this? basically exactly what you said in your github comment above would be really good:

```
I store the non-unit tail expression result before emitting scope drops so that
a tail call like foo() is evaluated before locals are dropped.
The function then returns this saved value instead of returning the original tail expression directly.

This changes the lowering from

drop(_x); 
return foo();

to

ret_slot = foo(); 
drop(_x); 
return ret_slot;
```

CohenArthur (approved):
Thank you!

gcc/rust/backend/rust-compile-base.cc
philberty: yeah thats probably better tbh

gcc/rust/backend/rust-compile-base.cc
Lishin1215: Thanks, done. 
I added the comment, resolved the rebase conflict, and pushed the updated branch.

CohenArthur (approved):
Thank you @Lishin1215 great work!

PR #4620, opened 2026-06-26
Add write_bytes and arith_offset

nsvke:
This patch implements `arith_offset` and `write_bytes` compiler intrinsics.

* `write_bytes`: Lowered to `__builtin_memset`.
* `arith_offset`: Lowered using `POINTER_PLUS_EXPR`.

`rustc` behavior:
- `arith_offset` -> https://godbolt.org/z/j631bMbT7
- `write_bytes` -> https://godbolt.org/z/q1K1Txc5v


philberty (approved):
LGTM

PR #4621, opened 2026-06-26
gccrs: Handle explicit-return drops with try-finally cleanup

Lishin1215:
This PR contains two commits
1. The first commit adds manual drop emission for explicit return.
2. The second commit switches to the function-scope try-finally cleanup and removes the manual drop-emission code.

It covers these cases:

1. Unit explicit return

```rust
fn f() {
    let _x = Droppable;
    return;
}
```

2. Unit explicit return expression

```rust
fn f() {
    let _x = Droppable;
    return make_unit();
}
```

4. Non-unit explicit return expression

```rust
fn f() -> i32 {
    let _x = Droppable;
    return make_value();
}
```

5. Nested explicit returns

```rust
fn f() {
    let _outer = Droppable;
    {
        let _inner = Droppable;
        return;
    }
}
```

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -271,6 +271,30 @@ CompileExpr::visit (HIR::ReturnExpr &expr)
 				    lvalue_locus, rvalue_locus);
     }
 
+  if (fncontext.retty->is_unit ())
+    {
+      if (expr.has_return_expr ())
+	{
+	  ctx->add_statement (return_value);
+	  return_value = unit_expression (expr.get_locus ());
+	}
+    }
+  else if (expr.has_return_expr ())
+    {
+      tree result_reference
+	= Backend::var_expression (fncontext.ret_addr, expr.get_locus ());
+
+      tree assignment
+	= Backend::assignment_statement (result_reference, return_value,
+					 expr.get_locus ());
+
+      ctx->add_statement (assignment);
+      return_value
+	= Backend::var_expression (fncontext.ret_addr, expr.get_locus ());
+    }
+
+  CompileDrop (ctx).emit_return_scope_drop_calls ();
+
   tree return_stmt = Backend::return_statement (fncontext.fndecl, return_value,
 						expr.get_locus ());
   ctx->add_statement (return_stmt);
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -293,11 +293,10 @@ CompileExpr::visit (HIR::ReturnExpr &expr)
 	= Backend::var_expression (fncontext.ret_addr, expr.get_locus ());
     }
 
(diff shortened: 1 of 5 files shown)


gcc/rust/backend/rust-compile-drop.cc
```diff
@@ -109,5 +108,21 @@ CompileDrop::emit_current_scope_drop_calls (Context *ctx)
...
 
+void
+CompileDrop::emit_current_scope_drop_calls (Context *ctx)
+{
+  emit_drop_candidate_calls (ctx, ctx->peek_block_drop_candidates ());
+}
+
+void
+CompileDrop::emit_return_scope_drop_calls (Context *ctx)
+{
```
Lishin1215: I added a separate emit function for explicit returns because they can leave 
multiple active block scopes.

It walks the drop candidate stack from the innermost scope outward. Each scope
still uses the normal reverse drop order. 
This keeps the current block-scope drop behavior unchanged.

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -142,6 +143,30 @@ CompileExpr::visit (HIR::ReturnExpr &expr)
 				    lvalue_locus, rvalue_locus);
     }
 
+  if (fncontext.retty->is_unit ())
+    {
```
Lishin1215, line 275: I split the handling for non-unit returns and unit returns.

For non-unit returns, the value is saved first, then drops are emitted, and then 
the saved value is returned. 
For unit returns with an explicit expression, the expression is emitted as a statement 
before emitting drops and returning `()`.

philberty:
merge conflicts but i am torn here because we have 2 options do this so we have complete control of the ordering of how drops are invoked.. but i also think there is value if we use:

TRY_FINALLY_EXPR

it means every time we enter a HIR::BlockExpr you setup this TRY Finally and use your drop context so you dont have to figure out how to put in the apropriate drop calls at specific points you just put them in the finally and because we have the HIR::Block expr following this convention means nested stuff which need drops this should just work

Lishin1215:
@philberty Thanks, that makes sense.
I rebased and pushed an update to resolve the merge conflicts!
I agree that TRY_FINALLY_EXPR may be the better general direction for block cleanup.

Do you prefer keeping this PR as the smaller explicit-return fix, or should I rework it 
to use TRY_FINALLY_EXPR for HIR::BlockExpr now?

CohenArthur:
We should keep this PR as the small fix, and then work on getting `TRY_FINALLY_EXPR` working in a later PR in my opinion. I think it is nice to build Drop piece by piece and then improve it as we understand more and more about what we are supposed to do, and make it better and more correct in the process

Lishin1215:
@CohenArthur Thank you!
I opened a separate PR for the TRY_FINALLY_EXPR work: https://github.com/Rust-GCC/gccrs/pull/4685.
For this PR, I will leave it as-is.

P-E-P (approved):

CohenArthur (approved):
LGTM!

PR #4622, opened 2026-06-26
gccrs: add while true lint

lucasly-ba:
This patch adds the while true lint, which warns when a while loop uses the literal `true` as its condition and suggests using `loop` instead.

gcc/testsuite/ChangeLog:

	* rust/compile/while-true_0.rs: New test.

### DIFF
--- a/gcc/rust/checks/lints/unused/rust-unused-checker.cc
+++ b/gcc/rust/checks/lints/unused/rust-unused-checker.cc
@@ -347,5 +347,70 @@ UnusedChecker::visit (HIR::LetStmt &stmt)
   walk (stmt);
 }
 
+namespace {
+
+// Collects the value of a boolean literal expression, if the visited
+// expression is one.
+class BoolLiteral : public HIR::HIRFullVisitorBase
+{
+public:
+  bool found = false;
+  bool value = false;
+
+  using HIR::HIRFullVisitorBase::visit;
+  void visit (HIR::LiteralExpr &lit) override
+  {
+    if (lit.get_literal ().get_lit_type () == HIR::Literal::LitType::BOOL)
+      {
+	found = true;
+	value
+	  = lit.get_literal ().as_string () == Values::Keywords::TRUE_LITERAL;
+      }
+  }
+};
+
+// Detects a predicate that is constantly true: the `true` literal, or a
+// comparison of two boolean literals that always holds.
+class ConstantTruth : public HIR::HIRFullVisitorBase
+{
+public:
+  bool is_true = false;
+
+  using HIR::HIRFullVisitorBase::visit;
+  void visit (HIR::LiteralExpr &lit) override
+  {
+    if (lit.get_literal ().get_lit_type () == HIR::Literal::LitType::BOOL)
+      is_true
+	= lit.get_literal ().as_string () == Values::Keywords::TRUE_LITERAL;
+  }
+
+  void visit (HIR::ComparisonExpr &cmp) override
+  {
+    BoolLiteral lhs, rhs;
+    cmp.get_lhs ().accept_vis (lhs);
+    cmp.get_rhs ().accept_vis (rhs);
+    if (lhs.found && rhs.found)
+      {
+	if (cmp.get_kind () == ComparisonOperator::EQUAL)
+	  is_true = lhs.value == rhs.value;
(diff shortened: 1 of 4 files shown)


gcc/rust/checks/lints/unused/rust-unused-checker.cc
```diff
@@ -323,5 +323,21 @@ UnusedChecker::visit (HIR::MatchExpr &expr)
...
+UnusedChecker::visit (HIR::WhileLoopExpr &expr)
+{
+  auto &predicate = expr.get_predicate_expr ();
+  if (predicate.get_expression_type () == HIR::Expr::ExprType::Lit)
+    {
+      auto &literal
+	= static_cast<HIR::LiteralExpr &> (predicate).get_literal ();
+      if (literal.get_lit_type () == HIR::Literal::LitType::BOOL
+	  && literal.as_string () == "true")
+	rust_warning_at (expr.get_locus (), OPT_Wunused_variable,
```
CohenArthur: I don't think `-Wunused-variable` is the right flag to check for this

CohenArthur (commented):
the implementation LGTM but just check the flag, I don't know that we want these warnings to be disabled if someone passes -Wno-unused-variable

gcc/rust/checks/lints/unused/rust-unused-checker.cc
```diff
@@ -323,5 +323,21 @@ UnusedChecker::visit (HIR::MatchExpr &expr)
...
+void
+UnusedChecker::visit (HIR::WhileLoopExpr &expr)
+{
+  auto &predicate = expr.get_predicate_expr ();
+  if (predicate.get_expression_type () == HIR::Expr::ExprType::Lit)
+    {
+      auto &literal
+	= static_cast<HIR::LiteralExpr &> (predicate).get_literal ();
+      if (literal.get_lit_type () == HIR::Literal::LitType::BOOL
+	  && literal.as_string () == "true")
```
P-E-P: Please use `Rust::Values::Keywords::TRUE_LITERAL`.

gcc/rust/checks/lints/unused/rust-unused-checker.cc
```diff
@@ -323,5 +323,21 @@ UnusedChecker::visit (HIR::MatchExpr &expr)
   walk (expr);
 }
 
+void
+UnusedChecker::visit (HIR::WhileLoopExpr &expr)
+{
+  auto &predicate = expr.get_predicate_expr ();
+  if (predicate.get_expression_type () == HIR::Expr::ExprType::Lit)
```
P-E-P: I wonder if we should check for a constant tautology instead of a literal, this way `true == true` would also get matched.

gcc/rust/checks/lints/unused/rust-unused-checker.cc
powerboat9: Maybe something like `-Wrust-wip-misc`?

gcc/rust/checks/lints/unused/rust-unused-checker.cc
lucasly-ba: frontend is built with -fno-rtti, so I can't dynamic_cast to tell a ComparisonExpr apart from the other OperatorExpr subclasses inside the lint.  any idea?

gcc/rust/checks/lints/unused/rust-unused-checker.cc
lucasly-ba: Settled this on Zulip with PEP

lucasly-ba:
@P-E-P

gcc/rust/checks/lints/unused/rust-unused-checker.cc
P-E-P: You either have to tag it or use a visitor to collect that information.

lucasly-ba:
Should be good @P-E-P @CohenArthur

philberty:
Hmm these all have merge conflicts can you fix up 2 or so then ping us on zulip

philberty (changes requested):
Hmm i am tempted to approve this but cant help but think that this could be done on the GCC tree's instead and intercept it either as a walk_tree helper or as a when we compile loop predicates it will do a build_fold and we can check for true on the gcc tree at that point hmm.. what does C/C++ do they check for this? not sure

PR #4623, opened 2026-06-26
gccrs: add path statement lint

lucasly-ba:
This patch adds the path statement lint, which warns on a statement made of a single path expression, as it has no effect.

gcc/testsuite/ChangeLog:

	* rust/compile/path-statements_0.rs: New test.

### DIFF
--- a/gcc/rust/checks/lints/unused/rust-unused-checker.cc
+++ b/gcc/rust/checks/lints/unused/rust-unused-checker.cc
@@ -400,5 +400,14 @@ UnusedChecker::visit (HIR::NegationExpr &expr)
   walk (expr);
 }
 
+void
+UnusedChecker::visit (HIR::ExprStmt &stmt)
+{
+  if (stmt.get_expr ().get_expression_type () == HIR::Expr::ExprType::Path)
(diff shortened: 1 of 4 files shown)


CohenArthur (approved):

powerboat9:
Does this catch cases where a path statement is used, and has an effect? Say,

```rust
mod a {
    pub(crate) const B: bool = true;
}

pub fn foo_is_active () -> bool {
    a::B
}
```

lucasly-ba:
> Does this catch cases where a path statement is used, and has an effect? Say,
> 
> ```rust
> mod a {
>     pub(crate) const B: bool = true;
> }
> 
> pub fn foo_is_active () -> bool {
>     a::B
> }
> ```

Yes. The lint only fires on a path in statement position, a path used as a tail/return expression like your a::B isn't an ExprStmt, so it's never flagged same as rustc's path_statements. I've added your example as a test to make that explicit.

lucasly-ba:
@P-E-P

philberty:
merge conflicts but other wise looks good

lucasly-ba:
should be good @philberty

PR #4624, opened 2026-06-26
gccrs: add unreachable pattern lint

lucasly-ba:
This patch adds the unreachable pattern lint, which warns on match arms that can never be reached because an earlier arm is irrefutable.

gcc/testsuite/ChangeLog:

	* rust/compile/unreachable-patterns_0.rs: New test.

### DIFF
--- a/gcc/rust/checks/lints/unused/rust-unused-checker.cc
+++ b/gcc/rust/checks/lints/unused/rust-unused-checker.cc
@@ -331,6 +331,26 @@ UnusedChecker::visit (HIR::MatchExpr &expr)
 			 "multiple ranges are one apart");
     }
 
+  // Every arm after an unguarded wildcard can never match. A bare identifier
+  // is not treated as irrefutable: the resolver does not yet disambiguate a
+  // unit-variant path (e.g. None) from a fresh binding.
+  bool seen_irrefutable = false;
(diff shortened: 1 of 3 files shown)


gcc/rust/checks/lints/unused/rust-unused-checker.cc
```diff
@@ -320,6 +320,29 @@ UnusedChecker::visit (HIR::MatchExpr &expr)
...
+	  continue;
+	}
+
+      auto type = arm.get_pattern ()->get_pattern_type ();
+      if (!arm.has_match_arm_guard ()
+	  && (type == HIR::Pattern::PatternType::WILDCARD
+	      || (type == HIR::Pattern::PatternType::IDENTIFIER
+		  && !static_cast<HIR::IdentifierPattern &> (
+			*arm.get_pattern ())
+			.has_subpattern ())))
```
CohenArthur: This also needs a test, not just the wildcard pattern. I think this is not enough to check that an identifier pattern is irrefutable, because for example this gets parsed as an identifier pattern:

```rust
enum Option<T> {
    Some(T),
    None,
}

use Option::{Some, None};

match some_option {
    None => {},
}
```

and is not irrefutable

CohenArthur (changes requested):

PR #4626, opened 2026-06-26
gccrs: add unused comparisons lint

lucasly-ba:
This patch adds the unused comparisons lint, which warns on comparisons that are always true or false due to the type limits of an unsigned integer.

gcc/testsuite/ChangeLog:

	* rust/compile/unused-comparisons_0.rs: New test.


lucasly-ba:
@P-E-P

P-E-P (approved):

PR #4628, opened 2026-06-27
gccrs: add non shorthand field patterns lint

lucasly-ba:
This patch adds the non shorthand field patterns lint, which warns when a struct pattern field binds to a variable of the same name, e.g. `Point { x: x }`, where the shorthand `Point { x }` should be used.

gcc/testsuite/ChangeLog:

	* rust/compile/non-shorthand-field-patterns_0.rs: New test.


philberty (approved):
LGTM

PR #4629, opened 2026-06-27
gccrs: add trivial numeric casts lint

lucasly-ba:
This patch adds the trivial numeric casts lint, which warns on a numeric cast where the source and target types are the same, e.g. `x as i32` when `x` is already an `i32`.

gcc/testsuite/ChangeLog:

	* rust/compile/trivial-numeric-casts_0.rs: New test.

### DIFF
--- a/gcc/rust/checks/lints/unused/rust-unused-checker.cc
+++ b/gcc/rust/checks/lints/unused/rust-unused-checker.cc
@@ -26,13 +26,24 @@
 #include "rust-keyword-values.h"
 #include "rust-attribute-values.h"
 #include "rust-rib.h"
+#include "rust-tyty.h"
 
 namespace Rust {
 namespace Analysis {
 UnusedChecker::UnusedChecker ()
   : nr_context (Resolver2_0::FinalizedNameResolutionContext::get ()),
-    mappings (Analysis::Mappings::get ()), unused_context (UnusedContext ())
+    mappings (Analysis::Mappings::get ()),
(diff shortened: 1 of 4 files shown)


lucasly-ba:
@P-E-P

gcc/rust/checks/lints/unused/rust-unused-checker.cc
```diff
@@ -347,5 +358,21 @@ UnusedChecker::visit (HIR::LetStmt &stmt)
   walk (stmt);
 }
 
+void
+UnusedChecker::visit (HIR::TypeCastExpr &expr)
+{
+  TyTy::BaseType *from;
+  TyTy::BaseType *to;
+  if (context.lookup_type (expr.get_casted_expr ().get_mappings ().get_hirid (),
```
philberty: You should split this if statement up...
```
bool found_from = lookup_expr;
bool found_expr = lookup
if (found_from && found_expr)
{
    bool is_numeric = ...
    bool type_names_match = ...
   // you might be able to get away with using TypeCompatable to be more generic not sure maybe this is meant to only work on numeric values...

  if (is_numeric && type_names_match)
  ... warrning
   
}

walk (expr)
```

This is jsut much easier to read down the line

philberty (changes requested):

gcc/rust/checks/lints/unused/rust-unused-checker.cc
lucasly-ba: should be good @philberty

PR #4630, opened 2026-06-27
gccrs: add unused unsafe lint

lucasly-ba:
This patch adds the unused unsafe lint, which warns about an `unsafe` block that does not contain any operation requiring an unsafe context. The unsafe checker now records which unsafe blocks are actually used and reports the rest.

gcc/testsuite/ChangeLog:

	* rust/compile/unused-unsafe_0.rs: New test.

### DIFF
--- a/gcc/rust/checks/errors/rust-unsafe-checker.h
+++ b/gcc/rust/checks/errors/rust-unsafe-checker.h
@@ -24,6 +24,8 @@
 #include "rust-hir-type-check.h"
 #include "rust-stacked-contexts.h"
 
+#include <unordered_set>
+
 namespace Rust {
 namespace HIR {
 class UnsafeChecker : public HIRFullVisitor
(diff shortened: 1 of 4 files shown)


gcc/rust/checks/errors/rust-unsafe-checker.h
```diff
@@ -51,7 +51,13 @@ class UnsafeChecker : public HIRFullVisitor
    */
   void check_function_attr (HirId node_id, location_t locus);
 
+  /**
+   * Mark the current unsafe block as required by an unsafe operation
+   */
+  void mark_unsafe_used ();
+
   StackedContexts<HirId> unsafe_context;
+  std::set<HirId> used_unsafe_blocks;
```
P-E-P: We do not require any ordering on those HirId, we could change this to `std::unordered_set`.

lucasly-ba: done

lucasly-ba:
@P-E-P

philberty (approved):
LGTM

PR #4632, opened 2026-06-27
gccrs: Add tests for local binding LIFO and nested block scopes

Lishin1215:
This PR adds two execute tests for local Drop ordering.

It covers:
- multiple local bindings in the same scope being dropped in LIFO order

```rust
fn f() {
    let _a = A;
    let _b = B;
    let _c = C;
}
```
- nested block scopes dropping inner bindings before outer bindings

```rust
fn f() {
    let _outer = Outer;
    {
        let _middle = Middle;
        {
            let _inner = Inner;
        }
        let _middle2 = Middle2;
    }
    let _outer2 = Outer2;
}
```

These cases are already supported by the current Drop implementation, and
this PR adds tests to make the expected order explicit.

-------
gcc/testsuite/ChangeLog:

	* rust/execute/drop-local-binding-lifo.rs: New test.
	* rust/execute/drop-nested-block-scope.rs: New test.

### DIFF
--- a/gcc/testsuite/rust/execute/drop-local-binding-lifo.rs
+++ b/gcc/testsuite/rust/execute/drop-local-binding-lifo.rs
@@ -1,0 +1,56 @@
+// { dg-output "C\r*\nB\r*\nA\r*\n" }
+// { dg-additional-options "-w" }
+#![feature(no_core)]
+#![feature(lang_items)]
+#![no_core]
+
+extern "C" {
+    fn printf(s: *const i8, ...);
+}
+
+#[lang = "sized"]
+pub trait Sized {}
+
+#[lang = "drop"]
+pub trait Drop {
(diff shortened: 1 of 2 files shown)


Polygonalr:
Just a small suggestion: maybe for the nested block scope test, it's ideal to also test out variables declared in outer scopes, after an inner scope, for example:

```rust
fn f() {
    let _outer = Outer;
    {
        let _middle = Middle;
        {
            let _inner = Inner;
        }
        let _middle2 = Middle2;
    }
    let _outer2 = Outer2;
}
```

Lishin1215:
@Polygonalr Thank you for the suggestion, that makes a lot of sense.
I’ll replace the current nested block test with this case.

Thanks!

philberty (approved):
LGTM

PR #4633, opened 2026-06-27
gccrs: add ellipsis inclusive range patterns lint

lucasly-ba:
This patch adds the ellipsis inclusive range patterns lint, which warns on the deprecated `...` range pattern syntax and suggests using `..=` instead.

It also propagates the ellipsis syntax flag when lowering range patterns to HIR, so the deprecated `...` form can be distinguished from `..=`.

gcc/testsuite/ChangeLog:

	* rust/compile/ellipsis-inclusive-range-patterns_0.rs: New test.


lucasly-ba:
@P-E-P

PR #4634, opened 2026-06-27
gccrs: add bare trait objects lint

lucasly-ba:
This patch adds the bare trait objects lint, which warns on trait objects written without an explicit `dyn`.

A single bare trait (e.g. `&Foo`) is lowered to a type-path that resolves to a trait definition, handled in `visit (HIR::TypePath)`; a bare trait object with several bounds (e.g. `&(Foo + Bar)`) is a `TraitObjectType` without the `dyn` keyword.

gcc/testsuite/ChangeLog:

	* rust/compile/bare-trait-objects_0.rs: New test.

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-type.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-type.cc
@@ -407,6 +407,11 @@ TypeCheckType::resolve_root_path (HIR::TypePath &path, size_t *offset,
 	  return root_tyty;
 	}
 
+      if (flag_unused_check_2_0 && lookup->is<TyTy::DynamicObjectType> ())
+	rust_warning_at (
+	  seg->get_locus (), OPT_Wdeprecated,
+	  "trait objects without an explicit %<dyn%> are deprecated");
+
       // if we have a previous segment type
       if (root_tyty != nullptr)
 	{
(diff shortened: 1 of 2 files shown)


gcc/rust/typecheck/rust-hir-type-check-type.cc
```diff
@@ -151,6 +151,25 @@ TypeCheckType::visit (HIR::TypePath &path)
       return;
     }
 
+  // The bare_trait_objects lint: a type-path that resolves directly to a trait
+  // definition is a trait object written without an explicit %<dyn%> (the
+  // explicit form is lowered to a TraitObjectType and handled separately).
+  if (flag_unused_check_2_0 && root->get_kind () == TyTy::TypeKind::DYNAMIC)
```
philberty: Do we need this? what part of the test case covers this

philberty: ah its for &Foo

philberty: A better way to do this is instead just add 

```
      if (lookup->is<TyTy::DynamicObjectType> ())
	rust_warning_at (
	  seg->get_locus (), OPT_Wunused_variable,
	  "trait objects without an explicit %<dyn%> are deprecated");
```

during 

```
TyTy::BaseType *
TypeCheckType::resolve_root_path (HIR::TypePath &path, size_t *offset,
				  bool *wasBigSelf)
```

just after it does query type should be a simpler way to solve this

philberty (changes requested):
that one check i think can be done simpler

lucasly-ba:
@philberty  should be good

lucasly-ba:
@P-E-P

PR #4637, opened 2026-06-27
gccrs: fix ICE on generic type aliases

lucasly-ba:
A type alias with generic parameters never resolved those parameters into the type context, so the reachability pass asserted (`rust_assert` in `visit_generic_predicates`) when looking up their types. This resolves the generic parameters as the other items do.

gcc/testsuite/ChangeLog:

	* rust/compile/type-alias-generic-params.rs: New test.


philberty (approved):
LGTM

PR #4638, opened 2026-06-27
gccrs: fix ICE on break with a label and a value

lucasly-ba:
A `break 'label value` expression looks up a temporary variable associated with the labeled loop to hold the broken value, but labeled loops never registered that temporary, causing an ICE in the code generator (`lookup_label_temp_var`). This registers the loop result temporary against the label.

gcc/testsuite/ChangeLog:

	* rust/compile/break-label-loop.rs: New test.


philberty (approved):
LGTM

PR #4639, opened 2026-06-27
gccrs: add type alias bounds lint

lucasly-ba:
This patch adds the type alias bounds lint, which warns on bounds applied to the generic parameters of a type alias, since those bounds are not enforced.

Note: this branch is based on #4637 (fix ICE on generic type aliases); the first commit is that fix, without which the construct cannot be compiled. It can be rebased once #4637 lands.

gcc/testsuite/ChangeLog:

	* rust/compile/type-alias-bounds_0.rs: New test.

### DIFF
--- a/gcc/rust/checks/lints/unused/rust-unused-checker.cc
+++ b/gcc/rust/checks/lints/unused/rust-unused-checker.cc
@@ -347,5 +347,24 @@ UnusedChecker::visit (HIR::LetStmt &stmt)
   walk (stmt);
 }
 
+void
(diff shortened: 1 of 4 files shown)


philberty (changes requested):
LGTM but there are merge conflicts

philberty:
merge conflicts again

lucasly-ba:
@philberty  should be good

lucasly-ba:
@P-E-P

philberty:
mer conflicts.. just combine a few of these to make it easier

lucasly-ba:
@philberty  just opened #4797

PR #4640, opened 2026-06-27
gccrs: add break with label and loop lint

lucasly-ba:
This patch adds the break with label and loop lint, which warns on a labeled `break` whose value is a loop expression, as it is easy to confuse with an unlabeled `break` of a labeled value.

Note: this branch is based on #4638 (fix ICE on break with a label and a value); the first commit is that fix, without which the construct cannot be compiled. It can be rebased once #4638 lands.

gcc/testsuite/ChangeLog:

	* rust/compile/break-with-label-and-loop_0.rs: New test.

### DIFF
--- a/gcc/rust/checks/lints/unused/rust-unused-checker.cc
+++ b/gcc/rust/checks/lints/unused/rust-unused-checker.cc
@@ -400,5 +400,18 @@ UnusedChecker::visit (HIR::NegationExpr &expr)
   walk (expr);
 }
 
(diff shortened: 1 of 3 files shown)


lucasly-ba:
@P-E-P

philberty (approved):
LGTM

philberty:
merge conflicts

lucasly-ba:
should be good @philberty

PR #4641, opened 2026-06-27
gccrs: add dyn drop lint

lucasly-ba:
This patch adds the dyn drop lint, which warns on a trait object carrying a `Drop` bound, since values are dropped automatically and the bound has no effect.

gcc/testsuite/ChangeLog:

	* rust/compile/dyn-drop_0.rs: New test.

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-type.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-type.cc
@@ -655,6 +655,16 @@ TypeCheckType::visit (HIR::TraitObjectType &type)
 	specified_bounds.push_back (std::move (predicate));
     }
 
+  // The dyn_drop lint: a trait object with a `Drop` bound is pointless, as
+  // values are dropped automatically regardless of the bound.
+  if (flag_unused_check_2_0)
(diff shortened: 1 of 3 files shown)


gcc/rust/typecheck/rust-hir-type-check-type.cc
```diff
@@ -655,6 +655,16 @@ TypeCheckType::visit (HIR::TraitObjectType &type)
 	specified_bounds.push_back (std::move (predicate));
     }
 
+  // The dyn_drop lint: a trait object with a `Drop` bound is pointless, as
+  // values are dropped automatically regardless of the bound.
+  if (flag_unused_check_2_0)
+    if (auto drop = mappings.lookup_lang_item (LangItem::Kind::DROP))
+      for (auto &bound : specified_bounds)
+	if (bound.get_id () == drop.value ())
+	  rust_warning_at (type.get_locus (), OPT_Wunused_variable,
```
P-E-P: `OPT_Wunused` is probably better here, `Drop` is a bound, not a variable. We could introduce `OPT_Wunused_bound` later down the line ? Maybe the C++ frontend already has something with concepts, I couldn't find any within `options.h`.

P-E-P (approved):

PR #4642, opened 2026-06-27
gccrs: add overlapping range endpoints lint

lucasly-ba:
This patch adds the overlapping range endpoints lint, which warns when two range patterns in a match overlap on a single endpoint, for example `0..=5` and `5..=10` (both match `5`).

gcc/testsuite/ChangeLog:

	* rust/compile/overlapping-range-endpoints_0.rs: New test.

### DIFF
--- a/gcc/rust/checks/lints/unused/rust-unused-checker.cc
+++ b/gcc/rust/checks/lints/unused/rust-unused-checker.cc
@@ -331,6 +331,21 @@ UnusedChecker::visit (HIR::MatchExpr &expr)
 			 "multiple ranges are one apart");
     }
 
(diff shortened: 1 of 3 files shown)


powerboat9:
It looks like these range checks are `O(n^2)` despite there probably being `O(nlogn)` solutions, but it should be fine for now.

powerboat9:
Actually, it looks like your changelog needs some tweaks. It should only mention the test you added and your changes to `(UnusedChecker::visit (MatchExpr))`

powerboat9 (changes requested):
Changelog

lucasly-ba:
@P-E-P

PR #4643, opened 2026-06-27
gccrs: add double negations lint

lucasly-ba:
This patch adds the double negations lint, which warns on a double arithmetic negation such as `- -x` (Rust has no `--` operator, so this is likely a mistake).

gcc/testsuite/ChangeLog:

	* rust/compile/double-negations_0.rs: New test.


lucasly-ba:
@P-E-P

P-E-P (approved):

lucasly-ba:
@P-E-P should be good

PR #4644, opened 2026-06-27
gccrs: add unused visibilities lint

lucasly-ba:
This patch adds the unused visibilities lint, which warns on a visibility qualifier applied to a `const _` item, as it has no effect.

gcc/testsuite/ChangeLog:

	* rust/compile/unused-visibilities_0.rs: New test.


philberty (approved):
LGTM

PR #4645, opened 2026-06-27
gccrs: add no mangle generic items lint

lucasly-ba:
This patch adds the no mangle generic items lint, which warns when `#[no_mangle]` or `#[export_name]` is applied to a generic function, since generic items must be mangled and the attribute has no effect.

gcc/testsuite/ChangeLog:

	* rust/compile/no-mangle-generic-items_0.rs: New test.


lucasly-ba:
@P-E-P

P-E-P (approved):

PR #4646, opened 2026-06-27
gccrs: add static mut refs lint

lucasly-ba:
This patch adds the static mut refs lint, which warns on taking a reference to a mutable static, as it can easily lead to undefined behaviour.

gcc/testsuite/ChangeLog:

	* rust/compile/static-mut-refs_0.rs: New test.


lucasly-ba:
@P-E-P

P-E-P (approved):

PR #4649, opened 2026-06-28
gccrs: add missing_abi lint

lucasly-ba:
Warn on an `extern` block written without an explicit ABI string, such as a bare `extern { ... }`, which silently defaults to the C ABI. Rust deprecates this in favour of spelling the ABI out, e.g. `extern "C"`.


philberty (approved):
LGTM but merge conflicts

philberty (approved):

PR #4650, opened 2026-06-28
#[repr(transparent)] initial implementation

Polygonalr:
~~This is needed for for #4565, because `CStr` is implemented with `#[repr(transparent)]` attribute.~~ #4565 got merged, this PR will also include a fix for a mistake I made there.

**Note 1**: Type-checking doesn't support extra zero-sized members yet, e.g. `Unique` struct in `core/src/ptr/unique.rs`:

```rust
#[repr(transparent)]
pub struct Unique<T: ?Sized> {
    pointer: *const T,
    _marker: PhantomData<T>, // zero-sized member!
}
```

I can open an issue for this after this gets merged.

**Note 2**: I need someone who's more experienced in Rust to confirm the behavior of transparent structs with 0 fields, since 0-field transparent structs are present in both `core` and `kernel`. Added a comment regarding in my commit that for now.

### DIFF
--- a/gcc/rust/backend/rust-compile-expr.cc
+++ b/gcc/rust/backend/rust-compile-expr.cc
@@ -769,10 +769,22 @@ CompileExpr::visit (HIR::StructExprStructFields &struct_expr)
 
   if (!adt->is_enum ())
     {
-      translated
-	= Backend::constructor_expression (compiled_adt_type, adt->is_enum (),
-					   arguments, union_disriminator,
-					   struct_expr.get_locus ());
+      auto repr_kind = adt->get_repr_options ().repr_kind;
+      if (repr_kind == TyTy::ADTType::ReprKind::TRANSPARENT)
+	{
+	  translated
+	    = fold_build1_loc (struct_expr.get_locus (), VIEW_CONVERT_EXPR,
+			       compiled_adt_type, arguments.front ());
+	}
+      else
+	{
+	  translated
+	    = Backend::constructor_expression (compiled_adt_type,
+					       adt->is_enum (), arguments,
+					       union_disriminator,
+					       struct_expr.get_locus ());
+	}
+
       return;
     }
 
@@ -843,6 +855,15 @@ CompileExpr::visit (HIR::FieldAccessExpr &expr)
       bool ok = variant->lookup_field (expr.get_field_name ().as_string (),
 				       nullptr, &field_index);
       rust_assert (ok);
+
+      auto repr_kind = adt->get_repr_options ().repr_kind;
+      if (repr_kind == TyTy::ADTType::ReprKind::TRANSPARENT)
+	{
+	  translated
+	    = compile_transparent_field_access (variant, expr.get_locus (),
+						receiver_ref);
+	  return;
+	}
     }
   else if (receiver->get_kind () == TyTy::TypeKind::REF)
     {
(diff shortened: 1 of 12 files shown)


powerboat9:
https://github.com/rust-lang/rust/issues/77841
https://github.com/rust-lang/rust/pull/115334

Seems like at most one field can be non-1-ZST (zero sized type with alignment 1). That non-1-ZST type can itself be a ZST. If there are no non-1-ZST types, it acts like `()` ABI wise.

https://github.com/rust-lang/rust/pull/155299

Also, it looks like much more recent versions of rustc warn/error when the user tries to rely on certain types being 1-ZSTs. I'd figure that could be left for later though.

gcc/rust/backend/rust-compile-expr.cc
```diff
@@ -859,22 +880,30 @@ CompileExpr::visit (HIR::FieldAccessExpr &expr)
...
+	      translated
+		= compile_transparent_field_access (variant, expr.get_locus (),
+						    indirect);
+	      return;
+	    }
+	  else
+	    {
+	      receiver_ref = indirect;
+	    }
 	}
```
powerboat9, line 895: Repeat check of `repr_kind`?

Polygonalr: I never should have pushed changes out at night, doesn't help with the fact that my IDE kept changing the tab sizes back to 2 instead of 8, nested if-else statements kept tripping me lol

gcc/rust/typecheck/rust-hir-type-check-base.cc
```diff
@@ -516,12 +516,12 @@ TypeCheckBase::parse_repr_options (const AST::AttrVec &attrs, location_t locus)
 	  bool is_align = false;
 	  bool is_c = false;
 	  bool is_integer = false;
+	  bool is_transparent = false;
 	  unsigned char value = 1;
 
 	  if (oparen == std::string::npos)
 	    {
 	      is_pack = inline_option.compare ("packed") == 0;
-	      is_align = inline_option.compare ("align") == 0;
```
powerboat9, line 525: Accidental removal?

gcc/rust/backend/rust-compile-expr.cc
powerboat9: No problem

gcc/rust/typecheck/rust-hir-type-check-base.cc
Polygonalr: Nope, I removed it because it is not possible to construct a struct with `#[repr(align)]` since the `align` representation requires an argument. On hindsight, I should be checking for this and throwing a specific error for it (like what `rustc` does).

Polygonalr: Added a commit to make it more explicit with new tests, to match `rustc` behaviour better.

gcc/rust/typecheck/rust-hir-type-check-base.cc
```diff
@@ -554,13 +564,26 @@ TypeCheckBase::parse_repr_options (const AST::AttrVec &attrs, location_t locus)
 	      value = strtoul (value_str.c_str () + 1, NULL, 10);
 	    }
 
-	  if (is_pack)
+	  if (is_transparent)
```
P-E-P, line 567: I wonder if this should be checked earlier within rust-attribute-checker ?

P-E-P (approved):
Looks great!
Have you looked into `scan-tree`/ `scan-rtl` deja gnu directives to compare the outer struct size/alignment with it's inner component ? It would be great to have a test for this behavior in case someone breaks transparent layout in the future.

gcc/rust/typecheck/rust-hir-type-check-base.cc
Polygonalr: Think it's neater to leave it here for now, we'll have to cover all ADTs if we shift the checks under `rust-attribute-checker`. These checks also feel more like type-checking than checking the attributes too.

P-E-P (approved):

P-E-P (approved):

PR #4651, opened 2026-06-28
gccrs: add repr_c_enums_larger_than_int lint

lucasly-ba:
Warn on a `#[repr(C)]` enum whose discriminant fits into neither a C `int` nor a C `unsigned int`. Such an enum is non-portable: C only permits enums whose discriminants do not all fit into one of those two types since C23, and Rust interprets `repr(C)` discriminants as `isize`, so the size may not match the C one.

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-item.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-item.cc
@@ -36,9 +36,51 @@
 #include "rust-type-util.h"
 #include "rust-tyty-variance-analysis.h"
 #include "rust-tyty.h"
+#include "options.h"
+#include "rust-compile-base.h"
+#include "rust-compile-context.h"
 
 namespace Rust {
 namespace Resolver {
+
+// Const-evaluate the discriminants of a repr(C) enum and warn when a value does
+// not fit into a C int/unsigned int. Done here, during type resolution, using
+// the compile context (a singleton shared with the backend).
+static void
+check_repr_c_enum_discriminants (Compile::Context *ctx, TyTy::BaseType *type)
+{
+  if (type->get_kind () != TyTy::TypeKind::ADT)
+    return;
+
+  auto &adt = static_cast<TyTy::ADTType &> (*type);
+  if (!adt.is_enum ()
+      || adt.get_repr_options ().repr_kind != TyTy::ADTType::ReprKind::C)
+    return;
+
+  for (auto &variant : adt.get_variants ())
+    {
+      if (!variant->has_discriminant ())
+	continue;
+
+      HIR::Expr &discriminant = variant->get_discriminant ();
+      TyTy::BaseType *discrim_ty = nullptr;
+      if (!ctx->get_tyctx ()->lookup_type (
+	    discriminant.get_mappings ().get_hirid (), &discrim_ty))
+	continue;
+
+      tree folded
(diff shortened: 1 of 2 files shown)


gcc/rust/typecheck/rust-hir-type-check-item.cc
```diff
@@ -410,6 +520,9 @@ TypeCheckItem::visit (HIR::Enum &enum_decl)
       variants.push_back (field_type);
     }
 
+  if (flag_unused_check_2_0 && repr.repr_kind == TyTy::ADTType::ReprKind::C)
```
philberty: This wont work because rustc supports stuff like:

```
#![feature(no_core)]
#![no_core]

const BASE: isize = 4;

const fn calculate(x: isize) -> isize {
    x * 2 + 1
}

pub enum Values {
    A = BASE + 1,
    B = calculate(3),
    C = {
        let x = 4;
        x * x
    },
}

```

So the HIR::EnumItemDiscriminant we can do const eval during type check by calling:
```
	  auto ctx = Compile::Context::get ();
	  tree capacity_expr
	    = Compile::HIRCompileBase::query_compile_const_expr (
	      ctx, capacity_type, type.get_size_expr ());
```
that should return the gcc tree for the value in theory it might do the type size checking also but yeah the enum discriminant HIR will need to store the resulting tree back again

philberty (changes requested):
Yeah i think we need to do const eval during type resolution of the enum disriminant variant

lucasly-ba:
@P-E-P

gcc/rust/typecheck/rust-hir-type-check-item.cc
```diff
@@ -36,10 +36,55 @@
...
+{
+  if (type->get_kind () != TyTy::TypeKind::ADT)
+    return;
+
+  auto &adt = static_cast<TyTy::ADTType &> (*type);
+  if (!adt.is_enum ()
+      || adt.get_repr_options ().repr_kind != TyTy::ADTType::ReprKind::C)
+    return;
+
+  const int64_t c_int_min = -2147483648LL;
```
P-E-P: `stdint` header provides a macro called `INT64_MAX`, we want to use it here.

gcc/rust/typecheck/rust-hir-type-check-item.cc
```diff
@@ -36,10 +36,55 @@
...
+  const int64_t c_uint_max = 4294967295LL;
```
P-E-P: Likewise

lucasly-ba: should be good @P-E-P

gcc/rust/typecheck/rust-hir-type-check-item.cc
```diff
@@ -36,10 +36,53 @@
 #include "rust-type-util.h"
 #include "rust-tyty-variance-analysis.h"
 #include "rust-tyty.h"
+#include "options.h"
+#include "rust-compile-base.h"
+#include "rust-compile-context.h"
+#include <cstdint>
```
P-E-P: You cannot include a system header like this, you need `rust-system.h`

lucasly-ba: should be good

philberty (approved):
LGTM

PR #4652, opened 2026-06-28
gccrs: add self_constructor_from_outer_item lint

lucasly-ba:
Warn when the `Self` constructor is used from an item nested inside an impl or trait, such as a function defined in a method body. Nested items do not inherit the `Self` of the surrounding impl. Gated behind `-frust-unused-check-2.0`.

### DIFF
--- a/gcc/rust/checks/errors/rust-hir-self-constructor.cc
+++ b/gcc/rust/checks/errors/rust-hir-self-constructor.cc
@@ -1,0 +1,82 @@
+// Copyright (C) 2025-2026 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#include "rust-hir-self-constructor.h"
+#include "rust-hir-item.h"
+#include "rust-hir-path.h"
+#include "rust-keyword-values.h"
+#include "rust-diagnostics.h"
+
+namespace Rust {
+namespace HIR {
+
+void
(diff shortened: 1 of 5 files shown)


philberty:
merge conflicts

powerboat9:
I think this should be an error unassociated with the unused checker

lucasly-ba:
@P-E-P

gcc/rust/checks/errors/rust-hir-self-constructor.cc
```diff
@@ -0,0 +1,82 @@
...
+  in_associated_scope = saved_associated;
+  impl_trait_nesting--;
+}
+
+void
+SelfConstructorChecker::visit (PathInExpression &path)
+{
+  if (self_ctor_from_outer && path.get_segments ().size () == 1
+      && path.get_segments ()[0].get_segment ().to_string ()
+	   == Values::Keywords::SELF_ALIAS)
```
CohenArthur, line 75: I think we have `get_segment ().is_big_self_segment ()`? that would be more readable here if it's available

gcc/rust/checks/errors/rust-hir-self-constructor.cc
```diff
@@ -0,0 +1,82 @@
...
+    rust_error_at (path.get_locus (),
+		   "cannot reference %<Self%> constructor from outer item");
```
CohenArthur, line 77: This isn't an error from rustc unless I'm mistaken? It's currently a warning and will be phased out, right?

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d0267b9a50d5ac5705f64912a5a8798a

so this should be changed to a similar message and a warning IMO

gcc/rust/checks/errors/rust-hir-self-constructor.cc
```diff
@@ -0,0 +1,82 @@
...
+#include "rust-hir-item.h"
+#include "rust-hir-path.h"
+#include "rust-keyword-values.h"
+#include "rust-diagnostics.h"
+
+namespace Rust {
+namespace HIR {
+
+void
+SelfConstructorChecker::go (Crate &crate)
```
CohenArthur, line 29: I think the name should change as there's other checks that would benefit from being implemented within this visitor. For example, using `Self` as a type name in outer items:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e23254c8936f2f674e637d1bbf53d319

```rust
struct Foo(i32);

impl Foo {
    fn foo() -> Self {
        fn bar() -> Self { Foo(15) }
        
        bar()
    }
}
```

PR #4653, opened 2026-06-28
gccrs: Create `ForeverStackBase`

powerboat9:
`ForeverStack::Node` doesn't have to be templated, so it can be moved to a new base class for `ForeverStack` (`ForeverStackBase`). This should help with handling Node objects that come from `ForeverStack` instances with different namespaces.


powerboat9:
Since the resolver code tends to have a lot of churn, I have another PR depending on this, and this has been up for a couple days, I'll merge this now

PR #4654, opened 2026-06-28
gccrs: add non_contiguous_range_endpoints lint

lucasly-ba:
Warn when an exclusive range pattern and the next arm are one value apart, such as `20..30` followed by `31..=40`, which silently skips `30`. Gated behind `-frust-unused-check-2.0`.


philberty (approved):
LGTM

PR #4659, opened 2026-06-29
gccrs: Add compatible version option

powerboat9:
GCCRS targets version 1.49 of rust. However, rust for linux expects rust 1.78 or later (soon 1.85 or later) and we will want to bump our main target version from 1.49 eventually. This patch adds an option for the user to specify a rust version to target for compatibility, so that we can account for backwards incompatible behavior between rust versions.

This should allow libcore 1.49 to compile without explicitly enabling post-1.49 features.

### DIFF
--- a/gcc/rust/checks/errors/feature/rust-feature-collector.cc
+++ b/gcc/rust/checks/errors/feature/rust-feature-collector.cc
@@ -18,6 +18,7 @@
 
 #include "rust-feature-collector.h"
 #include "rust-attribute-values.h"
+#include "rust-session-manager.h"
 
 namespace Rust {
 namespace Features {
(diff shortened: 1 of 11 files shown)


powerboat9:
This changes the interface added by #4425 (by removing an option). RfL devs should probably be informed before this is merged.

powerboat9:
This could be extended to cover `-frust-c-style-string-literals`, but I'll leave that for another PR

philberty (approved):
LGTM will leave @CohenArthur  to merge though to do final check

P-E-P (approved):

powerboat9:
@CohenArthur ok to merge?

PR #4660, opened 2026-06-29
backend: Fix incorrect impl block selection

### LINKED ISSUE #4600: Trait resolution mixes up methods for same-named structs in different modules
### Summary

The trait resolver fails to correctly differentiate between distinct structs that share the exact same name across different modules.

### Reproducer

I tried this code:

```rust
#![feature(no_core, lang_items)]
#![no_core]

#[lang = "sized"]
pub trait Sized {}

trait TraitA {
    fn do_a(&self) -> i32;
}

mod mod_a {
    use super::TraitA;
    
    pub struct StructX {
        pub val: i32,
    }
    
    impl TraitA for StructX {
        fn do_a(&self) -> i32 {
            1
        }
    }
    
    pub fn get_dyn(x: &StructX) -> &dyn TraitA {
        x as &dyn TraitA
    }
}

mod mod_b {
    use super::TraitA;
    
    pub struct StructX {
        pub val: i32,
    }
    
    impl TraitA for StructX {
        fn do_a(&self) -> i32 {
            0
        }
    }
    
    pub fn get_dyn(x: &StructX) -> &dyn TraitA {
        x as &dyn TraitA
    }
}

fn main() -> i32 {
    let b = mod_b::StructX { val: 10 };

    let dyn_b = mod_b::get_dyn(&b);

    dyn_b.do_a()
}

```

- [ ] Nightly

### Godbolt link

https://godbolt.org/z/Tz58Y1xq8

### Actual behavior

program returned: 1

### Expected behavior

program returned: 0

### GCC Version

commit-hash: 74c259427e971b8f18492974900d1c57cdc99ad5

nsvke:
During vtable generation in `compute_address_for_trait_item`, the compiler was picking the first `impl` block that satisfied the trait predicate, without verifying if the `impl` block actually belonged to the target receiver type. This caused dynamic dispatch to call the wrong methods when multiple types implemented the same trait in the same scope.

This patch adds a comparison between the impl block's resolved type and the `receiver` type to ensure we compute the address of the correct associated function.

Fixes #4600 

gcc/rust/ChangeLog:

	* backend/rust-compile.cc (HIRCompileBase::compute_address_for_trait_item): Ensure the impl block type matches the receiver type.

gcc/testsuite/ChangeLog:

	* rust/execute/trait-multi-impl.rs: New test.

### DIFF
--- a/gcc/rust/backend/rust-compile.cc
+++ b/gcc/rust/backend/rust-compile.cc
@@ -314,6 +314,15 @@ HIRCompileBase::compute_address_for_trait_item (
 
       rust_assert (ok);
 
+      TyTy::BaseType *mut_receiver = const_cast<TyTy::BaseType *> (receiver);
+      bool receiver_matches = Resolver::types_compatable (
+	TyTy::TyWithLocation (self, self->get_locus ()),
+	TyTy::TyWithLocation (mut_receiver, mut_receiver->get_locus ()),
+	UNDEF_LOCATION, false);
+
+      if (!receiver_matches)
+	continue;
+
       // Look through the relevant bounds on our type, and find which one our
       // impl block satisfies
       TyTy::TypeBoundPredicate *self_bound = nullptr;
--- a/gcc/testsuite/rust/execute/trait-multi-impl-generic.rs
+++ b/gcc/testsuite/rust/execute/trait-multi-impl-generic.rs
@@ -1,0 +1,46 @@
+#![feature(no_core, lang_items)]
+#![no_core]
+
+#[lang = "sized"]
+pub trait Sized {}
+
+trait TraitA {
+    fn do_a(&self) -> i32;
+}
+
+pub struct Container<T> {
+    pub val: i32,
+    pub _marker: T,
+}
+
+impl TraitA for Container<i32> {
+    fn do_a(&self) -> i32 {
+        self.val * 10
+    }
+}
+
+impl TraitA for Container<u32> {
+    fn do_a(&self) -> i32 {
+        self.val * 20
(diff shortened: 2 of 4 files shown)


gcc/rust/backend/rust-compile.cc
```diff
@@ -271,6 +271,9 @@ HIRCompileBase::compute_address_for_trait_item (
 
       rust_assert (ok);
 
+      if (self->get_ty_ref () != receiver->get_ty_ref ())
+	continue;
```
philberty, line 324: I dont think this will work for generics i think you should look at using the helper from type-util.cc types_compatable instead

nsvke: Thanks for the tip! I was actually working on a generic test case where ty ref failed, so this helper is exactly what I need.

nsvke: It passes for non-generics (even multi-module) and single-module generics, but the multi-module generic test still fails. I need to brainstorm this a bit. Do you have any suggestions?

philberty: I found the issue... in types_compatable i'll make an MR here then you can rebase your work ontop of it.. but you should only need types_compatable

philberty: this should fix it for you https://github.com/Rust-GCC/gccrs/pull/4663

nsvke: Good catch, thanks! It looks like it's working.

philberty (approved):
LGTM

PR #4662, opened 2026-06-29
forever-stack: Keep an unordered_map of Node children

CohenArthur:
There is no point in keeping a map instead of an unordered map here, and it harms
performance on big crates. Also slightly improve documentation and intent of the class.

gcc/rust/ChangeLog:

	* resolve/rust-forever-stack.h: Switch Node::children from map to unordered_map,
	improve documentation.


powerboat9:
I had thought `std::unordered_map` had weaker pointer invalidation guarantees, but it looks like only removal invalidates pointers to values or keys (https://en.cppreference.com/cpp/container/unordered_map). This does conflict with #4653 though.

PR #4672, opened 2026-06-30
Add size_of_val and min_align_of_val intrinsics

nsvke:
This patch adds the `size_of_val` and `min_align_of_val` intrinsics to the compiler. This intrinsics evaluates the size/align of a value at runtime, including DSTs.


philberty (approved):
LGTM

PR #4680, opened 2026-07-03
gccrs: Register `global_allocator` attribute

powerboat9:
This allows gccrs to recognize the `global_allocator` attribute and perform some sanity checks on its usage, but does not implement proper handling for the attribute.


powerboat9:
@nsvke thoughts? I don't want to step on your toes with the `alloc` work -- I salvaged this from an old branch I had laying around.

nsvke:
Please go ahead. Since the `#[global_allocator]` attribute is used by the consumers of `alloc`(like `std` or something else) rather than internally, it's not on my radar right now. You won't be stepping on my toes at all.

P-E-P (approved):

PR #4681, opened 2026-07-03
gccrs: Fix `IdentifierPathPass` on sub-patterns

powerboat9:
This should address #4116. It will probably fix it, but I want to double check before automatically closing #4116.

Supersedes #4117 and is a follow up to #4455.

### DIFF
--- a/gcc/rust/resolve/rust-identifier-path.cc
+++ b/gcc/rust/resolve/rust-identifier-path.cc
@@ -39,11 +39,14 @@ IdentifierPathPass::go (AST::Crate &crate, NameResolutionContext &ctx,
 void
 IdentifierPathPass::reseat (std::unique_ptr<AST::Pattern> &ptr)
 {
-  AST::IdentifierPattern *ident_pat;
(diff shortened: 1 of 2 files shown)


P-E-P (approved):

gcc/rust/resolve/rust-identifier-path.cc
```diff
@@ -43,7 +43,7 @@ IdentifierPathPass::reseat (std::unique_ptr<AST::Pattern> &ptr)
   if (ptr->get_pattern_kind () == AST::Pattern::Kind::Identifier)
     ident_pat = static_cast<AST::IdentifierPattern *> (ptr.get ());
   else
-    return;
+    goto fail;
```
P-E-P: I'd prefer if we explicitely call visit and then return. Gotos are confusing and people are not used to them.

powerboat9: Alright

P-E-P (approved):

PR #4683, opened 2026-07-04
gccrs: Handle ribs in `GlobbingVisitor`

powerboat9:
`GlobbingVisitor` can copy definitions between ribs, instead of creating new definitions from the AST.

### DIFF
--- a/gcc/rust/resolve/rust-finalize-imports-2.0.cc
+++ b/gcc/rust/resolve/rust-finalize-imports-2.0.cc
@@ -33,145 +33,66 @@ GlobbingVisitor::go (AST::GlobContainer *container)
   switch (container->get_glob_container_kind ())
     {
     case AST::GlobContainer::Kind::Module:
-      visit_module_container (static_cast<AST::Module &> (*container));
+      visit_container (static_cast<AST::Module &> (*container).get_node_id ());
       break;
     case AST::GlobContainer::Kind::Crate:
-      visit_crate_container (static_cast<AST::Crate &> (*container));
+      visit_container (static_cast<AST::Crate &> (*container).get_node_id ());
       break;
     case AST::GlobContainer::Kind::Enum:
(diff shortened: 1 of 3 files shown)


gcc/rust/resolve/rust-finalize-imports-2.0.cc
```diff
@@ -33,145 +33,66 @@ GlobbingVisitor::go (AST::GlobContainer *container)
...
-      if (variant->get_enum_item_kind () != AST::EnumItem::Kind::Struct)
-	ctx.insert_globbed (variant->get_identifier (), variant->get_node_id (),
-			    Namespace::Values);
+      auto globbed = glob_definition (ent.second);
+      if (globbed.has_value ())
+	{
+	  auto res = dst.insert (ent.first, globbed.value ());
+	  // inserting a globbed definition should (?) always succeed
+	  // TODO: double check
+	  // TODO: mark fixed point as dirty/changed?
```
CohenArthur, line 78: Yes, we need to mark the fixed point as dirty. This would insert a definition, hence in the current run of the resolver a usage could fail to resolve but would resolve to this newly inserted def in the run after. so we do need to run for at least one more turn

gcc/rust/resolve/rust-finalize-imports-2.0.cc
```diff
@@ -33,145 +33,66 @@ GlobbingVisitor::go (AST::GlobContainer *container)
...
-{}
-
-void
-GlobbingVisitor::visit (AST::ExternCrate &crate)
-{}
+  if (def.is_ambiguous ())
+    {
+      // TODO: error?
+      return tl::nullopt;
+    }
```
CohenArthur, line 93: This is per namespace right? So we probably need to assert that something isn't ambiguous here rather than returning a nullopt

CohenArthur (commented):
LGTM

PR #4684, opened 2026-07-04
gccrs: Fix type param forward declare check

powerboat9:
Type param bounds are allowed to use forward declared generic parameters.


philberty (approved):
LGTM

PR #4685, opened 2026-07-05
gccrs: Emit block scope drops through TRY_FINALLY_EXPR

Lishin1215:
This PR consists of 3 commits.

It changes block-scope Drop emission to use a TRY_FINALLY_EXPR cleanup instead of 
appending drop calls directly to the block body.

For example, a block like:
```rust
{
    let x = Droppable;
}
```

is lowered as:

```text
try {
    let x = Droppable;
} finally {
    drop(x);
}
```

This follows the same general cleanup/finally shape used by the Go frontend for defer handling.

For now, this PR only covers block-scope drops. Function-scope drops, explicit return cleanup, 
break/continue are left for follow-up work.

---
**commit 1**

gccrs: Add helper to build scope drop cleanup

gcc/rust/ChangeLog:

	* backend/rust-compile-drop.cc
	(CompileDrop::build_current_scope_drop_cleanup): New function to
	build current scope drop calls as a statement tree.
	(CompileDrop::emit_current_scope_drop_calls): Use it.
	* backend/rust-compile-drop.h
	(CompileDrop::build_current_scope_drop_cleanup): Declare.

**commit 2**

gccrs: Emit block scope drops through TRY_FINALLY_EXPR

gcc/rust/ChangeLog:

	* backend/rust-compile-block.cc
	(CompileBlock::visit): Build current scope drop cleanup and pass it
	when popping the block.
	* backend/rust-compile-context.h
	(Context::pop_block): Use pop_block_impl.
	(Context::pop_block_with_cleanup): New function for popping a block
	with cleanup code.
	(Context::pop_block_impl): New helper that adds the block statements
	directly, or wraps them in TRY_FINALLY_EXPR when cleanup is present.
	
**commit 3**

gccrs: Use EH_ELSE_EXPR in block Drop TRY_FINALLY_EXPR

gcc/rust/ChangeLog:

	* backend/rust-compile-context.h
	(Context::pop_block_impl): Wrap Drop cleanup in an EH_ELSE_EXPR.
	
---
**Note:**

`TRY_FINALLY_EXPR` cleanup originally caused `drop-nested-block-scope.rs` to fail
to link with an undefined reference to `__gccrs_personality_v0`.

This version wraps the cleanup operand in `EH_ELSE_EXPR`, using the Drop cleanup
as the normal cleanup and an empty exceptional cleanup for now.
With that, `drop-nested-block-scope.rs` links and passes locally.

### DIFF
--- a/gcc/rust/backend/rust-compile-context.h
+++ b/gcc/rust/backend/rust-compile-context.h
@@ -107,20 +107,11 @@ public:
     block_drop_candidates.emplace_back ();
   }
 
-  tree pop_block ()
+  tree pop_block () { return pop_block_impl (NULL_TREE, UNKNOWN_LOCATION); }
+
+  tree pop_block_with_cleanup (tree cleanup, location_t cleanup_locus)
   {
-    auto block = scope_stack.back ();
-    scope_stack.pop_back ();
-
-    auto stmts = statements.back ();
-    statements.pop_back ();
-
-    rust_assert (!block_drop_candidates.empty ());
-    block_drop_candidates.pop_back ();
-
-    Backend::block_add_statements (block, stmts);
-
-    return block;
+    return pop_block_impl (cleanup, cleanup_locus);
   }
 
   tree peek_enclosing_scope ()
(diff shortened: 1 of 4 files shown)


gcc/rust/backend/rust-compile-context.h
```diff
@@ -437,8 +437,14 @@ class Context
 	if (body == NULL_TREE)
 	  body = build_empty_stmt (cleanup_locus);
 
+	tree exceptional_cleanup = build_empty_stmt (cleanup_locus);
```
Lishin1215, line 440: I followed the cleanup shape described in gccint 11.7.6, the TRY_FINALLY_EXPR
cleanup operand can be an EH_ELSE_EXPR selector.

I did not add the empty volatile asm used by rustc_codegen_gcc to keep the
EH_ELSE_EXPR around, because the normal cleanup side here contains the actual
Drop cleanup, and the relevant tests pass with an empty exceptional cleanup.

P-E-P (approved):

PR #4686, opened 2026-07-06
gccrs: Fix duplicate "trailing semicolon in macro used" warnings

Polygonalr:
Compiling `core` is causing this warning to get spammed, this is just a hot fix to suppress the spam.

### DIFF
--- a/gcc/rust/expand/rust-macro-expand.cc
+++ b/gcc/rust/expand/rust-macro-expand.cc
@@ -1011,9 +1011,14 @@ transcribe_expression (Parser<MacroInvocLexer> &parser)
   // FIXME: make this an error for some edititons
   if (parser.peek_current_token ()->get_id () == SEMICOLON)
     {
-      rust_warning_at (
-	parser.peek_current_token ()->get_locus (), 0,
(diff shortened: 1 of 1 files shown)


gcc/rust/expand/rust-macro-expand.cc
```diff
@@ -1011,9 +1011,12 @@ transcribe_expression (Parser<MacroInvocLexer> &parser)
   // FIXME: make this an error for some edititons
   if (parser.peek_current_token ()->get_id () == SEMICOLON)
     {
-      rust_warning_at (
-	parser.peek_current_token ()->get_locus (), 0,
-	"trailing semicolon in macro used in expression context");
+      static std::unordered_set<location_t> warned_loc;
```
powerboat9, line 1016: Could you add a `TODO` comment for making this a member variable of `MacroExpander`?

CohenArthur: I think we need to be careful to still emit warnings different from the one we are currently keeping track of if we do this

CohenArthur (approved):
Good bandaid fix :D

PR #4693, opened 2026-07-08
testsuite: Add the alloc crate to the testsuite

nsvke:
This patch adds the `alloc` crate to the testsuite. A few temporary source-level workarounds were required to bypass current compiler limitations.

Currently, the entire crate successfully compiles until the name resolution phase


powerboat9:
Could you stick a copy of the 1.49 alloc crate in `libgrust/rustc-lib/`?

nsvke:
@CohenArthur @P-E-P

gcc/testsuite/rust/alloc/alloc.exp
```diff
@@ -0,0 +1,47 @@
...
+	set dg-do-what-default "compile"
+	set individual_timeout 600
+
+	dg-additional-files [ lsort [ glob -nocomplain \
+		$srcdir/$subdir/core.rs \
+		$srcdir/$subdir/prelude.rs \
+		$srcdir/$subdir/alloc/* \
+	]]
+
+	dg-runtest $srcdir/$subdir/alloc/src/lib.rs "-frust-edition=2015 -frust-crate=alloc -frust-compile-until=nameresolution -frust-cfg=target_has_atomic=\"ptr\" -w" ""
```
P-E-P, line 40: Shouldn't we compile this using edition 2018 ?

nsvke: yes but the 2018 edition causes unexpected errors.  I'll handle this, but I don't have time right now. I am temporarily using 2015 to protect it from regressions.

P-E-P (approved):

PR #4694, opened 2026-07-08
Resolve lang items from extern crate

P-E-P:
I've decided we should keep the current behavior of not visiting extern crate through their extern crate declaration and have an exception for the lang item collector.

### DIFF
--- a/gcc/rust/ast/rust-ast-dump.h
+++ b/gcc/rust/ast/rust-ast-dump.h
@@ -49,6 +49,16 @@ public:
       Dump,
       Hide,
     } dump_comments;
+    enum class Newline
+    {
+      Dump,
+      Hide,
+    } newline;
+    enum class Indentation
+    {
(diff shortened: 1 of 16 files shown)


gcc/rust/ast/rust-ast-dump.h
```diff
@@ -49,6 +49,12 @@ class Dump
       Dump,
       Hide,
     } dump_comments;
+    enum class Newline
+    {
+      Dump,
+      Hide,
+    } newline;
+    size_t indentation;
```
CohenArthur: I think we should have an enum here as well

```c++
enum class Indentation {
    None,
    Spaces4,
}
```

or something like that if we want more option. If we just want standard implementation then I would do

```c++
enum class Indentation {
    None,
    Standard,
}
```

CohenArthur (approved):
Good work :)

PR #4696, opened 2026-07-10
attr: Register needs_allocator

nsvke:
This patch registers the 'needs_allocator' attribute, which is required for compiling the 'alloc' crate.

Currently, it is implemented as a stub that emits a compiler warning when encountered. The reoson for this temporary stub is that full handling of 'needs_allocator' strictly depends on the 'global_allocator' infrastructure and the 'core::alloc::GlobalAlloc' trait. Since gccrs does not yet have full support for the 'core' crate, a complete implementation is deferred.

gcc/rust/ChangeLog:

	* checks/errors/rust-builtin-attribute-checker.cc (check_inner_attribute): Add warning for needs_allocator.
	* util/rust-attribute-values.h (class Attributes): Add NEEDS_ALLOCATOR constexpr.
	* util/rust-attributes.cc (__definitions): Add BuiltinAttrDefinition for new attribute.

gcc/testsuite/ChangeLog:

	* rust/compile/needs_allocator.rs: New test.

### DIFF
--- a/gcc/rust/checks/errors/rust-builtin-attribute-checker.cc
+++ b/gcc/rust/checks/errors/rust-builtin-attribute-checker.cc
@@ -38,6 +38,13 @@ check_inner_attribute (const AST::Attribute &attribute)
   if (Attributes::valid_outer_attribute (result.name))
     rust_error_at (attribute.get_locus (),
 		   "attribute cannot be used at crate level");
+
+  if (result.name == Values::Attributes::NEEDS_ALLOCATOR)
+    {
+      rust_warning_at (attribute.get_locus (), 0,
+		       "%<#[%s]%> is not implemented yet and has no effect",
+		       attribute.as_string ().c_str ());
+    }
 }
 
 /**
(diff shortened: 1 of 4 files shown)


gcc/rust/checks/errors/rust-builtin-attribute-checker.cc
```diff
@@ -38,6 +38,13 @@ check_inner_attribute (const AST::Attribute &attribute)
   if (Attributes::valid_outer_attribute (result.name))
     rust_error_at (attribute.get_locus (),
 		   "attribute cannot be used at crate level");
+
+  if (result.name == Values::Attributes::NEEDS_ALLOCATOR)
+    {
+      rust_warning_at (attribute.get_locus (), 0,
```
P-E-P, line 44: Do we want to abort compilation when this attribute is discovered by the compiler ? If we do, we'd better use `rust_sorry_at`.

nsvke: No, we don't want to abort compilation here. This attribute doesn't actually alter code generation; it is simply a marker to ensure that a global allocator is defined somewhere in the dependency graph when compiling a downstream binary.
​Emitting a warning is enough for now to indicate that the attribute's actual constraint checking is unimplemented, while still allowing the compilation of alloc to proceed.

P-E-P (approved):

PR #4698, opened 2026-07-12
gccrs: Support ZST type checking for `#[repr(transparent)]`

Polygonalr:
Zero-sized types such as `PhantomData` can be additional members of a `#[repr(transparent)]` struct besides the single non-zero-sized, so this PR supports type-checking them.

### DIFF
--- a/gcc/rust/typecheck/rust-tyty.cc
+++ b/gcc/rust/typecheck/rust-tyty.cc
@@ -927,6 +927,58 @@ BaseType::is_concrete () const
 }
 
 bool
+BaseType::is_zero_sized () const
+{
+  const TyTy::BaseType *x = destructure ();
+  switch (x->get_kind ())
+    {
+    // primitives that are always non-zero size
+    case FNPTR:
+    case FNDEF:
+    case SLICE:
+    case POINTER:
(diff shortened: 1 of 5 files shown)


gcc/rust/typecheck/rust-tyty.cc
```diff
@@ -926,6 +926,94 @@ BaseType::is_concrete () const
...
+    case PARAM:
+    case PROJECTION:
+    case PLACEHOLDER:
+    case ERROR:
+      return false;
+
+    case NEVER:
+      return true;
+    case TUPLE:
+      {
```
P-E-P, line 963: Switch cases easily get messy. Would it be possible to keep that switch within the current function and use a separate function for every case ?

Polygonalr: Hmm I was just following the same coding style as `is_unit()` too, but alright will abstract those checks out to their own functions.

P-E-P: There is a lot of places within the code were this isn't enforced, but we have to begin somewhere.

P-E-P (approved):

PR #4706, opened 2026-07-16
ci: Avoid trying to send emails on forked repos

powerboat9:
This should prevent `send_patches` from trying to send patches, failing to send patches, and emitting an error notification, on forked repos.


P-E-P (approved):

powerboat9:
@dkm good to merge?

PR #4710, opened 2026-07-16
gccrs: Run block cleanup for unlabeled break and continue

Lishin1215:
This PR builds on the existing `HIR::BlockExpr` `TRY_FINALLY_EXPR` cleanup infrastructure. 
It keeps unlabeled `break` and `continue` inside the block body, so block cleanup 
can run when they leave scopes.

It covers these unlabeled cases:

1. `continue`

```rust
loop {
    let _x = Droppable;
    continue;
}
```

2. tail-position `continue`

```rust
loop {
    let _x = Droppable;
    continue
}
```

3. `break`

```rust
loop {
    let _x = Droppable;
    break;
}
```

4. tail-position `break`

```rust
loop {
    let _x = Droppable;
    break
}
```

5. `break` with a value

```rust
let x = loop {
    let _x = Droppable;
    break 123;
};
```

The tests use `loop` instead of `while` to keep the coverage focused on cleanup behavior.

---
**Follow-up work:**
- Labeled `break` / `continue`
- Tail-position `return` cleanup

**While testing this change, I also found a separate issue that should be
tracked outside this PR:**

1. A loop-body local can be dropped twice in some cases. This likely needs 
   drop/initialized flags.

```rust
fn test_continue_double_drop() {
    let mut done = false;

    loop {
        if done {
            break;
        }

        let _outer = Droppable;

        {
            let _inner = Droppable;
            done = true;
            continue;
        }
    }
}
```

### DIFF
--- a/gcc/rust/backend/rust-compile-block.cc
+++ b/gcc/rust/backend/rust-compile-block.cc
@@ -25,6 +25,15 @@
 namespace Rust {
 namespace Compile {
 
+static bool
+is_control_flow_expr (HIR::Expr &expr)
+{
+  HIR::Expr::ExprType expr_type = expr.get_expression_type ();
+
+  return expr_type == HIR::Expr::ExprType::Break
+	 || expr_type == HIR::Expr::ExprType::Continue;
+}
+
 CompileBlock::CompileBlock (Context *ctx, Bvariable *result)
   : HIRCompileBase (ctx), translated (nullptr), result (result)
 {}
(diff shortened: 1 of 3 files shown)


gcc/rust/backend/rust-compile-block.cc
```diff
@@ -62,17 +71,25 @@ CompileBlock::visit (HIR::BlockExpr &expr)
...
-	  location_t locus = expr.get_final_expr ().get_locus ();
+	  location_t locus = final_expr.get_locus ();
 	  tree result_reference = Backend::var_expression (result, locus);
 
 	  tree assignment
 	    = Backend::assignment_statement (result_reference, compiled_expr,
 					     expr.get_locus ());
 	  ctx->add_statement (assignment);
 	}
+      else if (compiled_expr != nullptr && is_control_flow_expr (final_expr)
```
Lishin1215, line 87: This handles tail-position control flow like `{ break }` and `{ continue }`.
The generated jump still needs to be added to the block body, otherwise it is
created but never emitted into the final block.

Lishin1215:
I also opened an issue for the double-drop case I mentioned in the PR description:
https://github.com/Rust-GCC/gccrs/issues/4717

P-E-P (approved):
Great work!

PR #4714, opened 2026-07-20
gccrs: Backport `cfg_select!` macro

Polygonalr:
This macro is used in newer versions of `core` to compile `c_char`, that is used to compile `CStr` which has widespread use in Rust for Linux.

Aside from reviewing my code, let me know how I should properly gate this backported feature (e.g. maybe a compilation flag similar to `-frust-c-style-string-literals` from my `CStr` PR?), and whether there could be any improvements to the test cases.


powerboat9:
I wouldn't think it would need to be gated, since it only effects code which would error if `cfg_select!` wasn't built-in?

Polygonalr:
True that, though just want to confirm whether we'll need to gate it since it's a 1.95.0 feature (actually released very recently!).

CohenArthur:
Since this is a builtin macro, we will need to do something similar to `offset_of!()`. The issue will be that the `core` library we will use to compile the kernel will be 1.49, and will not have a builtin definition for the macro, while RfL will assume that it is present.

gcc/testsuite/rust/execute/torture/cfg_select2.rs
```diff
@@ -0,0 +1,31 @@
...
+}
+
+fn main() -> i32 {
+    cfg_select!
+    {
+        target_os = "windows" => {
+            unsafe {
+                let a = "windows\n\0";
+                printf(a as *const str as *const i8);
+            }
```
CohenArthur: isn't this test going to fail on Windows platforms? or am I overthinking this? Maybe we can turn this into something more reproducible with a `-frust-cfg=` option for the test?

CohenArthur (approved):

gcc/testsuite/rust/execute/torture/cfg_select2.rs
Polygonalr: You're probably overthinking this, I actually just assumed that our tests runs on Linux platforms when I was writing up the tests :grimacing:. I've updated them the tests, see whether they make more sense now.

(Now to think of it, we used to have a darwin tester in our CI that was removed a while ago too, it would have failed cfg_select1.rs)

CohenArthur:
For the two tests @Polygonalr I would rather have something we control rather than rely on target-specific variables, just for peace of mind. Something like this where we check a specific cfg seems sturdier to me:

```rust
// { dg-additional-options "-frust-compat-version=1.90 -frust-cfg=A=\"foo\"" }
// { dg-output "wildcard\r*\n" }
#![feature(no_core)]
#![no_core]

extern "C" {
    fn printf(s: *const i8, ...);
}

fn main() -> i32 {
    cfg_select! {
        A = "bar" => {
            unsafe {
                let a = "none\n\0";
                printf(a as *const str as *const i8);
            }
        },
        _ => {
            unsafe {
                let a = "wildcard\n\0";
                printf(a as *const str as *const i8);
            }
        }
    }
    return 0;
}
```

```rust
// { dg-additional-options "-frust-compat-version=1.90 -frust-cfg=A=\"foo\"" }
// { dg-output "pass\r*\n" }
#![feature(no_core)]
#![no_core]

extern "C" {
    fn printf(s: *const i8, ...);
}

fn main() -> i32 {
    cfg_select! {
        A = "foo" => {
            unsafe {
                let a = "pass\n\0";
                printf(a as *const str as *const i8);
            }
        }
        _ => {
            unsafe {
                let a = "fail\n\0";
                printf(a as *const str as *const i8);
            }
        }
    }
    return 0;
}
```

gcc/testsuite/rust/execute/torture/cfg_select2.rs
CohenArthur: On top of our CI, we have a lot of people in the wider GCC community that run our testsuite in weird configurations and on weird platforms, so I wouldn't put anything past them haha

gcc/rust/expand/rust-macro-builtins.cc
```diff
@@ -164,6 +165,8 @@ std::unordered_map<std::string, AST::MacroTranscriberFunc>
     /* offset_of is not declared in Rust 1.49 but still needed for
        Rust-for-Linux, so we still create a transcriber and warn the user */
     {"offset_of", MacroBuiltin::offset_of_handler},
+    /* cfg_select! is also not delcared in Rust 1.49 but also needed for RfL */
```
CohenArthur:
```suggestion
    /* cfg_select! is also not declared in Rust 1.49 but also needed for RfL */
```

gcc/rust/expand/rust-macro-expand.cc
```diff
@@ -329,6 +329,24 @@ MacroExpander::expand_invoc (AST::MacroInvocation &invoc,
...
+  if (assume_builtin_cfg_select)
+    {
+      fragment = MacroBuiltin::cfg_select_handler (invoc.get_locus (),
+						   invoc_data, semicolon)
+		   .value_or (AST::Fragment::create_empty ());
+
+      set_expanded_fragment (std::move (fragment));
+
+      return;
+    }
```
CohenArthur, line 348: perfect, if you want in a later PR we can think about factoring this as I expect we might add more assumed-builtin macros later on

CohenArthur (approved):
Minor nit but looks great to me, well done :)

CohenArthur (approved):
thank youuuuu this is great :D well done!

PR #4715, opened 2026-07-21
gccrs: Fix deriving with generic const args

powerboat9:
This doesn't fix an issue with the typechecker, so the included test only compiles up to the typechecking phase.

@philberty everything look good? This PR shouldn't be causing issues with the typechecker?


CohenArthur (approved):
LGTM :)

CohenArthur:
I can't speak for the typechecking part but the deriving part looks :+1:

Issue #4712, opened 2026-07-19
Make function-scope drop aligns to rustc


Lishin1215:
### Summary
Compared with rustc 1.49.0
gccrs and rustc have similar drop order, but their underlying structure is different.
This issue tracks making gccrs function drop scopes more like rustc.

### Background
The similar part is,
parameter drops are registered before body locals are lowered, so the final drop order becomes 
body locals first, then parameters.

  - In rustc,
  
  1. `construct_fn` creates an `arg_scope` for function parameters
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/mod.rs#L608-L628
  
  2. adds the parameter drops to that scope using `schedule_drop`:
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/mod.rs#L882-L888
  
  3. Body/local scopes are created while lowering the block, and local drops are scheduled into 
  those local scopes:
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/mod.rs#L943-L944
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/block.rs#L28-L43
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/block.rs#L104-L132
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/matches/mod.rs#L357-L375
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/matches/mod.rs#L561-L565

  4. When a scope is popped, rustc emits drops for the current top scope before popping it, 
  so inner body/local scopes are dropped before the outer `arg_scope`
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/block.rs#L200-L204
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/scope.rs#L539-L542
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/scope.rs#L556-L568
  https://github.com/rust-lang/rust/blob/1.49.0/compiler/rustc_mir_build/src/build/scope.rs#L656-L672

  Both parameter and local drops go through `schedule_drop`, which stores the drop 
in the matching scope's drop list:
[...]

PR #4718, opened 2026-07-23
gccrs: Separate argument and function body drop scopes

Lishin1215:
This PR makes gccrs closer to rustc by addressing the two differences described in #4712.

1. Registration timing

Before this patch, gccrs saved parameter drop candidates in a temporary vector. It registered them after creating the function block.

Now, gccrs creates the argument scope before compiling the parameters. 
Parameter drops are registered directly in this scope.

2. Scope structure

Before this patch, parameter drops and body local drops were stored in the same scope.
Now, gccrs uses two nested scopes. 
The outer scope stores parameter drops, and the inner scope stores body local drops.
The body scope is cleaned up first, so body locals are dropped before function parameters.


CohenArthur (approved):
This looks good to me, well done @Lishin1215! Good work!

P-E-P (approved):

PR #4726, opened 2026-07-29
gccrs: Handle ambiguous glob imports

powerboat9:


gcc/rust/resolve/rust-rib.cc
```diff
@@ -174,6 +174,40 @@ Rib::insert (std::string name, Definition def)
...
+  bool dirty = false;
+
+  const std::vector<NodeId> *ids_src;
+
+  if (!def.ids_shadowable.empty ())
+    ids_src = &def.ids_shadowable;
+  else if (!def.ids_non_shadowable.empty ())
+    ids_src = &def.ids_non_shadowable;
+  else
+    ids_src = &def.ids_globbed;
```
CohenArthur, line 189: Is there any reference for this in rustc source code or the Rust ref somewhere? or is it just due to how our specific NR algorithm works?

gcc/rust/resolve/rust-rib.h
powerboat9, line 156: It's the resolution order we use here. I think it does align with rustc, but I don't have anything specific to back that up.

gcc/rust/resolve/rust-rib.cc
powerboat9: I think it's just how our NR works, but it should align with rustc.

powerboat9:
Merging this, since it's been reviewed a bit, is blocking compilation of core, and should at the very least increase the correctness of our glob import handling.

Issue #4725, opened 2026-07-28
Identifier variant of enum incorrectly triggers duplicate identifier binding error


CohenArthur:
```rust
#![feature(no_core)]
#![feature(lang_items)]
#![feature(rustc_attrs)]
#![no_core]

enum Option {
    Some,
    None,
}

fn test_match(a: Option, b: Option) {
    match (a, b) {
        (None, None) => {}
        _ => {}
    }
}
```

Polygonalr:
investigated this out of boredom, this error is just because `None` is not imported as a prelude, fixable if we just add `use Option::None;` to the test code

`core` does this with [`use prelude::v1::*;`](https://github.com/Rust-GCC/gccrs/blob/master/libgrust/rustc-lib/core/src/lib.rs#L167), which includes [`pub use crate::option::Option::{self, None, Some};`](https://github.com/Rust-GCC/gccrs/blob/master/libgrust/rustc-lib/core/src/prelude/v1.rs#L43).

powerboat9:
Yeah, I think it was #4729 that fixed this

PR #4727, opened 2026-07-29
gccrs: Catch likely compilation error with `None`

powerboat9:
While technically allowed, matching with an irrefutable identifier pattern `None` is almost certainly an in-progress miscompilation of a refutable path pattern `None`.


powerboat9:
Related to #4725

CohenArthur (changes requested):
I don't think we should add this, we need to fix it shortly as this is an issue for compiling `core` 1.49. If we add it it will be removed extremely quickly and doesn't help with the issue I'm currently facing

powerboat9:
Marking this as a draft, since we should be on the verge of catching ~all such miscompilations

PR #4728, opened 2026-07-29
gccrs: Emit E0259 and E0260 for duplicate extern crate names

### LINKED ISSUE #4679: Compiler does not error out on extern crate defined multiple times
### Summary

The compiler should error out when multiple `extern crate` with the same name are used.

The core crate is injected automatically without the `#![no_core]` context, this means it will fail later with multiple same lang items defined if we ever import it a second time with `extern crate core`.

### Reproducer

I tried this code:

```rust
extern crate extern_crate;
extern crate extern_crate;
```

- [ ] Nightly

### Godbolt link

https://godbolt.org/z/3z4hfqxrz

### Actual behavior

The current behavior is no error.

### Expected behavior

I expected to see an error telling me that the extern crate is imported multiple times.

### GCC Version

e540b62

u7k4rs6:
Fixes #4679
Extern crate declarations never inserted their bound name into any namespace, so the existing duplicate-name check could not fire for them and both duplicate extern crates and item/extern crate collisions were accepted silently. Insert the bound name into the type namespace and select the error code based on whether the colliding declarations are extern crates. E0428 is unchanged for every other item kind.

gcc/rust/ChangeLog:

	* resolve/rust-toplevel-name-resolver-2.0.cc (TopLevel::visit (ExternCrate)): New function definition. (TopLevel::check_multiple_insertion_error): Select E0259, E0260 or E0428 based on the colliding declarations. (TopLevel::insert_or_error_out): Record extern crate nodes and forward the new parameter. (TopLevel::insert_enum_variant_or_error_out): Update call.
	* resolve/rust-toplevel-name-resolver-2.0.h (TopLevel::visit (ExternCrate)): New function declaration. (TopLevel::check_multiple_insertion_error) (TopLevel::insert_or_error_out): New parameter. (TopLevel::extern_crate_nodes): New member.

gcc/testsuite/ChangeLog:

	* rust/link/extern_crate_alias_0.rs: New test.
	* rust/link/extern_crate_alias_1.rs: New test.
	* rust/link/extern_crate_dup_0.rs: New test.
	* rust/link/extern_crate_dup_1.rs: New test.
	* rust/link/extern_crate_item_conflict_0.rs: New test.
	* rust/link/extern_crate_item_conflict_1.rs: New test.

in the git commit message.

- \[ ] Run `clang-format`

order to get feedback.
---

### DIFF
--- a/gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
+++ b/gcc/rust/resolve/rust-toplevel-name-resolver-2.0.cc
@@ -50,8 +50,18 @@ TopLevel::check_multiple_insertion_error (
     {
       rich_location rich_loc (line_table, locus);
       rich_loc.add_range (node_locations[result.error ().existing]);
+      auto &mappings = Analysis::Mappings::get ();
+      ErrorCode code;
+      if (mappings.is_extern_crate (node_id)
+	  && mappings.is_extern_crate (result.error ().existing))
+	code = ErrorCode::E0259;
+      else if (mappings.is_extern_crate (node_id)
+	       || mappings.is_extern_crate (result.error ().existing))
+	code = ErrorCode::E0260;
+      else
+	code = ErrorCode::E0428;
 
-      rust_error_at (rich_loc, ErrorCode::E0428, "%qs defined multiple times",
+      rust_error_at (rich_loc, code, "%qs defined multiple times",
 		     identifier.as_string ().c_str ());
     }
 }
@@ -125,6 +135,17 @@ TopLevel::visit (AST::Trait &trait)
   insert_or_error_out (trait.get_identifier (), trait, Namespace::Types);
 
   DefaultResolver::visit (trait);
+}
+
+void
+TopLevel::visit (AST::ExternCrate &crate)
+{
+  auto &name = crate.has_as_clause () ? crate.get_as_clause ()
+				      : crate.get_referenced_crate ();
+  Analysis::Mappings::get ().insert_extern_crate_id (crate.get_node_id ());
+  insert_or_error_out (name, crate, Namespace::Types);
+
+  DefaultResolver::visit (crate);
 }
 
 void
--- a/gcc/rust/resolve/rust-toplevel-name-resolver-2.0.h
+++ b/gcc/rust/resolve/rust-toplevel-name-resolver-2.0.h
@@ -167,6 +167,7 @@ private:
 
   void visit (AST::Module &module) override;
   void visit (AST::Trait &trait) override;
+  void visit (AST::ExternCrate &crate) override;
   void maybe_insert_big_self (AST::Impl &impl) override;
   void visit (AST::TraitItemType &trait_item) override;
   void visit (AST::MacroRulesDefinition &macro) override;
--- a/gcc/rust/util/rust-hir-map.cc
+++ b/gcc/rust/util/rust-hir-map.cc
@@ -1180,6 +1180,18 @@ Mappings::is_module (NodeId id)
   return module_ids.find (id) != module_ids.end ();
 }
 
+void
+Mappings::insert_extern_crate_id (NodeId id)
+{
+  extern_crate_ids.insert (id);
+}
+
+bool
+Mappings::is_extern_crate (NodeId id)
+{
+  return extern_crate_ids.find (id) != extern_crate_ids.end ();
+}
+
 tl::optional<AST::GlobContainer *>
 Mappings::lookup_glob_container (NodeId id)
 {
--- a/gcc/rust/util/rust-hir-map.h
+++ b/gcc/rust/util/rust-hir-map.h
@@ -328,6 +328,9 @@ public:
   void insert_module_id (NodeId);
   bool is_module (NodeId id);
 
+  void insert_extern_crate_id (NodeId);
+  bool is_extern_crate (NodeId id);
+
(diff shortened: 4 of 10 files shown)


u7k4rs6:
Heads-up for reviewers: this is going to change shape.

Discussed with @P-E-P on Zulip. This PR currently distinguishes E0259 from
E0260 using a set of extern crate NodeIds kept in `TopLevel`, alongside
`node_locations`. He'd rather `DuplicateNameError` itself carried the
information, which is what its own TODO anticipates.

I'll push that as a separate commit on this branch rather than opening a new
PR, per his suggestion. The side table and the parameter threading come back
out; the tests are unaffected either way, since they match on message text.

gcc/rust/resolve/rust-rib.cc
```diff
@@ -22,8 +22,9 @@
 namespace Rust {
 namespace Resolver2_0 {
 
-Rib::Definition::Definition (NodeId id, Mode mode, bool enum_variant)
-  : enum_variant (enum_variant)
+Rib::Definition::Definition (NodeId id, Mode mode, bool enum_variant,
+			     bool is_extern_crate)
+  : enum_variant (enum_variant), is_extern_crate (is_extern_crate)
```
P-E-P: I don't think we should store the extern crate status within the Definition structure. This structure will often be used for non crate definitions, what could be done instead is an extern crate mapping within rust-hir-mapping with the node id as a key.

gcc/rust/resolve/rust-rib.h
```diff
@@ -73,10 +73,15 @@ enum class Namespace
...
-  DuplicateNameError (std::string name, NodeId existing);
+  DuplicateNameError (std::string name, NodeId existing,
+		      bool existing_is_extern_crate = false,
+		      bool new_is_extern_crate = false);
 
   std::string name;
   NodeId existing;
+
+  bool existing_is_extern_crate;
+  bool new_is_extern_crate;
```
P-E-P: Those name are a bit confusing although I cannot find a better alternative right now. Would packing them in their own struct make it better ?

gcc/rust/resolve/rust-rib.cc
u7k4rs6: I've moved the extern-crate status into `Mappings` as a `NodeId`-keyed set, following the existing `insert_module_id` / `is_module` pattern. As a result, `rust-rib.h` and `rust-rib.cc` are back to their upstream state, and `git diff upstream/master` is empty for both files.

gcc/rust/resolve/rust-rib.h
u7k4rs6: The two bools no longer exist, so there's nothing left to pack. `check_multiple_insertion_error` now queries `Mappings` for both the existing node and the new node instead of carrying that state through `DuplicateNameError`.

One subtlety is the ordering in `visit (AST::ExternCrate)`: the node is recorded in `Mappings` before calling `insert_or_error_out`, since `check_multiple_insertion_error` runs inside that call and needs to know whether the new node is an extern crate.

u7k4rs6:
@P-E-P One question: your extern-crate name resolution changes landed recently. Would you prefer that I rebase this branch onto current `master` before you take another look, or is it fine to leave the current base?

P-E-P:
> @P-E-P One question: your extern-crate name resolution changes landed recently. Would you prefer that I rebase this branch onto current `master` before you take another look, or is it fine to leave the current base?

@u7k4rs6 I don't think the changes merged on upstream will  affect this PR much, the changes merged were related to lang items.
We usually rebase as soon as we can when a PR is not in a draft state anymore because the CI does not run on PRs with a big diff with master. So please rebase it :)

u7k4rs6:
@P-E-P Rebased onto master. No conflicts, and the testsuite is still clean locally.

P-E-P:
> @P-E-P Rebased onto master. No conflicts, and the testsuite is still clean locally.

This is great, could you squash your last commit within the other one ? Having a separate commit to undo something that has not been merged yet to master is only makes bisect longer and history grepping harder.

u7k4rs6:
@P-E-P Squashed into two commits. The combined commit ends up touching four files
rather than the ten the separate commit showed, since six of them netted to
zero and dropped out.

P-E-P:
You will probably need to update the changelog if you did not already do it.

u7k4rs6:
@P-E-P Already done, it went into the squashed commit's message and the ChangeLog
validator passes on both commits.

One thing while I'm here: by the same reasoning, commit 1 introduces the
extern_crate_nodes side table and commit 2 removes it, so bisecting to commit
1 still lands on something that never reaches master. Want me to fold them
into a single commit as well, or is the split between the diagnostics and
where the tracking lives worth keeping?

P-E-P:
> @P-E-P Already done, it went into the squashed commit's message and the ChangeLog validator passes on both commits.
> 
> One thing while I'm here: by the same reasoning, commit 1 introduces the extern_crate_nodes side table and commit 2 removes it, so bisecting to commit 1 still lands on something that never reaches master. Want me to fold them into a single commit as well, or is the split between the diagnostics and where the tracking lives worth keeping?

Following your description I would indeed squash both commits into one. I like to keep commits atomic but we must make sure that each commit brings value on its own and compiles correctly. If nobody gains anything from the commits being split in two we better squash them together.

u7k4rs6:
done !

P-E-P (approved):

PR #4729, opened 2026-07-29
gccrs: Generalize prelude import handling

powerboat9:
Replace some ad-hoc handling of prelude imports with a glob import in every module/crate.

### DIFF
--- a/gcc/rust/resolve/rust-default-resolver.cc
+++ b/gcc/rust/resolve/rust-default-resolver.cc
@@ -36,7 +36,10 @@ DefaultResolver::visit (AST::Crate &crate)
   if (!visited_crates.insert (crate.get_node_id ()).second)
     return;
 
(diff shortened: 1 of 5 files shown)


CohenArthur (commented):
This is for the lang prelude right? primitive types etc? Is this something rustc does as well or no? Why is this better than the current solution of checking the prelude if resolution fails? I am not against it, I think it is rather elegant, but yeah I don't know if this will cause issues in the future or not

powerboat9:
Nope, it's for the standard library prelude (https://doc.rust-lang.org/reference/names/preludes.html#r-names.preludes.std), which we currently only check when resolving identifier expressions (not path expressions, identifier patterns pending possible conversion to path patterns, etc).

powerboat9:
Merging this, since it's been reviewed a bit and is blocking late resolution of core.

PR #4730, opened 2026-07-30
gccrs: Add straight-line BIR drop style analysis

Lishin1215:
This patch adds an initial BIR Drop state analysis based on rustc's MIR
drop scheduling and Drop style classification.

It contains two commits:

**commit 1**
Fixes an incorrect BIR `IndexVec` size that could cause an out-of-bounds 
access when dumping BIR.

**commit 2**
Adds basic Drop state analysis for whole local variables in straight-line control flow. 
It adds Drop statements at scope exits and marks them as `Static` or `Dead`.

---
It covers these straight-line cases:

1. static local:

```rust
let x = Droppable;
```

```text
Drop(x): Static
```

2. whole-local move:

```rust
let x = Droppable;
let y = x;
```

```text
Drop(y): Static
Drop(x): Dead
```

3. Copy assignment:

```rust
let x = 1;
let y = x;
```

```text
Drop(y): Static
Drop(x): Static
```

4. Function argument:

```rust
fn f(x: Droppable) {}
```

```text
Drop(x): Static
```
---
**Current scope**

This patch supports:
- straight-line functions without branches or loops
- whole-local moves
- `copy` versus `move` assignments
- function arguments
- `Static` and `Dead` Drop styles

The following are left for follow-up patches:
- branches and conditional Drops
- loops
- partial moves
- `needs_drop` filtering
- backend integration

BIR Drop statements are currently emitted for all locals as analysis points.
Filtering them using `needs_drop` will be added separately.

### DIFF
--- a/gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
+++ b/gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
@@ -1,0 +1,114 @@
+// Copyright (C) 2026 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#include "rust-bir-drop-analysis.h"
+#include "rust-bir.h"
+
+namespace Rust {
+namespace BIR {
+
+void
+DropAnalysis::analyze (Function &function)
+{
+  std::vector<BasicBlockId> block_order;
+  std::set<BasicBlockId> visited;
+
+  BasicBlockId current = ENTRY_BASIC_BLOCK;
+
+  while (current != INVALID_BB)
+    {
+      // A repeated block indicates a cycle in straight-line control flow.
+      if (!visited.insert (current).second)
+	return;
+
+      block_order.push_back (current);
+
+      const BasicBlock &block = function.basic_blocks[current];
+
+      if (block.successors.empty ())
+	break;
+
+      if (block.successors.size () != 1)
+	return;
+
+      current = block.successors.front ();
+    }
+
+  std::vector<bool> initialized (function.place_db.size (), false);
+
+  for (PlaceId argument : function.arguments)
+    initialized[argument.value] = true;
+
+  for (BasicBlockId block_id : block_order)
+    {
+      BasicBlock &block = function.basic_blocks[block_id];
+
+      for (Statement &statement : block.statements)
+	{
+	  PlaceId place = statement.get_place ();
+
+	  switch (statement.get_kind ())
+	    {
+	    case Statement::Kind::STORAGE_LIVE:
+	      initialized[place.value] = false;
+	      break;
+
+	    case Statement::Kind::ASSIGNMENT:
+	      {
+		PlaceId lhs = place;
+		AbstractExpr &expr = statement.get_expr ();
+
+		if (expr.get_kind () == ExprKind::ASSIGNMENT)
+		  {
+		    PlaceId rhs = static_cast<Assignment &> (expr).get_rhs ();
+		    const Place &rhs_place = function.place_db[rhs];
+
+		    if (rhs_place.kind == Place::VARIABLE
+			&& rhs_place.should_be_moved ())
+		      initialized[rhs.value] = false;
+		  }
+
+		initialized[lhs.value] = true;
+		break;
+	      }
+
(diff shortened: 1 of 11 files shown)


gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
```diff
@@ -0,0 +1,114 @@
...
+	return;
+
+      block_order.push_back (current);
+
+      const BasicBlock &block = function.basic_blocks[current];
+
+      if (block.successors.empty ())
+	break;
+
+      if (block.successors.size () != 1)
```
Lishin1215, line 46: I limit this first version to straight-line control flow.
If the CFG contains branches or loops, the analysis returns and leaves the Drops unclassified.

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
```diff
@@ -0,0 +1,114 @@
...
+		  {
+		    PlaceId rhs = static_cast<Assignment &> (expr).get_rhs ();
+		    const Place &rhs_place = function.place_db[rhs];
+
+		    if (rhs_place.kind == Place::VARIABLE
+			&& rhs_place.should_be_moved ())
+		      initialized[rhs.value] = false;
+		  }
+
+		initialized[lhs.value] = true;
```
Lishin1215, line 86: I mark the lhs as initialized after an assignment.
I also mark a non-Copy variable rhs as uninitialized because its value was moved, and leave a copy assignment rhs initialized.

This first version only handles direct whole-local moves.

gcc/rust/checks/errors/borrowck/rust-bir-place.h
```diff
@@ -209,7 +209,7 @@ template <typename I, typename T> class IndexVec
 
 public:
   IndexVec () = default;
-  IndexVec (size_t size) { internal_vector.reserve (size); }
+  IndexVec (size_t size) : internal_vector (size) {}
```
Lishin1215, line 212: I changed the sized constructor to create the requested number of elements.
The old constructor only reserved capacity, so direct indexing during the BIR dump could access an empty vector.

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
```diff
@@ -0,0 +1,114 @@
...
+      if (block.successors.size () != 1)
+	return;
+
+      current = block.successors.front ();
+    }
+
+  std::vector<bool> initialized (function.place_db.size (), false);
+
+  for (PlaceId argument : function.arguments)
+    initialized[argument.value] = true;
```
Lishin1215, line 55: I handle function argument Drops separately because arguments are already 
initialized at function entry and do not use the normal local `StorageLive` 
and `StorageDead` handling.
I emit their Drops in reverse declaration order before the function returns.

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
```diff
@@ -0,0 +1,114 @@
...
+#include "rust-bir.h"
+
+namespace Rust {
+namespace BIR {
+
+void
+DropAnalysis::analyze (Function &function)
+{
+  std::vector<BasicBlockId> block_order;
+  std::set<BasicBlockId> visited;
```
P-E-P: Do we really need an ordering in here ? Wouldn't a unordered_set be better ?

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.h
```diff
@@ -0,0 +1,42 @@
...
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#ifndef RUST_BIR_DROP_ANALYSIS_H
+#define RUST_BIR_DROP_ANALYSIS_H
+
+namespace Rust {
+namespace BIR {
+
+struct Function;
```
P-E-P: Couldn't this forward declaration be avoided ?

gcc/rust/checks/errors/borrowck/rust-bir-place.h
P-E-P: I would add a comment on this line. This vector ctor is a bit misleading/tricky (many people think this only reserves the space) and it will either turn a warning in reader's mind or be misused. We need to be explicit about the elements default construction.

P-E-P (changes requested):
I'm not entirely familiar with BIR but this looks great. I've put a few suggestions.

gcc/rust/checks/errors/borrowck/rust-bir-place.h
Lishin1215: Thanks! I added a comment to make that clear.

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
Lishin1215: You're right, I think we don’t need ordering here. I changed it to `unordered_set`.

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.h
Lishin1215: Yes, I changed this to include `rust-bir.h` directly.

P-E-P (approved):
LGTM

PR #4743, opened 2026-08-01
gccrs: Test `libcore` compilation until lowering

### LINKED ISSUE #4692: [Meta] Late name resolution of `libcore 1.49`
- [x] #1922
- [x] #4664
- [x] #4699
- [x] #4702

powerboat9:
This slows down the testsuite a bit (4 minutes on my PC, without parallelization, about 10 minutes on github CI runners), but we need libcore to compile (and therefore need to be verifying that it does).

Closes #4692

### DIFF
--- a/gcc/testsuite/rust/core/core.exp
+++ b/gcc/testsuite/rust/core/core.exp
@@ -28,9 +28,9 @@ dg-init
 set saved-dg-do-what-default ${dg-do-what-default}
 
 set dg-do-what-default "compile"
-set individual_timeout 600
+set individual_timeout 1200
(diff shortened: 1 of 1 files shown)


P-E-P (approved):
I thought we still had a few issues around late name res. I would be happy to be proven wrong.

Patch is good though.

powerboat9:
> I thought we still had a few issues around late name res. I would be happy to be proven wrong.
> 
> Path is good though.

#4746, for example, fixes an issue in expansion. I'm just tracking based on what phase the compiler starts erroring/ICEing in.

PR #4745, opened 2026-08-01
gccrs: Implement basic support for #[repr(simd)] structs

Polygonalr:
Only type-checking support is implemented as I don't think we'll need strict SIMD representation for RfL at the moment - simd representation structs are still compiled as regular structs.

Note that this implementation is specific to 1.49.0, `#[repr(simd)]` structs uses a single array field from 1.50.0 onwards, I wrote a TODO comment in my changes for that.

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-base.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-base.cc
@@ -506,23 +506,24 @@ TypeCheckBase::parse_repr_options (const AST::AttrVec &attrs, location_t locus)
 	      continue;
 	    }
 
-	  const std::string inline_option = items.at (0)->as_string ();
+	  const std::string repr_option = items.at (0)->as_string ();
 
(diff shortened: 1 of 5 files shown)


gcc/rust/typecheck/rust-hir-type-check-base.cc
```diff
@@ -546,6 +547,7 @@ TypeCheckBase::parse_repr_options (const AST::AttrVec &attrs, location_t locus)
 			    || inline_option.compare ("u64") == 0
 			    || inline_option.compare ("u128") == 0);
 	      is_transparent = inline_option.compare ("transparent") == 0;
+	      is_simd = inline_option.compare ("simd") == 0;
```
CohenArthur: This variable should really be renamed haha, `repr_option` would be better

CohenArthur (approved):
LGTM

P-E-P (approved):

PR #4748, opened 2026-08-02
gccrs: Use BIR Drop analysis in backend cleanup

Lishin1215:
This patch connects the straight-line BIR Drop analysis to the existing backend cleanup.

It uses the BIR analysis result to skip backend Drops classified as `Dead`.
`Static` and unclassified Drops keep the existing cleanup behavior.

---
So now, for a basic move case like:

```rust
fn whole_move() {
    let x = Droppable;
    let y = x;
}
```

The backend uses the results `Drop(y): Static` and `Drop(x): Dead` to
skip `Drop(x)` while keeping `Drop(y)`.

A normal initialized local is still dropped.

### DIFF
--- a/gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
+++ b/gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
@@ -18,6 +18,7 @@
 
 #include "rust-bir-drop-analysis.h"
 #include "rust-bir.h"
+#include "rust-hir-map.h"
 
 #include <unordered_set>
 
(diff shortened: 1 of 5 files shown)


gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
```diff
@@ -0,0 +1,150 @@
...
+
+		  if (dropped_place.kind == Place::VARIABLE)
+		    {
+		      auto hir_id
+			= Analysis::Mappings::get ().lookup_node_to_hir (
+			  static_cast<NodeId> (
+			    dropped_place.variable_or_field_index));
+
+		      if (hir_id.has_value ())
+			definitely_dead.insert (hir_id.value ());
```
Lishin1215, line 127: I record the `HirId` when a whole-local Drop is classified as `Dead`.

The BIR analysis identifies the local using a `PlaceId`, while the existing
backend Drop candidates use `HirId`. This mapping lets the backend find the
corresponding analysis result.

`BIR PlaceId -> HirId -> backend DropCandidate`

I currently limit this to whole variables because projections and partial
moves are not supported yet.

philberty: that seems sensible

philberty (approved):
LGTM

PR #4749, opened 2026-08-03
gccrs: Move `SlicePattern` `RestPattern` handling

powerboat9:
This allows the AST to represent `SlicePattern` instances with more than one `RestPattern` child, which allows us to handle invalid `SlicePattern` instances which would be cfg'd out and mostly avoid dealing with the implications of a `RestPattern` until HIR lowering.

This should be useful for fixing the `SlicePattern` issue with `core`.

### DIFF
--- a/gcc/rust/ast/rust-ast-collector.cc
+++ b/gcc/rust/ast/rust-ast-collector.cc
@@ -3056,35 +3056,11 @@ TokenCollector::visit (GroupedPattern &pattern)
 }
 
 void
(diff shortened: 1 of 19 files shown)


Polygonalr:
LGTM, rustc only throws errors for multiple rest patterns in slice patterns during AST->HIR lowering anyways (didn't realise that when I was implementing it last year), so good that we match this behaviour too. Great that quite a bit of code was trimmed as a result too.

CohenArthur (approved):
LGTM!

PR #4751, opened 2026-08-03
forever-stack: Add dfs_cache

CohenArthur:
We do a *lot* of depth-first-search during name resolution, so having a temporary cache for the results speeds up the compilation of core by quite a lot. This is just a simple version that can be improved.

On my machine, this shaves off almost 3 minutes from compiling core.

gcc/rust/ChangeLog:

	* resolve/rust-forever-stack.h: Declare new cache API.
	* resolve/rust-forever-stack.hxx: Implement it and use it within dfs_cache.

Putting it in draft as I'm not sure we should be thinking about optimization already, but this has been sitting in my branch for a while.

### DIFF
--- a/gcc/rust/resolve/rust-forever-stack.h
+++ b/gcc/rust/resolve/rust-forever-stack.h
@@ -796,6 +796,10 @@ public:
   tl::optional<const Node &> dfs_node (const Node &starting_point,
 				       NodeId to_find) const;
 
+  std::unordered_map<NodeId, Node &> dfs_cache;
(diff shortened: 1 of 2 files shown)


P-E-P (approved):

P-E-P:
As long as we do not make the compiler more complex or change it's behavior specifically to match some speed criteria I'm fine with merging this. The behavior of  a cache is well known and easy enough, it won't impact development.

We still have a lot of low hanging fruits that could shave off a few seconds, unwanted copies etc... We should probably fix those before attempting anything that modifies the compiler's structure.

PR #4752, opened 2026-08-03
Better handle tool attributes

CohenArthur:
attributes: Check if an attribute is a tool attribute

Tool attributes are allowed in Rust code, and should be ignored by the
compiler and handled by the specified tools instead.

gcc/rust/ChangeLog:

	* util/rust-attributes.h: Change Attributes::is_known API to return an enum, add
	AttributeKnowledge result enum.
	* util/rust-attributes.cc (Attributes::is_known): Check for tool attributes and
	adapt return type.
	* hir/rust-ast-lower-base.cc (ASTLoweringBase::handle_outer_attributes): Use new
	API and ignore tool attributes.

gcc/testsuite/ChangeLog:

	* rust/compile/tool-attribute1.rs: New test.

### DIFF
--- a/gcc/testsuite/rust/compile/tool-attribute1.rs
+++ b/gcc/testsuite/rust/compile/tool-attribute1.rs
@@ -1,0 +1,22 @@
+#![feature(no_core)]
+#![feature(lang_items)]
+#![feature(rustc_attrs)]
+#![no_core]
+
+#[rustfmt::skip]
+pub fn foo() {}
+
+#[clippy::something::useful]
+#[diagnostic::diagnose_this]
+struct Boo;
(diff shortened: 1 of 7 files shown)


gcc/testsuite/rust/compile/tool-attribute1.rs
```diff
@@ -0,0 +1,7 @@
+#![feature(no_core)]
+#![feature(lang_items)]
+#![feature(rustc_attrs)]
+#![no_core]
+
+#[rustfmt::skip]
```
P-E-P, line 6: Could you add more tests (one per tool ?) ?

PR #4757, opened 2026-08-04
Fake PR to test the upstreaming of commit up to gerris/rebase-upstream/2026-08-04

dkm:
This is a fake PR, not meant to be merged. It tries to merge commits to upstream with an upstream base branch dkm/upstream-base.
We're only interested by the CI results.
-- [gerris](https://github.com/Rust-GCC/gerris) 🦀


CohenArthur:
Hmm, I wonder why the CI doesn't run on this. We had some issues with changelogs in this batch where the CI should have caught it. I fixed them right before pushing so there's no issue in the end, but it would be nice if it got caught ahead of time

dkm:
> Hmm, I wonder why the CI doesn't run on this. We had some issues with changelogs in this batch where the CI should have caught it. I fixed them right before pushing so there's no issue in the end, but it would be nice if it got caught ahead of time

Oh? The CI is not executed on the upstream branch, despite the name, it's in my todo since... But it's pretty easy and quick to fix...

But, the commit should have been flagged before merging. Do you have pointers to the commits?

dkm:
(closing, as you've sent everything upstream already)

PR #4773, opened 2026-08-08
gccrs: Implement compilation for `let <StructPattern>`

Polygonalr:


gcc/testsuite/rust/compile/let-structpattern.rs
```diff
@@ -0,0 +1,15 @@
+#![feature(no_core)]
+#![no_core]
+
+// TODO the warnings below shouldn't be emitted...?
+struct Foo {
+    x: i32, // { dg-warning "field is never read: .x." }
+    y: i32 // { dg-warning "field is never read: .y." }
```
CohenArthur, line 7: You can either add a bogus annotation here (I can't remember the exact directive) or keep it that way an open a new issue on GitHub so we remember to fix it

CohenArthur (commented):
At a first glance this looks good to me, I'll leave it to someone else to review and merge though as I'll be away for a few days and don't want to block this. Thank you!

gcc/testsuite/rust/compile/let-structpattern.rs
Polygonalr: The `dg-bogus` annotation seemed to be the opposite of `dg-warning` and `dg-error` (test case fails if compiler emits whatever message stated in the `dg-bogus` annotation?), so I don't think it's the correct way to do so.

P-E-P: Can't you pair it with a * xfail selector ? https://gcc.gnu.org/onlinedocs/gccint/Directives.html#Verify-compiler-messages

gcc/testsuite/rust/execute/torture/let-structpattern.rs
```diff
@@ -0,0 +1,19 @@
...
+struct Bar (i32, i32);
+
+fn main() -> i32 {
+    let mut my_foo = Foo{x: 32};
+    let mut my_bar = Bar(32, 64);
+    let Foo {x} = my_foo;
+    let Foo {x: another_x} = my_foo;
+    my_foo.x = 16;
+    let Foo {x: yet_another_x } = my_foo;
+    let Bar {0: w, ..} = my_bar;
```
P-E-P, line 17: should we add a binding pattern (@) test within a let ?

Polygonalr: Added a binding pattern to the execute test, see whether that is what you're looking for

P-E-P (approved):

PR #4775, opened 2026-08-11
gccrs: Handle conditional moves in BIR drop analysis

Lishin1215:
This patch extends the BIR Drop state analysis from straight-line control flow to conditional control flow.

It tracks whether each local variable may be `initialized` or `uninitialized` at the entry of each basic block. 

Take conditional move as example, 

              BB0
           initialize x
              /    \
             v      v
         BB1         BB2
        move x      no change
             \      /
              v    v
               BB3
              Drop(x)

For example,
```
BB0 -> BB1:
  (BB0): maybe_initialized(x)   = true
  (BB0): maybe_uninitialized(x) = false

  (BB1): reachable = false -> true
  (BB1): maybe_initialized(x)   = false -> true 
  (BB1): maybe_uninitialized(x) = false -> false 

BB1 -> BB3
  (pass through `let y = x`)
  (BB1): maybe_initialized(x) = true -> false 
  (BB1): maybe_unitialized(x) = false -> true
     
  (Assume BB2 has already reached BB3)
  (BB3): maybe_initialized(x)   = true -> true 
  (BB3): maybe_uninitialized(x) = false -> true 

BB3 -> END
  (BB3): maybe_initialized(x)   = true 
  (BB3): maybe_uninitialized(x) = true  
```
The analysis stops when the `worklist` becomes empty.

---

A Drop is classified as:

1. maybe initialized only:
1.1. Drop(x): Static
2. maybe uninitialized only:
2.1. Drop(x): Dead
3. maybe initialized and maybe uninitialized:
3.1. Drop(x): Conditional

---
It covers the following conditional control-flow cases:

1. conditional move:
```
  let x = A { i: 1 };

  if condition {
      let y = x;
  }
```
```
  Drop(x): Conditional
```
2. remains initialized after the join:
```
  let x = A { i: 1 };

  if condition {
      let y = 1;
  }
```
```
  Drop(x): Static
```

3. moved on both branches:
```
  let x = A { i: 1 };

  if condition {
      let y = x;
  } else {
      let z = x;
  }
```
```
  Drop(x): Dead
```

### DIFF
--- a/gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
+++ b/gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
@@ -1,0 +1,114 @@
+// Copyright (C) 2026 Free Software Foundation, Inc.
+
+// This file is part of GCC.
+
+// GCC is free software; you can redistribute it and/or modify it under
+// the terms of the GNU General Public License as published by the Free
+// Software Foundation; either version 3, or (at your option) any later
+// version.
+
+// GCC is distributed in the hope that it will be useful, but WITHOUT ANY
+// WARRANTY; without even the implied warranty of MERCHANTABILITY or
+// FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+// for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with GCC; see the file COPYING3.  If not see
+// <http://www.gnu.org/licenses/>.
+
+#include "rust-bir-drop-analysis.h"
+#include "rust-bir.h"
+
+namespace Rust {
+namespace BIR {
+
+void
+DropAnalysis::analyze (Function &function)
+{
+  std::vector<BasicBlockId> block_order;
+  std::set<BasicBlockId> visited;
+
+  BasicBlockId current = ENTRY_BASIC_BLOCK;
+
+  while (current != INVALID_BB)
+    {
+      // A repeated block indicates a cycle in straight-line control flow.
+      if (!visited.insert (current).second)
+	return;
+
+      block_order.push_back (current);
+
+      const BasicBlock &block = function.basic_blocks[current];
+
+      if (block.successors.empty ())
+	break;
+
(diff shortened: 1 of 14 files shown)


gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
```diff
@@ -0,0 +1,339 @@
...
+
+static void
+set_uninitialized (BlockInitializationState &state, PlaceId place)
+{
+  state.maybe_initialized[place.value] = false;
+  state.maybe_uninitialized[place.value] = true;
+}
+
+static bool
+merge_state (BlockInitializationState &into,
```
Lishin1215, line 82: This function combines states from incoming blocks.
Take conditional move as example, 

              BB0
           initialize x
              /    \
             v      v
         BB1      BB2
        move x   no change
             \      /
              v    v
               BB3
              Drop(x)

BB1 -> BB3
  (BB1): maybe_initialized(x) = false 
  (BB1): maybe_unitialized(x) =true
     
  (BB3): maybe_initialized(x)   = true || (BB1): maybe_initialized(x) -> true
  (BB3): maybe_uninitialized(x) = false ||  (BB1): maybe_unitialized(x) -> true

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
```diff
@@ -0,0 +1,338 @@
...
+}
+
+void
+DropAnalysis::analyze (Function &function)
+{
+  std::vector<BlockInitializationState> entry_states
+    = compute_entry_states (function);
+
+  // Keep the existing backend handling for straight-line CFGs.
+  const bool record_straight_line_backend_drops = is_straight_line (function);
```
Lishin1215, line 322: This is a specific bridge from BIR Drop analysis to the existing backend.
It stores the results using HirId and updates definitely_dead. 
It is only used for straight-line CFGs because the backend does not support Conditional Drops yet.

PR #4777, opened 2026-08-12
gccrs: Handle conditional moves in BIR drop analysis

Lishin1215:
This patch extends the BIR Drop state analysis from straight-line control flow to conditional control flow.

It tracks whether each local variable may be `initialized` or `uninitialized` at the entry of each basic block. 

Take conditional move as example, 

              BB0
           initialize x
              /    \
             v      v
         BB1         BB2
        move x      no change
             \      /
              v    v
               BB3
              Drop(x)

For example,
```
BB0 -> BB1:
  (BB0): maybe_initialized(x)   = true
  (BB0): maybe_uninitialized(x) = false

  (BB1): reachable = false -> true
  (BB1): maybe_initialized(x)   = false -> true 
  (BB1): maybe_uninitialized(x) = false -> false 

BB1 -> BB3
  (pass through `let y = x`)
  (BB1): maybe_initialized(x) = true -> false 
  (BB1): maybe_unitialized(x) = false -> true
     
  (Assume BB2 has already reached BB3)
  (BB3): maybe_initialized(x)   = true -> true 
  (BB3): maybe_uninitialized(x) = false -> true 

BB3 -> END
  (BB3): maybe_initialized(x)   = true 
  (BB3): maybe_uninitialized(x) = true  
```
The analysis stops when the `worklist` becomes empty.

---

A Drop is classified as:

1. maybe initialized only:
1.1. Drop(x): Static
2. maybe uninitialized only:
2.1. Drop(x): Dead
3. maybe initialized and maybe uninitialized:
3.1. Drop(x): Conditional

---
It covers the following conditional control-flow cases:

1. conditional move:
```
  let x = A { i: 1 };

  if condition {
      let y = x;
  }
```
```
  Drop(x): Conditional
```
2. remains initialized after the join:
```
  let x = A { i: 1 };

  if condition {
      let y = 1;
  }
```
```
  Drop(x): Static
```

3. moved on both branches:
```
  let x = A { i: 1 };

  if condition {
      let y = x;
  } else {
      let z = x;
  }
```
```
  Drop(x): Dead
```

### DIFF
--- a/gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
+++ b/gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
@@ -20,20 +20,277 @@
 #include "rust-bir.h"
 #include "rust-hir-map.h"
 
-#include <unordered_set>
-
 namespace Rust {
 namespace BIR {
-
 namespace {
 
-struct BasicBlockIdHash
+struct BlockInitializationState
 {
-  size_t operator() (BasicBlockId id) const
-  {
-    return std::hash<uint32_t> () (id.value);
-  }
+  explicit BlockInitializationState (size_t place_count)
+    : maybe_initialized (place_count, false),
+      maybe_uninitialized (place_count, false), reachable (false)
+  {}
+
+  std::vector<bool> maybe_initialized;
+  std::vector<bool> maybe_uninitialized;
+
+  // Whether this block is reached or not.
+  bool reachable;
 };
+
+static bool
+is_straight_line (const Function &function)
+{
+  std::set<BasicBlockId> visited;
+  BasicBlockId current = ENTRY_BASIC_BLOCK;
+
+  while (current != INVALID_BB)
+    {
+      // Revisiting a block means that the CFG contains a cycle.
+      if (!visited.insert (current).second)
+	return false;
+
+      const BasicBlock &block = function.basic_blocks[current];
+
+      if (block.successors.empty ())
+	return true;
+
+      if (block.successors.size () != 1)
+	return false;
+
+      current = block.successors.front ();
+    }
+
+  return true;
+}
+
+static void
+set_initialized (BlockInitializationState &state, PlaceId place)
+{
+  state.maybe_initialized[place.value] = true;
+  state.maybe_uninitialized[place.value] = false;
+}
+
+static void
+set_uninitialized (BlockInitializationState &state, PlaceId place)
+{
+  state.maybe_initialized[place.value] = false;
+  state.maybe_uninitialized[place.value] = true;
+}
+
+static bool
+merge_state (BlockInitializationState &into,
+	     const BlockInitializationState &from)
+{
+  bool changed = false;
+
+  if (!into.reachable)
+    {
+      into = from;
+      return !changed;
+    }
+
+  for (size_t i = 0; i < into.maybe_initialized.size (); i++)
+    {
+      bool maybe_initialized
+	= into.maybe_initialized[i] || from.maybe_initialized[i];
+
+      bool maybe_uninitialized
+	= into.maybe_uninitialized[i] || from.maybe_uninitialized[i];
+
+      changed |= maybe_initialized != into.maybe_initialized[i];
+      changed |= maybe_uninitialized != into.maybe_uninitialized[i];
+
+      into.maybe_initialized[i] = maybe_initialized;
+      into.maybe_uninitialized[i] = maybe_uninitialized;
+    }
+
+  return changed;
+}
+
+static void
+update_state_for_statement (Function &function, Statement &statement,
+			    BlockInitializationState &state)
+{
+  PlaceId place = statement.get_place ();
+
+  switch (statement.get_kind ())
+    {
+    case Statement::Kind::STORAGE_LIVE:
+      set_uninitialized (state, place);
+      break;
+
+    case Statement::Kind::ASSIGNMENT:
+      {
+	PlaceId lhs = place;
+	AbstractExpr &expr = statement.get_expr ();
+
+	if (expr.get_kind () == ExprKind::ASSIGNMENT)
+	  {
+	    PlaceId rhs = static_cast<Assignment &> (expr).get_rhs ();
+	    const Place &rhs_place = function.place_db[rhs];
+
+	    if (rhs_place.kind == Place::VARIABLE
+		&& rhs_place.should_be_moved ())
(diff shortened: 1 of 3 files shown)


gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
```diff
@@ -0,0 +1,339 @@
...
+
+static void
+set_uninitialized (BlockInitializationState &state, PlaceId place)
+{
+  state.maybe_initialized[place.value] = false;
+  state.maybe_uninitialized[place.value] = true;
+}
+
+static bool
+merge_state (BlockInitializationState &into,
```
Lishin1215, line 102: This function combines states from incoming blocks.
Take conditional move as example, 

              BB0
           initialize x
              /    \
             v      v
         BB1      BB2
        move x   no change
             \      /
              v    v
               BB3
              Drop(x)

BB1 -> BB3
  (BB1): maybe_initialized(x) = false 
  (BB1): maybe_unitialized(x) =true
     
  (BB3): maybe_initialized(x)   = true || (BB1): maybe_initialized(x) -> true
  (BB3): maybe_uninitialized(x) = false ||  (BB1): maybe_unitialized(x) -> true

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
```diff
@@ -0,0 +1,339 @@
...
+}
+
+void
+DropAnalysis::analyze (Function &function)
+{
+  std::vector<BlockInitializationState> entry_states
+    = compute_entry_states (function);
+
+  // Keep the existing backend handling for straight-line CFGs.
+  const bool record_straight_line_backend_drops = is_straight_line (function);
```
Lishin1215, line 343: This is a specific bridge from BIR Drop analysis to the existing backend.
It stores the results using HirId and updates definitely_dead.
It is only used for straight-line CFGs because the backend does not support Conditional Drops yet.

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
```diff
@@ -0,0 +1,339 @@
...
+
+#include "rust-bir-drop-analysis.h"
+#include "rust-bir.h"
+#include "rust-hir-map.h"
+
+namespace Rust {
+namespace BIR {
+namespace {
+
+struct BlockInitializationState
```
Lishin1215, line 27: In order to merge states from different blocks, I make the state associate to block
And I added a ‘reachable’ parameter to track if a block is visited. 

The previous straight-line analysis used one state while processing statements.
But in conditional cases, there are three different states that need to be represented (Static / Dead / Conditional).
The original representation could only represent two states.
So I added `maybe_initialized`, `maybe_uninitialized` to track these three states.

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
```diff
@@ -0,0 +1,339 @@
...
+
+      into.maybe_initialized[i] = maybe_initialized;
+      into.maybe_uninitialized[i] = maybe_uninitialized;
+    }
+
+  return changed;
+}
+
+static void
+update_state_for_statement (Function &function, Statement &statement,
```
Lishin1215, line 132: When implementing the conditional cases, I “update initialize/uninitialized based on statements” two times in different places.
1. The first time is because blocks need to merge states, so need to find the state and pass it to the next blocks.
2. The second time is to use the result to decide whether a Drop is Conditional, Static, or Dead.

Therefore, I extracted this logic from the old function into a new helper.

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
```diff
@@ -0,0 +1,339 @@
...
+    set_initialized (entry_state, argument);
+
+  std::vector<BasicBlockId> worklist;
+  std::vector<bool> queued (block_count, false);
+
+  worklist.push_back (ENTRY_BASIC_BLOCK);
+  queued[ENTRY_BASIC_BLOCK.value] = true;
+
+  // Propagate block states until the last block.
+  while (!worklist.empty ())
```
Lishin1215, line 222: To keep track of blocks that still need to be processed, I use a while loop, worklist and queued to achieve this.
So in the end

- Block that has not been reached yet 
- Block that already reached but its state changes after merging with another state.	  
 
will be added into the worklist.

(queued is to record whether a block is already in the worklist. )

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
```diff
@@ -0,0 +1,339 @@
...
+// its Drop statements.
+static void
+annotate_drop_statements (
+  Function &function, const std::vector<BlockInitializationState> &entry_states,
+  bool record_straight_line_backend_drops, std::set<HirId> &dead_drop_hir_ids,
+  std::set<HirId> &non_dead_drop_hir_ids)
+{
+  const size_t block_count = function.basic_blocks.size ();
+
+  for (size_t i = 0; i < block_count; i++)
```
Lishin1215, line 282: I loop here again because I need to wait until the state stops changing.
And I 
1. save state 
    * Because if the state might still change, the result may need to be overwritten many times. This will also increase time usage.
2. classify/set Drop
    * Because the classification depends on the state. If the state keeps  changing, the classification might be wrong.
3. handle Straight-line backend
    * Because the DropStyle is already known, so the result can be passed to the backend.
4. update state for statement
    * Because Drop classification needs the stable state, so need to update again here.

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
P-E-P: You should probably add this comment directly within the preamble of this function.

P-E-P (approved):
I'm not entirely familiar with BIR so it took a while to review. Another pair of eyes wouldn't hurt as all those `!maybe_initialized` & al are confusing sometimes but this looks good.

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
Lishin1215: I've added the comment directly to the function's preamble as suggested. Thanks!

P-E-P (approved):

PR #4783, opened 2026-08-16
gccrs: Improve `rust_debug` formatting

powerboat9:
This allows `rust_debug` to use (an ad-hoc implementation of) gcc diagnostic formatting. To keep this patch as small as possible, proper fixes for `-Wformat-diag` are mostly left for future patches.

### DIFF
--- a/gcc/rust/rust-session-manager.cc
+++ b/gcc/rust/rust-session-manager.cc
@@ -70,6 +70,9 @@
 #include "rust-target.h"
 #include "rust-system.h"
 
+// TODO: fix warnings and remove
+#pragma GCC diagnostic ignored "-Wformat-diag"
+
 extern bool saw_errors (void);
 
 extern Linemap *rust_get_linemap ();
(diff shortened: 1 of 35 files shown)


powerboat9:
@philberty is this good to merge?

gcc/rust/rust-session-manager.cc
```diff
@@ -70,6 +70,9 @@
 #include "rust-target.h"
 #include "rust-system.h"
 
+// TODO: remove
+#pragma GCC diagnostic ignored "-Wformat-diag"
```
philberty, line 74: whats going on with all the #pragma GCC diagnostic ignored "-Wformat-diag"

powerboat9: With the change from printf to GCC diagnostic formatting, GCC starts warning about diagnostic quality (excess punctuation, unquoted identifiers, etc).

philberty: i'm not sure what to review here whats the improvement maybe i cant see it with all the diff

powerboat9: It switches from printf style formatting (no `%qs`, `%<`, `%>`, `%m`, `%r`, `%R`, etc) to our implementation of GCC diagnostic style formatting (backwards compatible and with support for `%qs`, `%<`, `%>`, and `%m`, but not everything supported by GCC's main diagnostics system) for debug printing. #4784 is a follow up that switches us over to using GCC diagnostic formatting proper.

PR #4785, opened 2026-08-16
gccrs: Prevent warning on "unreachable" items

powerboat9:
Items declared in a block expression can't be unreachable.


philberty (approved):
LGTM

Issue #4766, opened 2026-08-04
Register `panic_info`, `panic_location`, `panic` and `panic_bounds_check` lang items


CohenArthur:
And eventually use them when expanding the various panicking builtin macros

Issue #4764, opened 2026-08-04
Register `new_unchecked` pinning lang item


CohenArthur:
...and see where the compiler needs to call it

PR #4792, opened 2026-08-17
gccrs: Add missing lang item declarations for core 1.49

### LINKED ISSUE #4763: Register `align_offset` lang item
We also need to think about where to use it as this is clearly required by the compiler

Polygonalr:
Closes #4763, #4764, #4766


nsvke:
if these are stubs, you can add them to `warn_if_stub_lang_item` at `hir/rust-ast-lower-base.cc`

CohenArthur:
Ideally we could do something like warn them for stubs when we see them in the lang item collector

CohenArthur:
(but in a later PR)

nsvke:
> Ideally we could do something like warn them for stubs when we see them in the lang item collector
> 

If I'm not misunderstanding you, this is already working that way 😅.

P-E-P (approved):

PR #4797, opened 2026-08-17
Combined unused lints

lucasly-ba:
As requested by @philberty on #4639, a single PR combining  #4622, #4623, #4624, #4626, #4629, #4639 and #4642

### DIFF
--- a/gcc/rust/checks/lints/unused/rust-unused-checker.cc
+++ b/gcc/rust/checks/lints/unused/rust-unused-checker.cc
@@ -26,12 +26,15 @@
 #include "rust-keyword-values.h"
 #include "rust-attribute-values.h"
 #include "rust-rib.h"
+#include "rust-tyty.h"
 
 namespace Rust {
 namespace Analysis {
 UnusedChecker::UnusedChecker ()
   : nr_context (Resolver2_0::FinalizedNameResolutionContext::get ()),
-    mappings (Analysis::Mappings::get ()), unused_context (UnusedContext ())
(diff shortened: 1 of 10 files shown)


philberty:
the commit format needs fixed up

lucasly-ba:
@philberty should be good

gcc/rust/checks/lints/unused/rust-unused-checker.cc
```diff
@@ -324,6 +335,41 @@ UnusedChecker::visit (HIR::MatchExpr &expr)
...
+	  rust_warning_at (arm.get_pattern ()->get_locus (), OPT_Wunused,
+			   "unreachable pattern");
+	  continue;
+	}
+
+      if (!arm.has_match_arm_guard ()
+	  && arm.get_pattern ()->get_pattern_type ()
+	       == HIR::Pattern::PatternType::WILDCARD)
+	seen_irrefutable = true;
+    }
```
CohenArthur, line 356: so this just checks if we've seen a wildcard pattern and calls it irrefutable, which seems wrong. we should check that we've seen an actually irrefutable pattern (which we have the infrastructure for, grep for `refutable` or `irrefutable` to find the impl). for example something like this is not a wildcard pattern but should still trigger the `unreachable pattern` warning

```rust
    let some_i32 = 15;

    match some_i32 {
        14 => {},
        15 => {},
        x => {},
        17 => {},
    };
```

CohenArthur (commented):
thanks for the work, that's really good!

PR #4798, opened 2026-08-18
gccrs: Generate backend drop flags for conditional moves

Lishin1215:
This patch connects the conditional BIR Drop analysis to the existing backend cleanup.

For now, in a conditional move case like:
```
fn f(cond: bool) {
    let x = Droppable { value: 1 };


    if cond {
        let y = x;
    }
}
```

The BIR Drop analysis classifies the cleanup of x as:

    Drop(x): Conditional

The backend then creates a Drop flag for x.
Conceptually, the generated backend code behaves like:
```
try {
    bool flag = false;
    let x = Droppable { value: 1 };
    flag = true;


    if (cond) {
        let y = x;
        flag = false;
    }
}
finally {
    if (flag) {
        flag = false;
        Drop(x);
    }
}
```

- When x is initialized, the flag is set to true.
- When x is moved, the flag is cleared to false, so Drop(x) is skipped during cleanup.
- When x is not moved, the flag remains true, so Drop(x) is called.

### DIFF
--- a/gcc/rust/checks/errors/borrowck/rust-bir-builder-internal.h
+++ b/gcc/rust/checks/errors/borrowck/rust-bir-builder-internal.h
@@ -280,17 +280,19 @@ protected: // Helpers to add BIR statements
   }
 
   void push_tmp_assignment (AbstractExpr *rhs, TyTy::BaseType *tyty,
-			    location_t location)
+			    location_t location,
+			    tl::optional<HirId> move_site = tl::nullopt)
   {
     PlaceId tmp = ctx.place_db.add_temporary (tyty);
     push_storage_live (tmp);
-    push_assignment (tmp, rhs, location);
+    push_assignment (tmp, rhs, location, move_site);
   }
 
-  void push_tmp_assignment (PlaceId rhs, location_t location)
+  void push_tmp_assignment (PlaceId rhs, location_t location,
+			    tl::optional<HirId> move_site = tl::nullopt)
   {
-    push_tmp_assignment (new Assignment (rhs), ctx.place_db[rhs].tyty,
-			 location);
+    push_tmp_assignment (new Assignment (rhs), ctx.place_db[rhs].tyty, location,
+			 move_site);
   }
 
   void push_switch (PlaceId switch_val, location_t location,
@@ -353,7 +355,8 @@ protected: // Helpers to add BIR statements
     return translated;
   }
 
-  PlaceId move_place (PlaceId arg, location_t location)
+  PlaceId move_place (PlaceId arg, location_t location,
+		      tl::optional<HirId> move_site = tl::nullopt)
   {
     auto &place = ctx.place_db[arg];
 
@@ -366,7 +369,7 @@ protected: // Helpers to add BIR statements
     if (place.is_rvalue ())
       return arg;
 
-    push_tmp_assignment (arg, location);
+    push_tmp_assignment (arg, location, move_site);
     return translated;
   }
 
@@ -379,12 +382,14 @@ protected: // Helpers to add BIR statements
   }
 
   template <typename T>
-  void move_all (T &args, std::vector<location_t> locations)
+  void move_all (T &args, std::vector<location_t> locations,
+		 tl::optional<HirId> move_site = tl::nullopt)
   {
     rust_assert (args.size () == locations.size ());
     std::transform (args.begin (), args.end (), locations.begin (),
-		    args.begin (), [this] (PlaceId arg, location_t location) {
-		      return move_place (arg, location);
+		    args.begin (),
+		    [this, move_site] (PlaceId arg, location_t location) {
+		      return move_place (arg, location, move_site);
 		    });
   }
 
--- a/gcc/rust/checks/errors/borrowck/rust-bir-builder-internal.h
+++ b/gcc/rust/checks/errors/borrowck/rust-bir-builder-internal.h
@@ -265,16 +265,18 @@ protected:
   }
 
 protected: // Helpers to add BIR statements
-  void push_assignment (PlaceId lhs, AbstractExpr *rhs, location_t location)
+  void push_assignment (PlaceId lhs, AbstractExpr *rhs, location_t location,
+			HirId move_site = UNKNOWN_HIRID)
   {
     ctx.get_current_bb ().statements.push_back (
-      Statement::make_assignment (lhs, rhs, location));
+      Statement::make_assignment (lhs, rhs, location, move_site));
     translated = lhs;
   }
 
-  void push_assignment (PlaceId lhs, PlaceId rhs, location_t location)
+  void push_assignment (PlaceId lhs, PlaceId rhs, location_t location,
+			HirId move_site = UNKNOWN_HIRID)
   {
-    push_assignment (lhs, new Assignment (rhs), location);
+    push_assignment (lhs, new Assignment (rhs), location, move_site);
   }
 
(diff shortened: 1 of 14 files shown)


gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.h
```diff
@@ -39,9 +39,13 @@ class DropAnalysis
   void analyze (Function &function);
 
   bool is_definitely_dead (HirId id) const;
+  bool needs_drop_flag (HirId id) const;
+  bool lookup_move_source (HirId move_site, HirId *source) const;
 
 private:
   std::set<HirId> definitely_dead;
+  std::set<HirId> conditionally_dropped;
```
Lishin1215: I added `conditionally_dropped` to save those that are classified as `Conditional`.
I need this information to determine whether a variable requires a `Drop flag`.
It is used when generating the backend code.

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.h
```diff
@@ -39,9 +39,13 @@ class DropAnalysis
...
+  std::map<HirId, HirId> move_sources;
```
Lishin1215: I added `move_sources` to map the move expression HirId to the source local HirId.


Since the move expression `x` and the source local `x` have different HirIds.
And also I need the source local `x`'s HirId to find the Drop flag, 
so this map lets me find the source HirId from the move expression HirId.

For example:
```rust
...
  let x = Droppable();  // source local HirId = 10
  x.flag = true;  // associate with source local HirId 10
...
  let y = x;  // move expression HirId = 30
```

Here I only have HirId 30, but the flag is associated with HirId 10.
So I need:

```rust
  move_sources[30] = 10
```

Then the backend can use HirId 10 to find the correct flag and clear it:
```rust
x.flag = false;
```

gcc/rust/checks/errors/borrowck/rust-bir-builder-internal.h
```diff
@@ -265,16 +265,18 @@ class AbstractBuilder
   }
 
 protected: // Helpers to add BIR statements
-  void push_assignment (PlaceId lhs, AbstractExpr *rhs, location_t location)
+  void push_assignment (PlaceId lhs, AbstractExpr *rhs, location_t location,
+			HirId move_site = UNKNOWN_HIRID)
```
P-E-P: Could we benefit from an optional ?

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
```diff
@@ -333,26 +344,45 @@ DropAnalysis::is_definitely_dead (HirId id) const
...
 void
 DropAnalysis::analyze (Function &function)
 {
   std::vector<BlockInitializationState> entry_states
     = compute_entry_states (function);
 
-  // Keep the existing backend handling for straight-line CFGs.
-  const bool record_straight_line_backend_drops = is_straight_line (function);
-
   std::set<HirId> dead_drop_hir_ids;
```
P-E-P: I've got a feeling it would be better to pack those into a class / structure ? This way we could add annotate_drop_statement and record_drop_label as member functions and avoid carrying this state around everywhere as 4 different arguments. It would be either `this` if we make them member function or a single argument if we pass that struct by ref.

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.h
P-E-P: What if multiple id are moved within the same location ? I guess any product type that move by construction would have multiple ids associated to the same hir id ?

Lishin1215: Hi, this is a good point.
The current patch only records move sources for direct whole-local moves such as let y = x. 
For these moves, push_assignment() is called with a move_site.
Product type will go through a different path and do not carry a move_site, so they are not recorded in move_sources. Therefore, multiple source HirIds will not currently be associated with the same key.

I am thinking of keeping the current map and adding a comment such as:
```C++
// This may need to be extended to std::map<HirId, std::set<HirId>>
// to support product moves in the future.
```
Or do you think it would still be better to use a set now?
Please let me know if I misunderstood anything. Thanks!

P-E-P: Either the comment or the change to map<set> would be good!

Also, is there any reason why you're using map/set and not unordered_set/unordered_map ?

Lishin1215: You're right. I don't think there is any ordering requirement here.
I'll change them to std::unordered_map and std::unordered_set.
Thanks!

Lishin1215: I've added the comment and changed `std::map` and `std::set` to `std::unordered_map` and `std::unordered_set`. Thanks!

gcc/rust/checks/errors/borrowck/rust-bir-builder-internal.h
Lishin1215: Yes, I've changed `move_site` to optional now. Thanks!

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.cc
Lishin1215: I've packed the four pieces of state into a `DropAnalysisResults` struct and now pass it by reference as a single argument!

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.h
Lishin1215: I've now made the case you mentioned explicitly unsupported in a new commit ([gccrs: Reject multiple move sources in product-type expressions](https://github.com/Rust-GCC/gccrs/pull/4798/changes/8ff0583686e9196b1cf96d61cb019d5ebf569c1a))
If multiple moved IDs are associated with the same move site, gccrs emits a diagnostic using `rust_sorry_at`. 
I've also added test! Thanks.

P-E-P: Great! Thank you!

gcc/rust/checks/errors/borrowck/rust-bir-builder-internal.h
P-E-P: Thank you, I feel it makes it clearer that no value can be provided instead of relying uniquely on function overloading.

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.h
```diff
@@ -21,6 +21,9 @@
 
 #include "rust-bir.h"
 
+#include <unordered_map>
+#include <unordered_set>
```
P-E-P: You cannot rely on system import, you instead need to rely on `rust-system.h`

P-E-P (changes requested):
Great work! One last nitpick and I'll merge this!

gcc/rust/checks/errors/borrowck/rust-bir-drop-analysis.h
Lishin1215: Thanks for pointing this out! 
I've updated it to rely on `rust-system.h`.

P-E-P (approved):

PR #4799, opened 2026-08-18
gccrs: Report an error on an empty path expression

### LINKED ISSUE #4790: Segfault in TypeCheckExpr::visit (PathInExpression&) on `let x = ::;`
### Summary

A `let` initialiser consisting of a bare `::` segfaults during type checking.
The crash is in the type checker, but the cause is in the parser: the path
parser returns an error node without emitting a diagnostic, and that error
node is indistinguishable from a real empty path further down.

Found while checking whether #1443 still reproduced. #1443's case,
`fn main() { ::; }`, is now correctly rejected. In a `let` initialiser the
same token sequence takes a different route and reaches type checking.

### Reproducer

```rust
pub fn main() {
    let x = ::;
}
```

```
crab1 mvce.rs -frust-incomplete-and-experimental-compiler-do-not-use \
      -frust-crate-attr=no_core -frust-crate-attr=feature(no_core)
```

- [x] Nightly

### Godbolt link

N/A, needs the experimental and no_core flags.

### Actual behavior

```
0x2d0753b internal_error(char const*, ...)
	../../gccrs/gcc/diagnostic-global-context.cc:787
0x16bd933 crash_signal
	../../gccrs/gcc/toplev.cc:325
0x7dc953e45caf ???
	./signal/../sysdeps/unix/sysv/linux/x86_64/libc_sigaction.c:0
0xe97a64 Rust::TyTy::BaseType::get_kind() const
	../../gccrs/gcc/rust/typecheck/rust-tyty.cc:264
0xf754bb Rust::Resolver::TypeCheckExpr::visit(Rust::HIR::PathInExpression&)
	../../gccrs/gcc/rust/typecheck/rust-hir-type-check-path.cc:228
0xef622d Rust::Resolver::TypeCheckExpr::Resolve(Rust::HIR::Expr&)
	../../gccrs/gcc/rust/typecheck/rust-hir-type-check-expr.cc:53
0xf0355e Rust::Resolver::TypeCheckStmt::visit(Rust::HIR::LetStmt&)
	../../gccrs/gcc/rust/typecheck/rust-hir-type-check-stmt.cc:113
0xf038e6 Rust::Resolver::TypeCheckStmt::Resolve(Rust::HIR::Stmt&)
	../../gccrs/gcc/rust/typecheck/rust-hir-type-check-stmt.cc:34
0xef7bf9 Rust::Resolver::TypeCheckExpr::visit(Rust::HIR::BlockExpr&)
	../../gccrs/gcc/rust/typecheck/rust-hir-type-check-expr.cc:717
0xef622d Rust::Resolver::TypeCheckExpr::Resolve(Rust::HIR::Expr&)
	../../gccrs/gcc/rust/typecheck/rust-hir-type-check-expr.cc:53
0xedff66 Rust::Resolver::TypeCheckItem::visit(Rust::HIR::Function&)
	../../gccrs/gcc/rust/typecheck/rust-hir-type-check-item.cc:865
0xed8e78 Rust::Resolver::TypeCheckItem::Resolve(Rust::HIR::Item&)
	../../gccrs/gcc/rust/typecheck/rust-hir-type-check-item.cc:102
0xed8e78 Rust::Resolver::TypeCheckItem::Resolve(Rust::HIR::Item&)
	../../gccrs/gcc/rust/typecheck/rust-hir-type-check-item.cc:88
0xe8cb9b Rust::Resolver::TypeResolution::Resolve(Rust::HIR::Crate&)
	../../gccrs/gcc/rust/typecheck/rust-hir-type-check.cc:72
0xd435b2 Rust::Session::compile_crate(char const*)
	../../gccrs/gcc/rust/rust-session-manager.cc:808
```

This is not #1443. `::;` in statement position already errors correctly and is
covered by compile/empty_path.rs. The `let` initialiser takes a different
parse route.

`parse_path_in_expression` (gcc/rust/parse/rust-parse-impl-path.hxx:388-395)
returns `PathInExpression::create_error ()` without emitting a diagnostic when
the initial segment fails; the in-source comment reads "don't necessarily
throw error but yeah". The sibling branch at :406-412, for subsequent
segments, does call `add_error`.

`create_error ()` builds a zero-segment path and `is_error ()` is
`!has_segments ()`, so nothing downstream distinguishes the sentinel from a
real path. The AST dump shows the initialiser reduced to nothing:

```
pub fn main() {
    let x =;
}
```

Name resolution is silent through every pass. `resolve_root_path`
(rust-hir-type-check-path.cc:243-387) then skips its segment loop and falls
through to `return root_tyty` at :386, returning the nullptr it was
initialised to at :247, which :228 dereferences.

Of that function's eight returns, five build a `new TyTy::ErrorType (...)` and
two check for non-null first. The fall-through at :386 is the only unguarded
exit.

### Expected behavior

This should be a parse error. rustc 1.95.0:

```
error: expected identifier, found `;`
 --> mvce.rs:2:15
  |
2 |     let x = ::;
  |               ^ expected identifier
```

### GCC Version

Built from Rust-GCC/gccrs at 08ec968b3da.

u7k4rs6:
Fixes Rust-GCC/gccrs#4790

The path parser previously returned an error node silently when the initial path segment failed to parse. The resulting zero-segment path can reach rust-hir-type-check-path.cc:228 through two routes: let x = ::; reaches TypeCheckStmt::visit (LetStmt&), while let :: = 1; reaches TypeCheckPattern::visit (PathInExpression&). Both eventually dereference the null result from resolve_root_path.

The new diagnostic is therefore gated on has_opening_scope_resolution. An unconditional guard was incorrect because it also catches bare $, which never ICEs and already receives more specific diagnostics downstream. The unconditional version broke issue-2225.rs and both issue-4140-*.rs by adding the less-specific expected identifier diagnostic.

The parse_stmt_or_expr guard from #4558 was also removed because it reports the same error at the same locus. Since add_error does not deduplicate diagnostics, keeping both guards produces duplicate errors and turns empty_path.rs red on excess errors.

Validation passes increased from 11,449 to 11,453 with zero failures. Comparing against the same-commit baseline showed exactly the four new PASS lines corresponding to the added tests.

### DIFF
--- a/gcc/testsuite/rust/compile/empty_path2.rs
+++ b/gcc/testsuite/rust/compile/empty_path2.rs
@@ -1,0 +1,9 @@
+#![feature(no_core)]
+#![no_core]
+
+// A path expression with no segments in a let initialiser segfaulted
+// during type checking, see Rust-GCC/gccrs#4790.
+fn main() {
+    let x = ::;
+    // { dg-error "expected identifier" "" { target *-*-* } .-1 }
+}
--- a/gcc/testsuite/rust/compile/empty_path2.rs
+++ b/gcc/testsuite/rust/compile/empty_path2.rs
@@ -1,0 +1,9 @@
+#![feature(no_core)]
+#![no_core]
+
+// A path expression with no segments in a let initialiser segfaulted
+// during type checking, see #4790.
+fn main() {
+    let x = ::;
+    // { dg-error "expected identifier" "" { target *-*-* } .-1 }
+}
--- a/gcc/testsuite/rust/compile/empty_path3.rs
+++ b/gcc/testsuite/rust/compile/empty_path3.rs
@@ -1,0 +1,12 @@
+#![feature(no_core)]
+#![no_core]
+
+// An empty path `::` in pattern position. The parser returns a path with no
+// segments without reporting anything, so type checking dereferences a null
+// root type and the compiler crashes instead of diagnosing. Same defect as
+// empty_path2.rs, which covers the let initialiser instead of the pattern.
+// See #4790.
+fn main() {
+    let :: = 1;
+    // { dg-error "expected identifier" "" { target *-*-* } .-1 }
+}
--- a/gcc/testsuite/rust/compile/empty_path3.rs
+++ b/gcc/testsuite/rust/compile/empty_path3.rs
@@ -1,0 +1,12 @@
+#![feature(no_core)]
+#![no_core]
+
+// An empty path `::` in pattern position. The parser returned a path with no
+// segments without reporting anything, so type checking dereferenced a null
+// root type and the compiler crashed instead of diagnosing. Same defect as
+// empty_path2.rs, which covers the let initialiser instead of the pattern.
+// See Rust-GCC/gccrs#4790.
+fn main() {
+    let :: = 1;
+    // { dg-error "expected identifier" "" { target *-*-* } .-1 }
+}
--- a/gcc/rust/parse/rust-parse-impl-path.hxx
+++ b/gcc/rust/parse/rust-parse-impl-path.hxx
@@ -389,8 +389,11 @@ Parser<ManagedTokenSource>::parse_path_in_expression ()
   AST::PathExprSegment initial_segment = parse_path_expr_segment ();
   if (initial_segment.is_error ())
     {
-      // skip after somewhere?
-      // don't necessarily throw error but yeah
+      if (has_opening_scope_resolution)
+	{
+	  Error error (locus, "expected identifier");
+	  add_error (std::move (error));
+	}
       return AST::PathInExpression::create_error ();
     }
   segments.push_back (std::move (initial_segment));
--- a/gcc/rust/parse/rust-parse-impl.hxx
+++ b/gcc/rust/parse/rust-parse-impl.hxx
@@ -7294,15 +7294,6 @@ Parser<ManagedTokenSource>::parse_stmt_or_expr ()
     case DOLLAR_SIGN:
       {
 	AST::PathInExpression path = parse_path_in_expression ();
-	if (path.is_error ())
-	  {
-	    Error error (t->get_locus (), "expected identifier");
-	    add_error (std::move (error));
-	    skip_after_semicolon ();
-	    return tl::unexpected<Parse::Error::Node> (
-	      Parse::Error::Node::CHILD_ERROR);
-	  }
-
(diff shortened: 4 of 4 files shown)


gcc/testsuite/rust/compile/empty_path3.rs
```diff
@@ -0,0 +1,12 @@
+#![feature(no_core)]
+#![no_core]
+
+// An empty path `::` in pattern position. The parser returns a path with no
+// segments without reporting anything, so type checking dereferences a null
+// root type and the compiler crashes instead of diagnosing. Same defect as
+// empty_path2.rs, which covers the let initialiser instead of the pattern.
+// See #4790.
```
powerboat9: Could you clarify that this is GCCRS github issue #4790? Upstream sometimes confuses these with bugzilla bug numbers.

gcc/testsuite/rust/compile/empty_path2.rs
```diff
@@ -0,0 +1,9 @@
+#![feature(no_core)]
+#![no_core]
+
+// A path expression with no segments in a let initialiser segfaulted
+// during type checking, see #4790.
```
powerboat9: Same here

gcc/testsuite/rust/compile/empty_path3.rs
```diff
@@ -0,0 +1,12 @@
+#![feature(no_core)]
+#![no_core]
+
+// An empty path `::` in pattern position. The parser returns a path with no
+// segments without reporting anything, so type checking dereferences a null
+// root type and the compiler crashes instead of diagnosing. Same defect as
+// empty_path2.rs, which covers the let initialiser instead of the pattern.
```
powerboat9: and could you make this past tense? Since you've fixed the bug.

powerboat9:
otherwise, LGTM

gcc/testsuite/rust/compile/empty_path3.rs
u7k4rs6: Yes, gccrs GitHub. Changed to Rust-GCC/gccrs#4790 so it can't be read as a bugzilla number upstream.

gcc/testsuite/rust/compile/empty_path2.rs
u7k4rs6: Same change here.

gcc/testsuite/rust/compile/empty_path3.rs
u7k4rs6: Done, past tense now.

PR #4803, opened 2026-08-20
gccrs: Resolve lang item types

powerboat9:
Lang items might not already be resolved when an expression depending on them is resolved.


CohenArthur (approved):
LGTM, can you add a testcase for it? also leaving it to @philberty for review

powerboat9:
Added test case

philberty (approved):
LGTM thats the correct fix

philberty:
good catch with that testcase it ices making an issue for myself for it

PR #4809, opened 2026-08-23
gccrs: Support some LLVM builtin functions

### LINKED ISSUE #4818: Compile `UNADJUSTED` ABI extern functions
Continuation of #4791, we do not support compilation of LLVM built-ins yet, e.g:

```rust
#[allow(improper_ctypes)]
extern "unadjusted" {
    #[link_name = "llvm.x86.rdrand.16"]
    fn x86_rdrand16_step() -> (u16, i32);
    #[link_name = "llvm.x86.rdrand.32"]
    fn x86_rdrand32_step() -> (u32, i32);
    #[link_name = "llvm.x86.rdseed.16"]
    fn x86_rdseed16_step() -> (u16, i32);
    #[link_name = "llvm.x86.rdseed.32"]
    fn x86_rdseed32_step() -> (u32, i32);
}
```

I do not have permissions to assign myself the issue, but work is already in progress in #4809.

Polygonalr:
Closes #4818, ~~currently it only supports the rdrand and rdseed functions, I'll need to write more function adapters to support other LLVM built-ins.~~

In theory it should support all the LLVM built-ins used in `stdarch`, though `llvm.x86.vcvtph2ps` and `llvm.x86.vcvtps2ph` requires the SIMD types to be compiled as well (which will also be compiled as part of `core_arch` within `stdarch`).

### DIFF
--- a/gcc/testsuite/rust/compile/llvm_builtins.rs
+++ b/gcc/testsuite/rust/compile/llvm_builtins.rs
@@ -1,0 +1,50 @@
+// { dg-options "-mrdrnd" }
+#![feature(no_core, abi_unadjusted)]
+#![no_core]
+
+#[allow(improper_ctypes)]
+extern "unadjusted" {
+    #[link_name = "llvm.x86.rdrand.16"]
+    fn x86_rdrand16_step() -> (u16, i32);
+    #[link_name = "llvm.x86.rdrand.32"]
+    fn x86_rdrand32_step() -> (u32, i32);
+    #[link_name = "llvm.x86.rdrand.64"]
+    fn x86_rdrand64_step() -> (u64, i32);
+    #[link_name = "llvm.x86.rdseed.16"]
+    fn x86_rdseed16_step() -> (u16, i32);
+    #[link_name = "llvm.x86.rdseed.32"]
+    fn x86_rdseed32_step() -> (u32, i32);
+    #[link_name = "llvm.x86.rdseed.64"]
+    fn x86_rdseed64_step() -> (u64, i32);
+
+    #[link_name = "llvm.x86.addcarry.32"]
+    fn llvm_addcarry_u32(a: u8, b: u32, c: u32) -> (u8, u32);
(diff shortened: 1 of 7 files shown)


powerboat9:
Could you replace "frontend" in your first commit with "front end"? Also, you should probably replace the mention of `(BuiltinsContext)` with a mention of `(BuiltinsContext::get)` and a `(BuiltinsContext::BuiltinsContext): Likewise.`.

Polygonalr:
Tests are not final yet.

* It seems like tests on the 32-bit arches are failing due to not having 64-bit built-ins such as `__builtin_ia32_rdrand64_step` and `__builtin_ia32_rdseed_di_step`, wondering if I should remove them or find a way to exclude these functions from the 32-bit arches.
* I'll also need to include some `dg-final` directives in my tests similar to what I did in #4650.

powerboat9:
I think the test should probably be restricted to run only on x86_64

philberty (approved):
LGTM

gcc/testsuite/rust/compile/llvm_builtins.rs
```diff
@@ -0,0 +1,62 @@
...
+    fn x86_rdseed32_step() -> (u32, i32);
+
+    #[link_name = "llvm.x86.addcarry.32"]
+    fn llvm_addcarry_u32(a: u8, b: u32, c: u32) -> (u8, u32);
+    #[link_name = "llvm.x86.addcarryx.u32"]
+    fn llvm_addcarryx_u32(a: u8, b: u32, c: u32, d: *mut u8) -> u8;
+    #[link_name = "llvm.x86.subborrow.32"]
+    fn llvm_subborrow_u32(a: u8, b: u32, c: u32) -> (u8, u32);
+
+    // TODO implement the SIMD types (i16x8, f32x4, f32x8) before
```
philberty, line 24: repr simd should be there they will be VECTOR_TYPES which are basically structs in a lot of  ways

PR #4837, opened 2026-09-03
attr: Allow `#[used]` attribute and its argument variant.

### LINKED ISSUE #4836: Handle `used` built-in attribute
https://rust-lang.github.io/rfcs/2386-used.html

Also support the `used_with_arg` feature which adds two extra variants, `#[used(compiler)]` and `#[used(linker)]`. These are used by the kernel.

CohenArthur:
Closes #4836, but I think it's missing some stuff in the backend to properly handle the attribute. We always mark `StaticItem`s with `TREE_USED`, not sure if that is enough?

This commit allows recognizing both forms of the `#[used]` attribute: The basic one, which forbids any argument, and the one gated behind `feature(used_with_arg)`, which allows specifying whether the compiler or linker should be responsible for marking the static item as used.

This does not do anything in the backend for actually marking the item as used even if the compiler decides it should not be used. This is a first step towards handling it as the kernel requires it.

gcc/rust/ChangeLog:

	* checks/errors/feature/rust-feature-defs.h (FEATURE_ACTIVE): Add `used_with_arg` feature.
	* checks/errors/feature/rust-feature-gate.cc (FeatureGate::check_used_attribute): New function. (FeatureGate::visit): Implement feature gating visits for static items.
	* checks/errors/feature/rust-feature-gate.h: Declare them.
	* checks/errors/rust-builtin-attribute-checker.cc (used): New handler. (check_valid_attribute_for_item): Handle `used` attribute for static items.
	* util/rust-attribute-values.h: Add new attribute value for `used`.
	* util/rust-attributes.cc: Likewise.

gcc/testsuite/ChangeLog:

	* rust/compile/used-attr1.rs: New test.
	* rust/compile/used-attr2.rs: New test.
	* rust/compile/used-attr3.rs: New test.
	
backend: Add proper handling for #[used] attribute.

We now support the full implementation of the #[used] attribute, which
can be used to conserve a static variable even if the compiler deems it
unused. This commit also implements #[used(linker)], which instructs the
linker to not garbage collect the static even if it seems unused.

This behavior is equivalent to the `used` and `retain` attributes in C.

gcc/rust/ChangeLog:

	* backend/rust-compile-item.cc (handle_used_attr): New function.
	(CompileItem::visit): Call it.

gcc/testsuite/ChangeLog:

	* rust/compile/used-attr4.rs: New test.


gcc/rust/checks/errors/feature/rust-feature-defs.h
```diff
@@ -521,6 +521,9 @@ FEATURE_ACTIVE ("destructuring_assignment", DESTRUCTURING_ASSIGNMENT, "1.49.0",
 		ISSUE_SOME (71126), EDITION_NONE)
 FEATURE_ACTIVE ("cfg_panic", CFG_PANIC, "1.49.0", ISSUE_SOME (77443),
 		EDITION_NONE)
+// FIXME: We need to enable this when the compatibility mode is >= 1.60 I guess?
+FEATURE_ACTIVE ("used_with_arg", USED_WITH_ARG, "1.60.0", ISSUE_SOME (93798),
+		EDITION_NONE)
```
powerboat9: This should go in `rust-feature-defs-rfl.h` for now

powerboat9:
> We always mark StaticItems with TREE_USED, not sure if that is enough?

It might be, could you add a comment around where that's happening which mentions the `#[used]` attribute, so that doesn't get removed by someone who isn't considering `#[used]`?

CohenArthur:
I ended up implementing all three forms of the attribute including the linker one, so we don't need to do anything particular. this last commit may break the CI on older platforms though

PR #4846, opened 2026-09-04
Recognize the `#[cfi_encoding]` attribute and warn that it does nothing 

CohenArthur:


philberty:
i dont think this is merged in yet: "https://gcc.gnu.org/pipermail/gcc-patches/2026-July/724091.html"

but i bet all this is will be to add another atribute onto the record type

CohenArthur:
@philberty yes exactly, I actually reached out to the patch author (@kees) to make sure I understood everything correctly, and you're right that we'll need to add another attribute to override this. For now Kees' patchset does not allow overriding the encoding but this is something we'll eventually work on together!

PR #4857, opened 2026-09-07
gccrs: use std::move to reduce copy ctor usage on type-bound-predicate

philberty:
gcc/rust/ChangeLog:

	* typecheck/rust-hir-type-check-expr.cc (TypeCheckExpr::visit): update
	* typecheck/rust-hir-type-check-item.cc (TypeCheckItem::resolve_impl_block_substitutions): likewise
	* typecheck/rust-hir-type-check-path.cc (TypeCheckExpr::visit): likewise
	* typecheck/rust-hir-type-check-type.cc (TypeCheckType::visit): likewise
	* typecheck/rust-hir-type-check.h: likewise
	* typecheck/rust-typecheck-context.cc (TypeCheckContext::insert_resolved_predicate): likewise
	* typecheck/rust-tyty-bounds.cc (TypeBoundPredicate::TypeBoundPredicate): likewise (TypeBoundPredicate::operator=): remove deadcode (TypeBoundsMappings::add_bound): use ref
	* typecheck/rust-tyty-call.cc (validate_call_argument_associated_impl_bounds): use ref
	* typecheck/rust-tyty.cc (BaseType::BaseType): likewise (BaseType::inherit_bound): likewise (ParamType::ParamType): likewiase
	* typecheck/rust-tyty.h: likewise

### DIFF
--- a/gcc/rust/typecheck/rust-hir-type-check-item.cc
+++ b/gcc/rust/typecheck/rust-hir-type-check-item.cc
@@ -1086,7 +1086,7 @@ TypeCheckItem::resolve_impl_block_substitutions (HIR::ImplBlock &impl_block,
 
   // inherit the bounds
   if (!specified_bound.is_error ())
-    self->inherit_bounds ({specified_bound});
+    self->inherit_bound (specified_bound);
 
   // check for any unconstrained type-params
   const TyTy::SubstitutionArgumentMappings trait_constraints
--- a/gcc/rust/typecheck/rust-typecheck-context.cc
+++ b/gcc/rust/typecheck/rust-typecheck-context.cc
@@ -574,13 +574,13 @@ TypeCheckContext::have_checked_for_unconstrained (HirId id, bool *result)
 }
 
 void
-TypeCheckContext::insert_resolved_predicate (HirId id,
-					     TyTy::TypeBoundPredicate predicate)
+TypeCheckContext::insert_resolved_predicate (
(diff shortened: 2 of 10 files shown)


powerboat9:
> likewiase

Lol. Also, you should really be doing `...: Likewise.` instead of `...: likewise`

gcc/rust/typecheck/rust-typecheck-context.cc
```diff
@@ -574,13 +574,13 @@ TypeCheckContext::have_checked_for_unconstrained (HirId id, bool *result)
...
-					     TyTy::TypeBoundPredicate predicate)
+TypeCheckContext::insert_resolved_predicate (
+  HirId id, const TyTy::TypeBoundPredicate &predicate)
 {
   // auto it = predicates.find (id);
   // rust_assert (it == predicates.end ());
 
-  predicates.insert ({id, predicate});
+  predicates.emplace (id, predicate);
 }
```
powerboat9, line 584: You should probably pass the predicate by-value (`TyTy::TypeBoundPredicate predicate`) so that `insert_resolved_predicate (id, std::move (predicate))` works. In that case, the emplace line would have to change to `predicated.emplace (id, std::move (predicate))`.

gcc/rust/typecheck/rust-hir-type-check-item.cc
powerboat9, line 1101: If you moved the `self->inherit_bound` call to just before the return, you could move from `specified_bound` instead of copying.

PR #4878, opened 2026-09-10
Attempt to clean up hir map

P-E-P:
Experimental draft PR to gather feedback

The Mapping class within rust-hir-map is slowly growing. A full clean up might be required in the future but in the meantime we may regroup related fields within specific classes.

I'm still unsure about the whole PtrMapping thing. I'd like to find a way to unify references and pointer mappings under the same class and then specialize it over the Id key.


gcc/rust/util/rust-hir-map.h
```diff
@@ -90,19 +100,157 @@ class Mappings
...
+class Mappings
+{
+public:
+  static Mappings &get ();
+  ~Mappings ();
+
+  CrateMappings crate_mapping;
+  ProcMacroMappings pmacro_mappings;
+  ASTMappings ast;
+  HIRMappings hir;
```
CohenArthur, line 238: I think it's completely fine to call these `crate` and `pmacro` instead of `crate_mapping` and `pmacro_mappings`, like you do for `ast` and `hir`. Even from an API point of view it makes sense to access `mappings.crate.lookup(blah)` or to do something like `auto &crate_mappings = Analysis::Mappings::get().crate;`. So I think be consistent and remove the `_mapping(s)` from all of these members

gcc/rust/util/rust-hir-map.h
```diff
@@ -90,19 +100,157 @@ class Mappings
...
+	  InsertionPolicy policy = InsertionPolicy::FORBID_DUPLICATES,
+	  InsertLocation insert_location = InsertLocation::NO>
+class PtrMapping
+{
+  std::unordered_map<HirId, Item *> storage;
+
+public:
+  tl::optional<Item *> lookup (HirId id);
+  void insert (Item *item);
+};
```
CohenArthur, line 197: I like this class and the way you use it a lot

gcc/rust/util/rust-hir-map.h
```diff
@@ -90,19 +100,157 @@ class Mappings
...
+  PtrMapping<HIR::GenericParam, InsertionPolicy::FORBID_DUPLICATES,
+	     InsertLocation::YES>
+    generic_param;
+  PtrMapping<HIR::Type> type;
+  PtrMapping<HIR::Stmt> stmt;
+  PtrMapping<HIR::FunctionParam, InsertionPolicy::ALLOW_DUPLICATES>
+    function_param;
+  PtrMapping<HIR::SelfParam> self_param;
+  PtrMapping<HIR::StructExprField> struct_expr_field;
+  PtrMapping<HIR::Pattern> pattern;
```
CohenArthur, line 226: I think all of these need to be plural, so `items` instead of `item`, `types` instead of `type` etc. We're talking about the types that we've mapped in the current mapping - it feels wrong to read `auto &whatever = mappings.hir.type.lookup(...)`

gcc/rust/util/rust-hir-map.h
```diff
@@ -90,19 +100,157 @@ class Mappings
...
+  std::map<CrateNum, std::vector<CustomDeriveProcMacro>> deriveMappings;
+  std::map<CrateNum, std::vector<BangProcMacro>> bangMappings;
+  std::map<CrateNum, std::vector<AttributeProcMacro>> attributeMappings;
+
+  std::map<NodeId, CustomDeriveProcMacro> procmacroDeriveMappings;
+  std::map<NodeId, BangProcMacro> procmacroBangMappings;
+  std::map<NodeId, AttributeProcMacro> procmacroAttributeMappings;
+  std::map<NodeId, CustomDeriveProcMacro> procmacroDeriveInvocations;
+  std::map<NodeId, BangProcMacro> procmacroBangInvocations;
+  std::map<NodeId, AttributeProcMacro> procmacroAttributeInvocations;
```
CohenArthur, line 122: you can remove `Mappings` from these, and also make sure the case conforms to the rest of the mappings since you are cleaning everything up. I would do `deriveMappings` -> `derives`, `bangMappings` -> `bangs` etc. and `procmacroAttributeMappings` -> `procmacro_attributes` and so on

gcc/rust/util/rust-hir-map.h
```diff
@@ -71,11 +71,21 @@ class NodeMapping
...
+class CrateMappings
 {
+  std::map<CrateNum, AST::Crate *> ast_crate_mappings;
+
+  // crate names
+  std::map<CrateNum, std::string> crate_names;
+
+  CrateNum crateNumItr;
+
+  CrateNum currentCrateNum;
```
CohenArthur, line 83: Change the case for these two as well

gcc/rust/util/rust-hir-map.h
```diff
@@ -71,11 +71,21 @@ class NodeMapping
   LocalDefId localDefId;
 };
 
-class Mappings
+class CrateMappings
 {
+  std::map<CrateNum, AST::Crate *> ast_crate_mappings;
```
CohenArthur, line 76:
```suggestion
  std::map<CrateNum, AST::Crate *> ast_crates;
```

gcc/rust/util/rust-hir-map.h
```diff
@@ -90,19 +100,157 @@ class Mappings
...
+public:
+  tl::optional<Item &> lookup (Id id)
+  {
+    auto it = storage.find (id);
+    if (it == storage.cend ())
+      return tl::nullopt;
+    return it->second;
+  }
+  void insert (Id id, Item item) { storage.insert ({id, item}); }
+};
```
CohenArthur, line 173: This is great

gcc/rust/util/rust-hir-map.h
P-E-P: I wish there was a way to unify it with the `Mapping` class, initially I wanted to do some mapping class and create two newtypes, one `HirMapping` which partially specialize to HirId and another `ASTMapping` which itself partially specialize to `NodeId`.

Issue #3868, opened 2025-07-04
Handle `#[rustc_legacy_const_generics(...)]`


CohenArthur:
I don't know if we really need to do anything other than stub it out: https://github.com/rust-lang/rust/issues/83167

philberty:
yeah this is super bizare tbh

```
#![feature(no_core, intrinsics, lang_items, staged_api)]
#![no_core]

#[lang = "sized"]
pub trait Sized {}

// #[rustc_legacy_const_generics(1)]
fn foo<const N: usize>(x: i32) -> i32 {
    x + N as i32
}

fn main() {
    let a = foo(10, 3);
}
```

actually means the argument in the foo call of 3 is the const generic argument

philberty:
chatgpt says:

```
Multiple const parameters are the interesting case:

#[rustc_legacy_const_generics(1, 3)]
fn foo<const N: usize, const M: usize>(x: i32, y: i32) {}

then:

foo(10, 3, 20, 4);

becomes:

foo::<3, 4>(10, 20);
```

PR #4889, opened 2026-09-11
gccrs: This is an utterly insane rustc shim needed for stdarch

philberty:
This doesnt fully implement the legacy const generics attribute but its the starting point.

What we actually need to do is understand:

    #[rustc_legacy_const_generics(1)]
    fn foo<const N: usize>(x: i32) -> i32 {
        x + N as i32
    }

    fn main() {
        let a = foo(10, 3);
    }

So you can see the call to foo has a 2nd param but its not a real 2nd param its meant to be turned into:

    foo::<3>(10)

Honestly this is just silly they probably implemented a bit of const generics wanted to use it in core and put this work around in to get away with it. Dumb.

Addresses #3868

gcc/rust/ChangeLog:

	* typecheck/rust-casts.cc (TypeCastRules::cast_rules): fix bad const cast
	* util/rust-attribute-values.h: add attribute
	* util/rust-attributes.cc: likewise

### DIFF
--- a/gcc/rust/typecheck/rust-casts.cc
+++ b/gcc/rust/typecheck/rust-casts.cc
@@ -18,6 +18,7 @@
 
 #include "rust-casts.h"
 #include "rust-tyty-util.h"
+#include "rust-tyty.h"
 
 namespace Rust {
 namespace Resolver {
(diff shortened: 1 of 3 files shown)


CohenArthur (approved):
code LGTM, just the issue number in the commit and it's all good

PR #4891, opened 2026-09-14
gccrs: Implement some platform-intrinsic simd function support

Polygonalr:
Only support basic binary operator functions at the moment, hunting down bugs in my code caused quite a wild goose chase


philberty (approved):
LGTM

Issue #4835, opened 2026-09-03
Implement `ExtractIf`/`extract_if` feature


CohenArthur:
We need to be able to handle the nightly `extract_if` feature for Rust-for-Linux, as it makes use of it 

https://doc.rust-lang.org/nightly/std/vec/struct.ExtractIf.html

One difficulty will be once again to "fake" this feature as it is not present in `core` 1.49. I'm not sure how we're going to inject an entire module into the compiler so we need to figure that out as well.

I don't think there is that much code that we can't handle so that should be okay

powerboat9:
I wonder if we couldn't just YOLO target rustc version bumps. Either that, or do minicore but the full 1.49 core extended with 1.85 core features.

powerboat9:
Misc footgun notes (up to 1.72, seems to be when `extract_if` was renamed from `drain_filter`):

- https://blog.rust-lang.org/2021/05/11/edition-2021/#intoiterator-for-arrays
- https://blog.rust-lang.org/2021/07/29/Rust-1.54.0/#attributes-can-invoke-function-like-macros
- https://blog.rust-lang.org/2022/01/13/Rust-1.58.0/#captured-identifiers-in-format-strings
- I think 1.59 removes `llvm_asm!`, also https://blog.rust-lang.org/2022/02/24/Rust-1.59.0/
- https://blog.rust-lang.org/2022/05/19/Rust-1.61.0/#custom-exit-codes-from-main
- https://blog.rust-lang.org/2022/06/30/Rust-1.62.0/#default-enum-variants
- https://blog.rust-lang.org/2022/08/11/Rust-1.63.0/#turbofish-for-generics-in-functions-with-impl-trait
- https://blog.rust-lang.org/2022/09/22/Rust-1.64.0/#c-compatible-ffi-types-in-core-and-alloc
- https://blog.rust-lang.org/2022/11/03/Rust-1.65.0/
- https://blog.rust-lang.org/2022/12/15/Rust-1.66.0/#explicit-discriminants-on-enums-with-fields
- https://blog.rust-lang.org/2023/01/26/Rust-1.67.0/#must-use-effective-on-async-fn
- https://blog.rust-lang.org/2023/03/09/Rust-1.68.0/#default-alloc-error-handler
- https://blog.rust-lang.org/2023/07/13/Rust-1.71.0/
[...]

PR #4924, opened 2026-09-20
gccrs: Initial implementation of hotpatching rustc-lib

Polygonalr:
ngl I feel icky about this implementation, making this PR to receive opinions on how to move #4835 forward, as it likely requires either backporting the feature to core, or bumping our target core version to 1.72, the latter being less than ideal. Patch file for #4835 is not done yet, just patching `CStr` as an example first.

I'm also not too sure about the stance of including .patch files in the gcc repo, since the root .gitignore ignores .patch files.


powerboat9:
You could try naming the file `0001-core-ffi-c-str.patch.keep`

CohenArthur:
or maybe something even clearer since this is just for the kernel like `0001-core-ffi-c-str.rfl-patch`? `0001-core-ffi-c-str.kernel-patch`?

CohenArthur (approved):
the Python code looks good to me but I'm really not good at Python :P for the README, I think it would be good to have a section on how to generate these patches and to re-iterate that these should only be present if they are *NEEDED* by the *KERNEL* to compile. this makes it clearer that we do intend to not patch core, but also want to start experimenting with the kernel and have to make some adjustments and concessions in that specific case

powerboat9:
I'm writing a script to update the vendored libraries you might want to integrate this with in some way. I'll have a PR up shortly.

Edit: #4934

Polygonalr:
@powerboat9 Hmmm not too sure what's there to change in my PR besides updating the patch files manually when we do a core version bump

